PR #10207, opened 2014-08-11
Use new people you may know and invite endpoints.

yipingliao:
Use new pymk and invite endpoints.

@2is10 Also includes small refactoring per your previous code review comments.
@leogrim

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/common/services/routeService.js
+++ b/kifi-backend/modules/shoebox/angular/src/common/services/routeService.js
@@ -54,8 +54,8 @@ angular.module('kifi.routeService', [])
       removeKeeps: route('/keeps/remove'),
       tagOrdering: route('/collections/ordering'),
       reorderTag: route('/collections/reorderTag'),
-      whoToInvite: route('/friends/wti'),
-      blockWtiConnection: route('/friends/wti/block'),
+      whoToInvite: route('/user/invite/recommended'),
+      blockWtiConnection: route('/user/invite/hide'),
       friends: function (page, pageSize) {
         return route('/user/friends') + '?page=' + page + '&pageSize=' + pageSize;
       },
@@ -65,7 +65,7 @@ angular.module('kifi.routeService', [])
       incomingFriendRequests: route('/user/incomingFriendRequests'),
       invite: route('/user/invite'),
       peopleYouMayKnow: function (page, pageSize) {
-        return route('/user/friends/find') + '?page=' + page + '&pageSize=' + pageSize;
+        return route('/user/friends/recommended') + '?page=' + page + '&pageSize=' + pageSize;
       },
       hideUserRecommendation: function (id) {
         return route('/user/' + id + '/hide');
--- a/kifi-backend/modules/shoebox/angular/src/friends/rightColFriendsView.js
+++ b/kifi-backend/modules/shoebox/angular/src/friends/rightColFriendsView.js
@@ -125,12 +125,8 @@ angular.module('kifi.friends.rightColFriendsView', [
               var name = '';
               var via = '';
 
-              if (person.name) {
-                name = person.name;
-                via = networkNamesMap[person.network];
-              } else if (person.network === 'email') {
-                name = person.identifier;  // Email address
-              }
+              name = person.name;
+              via = (person.network === 'email' && person.identifier) || networkNamesMap[person.network];
 
               peopleYouMayKnow.push({
                 networkType: person.network,
--- a/kifi-backend/modules/shoebox/angular/src/friends/rightColFriendsView.js
+++ b/kifi-backend/modules/shoebox/angular/src/friends/rightColFriendsView.js
@@ -122,20 +122,19 @@ angular.module('kifi.friends.rightColFriendsView', [
             var wtiList = wtiService.list;
 
             wtiList.forEach(function (person) {
-              var socialIdValues = person.fullSocialId.split('/');
               var name = '';
               var via = '';
 
               if (person.name) {
                 name = person.name;
-                via = networkNamesMap[socialIdValues[0]];
-              } else if (socialIdValues[0] === 'email') {
-                name = socialIdValues[1];
+                via = networkNamesMap[person.network];
+              } else if (person.network === 'email') {
+                name = person.identifier;  // Email address
               }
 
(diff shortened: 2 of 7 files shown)


kifi-backend/modules/shoebox/angular/src/layout/header/header.tpl.html
```diff
@@ -10,10 +10,10 @@
     <div class="sprite sprite-add-link" ng-click="addKeeps()">Add a link</div>
     <div class="kf-header-icon-hover-container">
       <a href="/friends" class="kf-header-icon-friends sprite sprite-friends">Friends</a>
-      <ul class="kf-dropdown-menu">
-        <li class="kf-dropdown-menu-item" ng-click="navigateToFriends()"><a class="kf-your-friends-header-link" href="javascript:" ng-class="{'active': isActive('/friends') }">Your Kifi Friends</a></li>
```
yipingliao, line 14: Removed kf-your-friends-header-link because it's not used.

kifi-backend/modules/shoebox/angular/src/common/services/routeService.js
```diff
@@ -54,8 +54,8 @@ angular.module('kifi.routeService', [])
       removeKeeps: route('/keeps/remove'),
       tagOrdering: route('/collections/ordering'),
       reorderTag: route('/collections/reorderTag'),
-      whoToInvite: route('/friends/wti'),
-      blockWtiConnection: route('/friends/wti/block'),
+      whoToInvite: route('/user/invite/recommended'),
+      blockWtiConnection: route('/user/invite/hide'),
```
2is10, line 58: In my mind, `/user/invite/...` has the same problem as `/user/friend/...` for who-to-invite routes: they’re “pre-cog” routes. The user hasn’t invited the people yet—just like we don’t know whether they’re actually friends.

Might consider `/user/people/recommended` and `/user/people/hide`. This is not critical to change for this pull request—just some feedback.

kifi-backend/modules/shoebox/angular/src/friends/rightColFriendsView.js
```diff
@@ -122,20 +122,19 @@ angular.module('kifi.friends.rightColFriendsView', [
...
               var via = '';
 
               if (person.name) {
                 name = person.name;
-                via = networkNamesMap[socialIdValues[0]];
-              } else if (socialIdValues[0] === 'email') {
-                name = socialIdValues[1];
+                via = networkNamesMap[person.network];
+              } else if (person.network === 'email') {
+                name = person.identifier;  // Email address
```
2is10: We often have a name associated with an email address (from a Gmail address book). In that case, can we display the person’s name (as with social network contacts) and the email address beneath, like:
Jared Jacobs
jared@kifi.com

The name could be omitted if it isn’t known.

kifi-backend/modules/shoebox/angular/src/invite/connectionCard.js
```diff
@@ -17,11 +17,11 @@ angular.module('kifi.invite.connectionCard', ['angularMoment'])
...
       var friend = scope.friend();
-      var network = friend.fullSocialId.split('/')[0];
-      var inNetworkId = friend.fullSocialId.split('/')[1];
+      var network = friend.network;
+      var inNetworkId = friend.identifier;
       var invited = (friend.lastInvitedAt != null);
 
-      if (friend.pictureUrl != null) {
+      if (friend.pictureUrl !== null) {
         scope.mainImage = friend.pictureUrl;
```
2is10, line 25: Probably safest to just check `if (friend.pictureUrl) {` since it needs to be a nonempty string to be valid.

kifi-backend/modules/shoebox/angular/src/friends/rightColFriendsView.js
yipingliao: Will do.

kifi-backend/modules/shoebox/angular/src/layout/header/header.tpl.html
```diff
@@ -10,10 +10,10 @@
...
-        <li class="kf-dropdown-menu-item" ng-click="navigateToInvite()"><a href="javascript:" ng-class="{'active': isActive('/invite') }">Invite a friend to Kifi</a></li>
-      </ul>
+      <div class="kf-dropdown-menu">
+        <a class="kf-dropdown-menu-item" ng-class="{'active': isActive('/friends')}" href="javascript:" ng-click="navigateToFriends()">Your Kifi Friends</a>
+        <a class="kf-dropdown-menu-item" ng-class="{'active': isActive('/invite')}" href="javascript:" ng-click="navigateToInvite()">Invite a friend to Kifi</a>
+      </div>
```
2is10: This `<div>` could be a `<menu>` (HTML 5) instead for more semantic markup.

kifi-backend/modules/shoebox/angular/src/friends/rightColFriendsView.js
leogrim: I have kept the existing semantics on the backend, that is when we don't have a name for an email address, we send the address itself in the name field. So you should always get a name. In general (FB / LinkedIn), the identifier field may not be human-readable.

2is10:
Thanks for making those changes!

lgtm when you’re happy with it

kifi-backend/modules/shoebox/angular/src/invite/connectionCard.js
yipingliao, line 25: Will do.

kifi-backend/modules/shoebox/angular/src/friends/rightColFriendsView.js
2is10: I don’t like that convention because it mislabels data, but it’s certainly not your fault. Fine to keep using that convention for now.

leogrim: I don't like it either. I made a new class for these recommendations (called InviteRecommendation), I'd be happy to tweak it to have something more explicit, e.g. make name an option.

yipingliao:
Made changes per code review. Tossed in tiny CSS fix (text alignment) per Jen's request.

2is10:
lgtm

PR #10208, opened 2014-08-11
Fixing double equals in tests

lawzlo:
@stkem @eishay @joshm1 

Those are all I found having double equals in tests. Actually not too many.


eishay:
Thanks! This is something important we should keep checking for.

PR #10218, opened 2014-08-11
Write unit tests for mutual friends modal.

yipingliao:
@2is10 Add some unit tests for the mutual friends modal last week.

### DIFF
--- a/kifi-backend/modules/shoebox/angular/test/unit/friends/seeMutualFriends.spec.js
+++ b/kifi-backend/modules/shoebox/angular/test/unit/friends/seeMutualFriends.spec.js
@@ -1,0 +1,179 @@
+'use strict';
+
+describe('kifi.friends.seeMutualFriends', function () {
+
+  beforeEach(module('kifi.friends.seeMutualFriends'));
+
+  var $q,
+      $compile,
+      $rootScope,
+      friendService,
+      inviteService,
+      savePymkService;
+
+  beforeEach(inject(function ($injector) {
+    $q = $injector.get('$q');
+    $compile = $injector.get('$compile');
+    $rootScope = $injector.get('$rootScope');
+    friendService = $injector.get('friendService');
+    inviteService = $injector.get('inviteService');
+    savePymkService = $injector.get('savePymkService');
+  }));
+
+  describe('kfSeeMutualFriends', function () {
+    var scope,
+        elem;
+
+    var testSavedPerson = null;
+    var savedPerson1 = {
+      id: '1a316f42-13be-4d86-a4a2-8c7efb3010b8',
+      fullName: 'Alexander Willis Schultz',
+      pictureUrl: '//djty7jcqog9qu.cloudfront.net/users/1a316f42-13be-4d86-a4a2-8c7efb3010b8/pics/200/xOMd7.jpg',
+      numMutualFriends: 2,
+      mutualFriends: [
+        {
+          id: '597e6c13-5093-4cba-8acc-93318987d8ee',
+          firstName: 'Stephen',
+          lastName: 'Kemmerling',
+          numFriends: 71,
+          pictureName: 'tEbqN.jpg'
+        },
+        {
+          id: 'ae139ae4-49ad-4026-b215-1ece236f1322',
+          firstName: 'Jen',
+          lastName: 'Granito Ruffner',
+          numFriends: 148,
+          pictureName: 'Upm5X.jpg'
+        }
+      ]
+    };
+    var savedPerson2 = {
+      id: 'fake-id-123',
+      fullName: 'John Smith',
+      pictureUrl: '//djty7jcqog9qu.cloudfront.net/users/fake-id-123/pics/200/0.jpg',
+      numMutualFriends: 2,
+      mutualFriends: [
+        {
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/angular/test/unit/friends/seeMutualFriends.spec.js
```diff
@@ -0,0 +1,179 @@
...
+
+    it('should reload when there is a new saved person', function () {
+      testSavedPerson = savedPerson1;
+      scope.$digest();
+      expect(elem.find('.kf-mutual-friends-pymk-name').text()).toBe('Alexander Willis Schultz');
+
+      testSavedPerson = savedPerson2;
+      scope.$digest();
+      expect(elem.find('.kf-mutual-friends-pymk-name').text()).toBe('John Smith');
+    });
```
2is10, line 110: The concept of a “saved person” exists only within this spec file, so making an assertion about it seems odd. Perhaps we could change this spec’s description to describe an actual use case that it corresponds to, or if there isn’t one, then remove this spec.

kifi-backend/modules/shoebox/angular/test/unit/friends/seeMutualFriends.spec.js
```diff
@@ -0,0 +1,179 @@
...
+
+      var deferred = $q.defer();
+      spyOn(inviteService, 'friendRequest').andReturn(deferred.promise);
+      elem.find('.kf-mutual-friends-action').click();
+      expect(elem.find('.kf-mutual-friends-action').hasClass('clickable')).toBe(false);
+
+      deferred.reject();
+      scope.$digest();
+      expect(elem.find('.kf-mutual-friends-action').hasClass('clickable')).toBe(true);
+    });
```
2is10, line 165: nice test case!

kifi-backend/modules/shoebox/angular/test/unit/friends/seeMutualFriends.spec.js
```diff
@@ -0,0 +1,179 @@
...
+      elem = $compile('<div kf-see-mutual-friends></div>')(scope);
+
+      spyOn(savePymkService, 'getSavedPersonYouMayKnow').andCallFake(function () {
+        return testSavedPerson;
+      });
+      spyOn(friendService, 'getPictureUrlForUser').andReturn('fake picture URL');
+    });
+
+    it('should have the correct pymk header', function () {
+      testSavedPerson = savedPerson1;
```
2is10, line 84: While it may feel repetitive, you might consider eliminating `testSavedPerson` and replacing these assignments with:

``` js
spyOn(savePymkService, 'getSavedPersonYouMayKnow').andReturn(savedPerson1);
```

It would make each test case feel more self-contained.

2is10:
Nice tests. Merge at will.

kifi-backend/modules/shoebox/angular/test/unit/friends/seeMutualFriends.spec.js
yipingliao, line 110: I've renamed this test to "should update modal content when saved person changes". The purpose of this test is to check that scope.$watch correctly updates the modal when we have a new saved person (when the user clicks on a different recommended friend's mutual friends link).

kifi-backend/modules/shoebox/angular/test/unit/friends/seeMutualFriends.spec.js
yipingliao, line 84: This was how I originally wrote the tests until I needed to test for scope.watch watching for a new saved person (discussed above). Calling spyOn a second time on the same function (to make it return savedPerson2 instead of savedPerson1) resulted in a Jasmine error (it won't let you make spy on functions twice). Hence using andCallFake with a function that returns a testSavedPerson that could be changed.

kifi-backend/modules/shoebox/angular/test/unit/friends/seeMutualFriends.spec.js
2is10, line 110: I see. I was confused because I thought “saved person” in the description was referring to `testSavedPerson`. I forgot about the saved person cache in the code. It sure adds complexity to expose that concept to the test. I’d really like to refactor the global modal to pass the person as a proper argument. Maybe we can look at together tomorrow.

kifi-backend/modules/shoebox/angular/test/unit/friends/seeMutualFriends.spec.js
2is10, line 84: I see. Another option would be to use the `.andReturn` pattern in most test cases and the `.andCallFake` pattern just in the test case(s) that require it. That would improve locality of concerns.

yipingliao, line 84: Okay, sounds good.

kifi-backend/modules/shoebox/angular/test/unit/friends/seeMutualFriends.spec.js
yipingliao, line 110: Indeed. It'd be great to visit that tomorrow. Thanks!

PR #10219, opened 2014-08-11
Correctly deal with merged uriId

stkem:
This fixes an issue where duplicate raw seed items were being created
whenever two uriIds where being merged in a migration. In the case of
items associated with a user this was causing a constraint violation in
the database (the other case should have caused one too, but
unfortunately MySQL has some limitations with constraints on nullable
columns)

Data will need to be reingested from scratch after this is merged
(takes about 90 minutes)

@lawzlo @yingjieMiao


stkem:
going to merge this to stop the exceptions in production. Please still give it a read when you get a chance.

PR #10221, opened 2014-08-11
remove (now unused) kifiHit from ShoeboxServiceClient

raymondng:
It's not really gone until it's gone from the service client ...

@yingjieMiao FYI


eishay:
:-)

PR #10229, opened 2014-08-12
final update of social refresh freq

joshm1:
all active social users have been updated to have their lastGraphRefresh set w/in the last 2 days

@stkem


2is10:
yay

PR #10231, opened 2014-08-12
get topic names for uri

yingjieMiao:
@stkem 

there is a simple threshold which we can tune. (empirically about 50% uris > threshold). If you think that is a problem, we can lower it a bit to x-percentile.


kifi-backend/modules/common/test/com/keepit/cortex/FakeCortexServiceClientImpl.scala
```diff
@@ -43,6 +43,7 @@ class FakeCortexServiceClientImpl(val airbrakeNotifier: AirbrakeNotifier) extend
   override def sampleURIsForTopic(topic: Int): Future[Seq[Id[NormalizedURI]]] = ???
   override def getSimilarUsers(userId: Id[User], topK: Int): Future[(Seq[Id[User]], Seq[Float])] = ???
   override def unamedTopics(limit: Int = 20): Future[(Seq[LDAInfo], Seq[Map[String, Float]])] = ???
+  override def getTopicNames(uris: Seq[Id[NormalizedURI]]): Future[Seq[Option[String]]] = ???
```
stkem: This is going to start being called by the curator tests, so will need some fake data. Just a `Seq.fill(uris.length, None)` should do.

stkem:
50% is good to start with, I think. Made one comment, lgtm otherwise.

yingjieMiao:
@stkem  updated and merge.

PR #10234, opened 2014-08-12
getHelpRankInfo => heimdal

raymondng:
Pretty much copy-and-paste for this one
Sorry no optimization yet 

@stkem let's move curator to this endpoint soon so we can remove another dependency on shoebox

@yingjieMiao @eishay


stkem:
lgtm

PR #10236, opened 2014-08-12
Multiple small changes

lawzlo:
check uriscore is empty, change vote API from POST to GET, add new test case for keep discoverability

@stkem @joshm1

### DIFF
--- a/kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
+++ b/kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
@@ -84,7 +84,9 @@ class RecommendationGenerationCommander @Inject() (
           userId = reco.userId,
           uriId = reco.uriId,
           {
-            val score = scoreCoefficients.recencyScore.getOrElse(defaultScore) * reco.allScores.recencyScore +
+            if (scoreCoefficients.isEmpty)
+              computeMasterScore(reco.allScores)
+            else scoreCoefficients.recencyScore.getOrElse(defaultScore) * reco.allScores.recencyScore +
               scoreCoefficients.overallInterestScore.getOrElse(defaultScore) * reco.allScores.overallInterestScore +
               scoreCoefficients.priorScore.getOrElse(defaultScore) * reco.allScores.priorScore +
               scoreCoefficients.socialScore.getOrElse(defaultScore) * reco.allScores.socialScore +
@@ -92,7 +94,6 @@ class RecommendationGenerationCommander @Inject() (
               scoreCoefficients.recentInterestScore.getOrElse(defaultScore) * reco.allScores.recentInterestScore +
               scoreCoefficients.rekeepScore.getOrElse(defaultScore) * reco.allScores.rekeepScore +
               scoreCoefficients.discoveryScore.getOrElse(defaultScore) * reco.allScores.discoveryScore
-            if (score != 0.0f) score else computeMasterScore(reco.allScores)
           },
           explain = Some(reco.allScores.toString)
         )
--- a/kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
+++ b/kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
@@ -63,14 +63,14 @@ class RecommendationGenerationCommander @Inject() (
   ).map(Id[User](_)) //will go away once we release, just saving some computation/time for now
 
   private def computeMasterScore(scores: UriScores): Float = {
-    7 * scores.socialScore +
-      5 * scores.overallInterestScore +
-      1 * scores.priorScore +
+    5 * scores.socialScore +
+      6 * scores.overallInterestScore +
+      2 * scores.priorScore +
       1 * scores.recencyScore +
       1 * scores.popularityScore +
-      5 * scores.recentInterestScore +
+      9 * scores.recentInterestScore +
       6 * scores.rekeepScore +
-      2 * scores.discoveryScore
+      3 * scores.discoveryScore
   }
 
   def getAdHocRecommendations(userId: Id[User], howManyMax: Int, scoreCoefficients: UriRecommendationScores): Future[Seq[RecommendationInfo]] = {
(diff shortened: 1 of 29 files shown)


kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
```diff
@@ -84,7 +84,9 @@ class RecommendationGenerationCommander @Inject() (
           userId = reco.userId,
           uriId = reco.uriId,
           {
-            val score = scoreCoefficients.recencyScore.getOrElse(defaultScore) * reco.allScores.recencyScore +
+            if (scoreCoefficients.isEmpty)
```
eishay: if we have a multi line if/else lets have explicit blocks

lawzlo: Cool. Will add it.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/RecommendationsController.scala
```diff
@@ -28,13 +28,27 @@ class RecommendationsController @Inject() (
     }
   }
 
-  def updateUriRecommendationFeedback(userId: Id[User], uriId: Id[NormalizedURI]) = JsonAction.authenticatedParseJsonAsync { request =>
+  def updateUriRecommendationFeedback(userId: Id[User], url: String) = JsonAction.authenticatedParseJsonAsync { request =>
     val feedback = request.body.as[UriRecommendationFeedback]
-    commander.updateUriRecommendationFeedback(userId, uriId, feedback).map(fkis => Ok(Json.toJson(fkis)))
+    val uriOpt = db.readOnlyMaster { implicit s =>
```
stkem: Database access should generally only occur in commanders and related classes, not controllers (there are some exceptions). Since there is already a commander to go with this controller I would move that logic there.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/RecommendationsController.scala
```diff
@@ -3,21 +3,21 @@ package com.keepit.controllers.website
...
 import scala.concurrent.Future
-
 import com.google.inject.Inject
 
 class RecommendationsController @Inject() (
     actionAuthenticator: ActionAuthenticator,
     commander: RecommendationsCommander,
-    userExperimentCommander: LocalUserExperimentCommander) extends WebsiteController(actionAuthenticator) with ShoeboxServiceController {
+    userExperimentCommander: LocalUserExperimentCommander,
+    //normalizedURIRepo: NormalizedURIRepo,
```
stkem: please don't leave commented out dependencies in the code

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/RecommendationsController.scala
lawzlo: Got example from `ShoeboxController`. Will move the logic to the commander.

stkem: Yeah, we still have a bunch of places in the code that do this. Most was written before we established the commander pattern.

lawzlo: Ah, I see. Will avoid it in the future.

kifi-backend/modules/shoebox/conf/shoebox.routes
```diff
@@ -273,7 +273,8 @@ POST    /site/libraries/move        @com.keepit.controllers.website.LibraryContr
 
 GET     /site/recos/adHoc           @com.keepit.controllers.website.RecommendationsController.adHocRecos(n: Int)
 POST    /site/recos/adHoc           @com.keepit.controllers.website.RecommendationsController.adHocRecos(n: Int)
-POST    /site/recos/updateUriRecommendationFeedback     @com.keepit.controllers.website.RecommendationsController.updateUriRecommendationFeedback(userId: Id[User], uriId: Id[NormalizedURI])
+POST    /site/recos/updateUriRecommendationFeedback     @com.keepit.controllers.website.RecommendationsController.updateUriRecommendationFeedback(userId: Id[User], url: String)
+GET     /site/recos/updateUriRecommendationUserInteraction     @com.keepit.controllers.website.RecommendationsController.updateUriRecommendationUserInteraction(userId: Id[User], url: String, vote: Option[Boolean])
```
2is10: I don’t know what this action does, but I see that it’s named "update...", and generally any kind of update should be a `POST`.

From [HTTP 1.1 Method Definitions](http://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html):

> the GET and HEAD methods SHOULD NOT have the significance of taking an action other than retrieval

stkem: It does indeed update something. Unfortunately it will be triggered from a link in an email (it will let you upvote/downvote recommendations in the internal recommendations email @joshm1 is working on) hence the GET request.

2is10: Unlike with our internal routes, where our convention is to use the controller action name as the route name, we use [RESTful](http://en.wikipedia.org/wiki/Representational_state_transfer#Applied_to_web_services) or “[clean URL](http://en.wikipedia.org/wiki/Clean_URL#Structure)” principles in choosing our public routes (e.g. those under `/site/` and `/ext/` and `/m/`).

Notice that most routes above and below use single words in each path segment, generally nouns. If the route is an action (a POST), the last path segment can optionally be a verb to clarify what the action is.

I don’t know what these specific recommendation-related actions do, but I’d be happy to help you choose cleaner routes. Maybe:

```
POST   /site/recos/uri/feedback     ...
POST   /site/recos/uri/interaction  ...
```

2is10: @stkem Emails can contain forms, which can send POST requests.
http://css-tricks.com/html-forms-in-html-emails/

lawzlo: That's really good to learn. So I guess in the case we still use POST request? What do you think? @joshm1

2is10: Also, the `/site/` prefix is for AJAX requests coming from the Angular app. The section heading `Web API` could probably be made clearer.
We should probably consider adopting a new convention for kifi.com actions from emails. Right now it looks like they’re mostly near the top of shoebox.routes (e.g. `/extmsg/email/mute` and `/unsubscribe/*token`) and their controllers are named using the convention `Email...Controller`.

stkem: Re POST from email: Cool, I wasn't aware that was possible. In that case we should definitely go with POST (sorry for the back and forth @lawzlo, I should have checked this). I like @2is10s route structure suggestion.
Re /site : This endpoint is in fact meant for the site, we are temporarily putting a call in the internal emails to test things out. If we end up keeping a feedback mechanism in the final email we'll make some separate mechanism.

lawzlo: No, thanks for considering change to GET, so that we learnt good things. I will change back to POST and rename routes.

joshm1: cool, I also didn't realize we  could send POST from email.

PR #10239, opened 2014-08-12
feed attribution: topic

yingjieMiao:
@stkem 

plugged in the api. not saving anything yet.


kifi-backend/modules/curator/app/com/keepit/curator/commanders/AttributionHelper.scala
```diff
@@ -0,0 +1,59 @@
...
+
+import scala.concurrent.Future
+
+@json case class UserAttribution(friends: Seq[Id[User]], others: Int)
+@json case class KeepAttribution(keeps: Seq[Id[Keep]])
+@json case class TopicAttribution(topicName: String)
+@json case class SeedAttribution(user: Option[UserAttribution], keep: Option[KeepAttribution], topic: Option[TopicAttribution])
+
+@ImplementedBy(classOf[SeedAttributionHelperImpl])
+trait SeedAttributionHelper {
```
stkem: Why a separate trait and class? Are you anticipating another version of this?

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
```diff
@@ -149,34 +150,47 @@ class RecommendationGenerationCommander @Inject() (
...
+            val attributionFut = toBeSavedItemsFut.flatMap { items => attributionHelper.getSeedAttribution(items) }
+
+            for {
+              toBeSavedItems <- toBeSavedItemsFut
+              attribution <- attributionFut
+            } yield {
+              db.readWrite { implicit s =>
+                (toBeSavedItems zip attribution) foreach {
+                  case (item, attr) =>
+                    // attr will be saved soon -- Yingjie
```
stkem: This isn't really such a big pull request yet. It would be good to see all the changes in one go, i.e. with saving and all. Otherwise it's a bit hard to say if intermediate types and such make sense.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/AttributionHelper.scala
```diff
@@ -0,0 +1,59 @@
...
+    }
+  }
+
+  def getUserAttribution(seeds: Seq[ScoredSeedItem]): Future[Seq[Option[UserAttribution]]]
+  def getKeepAttribution(seeds: Seq[ScoredSeedItem]): Future[Seq[Option[KeepAttribution]]]
+  def getTopicAttribution(seeds: Seq[ScoredSeedItem]): Future[Seq[Option[TopicAttribution]]]
+}
+
+@Singleton
+class SeedAttributionHelperImpl @Inject() (cortex: CortexServiceClient) extends SeedAttributionHelper {
```
stkem: A pattern used by all the other Helpers is to make them callable (i.e. with an `apply` method) and have everything else be private, thus making it possible to treat them a bit like a black box. In particular the caller doesn't need to know that there are different kinds of attributions or how they are handled.
In this case in particular I think it would help understandability of the code as a whole to follow the same pattern as the `UriScoringHelper`, i.e. take a `Seq[ScoredSeedItem]` and return an augmented version.

yingjieMiao: I could hide different attributions as private methods. `ScoredSeedItem` is not really an 'augmented' version of 'SeedItem'. The current method is exactly the same as `UriScoringHelper`.

stkem:
lgtm

PR #10241, opened 2014-08-12
switching help rank calls from shoebox to heimdal

stkem:
also 
increasing processing concurrency a bit

@raymondng @yingjieMiao @lawzlo

### DIFF
--- a/kifi-backend/modules/common/test/com/keepit/heimdal/FakeHeimdalServiceClientImpl.scala
+++ b/kifi-backend/modules/common/test/com/keepit/heimdal/FakeHeimdalServiceClientImpl.scala
@@ -59,7 +59,9 @@ class FakeHeimdalServiceClientImpl(val airbrakeNotifier: AirbrakeNotifier) exten
 
   def getKeepAttributionInfo(userId: Id[User]): Future[UserKeepAttributionInfo] = Future.successful(UserKeepAttributionInfo(Id[User](1), 0, 0, 0, 0, 0))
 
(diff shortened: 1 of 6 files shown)


kifi-backend/modules/common/test/com/keepit/heimdal/FakeHeimdalServiceClientImpl.scala
```diff
@@ -59,7 +59,9 @@ class FakeHeimdalServiceClientImpl(val airbrakeNotifier: AirbrakeNotifier) exten
 
   def getKeepAttributionInfo(userId: Id[User]): Future[UserKeepAttributionInfo] = Future.successful(UserKeepAttributionInfo(Id[User](1), 0, 0, 0, 0, 0))
 
-  def getHelpRankInfos(uriIds: Seq[Id[NormalizedURI]]): Future[Seq[HelpRankInfo]] = Future.successful(Seq.empty)
+  def getHelpRankInfos(uriIds: Seq[Id[NormalizedURI]]): Future[Seq[HelpRankInfo]] = Future.successful {
```
stkem, line 62: @raymondng when making a new endpoint please also have the fake version honor the contract

raymondng:
lgtm

kifi-backend/modules/common/test/com/keepit/heimdal/FakeHeimdalServiceClientImpl.scala
raymondng, line 62: sounds good -- already working on a more full fledged fake client

PR #10242, opened 2014-08-12
Rec discoverability should also be false when keep state is not ACTIVE + Unit tests

lawzlo:
@stkem

### DIFF
--- a/kifi-backend/modules/curator/app/com/keepit/curator/commanders/AllKeepsSeedIngestionHelper.scala
+++ b/kifi-backend/modules/curator/app/com/keepit/curator/commanders/AllKeepsSeedIngestionHelper.scala
@@ -104,13 +104,18 @@ class AllKeepSeedIngestionHelper @Inject() (
       ))
 
       val discoverable = {
-        {
-          if (keepInfo.discoverable && keep.isPrivate) keepInfoRepo.checkDiscoverableByUriId(keep.uriId)
-          else keepInfo.discoverable || !keep.isPrivate
-        } && {
-          if (keepInfo.state == CuratorKeepInfoStates.ACTIVE && keep.state != KeepStates.ACTIVE)
-            keepInfoRepo.checkActiveByUriId(keep.uriId)
-          else keepInfo.state == CuratorKeepInfoStates.ACTIVE || keep.state == KeepStates.ACTIVE
+        if (!rawSeedItems(0).discoverable) {
+          !keep.isPrivate && keep.state == KeepStates.ACTIVE
+        } else {
+          if (!keep.isPrivate && keep.state == KeepStates.ACTIVE) {
+            true
+          } else {
+            if (!keep.isPrivate != keepInfo.discoverable || keep.state.value != keepInfo.state.value) {
+              keepInfoRepo.checkDiscoverableByUriId(keep.uriId)
+            } else {
+              true
+            }
+          }
         }
       }
 
--- a/kifi-backend/modules/curator/app/com/keepit/curator/commanders/AllKeepsSeedIngestionHelper.scala
+++ b/kifi-backend/modules/curator/app/com/keepit/curator/commanders/AllKeepsSeedIngestionHelper.scala
@@ -100,11 +100,19 @@ class AllKeepSeedIngestionHelper @Inject() (
         uriId = keep.uriId,
         userId = keep.userId,
         state = State[CuratorKeepInfo](keep.state.value),
-        discoverable = !keep.isPrivate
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/curator/app/com/keepit/curator/commanders/AllKeepsSeedIngestionHelper.scala
```diff
@@ -116,7 +124,7 @@ class AllKeepSeedIngestionHelper @Inject() (
         userId = keep.userId,
         keepId = keep.id.get,
         state = State[CuratorKeepInfo](keep.state.value),
-        discoverable = !keep.isPrivate
+        discoverable = !keep.isPrivate && keep.state == KeepStates.ACTIVE
```
stkem: This is the discoverability of the keep, which should not be affected by wether it's active or not. Only the RawSeedItem needs to take that into account.

lawzlo: Semantically make sense. Will change it.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/AllKeepsSeedIngestionHelper.scala
```diff
@@ -103,8 +103,16 @@ class AllKeepSeedIngestionHelper @Inject() (
         discoverable = !keep.isPrivate
       ))
 
-      val discoverable = if (keepInfo.discoverable && keep.isPrivate) keepInfoRepo.checkDiscoverableByUriId(keep.uriId)
-      else (keepInfo.discoverable || (!keep.isPrivate && keep.state == KeepStates.ACTIVE))
+      val discoverable = {
```
stkem, line 106: I'm not sure I'm following this. Why not `val discoverable = if (keepInfo.discoverable && (((keep.isPrivate && keepInfo.discoverable) || (keep.state!=KeepStates.ACTIVE` && keepInfo.state==ACTIVE) keepInfoRepo.checkDiscoverableByUriId(keep.uriId)`
Still complicated, but then you don't need the separate db method.

stkem, line 106: I.e. if it's discoverable and either changed from public to private or from active to inactive/duplicate, check if we need to make the raw seed item private as well.

kifi-backend/modules/curator/app/com/keepit/curator/model/CuratorKeepInfoRepo.scala
```diff
@@ -49,7 +50,11 @@ class CuratorKeepInfoRepoImpl @Inject() (
   }
 
   def checkDiscoverableByUriId(uriId: Id[NormalizedURI])(implicit session: RSession): Boolean = {
-    (for (row <- rows if row.uriId === uriId && row.state === CuratorKeepInfoStates.ACTIVE && row.discoverable) yield row.id).firstOption.isDefined
+    (for (row <- rows if row.uriId === uriId && row.discoverable) yield row.id).firstOption.isDefined
```
stkem: Inactive items should no be considered as discoverable

lawzlo: I removed check ACTIVE here because similar semantic for Keep. Keep discoverable should not be affected by state. We check state separately for rawSeedItem I think?

stkem: Yes, that is correct, but this method check the discoverability of a uri, not of a particular keep, in order to determine what to do for a raw see ditem.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/AllKeepsSeedIngestionHelper.scala
lawzlo, line 106: updating as discussed.

kifi-backend/modules/curator/app/com/keepit/curator/model/CuratorKeepInfoRepo.scala
lawzlo: updating as discussed.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/AllKeepsSeedIngestionHelper.scala
stkem, line 106: :+1:

PR #10243, opened 2014-08-12
track feed recommendations sent to users in emails

joshm1:
@stkem @lawzlo 

also factored out more test setup code into CuratorTestHelpers.  there are factory methods for creating UriRecommendation, NormalizedUri, UriSummary, Users, Keeps

### DIFF
--- a/kifi-backend/modules/curator/app/com/keepit/curator/commanders/CuratorTasksPlugin.scala
+++ b/kifi-backend/modules/curator/app/com/keepit/curator/commanders/CuratorTasksPlugin.scala
@@ -31,7 +31,8 @@ class CuratorTasksPlugin @Inject() (
   }
 
   private def scheduleRecommendationEmail(): Unit = {
-    log.info("scheduleRecommendationEmail() called")
-    cronTaskOnLeader(quartz, emailActor.ref, "0 45 23 * * ?", EngagementEmailTypes.FEED)
+    // <sec> <min> <hr> <day of mo> <mo> <day of wk> <yr>
+    val cronTime = "0 0 14 * * ?" // 2pm UTC == 7am PDT
+    cronTaskOnLeader(quartz, emailActor.ref, cronTime, EngagementEmailTypes.FEED)
   }
 }
--- a/kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
+++ b/kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
@@ -73,12 +73,14 @@ class RecommendationGenerationCommander @Inject() (
       3 * scores.discoveryScore
   }
 
-  def getAdHocRecommendations(userId: Id[User], howManyMax: Int, scoreCoefficients: UriRecommendationScores): Future[Seq[RecommendationInfo]] = {
-    val recosFuture = db.readOnlyReplicaAsync { implicit session =>
+  def getTopRecommendations(userId: Id[User], howManyMax: Int): Future[Seq[UriRecommendation]] = {
+    db.readOnlyReplicaAsync { implicit session =>
       uriRecRepo.getByTopMasterScore(userId, Math.max(howManyMax, 1000))
     }
+  }
 
-    recosFuture.map { recos =>
+  def getAdHocRecommendations(userId: Id[User], howManyMax: Int, scoreCoefficients: UriRecommendationScores): Future[Seq[RecommendationInfo]] = {
+    getTopRecommendations(userId, howManyMax).map { recos =>
       recos.map { reco =>
         RecommendationInfo(
           userId = reco.userId,
@@ -98,7 +100,6 @@ class RecommendationGenerationCommander @Inject() (
         )
       }.sortBy(-1 * _.score).take(howManyMax)
     }
-
   }
 
   private def getPerUserGenerationLock(userId: Id[User]): ReactiveLock = {
--- a/kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
+++ b/kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
@@ -75,12 +75,12 @@ class RecommendationGenerationCommander @Inject() (
 
   def getTopRecommendations(userId: Id[User], howManyMax: Int): Future[Seq[UriRecommendation]] = {
     db.readOnlyReplicaAsync { implicit session =>
-      uriRecRepo.getByTopMasterScore(userId, Math.max(howManyMax, 1000))
+      uriRecRepo.getByTopMasterScore(userId, howManyMax)
     }
   }
 
   def getAdHocRecommendations(userId: Id[User], howManyMax: Int, scoreCoefficients: UriRecommendationScores): Future[Seq[RecommendationInfo]] = {
-    getTopRecommendations(userId, howManyMax).map { recos =>
+    getTopRecommendations(userId, Math.max(howManyMax, 1000)).map { recos =>
       recos.map { reco =>
         RecommendationInfo(
           userId = reco.userId,
--- a/kifi-backend/modules/curator/app/com/keepit/curator/commanders/email/EngagementFeedEmailSender.scala
+++ b/kifi-backend/modules/curator/app/com/keepit/curator/commanders/email/EngagementFeedEmailSender.scala
@@ -2,27 +2,30 @@ package com.keepit.curator.commanders.email
 
 import com.google.inject.{ ImplementedBy, Inject }
 import com.keepit.commanders.RemoteUserExperimentCommander
+import com.keepit.common.db.slick.Database
 import com.keepit.common.db.{ Id, LargeString }
 import com.keepit.common.healthcheck.AirbrakeNotifier
 import com.keepit.common.logging.Logging
 import com.keepit.common.mail.{ SystemEmailAddress, ElectronicMail }
 import com.keepit.curator.commanders.RecommendationGenerationCommander
-import com.keepit.curator.model.RecommendationInfo
+import com.keepit.curator.model.{ UriRecommendationRepo, UriRecommendation, RecommendationInfo }
 import com.keepit.model.{ URISummary, NormalizedURI, User, UriRecommendationScores, NotificationCategory, ExperimentType }
 import com.keepit.shoebox.ShoeboxServiceClient
 import com.keepit.common.time.Clock
 import org.joda.time.DateTimeZone
+import play.api
+import play.api.db
 import play.api.libs.concurrent.Execution.Implicits.defaultContext
 
 import concurrent.Future
 
-case class RecommendedUriSummary(reco: RecommendationInfo, uri: NormalizedURI, uriSummary: URISummary) {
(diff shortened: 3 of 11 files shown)


joshm1:
logic to not send keeps that've been sent before is coming next

kifi-backend/modules/curator/app/com/keepit/curator/commanders/CuratorTasksPlugin.scala
```diff
@@ -31,7 +31,8 @@ class CuratorTasksPlugin @Inject() (
   }
 
   private def scheduleRecommendationEmail(): Unit = {
-    log.info("scheduleRecommendationEmail() called")
-    cronTaskOnLeader(quartz, emailActor.ref, "0 45 23 * * ?", EngagementEmailTypes.FEED)
+    // <sec> <min> <hr> <day of mo> <mo> <day of wk> <yr>
+    val cronTime = "0 0 14 * * ?" // 2pm UTC == 7am PDT
```
eishay, line 35: good to have the documentation, its confusing. thanks!

kifi-backend/modules/curator/app/com/keepit/curator/commanders/email/EngagementFeedEmailSender.scala
```diff
@@ -68,19 +73,28 @@ class EngagementFeedEmailSenderImpl @Inject() (
...
+            subject = "Your Recommended Links from friends on Kifi",
             htmlBody = htmlBody,
             textBody = textBody,
             to = Seq(user.primaryEmail.get),
             from = SystemEmailAddress.ENG
           )
+
           log.info(s"sending email to ${user.id.get} with ${feedData.size} keeps")
-          shoebox.sendMail(email).map { EngagementFeedSummary(user.id.get, _, feedData) }
+          lazy val now = clock.now()(DateTimeZone.UTC)
```
eishay: if you import `com.keepit.common.time.zones.UTC` you can use it here as is and it's going to be future safe if we ever change the time library to something other then joda.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/email/EngagementFeedEmailSender.scala
```diff
@@ -68,19 +73,28 @@ class EngagementFeedEmailSenderImpl @Inject() (
...
+          shoebox.sendMail(email).map { sent =>
```
eishay, line 86: you don't need a `map` here, right? maybe `foreach`?
maybe actually do a 

``` scala
 shoebox.sendMail(email).filter(_.sent).foreach {
  db.readWrite { implicit rw =>
    ...
  }
}
```

kifi-backend/modules/curator/app/com/keepit/curator/model/UriRecommendation.scala
```diff
@@ -18,10 +18,12 @@ case class UriRecommendation(
     allScores: UriScores,
     seen: Boolean,
     clicked: Boolean,
-    kept: Boolean) extends Model[UriRecommendation] with ModelWithPublicId[UriRecommendation] with ModelWithState[UriRecommendation] {
+    kept: Boolean,
+    lastPushedAt: Option[DateTime] = None) extends Model[UriRecommendation] with ModelWithPublicId[UriRecommendation] with ModelWithState[UriRecommendation] {
 
   def withId(id: Id[UriRecommendation]): UriRecommendation = this.copy(id = Some(id))
   def withUpdateTime(updateTime: DateTime): UriRecommendation = this.copy(updateAt = updateTime)
+  def withLastPushedAt(pushedAt: Option[DateTime]): UriRecommendation = this.copy(lastPushedAt = pushedAt)
```
eishay: are you ok with having `pushedAt` be `None`?
would make sense to have a 

``` scala
def withLastPushedAt(pushedAt: DateTime): UriRecommendation = this.copy(lastPushedAt = Some(pushedAt))
```

kifi-backend/modules/curator/app/com/keepit/curator/commanders/email/EngagementFeedEmailSender.scala
eishay: Actually, UTC is the default time zone. Why should we be explicit about it?

kifi-backend/modules/curator/app/com/keepit/curator/model/UriRecommendationRepo.scala
```diff
@@ -43,7 +44,9 @@ class UriRecommendationRepoImpl @Inject() (
     def seen = column[Boolean]("seen", O.NotNull)
     def clicked = column[Boolean]("clicked", O.NotNull)
     def kept = column[Boolean]("kept", O.NotNull)
-    def * = (id.?, createdAt, updatedAt, state, vote.?, uriId, userId, masterScore, allScores, seen, clicked, kept) <> ((UriRecommendation.apply _).tupled, UriRecommendation.unapply _)
+    def lastPushedAt = column[DateTime]("last_pushed_at", O.Nullable)
```
eishay, line 47: not using lastPushedAt yet in this pull?

kifi-backend/modules/curator/app/com/keepit/curator/commanders/email/EngagementFeedEmailSender.scala
joshm1, line 86: the return value from the expression is used. the method returns `Future[EngagementFeedSummary]` which is created here

eishay, line 86: oh, missed it. thanks

kifi-backend/modules/curator/app/com/keepit/curator/commanders/email/EngagementFeedEmailSender.scala
joshm1: I didn't realize there was a time zone val defined in `common.time`, which is why I used this.  the clock.now() method doesn't have a default time zone unless the implicit is in scope, it would throw this error:

`could not find implicit value for parameter zone: org.joda.time.DateTimeZone`

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
```diff
@@ -73,12 +73,14 @@ class RecommendationGenerationCommander @Inject() (
       3 * scores.discoveryScore
   }
 
-  def getAdHocRecommendations(userId: Id[User], howManyMax: Int, scoreCoefficients: UriRecommendationScores): Future[Seq[RecommendationInfo]] = {
-    val recosFuture = db.readOnlyReplicaAsync { implicit session =>
+  def getTopRecommendations(userId: Id[User], howManyMax: Int): Future[Seq[UriRecommendation]] = {
+    db.readOnlyReplicaAsync { implicit session =>
       uriRecRepo.getByTopMasterScore(userId, Math.max(howManyMax, 1000))
```
stkem: I think this one should honor the `howManyMax` exactly....

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
```diff
@@ -73,12 +73,14 @@ class RecommendationGenerationCommander @Inject() (
...
     }
+  }
 
-    recosFuture.map { recos =>
+  def getAdHocRecommendations(userId: Id[User], howManyMax: Int, scoreCoefficients: UriRecommendationScores): Future[Seq[RecommendationInfo]] = {
+    getTopRecommendations(userId, howManyMax).map { recos =>
```
stkem: ....and the `Math.max(howManyMax, 1000)` should go here.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/email/EngagementFeedEmailSender.scala
stkem: We have a method in the time package called `currentDateTime` which does all this for you (including getting an implicit time zone), that way we never have the time zone in the code.

kifi-backend/modules/curator/app/com/keepit/curator/model/UriRecommendation.scala
stkem: +1. Makes it impossible to go from `Some` back to `None` which is probably desired.

kifi-backend/modules/curator/test/com/keepit/curator/CuratorTestHelpers.scala
```diff
@@ -1,15 +1,19 @@
...
+import com.keepit.common.db.slick.DBSession.RWSession
+import com.keepit.common.db.slick.Database
 import com.keepit.common.mail.EmailAddress
 import com.keepit.model._
-import com.keepit.shoebox.FakeShoeboxServiceClientImpl
-import model.{ UriRecommendationStates, UriScores, UriRecommendation }
+import com.keepit.shoebox.{ ShoeboxServiceClient, ShoeboxScraperClient, FakeShoeboxServiceClientImpl }
+import model.{ UriRecommendationRepo, UriRecommendationStates, UriScores, UriRecommendation }
 
-import scalaz.IndexedStateT
```
stkem, line 10: wah!

kifi-backend/modules/curator/app/com/keepit/curator/commanders/email/EngagementFeedEmailSender.scala
eishay: also, if you import `time._` you'll get `DEFAULT_DATE_TIME_ZONE` which is `UTC`

kifi-backend/modules/curator/app/com/keepit/curator/model/UriRecommendation.scala
joshm1: I kept it as an Option incase we wanted to use the method to nullify the value as @stkem said

kifi-backend/modules/curator/app/com/keepit/curator/commanders/email/EngagementFeedEmailSender.scala
stkem: which you can also import explicitly, that way you have an explicit implicit :-) (This sounds funny, but I'm not actually kidding)

kifi-backend/modules/curator/app/com/keepit/curator/model/UriRecommendationRepo.scala
joshm1, line 47: not yet. I'm working on the logic now to filter out the keeps that have this value

kifi-backend/modules/curator/app/com/keepit/curator/commanders/email/EngagementFeedEmailSender.scala
stkem: Also, it's won't make much of a difference here, but `lazy`s actually have a significant overhead, i.e. they only make sense performance wise if actually getting the value is expensive and it's rarely needed.

kifi-backend/modules/curator/app/com/keepit/curator/model/UriRecommendation.scala
eishay: i would add an explicit `withNoLastPushedAt()` method for that for explicitness and bugfreeness

PR #10245, opened 2014-08-12
Make GraphServiceClient a Singleton

leogrim:
@stkem


yingjieMiao:
@stkem @leogrim 
revert this? build fails

eishay:
if it was because of a merge where the engineer who merged is not here -
revert

PR #10251, opened 2014-08-13
New Angular asset router

andrewconner:
@stkem Would appreciate a thorough review. Of note, check out `KifiSiteRouter.scala` and `shoebox.routes`. `KifiSiteRouter` will likely grow, so analysis of the proposed pattern is helpful as well.


2is10:
Great not to have to maintain every Angular path in shoebox.routes.

stkem:
lgtm, modulo merge conflict

PR #10252, opened 2014-08-13
moving mac.css injection from api.js which sometimes runs before doc ready

2is10:


2is10:
api.js is sometimes injected at doc start via manifest into hidden, preloading tabs in Chrome, which can open ports but can’t have scripts or styles dynamically injected (using `chrome.tabs.executeScript` or `chrome.tabs.insertCSS`) so it’s best not to try to require a stylesheet so early.

I believe keeper_scout.js and notifier_scout.js are the only entry points to all of our `.kifi-root` extension UIs (the ones that can appear on any page), and they are both only injected after document ready. A document can become ready in a preloading/invisible/faux Chrome tab, but I think that happens less frequently. Hmm... I guess this really isn’t fully solving any problem. We’ll probably need to queue failed injection attempts for retrying later.

PR #10253, opened 2014-08-13
Bootstrap document as the Angular element in dev.html; update  references.

yipingliao:
Fixes: https://app.asana.com/0/15014070792041/15466640059101

@andrewconner As we discussed, there was a disparity between what `$rootElement` is in dev and in prod, and there are several references to `$rootElement` that was incorrectly assuming that `$rootElement` is `html` (while it's actually `document` in prod). `$rootElement` is now `document` in both dev and prod, and the references are fixed.

What was broken was:
- Upper left hamburger menu (when window is small) didn't work.
- Hovering over friend pic in a keep did not bring up tooltip.
- Dragging keeps to tags didn't show the nice tag highlighting UI effects.

cc @2is10

### DIFF
--- a/kifi-backend/modules/shoebox/angular/dev.html
+++ b/kifi-backend/modules/shoebox/angular/dev.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html ng-app="kifi">
+<html>
 <head>
   <meta charset="utf-8">
   <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
@@ -35,5 +35,6 @@
   </script>
   <script src="/dist/lib.js"></script>
   <script src="/dist/kifi.js"></script>
+  <script>angular.bootstrap(document, ['kifi']);</script>
 </body>
 </html>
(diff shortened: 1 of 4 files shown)


andrewconner:
Thanks!

2is10:
@yipingliao Did you consider making `$rootElement` the `<html>` element instead of the document?
The document isn’t actually an element (see [node types](https://developer.mozilla.org/en-US/docs/Web/API/Node.nodeType#Syntax)).

andrewconner:
Of note, [their recommended stub](https://docs.angularjs.org/guide/bootstrap) is:

``` javascript
    angular.element(document).ready(function() {
      angular.bootstrap(document, ['myApp']);
    });
```

Our current variation works because the scripts are at the end of the DOM. [Interesting link](http://stackoverflow.com/questions/16537783/which-method-should-i-use-to-manually-bootstrap-my-angularjs), specifically, we can do:

``` javascript
angular.bootstrap(angular.element("body")[0], ['TodoApp']);
```

yipingliao:
If we made `body` the root element, will we lose the gain in rendering speed we get by bootstrapping to `document`?

andrewconner:
I'd need to test. However, we could make `<html>` the root element using:

``` javascript
angular.bootstrap(angular.element('html')[0], ['kifi']);
```

2is10:
Native DOM shortcuts, FYI:
`<html>` ← `document.documentElement`
`<body>` ← `document.body`

PR #10255, opened 2014-08-13
Making kifeeeed access it's own experiment (separate from admin/whitelist)

stkem:
@lawzlo @yingjieMiao 
cc @eishay


eishay:
thanks!

PR #10259, opened 2014-08-13
vote API uses HtmlForm instead of json because of sending through email

lawzlo:
@joshm1 @stkem

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/RecommendationsController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/RecommendationsController.scala
@@ -31,7 +31,7 @@ class RecommendationsController @Inject() (
 
   def updateUriRecommendationUserInteraction(url: String) = HtmlAction.authenticatedAsync { request =>
     val form = request.body.asFormUrlEncoded.get
-    val vote = UriRecommendationUserInteraction(form.get("vote").flatMap(_.headOption.asInstanceOf[Option[Boolean]]))
+    val vote = UriRecommendationUserInteraction(form.get("vote").flatMap(_.headOption).map(_.toBoolean))
     commander.updateUriRecommendationUserInteraction(request.userId, url, vote).map(fkis => Ok(Json.toJson(fkis)))
   }
 }
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/RecommendationsController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/RecommendationsController.scala
@@ -2,12 +2,10 @@ package com.keepit.controllers.website
 
 import com.keepit.common.controller.{ ShoeboxServiceController, ActionAuthenticator, WebsiteController }
 import com.keepit.commanders.{ RecommendationsCommander, LocalUserExperimentCommander }
-import com.keepit.common.db.Id
 import com.keepit.common.db.slick.Database
 import com.keepit.model._
-import com.keepit.normalizer.NormalizedURIInterner
 import play.api.libs.concurrent.Execution.Implicits.defaultContext
-import play.api.libs.json.{ JsNull, Json }
+import play.api.libs.json.{ Json }
 import scala.concurrent.Future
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/app/com/keepit/controllers/website/RecommendationsController.scala
```diff
@@ -26,13 +24,14 @@ class RecommendationsController @Inject() (
...
-    commander.updateUriRecommendationFeedback(userId, url, feedback).map(fkis => Ok(Json.toJson(fkis)))
+    commander.updateUriRecommendationFeedback(request.userId, url, feedback).map(fkis => Ok(Json.toJson(fkis)))
   }
 
-  def updateUriRecommendationUserInteraction(userId: Id[User], url: String) = JsonAction.authenticatedParseJsonAsync { request =>
-    val vote = request.body.as[UriRecommendationUserInteraction]
-    commander.updateUriRecommendationUserInteraction(userId, url, vote).map(fkis => Ok(Json.toJson(fkis)))
+  def updateUriRecommendationUserInteraction(url: String) = HtmlAction.authenticatedAsync { request =>
+    val form = request.body.asFormUrlEncoded.get
+    val vote = UriRecommendationUserInteraction(form.get("vote").flatMap(_.headOption.asInstanceOf[Option[Boolean]]))
```
joshm1: `form.get("vote")` is going to return a String, which can't be cast to a Boolean even if the string is `"true"` or `"false"`. if `vote` was in the query string, Play could use QueryBindable to give you a boolean, but I don't think it'd work this way.

lawzlo: I think it can. 

scala> val vote = List(Some("true"))
vote: List[Some[String]] = List(Some(true))

scala> vote.flatMap(_.headOption.asInstanceOf[Option[Boolean]])
res1: List[Boolean] = List(true)

joshm1: interesting, it works with `Option[String]` but not with `String`.  I tested this:

```
scala> "true".asInstanceOf[Boolean]
java.lang.ClassCastException
```

but this worked

```
scala> Some("true").asInstanceOf[Option[Boolean]]
res8: Option[Boolean] = Some(true)

scala> Some("false").asInstanceOf[Option[Boolean]]
res9: Option[Boolean] = Some(false)
```

TIL

lawzlo: Good find. I think it's because Boolean is subtype of AnyVal, but String is AnyRef. So they can't convert type directly. I think there is simpler way to do this is: 
`scala> "true".toBoolean
res2: Boolean = true
`

stkem: `asInstanceOf` is actually really dangerous when used in conjunction with generics (like `Option[T]`). Due to JVM type erasure (generics are basically all converted to `Any` at runtime behind the scenes) this leads to code that compiles fine but fails at runtime.
I think the best way here is indeed as @lawzlo sugested using `toBoolean` and just letting it fail.

stkem: adding @andrewconner for forms expertise

andrewconner:
``` scala
scala> val a = Some(List(Some(1),Some(2)))
a: Some[List[Some[Int]]] = Some(List(Some(1), Some(2)))

scala> a.flatMap(_.headOption).flatten
res10: Option[Int] = Some(1)
```

stkem: and then `toBoolean`

lawzlo: Cool. Thanks guys.

PR #10260, opened 2014-08-13
check if image is legit & replace with default if not

ahsu1230:
(used in admin pages)

sorry this took so long
I got confused so many times along the way

@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/common/store/S3ImageStore.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/common/store/S3ImageStore.scala
@@ -163,8 +163,7 @@ class S3ImageStoreImpl @Inject() (
       })
       future onComplete {
         case Success(a) =>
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/app/com/keepit/common/store/S3ImageStore.scala
```diff
@@ -145,17 +145,26 @@ class S3ImageStoreImpl @Inject() (
...
+            val key = keyByExternalId(sizeName, externalId, usedImage)
+            val putObj = uploadToS3(key, response.getAHCResponse.getResponseBodyAsStream, label = originalImageUrl)
+            (usedImage, putObj)
         }
       })
       future onComplete {
-        case Success(_) =>
-          val usedImage = if (useDefaultImage) S3UserPictureConfig.defaultName else pictureName
+        case Success(a) =>
+          val usedDefault = a.foldLeft(false)((r, c) => r || c._1 == S3UserPictureConfig.defaultName)
```
andrewconner: `a.exists` perhaps?

PR #10262, opened 2014-08-13
allowing KifiVersion to have major.minor and no patch

2is10:
also using -1 to signify that a trailing ".0" was not provided
@stkem 
cc @jaredpetker


kifi-backend/modules/common/app/com/keepit/model/KifiVersion.scala
```diff
@@ -9,50 +9,75 @@ trait KifiVersion {
...
-    } else {
+    } else if (this.build > 0 || that.build > 0) {
       this.build - that.build
+    } else {
+      0
     }
   }
 
+  override def hashCode: Int = {
+    31 * (31 * (31 * major + minor) + (if (patch >= 0) patch else 0)) + (if (build >= 0) build else 0)
```
stkem, line 30: :-)

stkem:
lgtm

PR #10264, opened 2014-08-13
Rotate empty state connection badge.

yipingliao:
Rotate the connection badge on empty state (0 friends, 0 connections). Currently it's a static FB connection badge.

Also refactored out the rotating badge as its own directive so both empty state and non-empty state use it.

Fixes: https://app.asana.com/0/15014070792041/15493074535350
Screenshot of the two different badges at: https://app.asana.com/0/15014070792041/15493074535345

@2is10

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/friends/rightColConnectView.tpl.html
+++ b/kifi-backend/modules/shoebox/angular/src/friends/rightColConnectView.tpl.html
@@ -1,7 +1,5 @@
-<div class="kf-rightcol-friends-connect" ng-if="network">
+<div class="kf-rightcol-friends-connect" ng-show="connectSocial">
   <div class="kf-friends-content kf-friends-content-bigger">Find friends already on Kifi</div>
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/shoebox/angular/src/friends/rightColConnectView.tpl.html
```diff
@@ -1,7 +1,5 @@
-<div class="kf-rightcol-friends-connect" ng-if="network">
+<div class="kf-rightcol-friends-connect" ng-show="connectSocial">
```
yipingliao, line 1: Replaced `ng-if` with `ng-show` because I am two-way binding `network` between the parent scope and the isolate scope in `kfRotatingConnect` directive. This is so that the parent scope is aware of which network the `kfRotatingConnect` directive has connected to, if any. `ng-if` does not allow enable this two-way binding, but `ng-show` does.

2is10:
very nice!

PR #10265, opened 2014-08-13
vote buttons in feed email

joshm1:
@stkem

### DIFF
--- a/kifi-backend/modules/curator/app/views/email/feedRecommendationsInlined.scala.html
+++ b/kifi-backend/modules/curator/app/views/email/feedRecommendationsInlined.scala.html
@@ -1,14 +1,25 @@
-@(firstName: String, recos: Seq[com.keepit.curator.commanders.email.RecommendedUriSummary])
+@(emailData: com.keepit.curator.commanders.email.RecommendedEmailData)
 
 
-Hello @firstName,<br/><br/>
+Hello @emailData.toUser.firstName,<br/><br/>
 
-Kifi recommends @recos.size items for you:
+Kifi recommends @emailData.recos.size items for you:
 
 <ol>
-    @recos.map { reco =>
+    @emailData.recos.map { reco =>
     <li>
         <h3>@reco.title</h3>
+        <div>
+            How is this recommendation?
+            <form method="post" action="https://www.kifi.com/site/recos/vote?url=@helper.urlEncode(reco.url)">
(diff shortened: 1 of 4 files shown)


stkem:
:+1:

eishay:
nice

kifi-backend/modules/curator/app/views/email/feedRecommendationsInlined.scala.html
```diff
@@ -1,14 +1,25 @@
...
+        <div>
+            How is this recommendation?
+            <form method="post" action="https://www.kifi.com/site/recos/vote?url=@helper.urlEncode(reco.url)">
+                <input type="hidden" name="vote" value="true"/>
+                <input type="submit" value="Good"/>
+            </form>
+            <form method="post" action="https://www.kifi.com/site/recos/vote?url=@helper.urlEncode(reco.url)">
+                <input type="hidden" name="vote" value="false"/>
+                <input type="submit" value="Bad"/>
+            </form>
```
2is10, line 21: Something to keep in mind if we’re going to have this long-term: the URLs could be in hidden inputs too (not url-encoded). This would avoid potential URL length limit issues.

joshm1, line 21: Yep. including @lawzlo @stkem 

  The `form`'s are bad for emails anyway, Gmail will warn users before sending the request and some mail apps (like Airmail) try to handle the request themselves w/o spawning it in the browser (which breaks authentication).  If we ever need to do anything like this publically, it needs to get a GET and we should use a token for authentication instead of `*Action.authenticated` action handlers

lawzlo, line 21: This API will be only used for tester.

2is10, line 21: Agreed. Actions shouldn’t generally be taken directly from emails. The invite-a-specific-friend feature you built, Josh, is an example of a better way to handle it.

joshm1, line 21: Yes I agree in general. There are cases where it makes sense and would be bad UX otherwise. An example that comes to mind are Twitter emails with a clear "Follow" button in them that sends a GET.

PR #10266, opened 2014-08-13
shoebox clean-up

raymondng:
Clean-up remaining dependencies in shoebox runtime
Removed keepAttribution endpoint

Note that some tests are no longer relevant so they have been removed
Replacements can be found in branch rm_keepAttribution which will likely be the next pull request

@yingjieMiao @eishay


kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
```diff
@@ -157,7 +155,7 @@ class KeepsCommander @Inject() (
     collectionId: Option[ExternalId[Collection]],
     helprankOpt: Option[String],
     count: Int,
-    userId: Id[User]): (Seq[Keep], Option[Collection], Map[Id[Keep], Int], Map[Id[Keep], Int]) = {
+    userId: Id[User]): Future[(Seq[Keep], Option[Collection], Map[Id[Keep], Int], Map[Id[Keep], Int])] = {
```
yingjieMiao, line 159: it would be nice if we can break down this big method. also, some comments on the return values will be helpful. (Or, case class wrappers)

raymondng, line 159: Agreed. I will look into simplifying this next -- at least the helprank-specific part.

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
```diff
@@ -181,31 +179,44 @@ class KeepsCommander @Inject() (
 
     db.readOnlyReplica { implicit ro =>
       val collectionOpt = (collectionId map { id => collectionRepo.getByUserAndExternalId(userId, id) }).flatten
-      val keeps = collectionOpt match {
+      val keepsF = collectionOpt match {
```
yingjieMiao, line 182: you might want to match on the tuple `(collectionOpt, helpRankOpt)`. might make the branching clearer.

yingjieMiao:
lgtm

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
raymondng, line 182: good suggestion. I am trying with a few things -- will go over them with you.

PR #10267, opened 2014-08-13
fixing Android app version detection on websocket connection

2is10:
@stkem 
cc @jaredpetker

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/common/net/UserAgent.scala
+++ b/kifi-backend/modules/common/app/com/keepit/common/net/UserAgent.scala
@@ -24,7 +24,6 @@ case class UserAgent(
 object UserAgent extends Logging {
 
   val KifiIphoneAppTypeName = "kifi iphone app"
-  // val KifiAndroidAppTypeName = "kifi android app"
 
   private val MobileOses = Set("Android", "iOS", "Bada", "DangerOS", "Firefox OS", "Mac OS", "Palm OS", "BlackBerry OS", "Symbian OS", "webOS")
   private val TabletIndicators = Set("iPad", "Tablet")
--- a/kifi-backend/modules/common/app/com/keepit/common/net/UserAgent.scala
+++ b/kifi-backend/modules/common/app/com/keepit/common/net/UserAgent.scala
@@ -12,9 +12,10 @@ case class UserAgent(
     typeName: String,
     version: String) {
   lazy val isKifiIphoneApp: Boolean = typeName == UserAgent.KifiIphoneAppTypeName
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/eliza/app/com/keepit/eliza/controllers/AuthenticatedWebSocketsController.scala
```diff
@@ -169,9 +169,9 @@ trait AuthenticatedWebSocketsController extends ElizaServiceController {
           implicit val (enumerator, channel) = Concurrent.broadcast[JsArray]
 
           val typedVersionOpt: Option[KifiVersion] = try {
-            UserAgent.fromString(streamSession.userAgent).typeName match {
-              case UserAgent.KifiIphoneAppTypeName => versionOpt.map(KifiIPhoneVersion.apply)
-              case UserAgent.KifiAndroidAppTypeName => versionOpt.map(KifiAndroidVersion.apply)
+            UserAgent.fromString(streamSession.userAgent) match {
+              case ua if ua.isKifiIphoneApp => versionOpt.map(KifiIPhoneVersion.apply)
+              case ua if ua.isAndroid => versionOpt.map(KifiAndroidVersion.apply)
```
2is10: we’re currently assuming it’s the Kifi Android app if the user agent contains “Android”. would be better to have a dedicated Kifi Android app user agent string format, I think, like we do for iOS (see `UserAgent.iosAppRe` above)

stkem: +1. How hard would that be @jaredpetker?

kifi-backend/modules/common/app/com/keepit/common/net/UserAgent.scala
```diff
@@ -23,7 +24,7 @@ case class UserAgent(
 object UserAgent extends Logging {
 
   val KifiIphoneAppTypeName = "kifi iphone app"
-  val KifiAndroidAppTypeName = "kifi android app"
+  // val KifiAndroidAppTypeName = "kifi android app"
```
stkem: Is this line intended to be uncommented very soon? If not I think it's best not to leave it, lest it confuse the next person in there.

stkem:
made one very minor comment, lgtm otherwise.

kifi-backend/modules/common/app/com/keepit/common/net/UserAgent.scala
2is10: I thought it was less dangerous commented than not, and I thought it was a good hint about the intended direction for the Android app (in conjunction with the TODO comment on line 15), but rather than add more commenting to clarify, I’ve removed it.

PR #10269, opened 2014-08-14
Stop silently failing on certain web socket connect errors

stkem:
\\+ removes forgiving client version extraction (now strict again)
@2is10 @jaredpetker 
This should probably only be merged once android changes are confirmed, so @jaredpetker you give the thumbs up when ready.

### DIFF
--- a/kifi-backend/modules/eliza/app/com/keepit/eliza/controllers/AuthenticatedWebSocketsController.scala
+++ b/kifi-backend/modules/eliza/app/com/keepit/eliza/controllers/AuthenticatedWebSocketsController.scala
@@ -165,33 +165,34 @@ trait AuthenticatedWebSocketsController extends ElizaServiceController {
       }
     } else {
       authenticate(request) match {
-        case Some(streamSessionFuture) => streamSessionFuture.map { streamSession =>
-          implicit val (enumerator, channel) = Concurrent.broadcast[JsArray]
+        case Some(streamSessionFuture) =>
+          val iterateeAndEnumeratorFuture = streamSessionFuture.map { streamSession =>
+            implicit val (enumerator, channel) = Concurrent.broadcast[JsArray]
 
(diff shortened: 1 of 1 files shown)


jaredpetker:
I've made changes myself, of course any other clients out there may
suffer...should be fine though.

stkem:
Cool. Given that it's relatively late in the day and this has the potential of causing a bunch of airbrakes I'll merge it in the morning.

kifi-backend/modules/eliza/app/com/keepit/eliza/controllers/AuthenticatedWebSocketsController.scala
```diff
@@ -165,33 +165,34 @@ trait AuthenticatedWebSocketsController extends ElizaServiceController {
...
+              startMessages = startMessages :+ Json.arr("version", "new")
             }
+            onConnect(socketInfo)
+            (iteratee(streamSession, versionOpt, socketInfo, channel), Enumerator(startMessages: _*) >>> enumerator)
           }
-          val ipOpt: Option[String] = eipOpt.flatMap { eip =>
-            crypt.decrypt(ipkey, eip).toOption
+          iterateeAndEnumeratorFuture.onFailure {
+            case t: Throwable =>
+              airbrake.notify("Fatal error when establishing websocket connection", t)
```
2is10, line 193: I see. So Play closes the connection automatically on a failed `Future`, and this just ensures we report it. Feels a little odd that bad requests can trigger Airbrakes, but I guess only authenticated requests make it to this code path, so it’s better to assume we’ve made an error than to assume a user is playing with request parameters. Thanks!

stkem:
Deploy in progress. Please let me know if you notice issues @jaredpetker

stkem:
We are getting a bunch of errors fro iOS clients with a bad version. I'm rolling this back until we can confirm that these are being rejected legitimately (in an interview right now).

PR #10270, opened 2014-08-14
ios8 extension query (userInfo, pageInfo, tags, & friend connections)

ahsu1230:
@thassss @andrewconner


kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobilePageController.scala
```diff
@@ -34,4 +26,65 @@ class MobilePageController @Inject() (
...
+    // page/details (POST)
+    val pageFutures = SafeFuture { pageCommander.getPageDetails(url, request.userId, request.experiments) }
+    // user/me (GET)
+    val basicUserFutures = SafeFuture { userCommander.getUserInfo(request.user) }
+    val userAttributeFutures = userCommander.getKeepAttributionInfo(request.userId)
+    // collections/all (GET)
+    val numKeepsFuture = SafeFuture { db.readOnlyMaster { implicit s => keepRepo.getCountByUser(request.userId) } }
+    val collectionsFuture = SafeFuture { collectionCommander.allCollections(sortOrder, request.userId) }
+    // user/friendsDetails (GET)
+    val friendsFuture = SafeFuture { userCommander.getConnectionsPage(request.userId, 0, 20) }
```
ahsu1230: woops, will need to change this.
This function will require 2 more parameters (page number & page size) just like friends() function in MobileUserController

kifi-backend/modules/shoebox/test/com/keepit/controllers/mobile/MobilePageControllerTest.scala
```diff
@@ -106,5 +116,99 @@ class MobilePageControllerTest extends TestKit(ActorSystem()) with Specification
         Json.parse(contentAsString(result)) must equalTo(expected)
       }
     }
+
+    "query extension " in {
+      withDb(mobileControllerTestModules: _*) { implicit injector =>
+        val path = com.keepit.controllers.mobile.routes.MobilePageController.queryExtension().url
+        path === "/m/1/page/extensionQuery"
```
eishay, line 123: great, we must have these tests for endpoints

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobilePageController.scala
```diff
@@ -34,4 +26,65 @@ class MobilePageController @Inject() (
...
+
+    val tupleFuture = for {
+      pageInfo <- pageFutures
+      basicUserInfo <- basicUserFutures
+      userAttributeInfo <- userAttributeFutures
+      numKeeps <- numKeepsFuture
+      collections <- collectionsFuture
+      friendConnections <- friendsFuture
+    } yield {
+      (pageInfo, basicUserInfo, userAttributeInfo, numKeeps, collections, friendConnections)
```
andrewconner, line 53: the `yield` block here runs when all the futures are resolved. So, it may make sense to just put the body (starting at `val friendsJsons = connectionsPage ...`) inside this. It saves the next 9 lines and is more clear.

PR #10272, opened 2014-08-14
track digest email clicks; digest email more readable

joshm1:


joshm1:
@stkem @andrewconner 
cc @eishay 

I used the user_id column in electronic_mail table to track who we send the digest email TO.  The repo maps this column to the "senderUserId" field, which is a different meaning.  This field was previously unused for this email, and if we ever for any reason want to send the same digest to multiple users, this would have to change.  Thoughts?

stkem:
lgtm. I don't anticipate us sending bulk emails like that, so we can cross that bridge when we come to it.

PR #10273, opened 2014-08-14
Common core

andrewconner:
Previously, we had an odd pattern to enable implicit library augmentation: either importing `com.keepit.common._` (which puts all `common` packages in top-level scope) or importing `com.keepit.common.KestrelCombinator` (which isn't intuitive). Additionally, we had several full objects oddly placed inside the common package object.

Modeling after `scala.Predef`, this is the beginnings of a codebase wide standard library providing helpful conversations, augmentations, etc. To use, import `import com.keepit.common.core._`. I will be exploring more general purpose, helpful bits to add here.

@stkem @eishay @raymondng @ymatsuda @joshm1


kifi-backend/modules/common/app/com/keepit/common/Implicits.scala
```diff
@@ -0,0 +1,48 @@
...
+  // Kestrel Combinator, tee operator
+  // expr tap println
+  @inline def tap(f: A => Unit): A = {
+    f(x)
+    x
+  }
+}
+
+final class FuncExtensionOpts[A](f: => A) {
+  @inline def recover(g: Throwable => A): A = {
```
stkem: partial function?

stkem:
nice! lgtm.

eishay:
lgtm

raymondng:
:+1:

andrewconner:
By the way, the forward pipe is a very neat operator. Not super common in Scala, but should be. It makes for very readable control flow. Two common use cases: using it like `map` when you're not using monads:

``` scala
getUser() |> alertFriends |> createLibraries |> save
```

The other saves you from really nasty code when trying to update several fields dynamically. Think the Keep importer, which _may_ mutate one of several fields:

``` scala
keep |> { k =>
  if (newTitle.isDefined) k.copy(title = newTitle) else k
} |> { k =>
  if (newDate.isDefined) k.copy(createdAt = newDate) else k
} |> { k =>
  if (newUrl.isDefined) k.copy(url = newUrl) else k
} |> keepRepo.save
```

2is10:
I definitely like the concept, but like everything else, it needs to be used judiciously.

In your first example, the reader has to look up the methods to figure out what’s being saved.

And, for fun, here’s a concise alternative to your second example that makes only one copy and avoids saving if there are no changes.

``` scala
if (Seq(newDate, newTitle, newUrl).flatten.nonEmpty) {
  keepRepo.save k.copy(
    createdAt = newDate.orElse(k.createdAt),
    title = newTitle.orElse(k.title),
    url = newUrl.orElse(k.url))
}
```

There are so many ways to do everything in Scala!

andrewconner:
True. For the 2nd, our use cases are typically more complex than just the options. For my example, you're right.

PR #10274, opened 2014-08-14
Uri penalizing/boosting

lawzlo:
@stkem @yingjieMiao


kifi-backend/modules/common/conf/evolutions/shoebox/227.sql
```diff
@@ -0,0 +1,4 @@
+alter table raw_seed_item
+    add column url NOT NULL;
+
+insert into evolutions (name, description) values('227.sql', 'add url column');
```
yingjieMiao: this is not compliant to our convention of evolution file.

lawzlo: Will update format, and probably file number and name before merging, in case there are other evolution files come in before my.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
```diff
@@ -85,7 +88,7 @@ class RecommendationGenerationCommander @Inject() (
                 scoreCoefficients.rekeepScore.getOrElse(defaultScore) * reco.allScores.rekeepScore +
                 scoreCoefficients.discoveryScore.getOrElse(defaultScore) * reco.allScores.discoveryScore
             }
-          },
+          } * reco.allScores.weight,
```
yingjieMiao: this weight is already applied when you computing master score. Also, it might be better to refactor the long formula in the else branch into a function.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/UriWeightingHelper.scala
```diff
@@ -0,0 +1,37 @@
...
+    //----------------------------------Penalize---------------------------------------------------------------------------
+    ("""(?i)\b(https?|ftp|file)://[-A-Za-z0-9+&@#/%?=~_|!:,.;]*twitter.com[-A-Za-z0-9+&@#/%=~_|]""".r, 0.01f, "Twitter"),
+    ("""(?i)\b(https?|ftp|file)://[-A-Za-z0-9+&@#/%?=~_|!:,.;]*linkedin.com[-A-Za-z0-9+&@#/%=~_|]""".r, 0.01f, "LinkedIn"),
+    ("""(?i)\b(https?|ftp|file)://[-A-Za-z0-9+&@#/%?=~_|!:,.;]*mail.google.com[-A-Za-z0-9+&@#/%=~_|]""".r, 0.001f, "Google Mail"),
+    ("""(?i)\b(https?|ftp|file)://[-A-Za-z0-9+&@#/%?=~_|!:,.;]*wikipedia.org[-A-Za-z0-9+&@#/%=~_|]""".r, 0.1f, "Wikipedia"),
+    ("""(?i)\b(https?|ftp|file)://[-A-Za-z0-9+&@#/%?=~_|!:,.;]*facebook.com[-A-Za-z0-9+&@#/%=~_|]""".r, 0.001f, "Facebook"),
+
+    //----------------------------------Boost--------------------------------------------------------------------------------
+    ("""(?i)\b(https?|ftp|file)://[-A-Za-z0-9+&@#/%?=~_|!:,.;]*techcrunch.com[-A-Za-z0-9+&@#/%=~_|]""".r, 1.2f, "Techcrunch")
+  )
```
yingjieMiao, line 28: cool. Not an expert on regex. But I think the `ftp|file` part is not necessary here. Also, `[-A-Za-z0-9+&@#/%?=~_|!:,.;]` is a bit too verbose. We have a few examples in the url normalizers, e.g. `AmazonNormalizer`, `LinkedInNormalizer`

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
lawzlo: computeMasterScore is used by all users. this part is used by beta tester so they can change their scores from admin panel. I think they are separate and both places should be applied the weight. For long formula, sure, I will put it into a function.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/UriWeightingHelper.scala
```diff
@@ -0,0 +1,37 @@
...
+
+  def apply(items: Seq[SeedItem]): Seq[WeightedSeedItem] = items.map { item =>
+    var weight = 1.0f
+    weightScore.foreach(reg =>
+      weight = reg._1.findFirstIn(item.url) match {
+        case Some(data) => reg._2
```
yingjieMiao: would be nice if we can have a case class for that 3-tuple, so you don't have to use `_._2`

kifi-backend/modules/curator/app/com/keepit/curator/commanders/UriWeightingHelper.scala
```diff
@@ -0,0 +1,37 @@
...
+    ("""(?i)\b(https?|ftp|file)://[-A-Za-z0-9+&@#/%?=~_|!:,.;]*mail.google.com[-A-Za-z0-9+&@#/%=~_|]""".r, 0.001f, "Google Mail"),
+    ("""(?i)\b(https?|ftp|file)://[-A-Za-z0-9+&@#/%?=~_|!:,.;]*wikipedia.org[-A-Za-z0-9+&@#/%=~_|]""".r, 0.1f, "Wikipedia"),
+    ("""(?i)\b(https?|ftp|file)://[-A-Za-z0-9+&@#/%?=~_|!:,.;]*facebook.com[-A-Za-z0-9+&@#/%=~_|]""".r, 0.001f, "Facebook"),
+
+    //----------------------------------Boost--------------------------------------------------------------------------------
+    ("""(?i)\b(https?|ftp|file)://[-A-Za-z0-9+&@#/%?=~_|!:,.;]*techcrunch.com[-A-Za-z0-9+&@#/%=~_|]""".r, 1.2f, "Techcrunch")
+  )
+
+  def apply(items: Seq[SeedItem]): Seq[WeightedSeedItem] = items.map { item =>
+    var weight = 1.0f
```
yingjieMiao: this `var` is never modified in the `foreach`.  Also, the naming `weightScore` is not accurate. Something like `RegexBasedBooster` might be better.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/UriWeightingHelper.scala
```diff
@@ -0,0 +1,37 @@
...
+    weightScore.foreach(reg =>
```
yingjieMiao: `foreach` can be too expensive if you already find a match.

kifi-backend/modules/common/conf/evolutions/shoebox/227.sql
```diff
@@ -0,0 +1,4 @@
+alter table raw_seed_item
+    add column url NOT NULL;
```
yingjieMiao: column type? `varchar(3072)`. as in bookmark table.

yingjieMiao: Since this is NOT NULL, you would have exceptions before you deploy new curator.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/UriWeightingHelper.scala
yingjieMiao: sorry, mis-read the code. `var` is modified.

lawzlo: I think it does. if the link matched, then weight will be changed to new value. for the name. I'm not sure about `RegexBasedBooster` since it's not boosting only and don't see why to indicate it's RegexBased. Why weightScore is not accurate?

kifi-backend/modules/curator/app/com/keepit/curator/commanders/UriWeightingHelper.scala
lawzlo, line 28: Will remove `ftp|file`. and simplify `[-A-Za-z0-9+&@#/%?=~_|!:,.;]`. Thanks.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/UriWeightingHelper.scala
lawzlo: Cool. Thanks. Will move then into a case class.

kifi-backend/modules/common/conf/evolutions/shoebox/227.sql
lawzlo: oh. Thanks for letting me know.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/UriWeightingHelper.scala
yingjieMiao: something like this might work: `weightScore.find( tuple => item.url.matches(tuple._1))`. so you get the one you want. then get the weight once.

lawzlo: cool. Trying to optimize it.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/UriWeightingHelper.scala
yingjieMiao: Not feeling so strong about the naming. 'ScoringMultiplier'? In the `foreach`, you are using `reg` to denote the 3-tuple, a bit confusing on first look.

lawzlo: `ScoringMultiplier` works for me.

kifi-backend/modules/common/conf/evolutions/shoebox/227.sql
joshm1: on the topic of evolution file names... any reason we use this naming convention over filenames with a timestamps & description?  this way they can still be ordered.  it's practically immune to merge conflicts and gives a an high-level overview of what evolutions do w/o having to open each one.

example: `201408141835_curator_add_url_to_raw_seed_item.sql`

cc @andrewconner @stkem @eishay

andrewconner: Yep. It's required by Play's evolution plugin. We could write our own, but I imagine that's a low priority :)

joshm1: Simple enough :-)

lawzlo: First, I don't think it hurts us by using number for now, maybe it does in
the future. Second, for simple enough...if I create a file before another
engineer but another engineer modified db before me. what happen to my
evolution file?

kifi-backend/modules/curator/app/com/keepit/curator/commanders/UriWeightingHelper.scala
```diff
@@ -0,0 +1,41 @@
...
+  val scoringMultiplier = Seq(
+    //----------------------------------Penalize---------------------------------------------------------------------------
+    UrlPattern("""^https?://[-A-Za-z0-9+&@#/%?=~_|.]*twitter.com[-A-Za-z0-9+&@#/%=~_|]""".r, 0.01f, "Twitter"),
+    UrlPattern("""^https?://[-A-Za-z0-9+&@#/%?=~_|.]*linkedin.com[-A-Za-z0-9+&@#/%=~_|]""".r, 0.01f, "LinkedIn"),
+    UrlPattern("""^https?://[-A-Za-z0-9+&@#/%?=~_|.]*mail.google.com[-A-Za-z0-9+&@#/%=~_|]""".r, 0.001f, "Google Mail"),
+    UrlPattern("""^https?://[-A-Za-z0-9+&@#/%?=~_|.]*wikipedia.org[-A-Za-z0-9+&@#/%=~_|]""".r, 0.1f, "Wikipedia"),
+    UrlPattern("""^https?://[-A-Za-z0-9+&@#/%?=~_|.]*facebook.com[-A-Za-z0-9+&@#/%=~_|]""".r, 0.001f, "Facebook"),
+
+    //----------------------------------Boost--------------------------------------------------------------------------------
+    UrlPattern("""^https?://[-A-Za-z0-9+&@#/%?=~_|.]*techcrunch.com[-A-Za-z0-9+&@#/%=~_|]""".r, 1.2f, "Techcrunch")
```
andrewconner: Do we need such expensive regular expressions for domain matching?

kifi-backend/modules/curator/app/com/keepit/curator/commanders/UriWeightingHelper.scala
```diff
@@ -0,0 +1,41 @@
...
+import com.keepit.curator.model.{ WeightedSeedItem, SeedItem }
+
+import scala.util.matching.Regex
+
+case class UrlPattern(
+  regex: Regex,
+  weight: Float,
+  description: String)
+
+class UriWeightingHelper() {
```
andrewconner, line 14: `@Singleton` (regexes are expensive to compile)

kifi-backend/modules/common/conf/evolutions/shoebox/227.sql
andrewconner: No one should modify the db until merging in their code. This solves any issues with conflicts.

kifi-backend/modules/common/conf/evolutions/shoebox/226.sql
```diff
@@ -0,0 +1,10 @@
+# CURATOR
+
+# --- !Ups
+
+alter table raw_seed_item
+    add column url varchar(3072) NOT NULL;
```
andrewconner: This will fail.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/UriWeightingHelper.scala
lawzlo, line 14: Cool. Adding.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/UriWeightingHelper.scala
lawzlo: I'm not sure all cases of urls of those websites(and more in the future). So try to cover them all. If it's ok to miss some matches(very low possibility I think). I can simplify.

kifi-backend/modules/common/conf/evolutions/shoebox/226.sql
lawzlo: hmmm.. not sure why.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/UriWeightingHelper.scala
lawzlo: maybe like this: `"""^https?://[-A-Za-z0-9.]*facebook.com[-A-Za-z0-9+&@#/%=~_|]""".r`?

lawzlo: or even simpler: `"""^https?://[-A-Za-z0-9.]*facebook.com*""".r`

kifi-backend/modules/common/conf/evolutions/shoebox/226.sql
lawzlo: will add empty string as default value.

andrewconner: Mind leaving a comment on the field then in the case class and repo? A blank string for a URL may be unexpected.

lawzlo: Sure. Will add comments.

kifi-backend/modules/common/conf/evolutions/shoebox/226.sql
```diff
@@ -0,0 +1,10 @@
+# CURATOR
+
+# --- !Ups
+
+alter table raw_seed_item
+    add column url text NOT NULL DEFAULT '';
```
stkem, line 6: Once this is live and we have re-ingested all the data with urls we should remove the default. Help me remember.

lawzlo, line 6: will do. :)

kifi-backend/modules/curator/app/com/keepit/curator/model/SeedItem.scala
```diff
@@ -30,11 +31,21 @@ case class SeedItem(
     recencyScore: Float,
     priorScore: Float,
     rekeepScore: Float,
-    discoveryScore: Float) {
+    discoveryScore: Float,
+    weight: Float) {
```
stkem: I would prefer this to be a separate field in `ScoredSeedItem` in stead of being part of `UriScores`. Semantically it doesn't make much if a difference, but it plays a different role than all the other scores so it will be easier to understand for the next person in this code if it is separate.

stkem: Also, I think `multiplier` is a better name here, since we already use the term `weight` for the user supplied score coefficients.

kifi-backend/modules/curator/app/com/keepit/curator/model/SeedItem.scala
```diff
@@ -30,11 +31,21 @@ case class SeedItem(
...
 
-  override def toString = s"social:$socialScore --- popularity:$popularityScore --- overallInterest:$overallInterestScore --- recentInterest:$recentInterestScore --- recency:$recencyScore --- prior:$priorScore --- rekeep:$rekeepScore --- discovery:$discoveryScore"
+  override def toString = s"social:$socialScore --- popularity:$popularityScore --- overallInterest:$overallInterestScore --- recentInterest:$recentInterestScore --- recency:$recencyScore --- prior:$priorScore --- rekeep:$rekeepScore --- discovery:$discoveryScore --- weight:$weight"
 }
 
+case class WeightedSeedItem(
+  weightMultiplier: Float = 1.0f,
```
stkem: just `multiplier` is clearer, I think (see comment above)

lawzlo: will rename it.

kifi-backend/modules/curator/app/com/keepit/curator/model/SeedItem.scala
lawzlo: Yes, this was the first way I tried. And changed to current way because when calling `getAdHocRecommendations`, we lost multiplier. So I moved it in UriScores.

kifi-backend/modules/curator/app/com/keepit/curator/model/SeedItem.scala
```diff
@@ -30,11 +31,21 @@ case class SeedItem(
...
     rekeepScore: Float,
-    discoveryScore: Float) {
+    discoveryScore: Float,
+    weight: Float) {
 
-  override def toString = s"social:$socialScore --- popularity:$popularityScore --- overallInterest:$overallInterestScore --- recentInterest:$recentInterestScore --- recency:$recencyScore --- prior:$priorScore --- rekeep:$rekeepScore --- discovery:$discoveryScore"
+  override def toString = s"social:$socialScore --- popularity:$popularityScore --- overallInterest:$overallInterestScore --- recentInterest:$recentInterestScore --- recency:$recencyScore --- prior:$priorScore --- rekeep:$rekeepScore --- discovery:$discoveryScore --- weight:$weight"
 }
 
+case class WeightedSeedItem(
```
stkem: Maybe `SeedItemWithMultiplier` to keep with the field naming? If the IDE will do the rename for you, otherwise don't worry about it.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/UriWeightingHelper.scala
```diff
@@ -0,0 +1,43 @@
+package com.keepit.curator.commanders
+
+import com.keepit.curator.model.{ WeightedSeedItem, SeedItem }
+import com.google.inject.{ Singleton }
+
+import scala.util.matching.Regex
+
+case class UrlPattern(
```
stkem, line 8: :+1:

kifi-backend/modules/curator/app/com/keepit/curator/commanders/UriWeightingHelper.scala
```diff
@@ -0,0 +1,43 @@
...
+    UrlPattern("""^https?://[-A-Za-z0-9.]*linkedin.com[./?\#]""".r, 0.01f, "LinkedIn"),
+    UrlPattern("""^https?://[-A-Za-z0-9.]*mail.google.com[./?\#]""".r, 0.001f, "Google Mail"),
+    UrlPattern("""^https?://[-A-Za-z0-9.]*wikipedia.org[./?\#]""".r, 0.1f, "Wikipedia"),
+    UrlPattern("""^https?://[-A-Za-z0-9.]*facebook.com[./?\#]""".r, 0.001f, "Facebook"),
+
+    //----------------------------------Boost------------------------------------------------------------------------------
+    UrlPattern("""^https?://[-A-Za-z0-9.]*techcrunch.com[./?\#]""".r, 1.2f, "Techcrunch")
+  )
+
+  def apply(items: Seq[SeedItem]): Seq[WeightedSeedItem] = items.map { item =>
```
stkem: Actually, this should actually multiply the factors from all matching rules together. It will make more sense once we have more generic rules (e.e. boosting things with 'blog' in the domain that we may want to compound with other rules) and it also makes it independent of the order the rules are specifies in. (Order dependency can lead to some very hard to find bugs)

kifi-backend/modules/curator/app/com/keepit/curator/commanders/UriWeightingHelper.scala
lawzlo, line 8: @yingjieMiao 's idea. Thanks.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/UriWeightingHelper.scala
lawzlo: I was thinking add/multiple all matches together. It may hurt performance a little. But I think it's good idea.

stkem:
last thing is to multiply all matching rules together, as commented above. lgtm otherwise. Nice work!

kifi-backend/modules/curator/app/com/keepit/curator/commanders/UriWeightingHelper.scala
lawzlo: Will change it.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/UriWeightingHelper.scala
```diff
@@ -0,0 +1,43 @@
...
+
+import scala.util.matching.Regex
+
+case class UrlPattern(
+  regex: Regex,
+  weight: Float,
+  description: String)
+
+@Singleton
+class UriWeightingHelper() {
```
eishay, line 14: don't make stateless cheap to create classes as singletons. its a bad pattern that won't really help performance since the object will have to be looked up and short lived objects cost close to nothing.

kifi-backend/modules/curator/test/com/keepit/curator/UriWeightingHelperTest.scala
```diff
@@ -0,0 +1,70 @@
...
+import com.keepit.curator.model.{ Keepers, SeedItem }
+import com.keepit.model.{ NormalizedURI, User }
+import org.specs2.mutable.Specification
+import com.keepit.common.time._
+
+class UriWeightingHelperTest extends Specification with CuratorTestInjector {
+
+  private def makeSeedItems(): Seq[SeedItem] = {
+    val seedItem1 = SeedItem(userId = Id[User](42), uriId = Id[NormalizedURI](1),
+      url = """https://twitter.com/RoseHulmanAlums/status/494491541856337920""",
```
eishay, line 15: nice test

kifi-backend/modules/curator/app/com/keepit/curator/commanders/UriWeightingHelper.scala
```diff
@@ -0,0 +1,43 @@
...
+
+  val scoringMultiplier = Seq(
```
eishay, line 16: these companies have product and engineering blogs. can you make sure we don't block them as well?
e.g. 
https://engineering.linkedin.com/mobile/linkedin-connected-engineering-pre-meeting-intelligence
https://blog.twitter.com/2014/scalingtwitter-in-dublin-and-the-uk
https://code.facebook.com/posts/313033472212144/debugging-file-corruption-on-ios/

kifi-backend/modules/curator/app/com/keepit/curator/commanders/UriWeightingHelper.scala
```diff
@@ -0,0 +1,43 @@
...
+    //----------------------------------Penalize---------------------------------------------------------------------------
+    UrlPattern("""^https?://[-A-Za-z0-9.]*twitter.com[./?\#]""".r, 0.01f, "Twitter"),
+    UrlPattern("""^https?://[-A-Za-z0-9.]*linkedin.com[./?\#]""".r, 0.01f, "LinkedIn"),
```
eishay, line 19: linkedin its prodbably only `.*linkedin.com/(pub|in)/.*` or something of the sort

kifi-backend/modules/curator/app/com/keepit/curator/commanders/UriWeightingHelper.scala
lawzlo, line 16: Will make sure don't block useful information. l I think currently we just build mechanizer for boosting/penalizing urls. We definitely will add more url patterns and I think Mark will have a list for us. Also the boost and penalize scores will be changed later.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/UriWeightingHelper.scala
lawzlo, line 19: actually. I tried `linkedin.com#` `linkedin.com\` also will redirect to home page.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/UriWeightingHelper.scala
stkem, line 16: Yup, this is just to test the mechanism. Actually boosting/penalizing list is in the works (there is a google doc where Jen is collecting them). Good to include those cases in the test though.

lawzlo, line 16: Yes. I'm adding those patterns and test cases.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/UriWeightingHelper.scala
lawzlo, line 14: @andrewconner helped to point out the parse string to regex may expansive.(And we will have more and more regex in the future) So we have a singleton here.

lawzlo:
Tested, pushing to prod.

PR #10277, opened 2014-08-14
iOS 8 extension endpoint for keep one & array of collections

ahsu1230:
new controller + test

On the commander side, simply made a small change for keepWithMultipleTags()
this function was only used by UserCommander for defaultKeeps (but we don't have that anymore)

@thassss if you'd like the endpoint to return something different, let me know!
@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
@@ -600,7 +600,7 @@ class KeepsCommander @Inject() (
     searchClient.updateURIGraph()
 
     val keepCount = db.readOnlyMaster { implicit s =>
-      keepRepo.count
+      keepRepo.getCountByUser(userId)
     }
     (keepsByTag, keepCount)
   }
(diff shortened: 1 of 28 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
```diff
@@ -599,7 +599,10 @@ class KeepsCommander @Inject() (
 
     searchClient.updateURIGraph()
 
-    keepsByTag
+    val keepCount = db.readOnlyMaster { implicit s =>
```
andrewconner: I'm not really sure this is necessary. Could we go over the use case? @thassss When a user keeps, do you need the total keep count in the result?

thassss: I don't need it.

ahsu1230: took it out, thought it would be useful since addKeeps() has it in its response as well

ahsu1230:
@thassss
parameters: 
- KeepInfo title(optional), url, isPrivate, and optional other stuff
- array of tag names (String)

you get back list of tag external Ids that were added / updated with new keep

PR #10281, opened 2014-08-14
tolerating ?version=m17084 at websocket endpoint

2is10:
@stkem @thassss


2is10:
@stkem I think after this you should be able to revert the revert (#10279).

stkem:
Sorry for the delayed review. Had a big review queue today. This looks good to me. Let's get it on tomorrow morning including the revert revert.

PR #10283, opened 2014-08-14
fake stuff

raymondng:
@yingjieMiao @eishay -- this pull request added (replacement) tests wrt helprank integration with shoebox. The tests are pretty much the same as before but with hard-coded calls to repos replaced with calls via `FakeHeimdalServiceClientImpl`. I also removed `getDiscoveryCountByKeeper` as it's replaced by `getKeepAttributionInfo1 and is no longer used.

For everyone -- a bit more context: 

A couple of problems I needed to address:
- With APIs being ported over, `(Fake)HeimdalServiceClient` (similar to `ShoeboxServiceClient`) became a kitchen-sink that's hard to read/manage
- For integration tests to work I needed to put in some simple 'mock' impl for the repo-related methods -- creating duplicate code (counters, maps, etc.) for each repo result in boilerplate code that increases bloat

The general approach is similar to what we already do in `FakeShoeboxServiceClientImpl` (with fake counters and `MutableMap`-based repos, etc.) but with less boilerplate code and a bit more structure in place. In general `FakeXYZServiceClient` can mix-in `FakeServiceClient` and gain the ability to create (`makeRepo`) and persist (`save`) to fake repos, in a type-safe manner -- the rest of the methods (if needed) can be implemented on top of the repos.

If the service manages multiple repos (like Heimdal) then we can better structure the `XYZServiceClient` trait by separating out `ABCRepoAccess` (e.g. `ReKeepRepoAccess`), etc. for more readable and easier to maintain code. This is implementation specific and convention-driven (i.e. not enforced).

It's functional but definitely WIP. Comments (on overall approach or details) are appreciated!

@andrewconner @leogrim @stkem @ymatsuda @2is10


kifi-backend/modules/common/test/com/keepit/test/FakeRepoLike.scala
```diff
@@ -0,0 +1,43 @@
...
+import java.util.concurrent.atomic.AtomicLong
+
+import com.keepit.common.db.{ State, Id, Model }
+import collection.mutable.{ Map => MutableMap }
+
+trait FakeRepoLike[T <: Model[T]] {
+
+  def save(model: T): T
+  def get(id: Id[T])
+  def all()
```
yingjieMiao: return type.

raymondng: good catch. will fix.

yingjieMiao:
interesting. lgtm overall. a second pair of eyes might be helpful.

kifi-backend/modules/common/test/com/keepit/test/FakeRepoLike.scala
```diff
@@ -0,0 +1,43 @@
...
+}
+
+class FakeIdCounter[T <: Model[T]] {
+  private val counter = new AtomicLong(0)
+  def nextId(): Id[T] = { Id[T](counter.incrementAndGet) }
+}
+
+trait FakeRepoBase[T <: Model[T]] extends FakeRepoLike[T] {
+  def idCounter: FakeIdCounter[T]
+  def data: MutableMap[Id[T], T]
```
yingjieMiao: why this is `def`?

raymondng: to be a bit more flexible wrt implementation specifics.
in the simple use cases (such as in this pull request) they will be overridden with `val`

andrewconner: `ConcurrentHashMap` perhaps?

kifi-backend/modules/common/test/com/keepit/test/FakeRepoLike.scala
```diff
@@ -0,0 +1,43 @@
...
+
+  def save(model: T): T = {
+    val id = model.id.getOrElse(idCounter.nextId())
+    val updated = model.withId(id)
+    data(id) = updated
+    updated
+  }
+  def get(id: Id[T]) = data(id)
+  def all() = data.values.toSeq
```
andrewconner: If we want very similar to a real db, may want to sort this by ID.

kifi-backend/modules/common/test/com/keepit/test/FakeRepoLike.scala
raymondng: Yes. In fact I was wondering about this -- I noticed that we use `MutableMap`
today in `FakeShoeboxServiceClientImpl` so I followed suit. If you have a
few minutes I'd love to chat about this a bit.

kifi-backend/modules/common/test/com/keepit/test/FakeRepoLike.scala
raymondng: Oh we sort by default? I actually wondered about this one but didn't see it ... missing something?

andrewconner: No, but it does typically come out sorted-ish (think of `select * from user`). The hashing for the Map will effectively shuffle the results. It _shouldn't_ make a difference, but it may be unexpected for a test.

raymondng: I see. In that case I'd tend to think that we shouldn't do so here either -- that might unintentionally lead to code that makes assumptions that aren't legitimate in prod. WDYT?

andrewconner: Works for me.

eishay: if that's the case and we don't have guarantee by the db of any sorting unless we explicitly `order by` something. maybe its best that here we'll return a random list (even forcefully random) to catch implicit assumption about order.

raymondng: DB in general does not sort by default (for performance reasons) -- here due to hashing (as @andrewconner mentioned) it's already shuffled. IMHO it's probably good enough. WDYT?

andrewconner: Yeah this is fine. Just wanted to make sure that we were okay with the shuffle — sounds like it.

eishay: its fine, though i'll be happier to see instead something like

``` scala
scala.util.Random.shuffle(data.values.toSeq)
```

it will guarantee that there's no way that we trust order. since the hash is consistent with strings, ints and other  primitives, if the seq length is only 2 there's 50% that its exactly like the db in all test runs

raymondng: Thanks Eishay. As there's also cost to shuffling, I'd add a comment to that
effect for now. Will be monitoring and will change if hashing proves to be
insufficient.

PR #10288, opened 2014-08-14
Adding index hint to raw seed item lookup query

stkem:
As it turns out the mysql query planner was not using the obvious index for this query causing it to take 1-2s. With the hint it's about 20-40ms.
Also, H2 doesn't support index hints, hence the database dialect shenanigans in the repo.
@yingjieMiao @lawzlo @andrewconner

### DIFF
--- a/kifi-backend/modules/curator/app/com/keepit/curator/model/RawSeedItemRepo.scala
+++ b/kifi-backend/modules/curator/app/com/keepit/curator/model/RawSeedItemRepo.scala
@@ -2,7 +2,7 @@ package com.keepit.curator.model
 
 import com.keepit.common.db.slick.{ DbRepo, SeqNumberFunction, SeqNumberDbFunction, DataBaseComponent, Database }
 import com.keepit.common.db.slick.{ RepoWithDelete, DbRepoWithDelete }
-import com.keepit.common.db.{ Id, DbSequenceAssigner, SequenceNumber }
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/curator/app/com/keepit/curator/model/RawSeedItemRepo.scala
```diff
@@ -79,7 +96,13 @@ class RawSeedItemRepoImpl @Inject() (
   }
 
   def getBySeqNumAndUser(start: SequenceNumber[RawSeedItem], userId: Id[User], maxBatchSize: Int)(implicit session: RSession): Seq[RawSeedItem] = {
-    (for (row <- rows if row.seq > start && (row.userId === userId || row.userId.isNull)) yield row).sortBy(_.seq.asc).take(maxBatchSize).list
+    import StaticQuery.interpolation
+    val q = if (db.dialect == H2DatabaseDialect) {
+      sql"SELECT * FROM raw_seed_item WHERE seq > ${start.value} AND (user_id=$userId OR user_id IS NULL) ORDER BY seq LIMIT $maxBatchSize;"
+    } else {
+      sql"SELECT * FROM raw_seed_item USE INDEX (raw_seed_item_u_seq_user_id) WHERE seq > ${start.value} AND (user_id=$userId OR user_id IS NULL) ORDER BY seq LIMIT $maxBatchSize;"
```
eishay, line 103: :+1:

PR #10291, opened 2014-08-14
FortyTwo->Kifi

eishay:
@andrewconner @2is10

### DIFF
--- a/kifi-backend/modules/shoebox/public/about_us/culture.html
+++ b/kifi-backend/modules/shoebox/public/about_us/culture.html
@@ -13,7 +13,7 @@
 <meta name="viewport" content="width=device-width">
 <meta name="author" content="Alexandru Stoicescu"/>
 <meta name="viewport" content="width=device-width">
-<title>Our Culture at FortyTwo</title>
+<title>Our Culture at Kifi</title>
 <link rel="shortcut icon" type="image/png" href="images/kifi-favi-50.png">
 
 <!--[if lt IE 9]><script src="//html5shiv.googlecode.com/svn/trunk/html5.js"></script><![endif]-->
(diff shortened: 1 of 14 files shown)


kifi-backend/modules/shoebox/public/index.html
```diff
@@ -656,7 +656,7 @@ <h3 class="collection tag-new" data-name="{{name}}">
 							<span class="footer-sep">|</span>
 							<a href="/about">About kifi</a>
 							<span class="footer-sep">|</span>
-							<a href="http://www.42go.com">About FortyTwo</a></li>
+							<a href="https://www.kifi.com/about">About Kifi</a></li>
```
2is10: It looks like there’s already an “About Kifi” link just before this one. Maybe call this one “About us”?
Also, I think you may have intended to link to `/about/mission` instead (and you can omit the scheme and domain).

kifi-backend/modules/shoebox/public/html/index.html
```diff
@@ -1,7 +1,7 @@
 <!doctype html>
-<title>FortyTwo Inc.</title>
-<meta name="answer-to-life-the-universe-and-everything" content="FortyTwo">
-<meta name="entrance" content="jobs@42go.com">
+<title>Kifi</title>
+<meta name="answer-to-life-the-universe-and-everything" content="Kifi">
```
2is10: This joke doesn’t make sense anymore.

kifi-backend/modules/shoebox/public/about_us/culture.html
```diff
@@ -13,7 +13,7 @@
 <meta name="viewport" content="width=device-width">
 <meta name="author" content="Alexandru Stoicescu"/>
 <meta name="viewport" content="width=device-width">
```
2is10: Consider killing duplicate viewport tag and, perhaps, author tag?

2is10:
LGTM

kifi-backend/modules/shoebox/public/about_us/culture.html
andrewconner: And removing Alexandru.

kifi-backend/modules/shoebox/public/html/index.html
andrewconner:
``` html
<meta name="answer-to-life-the-universe-and-everything" content="Kifi. It's funny because the company's real name is actually FortyTwo. 42. Get it?">
```

kifi-backend/modules/shoebox/public/index.html
andrewconner: Perhaps outside of the scope, but we do have two about pages now. Consolidation time soon?

andrewconner:
Looks good.

kifi-backend/modules/shoebox/public/index.html
eishay: can you please take it to +mark@kifi.com ?

eishay: sure @2is10

PR #10292, opened 2014-08-14
Update feedback API per curator specs, Hide url from user for the API

lawzlo:
@stkem @joshm1 @yingjieMiao


lawzlo:
I just update SendgridCommanderTest and SendgridCommander to make tests pass. @joshm1 will help to correct logic for those classes since he has better understanding of how those logic work

kifi-backend/modules/common/app/com/keepit/model/UriRecommendationFeedback.scala
```diff
@@ -3,8 +3,10 @@ package com.keepit.model
...
 
 @json case class UriRecommendationFeedback(
-    seen: Option[Boolean],
-    clicked: Option[Boolean],
-    kept: Option[Boolean]) {
-  override def toString = s"UriRecommendationFeedback(seen:$seen, clicked:$clicked, kept:$kept)"
+    delivered: Option[Int] = None,
+    clicked: Option[Int] = None,
+    kept: Option[Boolean] = None,
+    deleted: Option[Boolean] = None,
```
stkem: Let's call this something else, since there is a notion of 'deletion' at the database level (I know the spec says 'deleted'), which might lead to confusion. How about 'trashed'?

stkem: consistently, including repo and such

stkem:
Made one comment on naming, lgtm otherwise. @joshm1 may want to weigh in on the email test.

kifi-backend/modules/shoebox/app/com/keepit/commanders/SendgridCommander.scala
```diff
@@ -148,7 +148,7 @@ class SendgridCommander @Inject() (
 
     val userId = email.senderUserId.get
     val keepUrl = event.url.get
-    val uriRecoFeedback = UriRecommendationFeedback(clicked = Some(true), seen = None, kept = None)
+    val uriRecoFeedback = UriRecommendationFeedback(clicked = Some(1), delivered = Some(1), kept = None)
```
joshm1: Setting this to 1 will overwrite whatever the existing `clicked` value is instead of incrementing it like I expect we'd prefer.  Should we have a way to pass the message "increment this by 1" instead of "set this to 1" since this was changed to a counter?

kifi-backend/modules/shoebox/conf/shoebox.routes
```diff
@@ -257,7 +257,7 @@ POST    /site/libraries/move        @com.keepit.controllers.website.LibraryContr
 
 GET     /site/recos/adHoc           @com.keepit.controllers.website.RecommendationsController.adHocRecos(n: Int)
 POST    /site/recos/adHoc           @com.keepit.controllers.website.RecommendationsController.adHocRecos(n: Int)
-POST    /site/recos/feedback        @com.keepit.controllers.website.RecommendationsController.updateUriRecommendationFeedback(url: String)
+POST    /site/recos/feedback        @com.keepit.controllers.website.RecommendationsController.updateUriRecommendationFeedback()
```
joshm1, line 263: this breaks the email feed that hardcodes the URL in the query string (since it's a test feed and we don't have access to the reverse router in curator).  Update the email HTML to pass "url" has a hidden value to this endpoint.

kifi-backend/modules/common/app/com/keepit/model/UriRecommendationFeedback.scala
lawzlo: Sure. Will rename it.

kifi-backend/modules/curator/app/com/keepit/curator/model/UriRecommendationRepo.scala
```diff
@@ -74,9 +76,11 @@ class UriRecommendationRepoImpl @Inject() (
   }
 
   def updateUriRecommendationFeedback(userId: Id[User], uriId: Id[NormalizedURI], feedback: UriRecommendationFeedback)(implicit session: RSession): Boolean = {
-    (if (feedback.seen.get) (for (row <- rows if row.uriId === uriId && row.userId === userId) yield (row.seen, row.updatedAt)).update((feedback.seen.get, currentDateTime)) > 0 else true) ||
-      (if (feedback.clicked.get) (for (row <- rows if row.uriId === uriId && row.userId === userId) yield (row.clicked, row.updatedAt)).update((feedback.clicked.get, currentDateTime)) > 0 else true) ||
-      (if (feedback.kept.get) (for (row <- rows if row.uriId === uriId && row.userId === userId) yield (row.kept, row.updatedAt)).update((feedback.kept.get, currentDateTime)) > 0 else true)
+    (if (feedback.delivered.isDefined) (for (row <- rows if row.uriId === uriId && row.userId === userId) yield (row.delivered, row.updatedAt)).update((feedback.delivered.get, currentDateTime)) > 0 else true) &&
```
stkem: Josh already mentioned this in another comment, I missed it earlier: This should increment the value not overwrite it. You may have to write actual SQL for that. Not sure if slick supports it.

kifi-backend/modules/common/app/com/keepit/model/UriRecommendationFeedback.scala
```diff
@@ -3,8 +3,10 @@ package com.keepit.model
 import com.kifi.macros.json
 
 @json case class UriRecommendationFeedback(
-    seen: Option[Boolean],
-    clicked: Option[Boolean],
-    kept: Option[Boolean]) {
-  override def toString = s"UriRecommendationFeedback(seen:$seen, clicked:$clicked, kept:$kept)"
+    delivered: Option[Int] = None,
+    clicked: Option[Int] = None,
```
stkem: These could remain booleans with the increment logic, (False and None mean do nothing, True means increment by one). That way the field is still optional in the json (which it wouldn't be if it were just a boolean) and you avoid the risk of someone sending negative numbers.

lawzlo: I think it's smart. :+1:

kifi-backend/modules/curator/app/com/keepit/curator/model/UriRecommendationRepo.scala
```diff
@@ -74,9 +76,11 @@ class UriRecommendationRepoImpl @Inject() (
...
+      (if (feedback.clicked.isDefined) (for (row <- rows if row.uriId === uriId && row.userId === userId) yield (row.clicked, row.updatedAt)).update((feedback.clicked.get, currentDateTime)) > 0 else true) &&
+      (if (feedback.kept.isDefined) (for (row <- rows if row.uriId === uriId && row.userId === userId) yield (row.kept, row.updatedAt)).update((feedback.kept.get, currentDateTime)) > 0 else true) &&
+      (if (feedback.deleted.isDefined) (for (row <- rows if row.uriId === uriId && row.userId === userId) yield (row.deleted, row.updatedAt)).update((feedback.deleted.get, currentDateTime)) > 0 else true) &&
+      (if (feedback.markedBad.isDefined) (for (row <- rows if row.uriId === uriId && row.userId === userId) yield (row.markedBad, row.updatedAt)).update((feedback.markedBad.get, currentDateTime)) > 0 else true)
```
joshm1: Each of these updates has the same where clause, you can use slick's query composition to factor that out.  This should also be more efficient since slick only has generate that part of the query once.

``` scala
val byUserAndUri = for (row <- rows if row.uriId === uriId && row.userId === userId) yield row
(if (feedback.seen.get) (for (row <- byUserAndUri) yield (row.seen, row.updatedAt).update(...
(if (feedback.clicked.get) (for (row <- byUserAndUri) yield (row.clicked, row.updatedAt).update(...
```

But that's a moot point if you'll have to write SQL to increment.

"There is currently no way to use scalar expressions or transformations of the existing data in the database for updates." - http://slick.typesafe.com/doc/2.0.0/queries.html

I guess it's a matter of code style and what is subjectively more readable, but I find code that make me scroll horizontally this much harder to read and understand.  These lines exceed 200 characters.

lawzlo: Actually. This was a wrong way I made too. It has race condition. If you read a row, and update. Before your updating finish, another read row request come in and update. Your previous update lost.

joshm1: I don't believe there are reads (SELECTs) then updates here. Just a batch of updates.  Every time you call .update(...) it compiles and runs the query `UPDATE <table> SET [column=value, ] WHERE user_id = ? AND uri_id = ?`  I don't see how these could run in parallel either since each update depends on the return value of the previous update, which is an Int instead of a Future.  Even if they did run in parallel,  you'd only overwrite the updated_at column.

I was also going to suggest trying to compose a single update query, but I don't see how to do that with slick.

I'd also extract currentDateTime to a val.

kifi-backend/modules/common/conf/evolutions/shoebox/227.sql
```diff
@@ -0,0 +1,27 @@
+# CURATOR
+
+# --- !Ups
+
+ALTER TABLE uri_recommendation
```
lawzlo: ran this in devo. no syntax error.

kifi-backend/modules/curator/app/com/keepit/curator/model/UriRecommendationRepo.scala
stkem: If you want to do increment with slick you'll have to first read the value and then write it back incremented, which is a race condition since slick does not support `select for update`. Hence the plain SQL.

kifi-backend/modules/common/conf/evolutions/shoebox/227.sql
stkem: why not just rename it?

kifi-backend/modules/shoebox/conf/shoebox.routes
stkem, line 263: bumping this comment

kifi-backend/modules/curator/app/com/keepit/curator/model/UriRecommendationRepo.scala
```diff
@@ -74,9 +80,20 @@ class UriRecommendationRepoImpl @Inject() (
   }
 
   def updateUriRecommendationFeedback(userId: Id[User], uriId: Id[NormalizedURI], feedback: UriRecommendationFeedback)(implicit session: RSession): Boolean = {
-    (if (feedback.seen.get) (for (row <- rows if row.uriId === uriId && row.userId === userId) yield (row.seen, row.updatedAt)).update((feedback.seen.get, currentDateTime)) > 0 else true) ||
-      (if (feedback.clicked.get) (for (row <- rows if row.uriId === uriId && row.userId === userId) yield (row.clicked, row.updatedAt)).update((feedback.clicked.get, currentDateTime)) > 0 else true) ||
-      (if (feedback.kept.get) (for (row <- rows if row.uriId === uriId && row.userId === userId) yield (row.kept, row.updatedAt)).update((feedback.kept.get, currentDateTime)) > 0 else true)
+    import StaticQuery.interpolation
+
+    (if (feedback.delivered.isDefined && feedback.delivered.get)
+      sql"UPDATE uri_recommendation SET delivered=delivered+1, updated_at=$currentDateTime WHERE user_id=$userId AND uri_id=${uriId}".asUpdate.first > 0
```
joshm1, line 86: since we have to write raw SQL why not compose a single query string for all possible updates instead of a separate for each column?

kifi-backend/modules/common/conf/evolutions/shoebox/227.sql
lawzlo: Didn't find play evolution rename syntax document, the example I found in our evolution files only have change name. Here I change both name and type. Drop and Add Just a easy and safe way for me to do.

kifi-backend/modules/curator/app/com/keepit/curator/model/UriRecommendationRepo.scala
```diff
@@ -74,9 +80,20 @@ class UriRecommendationRepoImpl @Inject() (
   }
 
   def updateUriRecommendationFeedback(userId: Id[User], uriId: Id[NormalizedURI], feedback: UriRecommendationFeedback)(implicit session: RSession): Boolean = {
-    (if (feedback.seen.get) (for (row <- rows if row.uriId === uriId && row.userId === userId) yield (row.seen, row.updatedAt)).update((feedback.seen.get, currentDateTime)) > 0 else true) ||
-      (if (feedback.clicked.get) (for (row <- rows if row.uriId === uriId && row.userId === userId) yield (row.clicked, row.updatedAt)).update((feedback.clicked.get, currentDateTime)) > 0 else true) ||
-      (if (feedback.kept.get) (for (row <- rows if row.uriId === uriId && row.userId === userId) yield (row.kept, row.updatedAt)).update((feedback.kept.get, currentDateTime)) > 0 else true)
+    import StaticQuery.interpolation
+
+    (if (feedback.delivered.isDefined && feedback.delivered.get)
```
stkem: Queries look correct to me, but please add a simple repo test for this method to flush out any potential syntax errors (in the SQL) and such.
Also, the chained logic (`&&`) is a bit hard to read. I suggest putting every query result into it's own variable and `&&` them together in the end. The compiler should be optimizing that away so there will be no overhead.

kifi-backend/modules/curator/app/com/keepit/curator/model/UriRecommendationRepo.scala
joshm1, line 86: generally unsafe to inject variables directly into the query.  Also, looks like you would get an invalid query like: `updated_at=2014-08-21 15:00` 

Use a parameterized query where the SQL looks like:

`updated_at = ?`

and you pass the parameter value(s) when you execute the query

kifi-backend/modules/curator/app/com/keepit/curator/model/UriRecommendationRepo.scala
lawzlo: There is a test for set clicked to None and delivered to Some(true). Test result is good. Maybe will add another test case to cover more logic. And will assign a val to each update query.

kifi-backend/modules/curator/app/com/keepit/curator/model/UriRecommendationRepo.scala
stkem, line 86: actually the `sql` part in front of the string takes care of that magically (i.e. with a macro)

joshm1, line 86: Cool, I didn't even notice that :)

kifi-backend/modules/shoebox/conf/shoebox.routes
lawzlo, line 263: @joshm1 can you point the name is email HTML class?

lawzlo, line 263: For hide url from user, I think it's better to just have one endpoint instead of create another endpoint contains explicit url for testers.
@joshm1 can you help to update the email html to have url in the body since you worked on that part?

lawzlo:
tested, pushing to prod.

PR #10293, opened 2014-08-15
Google OAuth2 update

raymondng:
Updating scopes only. Thanks @andrewconner for the tip about deprecation of `userinfo.email`

This takes care of the easy part, the other part about getting profile info is on the TODO list.

As a bonus, found that Google Contacts recently (finally!) added a readonly scope -- adopted as well.

Tested locally (with real google accounts)

@andrewconner @eishay


eishay:
nice

PR #10294, opened 2014-08-15
Enable users to edit name on kifi.com

yipingliao:
@2is10 cc @andrewconner 

Still to do: I'd like to get Jen to vet the actual experience of editing, canceling edits, etc., to make sure that the UI is intuitive, but those should be minor changes.

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/profile/profile.js
+++ b/kifi-backend/modules/shoebox/angular/src/profile/profile.js
@@ -6,6 +6,7 @@ angular.module('kifi.profile', [
   'util',
   'kifi.profileService',
   'kifi.profileInput',
+  'kifi.profileNameInput',
   'kifi.routeService',
   'kifi.profileEmailAddresses',
   'kifi.profileChangePassword',
@@ -49,6 +50,14 @@ angular.module('kifi.profile', [
       $scope.emailInput.value = val || '';
     });
 
+    $scope.name = {};
+    $scope.$watch('me.firstName', function (val) {
+      $scope.name.firstName = val;
+    });
+    $scope.$watch('me.lastName', function (val) {
+      $scope.name.lastName = val;
+    });
+
     $scope.addEmailInput = {};
 
     $scope.logout = function () {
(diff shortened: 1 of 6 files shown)


kifi-backend/modules/shoebox/angular/src/profile/profile.js
```diff
@@ -61,6 +70,14 @@ angular.module('kifi.profile', [
       });
     };
 
+    $scope.validateName = function (name) {
+      return profileService.validateNameFormat(name);
+    };
+
+    $scope.saveName = function (name) {
+      profileService.setNewName(name);
```
yipingliao, line 78: I am being very optimistic here about being able to save the name on the server; no error-detection whatsoever. :) Actually, @andrewconner , my question is: what kind of network/server errors can I expect from this operation?

kifi-backend/modules/shoebox/angular/src/profile/profileNameInput.tpl.html
```diff
@@ -0,0 +1,24 @@
+<div class="profile-input-box">
+  <!-- First Name -->
```
yipingliao, line 2: Let me know if these comments in HTML are too verbose. They help my eyes scan, but they do add to the size of the templates.

kifi-backend/modules/shoebox/angular/src/profile/profile.js
andrewconner, line 78: :) Right now, it's not ideal — you likely won't get any errors, and it'll be silent. _However_, it's on my list to write a new set of endpoints that lets you explicitly update parts of a user. It will have full validation and error handling.

yipingliao, line 78: Nice! Let me know when you do.

yipingliao:
Forgot to link to Asana task: https://app.asana.com/0/15014070792041/14750997211449
(wireframes in task)

kifi-backend/modules/shoebox/angular/src/profile/profileNameInput.tpl.html
2is10, line 2: The need for them could be considered a symptom of a different problem—that we’re using too much HTML.
I wonder if we couldn’t get away with a single input element and a single error element for each name. I understand that you may be going along with existing conventions for this page, so don’t feel the need to change any of this. Just saying it might be something to consider going forward, when working on new stuff.

2is10:
Cool, nice work. Feel free to merge.

PR #10297, opened 2014-08-15
feed explanation wanderer, integrated with curator

yingjieMiao:
@stkem  basic structure for review.

some implementation details need to tweak.


kifi-backend/modules/common/test/com/keepit/graph/FakeGraphServiceClientImpl.scala
```diff
@@ -10,7 +10,7 @@ import com.keepit.common.zookeeper.ServiceCluster
 import com.keepit.common.net.HttpClient
 import com.keepit.common.healthcheck.AirbrakeNotifier
 import com.keepit.graph.wander.{ Collisions, Wanderlust }
-import com.keepit.graph.model.{ RelatedEntities, ConnectedUserScore, ConnectedUriScore, GraphKinds }
+import com.keepit.graph.model._
```
stkem, line 13: Please try to avoid `._` imports when possible (I know intellij tends to add them automatically sometimes). It makes it harder to understand where things are coming from and you might be getting implicits you don't want.

kifi-backend/modules/graph/app/com/keepit/graph/wander/FeedExplanationCommander.scala
```diff
@@ -0,0 +1,120 @@
...
+
+object FeedExplanationWanderLust {
+
+  def wander(userId: Id[User]) =
+    Wanderlust(
+      startingVertexKind = "User",
+      startingVertexDataId = userId.id,
+      preferredCollisions = Set(),
+      avoidTrivialCollisions = true,
+      steps = 2000,
```
stkem: perhaps a bit low? I suppose it's worth a try though.

kifi-backend/modules/curator/app/com/keepit/curator/model/CuratorKeepInfoRepo.scala
```diff
@@ -49,4 +50,8 @@ class CuratorKeepInfoRepoImpl @Inject() (
     (for (row <- rows if row.uriId === uriId && row.state === CuratorKeepInfoStates.ACTIVE && row.discoverable) yield row.id).firstOption.isDefined
   }
 
+  def getUserURIsAndKeeps(userId: Id[User])(implicit session: RSession): Seq[(Id[NormalizedURI], Id[Keep])] = {
+    (for (r <- rows if r.userId === userId && r.state === CuratorKeepInfoStates.ACTIVE) yield (r.uriId, r.keepId)).list
```
stkem, line 54: This table currently has no index on `userId`, you'll need to add one to make this query work well.

kifi-backend/modules/graph/app/com/keepit/graph/wander/FeedExplanationCommander.scala
yingjieMiao: I agree. maybe 10K? don't want to kill graph though. it's not to hard to come up with a time estimate.

kifi-backend/modules/graph/app/com/keepit/graph/wander/FeedExplanationCommander.scala
```diff
@@ -0,0 +1,120 @@
...
+import com.keepit.graph.manager.GraphManager
+import com.keepit.graph.model._
+import com.keepit.model.{ Keep, User, NormalizedURI }
+import scala.collection.mutable.ListBuffer
+import scala.concurrent.Future
+import scala.concurrent.duration._
+import play.api.libs.concurrent.Execution.Implicits.defaultContext
+
+@Singleton
+class FeedExplanationCommander @Inject() (graph: GraphManager, clock: Clock) extends Logging {
```
stkem, line 17: I think it would be a good idea to have a short test for this class on the graph side. I think we have some helpers to set up a small fake graph, so it should not be too hard, hopefully.

kifi-backend/modules/curator/app/com/keepit/curator/model/CuratorKeepInfoRepo.scala
yingjieMiao, line 54: good point.

kifi-backend/modules/graph/app/com/keepit/graph/wander/FeedExplanationCommander.scala
yingjieMiao, line 17: I see. what does the test do? make sure no run time error? It's not trivial to check the math.

stkem:
Made a couple of comments. lgtm overall (index is essential though).
(@leogrim will probably have some comments (post merge) when he gets back.)

stkem:
sorry this took so long for review

kifi-backend/modules/graph/app/com/keepit/graph/wander/FeedExplanationCommander.scala
stkem, line 17: yup, basically just run time. You could also set up a fake graph were there is only one connection and make sure it's found.

yingjieMiao, line 17: i will add a test. should be fun.

stkem, line 17: :-)

yingjieMiao:
@stkem  updated with a test. tweak the num of steps to 5K.

PR #10299, opened 2014-08-15
Add user status tracking to user_clicked_page

yipingliao:
Also, add experiment tracking to user_clicked_page and user_viewed_page.

Task: https://app.asana.com/0/15014070792041/11359935698324

@stkem cc @2is10

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/common/services/mixpanelService.js
+++ b/kifi-backend/modules/shoebox/angular/src/common/services/mixpanelService.js
@@ -50,6 +50,21 @@
       return userStatus;
     }
 
+    function getExperiments() {
+      var experiments = [];
+      if (profileService && profileService.me && profileService.me.experiments) {
+        experiments = profileService.me.experiments || [];
+      }
+
+      // Remove any user status from the list of experiments. Currently user status
+      // is a kind of experiment. TODO: move user status out of experiments.
+      var realExperiments = _.reject(experiments, function (experiment) {
+        return (experiment === 'fake') || (experiment === 'admin') || (experiment === 'standard');
+      });
+
+      return realExperiments;
+    }
+
     function pageTrackForUser(mixpanel, path, origin) {
       if (userId) {
         var oldId = mixpanel.get_distinct_id && mixpanel.get_distinct_id();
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/angular/src/common/services/mixpanelService.js
```diff
@@ -50,6 +50,21 @@
       return userStatus;
     }
 
+    function getExperiments() {
+      var experiments = [];
```
yipingliao: Mixpanel says that they take arrays in properties (https://mixpanel.com/docs/properties-or-segments/property-data-types). We can see in a few days whether that breaks down nicely.

2is10: I think it will. We already send an experiments array with events originating from the backends.

stkem: Yup, this should work nicely.

kifi-backend/modules/shoebox/angular/src/common/services/mixpanelService.js
```diff
@@ -50,6 +50,21 @@
...
+        experiments = profileService.me.experiments || [];
+      }
+
+      // Remove any user status from the list of experiments. Currently user status
+      // is a kind of experiment. TODO: move user status out of experiments.
+      var realExperiments = _.reject(experiments, function (experiment) {
+        return (experiment === 'fake') || (experiment === 'admin') || (experiment === 'standard');
+      });
+
+      return realExperiments;
```
2is10: This could be a little more more concise and functional in style:

```
var experiments = profileService && profileService.me && profileService.me.experiments || [];
return _.without(experiments, 'fake', 'admin', 'standard');
```

I think we should also only start tracking once we know who the user is. I think this code should be restructured to stop checking `profileService && profileService.me`. If there’s a genuine race condition, it should queue tracking requests until the race condition is avoided. You don’t need to work on this—just FYI.

2is10:
nice work! lgtm

kifi-backend/modules/shoebox/angular/src/common/services/mixpanelService.js
yipingliao: Thanks! Will fix.

I'll make an Asana task to work on the restructuring bit.

stkem:
lgtm

PR #10300, opened 2014-08-15
moving our branding art to the console, removing optional <html>, <head>, <body> tags

2is10:
view diff with `?w=0` to see the effective changes more clearly

@andrewconner


2is10:
lgtm thanks

PR #10306, opened 2014-08-15
mobile endpoint and tests for PYMK

joshm1:
Extracted setup code to ShoeboxTestFactory as a way to make the setup code in our tests more DRY, which I also added as a def in ShoeboxTestInjector.

In many tests, setup code is often duplicated and takes up the much more LOC than the actual test code.  I think we should consider doing this more often, especially for common use cases like creating a users and user connections.

@stkem @andrewconner  @davoda


stkem:
:+1:

PR #10307, opened 2014-08-15
fixing GetResult field order

stkem:
this needs to match the database column order, otherwise very strange exception can result (in this case a NPE in our DateTime TypeMapper)
@lawzlo @yingjieMiao


yingjieMiao:
my understanding is that `def *` need to be in the same order as the case class. (I used to have two Boolean fields come together and switched order in the `def *` which caused bugs in Dev mode) `GetResult`'s `PositionedResult` comes in same order as your plain SQL query. Since you are using plain SQL `select *`, it comes in db column order, and slick's `def *` was not used. 

So in this case, I guess all you need to change is the `GetResult` part.

stkem:
Yeah, unfortunately the tests use the slick table definition to create the table, which causes it to have a different column order if the order of the field in the class does not match the column order as created by the evolutions. Thus in oder to make the `GetResult` work in both the tests and in prod this was needed.

yingjieMiao:
Ah, I see. Another solution is to use more verbose select, instead of 
select *.

stkem:
Good point. I didn't think of that.

PR #10310, opened 2014-08-15
eliza error reporting on failed (but authenticated) websocket request take 2

stkem:
same code as before (revert revert didn't work for some reason, got a github error)
@2is10


2is10:
lgtm

PR #10311, opened 2014-08-15
restructuring mixpanelService a bit

2is10:
@stkem cc @yipingliao 

This eliminates all of the tiptoeing around values injected via `.run`. They are safely initialized before any tracking occurs.

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/common/services/mixpanelService.js
+++ b/kifi-backend/modules/shoebox/angular/src/common/services/mixpanelService.js
@@ -1,9 +1,6 @@
-/**
-* inspired by https://github.com/luisfarzati/angulartics/blob/master/src/angulartics-mixpanel.js
-*/
-(function (angular) {
+// inspired by https://github.com/luisfarzati/angulartics/blob/master/src/angulartics-mixpanel.js
+(function () {
   'use strict';
-  var $window, $log, profileService, analyticsState;
 
   /**
    * @name kifi.mixpanel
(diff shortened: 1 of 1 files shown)


2is10:
view the pull request with `?w=0` in the URL for a smaller diff

kifi-backend/modules/shoebox/angular/src/common/services/mixpanelService.js
```diff
@@ -12,150 +9,131 @@
...
-        if (profileService && $window) {
-          var mixpanel = $window.mixpanel;
-          var origin = $window.location.origin;
-          pageTrackForVisitor(mixpanel, path, origin);
-          pageTrackForUser(mixpanel, path, origin);
-        }
-      });
-    });
-
-    angulartics.waitForVendorApi('mixpanel', 5000, function (/*mixpanel*/) {
```
2is10, line 134: These `.waitForVendorApi` calls were accomplishing nothing because we define a stub `window.mixpanel` before any of our Angular code ever runs.

kifi-backend/modules/shoebox/angular/src/common/services/mixpanelService.js
```diff
@@ -12,150 +9,131 @@
...
-      helpRankClicks: /^\/helprank\/clicks?$/,
-      helpRankReKeeps: /^\/helprank\/rekeeps?$/
-    };
-
-    function getLocation(path) {
-      for (var loc in locations) {
-        if (locations[loc].test(path)) {
-          return loc;
-        }
+      $analyticsProvider.firstPageview(false);
```
2is10, line 12: This line is new. It fixes double tracking of the initial page view.

kifi-backend/modules/shoebox/angular/src/common/services/mixpanelService.js
```diff
@@ -12,150 +9,131 @@
...
-          userStatus = 'fake';
-        }
-        else if (experiments.indexOf('admin') > -1) {
-          userStatus = 'admin';
+    return path;
+  }
+
+  function getExperiments() {
+    // 'fake' and 'admin' experiments are tracked separately in "userStatus" field
+    return _.without(profileService.me && profileService.me.experiments || [], 'fake', 'admin');
```
2is10, line 72: @yipingliao I edited the comment, removing the TODO, since I think it’s unlikely that we’ll decide to flag fake and admin users in a different way. "userStatus" is just a concept we invented for mixpanel analytics. Now that we’re sending all experiments, we may eventually discontinue the "userStatus" field.

Also, there is no `standard` experiment, so no need to filter it out.

yipingliao:
LGTM. Thanks!

kifi-backend/modules/shoebox/angular/src/common/services/mixpanelService.js
stkem, line 134: This seems correct, but I remember that there was some subtlety to this. Bringing in @eishay who I believe wrote this, just in case.

stkem:
lgtm, modulo possible mixpanel load issues (other comment)

2is10:
Merging because I’ve read the `waitForVendorApi` source and tested this locally. No issues. Happy to address any concerns that may surface if/when they do.

PR #10313, opened 2014-08-15
Quick for for iPhone version parsing failing

stkem:
The server is not detecting correctly that it’t an iOS client and thus
trying to parse it like the Browser Extension. Looks to be a UserAgent
issue.
@thassss can you figure out what UserAgent is being sent wiht the websocket request? (Not urgent, I know you've got a lot going on with the iOS 8 release right now)
cc @2is10


2is10:
LGTM. Lesson: Never trust smug-looking regular expressions.

PR #10321, opened 2014-08-15
load Kifi app on iPhone if it is installed, redirect to app store if not

joshm1:
@2is10 @andrewconner @stkem @davoda 

tested this URL, which is the HTML body from the test case: http://assets-b-prod.s3.amazonaws.com/test.html

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/HomeController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/HomeController.scala
@@ -121,7 +121,8 @@ class HomeController @Inject() (
     context.addRequestInfo(request)
     context += ("type", "landing")
     heimdalServiceClient.trackEvent(AnonymousEvent(context.build, EventType("visitor_viewed_page")))
-    Redirect("https://itunes.apple.com/app/id740232575")
+    val uriNoProto = request.uri.replaceFirst("^https?://", "")
+    Ok(views.html.mobile.iPhoneRedirect(uriNoProto))
   }
 
   def mobileLanding = HtmlAction(authenticatedAction = mobileLandingHandler(_), unauthenticatedAction = mobileLandingHandler(_))
--- a/kifi-backend/modules/shoebox/app/views/mobile/iPhoneRedirect.scala.html
+++ b/kifi-backend/modules/shoebox/app/views/mobile/iPhoneRedirect.scala.html
@@ -1,0 +1,43 @@
+@(url: String)
+
+<!DOCTYPE html>
+<html>
+<head>
+<title>Kifi</title>
+</head>
+<body>
+<p style="margin:3em auto;max-width:400px;">
+Loading Kifi app...
+<br/><br/>
+Kifi is the easiest way to organize and share the pages that matter to you online. Find something online that you love?
+Just keep the page, and it's instantly stored and organized in your kifi library where it's accessible from your
+favorite search engine, along with great finds from your friends' libraries, too.
+</p>
+<!-- source: http://aawaara.com/post/74543339755/smallest-piece-of-code-thats-going-to-change-the-world -->
+<script type="text/javascript">
+var timer, heartbeat, lastInterval;
+
+function clearTimers() {
+    clearTimeout(timer);
+    clearTimeout(heartbeat);
+}
+
+window.addEventListener('pageshow', clearTimers, false);
+window.addEventListener('pagehide', clearTimers, false);
+
+function getTime() { return (new Date()).getTime(); }
+
+// for browsers except Safari (which do not support pageshow and pagehide properly)
+function intervalHeartbeat() {
+    var now = getTime(), diff = now - lastInterval - 200;
+    lastInterval = now;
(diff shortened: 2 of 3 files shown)


andrewconner:
Cool! Just played with it, I like how it stops the 2nd redirect if the 1st worked.

kifi-backend/modules/shoebox/app/views/mobile/iPhoneRedirect.scala.html
```diff
@@ -0,0 +1,43 @@
...
+</head>
+<body>
+<p style="margin:3em auto;max-width:400px;">
+Loading Kifi app...
+<br/><br/>
+Kifi is the easiest way to organize and share the pages that matter to you online. Find something online that you love?
+Just keep the page, and it's instantly stored and organized in your kifi library where it's accessible from your
+favorite search engine, along with great finds from your friends' libraries, too.
+</p>
+<!-- source: http://aawaara.com/post/74543339755/smallest-piece-of-code-thats-going-to-change-the-world -->
```
2is10, line 16: I read the post. Most of the code there (and here) is for preventing the redirect from happening again if the user navigates back to the page, since the blog post author is trying to handle the case when a navigation is triggered by clicking a link on a meaningful page. I think we don’t need that code. If the user navigates back from where this page sends them, I think he/she would be wanting to get back to the previous page, not this temporary interstitial page, so we probably want to use `location.replace` instead. Then we can also simply remove the `pageshow`, `pagehide`, and interval code. We only need the timeout.

kifi-backend/modules/shoebox/app/views/mobile/iPhoneRedirect.scala.html
```diff
@@ -0,0 +1,43 @@
...
+<head>
+<title>Kifi</title>
+</head>
+<body>
+<p style="margin:3em auto;max-width:400px;">
+Loading Kifi app...
+<br/><br/>
+Kifi is the easiest way to organize and share the pages that matter to you online. Find something online that you love?
+Just keep the page, and it's instantly stored and organized in your kifi library where it's accessible from your
+favorite search engine, along with great finds from your friends' libraries, too.
```
2is10, line 14: They should be on the page less than 2 seconds, so will most likely not have a chance to read the content. Some people may fear they may have missed out on something important. We might want to consider just showing the “Loading Kifi app...” part.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/HomeController.scala
```diff
@@ -121,7 +121,8 @@ class HomeController @Inject() (
     context.addRequestInfo(request)
     context += ("type", "landing")
     heimdalServiceClient.trackEvent(AnonymousEvent(context.build, EventType("visitor_viewed_page")))
-    Redirect("https://itunes.apple.com/app/id740232575")
+    val uriNoProto = request.uri.replaceFirst("^https?://", "")
```
2is10, line 124: The protocol (scheme) ends with the colon, so just before the `//` would be a more natural breaking point.

2is10:
nice work

kifi-backend/modules/shoebox/app/views/mobile/iPhoneRedirect.scala.html
joshm1, line 16: Thanks for the input.  1 current issue with this impl. is that it doesn't close the browser tab they are on.  I didn't think of trying location.replace as an option (and possible solution to the open tab issue).  I don't know what location.replace will do if iOS doesn't load the app.  The timeout is there to prevent the app store from loading If the kifi:// url is found.

kifi-backend/modules/shoebox/app/views/mobile/iPhoneRedirect.scala.html
joshm1, line 14: Good point!

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/HomeController.scala
joshm1, line 124: will do!

kifi-backend/modules/shoebox/app/views/mobile/iPhoneRedirect.scala.html
2is10, line 16: I see. Not sure we’ll be able to close the browser tab. Browsers’s typically only let scripts close windows (tabs) that they’ve opened. Could be nice if iOS makes an exception for pages that launch an application.

As I read it, the timeout is here to load the app store if the kifi: url is _not_ found. The `pagehide` handler is there to prevent the app store page from loading if the kifi: url _is_ handled. I’m pretty certain the `pageshow` and interval code have no application to our use case. I think this code will be clearer if we eliminate any parts that are irrelevant/useless to us. Happy to discuss in person if you’d like.

If there’s no way for us to close the tab and the user comes back to this page later (`pageshow`), we may want to execute `history.back()` and if that fails, show a button/link that would give the user the option to re-execute the kifi: navigation.

PR #10322, opened 2014-08-15
Fix UI interactions in profile input boxes.

yipingliao:
@2is10 Thanks for your help on this!
Asana task: https://app.asana.com/0/15014070792041/15496334027991

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/profile/profile.js
+++ b/kifi-backend/modules/shoebox/angular/src/profile/profile.js
@@ -83,10 +83,6 @@ angular.module('kifi.profile', [
     };
 
     $scope.saveEmail = function (email) {
-      if ($scope.me && $scope.me.primaryEmail.address === email) {
-        return profileService.successInputActionResult();
-      }
-
       return getEmailInfo(email).then(function (result) {
         return checkCandidateEmailSuccess(email, result.data);
       }, function (result) {
--- a/kifi-backend/modules/shoebox/angular/src/profile/profile.styl
+++ b/kifi-backend/modules/shoebox/angular/src/profile/profile.styl
@@ -931,11 +931,11 @@ td.profile-email-separator
   .profile-input-container
     position: relative
     display: inline-block
-    width: 258px
+    width: 50%
 
     &:first-child
       margin-right: 20px
-      width: 220px
+      width: calc(50% - 20px)
 
     // Force focus-like styles on both the first name and last name
     // input fields when in editing mode, even though only one name field
--- a/kifi-backend/modules/shoebox/angular/src/profile/profile.styl
+++ b/kifi-backend/modules/shoebox/angular/src/profile/profile.styl
@@ -931,11 +931,11 @@ td.profile-email-separator
   .profile-input-container
     position: relative
     display: inline-block
-    width: 50%
+    width: 53%
 
     &:first-child
       margin-right: 20px
-      width: calc(50% - 20px)
+      width: calc(47% - 20px)
 
     // Force focus-like styles on both the first name and last name
     // input fields when in editing mode, even though only one name field
--- a/kifi-backend/modules/shoebox/angular/src/profile/profile.styl
+++ b/kifi-backend/modules/shoebox/angular/src/profile/profile.styl
@@ -927,7 +927,7 @@ td.profile-email-separator
   margin-right: auto
   width: 280px
 
-.profile-name-input
+.profile-name-input-box
   .profile-input-container
     position: relative
     display: inline-block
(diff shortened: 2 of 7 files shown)


kifi-backend/modules/shoebox/angular/src/profile/profile.js
```diff
@@ -83,10 +83,6 @@ angular.module('kifi.profile', [
     };
 
     $scope.saveEmail = function (email) {
-      if ($scope.me && $scope.me.primaryEmail.address === email) {
```
yipingliao, line 86: Moved the check for the same email out into the directive.

kifi-backend/modules/shoebox/angular/src/profile/profile.styl
```diff
@@ -927,7 +927,7 @@ td.profile-email-separator
   margin-right: auto
   width: 280px
 
-.profile-name-input
+.profile-name-input-box
```
yipingliao, line 930: Renamed some class names for consistency across name and email inputs.

kifi-backend/modules/shoebox/angular/src/profile/profile.tpl.html
```diff
@@ -20,7 +20,7 @@
           </div>
           <div kf-modal show="showEmailChangeDialog.value">
             <div kf-basic-modal-content with-cancel action-text="Continue" action="confirmSaveEmail()" cancel="cancelSaveEmail()">
-              <p>You are about to change your primary email to <span class="email-address">{{me.primaryEmail.address}}</span>.</p>
```
yipingliao, line 23: Fixed bug where the new email address displayed in the confirmation modal is actually the old email address.

kifi-backend/modules/shoebox/angular/src/profile/profileInput.js
```diff
@@ -17,75 +17,86 @@ angular.module('kifi.profileInput', ['util', 'kifi.profileService'])
...
-          })
           .on('focus', function () {
             $timeout(function () { setEditState(); });
           });
 
-        function cancelCancelEdit() {
-          if (cancelEditPromise) {
-            $timeout.cancel(cancelEditPromise);
-            cancelEditPromise = null;
+        element.find('.profile-input-save')
```
yipingliao: In the case where the user tabs into the save button, and then tabs again to focus on something else, cancel any edits.

kifi-backend/modules/shoebox/angular/src/profile/profileNameInput.js
```diff
@@ -48,36 +66,37 @@ angular.module('kifi.profileNameInput', ['util'])
           }
           scope.errorHeader = error.header || '';
           scope.errorBody = error.body || '';
-        };
+        }
 
-        var clearInputError = function () {
```
yipingliao, line 53: Converted a bunch of function expressions to function declarations because our code base in general uses function declarations.

kifi-backend/modules/shoebox/angular/src/profile/profile.tpl.html
2is10, line 23: Wow, bad bug. Good fix.

kifi-backend/modules/shoebox/angular/src/profile/profileInput.js
2is10: Shift-Tab would take focus back to the input. Don’t want to exit edit mode in that case.
You could listen for focus events on `$window` instead (with capture) and exit when it gets to a target outside the edit region.

kifi-backend/modules/shoebox/angular/src/profile/profileInput.js
```diff
@@ -17,75 +17,86 @@ angular.module('kifi.profileInput', ['util', 'kifi.profileService'])
...
+          // When user clicks outside the input and save button, cancel edits.
+          if (!event.target.classList.contains('profile-email-input') &&
+            !event.target.classList.contains('profile-input-save')) {
+            $document.off('mousedown', onClickOutsideInput);
+            scope.$apply(cancel);
           }
         }
 
-        function updateValue(value) {
+        function updateValue (value) {
```
2is10: our style is to use a space after the keyword `function`, but not after a function’s name

2is10:
Nice work!  LGTM.

yipingliao:
Thanks for the comments! I've made the suggested changes, and tossed in another small bug fix (the first and last names in the profile edit view were showing up as separate lines in less-wide screens because I hardcoded their widths without realizing that the middle column uses percentage widths). Going to merge this.

2is10:
:+1:

PR #10325, opened 2014-08-16
non-kifi users can be invited to Libraries

ahsu1230:
added email_address field to LibraryInvites & tested
also added a numKeeps to LibraryInfo

@andrewconner


andrewconner:
Cool, will review. Are we including the `authToken` as well (primary for email invites)?

ahsu1230:
just added!

andrewconner:
Great! Will review tonight.

ahsu1230:
thanks!

andrewconner:
Sorry for the slow review! Looks great. When you're next in here, mind putting an index on the email address and auth token?

ahsu1230:
oh ok sure thing!

PR #10327, opened 2014-08-16
Include attribution info with beta feed

stkem:
This is not very efficient (in particle the basic user lookup), but
should not live very long anyway.
@lawzlo @yingjieMiao


yingjieMiao:
lgtm

PR #10335, opened 2014-08-18
Test libraries

ahsu1230:
@andrewconner


kifi-backend/modules/common/conf/evolutions/shoebox/231.sql
```diff
@@ -3,10 +3,10 @@
...
 ALTER TABLE library_invite
-  ADD COLUMN email_address NULL;
+  ADD COLUMN email_address varchar(50) NULL;
 
 ALTER TABLE library_invite
-  ADD COLUMN auth_token NULL;
+  ADD COLUMN auth_token varchar(50) NULL;
 
 
 INSERT INTO evolutions (name, description) VALUES('231.sql', 'add email address and authentication token fields to library invites');
```
ahsu1230, line 12: now applied correctly

PR #10337, opened 2014-08-18
Refactor of modules

andrewconner:
This removes all the split modules from our codebase. After a discussion with several other engineers, we felt our split-module approach was hindering us and introducing unnecessary complexity.

A bit of background:
- An AngularJS module is like a package (jar). Modules can be created in a declarative manner, and also added to from many files.
- To "import" a parent module, it is listed during child module creation.
- Currently, we a new module for every file. This makes organization and dependency management a bit easier — given a file, it's simple to see what its dependencies are.
- Modules do not specifically help with any features such as lazy loading or organization.
- We currently organize code using many files, split by top level functionality. Our app needs all files to function: we do not have lazy loading support yet.
- The dependency tree started to look like a Set, rather than a balanced tree. See the `app.js` diff.
- This was a first sweep, putting everything in `kifi`. If you're working on a very self-contained module, feel free to split it.

Orthogonal to this: functional modules so we can mode towards lazy loading later — right now, AngularJS modules do not support this. I'm still investigating the best options, input greatly appreciated.

@stkem @joshm1 @ahsu1230 @2is10 @yipingliao


stkem:
:+1:

joshm1:
:+1:

yipingliao:
Yay!

PR #10341, opened 2014-08-18
More informative reporting when  a user tries to connect with a bad client version

stkem:
\+ letting them in right now

@2is10 @thassss @jaredpetker @eishay


2is10:
:+1:

PR #10342, opened 2014-08-18
update iPhone redirect page

joshm1:
@2is10 

the pageshow and pagehide events are useful for safari.  since location.replace() doesn't work in mobile, these are a good way to prevent the auto-redirect when the user navigates back/forward.  The content of the body is now just a button consistent with the Sign Up button.

### DIFF
--- a/kifi-backend/modules/shoebox/app/views/mobile/iPhoneRedirect.scala.html
+++ b/kifi-backend/modules/shoebox/app/views/mobile/iPhoneRedirect.scala.html
@@ -3,23 +3,48 @@
 <!DOCTYPE html>
 <html>
 <head>
+<meta name="viewport" content="width=device-width, user-scalable=no">
 <title>Kifi</title>
+<style type="text/css">
+#app-btn {
+font-family: 'Helvetica Neue', 'Open Sans', Arial, sans-serif;
+background: #93c059;
+text-align: center;
+color: #fff;
+display: inline-block;
+vertical-align: middle;
+padding: 0 1em;
+border: 1px solid #ccc;
+border-radius: 4px;
+font-weight: 400;
+height: 1.9em;
+line-height: 1.9em;
+font-size: 1.5em;
+text-decoration: none;
+outline: none;
+cursor: pointer;
+zoom: 1;
+margin: 0 auto;
+}
+</style>
 </head>
-<body>
-<p style="margin:3em auto;max-width:400px;">
-Loading Kifi app...
-<br/><br/>
-Kifi is the easiest way to organize and share the pages that matter to you online. Find something online that you love?
-Just keep the page, and it's instantly stored and organized in your kifi library where it's accessible from your
-favorite search engine, along with great finds from your friends' libraries, too.
-</p>
+<body style="display:none;text-align:center;padding:3em 0;">
+
+<div><a id="app-btn" href="https://www.kifi.com" onclick="return open();">Loading Kifi App...</a></div>
+
 <!-- source: http://aawaara.com/post/74543339755/smallest-piece-of-code-thats-going-to-change-the-world -->
 <script type="text/javascript">
-var timer, heartbeat, lastInterval;
+var timer,
+	heartbeat,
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/app/views/mobile/iPhoneRedirect.scala.html
```diff
@@ -34,10 +59,19 @@
...
+		heartbeat = setInterval(intervalHeartbeat, 200);
+		document.location = 'kifi:@url';
+		// try to prevent a "flicker" of the body immediately before the redirect
+		setTimeout(function () { document.body.style.display = ''; }, 200);
+		timer = setTimeout(function () { document.location = 'https://itunes.apple.com/app/id740232575'; }, 2000);
+	}
+	return false;
+}
+
+openKifiApp();
```
2is10, line 74: I don’t see the definition of `openKifiApp`.

kifi-backend/modules/shoebox/app/views/mobile/iPhoneRedirect.scala.html
```diff
@@ -34,10 +59,19 @@
     if (diff > 1000) clearTimers();
 }
 
-lastInterval = getTime();
-heartbeat = setInterval(intervalHeartbeat, 200);
-document.location = 'kifi://@url';
-timer = setTimeout(function () { document.location = 'https://itunes.apple.com/app/id740232575'; }, 2000);
+function open() {
```
2is10, line 62: `open` could be a confusing name, since [open](https://developer.mozilla.org/en-US/docs/Web/API/Window.open) is already defined in the global scope.

kifi-backend/modules/shoebox/app/views/mobile/iPhoneRedirect.scala.html
```diff
@@ -3,23 +3,48 @@
...
+
+<div><a id="app-btn" href="https://www.kifi.com" onclick="return open();">Loading Kifi App...</a></div>
+
 <!-- source: http://aawaara.com/post/74543339755/smallest-piece-of-code-thats-going-to-change-the-world -->
 <script type="text/javascript">
-var timer, heartbeat, lastInterval;
+var timer,
+	heartbeat,
+	lastInterval,
+	isRunning = false,
```
2is10, line 40: `isRunning` appears to never be set to anything but `false`.

kifi-backend/modules/shoebox/app/views/mobile/iPhoneRedirect.scala.html
```diff
@@ -34,10 +59,19 @@
...
-heartbeat = setInterval(intervalHeartbeat, 200);
-document.location = 'kifi://@url';
-timer = setTimeout(function () { document.location = 'https://itunes.apple.com/app/id740232575'; }, 2000);
+function open() {
+	if (!isRunning) {
+		lastInterval = getTime();
+		heartbeat = setInterval(intervalHeartbeat, 200);
+		document.location = 'kifi:@url';
+		// try to prevent a "flicker" of the body immediately before the redirect
+		setTimeout(function () { document.body.style.display = ''; }, 200);
```
2is10, line 68: Seems like `clearTimers` would be a more natural time to show the button. Might consider renaming it `abortNavigation` or something else a little higher level.

2is10:
Cool, thanks for adding the button.

joshm1:
@2is10 ouch, thanks for the thorough review! I made a last minute update and looks like I forgot to upload that version to s3 to test out before pushing.  fixing it now

kifi-backend/modules/shoebox/app/views/mobile/iPhoneRedirect.scala.html
joshm1, line 68: I chose to show the button earlier b/c if the app doesn't load w/in .2 seconds, 1.8 seconds is enough time for the user to notice the blank page and wonder what is going on.

PR #10344, opened 2014-08-18
Use isLikelyHuman detection in email invite recommendations

leogrim:
@raymondng @andrewconner


raymondng:
lgtm. add @2is10

2is10:
:+1:

PR #10347, opened 2014-08-18
updated endpoints for libraryInfos

ahsu1230:
(general details for libraries and invited libraries for a user), (full details for specific libraries - including a keepHolder)

@andrewconner


andrewconner:
Looks good. :+1:

PR #10354, opened 2014-08-18
Fix IE10 problem where IE10 does not recognize element.dataset.

yipingliao:
@2is10

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/home/home.js
+++ b/kifi-backend/modules/shoebox/angular/src/home/home.js
@@ -96,7 +96,7 @@ angular.module('kifi')
     };
 
     $scope.importBookmarks = function () {
-      var kifiVersion = angular.element($window.document.documentElement).data('kifiExt');
+      var kifiVersion = $window.document.documentElement.getAttribute('data-kifi-ext');
 
       if (!kifiVersion) {
         $rootScope.$emit('showGlobalModal','installExtension');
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/angular/src/layout/rightCol/rightCol.tpl.html
```diff
@@ -1,4 +1,4 @@
-<footer class="kf-col-inner" ng-controller="RightColCtrl">
+<div class="kf-col-inner" ng-controller="RightColCtrl">
```
yipingliao, line 1: Switched from footer to div for the right column since the right column is not a footer. This div's parent element is an aside.

2is10, line 1: nice improvement

kifi-backend/modules/shoebox/angular/src/home/home.js
```diff
@@ -96,7 +96,7 @@ angular.module('kifi')
     };
 
     $scope.importBookmarks = function () {
-      var kifiVersion = $window.document.documentElement.dataset.kifiExt;
+      var kifiVersion = angular.element($window.document.documentElement).data('kifiExt');
```
2is10: I’m not sure whether angular.element initializes an element’s data cache from the `data-` attributes. If it does, this should work fine. If not, we could just use `$` instead.

2is10:
:+1:

yipingliao:
Updated to use getAttribute. Will merge and deploy tomorrow morning. Hopefully it'll work as expected! Thanks for your help. :)

PR #10355, opened 2014-08-18
changed join/decline library endpoints

ahsu1230:
- joining / declining endpoints take in libraryId instead invite Id
- changed LibraryInfo & FullLibraryInfo to have url (LibraryInfo doesn't have slug anymore)

@andrewconner


kifi-backend/modules/common/app/com/keepit/model/Library.scala
```diff
@@ -55,6 +56,11 @@ object Library extends ModelWithPublicIdCompanion[Library] {
   def isValidName(name: String): Boolean = {
     (name != "") && !(name.length > maxNameLength) && !(name.contains("\"")) && !(name.contains("/"))
   }
+
+  def formatLibraryUrl(owner: BasicUser, slug: LibrarySlug): String = {
+    val usernameString = if (owner.username.isEmpty) owner.externalId else owner.username.get.value
+    s"/$usernameString/${slug.value}"
+  }
```
ahsu1230, line 63: look okay? =)

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
```diff
@@ -83,13 +83,20 @@ class LibraryController @Inject() (
 
   def getLibrariesByUser = JsonAction.authenticated { request =>
     val (accesses, invites) = libraryCommander.getLibrariesByUser(request.userId)
+    // rule out invites that are not duplicate invites to same library (only show library invite with highest access)
+    val invitesToShow = invites.groupBy(x => x._2).map { lib =>
+      val invites = lib._2.unzip._1
+      val highestInvite = libraryCommander.sortInvites(invites).last
+      (highestInvite, lib._1)
+    }.toSeq
```
ahsu1230, line 91: just for clarification, a user can receive multiple invites to the same library (and also multiple access levels)
1) we don't want all those invitations (so only show one)
2) we want the invite with the highest access level

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -178,6 +180,15 @@ class LibraryCommander @Inject() (
     }
   }
 
+  def sortInvites(invites: Seq[LibraryInvite]): Seq[LibraryInvite] = {
```
andrewconner: This can be replaced with `invites.sorted` if you put the ordering in the companion object and make it implicit. See `Id.scala`

kifi-backend/modules/common/app/com/keepit/model/LibraryMembership.scala
```diff
@@ -71,4 +71,8 @@ object LibraryAccess {
       case OWNER.value => OWNER
     }
   }
+
+  implicit def ord: Ordering[LibraryAccess] = new Ordering[LibraryAccess] {
```
andrewconner: Necessary? Or: use this instead of the ordering in `LibraryInvite`

kifi-backend/modules/common/app/com/keepit/model/Library.scala
andrewconner, line 63: `owner.externalId.id`

PR #10366, opened 2014-08-19
Minor syntactical tweaks

andrewconner:
Some very minor syntactical tweaks, @ahsu1230. IntelliJ was offering suggestions for improvements, and I made `formatLibraryUrl` not need a BasicUser (since it isn't always available).


ahsu1230:
Ok! I think I got the gist lol - will keep that in mind in the future.

andrewconner:
IntelliJ puts yellow lines in code areas it recommends for improvement. The big one here was unnecessary `{}` in multi-line match statements. I think we follow not using them in most (all?) of the code so just changed it to be consistent.

ahsu1230:
I see. I'll clean up as we add more BE stuff

ahsu1230:
I guess how much of yellow lined recommendations from IntelliJ do you follow then?

PR #10369, opened 2014-08-19
not allowing non http(s) protocol urls to be processed

eishay:
@andrewconner


kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepInterner.scala
```diff
@@ -156,8 +156,10 @@ class KeepInterner @Inject() (
   }
 
   val MAX_RANDOM_SCHEDULE_DELAY: Int = 600000
+  private val httpPrefix = "https?://.*".r
```
2is10: `.findPrefixOf` will return earlier if you remove `.*` from the regex

eishay: Right, thanks

andrewconner:
Looks good. Thanks.

PR #10380, opened 2014-08-19
Pymk updates

yipingliao:
Update PYMK in the following 2 ways:
(1) Switch endpoint to a new endpoint that uses offsets instead of pages (this increases flexibility in requesting people you may know)
(2) If the PYMK endpoint returns no results (sometimes the server takes a bit longer to put the data in the cache? @leogrim Correct me if I am not saying this right), then retry up to 3 times.

@2is10 cc @leogrim

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/friends/friendService.js
+++ b/kifi-backend/modules/shoebox/angular/src/friends/friendService.js
@@ -139,24 +139,22 @@ angular.module('kifi')
       },
 
       getPeopleYouMayKnow: function (offset, limit) {
-        var pymkDeferred = $q.defer();
+        var deferred = $q.defer();
 
         // If the people-you-may-know endpoint does not return a good result,
         // retry up to 2 times.
-        var pymkWaitTimes = [0, 5 * 1000, 10 * 1000];
-        function getPymkWithRetries(numRetries) {
-          numRetries = numRetries || 0;
+        var waitTimes = [5 * 1000, 10 * 1000];
+        function getPymkWithRetries() {
           kifiPeopleYouMayKnowService.get(offset || 0, limit || 10).then(function (people) {
             if (people.length > 0) {
-              pymkDeferred.resolve(people);
+              deferred.resolve(people);
             } else {
-              if (numRetries === 2) {
-                pymkDeferred.resolve([]);
+              if (waitTimes.length === 0) {
+                deferred.resolve([]);
               } else {
-                numRetries++;
                 $timeout(function () {
-                  getPymkWithRetries(numRetries);
-                }, pymkWaitTimes[numRetries]);
+                  getPymkWithRetries();
+                }, waitTimes.shift());
               }
             }
           });
@@ -164,7 +162,7 @@ angular.module('kifi')
 
         getPymkWithRetries();
 
-        return pymkDeferred.promise;
+        return deferred.promise;
       },
 
       hidePeopleYouMayKnow: function (id) {
--- a/kifi-backend/modules/shoebox/angular/src/friends/friendService.js
+++ b/kifi-backend/modules/shoebox/angular/src/friends/friendService.js
@@ -146,7 +146,7 @@ angular.module('kifi')
         var pymkWaitTimes = [0, 1000, 3 * 1000, 10 * 1000];
         function getPymkWithRetries(numRetries) {
           numRetries = numRetries || 0;
-          kifiPeopleYouMayKnowService.get(offset || 0, limit || 0).then(function (people) {
+          kifiPeopleYouMayKnowService.get(offset || 0, limit || 10).then(function (people) {
             if (people.length > 0) {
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/shoebox/angular/src/friends/friendService.js
```diff
@@ -139,8 +138,33 @@ angular.module('kifi')
...
       },
 
-      getPeopleYouMayKnow: function (page) {
-        return kifiPeopleYouMayKnowService.get(page || 0);
+      getPeopleYouMayKnow: function (offset, limit) {
+        var pymkDeferred = $q.defer();
+
+        // If the people-you-may-know endpoint does not return a good result,
+        // retry up to 3 times.
+        var pymkWaitTimes = [0, 1000, 3 * 1000, 10 * 1000];
```
yipingliao: @leogrim The retry times here are: 1 sec, 3 sec, and 10 secs. Please let me know if I should change these.

leogrim: That sounds a bit quick, given that the graph returns a result in ~2 seconds. I'd say 10, 30 and 1 minute.

yipingliao: I just talked to Jen. 30 seconds and 1 minute is too long to wait to drop in the people. We're going with retries in 5 seconds and then 10 seconds, and then if there are no results, display Who To Invite (and will change the heading accordingly to reflect that; won't be in this pull).

leogrim: WTI relies on the same data and won't be available if PYMK is not. Retrying too quickly will just add unnecessary load on the graph (we have request consolidators in place but they can't catch everything in a distributed environment). Ok for 5 seconds and 10 seconds.

yipingliao: Ah, I see.
However, there still be a case where we get no PYMK because there is just no Kifi user to recommend, but there are other social network contacts to invite. In that case, we will still try the WTI endpoint and show WTI to encourage the user to invite his (non-Kifi) contacts.

leogrim: Possible but unlikely, there should be several hundreds PYMK recommendations available :) 
So I don't know if you need to spend time doing this vs simply hiding the module in that rare scenario.

kifi-backend/modules/shoebox/angular/src/friends/friendService.js
```diff
@@ -139,8 +138,33 @@ angular.module('kifi')
...
 
-      getPeopleYouMayKnow: function (page) {
-        return kifiPeopleYouMayKnowService.get(page || 0);
+      getPeopleYouMayKnow: function (offset, limit) {
+        var pymkDeferred = $q.defer();
+
+        // If the people-you-may-know endpoint does not return a good result,
+        // retry up to 2 times.
+        var pymkWaitTimes = [0, 5 * 1000, 10 * 1000];
+        function getPymkWithRetries(numRetries) {
```
2is10: When I read just this function declaration, `numRetries` sounds like it’s the number or retries desired. The name `attempt` or `attemptNumber` might be clearer.

Alternatively, you might consider shifting off the first value from `pymkWaitTimes` on each call, and terminate if it’s empty. Then we wouldn’t need an argument and could eliminate the use of the value `2` below.

kifi-backend/modules/shoebox/angular/src/friends/friendService.js
```diff
@@ -139,8 +138,33 @@ angular.module('kifi')
...
       },
 
-      getPeopleYouMayKnow: function (page) {
-        return kifiPeopleYouMayKnowService.get(page || 0);
+      getPeopleYouMayKnow: function (offset, limit) {
+        var pymkDeferred = $q.defer();
+
+        // If the people-you-may-know endpoint does not return a good result,
+        // retry up to 2 times.
+        var pymkWaitTimes = [0, 5 * 1000, 10 * 1000];
```
2is10: Because this is a pretty small function scope and there’s no ambiguity, you might consider removing the `pymk` prefix from `pymkDeferred` and `pymkWaitTimes`.

2is10:
Nice work. LGTM. :+1:

kifi-backend/modules/shoebox/angular/src/friends/friendService.js
yipingliao: Ah yes, you're right. If we use `numTries` one might expect the value to decrease from 2 to 1 to 0.
I'll shift the array.

kifi-backend/modules/shoebox/angular/src/friends/friendService.js
yipingliao: Will do.

PR #10381, opened 2014-08-19
First Library ng code

andrewconner:
Starting `libraryService`, created `library` experiment.

`libraryByPathService` and `libraryByIdService` are implemented incorrectly for now until the API is out.

@ahsu1230


kifi-backend/modules/shoebox/angular/src/libraries/libraryService.js
```diff
@@ -0,0 +1,60 @@
...
+        return libraryByIdService.get(libraryId);
+      },
+      getLibraryByPath: function (path) {
+        return libraryByPathService.get(path);
+      }
+    };
+
+    return api;
+  }
+]);
```
ahsu1230, line 60: nicE!

PR #10386, opened 2014-08-19
Make name field optional in InviteRecommendation

leogrim:
@yipingliao @2is10

### DIFF
--- a/kifi-backend/modules/abook/app/com/keepit/abook/commanders/ABookRecommendationCommander.scala
+++ b/kifi-backend/modules/abook/app/com/keepit/abook/commanders/ABookRecommendationCommander.scala
@@ -219,7 +219,7 @@ class ABookRecommendationCommander @Inject() (
       case (socialUserId, score) if isRelevant(socialUserId) =>
         val friend = relevantSocialFriends(socialUserId)
         val lastInvitedAt = existingSocialInvites.get(socialUserId).flatMap(_.lastSentAt)
-        InviteRecommendation(friend.networkType, Right(friend.socialId), friend.fullName, friend.getPictureUrl(80, 80), lastInvitedAt, score)
+        InviteRecommendation(friend.networkType, Right(friend.socialId), Some(friend.fullName), friend.getPictureUrl(80, 80), lastInvitedAt, score)
     }
     recommendations.take(relevantSocialFriends.size)
   }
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/abook/app/com/keepit/abook/commanders/ABookRecommendationCommander.scala
```diff
@@ -258,15 +258,8 @@ class ABookRecommendationCommander @Inject() (
...
-        val contacts = relevantEmailAccounts(emailAccountId)
-        val (emailAddress, name) = contacts.find(_.name.isDefined) match {
-          case Some(preferredContact) => (preferredContact.email, preferredContact.name.get)
-          case _ => {
-            val emailAddress = contacts.head.email
-            (emailAddress, emailAddress.address)
-          }
-        }
-        InviteRecommendation(SocialNetworks.EMAIL, Left(emailAddress), name, None, lastInvitedAt, score)
+        val preferredContact = relevantEmailAccounts(emailAccountId).maxBy(_.name.map(_.length) getOrElse 0)
```
2is10, line 261: I’m not sure we want to prefer contacts with the longest names. Seems like that could introduce some bias, perhaps even toward non-human contacts. Maybe I’m misreading.

2is10:
So all clients are already prepared to handle no name?

2is10:
LGTM – just the two minor questions

kifi-backend/modules/abook/app/com/keepit/abook/commanders/ABookRecommendationCommander.scala
leogrim, line 261: This is in the situation where a user has several contacts coming from different address books for the same email address (unique emailAccountId). We pick the entry with the longest name to favor a full name over just a first or last name. This is just for display, it has no impact on ranking.

2is10, line 261: Sounds good. Thanks for explaining.

leogrim:
The site is the only client getting these at the moment, and it supports it. After chatting with @davoda and looking at the backend, my understanding is that mobile doesn't use WTI at all but relies on some code in typeahead to get connections in some deterministic order (not by relevance).

PR #10390, opened 2014-08-19
userHasVisibility & createLibraryFromTag

ahsu1230:
commander functions & tests
for (1) check to see if user has visibility to a library
for (2) create a library from a tag (follows the same structure as copy/move keeps to a library)
if a library of same tag name exists, any URIs that already exist in the library will have an error
these new keeps in the destination library will also retain the tag

@andrewconner 

controller API still in the works...

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
@@ -161,52 +161,37 @@ class LibraryCommander @Inject() (
     }
   }
 
-  def createLibraryFromCollection(userId: Id[User], tag: Collection): (Library, Seq[(Keep, LibraryError)]) = db.readWrite { implicit s =>
-    def saveKeep(lib: Library, keep: Keep) {
-      val newKeep = keepRepo.save(Keep(
-        userId = userId,
-        libraryId = lib.id,
-        title = keep.title,
-        uriId = keep.uriId,
-        urlId = keep.urlId,
-        url = keep.url,
-        isPrivate = keep.isPrivate,
-        bookmarkPath = keep.bookmarkPath,
-        source = KeepSource.site))
-      keepToCollectionRepo.save(KeepToCollection(keepId = newKeep.id.get, collectionId = tag.id.get))
-    }
-    val allKeeps2Tag = keepToCollectionRepo.getByCollection(tag.id.get)
-    libraryRepo.getByNameAndUserId(userId, tag.name) match {
-      case None =>
-        val lib = libraryRepo.save(Library(name = tag.name,
-          ownerId = userId,
-          visibility = LibraryVisibility.SECRET,
-          slug = LibrarySlug(tag.name),
-          kind = LibraryKind.USER_CREATED,
-          memberCount = 1))
-        libraryMembershipRepo.save(LibraryMembership(libraryId = lib.id.get, userId = userId, access = LibraryAccess.OWNER, showInSearch = true))
-        allKeeps2Tag.map { k2c =>
-          val keep = keepRepo.get(k2c.keepId)
-          saveKeep(lib, keep)
-        }
-        (lib, Seq.empty)
-      case Some(lib) =>
-        val existingURIs = keepRepo.getByLibrary(lib.id.get).map(_.uriId).toSet
-        val badKeeps = collection.mutable.Set[(Keep, LibraryError)]()
-        allKeeps2Tag.map { k2c =>
-          val keep = keepRepo.get(k2c.keepId)
-          if (!existingURIs.contains(keep.uriId)) {
-            saveKeep(lib, keep)
-          } else {
-            badKeeps += keep -> LibraryError.AlreadyExistsInDest
+  def copyKeepsFromCollectionToLibrary(libraryId: Id[Library], tagName: String): Either[LibraryFail, Seq[(Keep, LibraryError)]] = {
+    db.readWrite { implicit s =>
+      val lib = libraryRepo.get(libraryId)
+      val ownerId = lib.ownerId
+      collectionRepo.getByUserAndName(ownerId, tagName) match {
+        case None =>
+          Left(LibraryFail("tag not found"))
+        case Some(tag) =>
+          val badKeeps = collection.mutable.Set[(Keep, LibraryError)]()
+          val existingURIs = keepRepo.getByLibrary(libraryId).map(_.uriId).toSet
+          val keeps = keepToCollectionRepo.getByCollection(tag.id.get).map(k2c => keepRepo.get(k2c.keepId))
+          keeps.map { k =>
+            if (!existingURIs.contains(k.uriId)) {
+              val newKeep = keepRepo.save(Keep(
+                userId = ownerId,
+                libraryId = Some(libraryId),
+                title = k.title,
+                uriId = k.uriId,
+                urlId = k.urlId,
+                url = k.url,
+                isPrivate = k.isPrivate,
+                bookmarkPath = k.bookmarkPath,
+                source = KeepSource.site))
+              keepToCollectionRepo.save(KeepToCollection(keepId = newKeep.id.get, collectionId = tag.id.get))
+            } else {
+              badKeeps += k -> LibraryError.AlreadyExistsInDest
+            }
           }
-        }
-        (lib, badKeeps.toSeq)
+          Right(badKeeps.toSeq)
+      }
     }
-  }
-
-  def getLibraryById(id: Id[Library]): Library = db.readOnlyMaster { implicit s =>
-    libraryRepo.get(id)
   }
 
   def getLibraryByUserAndSlug(ownerId: Id[User], slug: LibrarySlug): Option[Library] = db.readOnlyMaster { implicit s =>
@@ -221,15 +206,18 @@ class LibraryCommander @Inject() (
     }
   }
 
-  def userHasVisibility(userId: Id[User], libraryId: Id[Library]): Boolean = {
+  def userAccess(userId: Id[User], libraryId: Id[Library]): Option[LibraryAccess] = {
     db.readOnlyMaster { implicit s =>
-      libraryRepo.get(libraryId).visibility match {
-        case LibraryVisibility.PUBLISHED =>
-          true
-        case LibraryVisibility.DISCOVERABLE =>
-          libraryMembershipRepo.getWithLibraryIdAndUserId(libraryId, userId).nonEmpty
-        case LibraryVisibility.SECRET =>
-          false
+      libraryMembershipRepo.getWithLibraryIdAndUserId(libraryId, userId) match {
+        case Some(mem) =>
+          Some(mem.access)
+        case None =>
+          if (libraryRepo.get(libraryId).visibility == LibraryVisibility.PUBLISHED)
+            Some(LibraryAccess.READ_ONLY)
+          else if (libraryInviteRepo.getWithLibraryIdAndUserId(libraryId, userId).nonEmpty)
+            Some(LibraryAccess.READ_ONLY)
+          else
+            None
       }
     }
   }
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -180,6 +221,19 @@ class LibraryCommander @Inject() (
...
 
+  def userHasVisibility(userId: Id[User], libraryId: Id[Library]): Boolean = {
+    db.readOnlyMaster { implicit s =>
+      libraryRepo.get(libraryId).visibility match {
+        case LibraryVisibility.PUBLISHED =>
+          true
+        case LibraryVisibility.DISCOVERABLE =>
+          libraryMembershipRepo.getWithLibraryIdAndUserId(libraryId, userId).nonEmpty
+        case LibraryVisibility.SECRET =>
+          false
```
andrewconner: What about owners of libraries? Or people that are followers of secret libraries?

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -160,16 +161,56 @@ class LibraryCommander @Inject() (
...
+        }
+        (lib, badKeeps.toSeq)
     }
   }
 
-  def getLibraryByUserAndSlug(ownerId: Id[User], slug: LibrarySlug): Option[Library] = {
-    db.readOnlyMaster { implicit s =>
-      libraryRepo.getBySlugAndUserId(userId = ownerId, slug = slug)
-    }
+  def getLibraryById(id: Id[Library]): Library = db.readOnlyMaster { implicit s =>
```
andrewconner: It may be better not to have this. It's not really doing anything besides opening a db connection and calling a method. Anywhere we have that needs the method call can just do it itself, and potentially re-use a db connection (instead of forcing a new one).

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -160,16 +161,56 @@ class LibraryCommander @Inject() (
     }
   }
 
-  def getLibraryById(id: Id[Library]): Library = {
-    db.readOnlyMaster { implicit s =>
-      libraryRepo.get(id)
+  def createLibraryFromCollection(userId: Id[User], tag: Collection): (Library, Seq[(Keep, LibraryError)]) = db.readWrite { implicit s =>
```
andrewconner: What if this is built on top of creating a library? So the user will pass in the name, slug, discoverability, etc, and that way they can do the transition however they like.

Also, we're on the way to dropping the `collection` table (I should have mentioned this), since tags will become just freeform text. This method may be better to just take a tag name, and call `collectionRepo.getByUserAndName` so we can migrate easily when it's time.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
ahsu1230: Ok, for now it's just only being used by LibraryController.getLibraryById

kifi-backend/modules/shoebox/test/com/keepit/commanders/LibraryCommanderTest.scala
```diff
@@ -345,6 +345,19 @@ class LibraryCommanderTest extends Specification with ShoeboxTestInjector {
       }
     }
 
+    "does user have visibility" in {
```
andrewconner, line 333: Mind adding a test checking for secret libraries? Users should have access to their own (and following) secret libraries

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
ahsu1230: oh right about owners. But people can follow a secret library?

andrewconner:
Nice, you're moving quickly. I didn't realize you were ready for these two. Tag -> Library is basically an import mechanism so that current users can take existing tags and create new libraries (and if we have time: add to existing libraries). It'll re-use the "create library" dialog on the front-end, but will make a different API call optionally. An idea is to have the client make two API calls: 1) creates the library as usual, no change, 2) tells the backend to copy all the keeps with a tag "tagname" to the library.

This way, adding to an existing library is easy. What do you think?

andrewconner:
For the 2nd one, user visibility, no need for a controller method. Just commander. It's for internal ACLs only. Clients won't make this request directly; rather, they'll request for what they need, and get an error if permission denied.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
andrewconner: Yep, if invited.

"Published" = library is visible to all, and is discoverable to all (see the Google doc for explicit meaning of discoverable)
"Discoverable" = visible to only members, but discoverable to friends
"Secret" = visible to only members, not discoverable to anyone but members

ahsu1230:
I actually like the calling the two APIs better. It would be easier to handle all the edge cases (like if a slug is already used, etc.) if the API is separated into the two instead of all one big chunk.

kifi-backend/modules/shoebox/test/com/keepit/commanders/LibraryCommanderTest.scala
ahsu1230, line 333: yup will do

andrewconner:
Cool let's do that then. So in the tag -> library API call, we take in a library ID and tag name.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -160,16 +161,41 @@ class LibraryCommander @Inject() (
...
+              val newKeep = keepRepo.save(Keep(
+                userId = ownerId,
+                libraryId = Some(libraryId),
+                title = k.title,
+                uriId = k.uriId,
+                urlId = k.urlId,
+                url = k.url,
+                isPrivate = k.isPrivate,
+                bookmarkPath = k.bookmarkPath,
+                source = KeepSource.site))
```
ahsu1230: would source be different?

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -160,16 +161,41 @@ class LibraryCommander @Inject() (
...
+              keepToCollectionRepo.save(KeepToCollection(keepId = newKeep.id.get, collectionId = tag.id.get))
```
ahsu1230: make sure new keeps copied to library retain tag?

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
andrewconner: Oo, good question. Simple solution is make a new `KeepSource` for tag import.

This brings up another idea: let users pick if they want the keeps affected to be copied, or moved. If moved, just trigger moving keeps between libraries. If copied, do exactly what you're doing. What do you think?

Also, any way to re-use move/copy commander methods, or would that be a pain?

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
andrewconner: For now, I think that's a good policy. Good thinking.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
ahsu1230: I might be able to reuse that code. the one in copy/move right now checks a lot of permissions between the source/destination libraries

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -160,16 +161,29 @@ class LibraryCommander @Inject() (
...
+          Left(LibraryFail("tag not found"))
+        case Some(tag) =>
+          val keeps = keepToCollectionRepo.getByCollection(tag.id.get).map(k2c => keepRepo.get(k2c.keepId))
+          def saveKeep(k: Keep, s: RWSession): Unit = {
+            implicit val session = s
+            val newKeep = keepRepo.save(Keep(title = k.title, uriId = k.uriId, url = k.url, urlId = k.urlId, isPrivate = k.isPrivate,
+              userId = k.userId, source = KeepSource.tagImport, libraryId = Some(libraryId)))
+            keepToCollectionRepo.save(KeepToCollection(keepId = newKeep.id.get, collectionId = tag.id.get))
+          }
+          val badKeeps = applyToKeeps(ownerId, lib, keeps, Set(), saveKeep)
```
ahsu1230, line 179: code reused!

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -349,7 +379,7 @@ class LibraryCommander @Inject() (
       case Some(_) =>
         def saveKeep(k: Keep, s: RWSession): Unit = {
           implicit val session = s
-          val newKeep = keepRepo.save(Keep(title = k.title, uriId = k.uriId, url = k.url, urlId = k.urlId,
+          val newKeep = keepRepo.save(Keep(title = k.title, uriId = k.uriId, url = k.url, urlId = k.urlId, isPrivate = k.isPrivate,
```
ahsu1230, line 382: make sure to keep privacy settings!

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
andrewconner, line 179: Could we `copyKeeps()` instead? As-is, `libraryId` isn't checked for permissions.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -160,16 +161,29 @@ class LibraryCommander @Inject() (
...
+          Right(badKeeps.toSeq)
+      }
     }
   }
 
-  def getLibraryByUserAndSlug(ownerId: Id[User], slug: LibrarySlug): Option[Library] = {
-    db.readOnlyMaster { implicit s =>
-      libraryRepo.getBySlugAndUserId(userId = ownerId, slug = slug)
-    }
+  def getLibraryByUserAndSlug(ownerId: Id[User], slug: LibrarySlug): Option[Library] = db.readOnlyMaster { implicit s =>
```
andrewconner, line 185: Same for this one, it's just calling a method inside a db session.

It makes it so that if the caller already had a db session open, we're opening two sessions.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
andrewconner, line 382: Good catch.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
ahsu1230, line 179: I think in copyKeeps() all tags are copied per keep
Here, we only want to retain the one tag for all new keeps

so if I have one keep with tags "USA" & "Murica"
then I create a library called "USA"
I should have one keep in library "USA" with only one tag of "USA" ? copyKeeps() will also tag new keep with "Murica"

ahsu1230, line 179: as for checking permissions, I should add that.

PR #10392, opened 2014-08-19
Fix build (increase Protractor page sync timeout).

yipingliao:
@2is10


2is10:
Okay. Long term, let’s look for ways to keep these tests fast.

PR #10395, opened 2014-08-19
feed digest includes users-kept information

joshm1:
also refactored the SenderImpl a bit.  Use of for-comprehensions and extracted methods (abstractions) easier to follow over the previous a deeply-nested maps and flatMaps.

@stkem


kifi-backend/modules/curator/app/com/keepit/curator/commanders/email/EngagementFeedEmailSender.scala
```diff
@@ -11,95 +11,179 @@ import com.keepit.curator.commanders.RecommendationGenerationCommander
 import com.keepit.curator.model.{ UriRecommendationRepo, UriRecommendation }
```
stkem: You may want to change the filename to match the new classname.

joshm1: I will in another commit. I intentionally didn't do it here so Git wouldn't think I deleted the old file and just created a new one (therefore losing change history for the file).  There were too many changes for git to think this was a rename.

stkem:
lgtm

PR #10399, opened 2014-08-20
removing old filter options

ymatsuda:
Removing old filter options (the time range filter, the custom user filter, and the collection filter)
I don't think these are used any more. And there is no plan to use them again in the future.
If these parameters are specified in the search URLs, the search backend will ignore them silently.

@eishay @2is10 @andrewconner @jaredpetker


2is10:
sounds good to me

andrewconner:
Sounds good. Does mobile have any references to it? @thassss

PR #10400, opened 2014-08-20
updateUsername & changes to createLibraryFromCollection

ahsu1230:
removed getLibraryByUserAndSlug from LibraryCommander

@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/UserController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/UserController.scala
@@ -239,6 +239,22 @@ class UserController @Inject() (
     }
   }
 
+  def updateUsername() = JsonAction.authenticatedParseJson { implicit request =>
+    val newUsername = (request.body \ "username").as[Username]
+    userCommander.setUsername(request.userId, newUsername) match {
+      case Left(error) => BadRequest(Json.obj("error" -> error))
+      case Right(username) => Ok(Json.obj("username" -> username))
+    }
+  }
+
+  def updateName() = JsonAction.authenticatedParseJson { implicit request =>
+    val newFirstName = (request.body \ "firstName").as[String]
+    val newLastName = (request.body \ "lastName").as[String]
+    val userData = UpdatableUserInfo(firstName = Some(newFirstName), lastName = Some(newLastName), description = None, emails = None)
+    userCommander.updateUserInfo(request.userId, userData)
+    Ok(JsString("success"))
+  }
+
   //private val emailRegex = """^[a-zA-Z0-9.!#$%&'*+\/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$""".r
   def updateCurrentUser() = JsonAction.authenticatedParseJsonAsync(allowPending = true) { implicit request =>
     request.body.validate[UpdatableUserInfo] match {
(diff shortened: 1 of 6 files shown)


andrewconner:
Nice! Looks good.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/UserController.scala
```diff
@@ -239,6 +239,22 @@ class UserController @Inject() (
...
 
+  def updateUsername() = JsonAction.authenticatedParseJson { implicit request =>
+    val newUsername = (request.body \ "username").as[Username]
+    userCommander.setUsername(request.userId, newUsername) match {
+      case Left(error) => BadRequest(Json.obj("error" -> error))
+      case Right(username) => Ok(Json.obj("username" -> username))
+    }
+  }
+
+  def updateName() = JsonAction.authenticatedParseJson { implicit request =>
```
andrewconner, line 250: Mind making a route for this?

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/UserController.scala
```diff
@@ -239,6 +239,22 @@ class UserController @Inject() (
...
+    val newFirstName = (request.body \ "firstName").as[String]
```
andrewconner, line 251: Mind making this and below an option?

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/UserController.scala
```diff
@@ -239,6 +239,22 @@ class UserController @Inject() (
...
+    val newLastName = (request.body \ "lastName").as[String]
+    val userData = UpdatableUserInfo(firstName = Some(newFirstName), lastName = Some(newLastName), description = None, emails = None)
+    userCommander.updateUserInfo(request.userId, userData)
```
andrewconner, line 254: We should move away from `UpdatableUserInfo`. It's vague what the caller wants to happen.

The commander should have a `updateName(userId: Id[User], firstName: Option[String], lastName: Option[String]` method.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/UserController.scala
```diff
@@ -239,6 +239,22 @@ class UserController @Inject() (
...
+    Ok(JsString("success"))
```
andrewconner, line 255: JSON actually is a subset of what's a possible js variable. Could this return an object?

kifi-backend/modules/shoebox/conf/shoebox.routes
```diff
@@ -266,6 +266,7 @@ POST    /site/recos/vote            @com.keepit.controllers.website.Recommendati
 
 GET     /site/user/me               @com.keepit.controllers.website.UserController.currentUser()
 POST    /site/user/me               @com.keepit.controllers.website.UserController.updateCurrentUser()
+POST    /site/user/username         @com.keepit.controllers.website.UserController.updateUsername()
```
andrewconner, line 269: `/site/user/me/username` may make more sense.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/UserController.scala
2is10, line 255: An object is certainly more flexible, but for the record, the JS string `"success"` is valid JSON, as are the non-string values `2` and `false`.

ahsu1230, line 255: what object do you think should be returned?
create a new User class that has first/last name, username, & email?

ahsu1230, line 255: I guess BasicUserInfo might do the trick

andrewconner, line 255: Interesting, I confused this with an empty response being invalid, apologies for the confusion. You're good.

PR #10402, opened 2014-08-20
pre-check if uri is parsable or we're getting something bad from the cli...

eishay:
...ent before passing it into the commander layer
@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtPageController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtPageController.scala
@@ -38,7 +38,7 @@ class ExtPageController @Inject() (
         Ok(Json.toJson(info))
       case Failure(e) =>
         log.error(s"Error parsing url: $url", e)
-        BadRequest(s"Error parsing url: $url")
+        BadRequest(Json.obj("error" -> s"Error parsing url: $url"))
     }
   }
 
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtPageController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtPageController.scala
@@ -22,6 +22,7 @@ import play.api.libs.json._
 
 import scala.concurrent.Await
 import scala.concurrent.duration._
+import scala.util.{ Failure, Success }
 
 class ExtPageController @Inject() (
   actionAuthenticator: ActionAuthenticator,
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtPageController.scala
```diff
@@ -31,8 +32,14 @@ class ExtPageController @Inject() (
...
     if (url.isEmpty) throw new Exception(s"empty url for json ${request.body} for user ${request.user}")
-    val info = pageCommander.getPageDetails(url, request.userId, request.experiments)
-    Ok(Json.toJson(info))
+    URI.parse(url) match {
+      case Success(_) =>
+        val info = pageCommander.getPageDetails(url, request.userId, request.experiments)
+        Ok(Json.toJson(info))
+      case Failure(e) =>
+        log.error(s"Error parsing url: $url", e)
+        BadRequest(s"Error parsing url: $url")
```
andrewconner: Mind making this JSON?

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobilePageController.scala
```diff
@@ -22,8 +25,14 @@ class MobilePageController @Inject() (
...
     val url = (request.body \ "url").as[String]
-    val info = pageCommander.getPageDetails(url, request.userId, request.experiments)
-    Ok(Json.toJson(info))
+    URI.parse(url) match {
+      case Success(_) =>
+        val info = pageCommander.getPageDetails(url, request.userId, request.experiments)
+        Ok(Json.toJson(info))
+      case Failure(e) =>
+        log.error(s"Error parsing url: $url", e)
+        BadRequest(s"Error parsing url: $url")
```
andrewconner: Same, I assume mobile would blow up on non-json

kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtPageController.scala
eishay: sure, thought i would expect the client to first check the response status before parsing and not try to parse if its not a 200.

eishay:
@andrewconner like this?

andrewconner:
Perfect thanks

2is10:
I think it’s sometimes useful to get reports like this, so that we’ll know how we can improve our parser. In this case, for example, Chrome said the fragment part contains a `#`. Chrome probably permits an unescaped `#` in the fragment because it doesn’t cause any ambiguity about the structure of the URL. The fragment begins at the first `#`. I think our parser would be better if it tolerated a `#` in a fragment, especially in light of the fact that Chrome does.

eishay:
- @ymatsuda 
  The reason why I added the intolerance for `#` was because we eventually use the apache http client which parse the url using the java URL class, and that one blows up.
  Another option is to let the URL in and know that we can't fetch content using it with the current libraries we use.

2is10:
Two observations: 1) The fragment is not sent in an HTTP request, so it can be discarded before constructing a `java.net.URL`. 2) If we do want to correct it for whatever reason, escaping `#` to `%23` would produce an equivalent, conforming URL.

eishay:
Didn't know about #1, I'll use it. We'll also do the escaping instead of just throwing exception. More pull requests will follow.

PR #10404, opened 2014-08-20
Return people you may know as an option all the way

leogrim:


leogrim:
@andrewconner @yipingliao it's a line change in the controller from this to different response codes.

PR #10407, opened 2014-08-20
Breaking out "special curator" boost score into it's own component

stkem:
instead of modifying the social score
This is currently only used to boost the items coming form the reddit bot.

@yingjieMiao @lawzlo


yingjieMiao:
lgtm

PR #10408, opened 2014-08-20
separated functionality of updateUserInfo into updating name, descriptio...

ahsu1230:
...n, & email

endpoints added
and controller test added


ahsu1230:
@andrewconner 
Note: userCommander.updateUserInfo is used both by UserController & MobileUserController
so if we wanna remove this function & endpoint, we'll have to change on site and mobile

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/UserController.scala
```diff
@@ -248,10 +248,22 @@ class UserController @Inject() (
...
+    val newDescription = (request.body \ "description").as[String]
+    userCommander.updateUserDescription(request.userId, newDescription)
+    Ok(JsString("success"))
+  }
+
+  def updateEmails() = JsonAction.authenticatedParseJson { implicit request =>
+    val newEmails = (request.body \ "emails").as[Seq[EmailInfo]]
+    val user = db.readOnlyMaster { implicit s => userRepo.get(request.userId) }
+    userCommander.updateEmailAddresses(request.userId, user.firstName, user.primaryEmail, newEmails)
     Ok(JsString("success"))
```
ahsu1230: only return JS String or want something more?

joshm1: `Ok` and 2xx usually imply success.  If the response body isn't useful to the client, how about using  205 (NoContent), which implies success as well but no body.

andrewconner: Returning the string is fine. If we have needs to differentiate between 200 and 205 later we can.

kifi-backend/modules/shoebox/app/com/keepit/commanders/UserCommander.scala
```diff
@@ -160,6 +160,15 @@ class UserCommander @Inject() (
     }
   }
 
+  def updateName(userId: Id[User], newFirstName: Option[String], newLastName: Option[String]): User = {
+    db.readWrite { implicit session =>
+      val user = userRepo.get(userId)
+      var newUser = if (newFirstName.nonEmpty) user.copy(firstName = newFirstName.get) else user
```
andrewconner: How about:

``` scala
val newUser = user.copy(firstName = newFirstName.getOrElse(user.firstName), lastName = newLastName.getOrElse(user.lastName))
```

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/UserController.scala
```diff
@@ -248,10 +248,22 @@ class UserController @Inject() (
...
+  def updateDescription() = JsonAction.authenticatedParseJson { implicit request =>
+    val newDescription = (request.body \ "description").as[String]
+    userCommander.updateUserDescription(request.userId, newDescription)
+    Ok(JsString("success"))
+  }
+
+  def updateEmails() = JsonAction.authenticatedParseJson { implicit request =>
+    val newEmails = (request.body \ "emails").as[Seq[EmailInfo]]
+    val user = db.readOnlyMaster { implicit s => userRepo.get(request.userId) }
+    userCommander.updateEmailAddresses(request.userId, user.firstName, user.primaryEmail, newEmails)
```
andrewconner: Mind making a new `userCommander` method for this? Basically, the whole old way of doing it is very ambiguous about what is needed. Specifically, this will actually remove any email not included in `newEmails`.

Check out https://app.asana.com/0/15523651812135/15641310296070 for more background. We want explicit endpoints for creating, updating, and deleting email addresses to fix the current ambiguity.

kifi-backend/modules/shoebox/conf/shoebox.routes
```diff
@@ -266,7 +266,10 @@ POST    /site/recos/vote            @com.keepit.controllers.website.Recommendati
 
 GET     /site/user/me               @com.keepit.controllers.website.UserController.currentUser()
 POST    /site/user/me               @com.keepit.controllers.website.UserController.updateCurrentUser()
-POST    /site/user/username         @com.keepit.controllers.website.UserController.updateUsername()
+POST    /site/user/me/username      @com.keepit.controllers.website.UserController.updateUsername()
+POST    /site/user/me/name          @com.keepit.controllers.website.UserController.updateName()
```
andrewconner, line 270: Thanks for splitting this one out!

andrewconner:
> Note: userCommander.updateUserInfo is used both by UserController & MobileUserController
> so if we wanna remove this function & endpoint, we'll have to change on site and mobile

Yep. We'll leave the old code in for now — you can add a comment next to it warning not to use it anymore, with today's date. I'll get with the mobile guys to work out a deprecation plan (they're on a slower release cycle).

kifi-backend/modules/shoebox/app/com/keepit/commanders/UserCommander.scala
ahsu1230: ah right duh, lol thanks

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/UserController.scala
ahsu1230: ok I just saw it. as for parameters, do you think we need to worry about whether an email is verified or not?

kifi-backend/modules/shoebox/app/com/keepit/commanders/UserCommander.scala
```diff
@@ -160,6 +160,79 @@ class UserCommander @Inject() (
...
+        ))
+        if (user.primaryEmail.isEmpty && isPrimary)
+          userValueRepo.setValue(userId, UserValueName.PENDING_PRIMARY_EMAIL, address)
+        Right()
+      } else {
+        Left("email already added")
+      }
+    }
+  }
+  def modifyEmail(userId: Id[User], address: EmailAddress, isPrimary: Boolean): Either[String, Unit] = {
```
ahsu1230: I actually noticed that isPrimary isn't really used.
Is there ever a use case where you would use modifyEmail(_,_,false?)
- because if you wanted to change primary emails anyway, you would just specify the new email (no need for boolean)

andrewconner: Probably not. You could call this `makeEmailPrimary` if that makes more sense. What do you think?

ahsu1230: yea I think so too. I'll go ahead and change it

kifi-backend/modules/shoebox/test/com/keepit/commanders/UserCommanderTest.scala
```diff
@@ -198,6 +198,50 @@ class UserCommanderTest extends Specification with ShoeboxTestInjector {
...
 
+    "handle emails" in {
+      withDb(modules: _*) { implicit injector =>
+        val userCommander = inject[UserCommander]
+        val userValueRepo = inject[UserValueRepo]
+        val user = db.readWrite { implicit session =>
+          userRepo.save(User(firstName = "Abe", lastName = "Lincoln", username = Some(Username("AbeLincoln"))))
+        }
+
+        val email1 = EmailAddress("vampireXslayer@gmail.com")
```
andrewconner, line 209: Just noticed the joke. Hahaha

ahsu1230, line 209: =D

PR #10410, opened 2014-08-20
notify curator on keep

yingjieMiao:
@stkem


stkem:
lgtm. We'll have to see if the multi keep scenario becomes a performance problem. If so, we'll have to add a batch version of the call, but should be good for now.

yingjieMiao:
agree. I guess "keep multiple" is not very common. Let's keep it as is for now.

PR #10413, opened 2014-08-20
Typeahead refactor && tests

raymondng:
Some long overdue refactoring && tests -- started long ago but got side-tracked, thanks to @andrewconner for the extra kick & patience to wait for this to be wrapped up

Aside from code clean-up/refactoring, most changes are test-related ... and caused the most merge conflicts (as FakeABookServiceClientImpl got moved to test)

@andrewconner @yingjieMiao @ymatsuda @leogrim


andrewconner:
Thanks, will review. Kind of a weird request, but mind squashing the commits since there's 35?

raymondng:
Hmm ... how do I do that? (looking)

andrewconner:
https://github.com/ginatrapani/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit

I guess we don't typically do it (some organizations do, but pretty much everyone needs to stick to it) so only do it if you want a clean resulting commit. Otherwise, I doubt it would bother anyone.

kifi-backend/modules/shoebox/app/com/keepit/commanders/TypeaheadCommander.scala
```diff
@@ -235,85 +217,63 @@ class TypeaheadCommander @Inject() (
...
-                      }
-                    }
-                  }
-                }
-              }
-            }
-          }
-          topF
+      case Some(limit) =>
+        val social: Future[Seq[(SocialNetworkType, TypeaheadHit[_])]] = socialF.map { hits =>
```
yingjieMiao: would be nice to have a type alias for `Seq[(SocialNetworkType, TypeaheadHit[_])]` . it's repeated quite a few times and makes code less readable.

raymondng: good idea.

kifi-backend/modules/shoebox/app/com/keepit/commanders/TypeaheadCommander.scala
```diff
@@ -235,85 +217,63 @@ class TypeaheadCommander @Inject() (
...
+          val (fb, lnkd) = hits.map(hit => (hit.info.networkType, hit)).partition(_._1 == SocialNetworks.FACEBOOK)
+          log.infoP(s"fb=${fb.mkString(",")} lnkd=${lnkd.mkString(",")}")
+          fb ++ lnkd
         }
+        val kifi: Future[Seq[(SocialNetworkType, TypeaheadHit[_])]] = kifiF.map { hits => hits.map(hit => (SocialNetworks.FORTYTWO, hit)) }
+        val abook: Future[Seq[(SocialNetworkType, TypeaheadHit[_])]] = abookF.map { hits => hits.filter(_.info.userId.isEmpty).map(hit => (SocialNetworks.EMAIL, hit)) }
+        val nf: Future[Seq[(SocialNetworkType, TypeaheadHit[_])]] = nfUsersF.map { hits => hits.map(hit => (SocialNetworks.FORTYTWO_NF, hit)) }
+        val futures: Seq[Future[Seq[(SocialNetworkType, TypeaheadHit[_])]]] = Seq(social, kifi, abook, nf)
```
yingjieMiao: why not use `Future.sequence`? then you probably don't need the hacky `processWhile` method.

kifi-backend/modules/shoebox/app/com/keepit/commanders/TypeaheadCommander.scala
```diff
@@ -235,85 +217,63 @@ class TypeaheadCommander @Inject() (
...
+          if (predicate(res)) processWhile(futures.tail, predicate, promised)
+          else promised.success()
+        case Failure(t) => promised.failure(t)
+      }
+    }
+    promised.future
+  }
+
+  private def fetchFirst(limit: Int, futures: Iterable[Future[(Seq[(SocialNetworkType, TypeaheadHit[_])])]]): Future[Seq[(SocialNetworkType, TypeaheadHit[_])]] = {
+    val zHits = new ArrayBuffer[(SocialNetworkType, TypeaheadHit[_])]
```
yingjieMiao: the naming of `zHits` is not clear to me.

kifi-backend/modules/shoebox/app/com/keepit/commanders/TypeaheadCommander.scala
```diff
@@ -235,85 +217,63 @@ class TypeaheadCommander @Inject() (
...
+        val kifi: Future[Seq[(SocialNetworkType, TypeaheadHit[_])]] = kifiF.map { hits => hits.map(hit => (SocialNetworks.FORTYTWO, hit)) }
+        val abook: Future[Seq[(SocialNetworkType, TypeaheadHit[_])]] = abookF.map { hits => hits.filter(_.info.userId.isEmpty).map(hit => (SocialNetworks.EMAIL, hit)) }
+        val nf: Future[Seq[(SocialNetworkType, TypeaheadHit[_])]] = nfUsersF.map { hits => hits.map(hit => (SocialNetworks.FORTYTWO_NF, hit)) }
+        val futures: Seq[Future[Seq[(SocialNetworkType, TypeaheadHit[_])]]] = Seq(social, kifi, abook, nf)
+        fetchFirst(limit, futures)
+    }
+  }
+
+  // somewhat specialized; pure side-effects
+  private def processWhile[T](futures: Iterable[Future[T]], predicate: T => Boolean, promised: Promise[Unit] = Promise()): Future[Unit] = {
```
andrewconner: Trying to follow this one, I might bug you... seems the same as Future.sequence?

kifi-backend/modules/shoebox/app/com/keepit/commanders/TypeaheadCommander.scala
raymondng: Performance. 
`Future.sequence` is good if you want to wait for all Futures to complete, but we don't want that here.

kifi-backend/modules/shoebox/app/com/keepit/commanders/TypeaheadCommander.scala
raymondng: Actually, no -- this one has short-circuitry. But sure please let me know if it's unclear (or has issues)

kifi-backend/modules/shoebox/app/com/keepit/commanders/TypeaheadCommander.scala
raymondng: z == zero. Let me add a comment ;p

kifi-backend/modules/shoebox/app/com/keepit/commanders/TypeaheadCommander.scala
raymondng: You meant `FutureHelpers`, right? Indeed there are some similarities (not the same, of course)

kifi-backend/modules/shoebox/app/com/keepit/commanders/TypeaheadCommander.scala
leogrim: Do you know which future is slow enough to impede performance?

kifi-backend/modules/shoebox/app/com/keepit/commanders/TypeaheadCommander.scala
leogrim: Yeah sorry, deleted my comment after understanding the difference.

kifi-backend/modules/shoebox/app/com/keepit/commanders/TypeaheadCommander.scala
raymondng: It's going to be highly variant depending on implementation, system/network load, etc. ... in general I'd like to avoid the pattern where we "wait for the slowest response to come back".

kifi-backend/modules/shoebox/app/com/keepit/commanders/TypeaheadCommander.scala
raymondng: No problem. Thanks for the review.

kifi-backend/modules/shoebox/app/com/keepit/commanders/TypeaheadCommander.scala
leogrim: Well it's trading performance for accuracy of results, and when I moved contact typeahead to abook I really couldn't feel a difference. I'm not really convinced we need to go hacky for performance unless proven necessary...

raymondng: AFAIK there are no trade-offs in accuracy. Note that the zero hits only collect the top matches. Let me know if I have missed anything.

For our current (small) use base, I readily admit it's difficult to argue for performance. It's an ingrained habit of mine (learned from painful experience) that I try to avoid certain "anti-performance" patterns (like this one).

Having said that, I'd argue (with no data because, well the change to abook is already done, and AFAIK we didn't collect A/B comparison test in a controlled environment) that you didn't "feel" a difference is because [1] we have small data set and [2] this logic helped -- we never wait for abook to return if we already have sufficient "good" (zero) hits. 

BTW this logic has always been there -- it's just more obvious after the refactoring.

Finally I'd object to the 'hacky' characterization :) It's certain not idiomatic functional Scala (not to mention it's not a method in standard library) but it's actually a pretty common practice for large systems with low latency requirements.

leogrim: I know it was already there, this is why I don't understand argument [2]. I can see contact results from ABook in the typeahead, which means they haven't been cut off and they come in fast enough. Now if these results were cut off, that would impact accuracy wouldn't it? Since we would return results without even knowing about potentially relevant contacts? If ABook indeed becomes too slow, these relevant contacts will disappear from results and we may not notice it, would we?

kifi-backend/modules/shoebox/app/com/keepit/commanders/TypeaheadCommander.scala
andrewconner: It makes sense now, I missed the short-circuitry.

kifi-backend/modules/shoebox/app/com/keepit/commanders/TypeaheadCommander.scala
raymondng: I see. The logic is as follows: if we have sufficient top hits from social, then we don't bother with contact results (that's lower priority, so will be discarded anyway), but if we don't have sufficient good hits, we _will_ wait for and process contact results, so there's no loss of accuracy, and the delay cost is only incurred if we need it (not if it comes within a certain threshold).

As the "argument" goes -- [1] dominates [2] likely in all scenarios. The lack of data is not helping, but we can easily (at a cost) simulate this by loading the system with millions of users/contacts, and hit one instance with thousands of concurrent typeahead requests, etc. -- there are tools to automate these sort of things ("load/stress testing"). At that point, a lot of things we don't see on a daily basis will start surfacing. I am used to the fact that we wouldn't release anything unless we pass the load/stress test, so I automatically think of these things and put them in place.

kifi-backend/modules/shoebox/app/com/keepit/commanders/TypeaheadCommander.scala
stkem: This seems fairly general purpose (i.e. not typeahead specific). Would it make sense to put it in `FutureHelpers`?

leogrim: +1 :)

2014-08-21 13:24 GMT-07:00 Stephen Kemmerling notifications@github.com:

raymondng: Alright gents -- moved to FutureHelpers (+ simple test)

raymondng:
Thanks for all the review feedback! I'd like to merge this soon (to avoid more merge conflicts) -- say by EOD today. Any objections?

yingjieMiao:
Ok with me.

leogrim:
:+1:

PR #10415, opened 2014-08-20
not setting Content-Type header if response status is 204 or 205

2is10:
@andrewconner 
cc @joshm1 

A baby step that should help some JS libraries consume our `204` responses better. Would be cool not to set it if the response body is empty, but I think that requires an extra `Iteratee` — not sure of the feasibility or performance implications.

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/common/controller/ActionsBuilder.scala
+++ b/kifi-backend/modules/common/app/com/keepit/common/controller/ActionsBuilder.scala
@@ -8,6 +8,7 @@ import play.api.libs.json.{ JsValue, JsNumber }
 import play.api.i18n.Messages
 import securesocial.core.SecuredRequest
 
+import play.api.http.Status.{ NO_CONTENT, RESET_CONTENT }
 import play.api.mvc.SimpleResult
 import com.keepit.common.logging.Logging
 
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/common/app/com/keepit/common/controller/ActionsBuilder.scala
```diff
@@ -140,4 +141,13 @@ class ActionsBuilder0(actionAuthenticator: ActionAuthenticator) extends Controll
       apply(parser = parse.tolerantJson, apiClient = apiClient, allowPending = allowPending)(authenticatedAction, unauthenticatedAction)
     }
   }
+
+  private def as(contentType: String)(result: SimpleResult): SimpleResult = {
+    result.header.status match {
+      case NO_CONTENT | RESET_CONTENT =>
```
eishay, line 147: nice

andrewconner:
Works for me. I actually would like to refactor this and come up with a better actions API. Right now it's confusing (`JsonAction` has confused everyone since I made it) and the unauthenticated actions can be handled a lot better. This is a good change though (I just wouldn't devote much more resources to it).

2is10:
After a little more thought, it seems like an `EmptyAction` might be a better convention. A `JsonAction` should return valid JSON; an `HtmlAction` should return HTML or redirect.

I guess this change might still be good for its own sake, even if we don’t intend to use it under normal circumstances. Might be good to add redirect status codes as well.

PR #10416, opened 2014-08-21
pre-compute public feed for kifi users who don't have enough keeps or friends

lawzlo:
It is a relative long pull request. I've read code a couple times by myself, but I may still miss things. Thanks for helping to review this long request. 
@stkem @yingjieMiao @joshm1 @andrewconner


kifi-backend/modules/common/conf/evolutions/shoebox/232.sql
```diff
@@ -1,3 +1,4 @@
+
```
andrewconner, line 1: What's this for?

lawzlo, line 1: Modified the conflict sql file, probably added a new line accidentally. I didn't do another commit since it doesn't affect anything. Will remove it before pushing code to prod.

kifi-backend/modules/curator/test/com/keepit/curator/RecommendationGenerationCommanderTest.scala
```diff
@@ -92,11 +128,39 @@ class RecommendationGenerationCommanderTest extends Specification with CuratorTe
...
+        val user1Keeps = makeKeeps(user1, 5, shoebox)
+        val user2Keeps = makeKeeps(user2, 5, shoebox)
+        val seedCommander = inject[SeedIngestionCommander]
+        val seedItemRepo = inject[RawSeedItemRepo]
         val commander = inject[RecommendationGenerationCommander]
+        Await.result(seedCommander.ingestAllKeeps(), Duration(10, "seconds"))
+        db.readWrite { implicit session => seedItemRepo.assignSequenceNumbers(1000) }
         commander.precomputeRecommendations()
-        1 === 1
+        Thread.sleep(5000)
```
stkem: Before anyone complains, this one is placeholder. I'm currently working on a way to await the "completeness" of a reactive lock, which will be used here (pre merge).

lawzlo: :+1:

kifi-backend/modules/common/conf/evolutions/shoebox/233.sql
```diff
@@ -0,0 +1,21 @@
+
+# CURATOR
+
+# --- !Ups
+
+CREATE TABLE if not exists public_feed (
```
stkem: Let's use the same naming pattern as the regular feed table, so something like `public_uri_recommendation`.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
```diff
@@ -30,18 +34,24 @@ class RecommendationGenerationCommander @Inject() (
...
     uriRecRepo: UriRecommendationRepo,
+    publicFeedRepo: PublicFeedRepo,
     genStateRepo: UserRecommendationGenerationStateRepo,
+    systemValueRepo: SystemValueRepo,
     experimentCommander: RemoteUserExperimentCommander) {
 
   val defaultScore = 0.0f
 
   val recommendationGenerationLock = new ReactiveLock(15)
+  val pubicFeedsGenerationLock = new ReactiveLock(10)
```
stkem: We are only generating one feed, so concurrency should be `1` here, otherwise we risk processing the same batch multiple times concurrently (won't do anything to the data, but it's a lot of wasted cycles)

kifi-backend/modules/curator/app/com/keepit/curator/model/RawSeedItemRepo.scala
```diff
@@ -108,6 +109,16 @@ class RawSeedItemRepoImpl @Inject() (
     q.as[RawSeedItem].list
   }
 
+  def getDiscoverableBySeqNum(start: SequenceNumber[RawSeedItem], maxBatchSize: Int)(implicit session: RSession): Seq[RawSeedItem] = {
+    import StaticQuery.interpolation
+    val q = if (db.dialect == H2DatabaseDialect) {
+      sql"SELECT * FROM raw_seed_item WHERE seq > ${start.value} AND (user_id IS NULL) AND discoverable=1 ORDER BY seq LIMIT $maxBatchSize;"
+    } else {
+      sql"SELECT * FROM raw_seed_item USE INDEX (raw_seed_item_u_seq_user_id) WHERE seq > ${start.value} AND (user_id IS NULL) AND discoverable=1 ORDER BY seq LIMIT $maxBatchSize;"
```
eishay, line 117: the sql statement is almost the same but for the `use index`. it may be more readable to use one statement so its clear we're not having different predicates. Can we do something like this? :

``` scala
sql"SELECT * FROM raw_seed_item ${if(db.dialect == H2DatabaseDialect) "USE INDEX (raw_seed_item_u_seq_user_id)" else ""} WHERE seq > ${start.value} AND (user_id IS NULL) AND discoverable=1 ORDER BY seq LIMIT $maxBatchSize;"
```

kifi-backend/modules/common/conf/evolutions/shoebox/233.sql
lawzlo: I was thinking since those feeds are not personalized, it's more like `Discover` but not `Recommendation` in semantics. So maybe `public_feed` is still better? Not sure. I'm ok with naming things.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
lawzlo: Make sense. Cool. Will change to 1.

kifi-backend/modules/curator/app/com/keepit/curator/model/SeedItem.scala
```diff
@@ -49,13 +65,15 @@ case class SeedItem(
 
 case class SeedItemWithMultiplier(
```
stkem, line 68: I think it's better to have a separate type here (`PublicSeedItemWithMultiplier`?), instead of optional fields. It's very boilerplaty, but that way you avoid accidentally passing an object with a `None` to a method that expects a `Some`.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/PublicUriScoringHelper.scala
```diff
@@ -0,0 +1,75 @@
...
+
+class PublicUriScoringHelper @Inject() (
+    graph: GraphServiceClient,
+    keepInfoRepo: CuratorKeepInfoRepo,
+    cortex: CortexServiceClient,
+    heimdal: HeimdalServiceClient) extends Logging {
+
+  private def getRawPopularityScores(items: Seq[SeedItemWithMultiplier]): Seq[Float] = items.map { item =>
+    val cappedPopularity = Math.min(item.timesKept, 100)
+    (cappedPopularity / 100.0).toFloat
```
eishay: same as doing 

``` scala
cappedPopularity / 100f
```

kifi-backend/modules/curator/app/com/keepit/curator/commanders/PublicUriScoringHelper.scala
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  private def getRawRecencyScores(items: Seq[SeedItemWithMultiplier]): Seq[Float] = items.map { item =>
+    val daysOld = Days.daysBetween(item.lastSeen, currentDateTime).getDays()
+    (1.0 / (Math.log(daysOld + 1.0) + 1)).toFloat
+  }
+
+  val uriHelpRankScores = TrieMap[Id[NormalizedURI], HelpRankInfo]() //This needs to go when we have proper caching on the help rank scores
```
eishay, line 35: i don't understand why `uriHelpRankScores` is a public `val` of this class. is there a reason why it should be visible outside the `getRawHelpRankScores` method and for it to be mutable?

kifi-backend/modules/curator/app/com/keepit/curator/commanders/PublicUriScoringHelper.scala
```diff
@@ -0,0 +1,75 @@
...
+  def getRawHelpRankScores(items: Seq[SeedItemWithMultiplier]): Future[(Seq[Float], Seq[Float])] = {
+    val helpRankInfos = heimdal.getHelpRankInfos(items.map(_.uriId).filterNot(uriHelpRankScores.contains))
+    helpRankInfos.map { infos =>
+      infos.foreach { info => uriHelpRankScores += (info.uriId -> info) }
```
eishay, line 39: can we have instead something like 

``` scala
val uriHelpRankScores = infos.map(info => info.uriId -> info).toMap
```

This way `uriHelpRankScores` is hidden and is immutable

kifi-backend/modules/curator/app/com/keepit/curator/model/PublicFeedRepo.scala
```diff
@@ -0,0 +1,58 @@
...
+  def deleteCache(model: PublicFeed)(implicit session: RSession): Unit = {}
+
+  def invalidateCache(model: PublicFeed)(implicit session: RSession): Unit = {}
+
+  def getByUri(uriId: Id[NormalizedURI], publicFeedState: Option[State[PublicFeed]])(implicit session: RSession): Option[PublicFeed] = {
+    (for (row <- rows if row.uriId === uriId && row.state === publicFeedState) yield row).firstOption
+  }
+
+  def getByTopMasterScore(maxBatchSize: Int, publicFeedState: Option[State[PublicFeed]])(implicit session: RSession): Seq[PublicFeed] = {
+    (for (row <- rows if row.state === publicFeedState) yield row).
```
eishay, line 52: I saw an index on `public_master_score` but not on `state`. Also, I'm not sure slick would give an efficient query here, you may want to verify.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
```diff
@@ -197,6 +232,62 @@ class RecommendationGenerationCommander @Inject() (
...
+              !publicSeedItems.isEmpty
+
+            }
+          }
+      }
+      res
+    }
+
+  }
+
```
joshm1, line 285: We're all guilty of it, and it's often rather difficult to avoid, but many great books on the topic (Clean Code, Code Complete) consider the # of lines in a method (and class/trait/module/etc) and the # of indentions to be impede code clarity/readability (among others like the # of public methods in a class, # of LOC in a class/file).  50 lines in a method is a lot, 10 indentations is a lot.  Consider how to extract logical segments into smaller, sensibly-named helper methods (abstractions!).  I'm curious what others think.  This really isn't that bad, I've seen worse cases (including my own).

@andrewconner @stkem @lawzlo @eishay @leogrim @raymondng @yingjieMiao @2is10

kifi-backend/modules/curator/app/com/keepit/curator/commanders/PublicUriScoringHelper.scala
lawzlo: Cool. Will make a change.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
yingjieMiao, line 285: Great comment @joshm1 .  I agree with you.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/UriWeightingHelper.scala
```diff
@@ -43,3 +45,22 @@ class UriWeightingHelper() {
...
 }
+
+@Singleton
+class PublicUriWeightingHelper() {
+
+  def apply(items: Seq[PublicSeedItem]): Seq[SeedItemWithMultiplier] = items.map { item =>
+    var masterWeight = 1.0f
+    UrlPatterns.scoringMultiplier.map { pattern =>
+      if (pattern.regex.findFirstIn(item.url).isDefined) masterWeight *= pattern.weight
+    }
```
joshm1, line 51: can we use fold to eliminate the mutable var and be a bit more functional?  (untested)

``` scala
val masterWeight = UrlPatterns.scoringMultiplier.fold(1.0f) { (weight, pattern) => 
  if (pattern.regex.findFirstIn(item.url).isDefined) weight * pattern.weight else weight
}
```

kifi-backend/modules/curator/app/com/keepit/curator/model/SeedItem.scala
lawzlo, line 68: Definitely, the way is  cleaner. The problem I had with this way is, when calling UriScoringHelper, the arg pass in Seq[SeedItemWithMultiplier], and this func need to call PublicUriScoringHelper first to get partial results which needs to take arg Seq[SeedItemWithMultiplier] as well. If We have `PublicSeedItemWithMultiplier` here. Not sure how to convert it to `SeedItemWithMultiplier`

kifi-backend/modules/curator/app/com/keepit/curator/commanders/UriWeightingHelper.scala
lawzlo, line 51: :+1: Nice. Let me try this way.

kifi-backend/modules/curator/app/com/keepit/curator/model/RawSeedItemRepo.scala
andrewconner, line 117: Unfortunately not possible. It's not normal string interpolation, it's making a prepared statement.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
stkem, line 285: Very good point. A general rule of thumb I try to use is for a function to fit entirely on screen in my editor sort of as a worst case. I'm guilty of violating that in lots of cases too though.
We need to be better about breaking down the recommendation generation logic, not just here but for the non-public recommendations as well. I would _not_ do it in this PR though. We'll need to think of a good abstraction to break it down. Just decomposing to make things shorter can yield lots of small functions that make no sense without all the context, which really isn't any better.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/UriWeightingHelper.scala
yingjieMiao, line 51: I think I mentioned this earlier. you can use something like `UrlPatterns.scoringMultipliers.find( rule => rule.regex.findFirstIn(item.url).isDefined)` to find the first regex matches the url. In this way, you avoid unnecessary iterations and the `var`.

stkem, line 51: We actually want to find all the regex that match the url, not just the first, which is why that wont work. I like the fold though.

lawzlo, line 51: Yes, I remember you mentioned this and I did this way. This way will find a matched rule and stop. But later we figured out we may want to all all matched rules and multiply them together because for example, if it's facebook.com, we will penalize it, but if it's eng.facebook.com or something like this. we want to boost it. Let me know if I was wrong.

kifi-backend/modules/curator/app/com/keepit/curator/model/SeedItem.scala
stkem, line 68: You could have a `asPublic` method on `SeedItemWithMultiplier` that makes a new instance of `PublicSeedItemWithMultiplier`.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
yingjieMiao, line 285: about the 'rule of thumb'. Fit on screen means differently for different people. People use different resolutions, have different size of screens, fonts, etc. In addition, sometimes people rotate screen by 90 degrees and can easily fit 100+ lines on screen.

kifi-backend/modules/common/conf/evolutions/shoebox/233.sql
stkem: You have a point. Let's keep it as is.

stkem: i.e. `public_feed`

kifi-backend/modules/common/conf/evolutions/shoebox/233.sql
```diff
@@ -0,0 +1,21 @@
...
+    id bigint(20) NOT NULL AUTO_INCREMENT,
+    created_at datetime NOT NULL,
+    updated_at datetime NOT NULL,
+    state varchar(20) NOT NULL,
+
+    uri_id bigint(20) NOT NULL,
+    public_master_score float(10) NOT NULL,
+    public_all_scores text NOT NULL,
```
stkem: The extra `public` prefix isn't really needed here, this should be clear from the table.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/PublicUriScoringHelper.scala
stkem, line 35: It's a stop gap cache for all the help rank lookups, persistent across function calls. Not very nice, but having this outside the function instead of inside is about a 100x performance difference in recommendation generation.
It'll go away as soon as @raymondng is done with "real" caching on the help rank scores.

kifi-backend/modules/curator/app/com/keepit/curator/model/SeedItem.scala
lawzlo, line 68: smart way. Will do it.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
eishay, line 285: :+1:

kifi-backend/modules/curator/app/com/keepit/curator/commanders/UriWeightingHelper.scala
stkem, line 51: No, you are right. We _do_ want to multiply them together. The code is currently correct. What @joshm1 suggested is just a little cleaner (but semantically equivalent)

yingjieMiao, line 51: I see. I wasn't aware of the context. It's a bit tricky. Ideally, we might want to find the 'best match', e.g. the regex that matches the longest substring in the url. So you wouldn't have a composition of penalty/boost. We can address that later. @stkem

kifi-backend/modules/curator/app/com/keepit/curator/model/PublicFeedRepo.scala
```diff
@@ -0,0 +1,58 @@
...
+trait PublicFeedRepo extends DbRepo[PublicFeed] with SeqNumberFunction[PublicFeed] {
+  def getByUri(uriId: Id[NormalizedURI], publicFeedState: Option[State[PublicFeed]])(implicit session: RSession): Option[PublicFeed]
+  def getByTopMasterScore(maxBatchSize: Int, publicFeedState: Option[State[PublicFeed]] = Some(PublicFeedStates.ACTIVE))(implicit session: RSession): Seq[PublicFeed]
+}
+
+@Singleton
+class PublicFeedRepoImpl @Inject() (
+  val db: DataBaseComponent,
+  val clock: Clock)
+    extends DbRepo[PublicFeed] with PublicFeedRepo with SeqNumberDbFunction[PublicFeed] with Logging {
```
stkem: This repo does not need a sequence number of it's own. We only need a sequence number to keep track of how many raw seed items we have read so far (which you are storing in the system value repo), not one for each item in the public feed, which is what this does.

kifi-backend/modules/curator/app/com/keepit/curator/model/PublicFeedRepo.scala
```diff
@@ -0,0 +1,58 @@
...
+
+  private val sequence = db.getSequence[PublicFeed]("public_feed_sequence")
```
stkem: not needed, see comment above

kifi-backend/modules/curator/app/com/keepit/curator/commanders/UriWeightingHelper.scala
```diff
@@ -26,16 +24,20 @@ class UriWeightingHelper() {
     UrlPattern("""^https?://[-A-Za-z0-9.]*.[-A-Za-z0-9.]*.[./?\#]kifi[-A-Za-z0-9.]*""".r, 1.2f, "Public page relate to Kifi"),
     UrlPattern("""^https?://(code|blog|engineering).[-A-Za-z0-9.]*.[./?\#][-A-Za-z0-9.]*""".r, 1.1f, "Tech")
   )
+}
+
+@Singleton
```
stkem: Singleton is no longer needed here since you broke out the UrlPatterns into an object (which is always a singleton)

kifi-backend/modules/curator/app/com/keepit/curator/commanders/UriWeightingHelper.scala
```diff
@@ -10,9 +10,7 @@ case class UrlPattern(
   weight: Float,
   description: String)
 
-@Singleton
-class UriWeightingHelper() {
-
+object UrlPatterns {
```
stkem, line 13: :+1:

kifi-backend/modules/curator/app/com/keepit/curator/commanders/UriWeightingHelper.scala
```diff
@@ -43,3 +45,22 @@ class UriWeightingHelper() {
       keepers = item.keepers)
   }
 }
+
+@Singleton
+class PublicUriWeightingHelper() {
```
stkem, line 46: You should be able to just have two `apply` methods on `UriWeightingHelper` with different type signature, thus avoiding the need for a second class.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/UriWeightingHelper.scala
lawzlo: Cool!

kifi-backend/modules/curator/app/com/keepit/curator/commanders/UriWeightingHelper.scala
lawzlo, line 46: The interesting thing is, it will complain "apply(scala.Seq) is already defined in the scope" I guess it can't detect different type inside of Seq.

kifi-backend/modules/curator/app/com/keepit/curator/model/PublicFeedRepo.scala
lawzlo: I should realize it when I had unit test. Will remove it.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
lawzlo, line 285: Good point. I had moved some components out as functions, but it's still too long. Will continue to refactor codes.

stkem, line 285: I suggest not speeding too much time on refactoring this in this PR, just to keep things separate (since the existing computation logic for the non-public feed should get the same refactor)

kifi-backend/modules/common/conf/evolutions/shoebox/233.sql
lawzlo: sure. removing.

kifi-backend/modules/curator/app/com/keepit/curator/model/PublicFeedRepo.scala
lawzlo, line 52: We need to check state anyway here. I guess the reason we don't add `state` as index because there are only 3 states, so it's not efficient to create index for it?  Not sure. @stkem please correct me if I'm wrong.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/PublicUriScoringHelper.scala
lawzlo, line 39: :+1:  This way looks more elegant.

lawzlo, line 39: But since we still have to keep `uriHelpRankScores` as a public val in class.  So maybe we still keep the current way.

lawzlo:
tested, pushing to prod.

kifi-backend/modules/curator/app/com/keepit/curator/model/PublicFeedRepo.scala
```diff
@@ -0,0 +1,56 @@
...
+  import DBSession._
+  import db.Driver.simple._
+
+  implicit val uriPublicScoresMapper = MappedColumnType.base[PublicUriScores, String](
+    { scores => Json.stringify(Json.toJson(scores)) },
+    { jstr => Json.parse(jstr).as[PublicUriScores] })
+
+  type RepoImpl = PublicFeedTable
+
+  class PublicFeedTable(tag: Tag) extends RepoTable[PublicFeed](db, tag, "public_feed") with SeqNumberColumn[PublicFeed] {
```
stkem, line 33: Why the `SeqNumberColumn`?

lawzlo, line 33: Sorry I didn't fully understand how SystemSeqNum and created seq num work since you told me last week so it's my mistake that I didn't remove seq column here but only removed assignSeqNum function. I think it doesn't need seq num here. Removing it.

kifi-backend/modules/curator/app/com/keepit/curator/model/PublicFeed.scala
```diff
@@ -0,0 +1,22 @@
...
+
+case class PublicFeed(
+    id: Option[Id[PublicFeed]] = None,
+    createdAt: DateTime = currentDateTime,
+    updatedAt: DateTime = currentDateTime,
+    state: State[PublicFeed] = PublicFeedStates.ACTIVE,
+    seq: SequenceNumber[PublicFeed] = SequenceNumber.ZERO,
+    uriId: Id[NormalizedURI],
+    publicMasterScore: Float,
+    publicAllScores: PublicUriScores) extends Model[PublicFeed] with ModelWithSeqNumber[PublicFeed] {
```
stkem, line 17: these still have double public names (class is `Public` and the methods) which doesn't match repo naming.

lawzlo, line 17: Yes. I left it here intentionally. I think it make sense to just name it master_score and all_scores in db since they are in separate table with uri_recommendation. But here I proper to have public prefix which I feel more consistence with class name. Also for maintenance if people search variable keyword in code base I guess?

PR #10419, opened 2014-08-21
kifeeed email WIP

joshm1:
@stkem @2is10 

I'm not happy with the duplication of passing around the unsubscribeUrl and the assetUrlBase being defined in multiple templates. I have a refactor in mind that involves creating a new EmailContext object that can be passed implicitly between templates, with many helper methods that are dynamic or will change with each template (unsubscribe URL, UTM tags, titles/headers, etc). Otherwise, this separates the header and footer and fixes some formatting issues

### DIFF
--- a/kifi-backend/modules/common/app/views/email/black/layout.scala.html
+++ b/kifi-backend/modules/common/app/views/email/black/layout.scala.html
@@ -1,0 +1,325 @@
+@(title: String = "Kifi", unsubscribeUrl: String)(content: Html)
+@assetUrlBase = @{ "https://djty7jcqog9qu.cloudfront.net/assets/digest" }
+
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/common/app/views/email/black/layout.scala.html
```diff
@@ -0,0 +1,325 @@
...
+                <td bgcolor="#ffffff">
+                    @content
+                </td>
+            </tr>
+            <tr>
+                <td height="40">&nbsp;</td>
+            </tr>
+            <tr>
+                <td valign="top">
+                    @views.html.email.black.footer(unsubscribeUrl)
```
eishay, line 317: great!

stkem:
Nice! I like the idea of an EmailContext! (not so sure about the implicit part, Play templates are confusing as is)

2is10:
EmailContext idea sounds interesting. One way might be to define a `CommonEmailTemplateValues` or `EmailTemplateFiller` type (trait or abstract class) that can be implemented differently at the point where each email template is invoked. I try to avoid the word `Context` when naming types because it’s so broad a term that many engineers, upon encountering it in unfamiliar code, interpret it as `EverythingButKitchenSink`.

PR #10420, opened 2014-08-21
escaping hash tags in a fragment instead of throwing on it

eishay:
@ymatsuda @2is10 @andrewconner


2is10:
LGTM. Thanks.

Looks like Chrome’s behavior here is consistent with Firefox and the WHATWG URL standard, specifically [the fragment state](http://url.spec.whatwg.org/#fragment-state) of parsing, which only uses [the simple encode set](http://url.spec.whatwg.org/#simple-encode-set).

You can verify it at the console of any modern browser like this:

```
new URL('http://example.com/#q=#s=r').hash
> "#q=#s=r"
```

eishay:
cool, thanks

2is10:
In fact, according to that spec, no printable ASCII characters require escaping in the fragment, so we might consider ditching the others we have listed too (`'@', ':', '/', '?'`).

```
new URL('http://example.com/?q=foo#@:/?bar').hash
> "#@:/?bar"
```

ymatsuda:
I don't think we escape '@', ':', '/', '?'. They are subtracted from the reserved character set.
Am I missing something?

2is10:
@ymatsuda You’re right, I didn’t read carefully. Might be clearer to define the sets without subtraction, but I understand if this code is intended to be a direct translation of set definitions from a spec.

PR #10421, opened 2014-08-21
return marker future when precomputing recos (for testing awaitability)

stkem:
@lawzlo  This is a different solution form what I mentioned earlier. It will only actually indicate the completion of the first batch, but that should be plenty for the tests and I think it's simpler.
In the test simply Await on the future returned by `precomputeRecommendations`. You can either merge this or port it into your PR. Le me know which one you chose.


lawzlo:
I will merge the request. Thanks!

PR #10428, opened 2014-08-21
Enhance feedback API

lawzlo:
@stkem @yingjieMiao


kifi-backend/modules/common/app/com/keepit/curator/model/RecommendationClientType.scala
```diff
@@ -0,0 +1,14 @@
...
+@json case class RecommendationClientType(value: String)
+
+object RecommendationClientType {
+  object Site extends RecommendationClientType("site")
+  object IOS extends RecommendationClientType("ios")
+  object Android extends RecommendationClientType("android")
+  object Extension extends RecommendationClientType("extension")
+  object Email extends RecommendationClientType("email")
+  object Unknown extends RecommendationClientType("unknown")
+}
```
lawzlo: will add new line

kifi-backend/modules/common/app/com/keepit/model/UriRecommendationFeedback.scala
```diff
@@ -1,18 +1,14 @@
...
 @json case class UriRecommendationFeedback(
-    delivered: Option[Boolean] = None,
     clicked: Option[Boolean] = None,
     kept: Option[Boolean] = None,
     trashed: Option[Boolean] = None,
-    markedBad: Option[MarkedAsBad] = None) {
-  override def toString = s"UriRecommendationFeedback(delivered:$delivered, clicked:$clicked, kept:$kept, trashed:$trashed, markedBad:$markedBad)"
+    vote: Option[Boolean] = None,
+    improvement: Option[String] = None,
+    fromClient: RecommendationClientType = RecommendationClientType.Unknown) {
```
stkem, line 12: This will also have to be an option, unfortunately, to allow clients to ommit it.

kifi-backend/modules/common/conf/evolutions/shoebox/234.sql
```diff
@@ -0,0 +1,19 @@
...
+  DROP COLUMN delivered;
+
+ALTER TABLE uri_recommendation
+  ADD COLUMN email_vote boolean DEFAULT NULL;
+
+ALTER TABLE uri_recommendation
+  ADD COLUMN improvement text DEFAULT NULL;
+
+ALTER TABLE uri_recommendation
+    ADD COLUMN from_client varchar(256) DEFAULT NULL;
```
yingjieMiao, line 15: "VARCHAR values are stored as a 1-byte or 2-byte length prefix plus data. The length prefix indicates the number of bytes in the value. A column uses one length byte if values require no more than 255 bytes, two length bytes if values may require more than 255 bytes."

If I understand this correctly, claiming this field as `varchar(256)` wastes one more byte per row. `varchar(255)` or even smaller should be fine.

stkem, line 15: last two are analytics only, as discussed

kifi-backend/modules/common/conf/evolutions/shoebox/234.sql
```diff
@@ -0,0 +1,19 @@
+# CURATOR
+
+# --- !Ups
+
+ALTER TABLE uri_recommendation
+  DROP COLUMN delivered;
+
+ALTER TABLE uri_recommendation
+  ADD COLUMN email_vote boolean DEFAULT NULL;
```
stkem, line 9: unifying, as discussed

kifi-backend/modules/common/conf/evolutions/shoebox/234.sql
```diff
@@ -0,0 +1,19 @@
+# CURATOR
+
+# --- !Ups
+
+ALTER TABLE uri_recommendation
+  DROP COLUMN delivered;
```
stkem, line 6: keeping for internal use

kifi-backend/modules/common/conf/evolutions/shoebox/234.sql
stkem, line 15: good point for future reference. In this it turns out we don't need the column at all :-)

lawzlo:
opened another request, close this one.

PR #10429, opened 2014-08-21
Update feedback API, remove user interaction API, add recClientType

lawzlo:
@stkem @yingjieMiao

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/curator/CuratorServiceClient.scala
+++ b/kifi-backend/modules/common/app/com/keepit/curator/CuratorServiceClient.scala
@@ -20,7 +20,6 @@ trait CuratorServiceClient extends ServiceClient {
   def updateUriRecommendationFeedback(userId: Id[User], uriId: Id[NormalizedURI], feedback: UriRecommendationFeedback): Future[Boolean]
   def triggerEmail(code: String): Future[String]
   def triggerEmailToUser(code: String, userId: Id[User]): Future[String]
-  def updateUriRecommendationUserInteraction(userId: Id[User], uriId: Id[NormalizedURI], interaction: UriRecommendationUserInteraction): Future[Boolean]
   def resetUserRecoGenState(userId: Id[User]): Future[Unit]
 }
 
@@ -53,12 +52,6 @@ class CuratorServiceClientImpl(
     call(Curator.internal.triggerEmailToUser(code, userId), callTimeouts = longTimeout).map { response =>
       response.json.as[String]
     }
-  }
-
-  def updateUriRecommendationUserInteraction(userId: Id[User], uriId: Id[NormalizedURI], interaction: UriRecommendationUserInteraction): Future[Boolean] = {
-    call(Curator.internal.updateUriRecommendationUserInteraction(userId, uriId), body = Json.toJson(interaction)).map(response =>
-      response.json.as[Boolean]
-    )
   }
 
   def resetUserRecoGenState(userId: Id[User]): Future[Unit] = {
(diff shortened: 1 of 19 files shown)


kifi-backend/modules/common/app/com/keepit/curator/CuratorServiceClient.scala
```diff
@@ -55,12 +54,6 @@ class CuratorServiceClientImpl(
     }
   }
 
-  def updateUriRecommendationUserInteraction(userId: Id[User], uriId: Id[NormalizedURI], interaction: UriRecommendationUserInteraction): Future[Boolean] = {
-    call(Curator.internal.updateUriRecommendationUserInteraction(userId, uriId), body = Json.toJson(interaction)).map(response =>
```
yingjieMiao, line 59: did you also remove it from `Router.scala`?

yingjieMiao:
lgtm

kifi-backend/modules/common/app/com/keepit/curator/CuratorServiceClient.scala
lawzlo, line 59: :+1: removing

stkem:
:+1:

kifi-backend/modules/common/app/com/keepit/model/UriRecommendationFeedback.scala
```diff
@@ -1,18 +1,14 @@
...
 @json case class UriRecommendationFeedback(
-    delivered: Option[Boolean] = None,
     clicked: Option[Boolean] = None,
     kept: Option[Boolean] = None,
     trashed: Option[Boolean] = None,
-    markedBad: Option[MarkedAsBad] = None) {
-  override def toString = s"UriRecommendationFeedback(delivered:$delivered, clicked:$clicked, kept:$kept, trashed:$trashed, markedBad:$markedBad)"
+    vote: Option[Boolean] = None,
+    improvement: Option[String] = None,
+    fromClient: Option[RecommendationClientType] = None) {
```
2is10, line 12: We might be able to improve the names of these new fields.
- `vote` — not immediately clear what `vote: true` or `vote: false` means. Maybe `liked`? Clearer, and the past tense would be consistent with the other fields.
- `improvement` — again, not immediately clear what this means. If it’s for an open-ended comment/feedback, we might consider `feedback` or `comment`, either of which, I think, would be clearer.
- `fromClient` — `clientType` would be more straightforward. Might also consider `channel` or `medium` (and rename the type to match).

lawzlo, line 12: nice suggestions for renaming `improvement` and `fromClient`. I will change them to comment and clientType. For name `vote`. `vote: true` is user like it and want to get more similar recommendations. `vote: false` is user mark the feed as bad recommendation. But I think `liked: false` maybe mean user not liked but not user doesn't like it?

2is10, line 12: Cool. I see your concern with `liked`. I was thinking that, because the value is an `Option[Boolean]`, it would be clear that `liked: false` means we know the user did not like it, whereas `None` would mean we don’t know. But now I see that other fields also use `Option[Boolean]` in this class even though `false` will never occur, which I think contributes to this tri-state interpretation problem.

I just spoke with Stephen. We’re thinking it might be clearer to combine all of the `Option[Boolean]` fields into a single field, perhaps named `type` or `eventType` or `events` (if it’s important that multiple events can be batched). Then we would see, for example `type: kept` or `type: liked` or `type: disliked` or `events: [clicked, liked, kept]`.

PR #10431, opened 2014-08-21
Email notifications (no popup) for library invitations! Just basics for ...

ahsu1230:
...now

@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
@@ -202,15 +202,18 @@ class LibraryCommander @Inject() (
     }
   }
 
-  def userAccess(userId: Id[User], libraryId: Id[Library]): Option[LibraryAccess] = {
+  def userAccess(userId: Id[User], libraryId: Id[Library], universalLinkOpt: Option[String]): Option[LibraryAccess] = {
     db.readOnlyMaster { implicit s =>
       libraryMembershipRepo.getWithLibraryIdAndUserId(libraryId, userId) match {
         case Some(mem) =>
           Some(mem.access)
         case None =>
-          if (libraryRepo.get(libraryId).visibility == LibraryVisibility.PUBLISHED)
+          val lib = libraryRepo.get(libraryId)
+          if (lib.visibility == LibraryVisibility.PUBLISHED)
             Some(LibraryAccess.READ_ONLY)
           else if (libraryInviteRepo.getWithLibraryIdAndUserId(libraryId, userId).nonEmpty)
+            Some(LibraryAccess.READ_ONLY)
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -415,6 +420,32 @@ class LibraryCommander @Inject() (
...
+      val inviter = userRepo.get(inviterId)
+      val invitee = userRepo.get(inviteeId)
+      val library = libraryRepo.get(libraryId)
+      (inviter, invitee, library)
+    }
+    invitee.primaryEmail match {
+      case None => {}
+      case Some(email) =>
+        //val libraryLink = s"www.kifi.com/users/${inviter.username.getOrElse(inviter.externalId)}/libraries/${library.slug}"
+        val libraryLink = s"www.kifi.com/users/${inviter.username.getOrElse(inviter.externalId)}/libraries?auth=${library.universalLink}"
```
ahsu1230: still figuring out the link for now

kifi-backend/modules/shoebox/app/views/email/libraryInvitation.scala.html
```diff
@@ -0,0 +1,49 @@
...
+                <tr>
+                    <td>
+                        Hey @myFirstName, <br><br>
+                        @{inviter.firstName} @{inviter.lastName} would like to share @libraryName with you <br><br>
+
+                        <!-- message to invitee here? -->
+                        <!-- libraryDescription here? -->
+
+                        <font color="#A3C879">Come check it out here!</font><br>
+                        <a href=@libraryLink><u>@libraryName</u></a>
```
andrewconner: Quotes around `@libraryLink`

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
andrewconner: We know it'll begin with https://www.kifi.com/(username)/(slug), and then have something like auth=(universalLink)

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -197,16 +202,19 @@ class LibraryCommander @Inject() (
     }
   }
 
-  def userAccess(userId: Id[User], libraryId: Id[Library]): Option[LibraryAccess] = {
+  def userAccess(userId: Id[User], libraryId: Id[Library], universalLinkOpt: Option[String]): Option[LibraryAccess] = {
```
ahsu1230, line 205: since this part requires a userId, for the time being, it can used only by kifi-users.
basically kifi-user (if does not have membership) can view the library (in read_only):
- it's published, user has received an invite before, or provides a universalLink

PR #10432, opened 2014-08-22
Feed digest updates

joshm1:
filter digest recos further; lazy load uri summary and attribution details individually instead of loading them in batch

refactor view and introduction of a Context object to encapsulate common dynamic email values ( @2is10 I know you don't like that name and I understand with your reasoning :) may rename in next commit )

### DIFF
--- a/kifi-backend/modules/curator/app/com/keepit/curator/commanders/email/FeedDigestEmailSender.scala
+++ b/kifi-backend/modules/curator/app/com/keepit/curator/commanders/email/FeedDigestEmailSender.scala
@@ -173,25 +173,31 @@ class FeedDigestEmailSenderImpl @Inject() (
   private def getDigestRecommendationsForUser(userId: Id[User]) = {
     getRecommendationsForUser(userId).flatMap { recos =>
       FutureHelpers.findMatching(recos, RECOMMENDATIONS_TO_DELIVER, isEmailWorthy, getDigestReco)
+    }.map { seq => seq.flatten }
+  }
+
+  private def isEmailWorthy(recoOpt: Option[DigestReco]) = {
+    recoOpt match {
+      case Some(reco) =>
+        val summary = reco.uriSummary
+        val uri = reco.uri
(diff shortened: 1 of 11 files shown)


joshm1:
@stkem

stkem:
:+1:

kifi-backend/modules/curator/app/com/keepit/curator/commanders/email/FeedDigestEmailSender.scala
```diff
@@ -103,27 +117,32 @@ class FeedDigestEmailSenderImpl @Inject() (
...
+    }
+    digestRecoMailF.flatten
   }
 
-  private def composeAndSendEmail(user: User, digestRecos: Seq[DigestReco], unsubscribeUrl: String): Future[DigestRecoMail] = {
+  private def composeAndSendEmail(user: User, digestRecos: Seq[DigestReco], _unsubscribeUrl: String): Future[DigestRecoMail] = {
     val userId = user.id.get
     val emailData = AllDigestRecos(toUser = user, recos = digestRecos)
-    val htmlBody: LargeString = views.html.email.feedDigest(emailData, unsubscribeUrl).body
+    val ctx = Context(campaign = "emailDigest", unsubscribeUrl = _unsubscribeUrl)
```
2is10, line 143: I believe adding the underscore was unnecessary. If it was a stylistic choice, I can go along, but personally I’d probably only mangle a name when necessary.

2is10:
very nice

kifi-backend/modules/curator/app/com/keepit/curator/commanders/email/FeedDigestEmailSender.scala
joshm1, line 143: I'm not a fan of this either. Only did b/c for some reason that I can't reproduce right now I thought there was a conflict I tried using named variables and variables of the same name. But I just renamed and it compiled, so nevermind that

PR #10434, opened 2014-08-22
tmp disable tracking

yingjieMiao:
@stkem FYI

will enable this tomorrow. (with some logs)


stkem:
I think I know what's happening: We send every keep to curator, in case it was a recommendation. The tracking logic needs to deal with the case where it wasn't, i.e. there is no recommendation for that uriId "keep" event. Not sure if that is the cause, but it could be. Let's go through it tomorrow.

PR #10435, opened 2014-08-22
using URIs

eishay:
There are many exceptions in the code that comes from bad url format. Since we use futures in may cases its very hard to trace the source of the bad urls.
This commit adds more usage of URI which is processed by our parser and is properly escaped. By using it early on, before passing it to the http client, we can trace sources of miss-formatted URLs earlier and can deal with them at the source.
@raymondng @ymatsuda @2is10


2is10:
good idea

PR #10436, opened 2014-08-22
Add renormalized visibility fields and userId value field

leogrim:
@ymatsuda

### DIFF
--- a/kifi-backend/modules/search/app/com/keepit/search/graph/library/LibraryIndexable.scala
+++ b/kifi-backend/modules/search/app/com/keepit/search/graph/library/LibraryIndexable.scala
@@ -17,6 +17,13 @@ object LibraryFields {
   val hiddenUsersField = "h"
   val recordField = "rec"
 
+  import LibraryVisibility.{SECRET, DISCOVERABLE, PUBLISHED}
+  @inline def toNumericCode(visibility: LibraryVisibility) = visibility match {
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/search/app/com/keepit/search/graph/library/LibraryIndexable.scala
```diff
@@ -17,6 +17,13 @@ object LibraryFields {
   val hiddenUsersField = "h"
   val recordField = "rec"
 
+  import LibraryVisibility.{ SECRET, DISCOVERABLE, PUBLISHED }
+  @inline def toNumericCode(visibility: LibraryVisibility) = visibility match {
+    case SECRET => 0
+    case DISCOVERABLE => 1
+    case PUBLISHED => 2
+  }
+
```
leogrim, line 26: I haven't found an exemple of a state being stored as a value field. So the idea here is to defend against potential changes made to LibraryVisibility:
1. The compiler will warn us about the introduction of new states (sealed class)
2. Using a numeric value instead of a String for compression and in case the visibility options are renamed.
Does this make sense?

ymatsuda, line 26: nice!

ymatsuda:
lgtm

PR #10437, opened 2014-08-22
Update client code to use new WTI query parameters.

yipingliao:
@eishay @leogrim


eishay:
thanks, please let me know when its in production.

PR #10438, opened 2014-08-22
All existing libraries do not have this field. Needs to be optional.

andrewconner:
@ahsu1230 Library endpoints were broken, this should fix it.

When we add new fields, we need to make sure we're backwards compatible with the existing db data.


ahsu1230:
ah whoops, my bad.
Thanks for fixing!

PR #10444, opened 2014-08-22
Making curator robust against normalization weirdness

stkem:
\+ minor code cleanup (breaking up a big function)
@yingjieMiao (cc @leogrim @lawzlo)


stkem:
merging now to stop exceptions in prod

PR #10447, opened 2014-08-22
Start of library.js, API json changes

andrewconner:
@ahsu1230 Included are some API changes. Specifically, the JSON has been flattened a bit to be easier to work with client side. I also noticed that keeps, followers, and collaborators were currently returning every of each, which will present some scaling problems. It's left as is, with a comment.


andrewconner:
Self-merging because I need the API updates for the client. Please review still :)

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
```diff
@@ -86,7 +86,7 @@ object KeepInfo {
     )
   }
 
-  def fromBookmark(bookmark: Keep): KeepInfo = {
+  def fromKeep(bookmark: Keep): KeepInfo = {
     KeepInfo(Some(bookmark.externalId), bookmark.title, bookmark.url, bookmark.isPrivate, libraryId = None) // todo(andrew): Add library external id
```
ahsu1230, line 90: just curious, could we fix this libraryId part yet? Or do we have to wait until all keeps receive a library?

ahsu1230:
yeah looks good! 
For viewing groupholders and keep holders, how do you think we should go about that?

andrewconner:
I am removing those. After using the API client side, it was kind of awkward: `library.keeps.keeps` and such.

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
andrewconner, line 90: We could, I'll do that. Thanks.

ahsu1230:
oh sorry should clarify. I'm talking about the view only first x keeps/users.
Would it work like the page APIs in admin where you view x objects at a time & use an index to keep track of which set of x objects you're looking at?

andrewconner:
Ah yeah. We'll need a method of pagination. I just added todos so we don't forget. As soon as libraries have 1000s of keeps, we'll need to be careful about how much data we send down.

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
ahsu1230, line 90: earlier was looking around & I think I saw one of your own reminders at line 205 in KeepInterner.scala about the same thing. It's for assigning a libraryId to a new keep, which could change that function & how RawBookmarkRepresentation works

kifi-backend/modules/shoebox/angular/src/libraries/libraryService.js
```diff
@@ -23,13 +23,30 @@ angular.module('kifi')
...
...\u1f4d\u1f59\u1f5b\u1f5d\u1f5f\u1f68-\u1f6f\u1fb8-\u1fbb\u1fc8-\u1fcb\u1fd8-\u1fdb\u1fe8-\u1fec\u1ff8-\u1ffb\u2102\u2107\u210b-\u210d\u2110-\u2112\u2115\u2119-\u211d\u2124\u2126\u2128\u212a-\u212d\u2130-\u2133\u213e\u213f\u2145\u2183\u2c00-\u2c2e\u2c60\u2c62-\u2c64\u2c67\u2c69\u2c6b\u2c6d-\u2c70\u2c72\u2c75\u2c7e-\u2c80\u2c82\u2c84\u2c86\u2c88\u2c8a\u2c8c\u2c8e\u2c90\u2c92\u2c94\u2c96\u2c98\u2c9a\u2c9c\u2c9e\u2ca0\u2ca2\u2ca4\u2ca6\u2ca8\u2caa\u2cac\u2cae\u2cb0\u2cb2\u2cb4\u2cb6\u2cb8\u2cba\u2cbc\u2cbe\u2cc0\u2cc2\u2cc4\u2cc6\u2cc8\u2cca\u2ccc\u2cce\u2cd0\u2cd2\u2cd4\u2cd6\u2cd8\u2cda\u2cdc\u2cde\u2ce0\u2ce2\u2ceb\u2ced\ua640\ua642\ua644\ua646\ua648\ua64a\ua64c\ua64e\ua650\ua652\ua654\ua656\ua658\ua65a\ua65c\ua65e\ua660\ua662\ua664\ua666\ua668\ua66a\ua66c\ua680\ua682\ua684\ua686\ua688\ua68a\ua68c\ua68e\ua690\ua692\ua694\ua696\ua722\ua724\ua726\ua728\ua72a\ua72c\ua72e\ua732\ua734\ua736\ua738\ua73a\ua73c\ua73e\ua740\ua742\ua744\ua746\ua748\ua74a\ua74c\ua74e\ua750\ua752\ua754\ua756\ua758\ua75a\ua75c\ua75e\ua760\ua762\ua764\ua766\ua768\ua76a\ua76c\ua76e\ua779\ua77b\ua77d\ua77e\ua780\ua782\ua784\ua786\ua78b\ua78d\ua790\ua7a0\ua7a2\ua7a4\ua7a6\ua7a8\uff21-\uff3a]/g; // jshint ignore:line
```
2is10, line 38: we may want to find an approximation for this that is significantly less than 2KB of JS code, or make an AJAX call for the suggested slug

PR #10448, opened 2014-08-22
chunkySequentialExec

raymondng:
Added chunky support to `sequentialExec` (i.e. `chunkySequentialExec`) + optional callback support -- typically for reporting/logging purposes but could be anything. Simple test included. Wired up in a couple places in HelpRank processing (replaces `sequentialExec`).

Also added short-circuitry logic to `foldLeft` (i.e. `foldLeftWhile`) -- an implementation detail (private) for now.

@ymatsuda @yingjieMiao @andrewconner @leogrim @stkem


kifi-backend/modules/common/app/com/keepit/common/concurrent/PimpedFuture.scala
```diff
@@ -60,16 +62,48 @@ object FutureHelpers {
...
+  private def chunkyExec[I, T](iter: Iterator[(Iterable[I], Int)], chunkSize: Int, promised: Promise[Unit] = Promise[Unit]())(f: I => Future[T], chunkCB: Option[Int => Unit] = None)(implicit ec: ScalaExecutionContext): Future[Unit] = {
+    if (iter.isEmpty) promised.success(())
+    else {
+      val items = iter.next
+      sequentialExec(items._1)(f) onComplete {
+        case Success(_) =>
+          chunkCB.foreach { _.apply(items._2) }
+          chunkyExec(iter, chunkSize, promised)(f, chunkCB)
+        case Failure(t) =>
+          log.error(s"[chunkyExec] Caught exception $t while processing chunk#${items._2} items=${items._1.take(5)}", t)
```
stkem: I don't think pure utility methods like this should be logging anything/side effecting internally. It's the responsibility fo the caller to deal with errors.

kifi-backend/modules/common/app/com/keepit/common/concurrent/PimpedFuture.scala
```diff
@@ -60,16 +62,48 @@ object FutureHelpers {
...
 
   def sequentialExec[I, T](items: Iterable[I])(f: I => Future[T])(implicit ec: ScalaExecutionContext): Future[Unit] = {
-    foldLeft(items)(()) { case ((), nextItem) => f(nextItem).map { _ => () } }
+    foldLeftWhile(items)(()) { case ((), nextItem) => f(nextItem).map { _ => () } }
+  }
+
+  private val noopChunkCB: Int => Unit = _ => Unit
+
+  // sequential execute in chunks + callback (optional)
+  def chunkySequentialExec[I, T](items: Iterable[I], chunkSize: Int = 10, chunkCB: Int => Unit = noopChunkCB)(f: I => Future[T])(implicit ec: ScalaExecutionContext) = {
```
stkem, line 71: I'm note sure I get the purpose of this (over regular `sequentialExec`), since it looks like it's still executing all all items sequentially. Is it just the reporting/callback functionality per chunk?

kifi-backend/modules/common/app/com/keepit/common/concurrent/PimpedFuture.scala
raymondng: I debated about this one as well. I tend to agree with you -- let me remove it.

kifi-backend/modules/common/app/com/keepit/common/concurrent/PimpedFuture.scala
raymondng, line 71: Yes that's pretty much it -- the utility is real though, at least for the use cases that I have in mind.

I should add that "chunky" approach also allows for better control/monitoring/error-handling -- something likely to be added later.

kifi-backend/modules/common/app/com/keepit/common/concurrent/PimpedFuture.scala
```diff
@@ -60,16 +62,46 @@ object FutureHelpers {
...
+        case Success(_) =>
+          chunkCB.foreach { _.apply(items._2) }
+          chunkyExec(iter, chunkSize, promised)(f, chunkCB)
+        case Failure(t) => promised.failure(t)
+      }
+    }
+    promised.future
   }
 
   def foldLeft[I, T](items: Iterable[I], promisedResult: Promise[T] = Promise[T]())(accumulator: T)(fMap: (T, I) => Future[T])(implicit ec: ScalaExecutionContext): Future[T] = {
```
andrewconner: Since this is now delegating to `foldLeftWhile`, should `promisedResult` be in here anymore?

raymondng: Good point. Agreed there's no longer a need for it. Will remove.

kifi-backend/modules/common/app/com/keepit/common/concurrent/PimpedFuture.scala
```diff
@@ -60,16 +62,46 @@ object FutureHelpers {
...
-    if (items.isEmpty) { promisedResult.success(accumulator) }
-    else fMap(accumulator, items.head).onComplete {
-      case Success(updatedAccumulator) => foldLeft(items.tail, promisedResult)(updatedAccumulator)(fMap)
-      case Failure(ex) => promisedResult.failure(ex)
+  def foldLeft[I, T](items: Iterable[I])(accumulator: T)(fMap: (T, I) => Future[T])(implicit ec: ScalaExecutionContext): Future[T] = {
+    foldLeftWhile(items)(accumulator)(fMap)
+  }
+
+  private def foldLeftWhile[I, T](items: Iterable[I], promised: Promise[T] = Promise[T]())(acc: T)(fMap: (T, I) => Future[T], pred: Option[(I) => Boolean] = None)(implicit ec: ScalaExecutionContext): Future[T] = {
+    if (items.isEmpty) { promised.success(acc) }
```
leogrim, line 94: We could avoid making pred an Option and have the default be Function.const(true)
Also would this be more flexible is pred was on T rather I? This means the predicate has to be on the original item, would be nice to be able to filter on the result of the async computation. What do you think?

raymondng, line 94: Thanks for the feedback! 
I like `Function.const` but I feel that `Option` is perhaps a bit cheaper to check. Not married to it at all -- please let me know if I missed anything.
Also, very good suggestion about T. How about make both T & I available?

raymondng:
Thanks for everyone's feedback! I'll plan to merge by EOD today & iterate.
Please let me know if you have any objections/concerns.

PR #10456, opened 2014-08-23
friends added to Feed digest

joshm1:
@stkem @leogrim


kifi-backend/modules/common/app/com/keepit/common/db/Id.scala
```diff
@@ -21,6 +21,27 @@ object Id {
   implicit def format[T]: Format[Id[T]] =
     Format(__.read[Long].map(Id(_)), new Writes[Id[T]] { def writes(o: Id[T]) = JsNumber(o.id) })
 
+  implicit def mapOfIdToObjectFormat[S, T](implicit t: Format[T]): Format[Map[Id[S], T]] = new Format[Map[Id[S], T]] {
```
2is10, line 24: Nice!

PR #10458, opened 2014-08-24
Reco endpoints and retrieval logic for the site phase 1

stkem:
Missing actual data flow for public recos since those are not in yet, but should be more or less feature complete otherwise.
@yingjieMiao @lawzlo @yipingliao


yipingliao:
Thanks for putting this together! 

I couldn't find what fields are in `uriRecommendationFeedback`, but it looks like that was already defined in:
https://github.com/kifi/keepit/blob/master/kifi-backend/modules/common/app/com/keepit/model/UriRecommendationFeedback.scala

It looks like the `fromClient` field is something that the server will determine based on user agent? Or is that something that clients should supply?

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationRetrievalCommander.scala
```diff
@@ -0,0 +1,55 @@
...
+    val adjustedScore = masterScore + recencyWeight * (4 * scores.recentInterestScore + 2 * scores.recencyScore)
+    (adjustedScore * finalPenaltyFactor).toFloat
+  }
+
+  def topRecos(userId: Id[User], more: Boolean = false, recencyWeight: Float = 0.5f, clientType: RecommendationClientType): Seq[RecoInfo] = {
+    require(recencyWeight <= 1.0f && recencyWeight >= 0.0f, "recencyWeight must be between 0 and 1")
+
+    val recos = db.readOnlyReplica { implicit session =>
+      uriRecRepo.getByTopMasterScore(userId, 1000)
+    } filterNot (_.kept) filter (_.trashed) map { reco =>
```
yingjieMiao: could combine the two filter into one to save some time. Also, second filter should be filterNot ?

kifi-backend/modules/common/app/com/keepit/curator/model/RecommendationInfo.scala
```diff
@@ -1,14 +1,40 @@
...
+  id: ExternalId[NormalizedURI],
+  title: Option[String],
+  url: String,
+  keepers: Seq[BasicUser],
+  others: Int,
+  siteName: Option[String],
+  uriSummary: URISummary)
+
+@json case class FullRecoInfo(
+  kind: String,
```
yingjieMiao: The `kind` here is different than the `kind` in `RecoAttributionInfo`? How about using trait and case class/object?

kifi-backend/modules/common/app/com/keepit/curator/model/RecommendationInfo.scala
```diff
@@ -1,14 +1,40 @@
...
+
+@json case class RecoAttributionInfo(
+  kind: String,
+  name: Option[String],
+  url: Option[String],
+  when: Option[DateTime])
+
+@json case class RecoMetaData(
+  attribution: Seq[RecoAttributionInfo])
+
```
yingjieMiao, line 39: The motivation/advantage of breaking `SeedAttribution` into a seq of `RecoAttributionInfo` is not clear to me. Maybe that's the API contract you had with the frontend?

kifi-backend/modules/shoebox/app/com/keepit/commanders/RecommendationsCommander.scala
```diff
@@ -66,4 +81,80 @@ class RecommendationsCommander @Inject() (
...
+        kind = "topic",
+        name = Some(topicAttr.topicName),
+        url = None,
+        when = None
+      )
+    } getOrElse keepAttrInfos
+  }
+
+  def topRecos(userId: Id[User], clientType: RecommendationClientType, more: Boolean, recencyWeight: Float): Future[Seq[FullRecoInfo]] = {
+    curator.topRecos(userId, clientType, more, recencyWeight).flatMap { recos =>
```
yingjieMiao, line 133: `topRecos` and `topPublicRecos` are also defined in `RecommendationRetrievalCommander`. The same method name and almost identical signatures look a bit confusing.

yingjieMiao:
a few comments. The data flow lgtm overall.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationRetrievalCommander.scala
stkem: ah, good catch. It's indeed `filterNot`. Fixing (and combining)

kifi-backend/modules/common/app/com/keepit/curator/model/RecommendationInfo.scala
stkem: It is different, yes. It's currently always "keep", but having it typed nonetheless is a good idea. I'll make that change.

kifi-backend/modules/common/app/com/keepit/curator/model/RecommendationInfo.scala
stkem, line 39: Yup, contract with the front end.

kifi-backend/modules/shoebox/app/com/keepit/commanders/RecommendationsCommander.scala
stkem, line 133: well, they basically do the same thing. This one is sort of a proxy for the one on curator, hence the same name.

stkem:
@yipingliao current plan is to figure out the client type from the route used (i.e. /site/.. and such). We might not be able to keep that up, but for now it keeps things simple.

stkem:
UserAgent does not work, because we would no be abel to tell the difference between the site and the extension.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationRetrievalCommander.scala
lawzlo: I'm thinking should do this filtering in repo level instead of in commander level since all those information(kept, trashed) are in same db table?

stkem: you have a point. moving it to the db.

PR #10459, opened 2014-08-25
Quick fix for tags being broken

andrewconner:
Routing precedence strikes again. Will make a test for this, and come up with a long-term solution.

@eishay @yipingliao


yipingliao:
Consolidating routes into a single file sounds like a great idea. Let me know if you want help writing tests.

PR #10460, opened 2014-08-25
Adding seq column in evolution file

lawzlo:
@stkem


stkem:
Huh? Why does this have a sequence number?

PR #10463, opened 2014-08-25
Caching for URISummaries used in notifications

stkem:
This currently errs on the side of speed very aggressively, i.e. if an item is not in the cache proceed immediately and fetch it asynchronously in the background (never waiting for a result).
I chose to keep this cache out of the service client and very Eliza specific for two reasons:
1) All the service client caches we have have "wait for the result if it's not cached" semantics, whereas here we want to proceed immediately if it's not there (and trigger a computation in the background)
2) Virtually all the service client level caches are key-value, whereas here we have a rather complex request object. Caching at the service level would require somehow encoding the request in the key, which is tricky and error prone.
@2is10

### DIFF
--- a/kifi-backend/modules/eliza/app/com/keepit/eliza/commanders/NotificationJsonMaker.scala
+++ b/kifi-backend/modules/eliza/app/com/keepit/eliza/commanders/NotificationJsonMaker.scala
@@ -137,4 +137,3 @@ case class InboxUriSummaryCacheKey(uriId: Id[NormalizedURI]) extends Key[URISumm
 class InboxUriSummaryCache(stats: CacheStatistics, accessLog: AccessLog, innermostPluginSettings: (FortyTwoCachePlugin, Duration), innerToOuterPluginSettings: (FortyTwoCachePlugin, Duration)*)
   extends JsonCacheImpl[InboxUriSummaryCacheKey, URISummary](stats, accessLog, innermostPluginSettings, innerToOuterPluginSettings: _*)
 
-//ZZZ remember cache module
--- a/kifi-backend/modules/eliza/app/com/keepit/eliza/commanders/NotificationJsonMaker.scala
+++ b/kifi-backend/modules/eliza/app/com/keepit/eliza/commanders/NotificationJsonMaker.scala
@@ -93,6 +93,23 @@ private[commanders] class NotificationJsonMaker @Inject() (
     }
   }
 
+  private def uriSummary(value: JsValue, uriId: Id[NormalizedURI]): Option[URISummary] = {
+    value.asOpt[String].map { url =>
+      val result = summaryCache.get(InboxUriSummaryCacheKey(uriId))
+      if (result.isEmpty) new SafeFuture(fetchAndCacheUriSummary(uriId, url), Some("Fetching URI summary for extension inbox"))
+      result
+    }.flatten
+  }
+
+  private def updateBasicUser(basicUser: BasicUser): Future[BasicUser] = {
+    shoebox.getUserOpt(basicUser.externalId) map { userOpt =>
+      userOpt.map(BasicUser.fromUser).getOrElse(basicUser)
+    } recover {
+      case _ =>
+        basicUser
+    }
+  }
+
   private def fetchAndCacheUriSummary(uriId: Id[NormalizedURI], url: String): Future[URISummary] = uriSummaryRequestLimiter.withLockFuture {
     shoebox.getUriSummary(
       URISummaryRequest(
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/eliza/app/com/keepit/eliza/commanders/NotificationJsonMaker.scala
```diff
@@ -82,21 +93,28 @@ private[commanders] class NotificationJsonMaker @Inject() (
...
+      )
+    ).map { uriSummary =>
+        summaryCache.set(InboxUriSummaryCacheKey(uriId), uriSummary)
+        uriSummary
+      }
+  }
+
+  private def uriSummary(value: JsValue, uriId: Id[NormalizedURI]): Option[URISummary] = {
+    value.asOpt[String].map { url =>
+      val result = summaryCache.get(InboxUriSummaryCacheKey(uriId))
```
2is10, line 98: Do our cache lookups usually block?

kifi-backend/modules/eliza/app/com/keepit/eliza/commanders/NotificationJsonMaker.scala
```diff
@@ -82,21 +93,28 @@ private[commanders] class NotificationJsonMaker @Inject() (
...
-          url = url,
-          imageType = ImageType.ANY,
-          minSize = ImageSize(65, 95),
-          withDescription = false,
-          waiting = true,
-          silent = false)) map Some.apply
-    } getOrElse {
-      Future.successful(None)
-    }
+  private def fetchAndCacheUriSummary(uriId: Id[NormalizedURI], url: String): Future[URISummary] = uriSummaryRequestLimiter.withLockFuture {
```
2is10: With the exception of this new method, this file is organized with private helpers defined in the same order that they’re encountered reading top-to-bottom. Moving this method beneath `uriSummary` below would continue the convention. If you don’t like that convention, don’t feel obligated to move it.

kifi-backend/modules/eliza/app/com/keepit/eliza/commanders/NotificationJsonMaker.scala
```diff
@@ -82,21 +93,28 @@ private[commanders] class NotificationJsonMaker @Inject() (
...
+    ).map { uriSummary =>
+        summaryCache.set(InboxUriSummaryCacheKey(uriId), uriSummary)
+        uriSummary
+      }
+  }
+
+  private def uriSummary(value: JsValue, uriId: Id[NormalizedURI]): Option[URISummary] = {
+    value.asOpt[String].map { url =>
+      val result = summaryCache.get(InboxUriSummaryCacheKey(uriId))
+      if (result.isEmpty) new SafeFuture(fetchAndCacheUriSummary(uriId, url), Some("Fetching URI summary for extension inbox backgrounds"))
```
2is10: let’s remove the word “backgrounds” (new design puts them in foreground)

2is10:
> This currently errs on the side of speed very aggressively, i.e. if an item is not in the cache proceed immediately and fetch it asynchronously in the background (never waiting for a result).

This is probably good for now. It may turn out that we’ll need to delay the notifications we push down the socket (for new messages) until we have the UriSummary.

Thanks for the quick turnaround!

kifi-backend/modules/eliza/app/com/keepit/eliza/commanders/NotificationJsonMaker.scala
stkem, line 98: Yes, all of them do. The underlying library is not very amicable to reactive patterns, unfortunately.

kifi-backend/modules/eliza/app/com/keepit/eliza/commanders/NotificationJsonMaker.scala
stkem: cool, removing

kifi-backend/modules/eliza/app/com/keepit/eliza/commanders/NotificationJsonMaker.scala
```diff
@@ -82,21 +93,12 @@ private[commanders] class NotificationJsonMaker @Inject() (
...
-          minSize = ImageSize(65, 95),
-          withDescription = false,
-          waiting = true,
-          silent = false)) map Some.apply
-    } getOrElse {
-      Future.successful(None)
-    }
+  private def uriSummary(value: JsValue, uriId: Id[NormalizedURI]): Option[URISummary] = {
+    value.asOpt[String].map { url =>
+      val result = summaryCache.get(InboxUriSummaryCacheKey(uriId))
```
ymatsuda, line 98: Why not using getOrElseFuture?

stkem, line 98: We don't want to wait for the result if the data is not in the cache, i.e. if there is a cache miss, this just return `None` immediately and asynchronously refreshed the cache.

ymatsuda, line 98: Doesn't getOrElseFuture work like that? It immediately returns a future. You can simply check if the future is completed already. You don't have to wait on the future. If it is completed, it is a cache hit, otherwise, it is a cache miss. Am I missing something?

stkem, line 98: You are right. I'll change it. Makes it simpler.

PR #10464, opened 2014-08-25
consolidate  multiple graph-relation caches with single relatedEntities cache

joshm1:
feed digest to wait for PYMK instead of depending on cache

@leogrim @stkem


kifi-backend/modules/common/app/com/keepit/graph/GraphServiceClient.scala
```diff
@@ -126,41 +124,39 @@ class GraphServiceClientImpl @Inject() (
...
-    }
+    getSociallyRelatedEntities(userId, bePatient).map(_.map(_.emailAccounts))
   }
 
   def explainFeed(userId: Id[User], uriIds: Seq[Id[NormalizedURI]]): Future[Seq[GraphFeedExplanation]] = {
     val payload = Json.obj("user" -> userId, "uris" -> uriIds)
     call(Graph.internal.explainFeed(), payload, callTimeouts = longTimeout).map { r => (r.json).as[Seq[GraphFeedExplanation]] }
   }
+
+  private def cachedEntities(id: Id[User]) = cacheProvider.relatedEntitiesCache.get(SociallyRelatedEntitiesCacheKey(id))
```
leogrim: I don't think it's necessary to factor this out since it's a one-liner and is only used once.

kifi-backend/modules/common/app/com/keepit/graph/GraphServiceClient.scala
```diff
@@ -126,41 +124,39 @@ class GraphServiceClientImpl @Inject() (
     call(Graph.internal.refreshSociallyRelatedEntities(id), callTimeouts = longTimeout).map(_ => ())
   }
 
+  def getSociallyRelatedEntities(userId: Id[User], bePatient: Boolean): Future[Option[SociallyRelatedPeople]] = {
+    cachedEntities(userId) match {
+      case Some(relatedEntities) => Future.successful(Some(relatedEntities))
+      case None => {
+        if (bePatient) call(Graph.internal.getSociallyRelatedEntities(userId), callTimeouts = longTimeout).map { r => Some(r.json.as[SociallyRelatedPeople]) }
+        else {
+          refreshSociallyRelatedEntities(userId)
```
leogrim: Can we just call Graph.internal.getSociallyRelatedEntities(userId) without mapping on the resulting future instead? And get rid of refreshSociallyRelatedEntities altogether? 
I plan on having a smarter endpoint for purely side-effecting refreshes, so the current endpoint won't be used anyway.

kifi-backend/modules/common/app/com/keepit/graph/model/RelatedEntities.scala
```diff
@@ -23,38 +23,26 @@ object RelatedEntities {
...
-}
-
-class SociallyRelatedUsersCache(stats: CacheStatistics, accessLog: AccessLog, innermostPluginSettings: (FortyTwoCachePlugin, Duration), innerToOuterPluginSettings: (FortyTwoCachePlugin, Duration)*)
-  extends JsonCacheImpl[SociallyRelatedUsersCacheKey, RelatedEntities[User, User]](stats, accessLog, innermostPluginSettings, innerToOuterPluginSettings: _*)
-
-case class SociallyRelatedLinkedInAccountsCacheKey(id: Id[User]) extends Key[RelatedEntities[User, SocialUserInfo]] {
-  override val version = 3
-  val namespace = "socially_related_linkedin_accounts"
-  def toKey(): String = id.id.toString
+case class SociallyRelatedPeople(
```
leogrim: Can we rename this to SociallyRelatedEntities? (it may contain more than people later on)

leogrim:
Cf comments, looks good otherwise!

PR #10465, opened 2014-08-25
Remove seq num from PublicFeed

lawzlo:
@stkem


stkem:
:+1:

stkem:
remember to update the actual db

lawzlo:
Thanks for reminding. Will do.

PR #10467, opened 2014-08-25
send keepers info to heimdal

yingjieMiao:
@raymondng  @stkem


raymondng:
Thanks @yingjieMiao @stkem 

It looks like the contract for Heimdal eventing is pretty weak -- I think you have mentioned this before. As this is for internal processing and persisting to relational db, I am thinking a stronger contract might be appropriate. That's admittedly a bigger change (so, no immediate impact). Thoughts?

stkem:
lgtm

stkem:
and yes, we should think about having some stronger contracts (ideally enfornced with types) on the heimdal events used in help rank. Not really sure how to go about that yet though.

raymondng:
Cool -- good to know there's consensus (in principle). We'll sort things out along the way.

PR #10468, opened 2014-08-25
commander APIs for type ahead for invites / sharing

ahsu1230:
ordering based on recent user interactions

in UserValueRepo, name key is RECENT_INTERACTION whose value is an array of Json objects in format:
[{"userType": "user", "id": 1, "action": "messaging"}, {"userType": "email", "id": "me@gmail.com", "action": 
messaging"}]
and it only contains the recent 100 interactions between any users
I added in an "action" field in case specific interactions should get heavier weights. Right now, all actions (inviting, sharing, messaging) are all the same weight

call usercommander.getInteractionScores to get a list of (user, score) sorted in decreasing order

@andrewconner


kifi-backend/modules/shoebox/app/com/keepit/commanders/UserCommander.scala
```diff
@@ -912,6 +913,63 @@ class UserCommander @Inject() (
   }
 
   protected def userAvatarImageUrl(user: User) = s3ImageStore.avatarUrlByUser(user)
+
+  def addInteraction(uid: Id[User], src: Either[Id[User], EmailAddress], interaction: UserInteraction) = {
+    val interactions = db.readOnlyMaster { implicit s =>
+      userValueRepo.getValue(uid, UserValues.recentInteractions)
+    }.as[List[JsObject]]
+    val newJson = src match {
+      case Left(id) => Json.obj("userType" -> "user", "id" -> id, "action" -> interaction.value)
```
andrewconner: Mind using `"type"`? "userType" implies that they're a user, and really it's just an identifier.

Alternatively, a smaller JSON would be `{"user": id, "action": interaction.value}` and `{"email": email, "action": interaction.value}`

kifi-backend/modules/shoebox/app/com/keepit/commanders/UserCommander.scala
```diff
@@ -912,6 +913,63 @@ class UserCommander @Inject() (
...
+      val sumScores = scores.groupBy(t => t._2).map { b =>
+        val sum = b._2.foldLeft(0.0)((r, c) => r + c._1)
+        (sum, b._1)
+      }
+      sumScores.toSeq.sortBy(-_._1)
+    }
+  }
+}
+
+sealed abstract class UserInteraction(val value: String, val score: Double)
```
andrewconner: Wow! Very nice. We'll be able to weight each interaction type... sweet.

kifi-backend/modules/shoebox/app/com/keepit/commanders/UserCommander.scala
```diff
@@ -912,6 +913,63 @@ class UserCommander @Inject() (
...
+  case object MESSAGE_USER extends UserInteraction("message", 1.0)
+
+  val maximumInteractions = 100
+
+  def getScoreForAction(action: String) = {
+    action match {
+      case INVITE_KIFI.value => INVITE_KIFI.score
+      case INVITE_LIBRARY.value => INVITE_LIBRARY.score
+      case MESSAGE_USER.value => MESSAGE_USER.score
+      case _ => 1.0
```
andrewconner: Not needed. `sealed` above saves you.

ahsu1230: oh right, thanks for the catch

kifi-backend/modules/shoebox/app/com/keepit/commanders/UserCommander.scala
```diff
@@ -912,6 +913,63 @@ class UserCommander @Inject() (
...
+    }
+  }
+
+  // given an index position in an array and weight of action, calculate score
+  private def calcInteractionScore(idx: Int, action: String): Double = {
+    val score = UserInteraction.getScoreForAction(action)
+    (15 * Math.pow(idx + 1.5, -0.7) + 0.5) * score
+  }
+
+  def getInteractionScores(uid: Id[User]): Seq[(Double, JsObject)] = {
```
andrewconner: It may be nice to have a nicer return type. The user of this API _probably_ doesn't care about what type of interaction it was, and _really_ cares about getting at the user/non-user type and identifier. Perhaps `Seq[(Double, Either[Id[User], EmailAddress])]`? Or create a super-type `InteractionEntity` around wrappers for both?

andrewconner:
This is really good. Nice job! Would you mind moving it out of `UserCommander`. Fortunately, it's relatively self-contained, and `UserCommander` is getting quite huge. Perhaps make a `UserInteractionCommander`?

andrewconner:
@raymondng finished the type ahead refactor, so that's ready. Feel free to merge when you're ready, and grab me if the type ahead requirements aren't clear.

andrewconner:
In terms of endpoints, we'll want one for the site, and one for messaging. But this can be exposed through the typeahead endpoint (ie, one that takes a query and returns results) — it's just the return when there's no query, so no need to expose this commander method directly

2is10:
Thanks, Aaron. Looks cool.

Q: If my last 100 interactions were 90 with my wife and 10 with Eishay, then will I only get two recommendations?

We might want to ensure we’re keeping a minimum of N contacts, for some small N. (I haven’t read this pull request closely, so I could be understanding incorrectly.)

Also, parsing an array of 100 JSON objects seems like a lot of work just to record a new interaction. Okay, I guess, if not blocking an HTTP response.

ahsu1230:
@2is10 yea, since the array only keeps track of your recent 100 interactions, in the scenario you described, you would only get the people you interacted with that appeared in the array. So, 2 interactions.
I think having a minimum of N contacts is definitely a good idea, which could mean we'll need an interaction repo table (due to not enough contacts from the 100 recent interactions)?

andrewconner:
@ahsu1230 @2is10 Yep. We'll improve this with time.

@ahsu1230 It is possible to append to the list without parsing the json, but it should be pretty fast, so we can wait.

Make sure the endpoint uses `readOnlyMaster` so that it gets cached. @eishay: we need to come up with a solution about the replica and caches, specifically `UserValue` (since almost all are in readOnly blocks).

PR #10472, opened 2014-08-25
cleanup(currently deactivate) low master score recos for user

lawzlo:
@stkem @yingjieMiao


kifi-backend/modules/curator/app/com/keepit/curator/model/UriRecommendationRepo.scala
```diff
@@ -87,6 +88,29 @@ class UriRecommendationRepoImpl @Inject() (
...
+	            SELECT master_score FROM uri_recommendation
+	            WHERE state=${UriRecommendationStates.ACTIVE} AND user_id=$userId
+	            ORDER BY master_score DESC LIMIT $limitNumRecosForUser
+              ) AS mScoreTable""".as[Float].first
+
+      val query =
+        sql"""UPDATE uri_recommendation
+              SET state=${UriRecommendationStates.INACTIVE}
+              WHERE (state=${UriRecommendationStates.ACTIVE} AND user_id=$userId AND master_score<$limitScore);""".asUpdate.first > 0
+
```
lawzlo, line 116: The function is not thread safe since the update process is consist of 3 queries. But this task is for roughly cleanup low master scores which may not recommend to users so I can't think of the reason will lead to any issue by myself. Please help to correct me if I missed anything. Thanks.

kifi-backend/modules/curator/test/com/keepit/curator/RecommendationCleanupCommanderTest.scala
```diff
@@ -0,0 +1,59 @@
...
+          repo.save(recs(5))
+        }
+
+        val commander = inject[RecommendationCleanupCommander]
+        val update = commander.cleanupLowMasterScoreRecos(Some(4))
+        update === true
+
+        db.readOnlyMaster { implicit s =>
+          val recos = repo.getByTopMasterScore(Id[User](42), 5)
+          recos.size === 4
```
yingjieMiao, line 54: it's best to check the actual id set.

yingjieMiao:
lgtm

kifi-backend/modules/curator/test/com/keepit/curator/RecommendationCleanupCommanderTest.scala
lawzlo, line 54: cool. adding

lawzlo:
tested, pushing to prod.

lawzlo:
I think I missed include recent recos as a condition. Adding on.

PR #10473, opened 2014-08-25
prepend "https" to email urls that start with "//"

joshm1:
@stkem


kifi-backend/modules/curator/app/com/keepit/curator/commanders/email/FeedDigestEmailSender.scala
```diff
@@ -132,10 +133,14 @@ class FeedDigestEmailSenderImpl @Inject() (
     val userId = user.id.get
     log.info(s"sending engagement feed email to $userId")
 
+    val recosF = getDigestRecommendationsForUser(userId)
+    val unsubUrlF = shoebox.getUnsubscribeUrlForEmail(user.primaryEmail.get)
+    val friendRecoF = getFriendRecommendationsForUser(userId)
```
joshm1, line 138: moved out of the for comprehension so they run in parallel

stkem:
:+1:

PR #10476, opened 2014-08-25
Cleanup recos but not include most recent 14 days

lawzlo:
@stkem @yingjieMiao

### DIFF
--- a/kifi-backend/modules/curator/app/com/keepit/curator/model/UriRecommendationRepo.scala
+++ b/kifi-backend/modules/curator/app/com/keepit/curator/model/UriRecommendationRepo.scala
@@ -20,7 +20,7 @@ trait UriRecommendationRepo extends DbRepo[UriRecommendation] {
   def getRecommendableByTopMasterScore(userId: Id[User], maxBatchSize: Int)(implicit session: RSession): Seq[UriRecommendation]
   def getNotPushedByTopMasterScore(userId: Id[User], maxBatchSize: Int, uriRecommendationState: Option[State[UriRecommendation]] = Some(UriRecommendationStates.ACTIVE))(implicit session: RSession): Seq[UriRecommendation]
   def updateUriRecommendationFeedback(userId: Id[User], uriId: Id[NormalizedURI], feedback: UriRecommendationFeedback)(implicit session: RSession): Boolean
-  def cleanupLowMasterScoreRecos(limitNumRecosForUser: Int)(implicit session: RSession): Boolean
+  def cleanupLowMasterScoreRecos(limitNumRecosForUser: Int, before: DateTime)(implicit session: RSession): Boolean
 }
 
 @Singleton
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/curator/app/com/keepit/curator/model/UriRecommendationRepo.scala
```diff
@@ -106,13 +105,17 @@ class UriRecommendationRepoImpl @Inject() (
 	              SELECT master_score
 	              FROM uri_recommendation
 	              WHERE state=${UriRecommendationStates.ACTIVE} AND user_id=$userId
+	                    AND created_at<$before
```
yingjieMiao: should we use 'updated_at' instead? @stkem

kifi-backend/modules/curator/app/com/keepit/curator/model/UriRecommendationRepo.scala
```diff
@@ -106,13 +105,17 @@ class UriRecommendationRepoImpl @Inject() (
...
 
       val query =
         sql"""UPDATE uri_recommendation
-              SET state=${UriRecommendationStates.INACTIVE}
-              WHERE (state=${UriRecommendationStates.ACTIVE} AND user_id=$userId AND master_score<$limitScore);""".asUpdate.first > 0
+              SET state=${UriRecommendationStates.INACTIVE},
+                  updated_at=$currentDateTime
+              WHERE state=${UriRecommendationStates.ACTIVE}
+                    AND user_id=$userId AND master_score<$limitScore
+                    AND created_at<$before;""".asUpdate.first > 0
```
yingjieMiao: I just noticed this: `(user, uriId)` pair is unique in the repo. Here we deactivate the item. But in `saveScoredSeedItems` method of `RecommendationGenerationCommander`,  we use 

`uriRecRepo.save(reco.copy(
            masterScore = computeMasterScore(item.uriScores),
            allScores = item.uriScores,
            attribution = item.attribution))`

which leaves the state the same as inactive, if we come to this `(user, uriId)` pair after it's been deactivated. This could be a problem. 

@stkem

kifi-backend/modules/curator/app/com/keepit/curator/model/UriRecommendationRepo.scala
lawzlo: I was thinking that. So currently updated_at will be updated when get users feedbacks. It seems not a reason to delay cleanup time?

stkem: Actually, here you don't need the date guard at all...

kifi-backend/modules/curator/app/com/keepit/curator/model/UriRecommendationRepo.scala
stkem: ...but here it should indeed be `updatedAt`, not `createdAt` (it's not perfect because of feedbacks, but better than created_at)
And yes, Yingjie is right, the saving in the `RecommendationGenerationCommander` will need to explicitly set the state back to active.

stkem:
please get this in asap (with the fixes based on comments). Whats currently merged will corrupt the data if deployed (after one hour, due to missing age guard).

kifi-backend/modules/curator/app/com/keepit/curator/model/UriRecommendationRepo.scala
lawzlo: @yingjieMiao good catch!

lawzlo: But why this pre-computation would be called after recos were generated 14 days later? I'm confused.

lawzlo: talked to @yingjieMiao and get clear.

kifi-backend/modules/curator/app/com/keepit/curator/model/UriRecommendationRepo.scala
lawzlo: hmm.. I'm not sure here, that seems two query based on different condition. I will remove it for checking in update code and think about edge case later.

PR #10487, opened 2014-08-26
Make cleanupLowMasterScoreRecos more functional

lawzlo:
@stkem @yingjieMiao


lawzlo:
tested, pushing to prod.

PR #10490, opened 2014-08-26
type ahead endpoint for contacts : email contacts & kifi-friends

ahsu1230:
@andrewconner @raymondng 

Type ahead for contacts
If there's no query, we grab a list of users & emails, ordered based on recent interactions
If there is a query, we search for kifi-friends & email contacts that fit the query


kifi-backend/modules/shoebox/app/com/keepit/commanders/UserInteractionCommander.scala
```diff
@@ -20,10 +20,11 @@ class UserInteractionCommander @Inject() (
       case Left(id) => Json.obj("user" -> id, "action" -> interaction.value)
       case Right(email) => Json.obj("email" -> email.address, "action" -> interaction.value)
     }
+    // append to head as most recent (will get the highest weight), remove from tail as least recent (lowest weight)
     val newInteractions = if (interactions.length + 1 > UserInteraction.maximumInteractions) {
-      interactions.drop(1) :+ newJson
+      newJson :: interactions.take(interactions.length - 1)
     } else {
-      interactions :+ newJson
+      newJson :: interactions
```
ahsu1230: woops messed up last time, we want to add recent interactions to head since the function gives a higher score on the items at the head of list

andrewconner: Nice catch. Also, statement could be:
`val newInteractions = newJson :: interactions.take(UserInteraction.maximumInteractions)`

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/TypeaheadController.scala
```diff
@@ -34,4 +34,10 @@ class TypeaheadController @Inject() (
     }
   }
 
+  def searchForContacts(query: Option[String], limit: Option[Int], pictureUrl: Boolean, dedupEmail: Boolean) = JsonAction.authenticatedAsync { request =>
```
andrewconner: We can probably streamline this and not have as many parameter options. Off the top of my head, `pictureUrl` and `dedupEmail` shouldn't change.

andrewconner:
Nice! Will review thoroughly after lunch.

andrewconner:
Let's also wire up eliza (messaging) for adding interactions, and then provide an endpoint for the browser extension.

ahsu1230:
so do you mean adding to eliza/controllers/ext or /site? both?

kifi-backend/modules/shoebox/conf/shoebox.routes
```diff
@@ -304,6 +304,7 @@ POST    /site/user/delighted/cancel @com.keepit.controllers.website.UserControll
 POST    /site/user/close            @com.keepit.controllers.website.UserController.closeAccount()
 
 GET     /site/user/connections/all/search    @com.keepit.controllers.website.TypeaheadController.searchWithInviteStatus(query:Option[String], limit:Option[Int], pictureUrl:Boolean ?= true, dedupEmail:Boolean ?= true)
+GET     /site/user/connections/contacts/search  @com.keepit.controllers.website.TypeaheadController.searchForContacts(query:Option[String], limit:Option[Int], pictureUrl:Boolean ?= true, dedupEmail:Boolean ?= true)
```
2is10: At one point we used “connections” specifically to mean non-Kifi social network connections (e.g. Facebook friends) and “contacts” to mean specifically imported address book entries. Having both here seems either contradictory or redundant, depending on how one chooses to interpret them. If this is a very general people search, maybe we should generalize the route accordingly to `/site/user/people/search`.

ahsu1230: I think this endpoint is for your kifi-friends & abook contacts. I think we're ignoring social connections for this kind of type-ahead. But, it's not as general as getting non-friend kifi users.
so maybe 'people' might be too general?

`/site/user/contacts/search` (?)

andrewconner:
Both.

kifi-backend/modules/shoebox/app/com/keepit/commanders/TypeaheadCommander.scala
```diff
@@ -337,6 +347,39 @@ class TypeaheadCommander @Inject() (
...
+          case Left(id) =>
+            val user = db.readOnlyMaster { implicit s =>
+              userRepo.get(id)
+            }
+            ConnectionStatus(user.fullName, interaction.score, if (pictureUrl) user.pictureName.map(_ + ".jpg") else None, s"fortytwo/${user.externalId}")
+          case Right(email) =>
+            ConnectionStatus("", interaction.score, None, emailId(email))
+        }
+      }
+      Future.successful(a)
```
andrewconner: Why wrap `a` in a `Future`?

kifi-backend/modules/shoebox/app/com/keepit/commanders/TypeaheadCommander.scala
```diff
@@ -344,3 +387,17 @@ class TypeaheadCommander @Inject() (
 }
 
 @json case class ConnectionWithInviteStatus(label: String, score: Int, networkType: String, image: Option[String], value: String, status: String, email: Option[String] = None, inviteLastSentAt: Option[DateTime] = None)
+
+@json case class ConnectionStatus(label: String, score: Double, image: Option[String], value: String)
```
andrewconner: What does "Status" mean in this?

ahsu1230: perhaps ConnectionFound instead?

kifi-backend/modules/shoebox/conf/shoebox.routes
2is10: Works for me! Thanks.

kifi-backend/modules/shoebox/app/com/keepit/commanders/TypeaheadCommander.scala
ahsu1230: so the aggregate method returns a Future ( of stuff)
and I figured in order to easily match working with the results from aggregate & the getInteractions(), this function should return a Future

kifi-backend/modules/shoebox/app/com/keepit/commanders/TypeaheadCommander.scala
ahsu1230: or maybe ContactsFound since we're not really talking about social users(connections) as @2is10 mentioned earlier

kifi-backend/modules/shoebox/app/com/keepit/commanders/TypeaheadCommander.scala
```diff
@@ -344,3 +387,17 @@ class TypeaheadCommander @Inject() (
 }
 
 @json case class ConnectionWithInviteStatus(label: String, score: Int, networkType: String, image: Option[String], value: String, status: String, email: Option[String] = None, inviteLastSentAt: Option[DateTime] = None)
+
+@json case class ContactFound(label: String, score: Double, image: Option[String], value: String)
```
andrewconner: It may be nice for the client to be able to know if it's a user or an email address. Not sure  `ConnectionWithInviteStatus` is the best template for us here. From the client perspective, they care about selecting and displaying these results. Having a "label" here mixes presentation with the data, and potentially introduces bugs such as:

For users, we have `BasicUser`, which is nice packaged form giving info needed to display a user. For email addresses, we primarily care about the name (if it exists) and email address of a contact.

It may be smart to return JSON that looks something like:

``` javascript
{"user": {/* basic user here*/}, "score": 4.323}
```

and

``` javascript
{"email": {"name": "Bob Fields", "address":"bfields@gmail.com"}, "score": 4.323}
```

What do you think?

2is10: Agreed that the data should be clearly labeled, and the format can differ for different kinds of results.

Do any clients need scores? If not, probably not worth including.

Also, in order to keep JSON small, we may not need to label what kind of result each one is. With that proposed format, clients would have to sniff for `.user` or `.email`. Sniffing for `.id` or `.address` would be just as easy.

kifi-backend/modules/shoebox/app/com/keepit/commanders/TypeaheadCommander.scala
```diff
@@ -344,3 +394,16 @@ class TypeaheadCommander @Inject() (
 }
 
 @json case class ConnectionWithInviteStatus(label: String, score: Int, networkType: String, image: Option[String], value: String, status: String, email: Option[String] = None, inviteLastSentAt: Option[DateTime] = None)
+@json case class ContactFound(name: String, eid: Option[ExternalId[User]], email: Option[EmailAddress], image: Option[String])
```
ahsu1230: @andrewconner 
@2is10 does this kind of fit better with what you were talking about?
So the Json would then give a name, and then maybe an externalId or email, so it should be easy to extract out those fields.
lists of `ContactFound` would already be sorted anyway, so there might not be a need for scores unless we want to display the values for experimental purposes. In which case, I can add anyway - no biggy.
What do you guys think?

andrewconner: I tend to err on the side of more types (to separate out use cases), but others may have another opinion. Technically, `name` is Optional, since for email addresses, we don't always have a name. If we just make the name the email address as a fallback, we can end up with the bug on the current invite page (that's what happened there).

Making it an option is better, but that creates a case class where all fields are optional. Seems like split classes (with a common trait) would be safer. @2is10?

2is10: I would tend to agree with @andrewconner and prefer using existing types instead of making a new one that tries to make different kinds of things look alike. Specifically, I was thinking of a JSON array, each of whose elements is either a `BasicUser` or a `BasicContact`.

2is10: If `BasicUser` is too verbose, then sure, we could introduce new types here. I suppose it would be desirable to send down a user’s name in the `name` field consistently, rather than sometimes splitting it into `firstName` and `lastName`.

P.S. Let’s name the `ExternalId` field `id`, as we do elsewhere. External IDs are the only kind of IDs clients ever know about.

2is10: Something like this?

``` scala
trait ContactSearchResult {}
case class UserResult(name: String, id: ExternalId[User], pic: URL) extends ContactSearchResult
case class EmailAddressResult(name: Option[String], email: EmailAddress) extends ContactSearchResult
```

ahsu1230: how would JSON serializing/deserializing work with ContactSearchResult?
would it just automatically figure out that it's a UserResult or EmailResult or would I have to implement a formatter?

2is10: We shouldn’t ever need to deserialize these on the server.

To serialize, maybe just add `@json` to each concrete type and use a match/case on each result? Looks like we already do a match/case on result type in `searchForContacts`.

kifi-backend/modules/eliza/app/com/keepit/eliza/commanders/MessagingCommander.scala
```diff
@@ -631,6 +631,10 @@ class MessagingCommander @Inject() (
       userRecipients <- userRecipientsFuture
       nonUserRecipients <- nonUserRecipientsFuture
     } yield {
+      userRecipients.map(id => shoebox.addInteraction(userId, Left(id), action = "message"))
```
andrewconner, line 634: After some thought, we could probably improve this by making it a batch add. Since most messages only have a few participants, not a super high priority, but probably a good things to do (make `addInteraction` take a `Seq[Either[Id[User], EmailAddress]]`)

andrewconner:
After playing with this a bit, it's really cool. The scoring seems to work as intended.

We might also want to add an interaction when you _receive_ a message from someone. I message you "Hey, know of any good restaurants around the city?" and you want to send me a message on another page — it would be nice if I was first (or close). We could potentially weight these lower than others. Thoughts?

PR #10492, opened 2014-08-26
Add url patterns for boosting and penalizing.

lawzlo:
Not a regex expert. Just verified several links per url regex through tool: http://regex101.com/#javascript


yingjieMiao:
escape of `.` ? your tests pass because the char  `.` matches the pattern `.`

lawzlo:
I thought if it doesn't need back slash for converting if we use """ in scala. But just verified, you are right. It needs converting. Adding back slash for all char need to convert.

yingjieMiao:
lgtm

stkem:
lgtm

lawzlo:
tested, pushing to prod.

PR #10494, opened 2014-08-26
add cache + better sql 

raymondng:
Improve performance of getHelpRankInfo via the following
1) Cache added
2) Better SQL -- the generated SQL does not look good (at all); the hand-written ones _should_ perform better

Note that in `KeepDiscoveryRepo` I use `PreparedStatement` directly and in `ReKeepRepo` (smaller dataset) I simply invoke an API one-by-one (for missing uriIds). The former is in all likelihood more performant but not sure if it makes a big difference with caching. Will deploy, monitor & iterate.

Also, as of now requests are directly sent to server (no client-side cache). I don't think it's necessary in this case but again will monitor and add if needed.

@ymatsuda (+ thanks for SQL consultation) 

@stkem @andrewconner FYI


kifi-backend/modules/heimdal/app/com/keepit/model/helprank/KeepDiscoveryRepo.scala
```diff
@@ -66,14 +72,38 @@ class KeepDiscoveryRepoImpl @Inject() (
...
-    uriIds.map { uriId => uriId -> getDiscoveryCountByURI(uriId, since) }.toMap
+  def getDiscoveryCountsByURIs(uriIds: Set[Id[NormalizedURI]], since: DateTime)(implicit session: RSession): Map[Id[NormalizedURI], Int] = {
+    if (uriIds.isEmpty) Map.empty
+    else {
+      val valueMap = uriDiscoveryCountCache.bulkGetOrElse(uriIds.map(UriDiscoveryCountKey(_)).toSet) { keys =>
+        val buf = collection.mutable.ArrayBuilder.make[(Id[NormalizedURI], Int)]
+        val missing = keys.map(_.uriId)
+        missing.grouped(20).foreach { ids =>
+          val params = Seq.fill(ids.size)("?").mkString(",")
+          val stmt = session.getPreparedStatement(s"select uri_id, count(distinct (hit_uuid)) from keep_click where uri_id in ($params) group by uri_id;")
```
andrewconner, line 86: Query interpolation doesn't work?

raymondng, line 86: I figure you'd ask :) No not really. I did look into it and my understanding (thus far) is that they basically lack support for this type of queries (variable # of bind variables for IN clause, etc. -- it's not well supported even for JDBC). I actually think we can make it work but it might end up even clunkier than plain JDBC.

raymondng:
Discussed with @ymatsuda and @andrewconner. Will merge/deploy/monitor + iterate.

stkem:
Sorry for the delayed review:
Afaik, both discovery count and rekeep count are always requested together. Why two separate caches?
Also, for these kinds of things it's great use the cache in the service client directly (we have many cases of this), to avoid http calls when the item is cached and allow the requesting service to cache things in memory which would be very beneficial here, since curator will look up the same uri many times.

raymondng:
It's more modular and easier to maintain -- we have various kinds of counts in HelpRank (well that's all we have for now) and I don't necessarily want to create a cache for each combination.

Interesting. Why does the curator lookup the same uri many times? 
Not objected to adding client cache -- just didn't see the need to do so. But can certainly add it.

yingjieMiao:
we does reference the cache directly on the client a lot. c.f. shoebox client.

stkem:
Curator will basically look up each uri's scores for each user (it's not actually all uris, there are a bunch of filters, but it's a lot).
I think client side caching is almost essential here, it could easily be a 10x speed improvement.

raymondng:
Sure let's talk to get on the same page. Adding a client-side cache is pretty simple.

stkem:
Well, basically the calls in the heimdal service client should not go through to heimdal if the data is in the cache. This does get a little tricky with the requests being batch, unfortunately, but we do have some examples dealing with that pattern in the code, e.g. `getUriSummaries` in the shoebox service client.

PR #10496, opened 2014-08-26
links for feed email to keep, send, and view; send UriRecommendationFeed...

joshm1:
...back from click actions

@stkem @2is10

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/email/EmailRecosController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/email/EmailRecosController.scala
@@ -4,7 +4,6 @@ import com.keepit.commanders.{ RawBookmarkRepresentation, KeepInterner }
 import com.keepit.common.controller.{ ActionAuthenticator, ShoeboxServiceController, WebsiteController }
 import com.keepit.common.db.ExternalId
 import com.keepit.common.db.slick.Database
-import com.keepit.controllers.ext.HandleDeepLinkRequests
 import com.keepit.curator.CuratorServiceClient
 import com.keepit.curator.model.RecommendationClientType
 import com.keepit.heimdal.HeimdalContextBuilderFactory
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/email/EmailRecosController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/email/EmailRecosController.scala
@@ -1,0 +1,65 @@
+package com.keepit.controllers.email
+
+import com.keepit.commanders.{ RawBookmarkRepresentation, KeepInterner }
+import com.keepit.common.controller.{ ActionAuthenticator, ShoeboxServiceController, WebsiteController }
+import com.keepit.common.db.ExternalId
+import com.keepit.common.db.slick.Database
+import com.keepit.controllers.ext.HandleDeepLinkRequests
+import com.keepit.curator.CuratorServiceClient
+import com.keepit.curator.model.RecommendationClientType
+import com.keepit.heimdal.HeimdalContextBuilderFactory
+import com.keepit.model._
+import play.api.libs.concurrent.Execution.Implicits.defaultContext
+import com.google.inject.Inject
+
+class EmailRecosController @Inject() (
+  db: Database,
+  actionAuthenticator: ActionAuthenticator,
+  keepInterner: KeepInterner,
+  uriRepo: NormalizedURIRepo,
+  curator: CuratorServiceClient,
+  heimdalContextBuilder: HeimdalContextBuilderFactory)
+    extends WebsiteController(actionAuthenticator) with ShoeboxServiceController with HandleDeepLinkRequests {
+
+  def viewReco(uriId: ExternalId[NormalizedURI]) = HtmlAction.authenticated { request =>
+    val source = KeepSource.emailReco
+    val hcb = heimdalContextBuilder.withRequestInfoAndSource(request, source)
+    implicit val context = hcb.build
+    val uri = db.readOnlyReplica { implicit s => uriRepo.get(uriId) }
+
+    curator.updateUriRecommendationFeedback(request.userId, uri.id.get, UriRecommendationFeedback(clicked = Some(true),
+      fromClient = Some(RecommendationClientType.Email)))
+
+    Found(uri.url)
+  }
+
+  def addKeep(uriId: ExternalId[NormalizedURI]) = HtmlAction.authenticated { request =>
+    val uri = db.readOnlyReplica { implicit s => uriRepo.get(uriId) }
+    val source = KeepSource.emailReco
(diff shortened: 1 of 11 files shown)


kifi-backend/modules/shoebox/conf/shoebox.routes
```diff
@@ -319,6 +319,14 @@ GET     /site/*path                 @com.keepit.controllers.website.HomeControll
 GET     /site/                      @com.keepit.controllers.website.HomeController.kifiSiteRedirect(path: String = "")
 
 ##########################################
+# Email-Link API
+##########################################
+
+GET     /e/1/recos/view             @com.keepit.controllers.email.EmailRecosController.viewReco(id: ExternalId[NormalizedURI])
+GET     /e/1/recos/keep              @com.keepit.controllers.email.EmailRecosController.addKeep(id: ExternalId[NormalizedURI])
```
2is10: `keepReco`?

kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtDeepLinkController.scala
```diff
@@ -16,6 +16,67 @@ import com.keepit.inject.FortyTwoConfig
 import java.util.NoSuchElementException
 import com.keepit.common.healthcheck.AirbrakeNotifier
 
+trait HandleDeepLinkRequests { this: ShoeboxServiceController =>
```
2is10: How about putting this in its own file in `com.keepit.controllers.email` next to your new controller?

Most of this controller belongs in that package too, but we can move it later.

kifi-backend/modules/shoebox/conf/shoebox.routes
```diff
@@ -319,6 +319,14 @@ GET     /site/*path                 @com.keepit.controllers.website.HomeControll
...
+GET     /e/1/recos/send             @com.keepit.controllers.email.EmailRecosController.sendReco(id: ExternalId[NormalizedURI])
```
2is10, line 327: Did you consider putting the ID in the path like `/e/1/recos/:id/send`?

kifi-backend/modules/shoebox/app/com/keepit/controllers/email/EmailRecosController.scala
```diff
@@ -0,0 +1,65 @@
...
+  def viewReco(uriId: ExternalId[NormalizedURI]) = HtmlAction.authenticated { request =>
+    val source = KeepSource.emailReco
+    val hcb = heimdalContextBuilder.withRequestInfoAndSource(request, source)
+    implicit val context = hcb.build
+    val uri = db.readOnlyReplica { implicit s => uriRepo.get(uriId) }
+
+    curator.updateUriRecommendationFeedback(request.userId, uri.id.get, UriRecommendationFeedback(clicked = Some(true),
+      fromClient = Some(RecommendationClientType.Email)))
+
+    Found(uri.url)
```
2is10, line 32: Instead of adding a second redirect here (in addition to the SendGrid one), could we link directly to the page and do our bookkeeping when SendGrid notifies us that the link was clicked?

joshm1, line 32: We need the request from email to go through us so we can redirect properly for mobile

kifi-backend/modules/shoebox/app/com/keepit/controllers/email/EmailRecosController.scala
```diff
@@ -0,0 +1,65 @@
...
+
+  def addKeep(uriId: ExternalId[NormalizedURI]) = HtmlAction.authenticated { request =>
+    val uri = db.readOnlyReplica { implicit s => uriRepo.get(uriId) }
+    val source = KeepSource.emailReco
+    val hcb = heimdalContextBuilder.withRequestInfoAndSource(request, source)
+    implicit val context = hcb.build
+
+    // todo(josh) check privacy w/ Product
+    val rawBookmark = RawBookmarkRepresentation(url = uri.url, isPrivate = false)
+    keepInterner.internRawBookmark(rawBookmark, request.userId, source, mutatePrivacy = false)
```
2is10, line 43: Another way to do this would be to load the Angular app and convey the user’s intent to it (external user ID and URL to keep). Then the app would execute the keep, if authorized, as if the user had clicked the “Keep” link in the app. This way the server would not be performing an action while handling this GET request. I think Twitter uses this approach with the Follow button in its emails, for example.

kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtDeepLinkController.scala
```diff
@@ -16,6 +16,67 @@ import com.keepit.inject.FortyTwoConfig
...
+      Redirect(uri.url)
+    } else if (isIphone) {
+      doHandleMobile(request, uri, locator)
+    } else {
+      log.info(s"sending unknown user to $uri")
+      Redirect(uri.url)
+    }
+  }
+
+  def doHandleMobile(request: Request[_], uri: NormalizedURI, locator: DeepLocator): SimpleResult = {
```
2is10: can this and `mobileCheck` be `protected`?

kifi-backend/modules/shoebox/conf/shoebox.routes
joshm1, line 327: I considered that. Decided not b/c I wasn't sure yet if the primary key for these URLs would  remain the external ID or if we should use an encoded value (which could hide and encapsulate any additional parameters we add in the future)

kifi-backend/modules/shoebox/app/com/keepit/controllers/email/EmailRecosController.scala
```diff
@@ -0,0 +1,65 @@
...
+  def sendReco(uriId: ExternalId[NormalizedURI]) = HtmlAction.authenticated { request =>
+    val source = KeepSource.emailReco
+    val hcb = heimdalContextBuilder.withRequestInfoAndSource(request, source)
+    implicit val context = hcb.build
+
+    val uri = db.readOnlyReplica { implicit s => uriRepo.get(uriId) }
+    curator.updateUriRecommendationFeedback(request.userId, uri.id.get, UriRecommendationFeedback(clicked = Some(true),
+      fromClient = Some(RecommendationClientType.Email)))
+
+    handleAuthenticatedDeepLink(request, uri, DeepLocator("#compose"), None)
```
2is10, line 61: clever code reuse

2is10:
nice work, good testing

kifi-backend/modules/shoebox/app/com/keepit/controllers/email/EmailRecosController.scala
joshm1, line 43: I like this idea. Thanks!

PR #10497, opened 2014-08-26
Recommend alpha

yipingliao:
Front-end recommendations code for alpha release.

Note that this is quite messy and not ready for "full" code review yet. There is quite a bit of refactoring that still needs to be done (recommendations and keeps are currently using separate services and separate HTML/directives that need to be consolidated), and there are good chunks of copy-and-paste code in this draft. I'm making this pull to get alpha testing started, but will refactor/clean up in the next two days.

@stkem @2is10 @andrewconner


kifi-backend/modules/shoebox/angular/src/common/filter.js
```diff
@@ -0,0 +1,8 @@
+'use strict';
+
+angular.module('kifi')
+  .filter('titlecase', function () {
+    return function(input) {
+      return input.slice(0, 1).toUpperCase() + input.slice(1).toLowerCase();
```
2is10, line 6: `.charAt(0)` might be slightly clearer for the first term.

If we want a robust `titlecase` (like [Rails’](http://api.rubyonrails.org/classes/ActiveSupport/Inflector.html#method-i-titlecase)), here are [some good candidates for Angular](https://gist.github.com/maruf-nc/5625869) (scroll down).

yipingliao, line 6: Thanks!

kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
```diff
@@ -49,6 +49,10 @@ angular.module('kifi')
           var loc = $location.path();
           return loc === path || util.startsWith(loc, path + '/');
         };
+
+        scope.inRecoExperiment = function () {
+          return profileService.me && profileService.me.experiments && profileService.me.experiments.indexOf('recos_beta') >= 0;
```
2is10, line 54: @andrewconner Are these precautions necessary? It’d be nice if we could guarantee that the app didn’t run until we knew who the user was. I know you’ve been working on embedding data in the HTML page. Could we start with the `BasicUser` + experiments?

kifi-backend/modules/shoebox/angular/src/recos/recos.js
```diff
@@ -5,38 +5,136 @@ angular.module('kifi')
...
+      $scope.loading = true;
+
+      recoService.getPopular().then(function (recos) {
+        $scope.loading = false;
+        $scope.recosState = 'hasPopularRecos';
+        $scope.recos = recos;
+      });
+    };
+
+    $scope.importBookmarks = function () {
```
2is10, line 72: Is this really needed here?

kifi-backend/modules/shoebox/angular/src/recos/recos.styl
```diff
@@ -0,0 +1,225 @@
+.kf-reco-recency-slider
+  margin: 20px auto
+  width: 75%
+
+  span
```
2is10, line 5: We should avoid using common tag names in descendent selectors to keep layouts fast. The new `span`, `p`, and `li` rules concern me; `label`, `button` and `h4` less so. Good rule of thumb is to give everything we style a class.

2is10:
Nice work. I’m excited to try it out.

The code duplication is a little concerning. It’s pretty easy to lose track of what was copied and from where after a pull request is merged.

kifi-backend/modules/shoebox/angular/src/recos/recoService.js
```diff
@@ -3,45 +3,229 @@
...
+          reason.name = formatTitleFromUrl(reason.url);
+        }
+      });
+
+      this.recoKeep = new Keep(rawReco.itemInfo);
+    }
 
     var clutchParams = {
-      cacheDuration: 20000
+      cacheDuration: 2  // Not using the cache so we can see the loading :) 20000
```
yipingliao: Change to 2 seconds.

kifi-backend/modules/shoebox/angular/src/common/filter.js
```diff
@@ -0,0 +1,8 @@
+'use strict';
+
+angular.module('kifi')
+  .filter('titlecase', function () {
```
andrewconner, line 4: Impressed!

kifi-backend/modules/shoebox/angular/src/recos/adhocService.js
```diff
@@ -0,0 +1,49 @@
...
+          keep.tagList = [];
+        });
+        util.replaceArrayInPlace(recos, res.data);
+        return recos;
+      });
+    }, clutchParams);
+
+    var api = {
+
+      fetchAdHocRecos: function (weights) {
```
andrewconner, line 29: :+1: Thanks for not increasing `keepService`'s size :)

andrewconner:
Overall very nice, great work. The code duplication is unfortunate. If you/we don't have an opportunity to polish `keepService` (and properly split responsibility) during this product cycle with recommendations, we'll definitely allocate time during libraries to do it right and not increase debt.

kifi-backend/modules/shoebox/angular/src/recos/recos.js
yipingliao, line 72: Maybe not. Will check on this.

kifi-backend/modules/shoebox/angular/src/recos/recos.js
```diff
@@ -5,38 +5,136 @@ angular.module('kifi')
...
+      if (recos.length > 0) {
+        $scope.loading = false;
+        $scope.recosState = 'hasRecos';
+        $scope.recos = recos;
+      } else {
+        $scope.recosState = 'noRecos';
+
+        // If the user has no recommendations, show some popular
+        // keeps/libraries as recommendations.
+        recoService.getPopular().then(function (recos) {
```
andrewconner, line 55: Primarily to @stkem: Can the backend just do this? Odd this would be a front-end concern. Or I'm late to the party?

stkem, line 55: You mean serve popular stuff implicitly form the same endpoint if there is nothing else? If that is what you mean, the reason it's separate is that it's displayed somewhat differently and we are talking about making it accessible independently.

PR #10499, opened 2014-08-26
add snapshot related columns

yingjieMiao:
@stkem


stkem:
lgtm

yingjieMiao:
note: with current code deployed, will trigger recompute for everyone: because the snapshots columns are defaulted to Null.

PR #10507, opened 2014-08-27
Update feedback API to use externalId

lawzlo:
@stkem @joshm1

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/SendgridCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/SendgridCommander.scala
@@ -1,13 +1,14 @@
 package com.keepit.commanders
 
 import com.google.inject.Inject
+import com.keepit.common.db.ExternalId
 import com.keepit.common.db.slick.Database
 import com.keepit.common.healthcheck.SystemAdminMailSender
 import com.keepit.common.logging.Logging
 import com.keepit.common.mail.{ ElectronicMail, ElectronicMailRepo, EmailAddress, SystemEmailAddress }
 import com.keepit.common.time.{ DEFAULT_DATE_TIME_ZONE, currentDateTime }
 import com.keepit.heimdal.{ HeimdalContextBuilderFactory, HeimdalServiceClient, NonUserEvent, NonUserEventTypes, UserEvent, UserEventTypes }
-import com.keepit.model.{ UriRecommendationFeedback, EmailOptOutRepo, NotificationCategory, UserEmailAddressRepo, UserEmailAddressStates }
+import com.keepit.model.{ NormalizedURI, UriRecommendationFeedback, EmailOptOutRepo, NotificationCategory, UserEmailAddressRepo, UserEmailAddressStates }
 import com.keepit.social.NonUserKinds
 import play.api.libs.concurrent.Execution.Implicits.defaultContext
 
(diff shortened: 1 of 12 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/SendgridCommander.scala
```diff
@@ -146,13 +147,12 @@ class SendgridCommander @Inject() (
...
     }
 
-    val userId = email.senderUserId.get
-    val keepUrl = event.url.get
-    val uriRecoFeedback = UriRecommendationFeedback(clicked = Some(true), kept = None)
-
-    recoCommander.updateUriRecommendationFeedback(userId, keepUrl, uriRecoFeedback).map { ok =>
-      if (!ok) log.warn(s"updateUriRecommendationFeedback($userId, $keepUrl, $uriRecoFeedback) returned false")
-    }
+    //    val userId = email.senderUserId.get
```
lawzlo: Since email will not use url anymore, @joshm1 can you update corresponding code to use externalId here?

kifi-backend/modules/shoebox/test/com/keepit/commanders/SendgridCommanderTest.scala
```diff
@@ -143,10 +143,11 @@ class SendgridCommanderTest extends Specification with ShoeboxTestInjector {
...
 
-          val (actualUserId, actualUriId, actualFeedback) = curator.updatedUriRecommendationFeedback.head
-          actualUriId === uri.id.get
-          actualUserId === user.id.get
-          actualFeedback === UriRecommendationFeedback(clicked = Some(true), kept = None)
+          //          val (actualUserId, actualUriId, actualFeedback) = curator.updatedUriRecommendationFeedback.head
+          //          actualUriId === uri.id.get
+          //          actualUserId === user.id.get
+          //          actualFeedback === UriRecommendationFeedback(clicked = Some(true), kept = None)
+          1 === 1
```
lawzlo: @joshm1 same here. :)

kifi-backend/modules/shoebox/app/com/keepit/model/NormalizedURIRepo.scala
```diff
@@ -208,6 +208,10 @@ class NormalizedURIRepoImpl @Inject() (
     (for (r <- rows if r.id.inSet(uriIds)) yield r.state).list.map(_.equals(NormalizedURIStates.SCRAPED))
   }
 
+  def getByExtId(extId: ExternalId[NormalizedURI])(implicit session: RSession): Option[NormalizedURI] = {
```
stkem: The repo already has a method that does exactly this (inherited from `ExternalIdColumnDbFunction`) called `getOpt`.

kifi-backend/modules/shoebox/app/com/keepit/commanders/SendgridCommander.scala
joshm1: we can get rid of this. we are no longer using sendgrid hooks to record these

kifi-backend/modules/shoebox/test/com/keepit/commanders/SendgridCommanderTest.scala
joshm1: you can delete this

kifi-backend/modules/shoebox/app/com/keepit/model/NormalizedURIRepo.scala
lawzlo: cool!

kifi-backend/modules/shoebox/app/com/keepit/commanders/SendgridCommander.scala
lawzlo: cool! will delete it.

stkem:
We'll need to sync with @yipingliao before deploying this so the changes on the site to match this can be made in short order.

PR #10510, opened 2014-08-27
[Draft] AugmentationCommander

leogrim:
@ymatsuda this has a few things missing: 
- actual tag retrieval 
- handling of published libraries 
- retrieval of titles and urls with fallback to ArticleIndex, need to think about this one if we want to return these first before the social information

But I'd be happy to know if you see any major flaw :)


ymatsuda:
lgtm!

PR #10512, opened 2014-08-27
shorter event name (for easier mixpanel consumption)

stkem:
user events already prepped `user_` to the event name, so we were ending
up with `user_recommendation_user_action` which is too long for some
mixpanel UI elements (and somewhat redundant)
@yingjieMiao


yingjieMiao:
thanks!

PR #10515, opened 2014-08-27
isPrivate removal, stage 1

andrewconner:
This is a bit tricky. This first PR accomplishes the following:
- Adds visibility (nullable) column to `bookmark` table
- Adds visibility field to `Keep` (non-Option). When it does not exist, it is calculated from isPrivate
- Removes isPrivate as a field from `Keep`
- Adds `isPrivate` as a method in `Keep`, calculated from the `visibility` field.
- Removes all assignments to `isPrivate`. Instead, assigns `visibility` and `libraryId`

Please review. I'm still testing.


kifi-backend/modules/common/app/com/keepit/model/Keep.scala
```diff
@@ -18,11 +18,11 @@ case class Keep(
     externalId: ExternalId[Keep] = ExternalId(),
     title: Option[String] = None,
     uriId: Id[NormalizedURI],
-    isPrimary: Boolean = true,
+    isPrimary: Boolean = true, // trick to let us have multiple inactive Keeps while keeping integrity constraints
     urlId: Id[URL],
     url: String, // denormalized for efficiency
     bookmarkPath: Option[String] = None,
-    isPrivate: Boolean = false, // This represents if the Keep is discoverable in search or the feed.
+    visibility: LibraryVisibility,
```
2is10: maybe add `// denormalized from this keep’s library`

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminLibraryController.scala
```diff
@@ -159,5 +159,39 @@ class AdminLibraryController @Inject() (
...
+              } else {
+                log.info(s"[lib-migrate:okay] It's good ${keep.id}, current: ${keep.visibility} + ${keep.libraryId}")
+              }
+            }
+            log.info(s"[lib-migrate] Updated batch of ${keepsAllFromOneUser.length} keeps for userId $userId")
+        }
+      }
+    }
+
+    Ok
```
ahsu1230, line 192: Nice!

andrewconner:
@ymatsuda @leogrim @stkem Mind going through this? It may be easiest to skip the tests while trying to understand the change (by LOC delta, most of this is getting existing tests to pass — we sure do create a ton of keeps for tests...).

I have a migrate endpoint here which will set the visibility and libraryId based on existing privacy for all keeps. Until I run that, this change possibly can still be destructive: when reading from the db, I use `is_privacy` to set visibility if it's not set. `Keep.isPrivate` is calculated based on visibility. On `save()`, `is_private` is set based on `Keep.visibility`. I'm reasonably certain it's correct, but would love another set of eyes.

PR #10519, opened 2014-08-27
recompute recos for user

lawzlo:
@stkem @yingjieMiao


kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
```diff
@@ -170,6 +172,15 @@ class RecommendationGenerationCommander @Inject() (
       ((seeds zip candidateURIs) filter (_._2) map (_._1), if (seeds.isEmpty) seq else seeds.map(_.seq).max)
     }
 
+  private def getRescoreSeedsForUser(userId: Id[User], state: UserRecommendationGenerationState) = {
+    for {
+      seeds <- seedCommander.getDiscoverableBySeqNumAndUser(state.seq, userId, 200)
+      recos <- db.readWriteAsync(implicit s => uriRecRepo.getByUserId(userId))
+    } yield {
+      ((recos zip seeds) filter (x => x._1.uriId == x._2.uriId) map (_._2), if (seeds.isEmpty) state.seq else seeds.map(_.seq).max)
```
yingjieMiao: why zip these two? are you sure they have same length and ordered in the same way?

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
```diff
@@ -322,8 +333,19 @@ class RecommendationGenerationCommander @Inject() (
...
+        }
       }
+
+      val state = getStateOfUser(userId)
+      val seedsAndSeqFuture = getRescoreSeedsForUser(userId, state)
+      val res: Future[Boolean] = getPrecomputationRecosResult(seedsAndSeqFuture, state, userId, Set.empty)
+      res.onFailure {
+        case t: Throwable => airbrake.notify("Failure during recommendation recomputation", t)
+      }
+      res.map(_ => ())
```
yingjieMiao: you can also use `res.onComplete` and match on success or fail state.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
lawzlo: oh right. I was only thinking they have different lengths which should be fine, it would zip to shorter length. But I didn't realize the order. Thanks for pointing out.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
```diff
@@ -170,6 +174,16 @@ class RecommendationGenerationCommander @Inject() (
       ((seeds zip candidateURIs) filter (_._2) map (_._1), if (seeds.isEmpty) seq else seeds.map(_.seq).max)
     }
 
+  private def getRescoreSeedsForUser(userId: Id[User], state: UserRecommendationGenerationState) = {
+    for {
+      seeds <- seedCommander.getDiscoverableBySeqNumAndUser(state.seq, userId, 200)
+      recos <- db.readWriteAsync(implicit s => uriRecRepo.getByUserId(userId))
+    } yield {
+      val recoUris = recos.map(_.uriId).toSet
+      (seeds filter (seed => recoUris.contains(seed.uriId)), if (seeds.isEmpty) state.seq else seeds.map(_.seq).max)
```
yingjieMiao: this is not quite efficient. You can use `recos` to create a `Map[Id, Boolean]`. Then the filtering step takes linear time.

lawzlo: I think this is linear time as well? check contains takes O(1), seeds filter takes O(n)?

yingjieMiao: maybe you are right. need to check manual.

andrewconner: [Set lookup is constant](http://docs.scala-lang.org/overviews/collections/performance-characteristics.html)

lawzlo: good chart :+1:

stkem: I don't think this will work correctly. If there are no matched recommendations in the first batch it will just do nothing, and even if there are matches it will stop after that.
You'll need to look up `rawSeedItem` form the db with the uriIds of the recos.

lawzlo: Sorry I confused about precomputation process would do per batch per ingestion batch. I'm updating it to get rawSeedItems. Thanks for pointing out.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
```diff
@@ -1,22 +1,6 @@
...
-  UriRecommendation,
-  UriScores,
-  PublicFeedRepo,
-  PublicSeedItem,
-  SeedItem,
-  PublicUriScores,
-  PublicFeed,
-  PublicScoredSeedItem
-}
+import com.keepit.curator.model.{ CuratorKeepInfoRepo, RawSeedItem, RawSeedItemRepo, UriRecommendationStates, ScoredSeedItemWithAttribution, RecoInfo, UserRecommendationGenerationStateRepo, UserRecommendationGenerationState, Keepers, UriRecommendationRepo, UriRecommendation, UriScores, PublicFeedRepo, PublicSeedItem, SeedItem, PublicUriScores, PublicFeed, PublicScoredSeedItem }
```
stkem: I think your IDE single-lined the imports again

andrewconner: 18 from the same package. `_` please :)

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
```diff
@@ -322,8 +350,19 @@ class RecommendationGenerationCommander @Inject() (
...
-        stateOpt.foreach { state => genStateRepo.save(state.copy(seq = SequenceNumber.ZERO)) }
+        stateOpt.foreach { state =>
+          genStateRepo.save(state.copy(seq = SequenceNumber.ZERO))
+        }
+      }
+
+      val state = getStateOfUser(userId)
+      val seedsAndSeqFuture = getRescoreSeedsForUser(userId, state)
+      val res: Future[Boolean] = getPrecomputationRecosResult(seedsAndSeqFuture, state, userId, Set.empty)
+      res.onComplete {
```
stkem: you could just use `onFailure` here and save the no-op success case.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
lawzlo: I think public feeds and recos computation should be separated into two classes. This class is too long. Will do refactoring later on.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
```diff
@@ -43,14 +29,16 @@ class RecommendationGenerationCommander @Inject() (
...
     uriRecRepo: UriRecommendationRepo,
     publicFeedRepo: PublicFeedRepo,
+    rawSeedItemRepo: RawSeedItemRepo,
     genStateRepo: UserRecommendationGenerationStateRepo,
     systemValueRepo: SystemValueRepo,
     experimentCommander: RemoteUserExperimentCommander) {
 
   val defaultScore = 0.0f
-
+  val MAX_INDIVIDUAL_KEEPERS_TO_CONSIDER = 100
```
stkem: The lookup logic (which uses this field) should go into the `SeedIngestionCommander`, so configuration and code are not duplicated.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
lawzlo: @yingjieMiao I think both ways are good. Just want to know which way is more convention so I can use it next time.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
```diff
@@ -170,6 +158,46 @@ class RecommendationGenerationCommander @Inject() (
       ((seeds zip candidateURIs) filter (_._2) map (_._1), if (seeds.isEmpty) seq else seeds.map(_.seq).max)
     }
 
+  private def recookSeedItem(userId: Id[User], rawItem: RawSeedItem, keepers: Keepers): SeedItem = SeedItem(
```
stkem: See comment above, this basically duplicates code that already exists in the `SeedIngestionCommander`.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
```diff
@@ -170,6 +158,46 @@ class RecommendationGenerationCommander @Inject() (
...
+    lastSeen = rawItem.lastSeen,
+    keepers = keepers,
+    discoverable = rawItem.discoverable
+  )
+
+  private def getRescoreSeedsForUser(userId: Id[User], state: UserRecommendationGenerationState) = {
+    val seedsFut = db.readOnlyReplicaAsync { implicit s =>
+      val recos = uriRecRepo.getByUserId(userId)
+      val rawSeedItems = recos.foldLeft(Seq.empty[RawSeedItem]) { (recosSeq, reco) =>
+        rawSeedItemRepo.getByUriIdAndUserId(reco.uriId, Some(userId)) match {
```
stkem: This might be making _a lot_ of repo calls. I think it's better to have one repo method that takes a set of uriIds.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
yingjieMiao: either works with me

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
```diff
@@ -170,6 +158,46 @@ class RecommendationGenerationCommander @Inject() (
...
+            Keepers.ReasonableNumber(keepInfoRepo.getKeepersByUriId(rawItem.uriId))
+          }
+        recookSeedItem(userId, rawItem, keepers)
+      }
+    }
+
+    for {
+      seeds <- seedsFut
+    } yield {
+      (seeds, if (seeds.isEmpty) state.seq else seeds.map(_.seq).max)
```
stkem: Re-scoring existing recommendation should not be updating the state, otherwise the full rescan which comes immediately after this will start in the wrong place.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
lawzlo: :+1: Will move it to `SeedIngestionCommander`

kifi-backend/modules/curator/app/com/keepit/curator/model/UriRecommendationRepo.scala
```diff
@@ -62,6 +64,10 @@ class UriRecommendationRepoImpl @Inject() (
   def table(tag: Tag) = new UriRecommendationTable(tag)
   initTable()
 
+  def getByUserId(userId: Id[User], uriRecommendationState: Option[State[UriRecommendation]])(implicit session: RSession): Seq[UriRecommendation] = {
```
stkem: I don't think this should filter by state. We want to consider everything previously recommended.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
lawzlo: Will create a batch check function in repo

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
```diff
@@ -322,8 +350,19 @@ class RecommendationGenerationCommander @Inject() (
...
         val stateOpt = genStateRepo.getByUserId(userId)
-        stateOpt.foreach { state => genStateRepo.save(state.copy(seq = SequenceNumber.ZERO)) }
+        stateOpt.foreach { state =>
+          genStateRepo.save(state.copy(seq = SequenceNumber.ZERO))
+        }
+      }
+
+      val state = getStateOfUser(userId)
+      val seedsAndSeqFuture = getRescoreSeedsForUser(userId, state)
+      val res: Future[Boolean] = getPrecomputationRecosResult(seedsAndSeqFuture, state, userId, Set.empty)
```
stkem: The batch you get here might be very large, which will cause the remote service calls in the scoring to time out. You'll need to break it up into blocks of, say, 200 or so.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
```diff
@@ -170,6 +158,46 @@ class RecommendationGenerationCommander @Inject() (
...
+    url = rawItem.url,
+    seq = SequenceNumber[SeedItem](rawItem.seq.value),
+    priorScore = rawItem.priorScore,
+    timesKept = rawItem.timesKept,
+    lastSeen = rawItem.lastSeen,
+    keepers = keepers,
+    discoverable = rawItem.discoverable
+  )
+
+  private def getRescoreSeedsForUser(userId: Id[User], state: UserRecommendationGenerationState) = {
```
stkem: Please try to always put an explicit return type on more complex methods like this. It helps a lot when reading the code.

lawzlo: cool. Will fix.

kifi-backend/modules/curator/app/com/keepit/curator/model/UriRecommendationRepo.scala
lawzlo: I was thinking maybe searching ACTIVE or ignore status don't affect much about result. Since if recos were deactivated before which are very unlikely to recommend so it's not necessary to do fast rescoring. Currently we have low score reaper to deactivate recos but we should assume those recos are deleted. Do I miss anything?

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
lawzlo: Cool. Make sense. Thanks.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
lawzlo: good point. fixing it.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
```diff
@@ -1,26 +1,10 @@
...
-  UriRecommendation,
-  UriScores,
-  PublicFeedRepo,
-  PublicSeedItem,
-  SeedItem,
-  PublicUriScores,
-  PublicFeed,
-  PublicScoredSeedItem
-}
+import com.keepit.curator.model.{ CuratorKeepInfoRepo, RawSeedItemRepo, UriRecommendationStates, ScoredSeedItemWithAttribution, RecoInfo, UserRecommendationGenerationStateRepo, UserRecommendationGenerationState, Keepers, UriRecommendationRepo, UriRecommendation, UriScores, PublicFeedRepo, PublicSeedItem, SeedItem, PublicUriScores, PublicFeed, PublicScoredSeedItem }
```
stkem: giant line still there

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
```diff
@@ -1,26 +1,10 @@
...
 import com.keepit.common.db.{ SequenceNumber, Id }
 import com.keepit.model.{ User, ExperimentType, UriRecommendationScores, SystemValueRepo, Name }
 import com.keepit.shoebox.ShoeboxServiceClient
-import com.keepit.common.concurrent.ReactiveLock
+import com.keepit.common.concurrent.{ ReactiveLock }
```
stkem: Why the `{ ... }`?

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
```diff
@@ -43,14 +27,15 @@ class RecommendationGenerationCommander @Inject() (
     attributionHelper: SeedAttributionHelper,
     db: Database,
     airbrake: AirbrakeNotifier,
+    keepInfoRepo: CuratorKeepInfoRepo,
```
stkem: unused dependency

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
```diff
@@ -170,6 +154,18 @@ class RecommendationGenerationCommander @Inject() (
       ((seeds zip candidateURIs) filter (_._2) map (_._1), if (seeds.isEmpty) seq else seeds.map(_.seq).max)
     }
 
+  private def getRescoreSeedsForUser(userId: Id[User]): Future[Seq[SeedItem]] = {
```
stkem, line 172: Why not have this entire method in the seed commander? That would avoid an dependency of the Reco Generation commander on the raw seed item repo, which it really should not know anything about.

lawzlo, line 172: because other scoring logics are in RecoGenCommander, so I kept this in same class. I can move it to seed commander.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
lawzlo: removing

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
lawzlo: I am going to break code into two classes, so I ignored this. But we can do it now.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
```diff
@@ -197,46 +193,49 @@ class RecommendationGenerationCommander @Inject() (
...
-          if (state.seq < newSeqNum) { precomputeRecommendationsForUser(userId, boostedKeepers) }
-          Future.successful(false)
-        } else {
-          val cleanedItems = seedItems.filter { seedItem => //discard super popular items and the users own keeps
-            seedItem.keepers match {
-              case Keepers.ReasonableNumber(users) => !users.contains(userId)
-              case _ => false
-            }
-          }
+  private def processSeeds(seedItems: Seq[SeedItem], newState: UserRecommendationGenerationState, userId: Id[User], boostedKeepers: Set[Id[User]]) = {
```
stkem: complex methods should get an explicit return type

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
```diff
@@ -197,46 +193,49 @@ class RecommendationGenerationCommander @Inject() (
...
+      processSeeds(seeds, newState, userId, boostedKeepers)
+    }
+  }
 
   private def precomputeRecommendationsForUser(userId: Id[User], boostedKeepers: Set[Id[User]]): Future[Unit] = recommendationGenerationLock.withLockFuture {
     getPerUserGenerationLock(userId).withLockFuture {
       val state = getStateOfUser(userId)
       val seedsAndSeqFuture = getCandidateSeedsForUser(userId, state)
-      val res: Future[Boolean] = getPrecomputationRecosResult(seedsAndSeqFuture, state, userId, boostedKeepers)
+      val res: Future[Boolean] = seedsAndSeqFuture.flatMap(seedsAndSeq => getPrecomputationRecosResult(seedsAndSeq._1, seedsAndSeq._2, state, userId, boostedKeepers))
```
stkem: `case (seeds, seq) => getPrecomputationRecosResult(seeds, seq, state, userId, boostedKeepers)`

lawzlo: cool.!

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
stkem, line 172: actually, not the entire method, but this commander should not have to deal with the raw seed item repo. Ideally the seed commander would have a method taking a `Seq[Id[NormalizedUri]]`.

lawzlo, line 172: I see. just move rawSeeds parts to seedCommander

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
stkem: Yes, let's break up the line now. Leaving code in a worse state hoping for a future refactor is not usually a good idea. Things happen and refactors never come, leading to the accumulation of problems.
As for breaking the class into two, which two parts would that be?

lawzlo: I'm breaking publicFeedGeneration to another commander. sounds good? Seems it doesn't have much overlap code with recoGenComm

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
```diff
@@ -322,8 +321,32 @@ class RecommendationGenerationCommander @Inject() (
...
+          genStateRepo.save(state.copy(seq = SequenceNumber.ZERO))
+        }
+      }
+
+      val state = getStateOfUser(userId)
+
+      val seedsFuture = getRescoreSeedsForUser(userId)
+      val res: Future[Seq[Boolean]] = seedsFuture.flatMap { seeds =>
+        val batches = seeds.grouped(200)
+        val allBatchesFutureResponses = batches.foldLeft(Future.successful(Seq.empty[Boolean])) { (allFutureResponses, batch) =>
```
stkem: This is mixing a synchronous loop like construct with an async body, which is hard to wrap your head around. I think it's correct, but I just spend nearly ten minutes convincing myself of that. At the least something like this will require thorough unit testing, however I think in this case `FutureHelpers.sequentialExec` (in `PimpedFuture.scala`) could make your life a lot easier.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
stkem: Yes, that is good idea. I would not do it in this PR though, to keep things separate.

kifi-backend/modules/curator/app/com/keepit/curator/model/UriRecommendationRepo.scala
stkem: We will eventually delete them when we run into size issue with the database, but until then we might as well use them. This computation is triggered when a users interests changed significantly, so it quite conceivable that a previous bad recommendation is now a good one.

lawzlo: Cool. I'm convinced. Will fix it. :)

stkem: :-)

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
lawzlo: cool. change to a simple way.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
lawzlo: fixing

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
```diff
@@ -170,6 +168,16 @@ class RecommendationGenerationCommander @Inject() (
       ((seeds zip candidateURIs) filter (_._2) map (_._1), if (seeds.isEmpty) seq else seeds.map(_.seq).max)
     }
 
+  private def getRescoreSeedsForUser(userId: Id[User]): Future[Seq[SeedItem]] = {
+    for {
+      recos <- db.readOnlyReplicaAsync(implicit s => uriRecRepo.getByUserId(userId))
+      rawSeeds <- seedCommander.getRawSeeds(userId, recos.map(_.uriId))
```
stkem: This call and the next could be combined. That way raw seed items are never exposed outside of the seed commander.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
```diff
@@ -322,8 +334,25 @@ class RecommendationGenerationCommander @Inject() (
...
+          genStateRepo.save(state.copy(seq = SequenceNumber.ZERO))
+        }
       }
+
+      val state = getStateOfUser(userId)
+
+      val seedsFuture = getRescoreSeedsForUser(userId)
+      val res: Future[Unit] = seedsFuture.flatMap { seeds =>
+        val batches = seeds.grouped(200)
+        FutureHelpers.sequentialExec(batches.toIterable)(batch => processSeeds(batch, state, userId, Set.empty))
```
stkem: This should still include the boostedKeepers just like the regular computation.

kifi-backend/modules/curator/test/com/keepit/curator/RecommendationGenerationCommanderTest.scala
```diff
@@ -167,5 +167,23 @@ class RecommendationGenerationCommanderTest extends Specification with CuratorTe
         feeds.size === 0
       }
     }
+
+    "re-compute recos for user" in {
```
stkem, line 171: I'm not sure how this tests that the re-computation works. You are putting some recos into the db and then just checking if they are there. You need to actually check if the recompilation logic does anything (this check should include checking that the sequence number for the user is reset to zero).

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
lawzlo: good call. fixing

kifi-backend/modules/curator/test/com/keepit/curator/RecommendationGenerationCommanderTest.scala
lawzlo, line 171: Yes, it checkes if recos saved before can be read again after reset user which for checking code path of reading from uriRecos repo and rawSeed repo and grab matched uris. I think this part is more important to be checked. set sequence number is straightforward here for me but sure I can check this also. I will change to run pre-computation and reset user, then I check seqnum and size of recos(but size will be 0 since we filter recos based on `shouldInclude`)

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
lawzlo: Sure. Adding it. Thanks.

stkem:
go for it

lawzlo:
wa ah! :+1:

lawzlo:
tested, pushing to prod.

PR #10520, opened 2014-08-27
check if feed is kept before sending heimdal event

yingjieMiao:
@stkem @raymondng

### DIFF
--- a/kifi-backend/modules/curator/app/com/keepit/curator/commanders/CuratorAnalytics.scala
+++ b/kifi-backend/modules/curator/app/com/keepit/curator/commanders/CuratorAnalytics.scala
@@ -37,7 +37,7 @@ class CuratorAnalytics @Inject() (
   private def toRecoUserActionContexts(userId: Id[User], uriId: Id[NormalizedURI], feedback: UriRecommendationFeedback): Seq[RecommendationUserActionContext] = {
     val modelOpt = db.readOnlyReplica { implicit s => uriRecoRepo.getByUriAndUserId(uriId, userId, None) }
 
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/curator/app/com/keepit/curator/commanders/CuratorAnalytics.scala
```diff
@@ -37,7 +37,7 @@ class CuratorAnalytics @Inject() (
   private def toRecoUserActionContexts(userId: Id[User], uriId: Id[NormalizedURI], feedback: UriRecommendationFeedback): Seq[RecommendationUserActionContext] = {
     val modelOpt = db.readOnlyReplica { implicit s => uriRecoRepo.getByUriAndUserId(uriId, userId, None) }
 
-    if (modelOpt.isDefined && modelOpt.get.delivered > 0) {
+    if (modelOpt.isDefined && modelOpt.get.delivered > 0 && !modelOpt.get.kept) {
```
raymondng: How about `modelOpt.exists(r => r.delivered > 0 && !r.kept))`?
Also how about `recOpt` instead of `modelOpt`?

stkem: I like `exists`. Also, this additional check should only happen if it's a keep event, it irrelevant otherwise.

stkem: I fact, we do want to allow votes and such on already kept items.

yingjieMiao:
@stkem  updated

stkem:
lgtm

PR #10522, opened 2014-08-27
Recommend alpha plus :)

yipingliao:
Mostly UI updates to the recos frontend. Please see commits for itemized changes. Also, updated calls to backend's feedback endpoints to use the new endpoints.

@stkem @2is10 cc @lawzlo


kifi-backend/modules/shoebox/angular/src/recos/recos.styl
```diff
@@ -1,18 +1,75 @@
...
+
+  // Reduce slider width to 50% when the screen is narrower but before the 
+  // three columns are reduced to two, and then increase it again when there are 
+  // only two columns on the page.
+  @media (min-width: 1024px) and (max-width: 1150px)
+    width: 50%
+
+  // Reduce slider width to 50% when the screen is really narrow.
+  @media (max-width: 880px)
+    width: 50%
```
2is10, line 21: We should avoid using `@media` queries for sizing and instead size relative to the parent container whenever possible. These `@media` query rules fail, sometimes badly, when the Kifi extension pane is open.

2is10:
:+1:

kifi-backend/modules/shoebox/angular/src/recos/recos.styl
yipingliao, line 21: Yikes! Thanks for letting me know.

PR #10525, opened 2014-08-28
allow redirect parameter to /recommendation after linking with FB

joshm1:
@stkem

### DIFF
--- a/kifi-backend/modules/curator/app/views/email/feedDigest.scala.html
+++ b/kifi-backend/modules/curator/app/views/email/feedDigest.scala.html
@@ -27,7 +27,7 @@
                                         <td align="left" valign="top"  style="font-family:Arial, sans-serif; font-size:16px;  text-align:left; color:#000000; line-height:21px; ">
                                             The digest is a recommended list based on your interests and recent Kifi activity.
                                             @if(!allRecos.isFacebookConnected) {
-                                            You can also <a href="https://www.kifi.com/link/facebook" target="_blank" style="line-height:16px; font-size:14px; color:#43a0f8; text-decoration:none;font-family:Arial, sans-serif;">Connect Facebook</a> to access recommendations from your friends’ best keeps.
+                                            You can also <a href="https://www.kifi.com/link/facebook?redirect=recommendation" target="_blank" style="color:#43a0f8; text-decoration:none;font-family:Arial, sans-serif;">Connect Facebook</a> to access recommendations from your friends’ best keeps.
                                             }
                                             <br/>
                                         </td>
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/app/com/keepit/controllers/core/AuthController.scala
```diff
@@ -51,8 +52,12 @@ class AuthController @Inject() (
     inviteCommander: InviteCommander,
     passwordResetRepo: PasswordResetRepo,
     heimdalServiceClient: HeimdalServiceClient,
+    config: FortyTwoConfig,
     implicit val secureSocialClientIds: SecureSocialClientIds) extends WebsiteController(actionAuthenticator) with ShoeboxServiceController with Logging {
 
+  // path is an Angular route
+  val LinkRedirects = Map("recommendation" -> s"${config.applicationBaseUrl}/recommendation")
```
2is10, line 59: Is it not possible to use the reverse router here instead of writing `/recommendation`? We try to use them whenever possible to make it easier to change routes.

kifi-backend/modules/curator/app/views/email/feedDigest.scala.html
```diff
@@ -60,7 +60,7 @@
                 <td align="center">
                     <table width="193" border="0" cellspacing="0" cellpadding="0" align="center">
                         <tr>
-                            <td><a href="https://www.kifi.com/kifeeeed" target="_blank"><img src="@assetUrl("digest/viewfulllist.jpg")" alt="View Full list" width="193" height="47"  style="display:block;" border="0" /></a></td>
+                            <td><a href="https://www.kifi.com/recommendation" target="_blank"><img src="@assetUrl("digest/viewfulllist.jpg")" alt="View Full list" width="193" height="47"  style="display:block;" border="0" /></a></td>
```
2is10, line 63: Can we change the `/recommendation` route to `/recommendations`?

kifi-backend/modules/shoebox/conf/shoebox.routes
```diff
@@ -38,7 +38,7 @@ POST    /auth/log-in                @com.keepit.controllers.core.AuthController.
 
 GET     /connect/:provider          @com.keepit.controllers.core.AuthController.popupBeforeLinkSocial(provider)
 GET     /connect/:provider/done     @com.keepit.controllers.core.AuthController.popupAfterLinkSocial(provider)
-GET     /link/:provider             @com.keepit.controllers.core.AuthController.link(provider)
+GET     /link/:provider             @com.keepit.controllers.core.AuthController.link(provider: String, redirect: Option[String] ?= None)
```
2is10, line 41: `then` or `dest` might also be nice choices for the parameter name. For example:
`/link/facebook?then=recommendations`

2is10:
:+1:

kifi-backend/modules/shoebox/app/com/keepit/controllers/core/AuthController.scala
joshm1, line 59: I'd prefer that too but this is an angular route without an identical play
route.

kifi-backend/modules/curator/app/views/email/feedDigest.scala.html
joshm1, line 63: I used the route previously defined in angular. I agree the plural sounds
better.

PR #10526, opened 2014-08-28
fixing primary emails being Null but users have verified emails

ahsu1230:
@leogrim @andrewconner 
will call this endpoint to fix this

@leogrim  As for `SocialUserImportEmail`, moving its functionality to UserCommander will require an internal endpoint since `SocialGraphPluginImpl` is using it & will require ShoeboxServiceClient to have it. Is it worth the move? =/


kifi-backend/modules/shoebox/app/com/keepit/admin/AdminUserController.scala
```diff
@@ -889,4 +889,19 @@ class AdminUserController @Inject() (
     NoContent
   }
 
+  def selectPrimaryEmails = AdminHtmlAction.authenticated { implicit request =>
+    val a = db.readWrite { implicit s =>
+      val usersNoPrimaries = userRepo.all.filter(_.primaryEmail.isEmpty)
+      usersNoPrimaries.map { u =>
+        val all = emailRepo.getAllByUser(u.id.get)
+        val selectedEmail = all.find(_.verified) match {
```
andrewconner: This is arbitrarily picking one. Is that what we want? (It may be, but having a picking function may make it clearer in the future — ie, pick the one with the lowest ID, or highest?)

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminUserController.scala
```diff
@@ -889,4 +889,19 @@ class AdminUserController @Inject() (
     NoContent
   }
 
+  def selectPrimaryEmails = AdminHtmlAction.authenticated { implicit request =>
+    val a = db.readWrite { implicit s =>
+      val usersNoPrimaries = userRepo.all.filter(_.primaryEmail.isEmpty)
```
andrewconner: We might want to do a sanity check here as well. Get all users, for those that have no primary email, do what you're doing. For those with one, check that a) the email exists / is active in emailRepo, b) the email belongs to the user, c) the email is verified. If not, remove it, and _attempt_ to select another (pick a verified one). Now seems like a good time to fix the data since we know there is/was a bug at one point.

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminUserController.scala
ahsu1230: how about most recently verified?

andrewconner: Usually, could work, but also may be an existing user who decided to connect a social network.

andrewconner: That may be best though.

andrewconner:
Great, thanks for taking initiative here to fix the data issue. For migrations, I tend to like to add a lot of logging, so you can be very clear what it did / is doing. (For more "serious" migrations, I also add a `readOnly` mode that lets me dry run it first.)

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminUserController.scala
```diff
@@ -889,4 +889,40 @@ class AdminUserController @Inject() (
...
+            }
+            userRepo.save(user.copy(primaryEmail = selectedEmail))
+
+          case Some(primary) => // sanity checking
+            // check email exists in repo
+            val allEmails = emailRepo.getAllByUser(user.id.get)
+            allEmails.find(e => e.address == primary) match {
+              case Some(e) if e.verified => user // good
+              case _ => // primary email does not exist in user's list of emails OR is not verified yet, select a new one
+                val selectedEmail = allEmails.filter(_.verified).sortBy(e => (e.lastVerificationSent, e.verifiedAt)).lastOption match {
```
ahsu1230, line 913: just like above (line 898),
we only look for verified emails and we first sort by lastVerificationSent then by verifiedAt which is are Option[DateTime]
So the order by lastVerificationSent looks like [None, None, Some(_), Some(_), Some(_)]
A None means email from social network (which we assume is verified)

So that means if we take the LAST of this list, we are taking the email (preference to real emails not social networks) that is most recently verified

joshm1:
Thanks!

PR #10530, opened 2014-08-28
small tweak

yingjieMiao:
@stkem

### DIFF
--- a/kifi-backend/modules/cortex/app/com/keepit/common/commanders/LDACommander.scala
+++ b/kifi-backend/modules/cortex/app/com/keepit/common/commanders/LDACommander.scala
@@ -236,7 +236,7 @@ class LDACommander @Inject() (
 
   def explainFeed(userId: Id[User], uris: Seq[Id[NormalizedURI]]): Seq[Seq[Id[Keep]]] = {
 
-    val MAX_KL_DIST = 0.6f // empirically this should be < 1.0
+    val MAX_KL_DIST = 0.5f // empirically this should be < 1.0
     val topK = 3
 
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/cortex/app/com/keepit/common/commanders/LDACommander.scala
```diff
@@ -248,7 +248,7 @@ class LDACommander @Inject() (
       scored.filter { _._2 < MAX_KL_DIST }.sortBy(_._2).take(topK).map { _._1 }
     }
 
-    val userFeats = db.readOnlyReplica { implicit s => uriTopicRepo.getUserRecentURIFeatures(userId, wordRep.version, min_num_words = 50, limit = 50) }
+    val userFeats = db.readOnlyReplica { implicit s => uriTopicRepo.getUserRecentURIFeatures(userId, wordRep.version, min_num_words = 50, limit = 100) }
```
stkem: Can we make it 200? Or do you think that will be too slow?

yingjieMiao: OK. will try. Note: with this kind of retrieval, we almost could have a recent interest score based on KL divergence.

yingjieMiao:
@stkem 

updated.

kifi-backend/modules/cortex/app/com/keepit/common/commanders/LDACommander.scala
stkem: True. However, this call is much less frequent than the interest scoring.

stkem:
:+1:

PR #10532, opened 2014-08-28
more safeguards on scraper for valid urls and more explicit about the re...

eishay:
...ason and source of the exception


eishay:
@raymondng

raymondng:
Sorry for the delay. Lgtm but many of the errors come from `java.net.URI` and not our URI parser. As I recall earlier we had explicitly put safeguards in place (in shoebox, using java.net.URI) -- so they should be caught prior to being sent to scraper. In fact, I think for a while that's the case -- we were catching the errors early. Need to check if those are still in place. There might be a regression.

PR #10538, opened 2014-08-28
HelpRank/Feed

raymondng:
Simple integration for reco clicked event. 

Local dev mode recommendation unfortunately still broken. Unit tests work but will need to deploy to prod to monitor & validate.

The current approach re-uses the search attribution mechanism: hit cache, etc. Once the discoveries are recorded, rekeep machinery kicks in.

reco keep event handling is not part of this PR. As discussed that might be a weak signal ... TBD

@stkem @yingjieMiao

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/curator/RecommendationUserAction.scala
+++ b/kifi-backend/modules/common/app/com/keepit/curator/RecommendationUserAction.scala
@@ -1,0 +1,13 @@
+package com.keepit.curator
+
+case class RecommendationUserAction(value: String)
+
+object RecommendationUserAction {
+  object Delivered extends RecommendationUserAction("delivered")
+  object Clicked extends RecommendationUserAction("clicked")
+  object Kept extends RecommendationUserAction("kept")
+  object MarkedGood extends RecommendationUserAction("marked_good")
+  object MarkedBad extends RecommendationUserAction("marked_bad")
+  object Trashed extends RecommendationUserAction("trashed")
+  object ImprovementSuggested extends RecommendationUserAction("improvement_suggested")
+}
--- a/kifi-backend/modules/heimdal/app/com/keepit/commander/EventTrackingCommander.scala
+++ b/kifi-backend/modules/heimdal/app/com/keepit/commander/EventTrackingCommander.scala
@@ -5,11 +5,10 @@ import com.keepit.common.db.{ ExternalId, Id }
 import com.keepit.common.db.slick.Database
 import com.keepit.common.healthcheck.AirbrakeNotifier
 import com.keepit.common.cache.TransactionalCaching
-import com.keepit.common.core._
 import com.keepit.common.logging.{ LogPrefix, Logging }
 import com.keepit.heimdal.{ HeimdalContext, KifiHitContext, SanitizedKifiHit, UserEvent }
 import com.keepit.model._
-import com.keepit.model.helprank.{ ReKeepRepo, KeepDiscoveryRepo }
+import com.keepit.model.helprank.{ KeepDiscoveryRepo }
 import com.keepit.search.ArticleSearchResult
 import com.keepit.shoebox.ShoeboxServiceClient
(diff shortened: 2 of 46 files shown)


kifi-backend/modules/common/app/com/keepit/curator/RecommendationUserAction.scala
```diff
@@ -0,0 +1,13 @@
...
+
+object RecommendationUserAction {
+  object Delivered extends RecommendationUserAction("delivered")
+  object Clicked extends RecommendationUserAction("clicked")
+  object Kept extends RecommendationUserAction("kept")
+  object MarkedGood extends RecommendationUserAction("marked_good")
+  object MarkedBad extends RecommendationUserAction("marked_bad")
+  object Trashed extends RecommendationUserAction("trashed")
+  object ImprovementSuggested extends RecommendationUserAction("improvement_suggested")
+}
```
yingjieMiao, line 13: mind removing the same class in `CuratorAnalytics`?

raymondng, line 13: Oh it didn't? I blame it on idea :) Will do.

kifi-backend/modules/heimdal/app/com/keepit/commander/EventTrackingCommander.scala
```diff
@@ -0,0 +1,93 @@
...
+        if (keeperIds.isEmpty) {
+          log.infoP("no keepers; skipped")
+          Future.successful[Unit]()
+        } else {
+          log.infoP(s"keepers=${keeperIds.mkString(",")}")
+          val hitUUID = ExternalId[ArticleSearchResult]()
+          val ts = userEvent.time
+          val numKeepers = keeperIds.size
+          implicit val dca = TransactionalCaching.Implicits.directCacheAccess
+          kifiHitCache.get(KifiHitKey(userId, uriId)) match {
```
yingjieMiao: I assume this cache is driven by 'click events', not just search 'hit' events?

yingjieMiao:
data flow lgtm. as discussed in person, this will record a recommendation keep event ONLY IF there is a click event shortly before (e.g. within 1 hour). (If there is no click event found in the cache, this recommendation keep is considered as a weak signal)

@stkem

kifi-backend/modules/heimdal/app/com/keepit/commander/EventTrackingCommander.scala
```diff
@@ -0,0 +1,93 @@
...
+
+object EventTrackingCommander {
+  implicit class HeimdalContextWrapper(underlying: HeimdalContext) {
+    // helpers/workarounds -- until we have a stronger contract for events
+    def getIdOpt[T](key: String): Option[Id[T]] = underlying.get[Double](key) map { id => Id[T](id.toLong) }
+    def getIdsOpt[T](key: String): Option[Seq[Id[T]]] = underlying.getSeq[Double](key).map { ids => ids.map(id => Id[T](id.toLong)) }
+  }
+}
+
+class EventTrackingCommander @Inject() (
```
stkem: It looks to me like this is help rank specific, which makes the name `EventTrackingCommander` a bit misleading. How about `HelpRankTrackingCommander`?

stkem:
made a small comment on naming. lgtm otherwise.
(I think the one hour window things is perfectly fine)

kifi-backend/modules/heimdal/app/com/keepit/commander/EventTrackingCommander.scala
raymondng: Sure thing.

kifi-backend/modules/heimdal/app/com/keepit/commander/EventTrackingCommander.scala
raymondng: yes

raymondng:
Thanks. Incorporated feedback.

PR #10543, opened 2014-08-28
Recommend front end: more UI changes

yipingliao:
Hi! This pull contains a bunch of UI changes and clean-up as pointed out by folks yesterday. I hope to get this deployed today to gather more UI feedback and get us into a ready state UI wise by Tuesday.

@2is10 cc @stkem


kifi-backend/modules/shoebox/angular/src/recos/recosView.tpl.html
```diff
@@ -34,17 +37,19 @@ <h1 class="kf-main-title">Recommendations</h1>
...
+      <footer class="kf-reco-bottom" ng-if="hasReason()" ng-class="{ 'kf-reco-multiple-reasons': hasTwoOrMoreReasons() }">
         <span ng-if="hasTwoOrMoreReasons()">
           <span class="kf-reco-icon kf-reco-reasons-right" ng-click="gotoNextReason()"></span>
           <span class="kf-reco-icon kf-reco-reasons-left" ng-click="gotoPrevReason()"></span>
         </span>
         <div class="kf-reco-reason" ng-repeat="reason in reasons" ng-if="$index === reasonIndex">
           <div ng-if="reason.kind === 'keep'">
-            Because you kept <a ng-href="{{reason.url}}" target="_blank" ng-attr-title="{{reason.name}}">{{reason.name}}</a>&nbsp;<time am-time-ago="reason.when"></time>
+            Because you kept <a ng-href="{{reason.url}}" target="_blank" ng-attr-title="{{reason.name}}">{{reason.name}}</a>
+            <span class="nowrap"><time am-time-ago="reason.when"></time></span>
```
2is10: Let’s please style the `<time>` elements instead of adding wrapper elements.

kifi-backend/modules/shoebox/angular/src/common/classes.styl
```diff
@@ -48,4 +48,7 @@
   height: 100%
 
 .clickable
-  cursor: pointer
\ No newline at end of file
+  cursor: pointer
+
+.nowrap
+  white-space: nowrap
```
2is10: This is what I call an “inline style” class. We avoid these and instead put style declarations in rules attached to semantic class names. `clickable` needs to die too.

kifi-backend/modules/shoebox/angular/src/keep/keep.styl
```diff
@@ -167,15 +167,18 @@ keepBorderRadius = 3px
...
 .kf-keep-time
 .kf-keep-read-time
+  font-family: "Georgia", Arial, sans-serif
   font-size: 13px
   color: #99a3b3
-  font-family: "Georgia", Arial, sans-serif
   white-space: nowrap
 
+.kf-keep-read-time
+  margin-left: -3px
```
2is10, line 180: curious about this. could the same effect be accomplished by instead decreasing the amount of whitespace declared elsewhere?

2is10:
:+1: nice work!

kifi-backend/modules/shoebox/angular/src/common/classes.styl
yipingliao: Removed "nowrap". Clickable appears in many places; I've made an Asana task to do this clean-up.

kifi-backend/modules/shoebox/angular/src/recos/recosView.tpl.html
yipingliao: Done!

yipingliao:
Slider is still ugly for about 40 pixels. Will definitely fix this before release, but am merging to push this out now.

2is10:
Sounds good. No rush on the `clickable` thing, and I’m happy to help with that. Just don’t want to see the “inline style” class pattern propagating further.

PR #10544, opened 2014-08-28
changed user's primaryEmails to be consistent with whatever email's we'v...

ahsu1230:
...e always been using

@andrewconner @leogrim @stkem

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/admin/AdminUserController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/admin/AdminUserController.scala
@@ -892,7 +892,7 @@ class AdminUserController @Inject() (
   def selectPrimaryEmails = AdminHtmlAction.authenticated { implicit request =>
     val a = db.readWrite { implicit s =>
 
-      def pickAnEmail(userId: Id[User]) = {
+      def pickAnEmail(user: User) = {
         /*
         val all = emailRepo.getAllByUser(user.id.get)
         all.filter(_.verified).sortBy(e => (e.lastVerificationSent, e.verifiedAt)).lastOption match {
@@ -900,13 +900,18 @@ class AdminUserController @Inject() (
           case _ => None // no verified emails to choose from, so still None
         }
         */
-        Some(emailRepo.getByUser(userId))
+        user.primaryEmail match {
+          case Some(e) => Some(e)
+          case None =>
+            val all = emailRepo.getAllByUser(user.id.get)
+            all.find(_.verified).map(_.address)
+        }
       }
 
       userRepo.all.map { user =>
         user.primaryEmail match {
           case None =>
-            val selectedEmail = pickAnEmail(user.id.get)
+            val selectedEmail = pickAnEmail(user)
             if (selectedEmail.nonEmpty) {
               log.info("Setting primary email address %s of userId %s".format(selectedEmail.get, user.id.get.toString))
             }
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/app/com/keepit/admin/AdminUserController.scala
```diff
@@ -891,14 +891,22 @@ class AdminUserController @Inject() (
...
+
+      def pickAnEmail(userId: Id[User]) = {
+        /*
+        val all = emailRepo.getAllByUser(user.id.get)
+        all.filter(_.verified).sortBy(e => (e.lastVerificationSent, e.verifiedAt)).lastOption match {
+          case Some(verifiedEmail) => Some(verifiedEmail.address)
+          case _ => None // no verified emails to choose from, so still None
+        }
+        */
+        Some(emailRepo.getByUser(userId))
```
ahsu1230: Here's the 

``` scala
def getByUser(userId: Id[User])(implicit session: RSession): EmailAddress = {
    val user = userRepo.get(userId)
    user.primaryEmail getOrElse {
      val all = getAllByUser(userId)
      all.find(_.verified) match {
        case Some(verifiedEmail) => verifiedEmail.address
        case None => all.headOption.getOrElse(throw new Exception(s"no emails for user $userId")).address
      }
    }
  }
```

ahsu1230:
checked by @stkem

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminUserController.scala
```diff
@@ -890,19 +890,34 @@ class AdminUserController @Inject() (
...
+    db.readWrite { implicit s =>
+
+      def pickAnEmail(user: User) = {
+        /*
+        val all = emailRepo.getAllByUser(user.id.get)
+        all.filter(_.verified).sortBy(e => (e.lastVerificationSent, e.verifiedAt)).lastOption match {
+          case Some(verifiedEmail) => Some(verifiedEmail.address)
+          case _ => None // no verified emails to choose from, so still None
+        }
+        */
```
joshm1, line 903: how come this is commented out?

andrewconner, line 903: We found out it runs faster when it's commented out.

kifi-backend/modules/shoebox/conf/shoebox.routes
```diff
@@ -476,7 +476,7 @@ GET     /admin/user/:id/mixpanel/reset    @com.keepit.controllers.admin.AdminUse
 POST    /admin/user/:id/uservalue   @com.keepit.controllers.admin.AdminUserController.userValue(id: Id[User])
 POST    /admin/user/:id/username   @com.keepit.controllers.admin.AdminUserController.setUsername(id: Id[User], username: String, overrideRestrictions: Boolean ?= false)
 DELETE  /admin/user/:id/username    @com.keepit.controllers.admin.AdminUserController.removeUsername(id: Id[User])
-POST    /admin/user/selectPrimaryEmails @com.keepit.controllers.admin.AdminUserController.selectPrimaryEmails()
```
andrewconner, line 479: For migrations, I do prefer `POST`s. Semantically correct, and less dangerous someone will accidentally call it.

stkem, line 479: I have to take the blame for this one. I suggested using a get to make it easier to call (and it's non destructive) :-/
You have a good point though.

PR #10549, opened 2014-08-29
Distributed Augmentation commander

leogrim:
@ymatsuda 
Comments similar to my previous pull request, that is:
- still need tag retrieval from the index
- the actual decoration logic producing AugmentedItem hasn't been thought out too much, waiting for better product input
- hope I got the distribution logic right :)


ymatsuda:
nice. lgtm!

PR #10550, opened 2014-08-29
new (dry) way to compose where clauses in repos

joshm1:
the motivation is that methods in repos contain a lot of duplicate code.  I'm introducing a syntax that is shorter, (subjectively) as readable , and could eliminate a lot of duplication between methods that have several identical filters.  i tested queries produced by slick and they are the same as before, so no database impact in this case.

the pipe `|>` syntax is a method @andrewconner created that essentially acts like the unix pipe and allows us to compose these `RepoQuery` methods without deeply nesting them like `h(g(f(x)))`

@eishay @stkem @ymatsuda @andrewconner @ahsu1230 @lawzlo @2is10 @raymondng @leogrim

### DIFF
--- a/kifi-backend/modules/curator/app/com/keepit/curator/model/UriRecommendationRepo.scala
+++ b/kifi-backend/modules/curator/app/com/keepit/curator/model/UriRecommendationRepo.scala
@@ -44,23 +44,23 @@ class UriRecommendationRepoImpl @Inject() (
 
   type RepoImpl = UriRecommendationTable
 
-  object QueryBuilder {
-    def recommendable(rows: RepoQuery = rows) =
+  private object QueryBuilder {
+    def recommendable(rows: RepoQuery) =
       for (row <- active(rows) if row.kept === false && row.trashed === false) yield row
 
-    def byUser(userId: Id[User])(rows: RepoQuery = rows) =
-      for (row <- rows if row.userId === userId) yield row
-
-    def byUri(uriId: Id[NormalizedURI])(rows: RepoQuery = rows) =
+    def byUri(uriId: Id[NormalizedURI])(rows: RepoQuery): RepoQuery =
       for (row <- rows if row.uriId === uriId) yield row
 
-    def active(rows: RepoQuery = rows) =
+    def byUser(userId: Id[User])(rows: RepoQuery): RepoQuery =
+      for (row <- rows if row.userId === userId) yield row
+
+    def active(rows: RepoQuery) =
       for (row <- rows if row.state === UriRecommendationStates.ACTIVE) yield row
 
-    def withThreshold(masterScoreThreshold: Float)(rows: RepoQuery = rows) =
+    def withThreshold(masterScoreThreshold: Float)(rows: RepoQuery) =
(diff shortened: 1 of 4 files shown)


andrewconner:
How bad is the generated SQL? :smile: 

edit: "i tested queries produced by slick" Oh wow, surprising. Nice.

yingjieMiao:
nice!

kifi-backend/modules/curator/app/com/keepit/curator/model/UriRecommendationRepo.scala
```diff
@@ -43,6 +44,27 @@ class UriRecommendationRepoImpl @Inject() (
 
   type RepoImpl = UriRecommendationTable
 
+  object QueryBuilder {
```
2is10: `private`?

kifi-backend/modules/curator/app/com/keepit/curator/model/UriRecommendationRepo.scala
```diff
@@ -63,22 +85,19 @@ class UriRecommendationRepoImpl @Inject() (
   initTable()
 
   def getByUriAndUserId(uriId: Id[NormalizedURI], userId: Id[User], excludeUriRecommendationState: Option[State[UriRecommendation]])(implicit session: RSession): Option[UriRecommendation] = {
-    (for (row <- rows if row.uriId === uriId && row.userId === userId && row.state =!= excludeUriRecommendationState.orNull) yield row).firstOption
+    (for (row <- byUser(userId)() |> byUri(uriId) if row.state =!= excludeUriRecommendationState.orNull) yield row).firstOption
```
2is10: For predicates whose parameter type leaves no ambiguity, we might consider shortening the method name to `with`:
`(for (row <- with(userId)() |> with(uriId) ...`

(Currently some of the methods begin with `by` and some with `with`. I think I prefer `with`.)

2is10:
very nice!

kifi-backend/modules/curator/app/com/keepit/curator/model/UriRecommendationRepo.scala
joshm1: great idea!

joshm1: actually can't use `with` as a method name, i'll use `by`

kifi-backend/modules/curator/app/com/keepit/curator/model/UriRecommendationRepo.scala
```diff
@@ -43,6 +44,27 @@ class UriRecommendationRepoImpl @Inject() (
 
   type RepoImpl = UriRecommendationTable
 
+  object QueryBuilder {
+    def recommendable(rows: RepoQuery = rows) =
```
andrewconner: I'm not sure these default arguments are a good pattern. It's very easy to forget to call it with the argument, and it "just works", even though in this case you may miss out on a previous clause. @stkem

PR #10554, opened 2014-08-29
wrapping the recommendations slider if short on horizontal space

2is10:
@yipingliao 

I’m not sure if we want this behavior, but I thought I’d try it out and show you what I meant in case.


stkem:
:+1:

PR #10556, opened 2014-08-29
search chunked response skeleton

ymatsuda:
- ExtSearchController.search2 uses Ok.chunked (am I using it correctly, @andrewconner ?)
- moved the decoration part SearchCommander.search2 to ExtSearchController
  @leogrim

### DIFF
--- a/kifi-backend/modules/search/app/com/keepit/controllers/ext/ExtSearchController.scala
+++ b/kifi-backend/modules/search/app/com/keepit/controllers/ext/ExtSearchController.scala
@@ -3,6 +3,7 @@ package com.keepit.controllers.ext
 import com.google.inject.Inject
 import com.keepit.common.concurrent.ExecutionContext._
 import com.keepit.common.controller.{ SearchServiceController, BrowserExtensionController, ActionAuthenticator }
+import com.keepit.common.db.Id
 import com.keepit.common.logging.Logging
 import com.keepit.model._
 import com.keepit.model.ExperimentType.ADMIN
@@ -12,12 +13,15 @@ import com.keepit.search.result.KifiSearchResult
 import com.keepit.search.result.ResultUtil
 import com.keepit.search.util.IdFilterCompressor
 import com.keepit.search.SearchCommander
+import com.keepit.shoebox.ShoeboxServiceClient
+import play.api.libs.concurrent.Execution.Implicits.defaultContext
 import play.api.libs.iteratee.Enumerator
 import play.api.libs.json._
 import scala.concurrent.Future
 
 class ExtSearchController @Inject() (
     actionAuthenticator: ActionAuthenticator,
+    shoeboxClient: ShoeboxServiceClient,
     searchCommander: SearchCommander) extends BrowserExtensionController(actionAuthenticator) with SearchServiceController with Logging {
 
   def search(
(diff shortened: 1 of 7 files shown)


kifi-backend/modules/search/app/com/keepit/search/result/KifiSearchResult.scala
```diff
@@ -52,6 +53,51 @@ object KifiSearchResult extends Logging {
...
+        log.error(s"can't serialize KifiPlainResult [uuid=$uuid][query=$query][hits=$hits][mayHaveMore=$mayHaveMoreHits][show=$show][experimentId=$experimentId][context=$context]", e)
+        throw e
+    }
+  }
+
+  def toKifiSearchHitsV2(hits: Seq[KifiShardHit]): JsArray = {
+    val v2Hits = hits.zipWithIndex.map {
+      case (h, index) =>
+        val json = JsObject(List(
+          "idx" -> JsNumber(index),
```
2is10: Probably okay to omit the hit index if it’s always the same as the index in the containing array. If it will sometimes start higher than zero (e.g. when someone clicks show more), then it’s probably a good idea to keep it.

kifi-backend/modules/search/app/com/keepit/controllers/ext/ExtSearchController.scala
```diff
@@ -59,9 +63,31 @@ class ExtSearchController @Inject() (
...
+    // TODO: augmentation
+    // decorationFutures += (plainResultFuture.map { ... })
+
+    // TODO: URI Summary
+    // if (withUriSummary) decorationFutures += (plainResultFuture.map { ... })
 
-    Ok(decoratedResult.json).withHeaders("Cache-Control" -> "private, max-age=10")
+    val decorationEnumerator = reactiveEnumerator(decorationFutures)
+
+    val resultEnumerator = plainResultEnumerator.andThen(decorationEnumerator).andThen(Enumerator.eof)
```
2is10: So that the entire, reconstituted response is valid JSON, we’ll probably want to precede `plainResultEnumerator` with a `[`, begin each decorationEnumerator with a `,` and precede `Enumerator.eof` with a `]`.

2is10: The exact format may require some experimentation with recent versions of Chrome & Firefox. It looks like there may still be [some issues](https://code.google.com/p/chromium/issues/detail?id=156023) with chunking non-HTML content types. Hopefully `application/json` will work fine.

2is10:
It’s exciting to see this coming together. Nice work!

andrewconner:
Yep, sounds be good. @2is10's suggestion about doing the `[` and `]` is good. You can do that with `Enumerator("[").andThen(...).andThen("]").andThen(Enumerator.eof)`

If we have problems chunking JSON, we can use jsonp as an immediate fix.

You can test this with:

``` bash
curl -N -vv http://your.endpoint/
```

(The `-N` flag tells curl not to buffer the chunks)

kifi-backend/modules/search/app/com/keepit/search/result/KifiSearchResult.scala
ymatsuda: The hit index is meant to be used to correlate a hit to a decoration information. An alternative I was thinking is to always keep a hit array (in 1st chunk) and a decoration info array (in subsequent chunks) parallel, i.e.  the same array index points to the same uri. I like the array index approach better. Which do you like? Do you have a better idea? @2is10

ymatsuda:
- removed hit index
- added "[" and "]"
- added UriSummary as a decoration chunk

@2is10 @andrewconner @leogrim

leogrim:
Cool! I'm excited to see how this behaves :)

PR #10561, opened 2014-08-29
binary search in sampling

yingjieMiao:
@stkem @leogrim 

To enable binary search, have to use array, which has to be invariant in `A`.


kifi-backend/modules/common/app/com/keepit/common/math/ProbabilityDensity.scala
```diff
@@ -4,19 +4,33 @@ import play.api.libs.functional.syntax._
...
+  def sample(x: Double): Option[A] = if (cumulative.size < 32) linearSample(x) else binarySample(x) // solve x = 8 * log2(x)
+
+  def linearSample(x: Double): Option[A] = cumulative.collectFirst { case (outcome, cdf) if x <= cdf => outcome }
+  def binarySample(x: Double): Option[A] = {
+    val n = cumulative.length
+    var low = 0
+    var up = n
+    var mid = 0
+    while (low < up) {
+      mid = low + (up - low) / 2
```
stkem, line 28: `<=> (up + low)/2` (one less operation :-)

stkem:
lgtm

kifi-backend/modules/shoebox/test/com/keepit/model/UserExperimentTest.scala
```diff
@@ -126,4 +126,20 @@ class UserExperimentTest extends Specification with ShoeboxTestInjector {
...
+      val density = ProbabilityDensity((0 until n - 1).map { i => (i, p) })
+      (0 until n - 1).map { i =>
+        val x = eps + i * p
+        density.linearSample(x).get === i
+        density.binarySample(x).get === i
+      }
+      density.linearSample(1.0 - eps) === None
+      density.binarySample(1.0 - eps) === None
+    }
+  }
```
leogrim, line 153: Cool! We may want to add a test with n not being a power of 2?

kifi-backend/modules/common/app/com/keepit/common/math/ProbabilityDensity.scala
yingjieMiao, line 28: one more operation to prevent overflow :-)

PR #10569, opened 2014-08-30
enabling thread page images for everyone on new extension

2is10:
@stkem


2is10:
@stkem This is kind of urgent because I just published the new extension for everyone. Won’t be available to Chrome users for at least 40 minutes, though.

PR #10572, opened 2014-09-02
Start refactoring recosService and keepService; get untag to work for recommendations.

yipingliao:
@stkem @andrewconner @2is10 

Hi! This is still a work-in-progress and rather messy, but I've started refactoring keepService and recoService and got untagging recommendations to work. :) I won't have a chance to finish all of the refactoring while traveling, so I am making the pull request now.

The idea is to have the recommendations and keeps controllers (RecosCtrl and KeepsCtrl) keep track of the recommendations and keeps being listed on the page. These controllers will make use of RecoService and (eventually) KeepService for data augmentation, and RecoNetworkService and KeepNetworkService for making the actual calls to the backend endpoints. I want to extend this pattern to the search results too. Please let me know what you think of this pattern.

Thanks!


kifi-backend/modules/shoebox/angular/src/recos/recoService.js
```diff
@@ -136,73 +133,13 @@ angular.module('kifi')
...
-
-      click: function (keep) {
-        $http.post(routeService.recoFeedback(keep.urlId), { clicked: true });
+      UserRecommendation: function (rawReco) {
+        return new Recommendation(rawReco, 'recommended');
       },
 
-      improve: function (keep, improvement) {
-        $http.post(routeService.recoFeedback(keep.urlId), { comment: improvement });
+      PopularRecommendation: function (rawReco) {
```
2is10: Let’s name these helper methods using lower camel case, so they don’t look like constructors. Maybe `createUserRecommendation` or `newUserRecommendation`?

kifi-backend/modules/shoebox/angular/src/recos/recoService.js
```diff
@@ -3,10 +3,7 @@
 angular.module('kifi')
 
 .factory('recoService', [
```
2is10: Looks like this service just decorates raw keeps as recommendations, primarily for display. Could we maybe call it `recoDecorator` or `recoAugmenter`? (Do service names have to end with `Service`? `recoDecoratorService`?)

andrewconner: They don't _have_ to, but it's convention.

kifi-backend/modules/shoebox/angular/src/keeps/keepNetworkService.js
```diff
@@ -0,0 +1,60 @@
...
+angular.module('kifi')
+
+.factory('keepNetworkService', [
+  '$analytics',
+  '$http',
+  '$location',
+  '$log',
+  '$q',
+  'env',
+  function ($analytics, $http, $location, $log, $q, env) {
```
2is10: If this service is going to be just for keep actions, maybe `keepActionService` or `keepAndUnkeepService`? When naming a type, I like to try to focus on the function it performs—the problem it can solve for callers. “network” answers “how?” more than “what?”.

2is10:
Nice work! I made a few suggestions. Andrew is better qualified to review.

I like the direction you’re going with this pull request. It will be good to have clear separation of responsibilities and naming of our services. It sounds like much of our keep service may soon be superseded by a library service that Andrew has in the works.

kifi-backend/modules/shoebox/angular/src/recos/recoService.js
yipingliao: Will do. Thanks!

kifi-backend/modules/shoebox/angular/src/recos/recoService.js
yipingliao: I'll try `recoDecoratorService`. `recoService` sounds generic and it's unclear what it's actually doing.

kifi-backend/modules/shoebox/angular/src/keeps/keepNetworkService.js
yipingliao: Sounds good. Will do.

PR #10573, opened 2014-09-02
refactor recoGenRecommander

lawzlo:
separate recoGen and publicFeedGen to make classes smaller and easy to maintain.

@stkem @andrewconner @yingjieMiao

### DIFF
--- a/kifi-backend/modules/curator/app/com/keepit/curator/commanders/PublicFeedGenerationCommander.scala
+++ b/kifi-backend/modules/curator/app/com/keepit/curator/commanders/PublicFeedGenerationCommander.scala
@@ -1,0 +1,119 @@
+package com.keepit.curator.commanders
+
+import com.google.inject.Inject
+import com.keepit.commanders.RemoteUserExperimentCommander
+import com.keepit.common.concurrent.ReactiveLock
+import com.keepit.common.db.{ Id, SequenceNumber }
+import com.keepit.common.db.slick.Database
+import com.keepit.curator.model.{
+  Keepers,
+  PublicUriScores,
+  PublicScoredSeedItem,
+  PublicFeed,
+  PublicFeedRepo,
+  PublicSeedItem
+}
+import com.keepit.model.{ ExperimentType, User, Name, SystemValueRepo }
+import com.keepit.shoebox.ShoeboxServiceClient
+
+import scala.concurrent.Future
+import play.api.libs.concurrent.Execution.Implicits.defaultContext
+
+class PublicFeedGenerationCommander @Inject() (
+    seedCommander: SeedIngestionCommander,
+    shoebox: ShoeboxServiceClient,
+    publicScoringHelper: PublicUriScoringHelper,
+    publicUriWeightingHelper: PublicUriWeightingHelper,
+    db: Database,
+    publicFeedRepo: PublicFeedRepo,
+    systemValueRepo: SystemValueRepo,
+    experimentCommander: RemoteUserExperimentCommander) {
+
+  val pubicFeedsGenerationLock = new ReactiveLock(1)
+  private val SEQ_NUM_NAME: Name[SequenceNumber[PublicSeedItem]] = Name("public_feeds_seq_num")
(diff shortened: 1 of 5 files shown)


lawzlo:
tested, pushing to prod.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/PublicFeedGenerationCommander.scala
```diff
@@ -0,0 +1,119 @@
...
+    db.readOnlyReplicaAsync { implicit session =>
+      publicFeedRepo.getByTopMasterScore(howManyMax)
+    }
+  }
+
+  private def computePublicMasterScore(scores: PublicUriScores): Float = {
+    (1 * scores.recencyScore +
+      1 * scores.popularityScore +
+      6 * scores.rekeepScore +
+      3 * scores.discoveryScore) *
```
stkem, line 47: Why did you remove the curation score?

lawzlo, line 47: oops. missed during refactoring. adding back.

stkem, line 47: Please be more careful with this. You shouldn't loose logic when moving a function from one class to another.

lawzlo, line 47: definitely. I will check again.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/PublicFeedGenerationCommander.scala
```diff
@@ -0,0 +1,119 @@
...
+  PublicFeedRepo,
+  PublicSeedItem
+}
+import com.keepit.model.{ ExperimentType, User, Name, SystemValueRepo }
+import com.keepit.shoebox.ShoeboxServiceClient
+
+import scala.concurrent.Future
+import play.api.libs.concurrent.Execution.Implicits.defaultContext
+
+class PublicFeedGenerationCommander @Inject() (
```
stkem, line 22: This needs to be a Singleton, otherwise the `pubicFeedsGenerationLock` does nothing, since there will be a new one created every time this class is injected somewhere.

lawzlo, line 22: make sense. Adding it. Thanks.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/PublicFeedGenerationCommander.scala
```diff
@@ -0,0 +1,119 @@
...
+    publicFeedRepo: PublicFeedRepo,
+    systemValueRepo: SystemValueRepo,
+    experimentCommander: RemoteUserExperimentCommander) {
+
+  val pubicFeedsGenerationLock = new ReactiveLock(1)
+  private val SEQ_NUM_NAME: Name[SequenceNumber[PublicSeedItem]] = Name("public_feeds_seq_num")
+  private def specialCurators(): Future[Seq[Id[User]]] =
+    experimentCommander.getUsersByExperiment(ExperimentType.SPECIAL_CURATOR).map(users => users.map(_.id.get).toSeq)
+
+  def getPublicFeeds(howManyMax: Int): Future[Seq[PublicFeed]] = {
```
lawzlo, line 37: I found this API never be used. Instead, recoRetrievalCommander calls feedRepo directly to get feeds. Is there anywhere else may use it? Or I can just remove it.

stkem, line 37: I think you can remove it

lawzlo, line 37: :+1:

PR #10574, opened 2014-09-02
Pushing `isPrivate` to the edges, ridding ourselves of bad KeepInfo input APIs

andrewconner:
Another tricky one, will be building some more tests to make sure this doesn't break current API.
- Pushing the concept of `isPrivate` away from `KeepInterner` and `KeepCommander`
- While in `KeepCommander`, I noticed some outdated and sloppy APIs, where methods were taking in `JsObject`s and `JsValue`s, making it very difficult to understand what the input format is. Additionally, I noticed that `KeepInfo` is being used as an input value class — this is confusing and potentially dangerous. So I switched all uses to `RawBookmarkRepresentation`, which is an input value class.
- All internal methods that deal with keeping now take a Library.
- Tests updated to reflect these changes.

This is a potentially messy PR / merge, so I'm working to get this in ASAP. Please review. @stkem @eishay @raymondng


andrewconner:
- @ahsu1230 @leogrim

andrewconner:
All tests pass, and checked common actions in dev mode from the site and extension. Mobile is more of a black box for me, continuing to check.

leogrim:
:+1:

PR #10578, opened 2014-09-02
fix a scraper error

ymatsuda:
I noticed there are many exceptions like this in a scraper's app.log.

17:21:36.684 [ForkJoinPool-4-worker-165] ERROR c.k.s.e.LinkProcessingExtractor - extraction failed: 
java.lang.IllegalArgumentException: Flat hash tables cannot contain null elements.
...
com.keepit.scraper.extractor.DefaultContentHandler.com$keepit$scraper$extractor$DefaultContentHandler$$startLink(DefaultExtractor.scala:127)
...

@leogrim @raymondng

### DIFF
--- a/kifi-backend/modules/scraper/app/com/keepit/scraper/extractor/DefaultExtractor.scala
+++ b/kifi-backend/modules/scraper/app/com/keepit/scraper/extractor/DefaultExtractor.scala
@@ -120,11 +120,12 @@ class DefaultContentHandler(maxContentChars: Int, handler: ContentHandler, metad
   }
 
   // link tag
-  private def startLink(uri: String, localName: String, qName: String, atts: Attributes) = {
+  private def startLink(uri: String, localName: String, qName: String, atts: Attributes): Unit = {
     super.startElement(uri, localName, qName, atts)
     val rel = atts.getValue("rel")
     val href = atts.getValue("href")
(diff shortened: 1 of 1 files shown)


raymondng:
Ah yes I remember identifying and fixing a similar error in YoutubeExtractor. :+1:

kifi-backend/modules/scraper/app/com/keepit/scraper/extractor/DefaultExtractor.scala
```diff
@@ -120,11 +120,12 @@ class DefaultContentHandler(maxContentChars: Int, handler: ContentHandler, metad
   }
 
   // link tag
-  private def startLink(uri: String, localName: String, qName: String, atts: Attributes) = {
+  private def startLink(uri: String, localName: String, qName: String, atts: Attributes): Unit = {
     super.startElement(uri, localName, qName, atts)
     val rel = atts.getValue("rel")
```
raymondng, line 125: As this happened more than once already, we can consider adding an implicit class (wraps `Attributes`) with methods (e.g. `getValueOpt`) that return Option[String] instead (and then replace the calls to `getValue`). This is assuming this code path is not performance bottleneck, of course. What do you think?

eishay:
:+1: good catch!

PR #10582, opened 2014-09-02
Add LibId in curatorKeepInfo

lawzlo:
@stkem


kifi-backend/modules/common/conf/evolutions/shoebox/238.sql
```diff
@@ -0,0 +1,10 @@
+# CURATOR
+
+# --- !Ups
+
+ALTER TABLE curator_keep_info
+  ADD COLUMN library_id BIGINT(20) NULL;
+
+insert into evolutions (name, description) values('238.sql', 'add library id column in curator_keep_info');
+
+# --- !Downs
```
stkem: newline

kifi-backend/modules/curator/app/com/keepit/curator/model/CuratorKeepInfo.scala
```diff
@@ -13,6 +13,7 @@ case class CuratorKeepInfo(
   uriId: Id[NormalizedURI],
   userId: Id[User],
   keepId: Id[Keep],
+  libraryId: Option[Id[Library]], // TODO: tan will change it to not option after re-ingesting
   state: State[CuratorKeepInfo], //library: Id[Library]
```
stkem: comment out `//library: Id[Library]` no needed any more

kifi-backend/modules/common/conf/evolutions/shoebox/238.sql
lawzlo: ya, I usually remember this add will add new line before pushing code to save a commit.

kifi-backend/modules/curator/app/com/keepit/curator/model/CuratorKeepInfo.scala
lawzlo: sure. will remove it.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/AllKeepsSeedIngestionHelper.scala
```diff
@@ -92,6 +92,7 @@ class AllKeepSeedIngestionHelper @Inject() (
       uriId = keep.uriId,
       userId = keep.userId,
       keepId = keep.id.get,
+      libraryId = keep.libraryId,
```
stkem, line 95: processUpdatedKeep also needs to write out the libraryId, since it can change.

stkem:
made a few comments, lgtm otherwise

kifi-backend/modules/curator/app/com/keepit/curator/commanders/AllKeepsSeedIngestionHelper.scala
lawzlo, line 95: cool. will add it.

lawzlo:
tested, pushing to prod.

PR #10585, opened 2014-09-02
stop using keep.isDiscoverable

ymatsuda:
@leogrim

### DIFF
--- a/kifi-backend/modules/search/app/com/keepit/search/graph/keep/KeepIndexable.scala
+++ b/kifi-backend/modules/search/app/com/keepit/search/graph/keep/KeepIndexable.scala
@@ -1,7 +1,7 @@
 package com.keepit.search.graph.keep
 
 import com.keepit.search.index.{ FieldDecoder, DefaultAnalyzer, Indexable }
-import com.keepit.model.{ NormalizedURI, Keep }
+import com.keepit.model.{ LibraryVisibility, NormalizedURI, Keep }
 import com.keepit.search.LangDetector
 import com.keepit.search.sharding.Shard
 import com.keepit.search.graph.library.LibraryFields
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/search/app/com/keepit/search/graph/keep/KeepIndexable.scala
```diff
@@ -37,9 +37,9 @@ case class KeepIndexable(keep: Keep, shard: Shard[NormalizedURI]) extends Indexa
 
     doc.add(buildKeywordField(libraryField, keep.libraryId.get.toString))
     doc.add(buildKeywordField(uriField, keep.uriId.toString))
-    if (keep.isDiscoverable) doc.add(buildKeywordField(uriDiscoverableField, keep.uriId.toString))
+    if (keep.visibility != LibraryVisibility.SECRET) doc.add(buildKeywordField(uriDiscoverableField, keep.uriId.toString))
     doc.add(buildKeywordField(userField, keep.userId.toString))
-    if (keep.isDiscoverable) doc.add(buildKeywordField(userDiscoverableField, keep.userId.toString))
+    if (keep.visibility != LibraryVisibility.SECRET) doc.add(buildKeywordField(userDiscoverableField, keep.userId.toString))
```
leogrim, line 43: So keeps with visibility != LibraryVisibility.SECRET include public libraries but also non-public discoverable libraries (e.g. main). So we have still have to load and check the library id and / or the contributor id to check whether the querying user has proper access to these non-public discoverable keeps (through library membership or friendship). We also need to search Secret libraries that the user belongs to in a different query.

I'm missing how these denormalizations pay out, could you help me understand this approach compared to the kind of naive branching done in AugmentationCommander using on the default libId / userId / visibility fields?

ymatsuda, line 43: uriDiscoverableField is used to determine if a hit in the "others" category is discoverable (= someone kept the uri publicly).

userDiscoverableField is used to find all discoverable (not necessarily searchable) keeps for each friend. We load URIs in to data buffer with score=0 at the beginning of a query exec. This allows us to correctly classify friend keeps.

In above both use cases, processing is done by scanning inverted lists with no additional lookups.

PR #10586, opened 2014-09-02
optimize

raymondng:
A follow-up. Bring `getReKeepCountsByURIs` up to par with `getDiscoveryCountsByURIs` based on prod results -- currently the latter beat the former by a wide margin in terms of performance. 

Now both utilizes the same approach (and almost same code except SQL) -- there might be room for a utility function here. Will look into it some more.

Also made `getHelpRankInfo` async -- issue both queries in parallel to further improve performance.

@ymatsuda @yingjieMiao @stkem @andrewconner


ymatsuda:
:+1:

PR #10587, opened 2014-09-02
removing /user/ from contact search path, adding deprecation notes

2is10:
@andrewconner


andrewconner:
Looks good. We can work on performance optimizations if you feel it's too slow.

PR #10589, opened 2014-09-02
shortening query parameter names as before

2is10:
@andrewconner ok with you?


2is10:
nevermind. just doesn’t matter.

PR #10590, opened 2014-09-02
view reco links from digest email work if unauthenticated

joshm1:
@stkem

### DIFF
--- a/kifi-backend/modules/common/test/com/keepit/common/controller/FakeActionAuthenticator.scala
+++ b/kifi-backend/modules/common/test/com/keepit/common/controller/FakeActionAuthenticator.scala
@@ -52,6 +52,7 @@ class FakeActionAuthenticator extends ActionAuthenticator with Logging {
 
   var fixedUser: Option[User] = None
   var fixedExperiments: Set[ExperimentType] = Set[ExperimentType]()
+  var isAuthenticated = true
 
   def setUser(user: User, experiments: Set[ExperimentType] = Set[ExperimentType]()): FakeActionAuthenticator = {
     fixedUser = Some(user)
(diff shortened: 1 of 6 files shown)


kifi-backend/modules/common/test/com/keepit/common/controller/FakeActionAuthenticator.scala
```diff
@@ -65,11 +66,18 @@ class FakeActionAuthenticator extends ActionAuthenticator with Logging {
     onSocialAuthenticated: SecuredRequest[T] => Future[SimpleResult],
     onUnauthenticated: Request[T] => Future[SimpleResult]): Action[T] = Action.async(bodyParser) { request =>
     try {
-      val user = fixedUser.getOrElse(User(id = Some(Id[User](1)), firstName = "Arthur", lastName = "Dent", username = None))
-      log.debug("running action with fake auth of user $user, request on path ${request.path} api: $apiClient")
-      val res = onAuthenticated(AuthenticatedRequest[T](FakeIdentity(user), user.id.get, user, request, fixedExperiments, None, None))
-      log.debug("executed action with res: $res")
-      res
+      if (isAuthenticated) {
```
joshm1, line 69: @andrewconner requiring `setUser` to be called broke about 7 controller tests (not sure how many actual test cases since they are skipped if 1 breaks).  not too bad, but I'd rather move on instead of refactor those for now.  This was an easy workaround.

andrewconner, line 69: Works for me

stkem:
:+1:

PR #10594, opened 2014-09-02
pagination for followers & collaborators

ahsu1230:
@andrewconner 

createFullLibraryInfo needs to have parameters (page#/pageSize for keeps and page#/pageSize for collaborators) so it can create library information with paginated keeps, collaborators & followers.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
@@ -37,16 +37,17 @@ class LibraryCommander @Inject() (
     implicit val publicIdConfig: PublicIdConfiguration,
     clock: Clock) extends Logging {
 
-  def createFullLibraryInfo(library: Library, pageKeeps: Int, pageSizeKeeps: Int, pageUsers: Int, pageSizeUsers: Int): FullLibraryInfo = {
+  def createFullLibraryInfo(library: Library): FullLibraryInfo = {
 
     val (lib, owner, collabs, follows, numCollabs, numFollows, keeps, keepCount) = db.readOnlyReplica { implicit s =>
       val owner = basicUserRepo.load(library.ownerId)
-      val collabs = libraryMembershipRepo.pageWithLibraryIdAndAccess(library.id.get, pageSizeUsers, pageUsers * pageSizeUsers, Set(LibraryAccess.READ_INSERT, LibraryAccess.READ_WRITE)).map(m => basicUserRepo.load(m.userId))
-      val follows = libraryMembershipRepo.pageWithLibraryIdAndAccess(library.id.get, pageSizeUsers, pageUsers * pageSizeUsers, Set(LibraryAccess.READ_ONLY)).map(m => basicUserRepo.load(m.userId))
+      val (firstFollows, firstCollabs) = libraryMembershipRepo.pageWithLibraryIdAndAccess(library.id.get, 10, 0, Set(LibraryAccess.READ_INSERT, LibraryAccess.READ_WRITE, LibraryAccess.READ_ONLY)).partition(u => u.access == LibraryAccess.READ_ONLY)
+      val collabs = firstCollabs.map(m => basicUserRepo.load(m.userId))
+      val follows = firstFollows.map(m => basicUserRepo.load(m.userId))
       val collabCount = libraryMembershipRepo.countWithLibraryIdAndAccess(library.id.get, Set(LibraryAccess.READ_WRITE, LibraryAccess.READ_INSERT))
       val followCount = libraryMembershipRepo.countWithLibraryIdAndAccess(library.id.get, Set(LibraryAccess.READ_ONLY))
 
-      val keeps = keepRepo.getByLibrary(library.id.get, pageSizeKeeps, pageKeeps * pageSizeKeeps).map(KeepInfo.fromKeep)
+      val keeps = keepRepo.getByLibrary(library.id.get, 10, 0).map(KeepInfo.fromKeep)
       val keepCount = keepRepo.getCountByLibrary(library.id.get)
       (library, owner, collabs, follows, collabCount, followCount, keeps, keepCount)
     }
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
@@ -37,21 +37,18 @@ class LibraryCommander @Inject() (
     implicit val publicIdConfig: PublicIdConfiguration,
     clock: Clock) extends Logging {
 
-  def createFullLibraryInfo(library: Library): FullLibraryInfo = {
-    val (lib, owner, collabs, follows, keeps, keepCount) = db.readOnlyReplica { implicit s =>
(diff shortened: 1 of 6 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -37,21 +37,18 @@ class LibraryCommander @Inject() (
     implicit val publicIdConfig: PublicIdConfiguration,
     clock: Clock) extends Logging {
 
-  def createFullLibraryInfo(library: Library): FullLibraryInfo = {
-    val (lib, owner, collabs, follows, keeps, keepCount) = db.readOnlyReplica { implicit s =>
+  def createFullLibraryInfo(library: Library, pageKeeps: Int, pageSizeKeeps: Int, pageUsers: Int, pageSizeUsers: Int): FullLibraryInfo = {
```
andrewconner: This may get confusing, where a "FullLibraryInfo" may have different amount of data. The keeps pattern previously was to include the first 10 keeps, and provide an endpoint to get more with `LibraryController.getKeeps`. This is probably the way to go for followers and collaborators as well.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -37,21 +37,18 @@ class LibraryCommander @Inject() (
...
-          case LibraryAccess.READ_ONLY => (c1, basicUserRepo.load(m.userId) :: f1)
-          case LibraryAccess.READ_INSERT => (basicUserRepo.load(m.userId) :: c1, f1)
-          case LibraryAccess.READ_WRITE => (basicUserRepo.load(m.userId) :: c1, f1)
-          case _ => (c1, f1)
-        }
-      }
-      val keeps = keepRepo.getByLibrary(library.id.get, 10, 0).map(KeepInfo.fromKeep)
+      val collabs = libraryMembershipRepo.pageWithLibraryIdAndAccess(library.id.get, pageSizeUsers, pageUsers * pageSizeUsers, Set(LibraryAccess.READ_INSERT, LibraryAccess.READ_WRITE)).map(m => basicUserRepo.load(m.userId))
+      val follows = libraryMembershipRepo.pageWithLibraryIdAndAccess(library.id.get, pageSizeUsers, pageUsers * pageSizeUsers, Set(LibraryAccess.READ_ONLY)).map(m => basicUserRepo.load(m.userId))
+      val collabCount = libraryMembershipRepo.countWithLibraryIdAndAccess(library.id.get, Set(LibraryAccess.READ_WRITE, LibraryAccess.READ_INSERT))
```
andrewconner, line 47: This is needed, great

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -37,21 +37,18 @@ class LibraryCommander @Inject() (
...
-      val (collabs, follows) = memberships.foldLeft(List.empty[BasicUser], List.empty[BasicUser]) {
-        case ((c1, f1), m) => m.access match {
-          case LibraryAccess.READ_ONLY => (c1, basicUserRepo.load(m.userId) :: f1)
-          case LibraryAccess.READ_INSERT => (basicUserRepo.load(m.userId) :: c1, f1)
-          case LibraryAccess.READ_WRITE => (basicUserRepo.load(m.userId) :: c1, f1)
-          case _ => (c1, f1)
-        }
-      }
-      val keeps = keepRepo.getByLibrary(library.id.get, 10, 0).map(KeepInfo.fromKeep)
+      val collabs = libraryMembershipRepo.pageWithLibraryIdAndAccess(library.id.get, pageSizeUsers, pageUsers * pageSizeUsers, Set(LibraryAccess.READ_INSERT, LibraryAccess.READ_WRITE)).map(m => basicUserRepo.load(m.userId))
```
andrewconner: This and the following lines can be combined into one query, and then split the results with `partition`

kifi-backend/modules/shoebox/test/com/keepit/controllers/website/LibraryControllerTest.scala
```diff
@@ -213,22 +213,24 @@ class LibraryControllerTest extends Specification with ShoeboxTestInjector {
...
+               |"url":"/ahsu/lib1",
+               |"ownerId":"${user1.externalId}",
+               |"collaborators":[],
+               |"followers":[],
+               |"keeps":[],
+               |"numKeeps":0,
+               |"numCollaborators":0,
+               |"numFollowers":0
+             |},
+             |"keepPageSize": 10,
```
andrewconner: It's probably easier to just deal with offsets and not pages, so we don't need to tell the client (and force the client to calculate page \* keepPageSize)

kifi-backend/modules/shoebox/test/com/keepit/controllers/website/LibraryControllerTest.scala
```diff
@@ -659,5 +658,83 @@ class LibraryControllerTest extends Specification with ShoeboxTestInjector {
...
+               |  "lastName":"Plumber",
+               |  "pictureName":"0.jpg"}
+               |  ],
+               |"followers":[
+               |  {"id":"${user3.externalId}",
+               |  "firstName":"Bowser",
+               |  "lastName":"Koopa",
+               |  "pictureName":"0.jpg"}
+               |  ],
+               |"numCollabs":1,
```
andrewconner, line 705: Mind keeping this format consistent with `FullLibraryInfo`? `numCollaborators` and `numFollowers`.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
```diff
@@ -236,6 +238,36 @@ class LibraryController @Inject() (
...
+
+        db.readOnlyReplica { implicit session =>
+          if (canView(request.userId, libraryRepo.get(libraryId), authToken)) {
+            val take = Math.min(count, 10)
+            val memberships = libraryMembershipRepo.pageWithLibraryIdAndAccess(libraryId, take, offset, Set(LibraryAccess.READ_WRITE, LibraryAccess.READ_INSERT, LibraryAccess.READ_ONLY))
+            val (f, c) = memberships.partition(_.access == LibraryAccess.READ_ONLY)
+            val followers = f.map(m => basicUserRepo.load(m.userId))
+            val collaborators = c.map(m => basicUserRepo.load(m.userId))
+
+            val numF = libraryMembershipRepo.countWithLibraryIdAndAccess(libraryId, Set(LibraryAccess.READ_ONLY))
```
andrewconner, line 256: Is this to update the client's store of the library in case it's been updated?

ahsu1230, line 256: it's to figure out how many total number of followers/collabs there are, right? Since using page() only gets you a few users

andrewconner, line 256: We can leave it, but at the point where clients are asking for more collaborator/follower, they know how many there are. If this is so that the clients update their internal count, we should probably think about a more holistic approach.

ahsu1230, line 256: ah I see. Gotcha.

PR #10597, opened 2014-09-02
using "pictureName" for clarity and consistency with BasicUser json

2is10:
ok @andrewconner?

Looks like the site isn’t using this yet.


andrewconner:
Good with me.

PR #10598, opened 2014-09-03
more code refactoring in curator

lawzlo:
@stkem @andrewconner @yingjieMiao @joshm1 

Make UriWeightingHelper take generic SeedItemType and return generic SeedItemWithMultiplierType.

I'm a newbie with Scala generic programming. Please help to review anywhere can do better than I did. So that we can make curator classes are more generic.


kifi-backend/modules/curator/app/com/keepit/curator/commanders/PublicUriScoringHelper.scala
```diff
@@ -5,13 +5,7 @@ import com.keepit.common.db.Id
...
 import com.keepit.common.time.currentDateTime
 import com.keepit.cortex.CortexServiceClient
-import com.keepit.curator.model.{
-  Keepers,
-  PublicSeedItemWithMultiplier,
-  CuratorKeepInfoRepo,
-  PublicUriScores,
-  PublicScoredSeedItem
-}
+import com.keepit.curator.model.{ SeedItemWithMultiplierType, Keepers, PublicSeedItemWithMultiplier, CuratorKeepInfoRepo, PublicUriScores, PublicScoredSeedItem }
```
lawzlo: I will fix format before checkin code.

stkem:
I'm not sure what the benefit of this is. It doesn't appear to save any code and having a Type names *Type is a little confusing (the data being passed is not actually a type, it's a value).
When you have a pull request open for a while (especially an exploratory one) and no-one reviews it, please bug someone in person instead of just merging. People miss emails sometimes.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/UriWeightingHelper.scala
```diff
@@ -77,5 +79,12 @@ class PublicUriWeightingHelper() {
       lastSeen = item.lastSeen,
       keepers = item.keepers)
   }
+
+  def apply[T <: SeedItemType](items: Seq[T]): Seq[SeedItemWithMultiplierType] = {
+    items match {
```
stkem, line 84: I'm afraid this pattern match does not work correctly. Due to type erasure the system can't distinguish different `Seq[T]` in the match, it will always go into the first case. The compiler generates a great deal of warnings about this.
Now, why that is just a warning and not a compiler error is beyond me.

stkem:
I'm sorry, but I'm going to revert this. It is causing incorrect behavior due to type erasure (see comments in the code). It should be a compiler error, but unfortunately is just a warning (which you should nonetheless pay attention to).

lawzlo:
I didn't know type erasure would potentially cause the problem, thanks for pointing out. The reason I doing this is for more refactoring later on. Currently PublicSeedItem, SeedItem, SeedItemWithMultiplier etc.  don't have inheritance relationships, and our each step of cook recos pipeline process is strongly coupled with each of those entity classes, I think the cook process should take more generic entity in and output the right entity out, so that we could easily add more features in the future(I believe we definitely will, like we currently add lib membership relationship for boosting and attribution). I will think about how to refactoring and maybe draw a picture to show you later, or there exists a reason that current codes is good enough and don't need refactoring now.

stkem:
I think the code structure is fairly descent right now. That being said, if you can come up with a good way to make it cleaner thats great! Unfortunately type erasure can make this kind of thing very difficult sometimes.
In this case I really blame the compiler. These kinds of erasure issues in match statements should really be a compile error. Do pay attention to the warnings in the future though.

lawzlo:
Thanks for tips. I will pay attention to the warnings.

PR #10601, opened 2014-09-03
stop using userInfo endpoint

raymondng:
Use openId endpoint instead, which is similar but (importantly) not deprecated

AFAIK everything works (using multiple accounts, testing locally) -- will deploy and monitor

As usual thanks @andrewconner for the tip about deprecation!

@andrewconner @eishay


andrewconner:
Great, thanks

PR #10606, opened 2014-09-03
Merging master into play23

andrewconner:
@raymondng I believe this resolves the conflicts. Also pushed it to Play 2.3.4.

Down to 13 deprecation warnings and the SecureSocial binary incompatibility.

```
[kifi-backend] $ common/compile
...
...
...
[error] bad symbolic reference. A signature in SecureSocial.class refers to type PlainResult
[error] in package play.api.mvc which is not available.
[error] It may be completely missing from the current classpath, or the version on
[error] the classpath might be incompatible with the version used when compiling SecureSocial.class.
[warn] 13 warnings found
[error] one error found
[error] (common/compile:compile) Compilation failed
[error] Total time: 115 s, completed Sep 2, 2014 9:04:03 PM
[kifi-backend] $
```


andrewconner:
Merging. Let's go over the migration tomorrow morning first thing.

raymondng:
:+1: Look forward to it.

PR #10607, opened 2014-09-03
reducing likelihood of image request timeouts from eliza

stkem:
@eishay @2is10 @yingjieMiao @andrewconner


eishay:
why is it slow from the first place? are we blocking on the remote site?

stkem:
No, we are not blocking afaik (don't know for sure about the parts inside scraper), but the calls to embedly and pagekeeper take a long time to come back in a lot of cases (I presume if they haven't seen the url before)

andrewconner:
The only insight I have is PagePeeker: it's a non-blocking but long running HTTP call. Up to 1 minute before it's "failed".

PR #10608, opened 2014-09-03
fixed bug - canView() didn't consider library invites before for read_on...

ahsu1230:
...ly access

@andrewconner


andrewconner:
Thanks!

PR #10610, opened 2014-09-03
changed library_invite repo to have nullable User_ID field so you can in...

ahsu1230:
...vite userId OR email address

@andrewconner


ahsu1230:
do not merge yet - still looking at evolutions dev mode

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryInviteRepo.scala
```diff
@@ -31,7 +31,7 @@ class LibraryInviteRepoImpl @Inject() (
 
   class LibraryInviteTable(tag: Tag) extends RepoTable[LibraryInvite](db, tag, "library_invite") {
     def libraryId = column[Id[Library]]("library_id", O.NotNull)
-    def ownerId = column[Id[User]]("owner_id", O.Nullable)
+    def ownerId = column[Id[User]]("owner_id", O.NotNull)
     def userId = column[Id[User]]("user_id", O.Nullable)
```
ahsu1230, line 35: recipient ID of an invite can be null (since recipient can also be an email) but ownerId of a library invite should not be null. only kifi users can invite to kifi libraries

PR #10611, opened 2014-09-03
fixing checkUrisDiscussed query

stkem:
@lawzlo @andrewconner


lawzlo:
:+1:  I was trying to fix it. You are so fast! Thanks!

PR #10612, opened 2014-09-03
Introducing Hashtag + tag indexing

leogrim:
@eishay @andrewconner I have typed the name of a Collection with the value class Hashtag. So that we can get into the habit of passing tags around as typed strings. See also my comments in KeepCommander. 

@ymatsuda tags are available in KeepIndexable, but I haven't spelled out the actual index fields yet (will do next as soon as we merge this). See also my comments in KeepCommander, I'm afraid they were some bugs in the current code for tag search.

I decided not to refactor the current code to inactivate tags together with inactivated keeps (we should) because the "undo" function of the website relies on them. I think we should wait until we get rid of the Collection model to clean this up. 

Thanks @andrewconner for your awesome job on KeepCommander, I agree with you that there's still some work left there :)

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
@@ -517,7 +517,10 @@ class KeepsCommander @Inject() (
       }
       val tagged = (activated ++ newK2C).toSet
       val taggingAt = currentDateTime
-      tagged.foreach(ktc => keptAnalytics.taggedPage(collection, keepsById(ktc.keepId), context, taggingAt))
+      tagged.foreach { ktc =>
+        keepRepo.save(keepsById(ktc.keepId)) // notify keep index
+        keptAnalytics.taggedPage(collection, keepsById(ktc.keepId), context, taggingAt)
+      }
       tagged
     }
     if (updateUriGraph) {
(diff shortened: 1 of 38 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
```diff
@@ -571,12 +576,14 @@ class KeepsCommander @Inject() (
       } {
         keepToCollectionRepo.remove(keepId = keep.id.get, collectionId = collection.id.get)
         collectionRepo.collectionChanged(collection.id.get)
+        keepRepo.save(keep) // notify keep index
         keptAnalytics.untaggedPage(collection, keep, context)
       }
     }
     searchClient.updateURIGraph()
   }
```
leogrim, line 585: The method below does not report to analytics (grrrrr)

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
```diff
@@ -603,6 +611,7 @@ class KeepsCommander @Inject() (
     }
   }
```
leogrim, line 613: The method below does not report to analytics either. It also did not bump up the sequence number of relevant collections, so I'm afraid the current collection index may not be consistent. @ymatsuda

ymatsuda:
:+1:

andrewconner:
Looks good.

PR #10613, opened 2014-09-03
Reco frontend fixes

yipingliao:
A few more frontend fixes for recommendations:
- Get untag to work for private keeps (forgot to fix this path when I fixed the public keeps)
- Renamed the tracking actions for recommendations so that it's clear that they are only for tracking
- When a user removes all the recommendations, reload a new set of recommendations
- Allow users to undo the removal of a recommendation

@stkem cc @2is10

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/recos/recoActionService.js
+++ b/kifi-backend/modules/shoebox/angular/src/recos/recoActionService.js
@@ -69,11 +69,11 @@ angular.module('kifi')
         $http.post(routeService.recoFeedback(keep.urlId), { vote: vote });
       },
 
-      keep: function (keep) {
+      trackKeep: function (keep) {
         $http.post(routeService.recoFeedback(keep.urlId), { kept: true });
       },
 
-      click: function (keep) {
+      trackClick: function (keep) {
         $http.post(routeService.recoFeedback(keep.urlId), { clicked: true });
       },
 
--- a/kifi-backend/modules/shoebox/angular/src/recos/recos.js
+++ b/kifi-backend/modules/shoebox/angular/src/recos/recos.js
@@ -102,10 +102,6 @@ angular.module('kifi')
       $scope.initialCardClosed = true;
     };
 
-    $scope.$watch(tagService.getTotalKeepCount, function (val) {
-      $scope.numKeptItems = val;
-    });
-
     recoActionService.get().then(function (rawRecos) {
       if (rawRecos.length > 0) {
         rawRecos.forEach(function (rawReco) {
--- a/kifi-backend/modules/shoebox/angular/src/recos/recos.js
+++ b/kifi-backend/modules/shoebox/angular/src/recos/recos.js
@@ -36,10 +36,10 @@ angular.module('kifi')
       _.pull($scope.recos, reco);
     };
 
-    $scope.keepReco = function (reco) {
+    $scope.keepReco = function (reco, isPrivate) {
       recoActionService.keep(reco.recoKeep);
-
-      keepActionService.keepPublic(reco.recoKeep).then(function (keptKeep) {
+      
+      keepActionService.keepOne(reco.recoKeep, isPrivate).then(function (keptKeep) {
         reco.recoKeep.id = keptKeep.id;
         reco.recoKeep.isPrivate = keptKeep.isPrivate;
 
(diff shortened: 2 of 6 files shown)


kifi-backend/modules/shoebox/angular/src/recos/recos.js
```diff
@@ -33,13 +44,28 @@ angular.module('kifi')
...
       recoActionService.trash(reco.recoKeep);
-      _.pull($scope.recos, reco);
+      
+      var trashedRecoIndex = _.findIndex($scope.recos, reco);
+      var trashedReco = $scope.recos.splice(trashedRecoIndex, 1)[0];
+
+      // If the user has trashed all the recommendations, reload a new set of
+      // recommendations.
+      if ($scope.recos.length === 0) {
+        $scope.getMore();
```
yipingliao, line 54: @stkem Here I am grabbing a new set of recommendations with `more` set to `true`. Let me know if we should be pinging the endpoint with `more` set to `false` instead.

kifi-backend/modules/shoebox/angular/src/recos/recos.js
```diff
@@ -33,13 +44,28 @@ angular.module('kifi')
...
+      } 
+      // If the user has recommendations remaining on the page, present her
+      // with the option to undo the trashing of the recommendation.
+      else {
+        undoService.add('Recommendation removed.', function () {
```
yipingliao: The undo option is not presented if the user has trashed the very last recommendation, because we would be loading a new set of recommendations already. What do you think of this user experience?

kifi-backend/modules/shoebox/angular/src/recos/recoActionService.js
```diff
@@ -69,11 +69,11 @@ angular.module('kifi')
         $http.post(routeService.recoFeedback(keep.urlId), { vote: vote });
       },
 
-      keep: function (keep) {
+      trackKeep: function (keep) {
         $http.post(routeService.recoFeedback(keep.urlId), { kept: true });
```
2is10, line 73: much better name, thanks

kifi-backend/modules/shoebox/angular/src/recos/recos.js
2is10: Consistency is important. If trashing a recommendation can be undone, then that should be the case every time one is trashed. I’ll take a look at the UI with you and see if we can come up with something.

kifi-backend/modules/shoebox/angular/src/recos/recos.js
stkem, line 54: It should not make a difference in this case, so we can leave it as `true` just to keep it uniform.

stkem:
lgtm

yipingliao:
Added a few more commits to this pull to address code review comments from @2is10  and new tasks from @stkem 
- Users can undo removal of recommendations even when the removed recommendation is the last one. The user experience is that the previously removed recommendation will reappear on the page in the top spot above the newly loaded recommendations.
- The dropdown menu on recommendations will hide when the user clicks anywhere outside the menu.
- Add animation on opacity on undoing removed recommendations so that the animation for removing and un-removing are symmetrical. @2is10 I will make an Asana task for adding the nice height animation.
- Change copies on initial state messaging and topic attribution messaging.
- Do not show recommendation dropdown menu on popular keeps since the menu actions are irrelevant for them.

2is10:
:+1:

Does undoing the last removal introduce a race condition? If the undo completes before the reload, could the new recommendations clobber the restored recommendation? Seems like they should either be appended beneath or simply discarded and never drawn in this case.

yipingliao:
Here is how I convinced myself that there is no race condition. Please let me know if I missed something.

When the last recommendation gets trashed, the first thing that happens is that `getMore` gets called, which wipes out `$scope.recos` (resets it to `[]`), and then initiates a call to the backend for more recommendations. While that's waiting, if the undo completes, then `$scope.recos` will now contain a single item (the previously removed recommendation). When the backend call completes, the new recommendations are added to `$scope.recos` via array `push`, so the previously removed recommendation remains intact as the first item in the array. In the event that the undo completes after the backend call completes, the previously removed recommendation will be spliced into the first position in `$scope.recos`, which has now been filled with the new recommendations. Since JavaScript runs to completion, I am not worried that in the middle of pushing the new recommendations on to `$scope.recos`, the undo complete will come in and splice items in that array. So either way, the behavior is that there will be 11 recommendation items, the first one being the one that was previously removed.

2is10:
Yes, `.push` and `.splice` will do the trick. Thanks for explaining!

PR #10615, opened 2014-09-03
add opt-out category for digest emails

joshm1:
@stkem


kifi-backend/modules/shoebox/app/views/website/optOutEmails.scala.html
```diff
@@ -23,6 +24,9 @@ <h2 class="page-title">Opt-out of Kifi Emails</h2>
       <label>
         <input type="checkbox" name="invite" value="true" @if(invite){checked}> Invitations to join Kifi from friends
       </label>
+      <label>
+        <input type="checkbox" name="invite" value="true" @if(digest){checked}> Kifi Digest recommendation emails
```
stkem: `name="invite"` intended?

stkem:
lgtm

PR #10618, opened 2014-09-03
a few curator fixes/adjustments

stkem:
Most notable:
-The new discussion filter was filtering all pages _without_ a discussion, instead of the other way around
-`resetUser` was using `withLock` instead of `withLockFuture` causing the lock to so nothing (and a test to become a randomly failing)
@lawzlo @yingjieMiao @andrewconner

### DIFF
--- a/kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
+++ b/kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
@@ -10,7 +10,6 @@ import com.keepit.curator.model.{
   UriRecommendationRepo,
   UriRecommendation,
   UriScores,
-  PublicFeedRepo,
   SeedItem
 }
 import com.keepit.common.db.{ SequenceNumber, Id }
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
```diff
@@ -141,7 +138,7 @@ class RecommendationGenerationCommander @Inject() (
     } yield {
       val candidateSeeds = (seeds zip candidateURIs) filter (_._2) map (_._1)
       eliza.checkUrisDiscussed(userId, candidateSeeds.map(_.uriId)).map { checkThreads =>
-        val candidates = (candidateSeeds zip checkThreads).collect { case (cand, hasChat) if hasChat => cand }
+        val candidates = (candidateSeeds zip checkThreads).collect { case (cand, hasChat) if !hasChat => cand }
```
stkem, line 141: @lawzlo This was backwards, filtering out all pages _without_ a conversation on them, instead of the one _with_ a conversation. Please add a test to cover this, so it doesn't happen again.

stkem:
I merging this to stop the build from breaking randomly

PR #10620, opened 2014-09-03
KeepCommander.allKeeps refactor

andrewconner:
@raymondng 

Re-wrote KeepCommander.allKeeps to make it clearer where data comes from. Not adding help rank attributions to all results right now because UI doesn't show it.

`withPageInfo` removed, it's always included now.


andrewconner:
Also added my goal, a `decorateKeepsIntoKeepInfos` @stkem

raymondng:
Looks good. We can add helprank later when needed

PR #10622, opened 2014-09-03
adding passcode to library_invites

ahsu1230:
@andrewconner


ahsu1230:
do not merge yet - still testing

kifi-backend/modules/common/conf/evolutions/shoebox/240.sql
```diff
@@ -0,0 +1,12 @@
+# SHOEBOX
+
+# --- !Ups
+
+ALTER TABLE library_invite
+  ADD COLUMN passcode varchar(20) NOT NULL;
+ALTER TABLE library_invite
+  MODIFY COLUMN auth_token varchar(40) NOT NULL;
```
ahsu1230, line 8: I'm removing the 2 test invites in the DB now so no invites will have NULL passcodes and since auth_tokens are always created anyway, we can change the column to be NOT NULL

kifi-backend/modules/common/app/com/keepit/model/LibraryInvite.scala
```diff
@@ -49,12 +50,20 @@ object LibraryInvite extends ModelWithPublicIdCompanion[LibraryInvite] {
...
 
   implicit def ord: Ordering[LibraryInvite] = new Ordering[LibraryInvite] {
     def compare(x: LibraryInvite, y: LibraryInvite): Int = x.access.priority compare y.access.priority
   }
+
+  def generatePasscode() = {
+    val randomNoun = Words.nouns(Random.nextInt(Words.nouns.length));
+    val randomAdverb = Words.adverbs(Random.nextInt(Words.adverbs.length));
+    val randomAdjective = Words.adjectives(Random.nextInt(Words.adjectives.length));
+    randomAdverb + " " + randomAdjective + " " + randomNoun
```
2is10, line 65: I thought we decided that two words is plenty. This isn’t for security. This is more like a captcha.

kifi-backend/modules/common/app/com/keepit/model/Words.scala
```diff
@@ -0,0 +1,8 @@
...
+}
```
2is10, line 8: Super long lines. Okay to wrap? Many editors can do it automatically.

I think we should limit word length to about 5 letters for library passcodes. Think about mobile. (These aren’t for security.)

We also need to be culturally sensitive. Words like “american” and “jewish” and certain verbs are likely to be offensive when paired with certain other words in these lists.

kifi-backend/modules/common/app/com/keepit/model/LibraryInvite.scala
ahsu1230, line 65: While we do have a lot of nouns (1850)
we only have like 200 adverbs and 600 adjectives and probably less if we take out longer words,
so if we use two words, we'll probably have a lot of duplicate combos. Do you think that's a problem?

ahsu1230, line 65: I guess on second thought, having duplicate passcodes for different invites isn't a huge deal

kifi-backend/modules/common/app/com/keepit/model/Words.scala
ymatsuda, line 8: Can you put the word lists in S3 rather than in the source file?

andrewconner, line 8: It's around ~18k and simplifies quite a bit to keep in source. Is it a size concern?

ymatsuda:
Putting large data in source is a bad practice in general.

I take S3 back. Why don't you use DB? That should be simple enough and easy to maintain.

If you still think the source code is the right place, please try to take it out of the "common" module.

andrewconner:
Agreed, minimally out of common. Good catch Yasuhiro.

PR #10624, opened 2014-09-03
Decorating library keeps!

andrewconner:
@ahsu1230


ahsu1230:
nice! if all the tests pass even from the Controller Future changes, looks good to me!

PR #10629, opened 2014-09-04
added can view library if have invite AND passcode is correct

ahsu1230:


ahsu1230:
sensitive words removed include:
anything racist & relating to religion
curse words (f bomb and s bomb)
words like "rape", "kill, "murder"

PR #10631, opened 2014-09-04
faster sampler?

yingjieMiao:
@stkem @leogrim 

a hybrid sampling:
1) use 20% of the trials to do random walk, estimate the reaching probability of users and topics.
2) distribute remaining sampling quota to related users/topics, proportional to the reaching probability. Do 'repeated sampling' at these vertexes. 
3) combine uri scores from step 1) and step 2)


stkem:
Nice! lgtm. 
Can you hook this up to an admin page similar to https://admin.kifi.com/admin/graph/discover so we can evaluate it (for speed, result set size and quality)? 
(doesn't need to be in this PR)

yingjieMiao:
of course, I will. Note the current admin page already 'benefit?' from binary search in sampling.

stkem:
right, thanks!

PR #10639, opened 2014-09-04
Pagination of library keep list

andrewconner:
@ahsu1230 @yipingliao


kifi-backend/modules/shoebox/angular/src/libraries/library.js
```diff
@@ -39,36 +43,33 @@ angular.module('kifi')
...
-      return 'Showing the ' + numShown + ' latest Keeps';
+      return 'Showing ' + numShown + ' of ' + $scope.library.numKeeps + ' latest Keeps';
     };
 
     $scope.scrollDistance = '100%';
     $scope.scrollDisabled = false;
 
     $scope.getNextKeeps = function () {
-      if ($scope.loading) {
+      if ($scope.loading || !$scope.library || $scope.keeps.length == 0) {
```
yipingliao: Use `===`.

kifi-backend/modules/shoebox/angular/src/libraries/library.js
```diff
@@ -12,16 +12,20 @@ angular.module('kifi')
 
     var libraryP = libraryService.getLibraryByUserSlug(username, librarySlug);
 
+    $scope.loading = true;
     libraryP.then(function (library) {
+      _.forEach(library.keeps, keepService.buildKeep);
```
yipingliao, line 17: I just made a PR that should make this call to `buildKeep` unnecessary.

kifi-backend/modules/shoebox/angular/src/libraries/library.js
```diff
@@ -39,36 +43,33 @@ angular.module('kifi')
...
         return;
       }
 
       $scope.loading = true;
-      return libraryService.getLibraryByUserSlug(username, librarySlug).then(function (list) {
+      return libraryService.getKeepsInLibrary($scope.library.id, $scope.keeps.length).then(function (res) {
         $scope.loading = false;
+        // $scope.scrollDisabled = true;
+        if (res.keeps) {
+          _.forEach(res.keeps, keepService.buildKeep);
```
yipingliao, line 62: Another place where `buildKeep` call may not be needed.

yipingliao:
:+1:

kifi-backend/modules/shoebox/angular/src/libraries/library.js
andrewconner, line 17: Thanks! I actually still want the augmentation for now (for the other fields), but will keep that in mind.

kifi-backend/modules/shoebox/angular/src/libraries/library.js
andrewconner: Thanks!

PR #10642, opened 2014-09-04
A few small curator adjustements

stkem:
see individual commits for specific changes
@yingjieMiao @lawzlo

### DIFF
--- a/kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationRetrievalCommander.scala
+++ b/kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationRetrievalCommander.scala
@@ -8,6 +8,8 @@ import com.keepit.common.akka.SafeFuture
 
 import play.api.libs.concurrent.Execution.Implicits.defaultContext
 
+import scala.util.Random
+
 import com.google.inject.{ Inject, Singleton }
 
 @Singleton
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationRetrievalCommander.scala
```diff
@@ -53,7 +55,11 @@ class RecommendationRetrievalCommander @Inject() (db: Database, uriRecoRepo: Uri
   }
 
   def topPublicRecos(): Seq[RecoInfo] = {
-    db.readOnlyReplica { implicit session => publicFeedRepo.getByTopMasterScore(1000) }.sortBy(-1 * _.updatedAt.getMillis).take(10).map { reco =>
+    val candidates = db.readOnlyReplica { implicit session =>
+      publicFeedRepo.getByTopMasterScore(1000)
+    }.sortBy(-1 * _.updatedAt.getMillis).take(100)
+
+    Random.shuffle(candidates).take(10).map { reco =>
```
lawzlo, line 62: Why we shuffle here? Do we want to show recos to user from highest to lowest interests so they will potentially read more?

stkem, line 62: This is just to make the public feel a bit more dynamic, making Kifi look more alive to new users.

stkem:
I just wanted to get some dynamism into the public feed, which can feel very static (there is no UI for the user to get more items). While sampling based on the score would indeed be better, I don't think it's worth the effort in this case.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationRetrievalCommander.scala
lawzlo, line 62: I see. :+1:

stkem:
I've added some recommendation score cutoff adjustments to this PR

yingjieMiao:
cool. Maybe we can consolidate all these cutoffs in one object, like SearchConfig does. So we don't have to jump around the code to modify those numbers.

stkem:
Good idea. 
Need to get this out to get some recos to the people who complained, but we should definitely do this. Will also allow us to have different configs for different people.

PR #10645, opened 2014-09-04
Implement library membership ingestion part + tests

lawzlo:


kifi-backend/modules/common/conf/evolutions/shoebox/241.sql
```diff
@@ -0,0 +1,22 @@
...
+CREATE TABLE if not exists curator_library_membership_info (
+    id bigint(20) NOT NULL AUTO_INCREMENT,
+    created_at datetime NOT NULL,
+    updated_at datetime NOT NULL,
+    state varchar(20) NOT NULL,
+
+    user_id bigint(20) NOT NULL,
+    library_id bigint(20) NOT NULL,
+    library_kind text NOT NULL,
+    library_acess text NOT NULL,
```
stkem: `access`. Also, let's keep the same SQL data types as used by the incoming data (i.e. `VARCHAR(20)` and `VARCHAR(64)` in this case instead of `text`)

kifi-backend/modules/curator/app/com/keepit/curator/commanders/LibraryMembershipSeedIngestionHelper.scala
```diff
@@ -0,0 +1,51 @@
...
+class LibraryMembershipSeedIngestionHelper @Inject() (
+    systemValueRepo: SystemValueRepo,
+    libInfoRepo: CuratorLibraryMembershipInfoRepo,
+    db: Database,
+    shoebox: ShoeboxServiceClient) extends GlobalSeedIngestionHelper with Logging {
+
+  private val SEQ_NUM_NAME: Name[SequenceNumber[Library]] = Name("all_library_seq_num")
+
+  private def processLibraryMemberships(libMembership: LibraryAndMemberships)(implicit session: RWSession): Unit = {
+    libMembership.memberships.foreach(membership =>
```
stkem: Not a big thing, but the convention is to use curly braces for blocks like this, i.e. `.foreach {`

lawzlo: cool. will change.

kifi-backend/modules/common/conf/evolutions/shoebox/241.sql
lawzlo: cool. fixing.

lawzlo:
updated evolution file and format.

kifi-backend/modules/common/conf/evolutions/shoebox/241.sql
```diff
@@ -0,0 +1,22 @@
...
+    updated_at datetime NOT NULL,
+    state varchar(20) NOT NULL,
+
+    user_id bigint(20) NOT NULL,
+    library_id bigint(20) NOT NULL,
+    library_kind VARCHAR(20) NOT NULL,
+    library_access VARCHAR(64) NOT NULL,
+
+    PRIMARY KEY (id),
+    INDEX curator_library_info_i_user_id_library_id (user_id, library_id)
```
stkem: This should be a unique key

kifi-backend/modules/curator/app/com/keepit/curator/commanders/LibraryMembershipSeedIngestionHelper.scala
```diff
@@ -0,0 +1,51 @@
...
+
+class LibraryMembershipSeedIngestionHelper @Inject() (
+    systemValueRepo: SystemValueRepo,
+    libInfoRepo: CuratorLibraryMembershipInfoRepo,
+    db: Database,
+    shoebox: ShoeboxServiceClient) extends GlobalSeedIngestionHelper with Logging {
+
+  private val SEQ_NUM_NAME: Name[SequenceNumber[Library]] = Name("all_library_seq_num")
+
+  private def processLibraryMemberships(libMembership: LibraryAndMemberships)(implicit session: RWSession): Unit = {
```
stkem: This also needs to deal with the case when we have seen the library already and it's just an update. This way you'll get an integrity constraint violation from the database (after index fix) if you save the item.

kifi-backend/modules/common/test/com/keepit/shoebox/FakeShoeboxServiceClientImpl.scala
```diff
@@ -630,7 +630,24 @@ class FakeShoeboxServiceClientImpl(val airbrakeNotifier: AirbrakeNotifier) exten
...
+    val membership2 = LibraryMembership(libraryId = Id[Library](1), userId = Id[User](42), access = LibraryAccess.OWNER, showInSearch = false)
+    val membership3 = LibraryMembership(libraryId = Id[Library](2), userId = Id[User](43), access = LibraryAccess.OWNER, showInSearch = false)
+    val membership4 = LibraryMembership(libraryId = Id[Library](2), userId = Id[User](43), access = LibraryAccess.OWNER, showInSearch = false)
+    val membership5 = LibraryMembership(libraryId = Id[Library](3), userId = Id[User](43), access = LibraryAccess.OWNER, showInSearch = false)
+
+    val libAndMembership1 = LibraryAndMemberships(lib1, Seq(membership1, membership2))
+    val libAndMembership2 = LibraryAndMemberships(lib2, Seq(membership3, membership4))
+    val libAndMembership3 = LibraryAndMemberships(lib3, Seq(membership5))
+    Future.successful(Seq(libAndMembership1, libAndMembership2, libAndMembership3))
+  }
```
yingjieMiao, line 646: this is not the right way to test. The `FakeShoeboxServiceClient` will need to have methods to save `library` and `library membership`. And this method should not be faked with fixed return data. 

c.f. `getUserIndexable` in the same class.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/SeedIngestionCommander.scala
```diff
@@ -55,6 +58,10 @@ class SeedIngestionCommander @Inject() (
     log.info("Ingested one batch of keeps.")
   }
 
+  def ingestLibraries(batchSize: Option[Int] = Some(INGESTION_BATCH_SIZE)): Future[Unit] = FutureHelpers.whilef(libraryIngestor(batchSize.getOrElse(INGESTION_BATCH_SIZE))) {
```
stkem: please also hook this up in the `ingestAll` method (after keeps, before graph)

kifi-backend/modules/common/test/com/keepit/shoebox/FakeShoeboxServiceClientImpl.scala
lawzlo, line 646: I checked saveLibrary or saveLibraryMembership at first time and couldn't find them or similar API in shoebox that could save library so I create fake return for `getLibrariesAndMembershipsChanged` API. Does this way will cause any problem?

andrewconner, line 646: There won't be a `saveLibrary` (at least, no use case for it now). A technique used elsewhere is to make a method that lets the testing class set the data it will later read.

So you add a new method, `setLibrariesAndMembershipsChanged(libraryAndMemberships: Seq[LibraryAndMemberships])` and let it mutate the fake shoebox client's internal state.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/LibraryMembershipSeedIngestionHelper.scala
lawzlo: Ya, I thought about this part before and was thinking just let the new memberships override old ones. but I realize now CuratorLibraryMembershipInfo id will increase every time so it doesn't override but create a lot of duplicates. I will have update CuratorLibraryMembershipInfo function based on userId as well. Thanks!

kifi-backend/modules/common/test/com/keepit/shoebox/FakeShoeboxServiceClientImpl.scala
andrewconner, line 646: The problem with returning fixed data is that the data you're returning is wrong — there is no User ID 42, these libraries don't have 10 members, etc. This causes issues later on, ie, if a consumer tries to look up the user.

lawzlo, line 646: Make sense. I will make the tests more dynamic so that the fake API can be used by other tests as well but not just fit my test case. Thanks.

yingjieMiao, line 646: Looks like the real `shoebox client` returns a bundle of `library` and `library memberships'. It is currently used by search indexer. Here, you want to copy two tables from shoebox to curator. I am not sure it's necessary (or right) to reuse that endpoint. Note that, a library's seq num could be bumped up, without any change of library memberships. To me it's clearer to have two tables on the curator side to follow shoebox's change, independently.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/SeedIngestionCommander.scala
lawzlo: hooking up.

kifi-backend/modules/common/test/com/keepit/shoebox/FakeShoeboxServiceClientImpl.scala
lawzlo, line 646: @yingjieMiao has great point about improving performance of ingesting libraryMembership. I will create a another internal endpoint in shoebox to listen to libraryMembership changes only and ingest to curator as discussed.

PR #10651, opened 2014-09-04
joining / leaving library changes library memberCount field

ahsu1230:
@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
@@ -337,7 +337,7 @@ class LibraryCommander @Inject() (
           case Some(mem) =>
             libraryMembershipRepo.save(mem.copy(state = LibraryMembershipStates.ACTIVE))
         }
-        libraryRepo.save(lib.copy(memberCount = lib.memberCount + 1))
+        libraryRepo.updateMemberCount(libraryId)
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -331,7 +331,13 @@ class LibraryCommander @Inject() (
...
       else {
         val maxAccess = if (listInvites.isEmpty) LibraryAccess.READ_ONLY else listInvites.sorted.last.access
-        libraryMembershipRepo.save(LibraryMembership(libraryId = libraryId, userId = userId, access = maxAccess, showInSearch = true))
+        libraryMembershipRepo.getOpt(userId, libraryId) match {
+          case None =>
+            libraryMembershipRepo.save(LibraryMembership(libraryId = libraryId, userId = userId, access = maxAccess, showInSearch = true))
+          case Some(mem) =>
+            libraryMembershipRepo.save(mem.copy(state = LibraryMembershipStates.ACTIVE))
+        }
+        libraryRepo.save(lib.copy(memberCount = lib.memberCount + 1))
```
andrewconner: For now, this is fine, but the race condition here may present problems later. It would be nice to make it calculated _in_ the db. ie, something like:

``` sql
UPDATE library SET member_count = (SELECT COUNT(id) FROM library_membership WHERE library_id = <id>) where id = <id>;
```

andrewconner:
Thanks!

PR #10662, opened 2014-09-05
lower the min requirement

yingjieMiao:
@stkem @eishay 

that was the cause (empty reco for those users). 

I need to find a way to fix these guys' interest profiles.


stkem:
:+1:

PR #10665, opened 2014-09-05
bring back lda update

yingjieMiao:
@stkem @leogrim 

Turned off this when there were tons of exceptions. hope this won't bring back exceptions.


stkem:
:+1:
I think it should be fine, the problems previously were likely all caused by the iterator modification issue (which presumably affected LDA more because of the high degree -> long iterator)

PR #10667, opened 2014-09-05
play23 -- test passed!

raymondng:
Dependencies updated

@andrewconner


andrewconner:
Yep

kifi-backend/modules/common/conf/dev-akka-sync.conf
```diff
@@ -9,6 +9,7 @@ play {
     loggers = ["akka.event.slf4j.Slf4jLogger"]
     loglevel = DEBUG
     log-config-on-start = false
+    jvm-exit-on-fatal-error = off
```
eishay, line 12: i thought we already have it

kifi-backend/project/Global.scala
```diff
@@ -50,11 +50,11 @@ object Global {
     jdbc, // todo(andrew): move to sqldb when we discover a way to get Play to support multiple play.plugins files.
     cache,
     ws,
-    "com.typesafe.play.plugins" %% "play-statsd" % "2.2.0" exclude("play", "*"),
-    "com.typesafe" %% "play-plugins-mailer" % "2.2.0" exclude("play", "*"),
+    "com.typesafe.akka" %% "akka-testkit" % "2.3.4" exclude("play", "*"),
```
eishay, line 53: great!

kifi-backend/modules/common/conf/dev-akka-sync.conf
raymondng, line 12: we have it in prod but not dev. dev was bombing out due to akka upgrade breakage, so i added this to debug. turns out we had some incompatible libraries -- see `Global.scala` -- fixed.

PR #10669, opened 2014-09-05
using the "private" page title in messaging emails

stkem:
@andrewconner 
also disallowing screenshots for the image in the small picture version
of the email, which look pretty bad (since they are tiny) and time out 
most of the time anyway.
The email code here is a mess (significant parts of it my fault).
Hopefully we can bring some better structure into this some time soon with @joshm1 new email tools.


andrewconner:
Awesome, thanks!

PR #10671, opened 2014-09-05
fixing suggested contacts endpoint to honor limit and desired ordering

2is10:
@andrewconner


andrewconner:
Why the sorting?

2is10:
Thanks.

Sorting because the UI currently groups Kifi users at the top and email address contacts beneath. This is how the typeahead branch works too, so we ought to be consistent. We may decide to relax that constraint in the future.

ahsu1230:
:thumbsup:

PR #10672, opened 2014-09-05
Using a neural network to size keep cards

andrewconner:
This is exploratory, in an attempt to calculate optimal keep card sizes more efficiently. It is over 10 times faster than the previous implementation, able to size 250 keep cards in 25ms.

The idea is that given (a) the card size (fixed, per window size), (b) the image's aspect ratio (fixed), (c) the description text's size (as measured in px, on one line), and (d) the sizing of the text area (a guess), a NN can calculate an approximate score (lower is better, think of it as error from optimal size, where text height = image height).

Previously, we drew a hidden keep card and made measurements. This is very slow because it makes the browser's rendering engine do a lot of work.

After scoring for 8 different sizes (variable (d) above), we pick the one with the lowest score (early termination on a particularly good score).

This NN was trained on 5k keep cards, sized using the old method. Error is 0.0006. The resulting JS was generated code (https://github.com/harthur/brain) and has not been modified yet. It does not currently pass the linter, as-is.

Before merging, I will keep a reference implementation, and document exactly how to re-train the model.

Shortcomings:
- Does not even attempt to model text wrapping, which happens on word-breaks, not based on length. This matters if the description has particularly long words, or the available screen space is very small.
- Not transparent regarding _why_ a card is sized how it is.

This isn't ready to be merged, but I would appreciate feedback, especially if this is a silly idea. @stkem @yipingliao @yingjieMiao @ymatsuda @leogrim @eishay @2is10


ymatsuda:
Do you anticipate any i18n issue?

andrewconner:
@ymatsuda I've tested it with a few Chinese and Japanese keeps and it seems to work well. That was the reason I use the description length (measured in pixels, if it were on one line) instead of character length. I'll test more before rolling it out. If you could send me a list of pages, that would be great.

yingjieMiao:
nice! I like the idea. some questions:
1) was the error measured on training set or testing set? 
2) did you formulate this as a regression or classification problem?
3) size of the hidden layer?

andrewconner:
1) Testing
2) Regression, calculating the score of the "guess" (width of text area). Here's a sample input and output:

```
{input: {cardWidth: 0.532, aspectRatio: 0.26666666666666666, trimmed: 1.2976, textWidth: 0.36}, output: {score: 0.311}}
```

3) One layer, 5 nodes

yingjieMiao:
:+1:

joshm1:
:+1: love it!

yipingliao:
Nice! :) We should definitely try it; we can always revert if we start to see strange behavior.

kifi-backend/modules/shoebox/angular/src/libraries/library.js
```diff
@@ -16,9 +16,8 @@ angular.module('kifi')
     libraryP.then(function (library) {
       _.forEach(library.keeps, keepService.buildKeep);
       $scope.library = library;
-      $scope.keeps = library.keeps || [];
+      $scope.keeps.push.apply($scope.keeps, library.keeps);
```
yipingliao, line 30: Will `$scope.keeps = $scope.keeps.concat(library.keeps)` work? It may be slightly clearer than using `apply`.

andrewconner, line 30: Whoops, this wasn't intended for this PR. concat creates a new array, while push.apply appends to the existing. I'd like to keep the same object if possible (hence this change)

kifi-backend/modules/shoebox/angular/src/layout/nav/nav.tpl.html
```diff
@@ -1,19 +1,32 @@
...
     <a class="kf-nav-link" href="/">
       <span class="sprite sprite-my-keeps-library"></span>
       <span class="kf-nav-name">Keeps Library</span>
-      <span class="kf-nav-count" ng-class="{ empty: !counts.keepCount }">{{counts.keepCount}}</span>
+      <span class="kf-nav-count allcaps" ng-class="{ empty: !counts.keepCount }">{{counts.keepCount}}</span>
     </a>
   </li>
 </ul>
 <ul class="kf-nav-group" ng-if="librariesEnabled">
+  <li class="kf-nav-item kf-nav-recos" ng-class="{ active: isActive('/recommendations') }">
```
stkem, line 17: Is this block related?

stkem:
:+1:

andrewconner:
Closing because now's not the time. Will revisit later.

PR #10678, opened 2014-09-05
new email API and email template processor

joshm1:
The purpose of this API and the email template is to abstract away the process of fetching common dynamic values.  As we add more emails, we don't want each emailer to worry with fetching common user attributes (firstName, lastName, avatar/image Url, email address). We should see a performance improvement as well (less service calls and database requests).  If we have to fetch a user object or BasicUser objects for multiple purposes, there could be duplicate service calls and/or database requests.  

For example, using placeholders in the email template, the sender will fetch a `Set` of `Id[User]` that are needed in the template (regardless of why they are needed).  We can also add dynamic information that differs per email in the header/footer (which different emails shouldn't be concerned with) without the additional complexity of having to deal with multiple parameters that are passed into each template.. like the "campaign" that's put in URLs for tracking, or the unsubscribeUrl in the footer.

I refactored the Digest Email to use this new email API, but still want to refactor some more to use placeholders for BasicUser information and avatar image Urls.

@stkem @eishay @2is10 @andrewconner 

P.S. I'm terrible at names, feel free to suggest better names for these classes.  ( @2is10  usually has good suggestions ; ) ). I particularly don't like the "EmailModule"  (since Module is usually used in our code for DI).


kifi-backend/modules/shoebox/app/com/keepit/commanders/emails/EmailHtmlPreProcessor.scala
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  private def getUserImageUrls(userIds: Seq[Id[User]], width: Int = 100) = {
+    Future.sequence(
+      userIds map (userId => userCommander.getUserImageUrl(userId, width).map(url => (userId, url)))
+    ) map (_.toMap)
+  }
+
+  private def getUnsubUrl(emailAddr: EmailAddress) = {
+    // tokens end with \r\n and I don't like it
```
andrewconner: We should probably fix this :) (instead of this workaround)

kifi-backend/modules/shoebox/test/com/keepit/commanders/emails/EmailTemplateProcessorImplTest.scala
```diff
@@ -0,0 +1,83 @@
...
+          htmlContent = Seq(html1, html2),
+          unsubscribableEmail = Some(user3.primaryEmail.get)
+        )
+
+        val outputF = processor.process(emailModule)
+        val output = Await.result(outputF, Duration(5, "seconds")).body
+
+        output must contain("Aaron Paul and Bryan Cranston joined!")
+        output must contain("""<img src="http://cloudfront/1_100x100_0" alt="Aaron Paul"/>""")
+        output must contain("""<img src="http://cloudfront/2_100x100_0" alt="Bryan Cranston"/>""")
```
andrewconner, line 77: :+1:

kifi-backend/modules/common/app/com/keepit/common/mail/ElectronicMail.scala
```diff
@@ -33,6 +34,45 @@ object ElectronicMailCategory {
...
+  from: EmailAddress,
+  fromName: Option[String] = None,
+  to: Seq[EmailAddress] = Seq[EmailAddress](),
+  cc: Seq[EmailAddress] = Seq[EmailAddress](),
+  subject: String,
+  htmlContent: Seq[Html],
+  category: ElectronicMailCategory,
+  campaign: Option[String] = None,
+  senderUserId: Option[Id[User]] = None,
+  unsubscribableEmail: Option[EmailAddress] = None)
```
2is10: mostly stuff we already discussed:
- maybe name class `EmailToSend`? other possibilities: `EmailSkeleton`, `EmailSpec`, `EmailDraft`
- limit `to` to one `EmailAddress` and eliminate `unsubscribableEmail`
- name the `Option[Id[User]]` field accurately (`toUserId` or `recipientUserId` perhaps?) or eliminate it
- `htmlContent` might be better named `partials` or `templates` ("html" maybe be unnecessary, already in the type)

2is10:
I really like this system for specifying and populating emails. Very nice.

kifi-backend/modules/shoebox/app/com/keepit/commanders/emails/EmailHtmlPreProcessor.scala
joshm1: You're right. I did this before I realized it was a production issue as well (at first, thought it might be a dev issue related to the crypt key).

kifi-backend/modules/common/app/com/keepit/model/Email.scala
```diff
@@ -0,0 +1,128 @@
+package com.keepit.model
+
+import com.keepit.common.db.Id
+import com.keepit.common.mail.EmailAddress
+import play.api.libs.json._
+import play.api.templates.Html
+
+import scala.util.matching.Regex
+
+object Email {
```
stkem, line 10: good use case for a `package object`?

kifi-backend/modules/common/app/com/keepit/shoebox/ShoeboxServiceClient.scala
```diff
@@ -111,6 +111,7 @@ trait ShoeboxServiceClient extends ServiceClient {
   def getInvitations(senderId: Id[User]): Future[Seq[Invitation]]
   def getSocialConnections(userId: Id[User]): Future[Seq[SocialUserBasicInfo]]
   def addInteractions(userId: Id[User], actions: Seq[(Either[Id[User], EmailAddress], String)]): Unit
+  def sendMailModule(email: EmailToSend): Future[Boolean]
```
stkem: You may want to get rid of the `module` naming here too. Maybe `processAndSendMail`? (since it does some extra steps before sending.

kifi-backend/modules/common/app/com/keepit/model/Email.scala
joshm1, line 10: hmm, good idea. hadn't thought of that.  won't refactor for this PR, may do it in the future

kifi-backend/modules/common/app/com/keepit/model/Email.scala
```diff
@@ -0,0 +1,132 @@
...
+
+    override def toString() =
+      tagLeftDelim + JsArray(Seq(JsString(label.value)) ++ args).toString() + tagRightDelim
+  }
+
+  case class Tag0(label: TagLabel) extends Tag {
+    def args = Seq.empty
+  }
+
+  case class Tag1[S](label: TagLabel, arg: S)(implicit val writer: Writes[S]) extends Tag {
```
stkem, line 50: How come there is no `Tag2`?

joshm1, line 50: Good observation. Didn't have a needed for it. Was prototyping during implementation and had a temporary need for a Tag3 but eventually took that out.

kifi-backend/modules/common/app/com/keepit/model/Email.scala
```diff
@@ -0,0 +1,132 @@
...
+    val unsubscribeUrl = TagLabel("unsubscribeUrl")
+    val unsubscribeUserUrl = TagLabel("unsubscribeUserUrl")
+    val unsubscribeEmailUrl = TagLabel("unsubscribeEmailUrl")
+    val userExternalId = TagLabel("userExternalId")
+    val campaign = TagLabel("campaign")
+    val title = TagLabel("title")
+    val baseUrl = TagLabel("baseUrl")
+  }
+
+  val tagLeftDelim: String = "<% "
```
stkem, line 72: One thing you may want to test for is what happens when any (or both) of the delimiters appear in the payload of an email somehow (I'm thinking page description of some Article on php, for example)

2is10, line 72: We HTML-escape titles, descriptions, and any other text we insert into HTML, or should be.

Harder would be discriminating these kinds of tags in plain-text email partials, but it looks like we don’t currently use any of those.

joshm1, line 72: Good point. I'll add more unique delimiters.

On Fri, Sep 5, 2014 at 2:58 PM, Stephen Kemmerling <notifications@github.com

2is10, line 72: My main point was that the use of `<` and `>` in your delimiters `<%` and `%>` makes them safe for use in our HTML templates, since we always HTML-escape those characters to `&lt;` and `&gt;` when inserting third-party content into our HTML. So we could never mistakenly interpret any third-party content as our tags.

joshm1, line 72: Understood. Thanks!

PR #10685, opened 2014-09-05
New keep cards icon

yipingliao:
- Use new asset (cards icon) to represent keep cards
- Per Danny's feedback, shorten the recommendations slider, and when it's too short to display the words fully, display 'At' for 'All Time' and 'Re' for 'Recent'
- Remove `allcaps` styles-only class.

@2is10


kifi-backend/modules/shoebox/angular/src/layout/main/main.styl
```diff
@@ -28,12 +28,19 @@
   .kf-main-search &
     padding-top: 10px
 
+  .kf-keeps-card-icon
+    display: inline-block
+    margin: -8px 8px 0 0
```
2is10, line 33: a named or px value for `vertical-align` might be a little more stable and self-describing than a negative top margin

kifi-backend/modules/shoebox/angular/src/layout/main/main.styl
```diff
@@ -28,12 +28,19 @@
...
+    margin: -8px 8px 0 0 
+
 .kf-main-title
   margin: 0 0 2px
   font-size: 25px
   line-height: 30px
   font-weight: 300
 
+  .kf-main-head &
+    display: inline-block
```
2is10, line 42: Putting the icon inside the `h1` would make this unnecessary.

kifi-backend/modules/shoebox/angular/src/layout/main/main.styl
```diff
@@ -51,6 +58,9 @@
   .sprite-edit-pencil-icon
     margin-right: 5px
 
+  .kf-main-head &
+    margin: -3px 0 0 40px
```
2is10, line 62: Context-specific rule overrides like this can really become a drag over time. They’re harder to find and maintain properly. Maybe we can look at the layout together.

kifi-backend/modules/shoebox/angular/src/recos/recos.styl
```diff
@@ -49,9 +49,21 @@
 .kf-reco-recency-all-time
   left: 0
 
+  &::before
+    content: 'All time'
+
+    @media (min-width: 1024px) and (max-width: 1100px)
+      content: 'At'
```
2is10, line 56: I like the idea, but these @media queries just aren’t reliable. We’ll need to the use script to do it reliably.

2is10:
Nice work. Thanks for discussing with me in person. I know it can be challenging to address everyone’s concerns. Feel free to merge when you’re done.

yipingliao:
Closing this for now; will make another pull.

PR #10688, opened 2014-09-05
reduce topic connections of uri in graph

yingjieMiao:
@stkem @leogrim 

this reduces the size of graph, and helps with speed of random walk.


stkem:
:+1

PR #10689, opened 2014-09-05
quartz && netty

raymondng:
Found a couple issues via ad hoc testing locally. 

Turns out our quartz dependency does not work with the latest play/akka versions. I upgraded to 0.3.0 but got unresolved dependency error. Adding the quartz repo addressed that, but not sure if that's expected.

Question for @eishay: do/should we need to manually build & add the jar to our nexus repo? I asked because I noticed that our previous quartz version was somewhat "special".

Also removed one cast to `NingWSResponse`.

I'd like to do a bit local testing, and we can try test instance after lunch. 

@andrewconner @eishay


eishay:
I don't know/remember

raymondng:
Okay assuming no major objections we'll go with the current approach. Of course I'll continue testing effort.

PR #10693, opened 2014-09-05
adding beta feature label to Recommendations page

2is10:
@yipingliao


2is10:
merging to avoid conflicts. we can tweak if necessary.

PR #10697, opened 2014-09-05
filter out long tail noise during pre-score

yingjieMiao:
@stkem FYI


stkem:
:+1: 
Adding @leogrim just saw he has a complete history of all that happened

PR #10701, opened 2014-09-05
New keep cards icon

yipingliao:
@2is10 

Second try at this. :)
- Merged in your "beta" addition, but modified it to say just "beta" because the longer phrase looked really squished in there.
- Addressed code review comments from the previous pull.
- Made slider even smaller (in width), but kept it to 5 settings (per @stkem )

Hopefully this will be it for recos frontend for at least a week!

Thanks,
Yiping


2is10:
Great!

PR #10703, opened 2014-09-05
removing pre-Angular Kifi web app

2is10:
@andrewconner


andrewconner:
Thanks! I _think_ `/img` is old too, but it's hard to be sure.

raymondng:
:+1:

PR #10707, opened 2014-09-05
endpoint to recompute user interests

yingjieMiao:
@stkem 

the endpoint will trigger a scan of all users in 'interest profile' repo. However, in the actual update plugin, there is a guard to test if it's necessary to compute the feature. (Mainly used to defend people like Jeff. So for people with few keeps, the guard will let it go).


stkem:
lgtm

PR #10708, opened 2014-09-05
typeahead -- dedup email addresses

raymondng:
This should do it

@andrewconner @leogrim


2is10:
thanks

leogrim:
I think this doesn't account for potential case differences in email addresses and doesn't guarantee that we pick the duplicate address with the highest score. The original code did, see https://github.com/kifi/keepit/commit/dc6102371c7d3cc5aae6158484072c4cafb21c68

raymondng:
Thanks. Will take care of it.

Didn't know that we now persist in varying cases. I distinctly remember
that we made an earlier decision to persist uniformly (via toLower). There
are different trade-offs ... I'll catch up with you when you come back.

andrewconner:
Feel free to merge when you're happy — wasn't sure if @leogrim's comment needed new code?

raymondng:
Thanks. Yes new code (or porting old code) will be required. I might merge
this first and handle it later.

leogrim:
@raymondng about case, @2is10 and I then decided that it was best to keep user's data intact, even though we should always manipulate email addresses in a case insensitive fashion, like the database does.

raymondng:
Thanks for the feedback. Will merge & iterate.

PR #10710, opened 2014-09-05
more timing in graph sampling

yingjieMiao:
@stkem 

This is temporary. The point is: if actual sampling dominate the time, then we cannot do much (unless we do a sparse representation of edge probabilities). 

(this won't cause log flood. every uriWander call generates at most a few hundreds of these. )


stkem:
:+1:

yingjieMiao:
@stkem @leogrim  
as the log shows, current random walk has a heavy bottleneck in gathering transition probabilities at a vertex. Actual sampling takes no time. 

c.f.

sampling destination for (UserReader,KeepReader,TimestampEdgeReader): 126 millis to gather edge weights, edge size: 4917, 1 millis to sample 1575 times 

sampling destination for (LDATopicReader,UriReader,WeightedEdgeReader): 440 millis
 to gather edge weights, edge size: 205833, 1 millis to sample 1319 times

PR #10712, opened 2014-09-05
data sanity checking: keep.visibility must match library visibility & in...

ahsu1230:
...ternal system generated libraries (1 main & 1 secret per user, ownership access for both libraries)

@andrewconner @ymatsuda


kifi-backend/modules/shoebox/app/com/keepit/integrity/OrphanCleaner.scala
```diff
@@ -173,6 +174,15 @@ class OrphanCleaner @Inject() (
       }
 
       db.readWriteSeq(bookmarks, collector) { (s, bookmark) =>
+        bookmark.libraryId match {
```
ahsu1230, line 177: even though this part doesn't relate to normalized URI checking, it goes through batches of bookmarks anyway - so I'll add a visibility checker. The keep's visibility must match the corresponding library's visibility.
What do you think? @ymatsuda

PR #10715, opened 2014-09-05
Fixing ng assets

andrewconner:
This should do it. It's a short term hack, I can explain what the issue was and why this fixes it.

@raymondng


raymondng:
Tricky. Thanks @andrewconner for sorting this out.

PR #10719, opened 2014-09-06
fix digest QA email that depends on real user ids

joshm1:
@stkem

### DIFF
--- a/kifi-backend/modules/curator/app/com/keepit/curator/commanders/email/FeedDigestEmailSender.scala
+++ b/kifi-backend/modules/curator/app/com/keepit/curator/commanders/email/FeedDigestEmailSender.scala
@@ -191,24 +191,26 @@ class FeedDigestEmailSenderImpl @Inject() (
   }
 
   private def sendAnonymoizedEmailToQa(module: EmailToSend, emailData: AllDigestRecos): Unit = {
-    def fakeUserId = Id[User](Random.nextInt(Int.MaxValue))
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/curator/app/com/keepit/curator/commanders/email/FeedDigestEmailSender.scala
```diff
@@ -191,24 +191,26 @@ class FeedDigestEmailSenderImpl @Inject() (
   }
 
   private def sendAnonymoizedEmailToQa(module: EmailToSend, emailData: AllDigestRecos): Unit = {
-    def fakeUserId = Id[User](Random.nextInt(Int.MaxValue))
+    val userIds = Seq(1, 3, 9, 48, 61, 100, 567, 2538, 3466, 7100, 7456).map(i => Id[User](i.toLong)).sortBy(_ => Random.nextInt())
```
stkem, line 194: Bit of a hack. Deserves at least a comment of what's going on here, imho.

joshm1, line 194: It is, and I don't usually settle for hacks like this, but given this is an internal-QA email I didn't think it was worth the additional effort to query a set of random users for this.  I'll add a comment, and can add a less-hacky "fake friends for QA" implementation if you'd prefer.

PR #10721, opened 2014-09-08
Fixing frontend icons, backend url (pre) normalization

andrewconner:
@leogrim Is this the recommended way to get a URI id, given a URL string nowdays?


andrewconner:
Merging to test with prod data. Shouldn't hurt anything :8ball:

PR #10730, opened 2014-09-08
email contact notification bug fix

joshm1:
@2is10 @stkem

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/HomeController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/HomeController.scala
@@ -250,7 +250,7 @@ class HomeController @Inject() (
 
   def install = HtmlAction.authenticated { implicit request =>
     SafeFuture {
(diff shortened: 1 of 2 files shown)


stkem:
lgtm

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/HomeController.scala
```diff
@@ -250,7 +250,7 @@ class HomeController @Inject() (
 
   def install = HtmlAction.authenticated { implicit request =>
     SafeFuture {
-      userCommander.tellUsersWithContactOfNewUserImmediate(request.user)
+      if (!hasSeenInstall) userCommander.tellUsersWithContactOfNewUserImmediate(request.user)
```
2is10, line 253: This seems like an odd place for this. This is a page that any user can view at any time, and new users can avoid (e.g. by closing a tab after signup or by signing up using our mobile apps).

Perhaps a better place would be at the end of the signup flow code path, like in `SignupHelper.finishSignup`, done asynchronously.

2is10:
thanks

PR #10739, opened 2014-09-08
trying to improve getKeepsAndTagsChanged

ymatsuda:
@leogrim 
- There were too many db calls. This change uses joins (collection and keep_to_collection) in db.
- We should bypass caches for indexing.


leogrim:
Ok!

PR #10740, opened 2014-09-08
Fixing less compilation

andrewconner:
@raymondng


andrewconner:
Also:

```
java.lang.ClassCastException: com.ning.http.client.providers.netty.NettyResponse cannot be cast to play.api.libs.ws.ning.NingWSResponse
PagePeekerClient.scala:53:in com.keepit.common.pagepeeker.PagePeekerClientImpl[a]#apply
PagePeekerClient.scala:52:in com.keepit.common.pagepeeker.PagePeekerClientImpl[a]#apply
```

PR #10747, opened 2014-09-09
add visibility and library id to cortex keep table

yingjieMiao:
@stkem 

both fields are optional for backward compatibility.


stkem:
lgtm

yingjieMiao:
will merge this tomorrow morning in case bad things happen.

PR #10751, opened 2014-09-09
removing Google I/O choice from the guide b/c it breaks the guide now

2is10:
@ahsu1230 

This will reduce the number of page choices from 4 to 3, for now. (Product has been considering dropping it to 2.)


ahsu1230:
oh thanks. Wonder why that particular page breaks the guide

PR #10753, opened 2014-09-09
Change 'trimRight()' to 'trim()' for better browser support.

yipingliao:
Also, call 'formatTitleFromUrl' for recommendation attributions only if there is a url.
https://app.asana.com/0/15589946826415/15785999026810

I found a bug while testing recommendations on IE10 where the browser does not recognize the `trimRight()` method on JavaScript strings. This is also a problem for keeps, but the code path that contains `trimRight()` is accessed more frequently for recommendations. Switching to `trim()`, which is supported by IE9 and up.

@2is10 cc @stkem

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/keeps/keepActionService.js
+++ b/kifi-backend/modules/shoebox/angular/src/keeps/keepActionService.js
@@ -33,7 +33,7 @@ angular.module('kifi')
           };
         })
       };
-      $log.log('keepService.keep()', data);
+      $log.log('keepActionService.keep()', data);
 
       var url = env.xhrBase + '/keeps/add';
       var config = {
--- a/kifi-backend/modules/shoebox/angular/src/recos/recoDecoratorService.js
+++ b/kifi-backend/modules/shoebox/angular/src/recos/recoDecoratorService.js
@@ -29,7 +29,7 @@ angular.module('kifi')
         fileNameIdx += fileNameMatch.index;
         fileName = fileNameMatch[0];
       }
-      fileName = fileName.replace(fileNameToSpaceRe, ' ').trimRight().trimLeft();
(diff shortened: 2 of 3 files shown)


kifi-backend/modules/shoebox/angular/src/keeps/keepActionService.js
```diff
@@ -33,7 +33,7 @@ angular.module('kifi')
           };
         })
       };
-      $log.log('keepService.keep()', data);
+      $log.log('keepActionService.keep()', data);
```
yipingliao, line 36: This is just for clarification in logging.

kifi-backend/modules/shoebox/angular/src/recos/recoDecoratorService.js
```diff
@@ -29,7 +29,7 @@ angular.module('kifi')
         fileNameIdx += fileNameMatch.index;
         fileName = fileNameMatch[0];
       }
-      fileName = fileName.replace(fileNameToSpaceRe, ' ').trimRight().trimLeft();
+      fileName = fileName.replace(fileNameToSpaceRe, ' ').trim();
```
yipingliao, line 32: On the branch where I am doing refactoring, all 3 instances of `trim()` is DRYed up into one.

kifi-backend/modules/shoebox/angular/src/recos/recoDecoratorService.js
```diff
@@ -126,7 +126,7 @@ angular.module('kifi')
       };
 
       this.recoData.reasons.forEach(function (reason) {
-        if (!reason.name) {
+        if (!reason.name && reason.url) {
```
yipingliao, line 129: Only convert url to title if there is a url to begin with.

2is10:
Great!

PR #10756, opened 2014-09-09
endpoint for adding keeps to a library

ahsu1230:
@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
@@ -347,31 +347,24 @@ class LibraryController @Inject() (
     }
   }
 
-  def addKeeps(pubId: PublicId[Library], separateExisting: Boolean = false) = JsonAction.authenticatedParseJson { request =>
+  def addKeeps(pubId: PublicId[Library]) = JsonAction.authenticatedParseJson { request =>
     Library.decodePublicId(pubId) match {
       case Failure(ex) =>
         BadRequest(Json.obj("error" -> "invalid id"))
       case Success(libraryId) =>
-        try {
-          Json.fromJson[RawBookmarksWithCollection](request.body).asOpt map { fromJson =>
-            val source = KeepSource.site
-            implicit val context = heimdalContextBuilder.withRequestInfoAndSource(request, source).build
+        (request.body \ "keeps").asOpt[Seq[RawBookmarkRepresentation]] map { fromJson =>
+          val source = KeepSource.site
+          implicit val context = heimdalContextBuilder.withRequestInfoAndSource(request, source).build
 
-            val (keeps, addedToCollection, failures, alreadyKeptOpt) = keepsCommander.keepMultiple(fromJson.keeps, libraryId, request.userId, source, fromJson.collection, separateExisting)
-            log.info(s"kept ${keeps.size} keeps")
-            Ok(Json.obj(
-              "keeps" -> keeps,
-              "failures" -> failures,
-              "addedToCollection" -> addedToCollection
-            ) ++ (alreadyKeptOpt map (keeps => Json.obj("alreadyKept" -> Json.toJson(keeps))) getOrElse Json.obj()))
-          } getOrElse {
-            log.error(s"can't parse object from request ${request.body} for user ${request.user}")
-            BadRequest(Json.obj("error" -> "Could not parse object from request body"))
-          }
-        } catch {
-          case e: Throwable =>
-            log.error(s"error keeping ${request.body}", e)
-            throw e
+          val (keeps, _, failures, alreadyKeptOpt) = keepsCommander.keepMultiple(fromJson, libraryId, request.userId, source, None, true)
+          log.info(s"kept ${keeps.size} keeps")
+          Ok(Json.obj(
+            "keeps" -> keeps,
+            "failures" -> failures
+          ) ++ (alreadyKeptOpt map (keeps => Json.obj("alreadyKept" -> Json.toJson(keeps))) getOrElse Json.obj()))
+        } getOrElse {
+          log.error(s"can't parse object from request ${request.body} for user ${request.user}")
+          BadRequest(Json.obj("error" -> "Could not parse object from request body"))
         }
     }
   }
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
@@ -9,6 +9,7 @@ import com.keepit.common.db.slick.Database
 import com.keepit.common.mail.EmailAddress
 import com.keepit.common.social.BasicUserRepo
 import com.keepit.common.time.Clock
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/conf/shoebox.routes
```diff
@@ -263,6 +263,7 @@ POST    /site/libraries/:id/join    @com.keepit.controllers.website.LibraryContr
 POST    /site/libraries/:id/decline @com.keepit.controllers.website.LibraryController.declineLibrary(id: PublicId[Library])
 POST    /site/libraries/:id/leave   @com.keepit.controllers.website.LibraryController.leaveLibrary(id: PublicId[Library])
 GET     /site/libraries/:id/keeps   @com.keepit.controllers.website.LibraryController.getKeeps(id: PublicId[Library], count: Int, offset: Int ?= 0, auth:Option[String] ?= None, passcode:Option[String] ?= None)
+POST    /site/libraries/:id/addKeeps    @com.keepit.controllers.website.LibraryController.addKeeps(id: PublicId[Library], separateExisting: Boolean ?= false)
```
andrewconner: Since you're using POST, you could just do `/site/libraries/:id/keeps` to keep the same URL as getting keeps.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
```diff
@@ -344,5 +346,35 @@ class LibraryController @Inject() (
...
+              "keeps" -> keeps,
+              "failures" -> failures,
+              "addedToCollection" -> addedToCollection
+            ) ++ (alreadyKeptOpt map (keeps => Json.obj("alreadyKept" -> Json.toJson(keeps))) getOrElse Json.obj()))
+          } getOrElse {
+            log.error(s"can't parse object from request ${request.body} for user ${request.user}")
+            BadRequest(Json.obj("error" -> "Could not parse object from request body"))
+          }
+        } catch {
+          case e: Throwable =>
```
andrewconner: Did you notice it throwing exceptions?

ahsu1230: No, but its what KeepsController does, so I figured there might be a reason the exception handler is there.

kifi-backend/modules/shoebox/conf/shoebox.routes
ahsu1230: yea, maybe - just wanted to be a bit more explicit lol

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
```diff
@@ -344,5 +346,35 @@ class LibraryController @Inject() (
...
     }
   }
+
+  def addKeeps(pubId: PublicId[Library], separateExisting: Boolean = false) = JsonAction.authenticatedParseJson { request =>
+    Library.decodePublicId(pubId) match {
+      case Failure(ex) =>
+        BadRequest(Json.obj("error" -> "invalid id"))
+      case Success(libraryId) =>
+        try {
+          Json.fromJson[RawBookmarksWithCollection](request.body).asOpt map { fromJson =>
```
andrewconner: This is probably not a good class to use, especially since tags are changing. Supporting titles is fine, though. Usually, though, the site won't have any title, but I suppose it doesn't hurt anything.

From the site's point of view, a nice API might be something like:

``` json
{"keeps": [{"url": "http://site.com", "title":"this is optional, too"}]}
```

If we support tags later in this endpoint, it probably won't be `Either[ExternalId[Collection], String]`.

andrewconner:
Nice. Left a few comments, feel free to merge when ready.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
```diff
@@ -344,5 +346,35 @@ class LibraryController @Inject() (
...
+            val source = KeepSource.site
+            implicit val context = heimdalContextBuilder.withRequestInfoAndSource(request, source).build
+
+            val (keeps, addedToCollection, failures, alreadyKeptOpt) = keepsCommander.keepMultiple(fromJson.keeps, libraryId, request.userId, source, fromJson.collection, separateExisting)
+            log.info(s"kept ${keeps.size} keeps")
+            Ok(Json.obj(
+              "keeps" -> keeps,
+              "failures" -> failures,
+              "addedToCollection" -> addedToCollection
```
andrewconner: Not needed.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
andrewconner: Hm, not sure. KeepsController / KeepsCommander is not very good, it's super fragmented — the same thing is done several times, because use cases were _slightly_ different or someone didn't know about the other parts. Ideally we'll be able to fix a lot of the bad patterns.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
```diff
@@ -344,5 +346,35 @@ class LibraryController @Inject() (
         }
     }
   }
+
+  def addKeeps(pubId: PublicId[Library], separateExisting: Boolean = false) = JsonAction.authenticatedParseJson { request =>
```
andrewconner: Let's stick to one or the other with `separateExisting`. I think this was a backwards compatibility thing.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
ahsu1230: Ah in `KeepInterner.internUriAndBookmark` failed kept URIs throw exceptions, but the exception is caught inside the function anyway.
Sooo we might not need an exception handler anyway

kifi-backend/modules/shoebox/conf/shoebox.routes
```diff
@@ -263,6 +263,7 @@ POST    /site/libraries/:id/join    @com.keepit.controllers.website.LibraryContr
 POST    /site/libraries/:id/decline @com.keepit.controllers.website.LibraryController.declineLibrary(id: PublicId[Library])
 POST    /site/libraries/:id/leave   @com.keepit.controllers.website.LibraryController.leaveLibrary(id: PublicId[Library])
 GET     /site/libraries/:id/keeps   @com.keepit.controllers.website.LibraryController.getKeeps(id: PublicId[Library], count: Int, offset: Int ?= 0, auth:Option[String] ?= None, passcode:Option[String] ?= None)
+POST    /site/libraries/:id/addKeeps    @com.keepit.controllers.website.LibraryController.addKeeps(id: PublicId[Library])
```
2is10, line 266: I’d second Andrew’s recommendation for POST to `/libraries/:id/keeps`—standard REST semantics.

When we do need a verb in a public route, we put it last (e.g. `/libraries/:id/keeps/add`).

ahsu1230, line 266: I see. okay will change in next push

PR #10757, opened 2014-09-09
switching over curator ingestion to new graph endpoint

stkem:
if all goes well, I’ll remove the old logic
@yingjieMiao (cc @leogrim)

### DIFF
--- a/kifi-backend/modules/curator/app/com/keepit/curator/commanders/TopUriSeedIngestionHelper.scala
+++ b/kifi-backend/modules/curator/app/com/keepit/curator/commanders/TopUriSeedIngestionHelper.scala
@@ -92,9 +92,10 @@ class TopUriSeedIngestionHelper @Inject() (
             }
           }
 
-          val normalizationFactor = if (uriScores.isEmpty) 0.0f else uriScores.values.max.toFloat
+          val rescaledUriScores = uriScores.mapValues(score => Math.log(score.toDouble + 1).toFloat)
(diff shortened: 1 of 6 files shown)


kifi-backend/modules/curator/app/com/keepit/curator/commanders/TopUriSeedIngestionHelper.scala
```diff
@@ -93,9 +92,12 @@ class TopUriSeedIngestionHelper @Inject() (
             }
           }
 
-          uriScores.foreach { uriScore =>
-            log.debug(s"ingesting uri score is: ${uriScore.score.toFloat}, related user id is: ${uriScore.uriId.toString}")
-            processUriScores(uriScore, userId)
+          val normalizationFactor = if (uriScores.isEmpty) 0.0f else uriScores.values.max.toFloat
+
```
yingjieMiao, line 97: you might want to tune this. with current settings,  high score could be ~ 1000, tails have score 1~2.  This makes those scores effectively useless.

yingjieMiao, line 97: e.g. consider them in log scale

stkem, line 97: good idea, I'll do that

PR #10764, opened 2014-09-09
shoebox library data pipe

yingjieMiao:
@stkem

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/internal/ShoeboxDataPipeController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/internal/ShoeboxDataPipeController.scala
@@ -163,4 +163,14 @@ class ShoeboxDataPipeController @Inject() (
     }
     Ok(Json.toJson(keepAndTagsChanged))
   }
+
+  def getLibrariesChanged(seqNum: SequenceNumber[Library], fetchSize: Int) = Action { request =>
+    val libs = db.readOnlyReplica(2) { implicit s => libraryRepo.getBySequenceNumber(seqNum, fetchSize) } map { LibraryView.fromLibrary(_) }
+    Ok(Json.toJson(libs))
+  }
+
+  def getLibraryMembershipsChanged(seqNum: SequenceNumber[LibraryMembership], fetchSize: Int) = Action { request =>
+    val mem = db.readOnlyReplica(2) { implicit s => libraryMembershipRepo.getBySequenceNumber(seqNum, fetchSize) }
+    Ok(Json.toJson(mem))
+  }
 }
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/internal/ShoeboxDataPipeController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/internal/ShoeboxDataPipeController.scala
@@ -165,12 +165,12 @@ class ShoeboxDataPipeController @Inject() (
   }
 
   def getLibrariesChanged(seqNum: SequenceNumber[Library], fetchSize: Int) = Action { request =>
-    val libs = db.readOnlyReplica { implicit s => libraryRepo.getBySequenceNumber(seqNum, fetchSize) } map { LibraryView.fromLibrary(_) }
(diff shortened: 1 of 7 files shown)


kifi-backend/modules/shoebox/app/com/keepit/controllers/internal/ShoeboxDataPipeController.scala
```diff
@@ -163,4 +163,14 @@ class ShoeboxDataPipeController @Inject() (
     }
     Ok(Json.toJson(keepAndTagsChanged))
   }
+
+  def getLibrariesChanged(seqNum: SequenceNumber[Library], fetchSize: Int) = Action { request =>
+    val libs = db.readOnlyReplica(2) { implicit s => libraryRepo.getBySequenceNumber(seqNum, fetchSize) } map { LibraryView.fromLibrary(_) }
```
stkem: Why the `2` here? read only sessions should generally not need retries (If a read session fails we likely have a much bigger problem than if that happens to a write session)

kifi-backend/modules/shoebox/app/com/keepit/controllers/internal/ShoeboxDataPipeController.scala
```diff
@@ -163,4 +163,14 @@ class ShoeboxDataPipeController @Inject() (
...
+    Ok(Json.toJson(libs))
+  }
+
+  def getLibraryMembershipsChanged(seqNum: SequenceNumber[LibraryMembership], fetchSize: Int) = Action { request =>
+    val mem = db.readOnlyReplica(2) { implicit s => libraryMembershipRepo.getBySequenceNumber(seqNum, fetchSize) }
```
stkem: Some here (the `2`). Also, `LibraryMembership` is a db model and should not be sent over the wire either, i.e. we'll need a View for that just like for the library.

stkem:
made two comments, lgtm otherwise

kifi-backend/modules/shoebox/app/com/keepit/controllers/internal/ShoeboxDataPipeController.scala
yingjieMiao: was following the convention. I can remove them all.

kifi-backend/modules/shoebox/app/com/keepit/controllers/internal/ShoeboxDataPipeController.scala
yingjieMiao: as of now, transferring `LibraryMembership` is not a 'waste'. Most of the fields will be used. It's like how we do the keep ingestion. (wouldn't mind to create a view though, if that's the pattern we are adopting now).

kifi-backend/modules/shoebox/app/com/keepit/controllers/internal/ShoeboxDataPipeController.scala
stkem: Yes, I think they should be removed. Unless I'm missing something (cc @andrewconner @ahsu1230)

kifi-backend/modules/shoebox/app/com/keepit/controllers/internal/ShoeboxDataPipeController.scala
stkem: Oh, the point was not waste, but rather to decouple the two things, so that we are free to change the db model without having to change anything on other Services and vice verse.
I know we still send the DB model in a bunch of places, but we should try to move away from that.

yingjieMiao: I see. Let's make this pattern consistent in the future.

kifi-backend/modules/shoebox/app/com/keepit/controllers/internal/ShoeboxDataPipeController.scala
ahsu1230: not that I know of =)

andrewconner: Yeah I think it's fine.

yingjieMiao:
@stkem  

updated. 
created membershipView. 
use toView() instead of from(). I will leave the type issue for now. It's quite cumbersome as is.

stkem:
lgtm

PR #10770, opened 2014-09-09
add keeps (libraries only) on kifi.com [FE]

ahsu1230:
@andrewconner @yipingliao 

numKeeps per library on left panel increases with addKeepsToLibrary()
may still need to work on several other bugs
(keeping urls in different libraries, unkeeping from Library, etc.)

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/keep/addKeep.js
+++ b/kifi-backend/modules/shoebox/angular/src/keep/addKeep.js
@@ -3,8 +3,8 @@
 angular.module('kifi')
 
 .directive('kfAddKeep', [
-  '$document', '$rootScope', '$location', 'keyIndices', 'keepService',
-  function ($document, $rootScope, $location, keyIndices, keepService) {
+  '$document', '$rootScope', '$location', 'keyIndices', 'keepService', 'libraryService',
+  function ($document, $rootScope, $location, keyIndices, keepService, libraryService) {
 
     return {
       restrict: 'A',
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/shoebox/angular/src/keep/addKeep.tpl.html
```diff
@@ -7,11 +7,19 @@
...
+                    value="{{lib.id}}">{{lib.name}}</option>
+          </select>
+        </div>
       </div>
     </div>
   </div>
   <div class="dialog-buttons">
-    <button class="kf-action-button" ng-click="keepUrl()" tabindex="0">Keep it</button>
+    <button class="kf-action-button" ng-if="!librariesEnabled" ng-click="keepUrl()" tabindex="0">Keep it</button>
+    <button class="kf-action-button" ng-if="librariesEnabled" ng-click="keepToLibrary()" tabindex="0">Keep To Library</button>
```
yipingliao: I noticed that this text doesn't show up entirely on the button. Maybe we need to style the button to be longer to accommodate for this?

yipingliao:
Looks good!

kifi-backend/modules/shoebox/angular/src/keep/addKeep.js
```diff
@@ -94,6 +94,32 @@ angular.module('kifi')
...
+          } else {
+            scope.state.invalidUrl = true;
+          }
+        };
+
+        scope.libraries = _.filter(libraryService.librarySummaries, function(lib) {
+          return lib.access !== "read_only";
+        });
+        scope.data = scope.data || {};
+        scope.data.selectedLibraryId = scope.libraries[0].id;
```
andrewconner: May be nice to look for the Main (look up `_.find`)

kifi-backend/modules/shoebox/angular/src/keeps/keepService.js
```diff
@@ -473,6 +473,33 @@ angular.module('kifi')
         return processKeepAction(data);
       },
 
+      keepToLibrary: function (keepUrls, libraryId) {
```
andrewconner, line 476: @yipingliao Is keepService the place to do this? Up to you, I know you're moving networking aspects of keepService around?

kifi-backend/modules/shoebox/angular/src/keep/addKeep.tpl.html
ahsu1230: We could also leave it as just "Keep It"
I don't think we need to explicitly say Keep To Library - I had that just to make sure I was in the library experiment.

andrewconner:
Feel free to merge when ready.

kifi-backend/modules/shoebox/angular/src/keep/addKeep.js
```diff
@@ -94,6 +94,38 @@ angular.module('kifi')
           }
         };
 
+        scope.keepToLibrary = function () {
+          var url = (scope.state.input) || '';
+          if (url && keepService.validateUrl(url)) {
+            var libUrl = _.find(scope.libraries, function(lib) {
+                return lib.id === scope.data.selectedLibraryId
+              }).url;
+            $location.path(libUrl);
```
yipingliao: Maybe not re-route? Stay on currently working view.

kifi-backend/modules/shoebox/angular/src/keeps/keepService.js
yipingliao, line 476: Just discussed with @ahsu1230 . It's good there for now.

kifi-backend/modules/shoebox/angular/src/keep/addKeep.js
andrewconner: Yeah, not sure. It was there before for one case... you think it's better to stay? I'm cool with that.

andrewconner: Also, if we do redirect, probably should _after_ adding it. And the keep specific redirect below should go.

PR #10773, opened 2014-09-09
using dedicated endpoint for logging out from extension

2is10:
@andrewconner


kifi-backend/modules/shoebox/conf/shoebox.routes
```diff
@@ -81,9 +81,9 @@ POST    /tags/:id/addToKeep         @com.keepit.controllers.ext.ExtBookmarksCont
 POST    /tags/clear                 @com.keepit.controllers.ext.ExtBookmarksController.clearTags()
 
 POST    /kifi/start                 @com.keepit.controllers.ext.ExtAuthController.start
-GET     /session/end                @com.keepit.controllers.ext.ExtAuthController.logOut
```
2is10, line 84: this hasn’t been used in ages

kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtAuthController.scala
```diff
@@ -139,14 +139,19 @@ class ExtAuthController @Inject() (
     } getOrElse BadRequest(Json.obj("error" -> "no_such_provider"))
   }
 
-  // where SecureSocial sends users if it can't figure out the right place (see securesocial.conf)
-  def welcome = JsonAction.authenticated { implicit request =>
```
2is10, line 143: this method wasn’t hooked up to a route

andrewconner:
Nice

PR #10774, opened 2014-09-10
splitting ExtDeepLinkController in two

2is10:
@joshm1


2is10:
`com.keepit.controllers.email.EmailDeepLinkController`: public kifi.com routes (from emails)
`com.keepit.controllers.internal.ShoeboxDeepLinkController`: internal routes for other services

`ExtDeepLinkController` was one of our older controllers, predating our internal routes convention, and inappropriately named, since it doesn’t handle any requests from the extension.

andrewconner:
Works for me.

PR #10775, opened 2014-09-10
remove play23/akka23 warning

raymondng:
play23/akka23 update

never mind about the change history -- I tried to rebase and apparently made things worse (?!)

@andrewconner


andrewconner:
Interesting, `BroadcastGroup` vs `RoundRobinGroup`. We want to broadcast to all scraping actors?

raymondng:
For scrape agents, its only (current) use is for broadcast. I was thinking about getting rid of the router all together, but this (appears to be) a simpler change.

andrewconner:
Looks good

andrewconner:
Merging to rid the warnings — since you're out, if there's problems, I'll revert this PR.

PR #10779, opened 2014-09-10
New Recommendation "recompute" logic

stkem:
All it does it trigger another fetch from the graph. Even if there are no new items, it will bump up the sequence number of a bunch of raw seed items causing the regular reco computation logic to pick it up on it's next run.
It places a bit of an additional burden on the graph (and the ingestion logic) but they can handle it quite well right now and it simplifies the recompute logic a lot.

Also renamed it from "reset..." to "refresh..." and added a button to each users admin page to trigger a reco refresh.

@yingjieMiao


yingjieMiao:
lgtm

PR #10780, opened 2014-09-10
adding keeps to library can return which keeps already existed IN a libr...

ahsu1230:
...ary (before, it was keeps USER has kept)


ahsu1230:
@andrewconner

andrewconner:
Great thanks

PR #10782, opened 2014-09-10
Whoops this broke tests

andrewconner:
@ahsu1230 My bad, I won't gut out all your tests to stop the exceptions.


ahsu1230:
ah sorry should've told you about this PR
https://github.com/kifi/keepit/pull/10780 - it should fix those airbrakes

PR #10783, opened 2014-09-10
getting more context into scraper fetch errors

eishay:


eishay:
@raymondng

andrewconner:
Merging, seems low risk :)

PR #10785, opened 2014-09-10
1 retry in the sequence number assigner

stkem:
I think the raw_seed_item table in curator is currently the most
insert/save heavy table in the system. We’ve been seeing the occasional
lock wait timeout exceeded error for the sequence number assigner there
when big batch updates/inserts are going on. I think it’s likely only a
matter of time before this happens elsewhere as well, this I would like
to add one retry to the logic.
Thoughts?

@leogrim @ymatsuda (cc @andrewconner @eishay)

### DIFF
--- a/kifi-backend/modules/sqldb/app/com/keepit/common/db/DbSequenceAssigner.scala
+++ b/kifi-backend/modules/sqldb/app/com/keepit/common/db/DbSequenceAssigner.scala
@@ -14,7 +14,7 @@ abstract class DbSequenceAssigner[M <: ModelWithSeqNumber[M]](
 
   override def assignSequenceNumbers(): Unit = {
     try {
-      while (db.readWrite { implicit session => repo.assignSequenceNumbers(batchSize) } > 0) {}
+      while (db.readWrite(attempts = 2) { implicit session => repo.assignSequenceNumbers(batchSize) } > 0) {}
     } catch {
       case e: UnsupportedOperationException =>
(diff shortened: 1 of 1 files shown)


andrewconner:
Why is there contention on those rows?

Besides that, seems fine.

yingjieMiao:
would batch size matter? e.g. batch of curator updates and batch of the seqNum assigner?

yingjieMiao:
I also noticed that in `TopUriSeedIngestionHelper`, after we get (potentially ~10K) scored uris from graph, we are trying to saving all these scored items in one db session. That could be problematic by itself.  Batch processing them?

ymatsuda:
Looks ok to me.
I am curious why this happens. Is the writer's transaction (update/insert) taking very long?

stkem:
As @yingjieMiao said, the main culprit is the ingestion from the graph, which comes in a possibly very large (i.e. easily several thousand uris) batch. The reason it's in one transaction is that we also keep track of the last time we successfully ingested this data. If we brake it up into multiple transaction the state logic will have to deal with partially completed ingestions, which makes it significantly more complicated. If anyone feels strongly about it I can do that too, but I though this would be the simpler solution given that it's a fairly rare occurrence and won't do any damage elsewhere (and in fact might help when longer transactions creep up in other places)

stkem:
Discussed this some more with @yingjieMiao and decided to close it for now.

PR #10792, opened 2014-09-10
library and membership in fake shoebox client

yingjieMiao:
@stkem  @andrewconner @leogrim


yingjieMiao:
will be used later in unit tests.

PR #10794, opened 2014-09-10
Reco refresh can take some time. No need for the caller to wait for that.

stkem:
It can cause timeouts and there is no response that the caller would use.
@yingjieMiao


yingjieMiao:
thanks

PR #10800, opened 2014-09-10
Kifi black reset password email

joshm1:
This does not update the production email.  I will test it with the new emailSenderTest endpoint first, then update the production template in another PR.

I made no changes to AuthController.forgotPassword, but I went ahead and added a unit/regression test for it, which should continue to pass when I change it to use ResetPasswordEmailSender

@stkem


stkem:
lgtm

PR #10806, opened 2014-09-10
Create libraries frontend

andrewconner:
@yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/layout/main/MainCtrl.js
+++ b/kifi-backend/modules/shoebox/angular/src/layout/main/MainCtrl.js
@@ -142,6 +142,11 @@ angular.module('kifi')
         case 'installExtensionError':
           $scope.modal = 'install_extension_error';
           $scope.data.showInstallErrorModal = true;
+          break;
+        case 'createLibrary':
+          $scope.modal = 'create_library';
+          $scope.data.showCreateLibraryModal = true;
+          break;
       }
     });
 
@@ -178,6 +183,10 @@ angular.module('kifi')
     $scope.cancelImport = function () {
       $window.postMessage('import_bookmarks_declined', '*');
       $scope.data.showImportModal = false;
+    };
+
+    $scope.hideModal = function () {
+      $scope.modal = null;
     };
 
     $scope.disableBookmarkImport = true;
--- a/kifi-backend/modules/shoebox/angular/src/libraries/library.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/library.js
@@ -39,11 +39,6 @@ angular.module('kifi')
         return 'Loading...';
       }
 
-      var subtitle = false; // todo? keepService.getSubtitle($scope.mouseoverCheckAll);
-      if (subtitle) {
-        return subtitle;
-      }
-
       var numShown = $scope.keeps.length;
       switch (numShown) {
       case 0:
--- a/kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.js
@@ -41,15 +41,12 @@ angular.module('kifi')
 
 
     $scope.createLibrary = function () {
-      if ($scope.submitting) {
-        return;
-      }
       var newError = false;
       if ($scope.library.name.length < 3) {
         $scope.$error.name = ' Try a longer name';
         newError = true;
       }
-      if (newError) {
+      if (newError || $scope.submitting) {
         return;
       }
       $scope.submitting = true;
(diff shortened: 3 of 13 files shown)


kifi-backend/modules/shoebox/angular/src/layout/main/MainCtrl.js
```diff
@@ -180,6 +185,10 @@ angular.module('kifi')
       $scope.data.showImportModal = false;
     };
 
+    $scope.hideModal = function () {
```
andrewconner, line 188: It's been a while since I worked with modals, and it seems a bit of a mess — there's both a mutable variable `$scope.modal` and something like `$scope.data.showXModal` for each modal. Closing the modals doesn't correctly unset everything to previous state.

What about a quick port to using https://github.com/likeastore/ngDialog#api instead? Since modals are almost always in their own template file anyways, shouldn't be that hard.

andrewconner, line 188: Or this one: https://github.com/dwmkerr/angular-modal-service (less stars, though)

yipingliao, line 188: Agreed on modals being messy. I started a modal refactoring a few weeks back but haven't had the chance to finish it yet. We actually have quite a few modals that are not in their own template files and depend on their parent scopes. Not sure how ngDialog deals with that but we can take a look.

kifi-backend/modules/shoebox/angular/src/libraries/library.js
```diff
@@ -39,11 +39,6 @@ angular.module('kifi')
         return 'Loading...';
       }
 
-      var subtitle = false; // todo? keepService.getSubtitle($scope.mouseoverCheckAll);
```
yipingliao, line 42: I've been looking into keepService.getSubtitle. :) By the way, I've not found any place where it's actually called with a mouseover. Do you know about this feature?

kifi-backend/modules/shoebox/angular/src/layout/main/MainCtrl.js
andrewconner, line 188: Yeah, scopes are probably important. I see ours makes an interesting mess of the scopes — would be _great_ if the modal's scope just inherited the parent controller (almost always `MainCtrl`)

kifi-backend/modules/shoebox/angular/src/libraries/library.js
andrewconner, line 42: No clue. Couldn't find a use case for this here, so it's out.

kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.js
```diff
@@ -0,0 +1,86 @@
...
+      if ($scope.submitting) {
+        return;
+      }
+      var newError = false;
+      if ($scope.library.name.length < 3) {
+        $scope.$error.name = ' Try a longer name';
+        newError = true;
+      }
+      if (newError) {
+        return;
```
yipingliao, line 50: Can this be moved to after line 50 instead?

kifi-backend/modules/shoebox/angular/src/libraries/library.js
yipingliao, line 42: Great.

kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.js
andrewconner, line 50: Sure, done.

yipingliao:
:+1:

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -78,9 +78,8 @@ class LibraryCommander @Inject() (
 
   def addLibrary(libAddReq: LibraryAddRequest, ownerId: Id[User]): Either[LibraryFail, Library] = {
     val badMessage: Option[String] = {
-      if (!libAddReq.collaborators.intersect(libAddReq.followers).isEmpty) { Some("collaborators & followers overlap!") }
```
andrewconner, line 81: @ahsu1230 This was removed because collaborators aren't a thing yet, and I'd prefer the code to be out so that we deal with this validation holistically when we're ready in a couple months.

kifi-backend/modules/shoebox/angular/src/layout/main/MainCtrl.js
yipingliao, line 188: Actually over 50% of our modals are scattered throughout the templates and inherit from something deeper than `MainCtrl`. I wrote a modal service that takes in the scope of the invoking function and creates a modal using that scope. I really should go finish it. :)

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
ahsu1230, line 81: sure, good idea

PR #10808, opened 2014-09-10
Buildfix part 2, return of the old tests

andrewconner:


andrewconner:
The original was better.

PR #10810, opened 2014-09-10
and yet more logging/reporting on the public feed

stkem:
almost got it narrowed down, I think
@yingjieMiao

### DIFF
--- a/kifi-backend/modules/curator/app/com/keepit/curator/commanders/PublicFeedGenerationCommander.scala
+++ b/kifi-backend/modules/curator/app/com/keepit/curator/commanders/PublicFeedGenerationCommander.scala
@@ -59,16 +59,24 @@ class PublicFeedGenerationCommander @Inject() (
       log.info(s"public feed: saving items")
       items foreach { item =>
         log.info(s"public feed: saving $item")
-        val feedOpt = publicFeedRepo.getByUri(item.uriId, None)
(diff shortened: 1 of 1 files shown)


yingjieMiao:
:+1:

kifi-backend/modules/curator/app/com/keepit/curator/commanders/PublicFeedGenerationCommander.scala
```diff
@@ -59,16 +59,24 @@ class PublicFeedGenerationCommander @Inject() (
...
+              publicAllScores = item.publicUriScores))
+          } getOrElse {
+            publicFeedRepo.save(PublicFeed(
+              uriId = item.uriId,
+              publicMasterScore = computePublicMasterScore(item.publicUriScores),
+              publicAllScores = item.publicUriScores))
+          }
+        } catch {
+          case t: Throwable => {
+            log.info(s"public feed: something went wrong here: $t")
```
andrewconner, line 76: log.error? :) (stack trace for free)

stkem, line 76: that would be better indeed :-)
This was just to track down a weird prod issue (which I think I just found). Sending a PR to remove this as soon as the tests finish locally.

PR #10824, opened 2014-09-11
library feature updater

yingjieMiao:
@stkem 
1. the updater listens on keepRepo change
2. currently, when a library is deleted on shoebox, its state becomes inactive. But keep repo doesn't change. Talked to @ahsu1230 (cc: @andrewconner ), was told that library deletion behavior was not fully decided. This means that when library was removed, the updater will not detect this change. We can easily add a reaper to clean active library features for inactive libraries. (alternatively, we can also listen on libraryRepo's seqNum, but that might be more complicated)

### DIFF
--- a/kifi-backend/modules/cortex/app/com/keepit/cortex/models/lda/LDALibraryUpdater.scala
+++ b/kifi-backend/modules/cortex/app/com/keepit/cortex/models/lda/LDALibraryUpdater.scala
@@ -16,19 +16,6 @@ import com.keepit.model.{ LibraryStates, LibraryKind, Library }
 import com.keepit.cortex.utils.MatrixUtils._
 import org.joda.time.DateTime
 
-class LDALibraryUpdaterActor @Inject() (airbrake: AirbrakeNotifier, updater: LDALibraryUpdater) extends FeatureUpdateActor(airbrake, updater)
-
-trait LDALibraryUpdaterPlugin extends FeatureUpdatePlugin[Library, DenseLDA]
-
-@Singleton
-class LDALibraryUpdaterPluginImpl @Inject() (
-  actor: ActorInstance[LDALibraryUpdaterActor],
-  discovery: ServiceDiscovery,
-  val scheduling: SchedulingProperties) extends BaseFeatureUpdatePlugin(actor, discovery) with LDALibraryUpdaterPlugin
-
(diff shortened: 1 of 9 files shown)


kifi-backend/modules/cortex/app/com/keepit/cortex/models/lda/LDALibraryUpdater.scala
```diff
@@ -0,0 +1,106 @@
...
+        case Some(m) => m.copy(numOfEvidence = feats.size, topic = mean, state = state)
+        case None => LibraryLDATopic(libraryId = libId, version = representer.version, numOfEvidence = feats.size, topic = mean, state = state)
+      }
+      db.readWrite { implicit s => libraryLDARepo.save(toSave) }
+    }
+  }
+
+  private def shouldComputeFeature(libId: Id[Library], model: Option[LibraryLDATopic]): Boolean = {
+    def isApplicableLibrary(lib: Option[CortexLibrary]): Boolean = lib.exists(x => x.state.value == LibraryStates.ACTIVE.value && x.kind != LibraryKind.SYSTEM_MAIN && x.kind != LibraryKind.SYSTEM_SECRET)
+    def isOld(updatedAt: DateTime) = updatedAt.plusMinutes(5).getMillis < currentDateTime.getMillis
```
stkem: Doesn't this suffer form the same problem that we had with user interest profiles, i.e. if any keeps were added in the five minute window since the last computation and then nothing we will never update?

yingjieMiao: true. I thought it is not a big issue. On the other hand, I can remove this guard as well. It was mainly used to defend Super Keepers. For libraries, I don't see a clear issue to recompute every time.

stkem: let's do that then. If it becomes a problem we can always add a guard later.

stkem:
lgtm (minus guard). We can deal with library deletion when thats been fully figured out.

yingjieMiao:
@stkem  guard removed.

andrewconner:
> But keep repo doesn't change

Keeps are marked as inactive (as of today, before was an oversight, not product decision)

yingjieMiao:
thanks @andrewconner

PR #10827, opened 2014-09-11
display library names of every keep card

ahsu1230:
@andrewconner @yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/keep/keep.js
+++ b/kifi-backend/modules/shoebox/angular/src/keep/keep.js
@@ -34,8 +34,8 @@ angular.module('kifi')
 ])
 
 .directive('kfKeep', [
-  '$document', '$rootScope', '$rootElement', '$timeout', 'tagService', 'keepService', 'util',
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/angular/src/keep/keep.js
```diff
@@ -61,6 +61,35 @@ angular.module('kifi')
...
+
+          scope.getLibraryUrl = function() {
+            if (scope.keep.libraryInfo) {
+              return scope.keep.libraryInfo.url;
+            } else {
+              return '/';
+            }
+          };
+
+          scope.$watch('libraries.length', function() {
```
ahsu1230, line 86: as discussed with @yipingliao on loading the screen, the first few keep cards do not have library names displayed because `libraryService.librarySummaries` hasn't returned yet. And so, the target library is not found & nothing is displayed. With the watch, we can fix the first few keep cards once `libraries` populates

yipingliao:
:+1:

PR #10830, opened 2014-09-11
dedup -- ignoreCase

raymondng:
Based on earlier review feedback from @leogrim 

cc: @andrewconner @2is10

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/TypeaheadCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/TypeaheadCommander.scala
@@ -56,7 +56,7 @@ class TypeaheadCommander @Inject() (
       case Some(query) => abookServiceClient.prefixQuery(userId, query, Some(limit)).map { hits => hits.map(_.info) }
       case None => abookServiceClient.getContactsByUser(userId, pageSize = Some(limit))
     }
-    futureContacts map { items => items.dedupBy(_.email.address.toLowerCase).toSeq }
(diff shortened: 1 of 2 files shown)


andrewconner:
That's a cool utility method.

leogrim:
Indeed, but why not reuse the original method from RichContact?

leogrim:
Actually, feel free to replace the one in RichContact with yours :)
:+1:

kifi-backend/modules/shoebox/app/com/keepit/commanders/TypeaheadCommander.scala
```diff
@@ -258,7 +258,7 @@ class TypeaheadCommander @Inject() (
         val kifi: Future[Seq[NetworkTypeAndHit]] = kifiF.map { hits => hits.map(hit => (SocialNetworks.FORTYTWO, hit)) }
         val abook: Future[Seq[NetworkTypeAndHit]] = abookF.map { hits =>
           val nonUsers = hits.filter(_.info.userId.isEmpty)
-          val distinctEmails = nonUsers.groupBy(_.info.email.address).map(_._2.head).toSeq.sortBy(_.score)
+          val distinctEmails = dedupBy(nonUsers) { _.info.email.address.toLowerCase }
```
2is10: Nice!

I think we normally use parentheses when passing a function that doesn’t name its argument, like on the previous line.

raymondng:
Thanks for the feedback! Will revise based on input.

PR #10833, opened 2014-09-11
Managing libraries

andrewconner:
@yipingliao


kifi-backend/modules/shoebox/angular/src/libraries/library.js
```diff
@@ -3,13 +3,28 @@
...
     $scope.keeps = [];
     $scope.library = {};
     var username = $routeParams.username;
     var librarySlug = $routeParams.librarySlug;
 
+    $scope.manageLibrary = function () {
+      libraryService.libraryState = {
+        library: $scope.library,
+        returnAction: function (resp) {
+          libraryService.getLibraryById($scope.library.id, true).then(function (data) {
```
andrewconner, line 17: I need to polish Clutch a bit. The issue here is that `libraryService.getLibraryById` and `libraryService.getLibraryByUserSlug` both return libraries, but since there's no global state for library objects, one returning updated data doesn't update the cache of the other one. This basically calls both on save to make sure the caches are updated.

yipingliao, line 17: Okay, sounds good.

yipingliao:
:+1:

PR #10841, opened 2014-09-11
digest emails to be queued and consumed from SQS

joshm1:
@stkem anything else I need to do for this new queue?  add the retry limit via AWS console after the queue is created?

### DIFF
--- a/kifi-backend/modules/curator/app/com/keepit/curator/commanders/email/EngagementEmailActor.scala
+++ b/kifi-backend/modules/curator/app/com/keepit/curator/commanders/email/EngagementEmailActor.scala
@@ -1,27 +1,31 @@
 package com.keepit.curator.commanders.email
 
-import com.google.inject.{ Inject, Singleton }
-import com.keepit.commanders.RemoteUserExperimentCommander
+import com.google.inject.Inject
 import com.keepit.common.akka.FortyTwoActor
 import com.keepit.common.healthcheck.AirbrakeNotifier
 import com.keepit.common.logging.Logging
-import com.keepit.curator.commanders.RecommendationGenerationCommander
 import play.api.libs.concurrent.Execution.Implicits.defaultContext
 
-object EngagementEmailTypes {
-  object FEED
+object FeedDigestMessage {
+  object Queue
+  object Send
 }
 
 class EngagementEmailActor @Inject() (
-    engagementFeedSender: FeedDigestEmailSender,
+    feedDigestSender: FeedDigestEmailSender,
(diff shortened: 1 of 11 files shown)


kifi-backend/modules/curator/app/com/keepit/curator/commanders/email/EngagementEmailActor.scala
```diff
@@ -1,27 +1,31 @@
...
+  import FeedDigestMessage._
 
-  def receive(): PartialFunction[Any, Unit] = {
-    case FEED => {
-      log.info("calling FeedDigestEmailSender.send() from " + getClass.getName)
-      engagementFeedSender.send()
+  def receive: PartialFunction[Any, Unit] = {
+    case Queue => {
+      log.info("calling FeedDigestEmailSender.send()")
+      feedDigestSender.addToQueue().foreach(_ => self ! Send)
```
joshm1, line 27: @stkem see any issues with doing this?  Send is called every 10 minutes via the scheduler, but this will call it as well immediately after messages are queued

andrewconner, line 27: Could there be a delay? I didn't read this closely, but if you add to a queue and very quickly check, it may miss it? @stkem

stkem, line 27: Ah yes, good point. That is correct. If try to retrieve something before the send request has returned you'll very likely get nothing.

andrewconner, line 27: This does wait until it returns, but immediately requests out — probably 1-2ms later.

stkem, line 27: Hmm, I left a longer message here earlier but it seems to have disappeared. Here it is again: There is a problem with calling it here and in the scheduler. Since `processQueue` is async the single threadedness of the actor doesn't help you, which could cause multiple simultaneous calls to `processQueue`, which in turn (although unlikely) could lead to double sending of emails. A simple solution would be to have a reactive lock in `processQueue` with a concurrency of 1.

andrewconner, line 27: Just keep local state if it should call again, no?

stkem, line 27: Yup, that works too, basically have a local variable (boolean, AtomicBoolean) in the actor that tracks if the queue processing is currently running.

PR #10846, opened 2014-09-12
admin page to show library topic

yingjieMiao:
@stkem

### DIFF
--- a/kifi-backend/modules/common/test/com/keepit/cortex/FakeCortexServiceClientImpl.scala
+++ b/kifi-backend/modules/common/test/com/keepit/cortex/FakeCortexServiceClientImpl.scala
@@ -4,7 +4,7 @@ import com.keepit.cortex.dbmodel.LDAInfo
 
 import scala.concurrent.Future
 import com.keepit.common.db.Id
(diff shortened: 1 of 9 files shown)


kifi-backend/modules/common/test/com/keepit/cortex/FakeCortexServiceClientImpl.scala
```diff
@@ -4,7 +4,7 @@ import com.keepit.cortex.dbmodel.LDAInfo
 
 import scala.concurrent.Future
 import com.keepit.common.db.Id
-import com.keepit.model.{ Keep, User, NormalizedURI, Word2VecKeywords }
+import com.keepit.model._
```
stkem, line 7: I'm not a big fan of `_` imports, especially for such big packages. It makes the code harder to reason about.

stkem:
made one small comment, lgtm otherwise

kifi-backend/modules/common/test/com/keepit/cortex/FakeCortexServiceClientImpl.scala
yingjieMiao, line 7: this is in `FakeCortexService`. will pay attention in the future.

PR #10849, opened 2014-09-12
RIP keepService

yipingliao:
DO NOT MERGE
- I want to squash these commits before merging.
- Since this is a big change, I need to spend more time testing.
- One more known issue needs to be fixed: when an URL is added as a keep, the current view on kifi.com (whether it be the keep stream or a specific library) needs to be updated.

@andrewconner cc @2is10 @ahsu1230 
This is the giant pull request that kills `keepService`, which was responsible for calling the server for keeps and about 20 other things. I have split it up into:
- `keepActionService`
- `keepDecoratorService`
- `searchActionService`
- `selectionService`

In addition, the controllers now keep state for the keeps that are being shown, which was previously kept in `keepService`. The affected controllers are: `HomeCtrl`, `LibraryCtrl`, `SearchCtrl`, `TagKeepsCtrl`, `KeepViewCtrl`, and `HelpRankCtrl`. A couple of things also got moved into `util`.

Library functionality that depended on `keepService` has been migrated over.

Maybe a better way to review this is to sit with me for a few minutes and help out with testing (even 10 minutes of your time would be very good). :)

Thanks!


andrewconner:
If anyone else is having problems to build this, I needed to do this:

```
rm -rf node_modules
npm cache clean -f
npm install
```

May have been a local fluke, but I was having dependency exceptions without that.

kifi-backend/modules/shoebox/angular/src/keep/keepDecoratorService.js
```diff
@@ -0,0 +1,158 @@
...
+          var found = _.find(read_times, function (t) { return minutesEstimate < t; });
+          return found ? found + ' min' : '> 1 h';
+        }
+      }
+
+      // Add new properties to the keep.
+      this.titleAttr = this.title || this.url;
+      this.titleHtml = this.title || util.formatTitleFromUrl(this.url);
+      this.hasSmallImage = this.summary && shouldShowSmallImage(this.summary) && hasSaneAspectRatio(this.summary);
+      this.hasBigImage = this.summary && (!shouldShowSmallImage(this.summary) && hasSaneAspectRatio(this.summary));
```
2is10: To avoid duplication and clarify that a keep has either a small image or a big image, consider:

``` javascript
if (this.summary && hasSaneAspectRatio(this.summary)) {
  this[shouldShowSmallImage(this.summary) ? 'hasSmallImage' : 'hasBigImage'] = true;
}
```

or, even clearer, `.imageSize = 'big'` or `'small'` or `'none'` or something like that.

kifi-backend/modules/shoebox/angular/src/keep/keepDecoratorService.js
```diff
@@ -0,0 +1,158 @@
...
+          }
+        });
+      }
+    };
+
+    // This is needed only for tag drop in tagItem.js.
+    // The reason is that angular.toJSON does not copy over
+    // the Keep prototype methods.:(
+    // TODO(yiping): figure out a more elegant way to do this.
+    function buildKeepForTagDrop(keep, isMyBookmark) {
```
2is10: I think this function could just be a one-liner:

``` javascript
return _.extend(new Keep, keep, {isMyBookmark: isMyBookmark !== false});
```

kifi-backend/modules/shoebox/angular/src/keeps/keepActionService.js
```diff
@@ -9,12 +9,84 @@ angular.module('kifi')
...
+
+      var config = {
+        params: params
+      };
+
+      return keepList.get(url, config).then(function (data) {
+        var result = {};
+
+        result.keeps = data.keeps;
+        result.mayHaveMore = (data.keeps.length > 0) && (data.keeps.length >= params.count - 1);
```
2is10: I think this is equivalent and might be clearer:

``` javascript
result.mayHaveMore = data.keeps.length > Math.max(0, params.count - 2);
```

I’m not clear on why it’s `>= params.count - 1` and not `>= params.count`. If we asked for 3 and got 2, why would we think there might be more?

kifi-backend/modules/shoebox/angular/src/keep/keepDecoratorService.js
yipingliao: Thanks! Will do.

kifi-backend/modules/shoebox/angular/src/keeps/keepActionService.js
2is10: Come to think of it, if we’re concerned that `params.count` might be zero or less, then we should really check for that before making a server request, no?

kifi-backend/modules/shoebox/angular/src/libraries/library.js
```diff
@@ -3,42 +3,71 @@
...
       }
 
+      // If there are selected keeps, display the number of keeps
+      // in the subtitle.
+      if (selectedCount > 0) {
+        switch (selectedCount) {
+          case 0:
+            return null;
+          case 1:
+            return selectedCount + ' Keep selected';
```
2is10: A comment I made earlier applies here too.

kifi-backend/modules/shoebox/angular/src/libraries/library.js
```diff
@@ -51,30 +80,41 @@ angular.module('kifi')
...
+
+    //
+    // On LibraryCtrl initialization.
+    //
+
+    // librarySummaries has a few of the fields we need to draw the library.
+    // Attempt to pre-populate the library object while we wait
+    if (libraryService.librarySummaries) {
+      var lib = _.find(libraryService.librarySummaries, function (elem) {
+        return elem.url === '/' + username + '/' + librarySlug;
```
2is10: would be better to build the path we’re look for only once

kifi-backend/modules/shoebox/angular/src/search/search.js
```diff
@@ -157,73 +85,92 @@ function ($http, $scope, keepService, $routeParams, $location, $window, routeSer
...
-      if (subtitle) {
-        return subtitle;
+      // If there are selected keeps, display the number of keeps
+      // in the subtitle.
+      if (selectedCount > 0) {
+        switch (selectedCount) {
+          case 0:
+            return null;
+          case 1:
+            return selectedCount + ' Keep selected';
```
2is10: again

kifi-backend/modules/shoebox/angular/src/tagKeeps/tagKeeps.js
```diff
@@ -3,82 +3,106 @@
...
+      }
+
+      // If there are selected keeps, display the number of keeps
+      // in the subtitle.
+      if (selectedCount > 0) {
+        switch (selectedCount) {
+          case 0:
+            return null;
+          case 1:
+            return selectedCount + ' Keep selected';
```
2is10: again, wow

2is10:
Very nice!

kifi-backend/modules/shoebox/angular/src/keep/keepDecoratorService.js
2is10: Maybe call it `reconstituteKeep` or `reconstituteKeepFromJson`? The `ForTagDrop` part is specifying who the intended caller is. Better to name solely based on what a function does. Comments are a good place to explain the “why”, when necessary.

yipingliao: Done. Thanks!

kifi-backend/modules/shoebox/angular/src/libraries/library.js
yipingliao: Done. Thanks!

PR #10860, opened 2014-09-12
configurable steps from admin

yingjieMiao:
@stkem 

noticed that the lists gets longer and longer. (and cause too much load on shoebox for DB lookup).


stkem:
:+1:

PR #10867, opened 2014-09-12
[DO NOT MERGE] async -> tryAccept

raymondng:
For review only. This takes care of the last play23 deprecation warning, but it's a behavior change and I don't know what the impact is. Chances are we cannot merge this right now -- nonetheless would appreciate your feedback.

All tests passed; local dev mode tests seem ok.

@stkem @andrewconner @2is10

### DIFF
--- a/kifi-backend/modules/eliza/app/com/keepit/eliza/controllers/AuthenticatedWebSocketsController.scala
+++ b/kifi-backend/modules/eliza/app/com/keepit/eliza/controllers/AuthenticatedWebSocketsController.scala
@@ -156,12 +156,12 @@ trait AuthenticatedWebSocketsController extends ElizaServiceController {
     }
   }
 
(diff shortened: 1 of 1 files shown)


2is10:
I think this was a good API change on Play’s part, but we won’t be able to merge this until I build future compatibility into the extension and most users have it. Otherwise I think we’ll once again run into load crises every time we deploy eliza. We’ll need to give the mobile guys a heads up too.

raymondng:
Thanks @2is10. That sounds about right. Also add @thassss and @jaredpetker.
_Definitely_ not merging right now, but something to keep in mind.

Will close this PR by EOD today. This is FYI.

andrewconner:
We _could_ use `tryAccept`, always accept, and just reply along with the existing API.

That way the deprecation warning is gone, and we have more time for the clients.

raymondng:
Yup. That'll be another PR (want to keep this one the way it is, as a ref)

PR #10871, opened 2014-09-12
Campaign Tracking Phase 1

stkem:
Includes:
- Actually recording of a new users most recent pre signup campaign id (by way of the "kcid" query parameter on all un authenticated requests, which causes setting of a cookie, which is read at signup and stored as a user value if absent)
- Augmentation of all server side events with the campaign id, broken down into segments (separator "-") for easier analysis in mixpanel.

Does not include:
- Augmentation of events that go directly do mixpanel (from kifi.com, the extension and possible mobile). We should probably just bite the bullet and have our own event proxy, but it might end up just being exposed to the clients for them to send directly, depending on time availability.
  (There has been some talk from higher up that we may want to use campaign ids to customize a users experience in a fine grained manner, in which case the clients will likely need it anyway, so I suppose that something to consider as well. I'll do my best to get all this stuff decided before making a call)

@leogrim @andrewconner @raymondng


2is10:
Might consider dropping the "id" everywhere and just call it a Kifi campaign. `"kc"`, `KIFI_CAMPAIGN`, etc. I think it’d still be clear.

stkem:
You have a good point. The reason it has "id" in it all over is that we are planning to also have a `KifiCampaign` type in the not too far future with descriptive information about the campaign.

2is10:
I see. Might still consider `kc=...` for the query string.

PR #10872, opened 2014-09-12
play23 deprecation (async)

raymondng:
Not really taking advantage of `tryAccept` but it does get rid of the deprecation warning.

@andrewconner @2is10 @stkem


andrewconner:
Looks good.

2is10:
:+1:

PR #10874, opened 2014-09-12
hiding model UserSession in shoebox while exposing a smaller representation of it to the outer world

eishay:
Namely, outside the shoebox - class UserSessionView is going to be used instead of UserSession.
UserSession serialization is still supported for backward compatibility until the clients will be bounced as well.

Please review
@leogrim @raymondng @ymatsuda @2is10 @stkem @andrewconner @ahsu1230 @yingjieMiao


kifi-backend/modules/shoebox/conf/shoebox.routes
```diff
@@ -706,6 +706,7 @@ GET     /internal/shoebox/database/getExperiment      @com.keepit.controllers.in
 POST    /internal/shoebox/database/saveExperiment     @com.keepit.controllers.internal.ShoeboxController.saveExperiment
 GET     /internal/shoebox/database/socialUserInfoByNetworkAndSocialId @com.keepit.controllers.internal.ShoeboxController.getSocialUserInfoByNetworkAndSocialId(id: String, networkType: String)
 GET     /internal/shoebox/database/socialUserInfosByUserId @com.keepit.controllers.internal.ShoeboxController.getSocialUserInfosByUserId(id: Id[User])
+GET     /internal/shoebox/database/sessionByExternalId @com.keepit.controllers.internal.ShoeboxController.getSessionViewByExternalId(sessionId: UserSessionExternalId)
 GET     /internal/shoebox/database/sessionByExternalId @com.keepit.controllers.internal.ShoeboxController.getSessionByExternalId(sessionId: ExternalId[UserSession])
```
stkem, line 710: Duplicate route on purpose? (the second one is shadowed)

eishay, line 710: ooops, fixing

kifi-backend/modules/common/app/com/keepit/model/view/UserSessionView.scala
```diff
@@ -0,0 +1,40 @@
...
+
+object UserSessionView {
+  private implicit val idFormat = Id.format[UserSessionRemoteModel]
+  private implicit val userIdFormat = Id.format[User]
+  private implicit val externalIdFormat = ExternalId.format[UserSessionRemoteModel]
+  private implicit val stateFormat = State.format[UserSessionRemoteModel]
+
+  implicit val userSessionFormat: Format[UserSessionView] = (
+    (__ \ 'id).format[Id[UserSessionRemoteModel]] and
+    (__ \ 'externalId).format[ExternalId[UserSessionRemoteModel]] and
```
2is10: I’m not sure I like introducing new remote (fake) model types. The two identifiers above identify objects of type `UserSession`. Once `UserSession` is no longer JSON-serializable, why will it be important to avoid referencing it here?

2is10:
This is great. Looking forward to confining database models to their home services.

kifi-backend/modules/common/app/com/keepit/model/id.scala
```diff
@@ -0,0 +1,15 @@
...
+
+  sealed trait UserSessionRemoteModel
+
+  object Types {
+    type UserSessionId = Id[UserSessionRemoteModel]
+    type UserSessionExternalId = ExternalId[UserSessionRemoteModel]
+    def UserSessionExternalId(id: String) = ExternalId[UserSessionRemoteModel](id)
+  }
+
+}
```
ahsu1230: new line? =0

eishay:
@2is10 I see that we can get away without sending the ids in the view class which makes it even smaller and faster to send around, will commit it soon.
Still, there's the question of using a reference to this object using typed id objects for pulling the model projections from cache or service clients. 
I thought we should hide the models in the service and not make them exposed in the common. If we'll have them in common then over time common will be crowded with models that most services should not care about and the IDEs will suggest using them when they shouldn't. Also, if we make the models viewable by all services it means that everything they may refer to (e.g. some type private to that service) will have to be dragged to the common module though only that service needs to see it.
I'm still looking for alternative ideas...

2is10:
Another option might be to define a `UserSessionId` type that can be easily converted to `Id[UserSession]` (implicitly, perhaps), but that has no direct relation to our `Id` type and isn’t generic, so also has no direct reference to either `UserSession` or a potentially misleading knockoff of `UserSession`.

However we decide to do it, I think the new “public” types we choose probably belong in service-specific packages like `com.keepit.shoebox.model.ids` or something.

eishay:
I like the service-specific packaging, will do.
About `UserSessionId`, its a good idea. 
I used a type like `type UserSessionId = Id[UserSessionRemoteModel]` but I don't think we actually need it to be of type `Id`. I'll try it out.

eishay:
Looking at it a bit more I see that using Id and External Ids does give us some nice things like the `queryStringBinder`, `pathBinder`, json formatters and external id format verifications.

2is10:
True, but I don’t think that justifies using them in a misleading way (writing an identifier of one type as if it were an identifier of a different type). Those are all fixed costs, easy to reimplement or bridge for a new kind of identifier that doesn’t reference the entity type.

eishay:
ok

PR #10878, opened 2014-09-12
Fix bug where keeping an already kept URL does not close the add-keep modal.

yipingliao:
I found a small bug: when a user adds an already-kept URL in the add keep modal, the modal doesn't automatically close. This pull addresses that.

@andrewconner @ahsu1230


andrewconner:
Thanks

PR #10881, opened 2014-09-13
get Page endpoint (now with libraries!)

ahsu1230:
still without otherKeepers (non-friends)

Thanks @leogrim for all the help in search

@andrewconner @2is10 @leogrim


kifi-backend/modules/shoebox/app/com/keepit/commanders/KeeperInfo.scala
```diff
@@ -33,3 +35,53 @@ object KeeperInfo {
...
+    (__ \ 'removable).format[Boolean] and
+    (__ \ 'library).format[LibraryData]
+  )(KeepData.apply, unlift(KeepData.unapply))
+}
+
+case class LibraryData(
+  id: PublicId[Library],
+  name: String,
+  visibility: LibraryVisibility,
+  url: String)
```
2is10, line 78: Would `@json` work here?

kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtPageController.scala
```diff
@@ -42,4 +28,17 @@ class ExtPageController @Inject() (
     }
   }
 
+  def getPage() = JsonAction.authenticatedParseJsonAsync { request =>
+    val url = (request.body \ "url").as[String]
+    if (url.isEmpty) throw new Exception(s"empty url for json ${request.body} for user ${request.user}")
+    URI.parse(url) match {
+      case Success(_) =>
+        pageCommander.getPageInfo(url, request.userId, request.experiments).map { info =>
```
2is10, line 36: If we’ve already done URI parsing, why not pass in the parsed URI instead of the raw string, which `PageCommander` will then have to re-parse? That way we can also eliminate one of the checks for whether the URL is empty.

kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtPageController.scala
```diff
@@ -42,4 +28,17 @@ class ExtPageController @Inject() (
     }
   }
 
+  def getPage() = JsonAction.authenticatedParseJsonAsync { request =>
+    val url = (request.body \ "url").as[String]
+    if (url.isEmpty) throw new Exception(s"empty url for json ${request.body} for user ${request.user}")
```
2is10, line 33: Why treat the empty string differently from any other invalid URL?

kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtPageController.scala
```diff
@@ -42,4 +28,17 @@ class ExtPageController @Inject() (
     }
   }
 
+  def getPage() = JsonAction.authenticatedParseJsonAsync { request =>
```
2is10, line 31: Probably clearer to name this method `getPageInfo()` like the commander method.

kifi-backend/modules/shoebox/app/com/keepit/commanders/PageCommander.scala
```diff
@@ -95,4 +97,63 @@ class PageCommander @Inject() (
     val domain: Option[Domain] = db.readOnlyMaster { implicit s => host.flatMap(domainRepo.get(_)) }
     domain.flatMap(_.sensitive) orElse host.flatMap(domainClassifier.isSensitive(_).right.toOption) getOrElse false
   }
+
+  def getPageInfo(url: String, userId: Id[User], experiments: Set[ExperimentType]): Future[PageInfo] = {
+    if (url.isEmpty) throw new Exception(s"empty url for user $userId")
```
andrewconner, line 102: Same as Jared's comment below

kifi-backend/modules/shoebox/app/com/keepit/commanders/PageCommander.scala
```diff
@@ -95,4 +97,63 @@ class PageCommander @Inject() (
...
+
+            val keeperOpt = keeperId.map(basicUserRepo.load(_))
+            val libOpt = libId.map(libraryRepo.get(_))
+            val libDataOpt = libOpt.map { lib =>
+              val owner = basicUserRepo.load(lib.ownerId)
+              LibraryData(
+                Library.publicId(lib.id.get),
+                lib.name,
+                lib.visibility,
+                Library.formatLibraryUrl(owner.username, owner.externalId, lib.slug))
```
2is10, line 146: Probably not in this pull request, but `Library.formatLibraryUrl` ought to be renamed to `formatLibraryPath` or `formatLibraryUrlPath` since it only returns the path part of a URL.

kifi-backend/modules/shoebox/app/com/keepit/commanders/PageCommander.scala
```diff
@@ -95,4 +97,63 @@ class PageCommander @Inject() (
...
+            }
+            val keepDataOpt = libDataOpt.map(KeepData(keepId, true, true, _))
```
2is10, line 148: This line assumes the owner is the authenticated user, but line 141 above doesn’t (loads the owner from the database). This method ought to be consistent in its assumptions, and if there’s an assumption about ownership, then a TODO comment for eliminating the assumption would be in order. Thanks.

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeeperInfo.scala
ahsu1230, line 78: I was having trouble getting it work... =/

2is10:
Minor comments. Nice work. Thanks, Aaron!

ahsu1230:
Thanks Jared,
Sorry I kind of missed these comments after I merged the PR. But in a moment, I'll be rolling out a new PR with these changes

PR #10882, opened 2014-09-13
curator library membership ingestion

stkem:
@yingjieMiao


kifi-backend/modules/common/conf/evolutions/shoebox/246.sql
```diff
@@ -0,0 +1,21 @@
...
+
+CREATE TABLE curator_library_membership_info (
+    id bigint(20) NOT NULL AUTO_INCREMENT,
+    created_at datetime NOT NULL,
+    updated_at datetime NOT NULL,
+    state varchar(20) NOT NULL,
+
+    user_id bigint(20) NOT NULL,
+    library_id bigint(20) NOT NULL,
+    library_access VARCHAR(64) NOT NULL,
```
yingjieMiao: this `varchar(20)` on shoebox

yingjieMiao:
do you have a test for ingestion ?  lgtm otherwise.

kifi-backend/modules/common/conf/evolutions/shoebox/246.sql
stkem: oh, good find. fixing.

stkem:
test is in the works. Will add one before merging.

PR #10883, opened 2014-09-13
Do not reload raw recommendations unless there is a page refresh.

yipingliao:
@stkem Currently there is a bug with recommendations front end where an Angular re-route back to /recommendations reloads the raw recommendations we have previously received from the server. While there is no extra call to the server, what this means is that user changes on the client side to the recommendations (trashing, keeping) are not preserved.

I made a new service called `recoStateService` to store state about the client side recommendations, so that routing within the Angular app will maintain those changes.

Thanks to @ahsu1230 @jaredpetker and @andrewconner  for looking at that subtle JS bug with me. :)


stkem:
:+1:

PR #10884, opened 2014-09-14
Uri summary burst

eishay:
trying to understand what is the source of the calls to this client service api.
we seem to call in in bursts that makes the system timeout on us and create rippling effects. 
@stkem please review


stkem:
lgtm

PR #10885, opened 2014-09-15
tracing stack trace of all service clients. for some reason the onFailur...

eishay:
...e that the caller to the call method is doing was not triggered.
The change I did yesterday did not catch the exceptions full trace we still have in prod all the time.
@ymatsuda @stkem


stkem:
lgtm

PR #10895, opened 2014-09-15
removing repo of keeper rules for auto-engagement, retaining do-not-engage URL patterns

2is10:
@andrewconner


2is10:
By “engagement” I mean the automatic showing of the friends-who-kept callout above the keeper.

Before this pull request, there were only two rules remaining:
- _Don’t engage if the keeper has already been shown recently on this page._ We’ll never turn this rule off because it would make Kifi way too annoying. (That’s why we built this functionality.)
- _Don’t engage if a page’s URL matches a list of patterns._ The list is already configurable, so the ability to toggle this rule isn’t buying us any real flexibility.

So I’m removing this rule functionality because it isn’t serving any real purpose and hasn’t been used forever.

2is10:
Merging, happy to discuss in person if you’d like.

PR #10901, opened 2014-09-15
adding IDs of main and secret libraries to /kifi/start response

2is10:
@andrewconner 
also removing long-unused `"joined"` field from the response


andrewconner:
Works for me

PR #10904, opened 2014-09-15
Keep image (you can merge)

andrewconner:
The next generation of the keeper will let the user select which image to use for a keep.

Requirements for new keep image:
- Two keeps of the same URL can have different images
- Images should be sized to several standard sizes
- Once uploaded, an image should never change. As such, it will have a "cache forever" header.

The basic flow for fetching an image:
1. Receive image URL from user or make request to external services
2. Hash the image file. This can be seen as its "key".
3. Check if we already have KeepImage records for that hash. If yes, skip to 7. If not:
4. Verify that the image indeed is an image that we can read.
5. Make several size variations of the original.
6. Upload the original and all size variations to S3, keyed by the hash and their dimensions.
7.  Create `KeepImage` records for every size for this keep.

Opening this up for feedback. @stkem @raymondng @eishay @ahsu1230 

Left to do:
- Any copying of a keep should also duplicate KeepImage records


kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepImageCommander.scala
```diff
@@ -0,0 +1,492 @@
...
+    while (len > 0) {
+      digest.update(buffer, 0, len)
+      len = is.read(buffer)
+    }
+    val md5sum = digest.digest()
+
+    is.close()
+
+    val bigInt = new BigInteger(1, md5sum)
+    var bigIntStr = bigInt.toString(16)
```
2is10: Okay if we use radix 36 instead?

andrewconner: Yes. I used 16 so that it's very easy to calculate the filename manually from the command like: `md5 file.png`

2is10: okay, whatever

kifi-backend/modules/common/app/com/keepit/common/store/S3AsyncHelper.scala
```diff
@@ -0,0 +1,64 @@
...
+
+  def s3Client: AmazonS3
+
+  def asyncUpload(bucketName: String, key: String, is: InputStream, objectMetadata: ObjectMetadata)(implicit tm: TransferManager): Future[UploadResult] = {
+    val p = Promise[UploadResult]()
+    val upload = tm.upload(bucketName, key, is, objectMetadata)
+    upload.addProgressListener(new ProgressListener {
+      override def progressChanged(p1: ProgressEvent): Unit = {
+        if (!p.isCompleted) {
+          if (p1.getEventCode == ProgressEvent.COMPLETED_EVENT_CODE) {
```
yingjieMiao: I would prefer pattern matching here and below.

kifi-backend/modules/common/conf/evolutions/shoebox/247.sql
```diff
@@ -0,0 +1,26 @@
...
+CREATE TABLE keep_image (
+  id bigint(20) NOT NULL AUTO_INCREMENT,
+  created_at datetime NOT NULL,
+  updated_at datetime NOT NULL,
+  state varchar(20) NOT NULL,
+  keep_id bigint(20) NOT NULL,
+  image_url varchar(92) NOT NULL,
+  image_format varchar(16) NOT NULL,
+  width int NOT NULL,
+  height int NOT NULL,
```
yingjieMiao: `small int unsigned`, for width and height?

kifi-backend/modules/common/conf/evolutions/shoebox/247.sql
```diff
@@ -0,0 +1,26 @@
...
+  source varchar(32) NOT NULL,
+  source_file_hash varchar(32),
+  source_image_url varchar(3072),
+  is_original boolean,
+
+  PRIMARY KEY (id),
+  INDEX keep_image_f_keep_id (keep_id),
+  UNIQUE INDEX keep_image_u_source_file_hash_size_keep_id (source_file_hash, width, height, keep_id)
```
yingjieMiao: would put keepId first in the 4-tuple give better performance?

stkem: At the very least, that would obviate the need for a separate index in keep_id, but you would lose the ability to look things up efficiently just by hash and size.

andrewconner: Looking up by hash will be common.

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepImageCommander.scala
```diff
@@ -0,0 +1,492 @@
...
+    }
+    if (existingImagesForKeep.nonEmpty && !overwriteUserSelectedImage) {
+      Future.successful(ImageProcessState.ExistingStoredImagesFound(existingImagesForKeep))
+    } else {
+      fetcher.flatMap {
+        case Right(loadedImage) =>
+          val existingSameHash = db.readOnlyReplica { implicit session =>
+            keepImageRepo.getBySourceHash(loadedImage.hash)
+          }
+          if (existingSameHash.isEmpty) {
```
yingjieMiao: I feel the code might be more readable if you do a pattern matching on the tuple `(existingSameHash, forThisKeep)`

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepImageCommander.scala
```diff
@@ -0,0 +1,492 @@
...
+    }
+  }
+
+  private def createDerivatesUploadAndSave(loadedImage: ImageProcessState.ImageLoadedAndHashed, keepId: Id[Keep], source: KeepImageSource, overwriteUserSelectedImage: Boolean): Future[ImageProcessDone] = {
+    buildPersistSet(loadedImage) match {
+      case Right(toPersist) =>
+        val uploads = toPersist.map { image =>
+          log.info(s"Persisting ${image.key} (${image.bytes} B)")
+          keepImageStore.put(image.key, image.is, image.bytes, imageFormatToMimeType(image.format)).map { r =>
+            image.is.close()
```
yingjieMiao: I feel like low-level operation like closing an `inputStream` should be done in the `put()` method

kifi-backend/modules/common/conf/evolutions/shoebox/247.sql
```diff
@@ -0,0 +1,26 @@
...
+
+# --- !Ups
+
+CREATE TABLE keep_image (
+  id bigint(20) NOT NULL AUTO_INCREMENT,
+  created_at datetime NOT NULL,
+  updated_at datetime NOT NULL,
+  state varchar(20) NOT NULL,
+  keep_id bigint(20) NOT NULL,
+  image_url varchar(92) NOT NULL,
```
yingjieMiao: not clear about the size `92` here. Is it derived from any existing column, or gauranteed by some code logic?

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepImageCommander.scala
```diff
@@ -0,0 +1,492 @@
...
+            case uploadedImage =>
+              // todo isOriginal
+              val ki = KeepImage(keepId = keepId, imageUrl = uploadedImage.key, format = uploadedImage.format, width = uploadedImage.image.getWidth, height = uploadedImage.image.getHeight, source = source, sourceImageUrl = loadedImage.sourceImageUrl, sourceFileHash = loadedImage.hash, isOriginal = false)
+              uploadedImage.image.flush()
+              ki
+          }
+          val existingImagesForKeep = db.readOnlyMaster { implicit session =>
+            keepImageRepo.getAllForKeepId(keepId)
+          }
+          if (existingImagesForKeep.isEmpty || overwriteUserSelectedImage) {
```
yingjieMiao: again, pattern matching on the tuple would give much more readable code and structure, IMHO.

kifi-backend/modules/common/conf/evolutions/shoebox/247.sql
andrewconner: This is our CDN url. We control it. An example: http://d1scct5mnc9d9m.cloudfront.net/keep/57e396baedfe1a034590339082b9abce_538x190_o.png

I can make it longer, but I gave us several extra characters in case cloudfront extends their URLs.

yingjieMiao: thanks! :+1:

kifi-backend/modules/common/conf/evolutions/shoebox/247.sql
```diff
@@ -0,0 +1,26 @@
+# SHOEBOX
+
+# --- !Ups
+
+CREATE TABLE keep_image (
```
raymondng: does this supercede `image_info`? how are the two related?

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepImageCommander.scala
```diff
@@ -0,0 +1,492 @@
...
+import play.api.libs.concurrent.Execution.Implicits.defaultContext
+import play.api.libs.iteratee.Iteratee
+import play.api.libs.ws.WS
+import com.keepit.common.core._
+
+import scala.concurrent.Future
+import scala.util.{ Failure, Success, Try }
+
+@ImplementedBy(classOf[KeepImageCommanderImpl])
+trait KeepImageCommander {
```
ahsu1230, line 27: Just curious, why use a trait to define the commander as opposed to just using class?

kifi-backend/modules/common/conf/evolutions/shoebox/247.sql
andrewconner: Sort of, maybe. Tangential, for now. Keeps will have user-selected images. Not sure what messages and stuff will have.

kifi-backend/modules/common/conf/evolutions/shoebox/247.sql
andrewconner: Thanks.

kifi-backend/modules/common/conf/evolutions/shoebox/247.sql
raymondng: A bit concerned as the two are _sufficiently similar_ that might generate confusion. Is it possible to leverage/enhance `image_info` to achieve what is desired without causing breakage?

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepImageCommander.scala
andrewconner, line 27: 1. Locking down the API. If everyone is injecting the trait, then accidental public methods don't cause a re-compile. This includes locking down the type signature of methods.
2. To serve as a table of contents, of sort. It's a large class, and it's easier to see this way what's available to use.
3. We can mock it easier.

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepImageCommander.scala
```diff
@@ -0,0 +1,492 @@
...
+  }
+
+}
+
+sealed trait ImageProcessState
+sealed trait ImageProcessDone extends ImageProcessState
+sealed trait KeepImageStoreInProgress extends ImageProcessState
+sealed abstract class KeepImageStoreFailure(reason: String) extends ImageProcessState with ImageProcessDone
+sealed abstract class KeepImageStoreFailureWithException(ex: Throwable, reason: String) extends KeepImageStoreFailure(reason)
+object ImageProcessState {
```
yingjieMiao: wow! 12 states.

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepImageCommander.scala
ahsu1230, line 27: I see. Should this be the convention for new commanders in the future?

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepImageCommander.scala
```diff
@@ -0,0 +1,492 @@
...
+      }
+    }
+  }
+
+  @inline
+  private def maxDivergence(s1: ImageSize, s2: ImageSize): Int = {
+    Math.max(Math.abs(s1.height - s2.height), Math.abs(s1.width - s2.width))
+  }
+}
+
```
yingjieMiao, line 364: please consider splitting this class into smaller classes, if makes sense. It should be helpful for unit tests and future maintenance.

kifi-backend/modules/common/conf/evolutions/shoebox/247.sql
andrewconner: Not easily / in a beneficial way. There are a few issues with the current implementation, and even if it was meeting our current needs, it's URI centric — this is keep (or, image file) centric.

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepImageCommander.scala
raymondng, line 27: :+1:  for using traits to lock down API

If it makes sense, we could also rename `KeepImageCommander` trait to something else (say `KeepImageAPI`) and then we simply have the impl be `KeepImageCommander`

kifi-backend/modules/common/conf/evolutions/shoebox/247.sql
2is10: How about storing just the path in this column?

We’ll often need to use `https:`. And we might conceivably change hosts. And repeating the same URL prefix in each row seems a little wasteful.

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepImageCommander.scala
andrewconner, line 27: Yeah, in the future, using a trait would be great. It _really_ speeds up compile times, because it means `scalac` has a lot less type inference to do. Other big improvements are explicitly labeling types and using private methods on anything that isn't accessed outside the class. Easy to forget in Scala.

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepImageCommander.scala
```diff
@@ -0,0 +1,513 @@
...
+          val tempFile = TemporaryFile(prefix = "remote-file", suffix = "." + formatOpt.get.value)
+          val outputStream = new FileOutputStream(tempFile.file)
+
+          val maxSize = 1024 * 1024 * 16
+
+          var len = 0
+          val iteratee = Iteratee.foreach[Array[Byte]] { bytes =>
+            len += bytes.length
+            if (len > maxSize) { // max original size
+              throw new Exception(s"Original image too large (> $len bytes): $imageUrl")
```
andrewconner: This is possibly to do pure-functionally. Here is my implementation:

``` scala
  private def foldT[E, A](zero: A)(f: (A, E) => Option[A]): Iteratee[E, Option[A]] = {
    def step(incr: Option[A])(i: Input[E]): Iteratee[E, Option[A]] = i match {
      case Input.EOF | Input.Empty => Done(incr, Input.EOF)
      case Input.El(e) =>
        if (incr.isDefined) {
          val run = f(incr.get, e)
          if (run.isDefined) {
            Cont[E, Option[A]](i => step(run)(i))
          } else {
            Done(None, Input.EOF)
          }
        } else {
          Done(None, Input.EOF)
        }
    }
    Cont[E, Option[A]](i => step(Some(zero))(i))
  }
    // ... then use it:
    val iteratee = foldT[Array[Byte], Int](0) { case (len, bytes) =>
    val lenN = len + bytes.length
    if (lenN > 10000) { // max original size
      None
    } else {
      outputStream.write(bytes)
      Some(lenN)
    }
  }
```

I felt this was less clear and not as maintainable. Thoughts?

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepImageCommander.scala
```diff
@@ -0,0 +1,513 @@
...
+    }.filter(_.isDefined).map(_.get)
+
+    sizes.foldRight(Set[Int]()) {
+      case (b, a) =>
+        a.find(existingSize => Math.abs(b - existingSize) < 100) match {
+          case Some(alreadyHasACloseEnoughSize) => a
+          case None => a + b
+        }
+    }.filterNot(i => i == Math.max(image.getWidth, image.getHeight)) // throw out original size
+  }
```
yingjieMiao: `n^2` complexity? how about sorting first? more efficient, and might give shorter code? In addition, current code's behavior depends on the order of `sizes`.

yingjieMiao: also, you might want to throw away original sizes first. Cause otherwise, a close enough candidate may not be considered (b/c of original size was there). After you throw away original sizes, you can have a gap.

andrewconner: I could, but the code would be more complex, no? There will only ever be a few sizes (~4).

Order mattering is intentional.

andrewconner: Not throwing away original size first is intentional, because in addition to these sizes, we upload an image with the original dimensions (filling in the gap)

yingjieMiao:
```
var t = -100
sizes.flatMap{ x => if ( x - t > 100) { t = x; Some(x)} else None}
```

assuming sorted. (is this somewhat equivalent?)
Agree with what you said, complexity might not be an issue here.

kifi-backend/modules/common/app/com/keepit/common/store/KeepImageStore.scala
```diff
@@ -0,0 +1,47 @@
...
+class KeepImageStoreImpl @Inject() (
+    s3ImageConfig: S3ImageConfig,
+    val s3Client: AmazonS3,
+    implicit val transferManager: TransferManager) extends KeepImageStore with S3AsyncHelper {
+
+  def put(key: String, is: InputStream, contentLength: Int, mimeType: String): Future[UploadResult] = {
+    val om = new ObjectMetadata()
+    om.setContentType(mimeType)
+    if (contentLength > 0) {
+      om.setCacheControl("public, max-age=31556926")
```
stkem: I know this is standardized, but a comment here as to the time unit used (and possibly a human readable translation) could be quite helpful for future readers.

kifi-backend/modules/common/app/com/keepit/common/store/S3AsyncHelper.scala
stkem: +1, that will also avoid calling `getEventCode` (a method) multiple times. 
Also, the aws docs say `getEventCode` is deprecated and you should use `getEventType`

kifi-backend/modules/common/app/com/keepit/common/store/S3AsyncHelper.scala
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  def asyncDownload(bucketName: String, key: String)(implicit tm: TransferManager): Future[TemporaryFile] = {
+    val p = Promise[TemporaryFile]()
+    val tf = TemporaryFile()
+    val transfer = Try {
+      val download = tm.download(bucketName, key, tf.file)
+      download.addProgressListener(new ProgressListener {
+        override def progressChanged(p1: ProgressEvent): Unit = {
+          if (p1.getEventCode == ProgressEvent.COMPLETED_EVENT_CODE) {
```
stkem: same as above, `match` on `p1.getEventType`?

kifi-backend/modules/common/app/com/keepit/common/store/S3AsyncHelper.scala
andrewconner: Our version of the AWS lib does not have `getEventType`

kifi-backend/modules/common/conf/evolutions/shoebox/247.sql
stkem: Small thing, but if the length of the field is relatively constant, a `char(??)` will be significantly more efficient than a `varchar(??)`.

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepImageCommander.scala
stkem, line 27: +1 on the pattern and on possibly different naming. I've never been a big fan of the `*Impl` scheme.

andrewconner, line 27: Consumers will never see it. We can name it whatever, as long as no one tries to use it directly.

kifi-backend/modules/common/app/com/keepit/common/store/S3AsyncHelper.scala
stkem: :-( Maybe we should upgrade?

raymondng: +1. If it's straightforward.

Upgrading can be tricky though esp. when additional dependencies come into play (pun not intended)

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepImageCommander.scala
stkem: Very much _not_ worth it! I don't like the notion that functional programming is somehow inherently better. It leads to hard to understand (and often slow code). The imperative way is indeed clearly superior here, imho.

kifi-backend/modules/shoebox/conf/shoebox.routes
```diff
@@ -1,4 +1,7 @@
 
+POST     /test1                      @com.keepit.controllers.ext.ExtKeepImageController.autoFetch()
```
stkem: nice routes

kifi-backend/modules/common/conf/evolutions/shoebox/247.sql
raymondng: Good point. How about using `image_info` as source of truth for images associated with a URI (+ sizing, etc.) and then `keep_image` would keep pointers to it?

andrewconner:
I would like to merge this tomorrow morning and start the migration slowly, especially since this PR is now in the "massive" category. Tests and the initial migration code have been added.

PR #10905, opened 2014-09-15
compute library induced user interest

yingjieMiao:
@stkem 

created an endpoint to test against production database. Once this becomes mature, I can add a third field into the current `LDAUserURIInterestScores`.


stkem:
lgtm

PR #10911, opened 2014-09-15
added extension endpoint to unkeep from library

ahsu1230:
also added new commander function in KeepsCommander for unkeeping for libraries because original
1) doesn't check library memberships (can unkeep if owner or RW access)
2) assumed the user unkeeping is the same as owner of the keep (who is also owner of the library). 

includes tests
@2is10 @andrewconner


kifi-backend/modules/common/app/com/keepit/model/LibraryMembership.scala
```diff
@@ -26,6 +26,10 @@ case class LibraryMembership(
   def withState(newState: State[LibraryMembership]): LibraryMembership = this.copy(state = newState)
 
   override def toString: String = s"LibraryMembership[id=$id,libraryId=$libraryId,userId=$userId,access=$access,state=$state]"
+
+  def hasWriteAccess(): Boolean = {
```
2is10: no side effects, so I think the convention is to omit parentheses

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
```diff
@@ -458,6 +459,26 @@ class KeepsCommander @Inject() (
...
   }
 
+  def unkeepFromLibrary(libId: Id[Library], extId: ExternalId[Keep], userId: Id[User])(implicit context: HeimdalContext): Either[String, KeepInfo] = {
+    db.readOnlyMaster { implicit session =>
+      libraryMembershipRepo.getWithLibraryIdAndUserId(libId, userId)
+    } match {
+      case Some(mem) if mem.hasWriteAccess() =>
+        db.readWrite { implicit session =>
+          val ownerId = libraryRepo.get(libId).ownerId
+          keepRepo.getByExtIdAndUser(extId, ownerId) match {
```
2is10: It seems pointless to look up the owner ID just to look up the keep. I suspect `.getByExtIdAndUser` was meant to help cross-check that a user owns a keep before libraries. How about adding a new `KeepRepo.getByExtIdAndLibraryId` method?

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
```diff
@@ -458,6 +459,26 @@ class KeepsCommander @Inject() (
...
+            case Some(keep) if libId == keep.libraryId.get =>
+              // report who unkept
+              setKeepStateWithSession(keep, KeepStates.INACTIVE, userId)
+              Right(finalizeUnkeeping(Seq(keep), userId).head)
```
2is10: `finalizeUnkeeping` looks like it doesn’t require the database, and so probably ought to be outside the database session.

kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtLibraryController.scala
```diff
@@ -4,10 +4,11 @@ import com.google.inject.Inject
 import com.keepit.commanders.{ KeepsCommander, RawBookmarkRepresentation, LibraryCommander }
 import com.keepit.common.controller.{ ShoeboxServiceController, BrowserExtensionController, ActionAuthenticator }
 import com.keepit.common.crypto.{ PublicId, PublicIdConfiguration }
+import com.keepit.common.db.ExternalId
 import com.keepit.common.db.slick.Database
 import com.keepit.common.social.BasicUserRepo
 import com.keepit.heimdal.HeimdalContextBuilderFactory
-import com.keepit.model.{ LibraryMembershipRepo, KeepSource, Library, LibraryAccess }
+import com.keepit.model._
```
2is10: I like seeing the types here. Okay if we add them back?

andrewconner:
Mind making it take a set of keeps? I can see us needing that version soon.

KeepsCommander is getting out of control (not your fault). `unkeepMultiple`, `unkeepBulk`, `unkeepBatch`, `unkeep`, `unkeepFromLibrary`. I think we need to consolidate. All the other unkeeps should become library-aware, right?

kifi-backend/modules/shoebox/test/com/keepit/controllers/ext/ExtLibraryControllerTest.scala
```diff
@@ -159,5 +161,69 @@ class ExtLibraryControllerTest extends Specification with ShoeboxTestInjector wi
...
+        status(result2) must equalTo(BAD_REQUEST)
+        db.readOnlyMaster { implicit s => keepRepo.getByLibrary(lib1.id.get, 10, 0).map(_.title).flatten === Seq("NotChoke", "Throw") }
+
+        // test unkeep from someone else's library (have RW access)
+        inject[FakeActionAuthenticator].setUser(user2)
+        val request3 = FakeRequest("POST", com.keepit.controllers.ext.routes.ExtLibraryController.removeKeep(pubId1, keep3.externalId).url)
+        val result3 = extLibraryController.removeKeep(pubId1, keep3.externalId)(request3)
+        status(result3) must equalTo(OK)
+        contentType(result3) must beSome("application/json")
+        db.readOnlyMaster { implicit s => keepRepo.getByLibrary(lib1.id.get, 10, 0).map(_.title).flatten === Seq("Throw") }
```
2is10, line 246: nice test cases

ahsu1230:
Right now debating on unkeeping multiple keeps from multiple libraries
or unkeeping multiple keeps from single library.
Thoughts?

kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtLibraryController.scala
```diff
@@ -63,4 +64,22 @@ class ExtLibraryController @Inject() (
...
+      case Failure(ex) =>
+        BadRequest(Json.obj("error" -> "invalid_library_id"))
+      case Success(libraryId) =>
+        implicit val context = heimdalContextBuilder.withRequestInfoAndSource(request, KeepSource.keeper).build
+        keepsCommander.unkeepFromLibrary(Seq(keepExtId), libraryId, request.userId) match {
+          case Left(failMsg) => BadRequest(Json.obj("error" -> failMsg))
+          case Right((unkeptKeeps, failures)) =>
+            Ok(Json.obj(
+              "unkept" -> Json.toJson(unkeptKeeps),
+              "failed" -> Json.toJson(failures)
```
2is10, line 80: The fact that the server implements “unkeep one” using “unkeep many” has leaked into the response JSON format. It shouldn’t leak into the controllers or the response. It should be entirely encapsulated within the commander.

ahsu1230, line 80: Ah you're right - my bad. Anything in particular this `removeKeep` function should respond with?

2is10, line 80: It’s okay, not a problem. In the success case, a 204 (`NoContent`) would be fine. A 400, 403 (not authorized), or 404 (not found) with `{"error":"description_code"}` would be good if the request is bad. Thanks!

PR #10913, opened 2014-09-15
"fake" library reco data

stkem:
actual logic will stay the same, but the list of libraries will change
@yingjieMiao @yipingliao

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/curator/model/RecommendationInfo.scala
+++ b/kifi-backend/modules/common/app/com/keepit/curator/model/RecommendationInfo.scala
@@ -1,10 +1,13 @@
 package com.keepit.curator.model
 
 import com.keepit.common.db.{ Id, ExternalId }
-import com.keepit.model.{ NormalizedURI, User, URISummary }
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/common/app/com/keepit/curator/model/RecommendationInfo.scala
```diff
@@ -37,17 +41,42 @@ object RecoKind {
...
+  numKeeps: Int) extends RecoItemInfo
 
-@json case class FullRecoInfo(
+object RecoItemInfo {
+  implicit val writes = new Writes[RecoItemInfo] {
+    def writes(obj: RecoItemInfo) = obj match {
+      case uri: UriRecoItemInfo => Json.toJson(uri)
+      case lib: LibRecoItemInfo => Json.toJson(lib)
+    }
+  }
```
yingjieMiao, line 71: what about `reads`? How would that be defined?

stkem, line 71: It doesn't need to be. This is a server->client data format only.

PR #10914, opened 2014-09-15
email placeholders can be applied to the "subject" and "fromName"

joshm1:
@stkem 

It's common that we want an email's fromName to be "first last (via Kifi)".  I created a short method to default to this fromName with only a user id.
- `EmailToSend.fromName` is now `Option[Either[Id[User], String]]`.  instead of `Option[String]`.

If `Left(id)` is used, it will be preprocessed as "<% fullName(id) %> (via Kifi)" and then feed through the template processor

`Left(id)` is just shorthand for `Right(fullName(id) + " (via Kifi)")`

If `Right(str)` is used, it will still go through the template processor and any tags will be replaced

It was also common to put first and last names in the subject line. So I made this work with tags as well.

I was able to drop the database dependency from several email senders using these updates.


stkem:
lgtm

PR #10916, opened 2014-09-16
processKifiHit bug fix

raymondng:
Revisited the code again and this issue (db scoping) jumped out -- I missed it while porting this API from shoebox. My bad.

Can't say for sure, but it's possible that this is related to heimdal connection leakage issue.

@andrewconner @stkem @yingjieMiao @eishay


stkem:
oh, good candidate

raymondng:
Will push and see if this helps.

PR #10921, opened 2014-09-16
increase econtact cache ttl to 30 days

eishay:
@raymondng @leogrim  increased cache ttl since miss rate is pretty high - h67059,m19357,s19838


raymondng:
lgtm

PR #10922, opened 2014-09-16
moving the class definition of FriendRequest to shoebox, keeping backwards compatibility until the clients are updated.

eishay:
like with other models we have in the common module, FriendRequest should not have been there. Other services are getting the entire object of seven fields when they actually need only the recipientId of active objects.

@yingjieMiao @leogrim @raymondng @andrewconner


yingjieMiao:
lgtm

kifi-backend/modules/common/app/com/keepit/shoebox/ShoeboxServiceClient.scala
```diff
@@ -86,7 +86,7 @@ trait ShoeboxServiceClient extends ServiceClient {
   def getDeepUrl(locator: DeepLocator, recipient: Id[User]): Future[String]
   def getNormalizedUriUpdates(lowSeq: SequenceNumber[ChangedURI], highSeq: SequenceNumber[ChangedURI]): Future[Seq[(Id[NormalizedURI], NormalizedURI)]]
   def getHelpRankInfos(uriIds: Seq[Id[NormalizedURI]]): Future[Seq[HelpRankInfo]]
-  def getFriendRequestsBySender(senderId: Id[User]): Future[Seq[FriendRequest]]
+  def getFriendRequestsRecipientIdBySender(senderId: Id[User]): Future[Seq[Id[User]]]
```
yingjieMiao, line 92: minor: `RecipientIds`, plural form?

PR #10927, opened 2014-09-16
check starting vertex in WanderingCommander

yingjieMiao:
@leogrim @stkem 

this is a cheap solution before we go to job queue...

### DIFF
--- a/kifi-backend/modules/graph/app/com/keepit/graph/wander/WanderingCommander.scala
+++ b/kifi-backend/modules/graph/app/com/keepit/graph/wander/WanderingCommander.scala
@@ -5,7 +5,6 @@ import com.keepit.graph.manager.GraphManager
 import com.keepit.graph.model._
 import com.keepit.common.logging.Logging
 import com.keepit.common.time._
-import com.keepit.graph.simple.Vertex
 import scala.concurrent.Future
 import com.keepit.common.service.RequestConsolidator
 import scala.concurrent.duration._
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/graph/app/com/keepit/graph/wander/WanderingCommander.scala
```diff
@@ -50,7 +50,11 @@ class WanderingCommander @Inject() (graph: GraphManager, clock: Clock) extends L
         val wanderer = reader.getNewVertexReader()
         val scout = reader.getNewVertexReader()
         val scoutingWanderer = new ScoutingWanderer(wanderer, scout)
-        scoutingWanderer.wander(wanderlust.steps, teleporter, resolver, journal)
+        if (wanderer.hasVertex(startingVertexId)) {
+          scoutingWanderer.wander(wanderlust.steps, teleporter, resolver, journal)
+        } else {
+          log.error(s"trying to start a random walk from non-existing vertex: ${startingVertexKind}: ${wanderlust.startingVertexDataId}. Return empty journal now.")
```
stkem, line 56: Could returning an empty result be problematic with the PYMK caching?

yingjieMiao, line 56: hmmm, I assume PYMK uses `SocialWanderingCommander`. This fix should only affect uriWandering problem. But this is a good point.

stkem, line 56: oh, I think you are right. Just thought I'd bring it up. Can you confirm @leogrim

leogrim, line 56: Yep, @yingjieMiao is right :)

yingjieMiao:
thanks for the feedback. merge this, hope to reduce some production noise.

PR #10928, opened 2014-09-16
Implement basic UI for sharing libraries.

yipingliao:
This is not according to the design wireframes yet, because they are still not finalized. However, I wanted to get the basic functionality coded up; this will allow us (people in the library experiment) to actually invite people through the website.

@andrewconner cc @2is10 
https://app.asana.com/0/15523651812135/15523601193412


andrewconner:
Wahoo!

kifi-backend/modules/shoebox/angular/src/libraries/library.js
```diff
@@ -134,4 +136,157 @@ angular.module('kifi')
...
+          // For now, we are only supporting inviting one person at a time.
+          var invitees = [
+            {
+              type: !!result.id ? 'user' : 'email',
+              id: !!result.id ? result.id : result.email,
+              access: 'read_only'  // Right now, we're only supporting read-only access.
+            }
+          ];
+
+          // console.log('Library ID: ' + scope.library.id + ' Invitee: ' + JSON.stringify(invitees));
```
yipingliao, line 275: @ahsu1230 and I found yesterday that the endpoint is a little flaky (sometimes it returns 500, sometimes it processes the invite successfully). Leaving the log messages in for debugging for a while longer.

kifi-backend/modules/shoebox/angular/src/libraries/library.styl
```diff
@@ -0,0 +1,7 @@
+.kf-library-share
+  .social-invite-element
+    &:hover
+      cursor: pointer
+
+    &.selected
+      background-color: #eee
```
yipingliao: Oops, will add newline.

kifi-backend/modules/shoebox/angular/src/libraries/library.js
andrewconner, line 275: Interesting. Can you trigger it intentionally?

kifi-backend/modules/shoebox/angular/src/libraries/libraryShareSearch.tpl.html
```diff
@@ -0,0 +1,31 @@
+<div class="social-invite-search">
```
yipingliao, line 1: Some CSS classes are borrowed from the friend search dropdown right now. There will be a lot of CSS changes later when we implement the design, so I have not cleaned this up yet.

kifi-backend/modules/shoebox/angular/src/libraries/library.js
yipingliao, line 275: Unfortunately, no. :(

kifi-backend/modules/shoebox/angular/src/common/services/routeService.js
```diff
@@ -62,6 +62,9 @@ angular.module('kifi')
       friendRequest: function (id) {
         return env.xhrBase + '/user/' + id + '/friend';
       },
+      contactSearch: function (opt_query) {
+        return route('/user/contacts/search' + (!!opt_query ? '?query=' + opt_query : ''));
```
2is10: FYI—coercion to boolean is implicit in this context, so `!!` is unnecessary

yipingliao: I got a little carried away with `!!` in this pull request. :) I'll go back and remove any that are not necessary.

PR #10929, opened 2014-09-16
couple of small changes

ahsu1230:
- fixed response for extension API unkeeping from a library (returns NoContent 204 on success, 4xx otherwise) @2is10 
- include `kind` field for LibraryInfo - better for front-end to distinguish system generated libraries @yipingliao 
- fixing `passcode` bug (extending column size to 30) @andrewconner


kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtLibraryController.scala
```diff
@@ -74,11 +74,7 @@ class ExtLibraryController @Inject() (
         implicit val context = heimdalContextBuilder.withRequestInfoAndSource(request, KeepSource.keeper).build
         keepsCommander.unkeepFromLibrary(Seq(keepExtId), libraryId, request.userId) match {
           case Left(failMsg) => BadRequest(Json.obj("error" -> failMsg))
-          case Right((unkeptKeeps, failures)) =>
-            Ok(Json.obj(
-              "unkept" -> Json.toJson(unkeptKeeps),
-              "failed" -> Json.toJson(failures)
-            ))
+          case Right(_) => NoContent
```
2is10, line 77: This is incorrect. Need to return a 4xx if the user can write to the library but the keep ID is invalid. This is why I think commanders should expose simple dedicated methods for simple operations on one thing. Using complex methods for simple things is error-prone. If we feel strongly about re-using the “many” operation to achieve the “one” operation, then let’s only write that code once, inside the commander, so that each controller doesn’t have to deal with the complexity and risk getting it wrong.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -558,7 +558,8 @@ case class LibraryInfo(
   shortDescription: Option[String],
   url: String,
   ownerId: ExternalId[User],
-  numKeeps: Int)
+  numKeeps: Int,
+  kind: LibraryKind)
```
yipingliao, line 562: Awesome, thank you. I see that kind can be `user_created`. What is the kind when it's not user created? I'll change the front end code on my next check in to use this.

andrewconner, line 562: Right now the others are `system_main` and `system_secret`.

PR #10930, opened 2014-09-16
never filter out uris that were previously recommended to a user

stkem:
otherwise recommendations with incorrectly high scores will never get
corrected if even if we fix the algorithm if the new score is below the
threshold for inclusion

also refactored the code a bit to make it clearer

@yingjieMiao


yingjieMiao:
lgtm. 

unrelated to this PR: the fact that `precomputeRecommendationsForUser` and `processSeeds` calls each other seems unusual. Especially one of them is a recursive function. The behavior looks a bit confusing.

PR #10935, opened 2014-09-16
renaming "url" to "path" in LibraryData and using it in getLibraries

2is10:
Also adding a new `LibraryCommander` method `getLibrariesUserCanKeepTo` that does library access level filtering in the database instead of in memory and avoids loading library invitations unnecessarily.

@ahsu1230

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtLibraryController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtLibraryController.scala
@@ -8,8 +8,7 @@ import com.keepit.common.db.ExternalId
 import com.keepit.common.db.slick.Database
 import com.keepit.common.social.BasicUserRepo
 import com.keepit.heimdal.HeimdalContextBuilderFactory
-import com.keepit.model.{ LibraryMembershipRepo, Keep, KeepSource, Library, LibraryAccess }
-import com.keepit.social.BasicUser
+import com.keepit.model.{ Keep, KeepSource, Library, LibraryAccess, LibraryMembershipRepo }
 import play.api.libs.json._
 
 import scala.util.{ Success, Failure }
@@ -26,12 +25,8 @@ class ExtLibraryController @Inject() (
     extends BrowserExtensionController(actionAuthenticator) with ShoeboxServiceController {
 
   def getLibraries() = JsonAction.authenticated { request =>
-    val basicSelf = BasicUser.fromUser(request.user)
     val datas = libraryCommander.getLibrariesUserCanKeepTo(request.userId) map { lib =>
-      val owner = if (lib.ownerId == request.userId)
-        basicSelf
-      else
-        db.readOnlyMaster { implicit s => basicUserRepo.load(lib.ownerId) }
+      val owner = db.readOnlyMaster { implicit s => basicUserRepo.load(lib.ownerId) }
       LibraryData(
         id = Library.publicId(lib.id.get),
         name = lib.name,
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtLibraryController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtLibraryController.scala
@@ -1,7 +1,7 @@
 package com.keepit.controllers.ext
 
 import com.google.inject.Inject
-import com.keepit.commanders.{ KeepsCommander, RawBookmarkRepresentation, LibraryCommander }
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtLibraryController.scala
```diff
@@ -25,18 +26,19 @@ class ExtLibraryController @Inject() (
...
-        "id" -> Library.publicId(lib.id.get).id,
-        "name" -> lib.name,
-        "path" -> Library.formatLibraryPath(owner.username, owner.externalId, lib.slug),
-        "visibility" -> Json.toJson(lib.visibility))
+    val basicSelf = BasicUser.fromUser(request.user)
+    val datas = libraryCommander.getLibrariesUserCanKeepTo(request.userId) map { lib =>
+      val owner = if (lib.ownerId == request.userId)
+        basicSelf
+      else
+        db.readOnlyMaster { implicit s => basicUserRepo.load(lib.ownerId) }
```
2is10: @ahsu1230 @andrewconner Do you like `basicSelf` here? It just seems like a waste to use database code and possibly access memcache or the database when, in the common case (user’s own libraries), you’ve got the `User` you need sitting right there in memory. Does the fact the request has authenticated successfully guarantee that the user will also be in an in-memory cache in `BasicUserRepo`/`UserRepo`? If so, the database code path probably never adds a measurable delay for the user’s own libraries, and there’s no reason to introduce `basicSelf`.

andrewconner: Why not just do `BasicUser.fromUser(request.user)` only in the if-true branch?

andrewconner: To answer your question, if a user is authenticated, it can be fairly safely assumed that they're in the caches.

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryRepo.scala
```diff
@@ -17,7 +17,7 @@ import scala.slick.jdbc.StaticQuery.interpolation
 trait LibraryRepo extends Repo[Library] with SeqNumberFunction[Library] {
   def getByIdAndOwner(libraryId: Id[Library], ownerId: Id[User], excludeState: Option[State[Library]] = Some(LibraryStates.INACTIVE))(implicit session: RSession): Option[Library]
   def getByNameAndUserId(userId: Id[User], name: String, excludeState: Option[State[Library]] = Some(LibraryStates.INACTIVE))(implicit session: RSession): Option[Library]
-  def getByUser(userId: Id[User], excludeState: Option[State[Library]] = Some(LibraryStates.INACTIVE))(implicit session: RSession): Seq[(LibraryAccess, Library)]
+  def getByUser(userId: Id[User], excludeState: Option[State[Library]] = Some(LibraryStates.INACTIVE), excludeAccess: Option[LibraryAccess] = None)(implicit session: RSession): Seq[(LibraryAccess, Library)]
```
ahsu1230, line 20: instead of a single excludeAccess `Option` we could use an excludeSet instead so we can exclude multiple access rights in one call
If no access rights need to be excluded, just pass in an empty Set

2is10, line 20: That’s a good idea. I considered it. I’d prefer to wait until there’s a use for it. I believe queries that use `in` with a variable-length list either can’t be pre-compiled or can’t use a prepared statement—I forget which. It looks like this query is already not pre-compiled for some reason, so it might not matter. I just don’t want to risk making it harder to optimize this query until we have a need for it.

Okay, here’s [a good explanation](http://stackoverflow.com/questions/19926258/are-parameters-of-type-list-or-set-in-slick-query-templates-possible-lifted-ap) of the issues (see the top two answers). You can’t precompile. You can get the benefit of prepared statements if you use `inSetBind`—there’s just a different prepared statement for each set size actually passed.

kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtLibraryController.scala
2is10: Okay, I’ll eliminate the `if` branch entirely.

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryRepo.scala
ahsu1230, line 20: :ok_hand: gotcha, let's wait then

PR #10936, opened 2014-09-16
adding message field to library_invite

ahsu1230:
message is an Option[String] and is a `text` in database

@yipingliao @andrewconner


ahsu1230:
I actually right now have a message per invitee
But I see Asana has an overall message for all invites, changing it...

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
```diff
@@ -188,13 +188,14 @@ class LibraryController @Inject() (
 
         val validInviteList = db.readOnlyMaster { implicit s =>
           invites.map { i =>
+            val message = (i \ "message").asOpt[String]
```
andrewconner: Mind making `""` be a `None` also?

yipingliao:
Thanks! Is there a length limit on this string/what length should the front end limit this string to?

ahsu1230:
http://help.scibit.com/mascon/masconMySQL_Field_Types.html

I think it says a normal `text` can have up to 16k characters should be enough right?

PR #10937, opened 2014-09-16
add public id to the library view

ymatsuda:
Added public id to the admin library view
fyi @andrewconner @ahsu1230


andrewconner:
Looks good.

PR #10939, opened 2014-09-16
refactor digest email queue to use nextWithLock instead of nextBatchWithLock

joshm1:
should be functionally identical to before

@stkem


stkem:
lgtm, but why the change? Throttling?

joshm1:
@stkem yeah. I discussed it with @eishay earlier, we decided even 5 in parallel wasn't necessary (although it actually wasn't 5 this morning, it was using a batch size of 1 by mistake).

PR #10944, opened 2014-09-16
hiding user session model in the shoebox

eishay:
Applying some comments from @2is10 and all.
Note changes with 

``` scala
case class UserSessionExternalId(id: String) extends SurrogateExternalId
```

/cc @andrewconner @leogrim @ymatsuda @raymondng


eishay:
@2is10

kifi-backend/modules/abook/app/com/keepit/common/cache/ABookCacheModule.scala
```diff
@@ -66,7 +68,7 @@ case class ABookCacheModule(cachePluginModules: CachePluginModule*) extends Cach
   @Singleton
   @Provides
   def userSessionExternalIdCache(stats: CacheStatistics, accessLog: AccessLog, innerRepo: InMemoryCachePlugin, outerRepo: FortyTwoCachePlugin) =
```
2is10, line 70: make method name consistent? inconsequential, I realize

kifi-backend/modules/common/app/com/keepit/common/db/ExternalId.scala
```diff
@@ -55,3 +60,51 @@ object ExternalId {
...
+
+  implicit def queryStringBinder(implicit stringBinder: QueryStringBindable[String]) = new QueryStringBindable[T] {
+    override def bind(key: String, params: Map[String, Seq[String]]): Option[Either[String, T]] = {
+      stringBinder.bind(key, params) map {
+        case Right(id) =>
+          asOpt(id) match {
+            case Some(extId) => Right(extId)
+            case None => Left(s"Unable to bind an ExternalId with $id")
+          }
+        case _ => Left("Unable to bind an ExternalId")
```
2is10, line 91: should we change these methods to say `SurrogateExternalId`?

kifi-backend/modules/common/app/com/keepit/common/routes/Router.scala
```diff
@@ -1,7 +1,9 @@
 package com.keepit.common.routes
 
-import com.keepit.common.db.{ SequenceNumber, ExternalId, Id, State }
+import com.keepit.common.db._
```
2is10: :frowning:  I like seeing the types listed.

kifi-backend/modules/common/app/com/keepit/common/routes/Router.scala
```diff
@@ -1,7 +1,9 @@
...
 import com.keepit.model._
+import com.keepit.shoebox.model.Ids
+import Ids.UserSessionExternalId
```
2is10: Since these are going to be imported in a lot of places, it would be nice to choose a code organization that allows for simple one-line, standard-looking imports. Did you consider an `ids` package object?

I’m not going to push hard on this—just a thought.

kifi-backend/modules/common/app/com/keepit/shoebox/model/Ids.scala
```diff
@@ -0,0 +1,12 @@
+package com.keepit.shoebox.model
+
+import com.keepit.common.db.{SurrogateExternalId, SurrogateExternalIdCompanion}
+
+object Ids {
+  case class UserSessionExternalId(id: String) extends SurrogateExternalId
+
+  object UserSessionExternalId extends SurrogateExternalIdCompanion[Ids.UserSessionExternalId] {
+    def create(id: String): Ids.UserSessionExternalId = Ids.UserSessionExternalId(id)
```
2is10: This should work without the `Ids.` qualifiers, no?

kifi-backend/modules/common/app/com/keepit/shoebox/model/Ids.scala
```diff
@@ -0,0 +1,12 @@
...
+  }
```
2is10: Nice work, very little boilerplate per type.

I wonder if we could reduce the companion type declaration to one line. Could we pass the ID type constructor as an argument instead? Something like:

``` scala
object UserSessionExternalIdComp extends SurrogateExternalIdCompanion[UserSessionExternalId](UserSessionExternalId)
```

Instantiating by reflection is another possibility, but would nice to avoid.

2is10:
Minor comments. I’m happy with the approach. Nicely done.

kifi-backend/modules/common/app/com/keepit/common/routes/Router.scala
eishay: I'm with you, my IDE is collapsing imports if there are more then five in the list. Updated the configs to ten, see how it works.

kifi-backend/modules/common/app/com/keepit/common/routes/Router.scala
eishay: yes, I had to use object for the `type` but since we use case classes now we can go back to `package`, will do. Thanks!

kifi-backend/modules/common/app/com/keepit/shoebox/model/Ids.scala
eishay: It did have a problem with it. Getting the wrapper object out of the way made the compiler happy without the `Ids.`

eishay:
@2is10 reduce the `SurrogateExternalIdCompanion` to one-liner, thanks for pushing for it - much cleaner now. thanks for @andrewconner for the `apply` tip

PR #10946, opened 2014-09-16
changing extension response for unkeeping from libraries

ahsu1230:
@2is10 @andrewconner 
Here's some refactoring we discussed earlier. In the commander there's an `unkeepOne` and `unkeepMany` function which wrap around a common private unkeeping function. With this, it's easy for controllers to call the commander functions and be able to parse out errors, or failed/successful unkeeps

While there are already lots of unkeeping functions in KeepCommander, we'll consolidate them and do some cleanups later so any unkeep APIs are library-aware and like better named lol

Let me know if you have any thoughts

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
@@ -459,25 +459,42 @@ class KeepsCommander @Inject() (
     }
   }
 
-  def unkeepFromLibrary(keeps: Seq[ExternalId[Keep]], libId: Id[Library], userId: Id[User])(implicit context: HeimdalContext): Either[String, (Seq[KeepInfo], Seq[ExternalId[Keep]])] = {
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
```diff
@@ -459,25 +459,42 @@ class KeepsCommander @Inject() (
...
+        val (infos, failures) = unkeepFromLibrary(Seq(keep), libId, userId)
+        if (infos.isEmpty)
+          Left("invalid_keep_id")
+        else
+          Right(infos.head)
+      case _ =>
+        Left("permission_denied")
+    }
+  }
+  def unkeepManyFromLibrary(keeps: Seq[ExternalId[Keep]], libId: Id[Library], userId: Id[User])(implicit context: HeimdalContext): Either[String, (Seq[KeepInfo], Seq[ExternalId[Keep]])] = {
```
ahsu1230, line 473: still wondering if there's a better return type signature

2is10:
Nice. The signature for `unkeepOneFromLibrary` makes calling from controllers much easier and less error-prone.

I’m not seeing the benefit of factoring out a private helper method in the commander. I think `unkeepOneFromLibrary` could simply call `unkeepManyFromLibrary` and then translate the complicated result into its simpler return type.

ahsu1230:
removed private function! looks a lot better now =)

PR #10948, opened 2014-09-16
adding follow/unfollow capabilities to library reco card

stkem:
not sure if I’m following current patterns well

@yipingliao @andrewconner @ahsu1230

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
@@ -31,13 +31,16 @@ angular.module('kifi')
 
         // Data augmentation. May want to move out to own decorator service
         // like the keepDecoratorService.
-        if (scope.library.owner) {
-          scope.library.owner.image = friendService.getPictureUrlForUser(scope.library.owner);
-        }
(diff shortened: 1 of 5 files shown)


stkem:
That is pretty much all I can do for now. The rest will have to wait for the designs (it is functionally complete, just ugly)

kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
```diff
@@ -38,6 +38,34 @@ angular.module('kifi')
...
+        };
+
+        scope.follow = function () {
+          if (!scope.followed()) {
+            libraryService.joinLibrary(scope.library.id);
+            scope.library.followers.push({
+              id: profileService.me.id,
+              firstName: profileService.me.firstName,
+              lastName: profileService.me.lastName,
+              image: friendService.getPictureUrlForUser(profileService.me)
```
yipingliao: replace with `pictureName: profileService.me.pictureName` to be consistent with the follower objects we get from the server. Line 39 converts `pictureName` to the URL for all followers.

yipingliao:
:+1:

PR #10956, opened 2014-09-17
Clean keep.js

yipingliao:
This pull is a follow-up to the `keepService` refactoring pull last week, and cleans up a couple of things:
- Reorganized the code in `keep.js` a bit; should be fairly easy to drop in the neural networks code now.
- Renamed `kfKeepContent`, which was made up just to be different from `kfKeep` when we had two copies, to `kfKeepCard` (to be consistent with the new `kfLibraryCard`)
- Reduced number of calls to update keep tags.

@andrewconner


kifi-backend/modules/shoebox/angular/src/common/selectionService.js
```diff
@@ -67,10 +67,10 @@ angular.module('kifi')
     };
 
     Selection.prototype.getSelected = function (keeps) {
-      var self = this;
```
yipingliao, line 70: I noticed that elsewhere in that code we use `that`, so changing to be consistent.

kifi-backend/modules/shoebox/angular/src/keeps/keeps.js
```diff
@@ -8,14 +8,23 @@ angular.module('kifi')
     $scope.me = profileService.me;
     $scope.data = {draggedKeeps: []};
 
-    $scope.$watch(function () {
-      // TODO: is this too inefficient? Will this be called too many times?
-      return $scope.keeps.length + ',' + tagService.allTags.length;
-    }, function () {
+    // Whenever new keeps are loaded or when tags have been added or removed,
```
yipingliao, line 11: Previously, before the refactor, this watch was triggered whenever `keepService` was reset, which is effectively once per load. When I removed `keepService`, I triggered this whenever the length of the keeps in a controller changed. This was much more frequent because the keeps in the controllers are pushed in one by one. The new code triggers this watch only on the falling edge of `keepsLoading`, which brings the frequency of joining tags back down.

kifi-backend/modules/shoebox/angular/src/common/selectionService.js
```diff
@@ -67,10 +67,10 @@ angular.module('kifi')
...
+      var that = this;
 
       return keeps.filter(function (keep) {
-        return keep.id in self.selected;
+        return keep.id in that.selected;
```
2is10: might also consider:

``` javascript
var selected = this.selected;
return keeps.filter(function (keep) {
  return keep.id in selected;
});
```

Note that the `|| []` at the end is useless.

kifi-backend/modules/shoebox/angular/src/keep/keep.js
```diff
@@ -303,26 +333,16 @@ angular.module('kifi')
...
-
-        // TODO: bind to 'drop' event
-        scope.onTagDrop = function (tag) {
-          tagService.addKeepToTag(tag, scope.keep);
-          scope.isDragTarget = false;
-        };
+        // Dragging.
+        scope.$watch('editMode.enabled', function () {
+          element.attr('draggable', true);
+        });
```
2is10: Would it be possible to write this into the template instead? (I don’t know Angular well enough.)

``` html
draggable="!!editMode.enabled"
```

2is10:
good changes, thanks

kifi-backend/modules/shoebox/angular/src/keep/keep.js
yipingliao: Yes, it will actually look like `draggable="{{!!editMode.enabled}}"`. However, after digging into this, I discovered that this code always results in `draggable` being true, so I am going to remove this and add the attribute to the template. Thanks for spotting this!

kifi-backend/modules/shoebox/angular/src/common/selectionService.js
yipingliao: Updating, thanks.

PR #10957, opened 2014-09-17
if timing out when trying to get uri resources (i.e. images) then just i...

eishay:
...gnore value and not try again until cache expires
@stkem


stkem:
Actually, the second request often (not always) succeeds because embedly now knows about the uri and responds a lot quicker. This way we won't try again for 30 days. Just making sure that that is intended.

eishay:
2 minutes ttl is very long, and  in cases of repeating timeouts for the same url its a churn on the system. How about reducing that cache to one or two days?

stkem:
That could increase total load significantly (even if reducing timeouts). An alternative could be to set `waiting` to false in the request here (causing it to return immediately if there is nothing and fetch it in the background) and _not_ cache empty responses.

eishay:
Good idea, I'm going to hold on this pull request for now as we're trying
alternative options.

PR #10958, opened 2014-09-17
making unkeep from library idempotent

2is10:
@ahsu1230

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
@@ -437,6 +437,7 @@ class KeepsCommander @Inject() (
       keeps.map(setKeepStateWithSession(_, KeepStates.INACTIVE, userId))
     }
     finalizeUnkeeping(keeps, userId)
+    keeps map KeepInfo.fromKeep
   }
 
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
```diff
@@ -472,30 +475,32 @@ class KeepsCommander @Inject() (
...
         Left("permission_denied")
     }
   }
 
-  private def finalizeUnkeeping(keeps: Seq[Keep], userId: Id[User])(implicit context: HeimdalContext): Seq[KeepInfo] = {
+  private def finalizeUnkeeping(keeps: Seq[Keep], userId: Id[User])(implicit context: HeimdalContext): Unit = {
     // TODO: broadcast over any open user channels
     keptAnalytics.unkeptPages(userId, keeps, context)
     searchClient.updateURIGraph()
-    keeps map KeepInfo.fromKeep
```
ahsu1230, line 498: just curious, what's the advantage of taking this out of finalizeUnkeeping?

ahsu1230:
looks good, thanks

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
2is10, line 498: I felt it made the code more confusing. It’s unrelated to finalizing the unkeeping. It’s just about preparing a method’s return value. In particular, in `unkeepManyFromLibrary`, the set of keeps that we need to return is now potentially different from the set of keeps that need “unkeep finalizing”.

PR #10959, opened 2014-09-17
Auth refractor: baby steps

raymondng:
Introducing new `Action` and `Request` types that allow us to take full advantage of Play 2.3 `Action` API (with composition support, among other things) and remove some unnecessary dependencies on `SecureSocial`.

The end goal is to provide a simpler and more efficient API to ultimately replace the current `ActionsBuilder`-based API.

Implemented basic primitives described in @andrewconner's design spec:
https://docs.google.com/a/42go.com/document/d/1zYmCC0C4VTG4-fVZSXgtb0KBlJ1H-D4FDLugn15aSxw

Added `WebController` that wired in `UserActions` and wired up one endpoint (typeahead) only as a sanity check. 

Baby steps; much more to come

@andrewconner @ymatsuda @stkem @2is10 @eishay


2is10:
​cool!

kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  def isAdmin(userId: Id[User]): Boolean
+
+  def getUserOpt(implicit request: Request[_]): Future[Option[User]]
+
+  def getUserExperiments(implicit request: Request[_]): Future[Set[ExperimentType]]
+}
+
+trait UserActions { self: Controller =>
```
andrewconner, line 37: Suggestion: `UserAwareAPIActions` or similar. These are endpoints for users or non-users from clients (not direct web browsing).

kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
```diff
@@ -0,0 +1,76 @@
+package com.keepit.common.controller
+
+import com.keepit.common.db.{ ExternalId, Id }
+import com.keepit.model.{ ExperimentType, KifiInstallation, User }
+import play.api.mvc._
+
+import scala.concurrent.ExecutionContext.Implicits.global
```
andrewconner: We prefer to use: `import play.api.libs.concurrent.Execution.Implicits.defaultContext`

Any current usage of Scala's is a test or a mistake.

andrewconner:
SO much cleaner!

kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
raymondng: Thanks. Will fix.

kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
raymondng, line 37: This is a bit of an implementation detail and I struggled to find a good name for it -- Helper is admittedly a bit of a cop-out. Will look into it.

andrewconner:
Of note, looks like `SecureSocialAuthenticatorPluginImpl` is the only class that really uses `UserSession`. So: we should be able to cut that fairly easily as well.

raymondng:
Would be more than happy to get rid of it!

kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
andrewconner, line 37: I see. Perhaps below a better naming would be `UserAPIController` instead of `WebController`.

kifi-backend/modules/common/app/com/keepit/common/controller/WebsiteController.scala
```diff
@@ -17,3 +17,6 @@ abstract class WebsiteController(val actionAuthenticator: ActionAuthenticator) e
   object AnyAction extends Actions.AuthenticatedActions with Actions.NonAuthenticatedActions
 
 }
+
+abstract class WebController(val userActionsHelper: UserActionsHelper) extends ServiceController with UserActions {
```
andrewconner: I _think_ we can lose the idea of a WebsiteController vs BrowserExtensionController vs MobileController. Thoughts?

raymondng: I have no problem with it but I don't know enough about context/usage ... other experts please chime in.

2is10: Okay with me. Package and class names make it clear what’s what.

stkem:
Nice! The current unauthenticated requests setup include a campaign tracking feature which will also have to be added here before putting this into prod (otherwise we'll loose data). Should hopefully be a simple copy/paste. See `ActionsBuilder.scala:51` (`augmentedUnauthenticatedRequest`)

raymondng:
Thanks! Yes I have taken note of it. That is why I have only wired up one
endpoint. Is that still a concern? If so I can remove that wiring

stkem:
The endpoint you wired up is fine, since it's authed. My concern is that it will be used for other (unauthed) endpoints when it's merged when it's in fact not ready, i.e. I think the campaign tracking needs to be in there before it goes into prod. I'd be happy to help if it turns out to be more complex.

andrewconner:
We will not use it for anything serious until it's ready.

raymondng:
Yup. Miles to go. Just baby steps.

kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
```diff
@@ -0,0 +1,64 @@
+package com.keepit.common.controller
+
+import com.keepit.common.db.{ ExternalId, Id }
+import com.keepit.model.{ ExperimentType, KifiInstallation, User }
+import play.api.mvc._
```
stkem: How many things from this package are you actually using? I'm not a big fan of `._` imports in general (with a few exceptions) as they make the code harder to read (and you might be getting implicits you don't want).

kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
raymondng, line 37: Done

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/TypeaheadController.scala
```diff
@@ -4,7 +4,7 @@ import com.google.inject.Inject
 import com.keepit.common.db.slick.Database
 import com.keepit.common.healthcheck.AirbrakeNotifier
 import com.keepit.common.logging.Logging
-import com.keepit.common.controller.{ ShoeboxServiceController, ActionAuthenticator, WebsiteController }
+import com.keepit.common.controller._
```
stkem: Same as above, is the `._` really needed?

stkem:
Someone else might. I get the point about having both the old and the new setup in prod in parallel for a bit to test things, but I don't see the benefit of putting the new setup in prod before it's complete. Besides, this particular thing will take 2 minutes.

kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
raymondng: Let me check. Actually I share your preference -- I blame it on IDEA.

kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
```diff
@@ -0,0 +1,64 @@
...
+        case Some(userId) => block(buildUserRequest(userId))
+        case None => Future.successful(Unauthorized)
+      }
+    }
+  }
+
+  object MaybeUserAction extends ActionBuilder[MaybeUserRequest] {
+    def invokeBlock[A](request: Request[A], block: (MaybeUserRequest[A]) => Future[Result]): Future[Result] = {
+      implicit val req = request
+      userActionsHelper.getUserIdOpt match {
```
stkem, line 57:
``````
.map { res =>
        request.queryString.get("kcid").flatMap(_.headOption).map { kcid =>
             res.addingToSession("kcid" -> kcid)
        } getOrElse res
}```
at the end here should do the trick for campaign tracking if I'm reading things correctly.
``````

raymondng, line 57: Thanks. Let me give it a shot.

raymondng:
Thanks all for the feedback. Will merge and build upon it -- a lot more work to do.

PR #10960, opened 2014-09-17
reducing the amount of information about library sent to extension with each keep

2is10:
The extension really only needs to know the library’s ID and whether it’s secret.
@ahsu1230


ahsu1230:
:+1:

PR #10962, opened 2014-09-17
Minor changes (logging primarily)

andrewconner:
This disables info logging during tests — makes it a LOT easier to read. If you need them enabled, just modify `test-logger.xml`.


kifi-backend/.gitignore
```diff
@@ -11,7 +11,6 @@ app_version.txt
 access*
 cx.properties
 bin
-data
```
andrewconner, line 14: @2is10 needed to remove this because we have an existing test folder in tests, and it looks like ignoring all "data" folders keeps interacting with the files inside.

andrewconner, line 14: This didn't add any files, so should be okay?

2is10, line 14: fine w/me

PR #10966, opened 2014-09-17
Removing println in tests

andrewconner:
@raymondng @ymatsuda @stkem @eishay @leogrim 

Tests are quite noisy. This comments out all the printlns in tests, with an easy to grep comment asking about removal.


andrewconner:
Reviewed the idea with @leogrim, @raymondng, and @ymatsuda. No complains, so merging to avoid conflicts.

raymondng:
:+1:

PR #10970, opened 2014-09-17
+ crossOrigin

raymondng:
I think this should do it.

Mostly copy-and-paste from `ActionAuthenticator`, plus some minor refactoring

@andrewconner @yipingliao

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
+++ b/kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
@@ -1,6 +1,7 @@
 package com.keepit.common.controller
 
 import com.keepit.common.db.{ ExternalId, Id }
+import com.keepit.common.net.URI
 import com.keepit.model.{ ExperimentType, KifiInstallation, User }
 import play.api.mvc.{ Request, WrappedRequest, Controller, ActionBuilder, Result }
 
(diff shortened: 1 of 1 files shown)


yipingliao:
Thanks!

raymondng:
cc: @2is10 and @eishay -- I think the changes are safe nonetheless would appreciate a quick once-over.

raymondng:
Would appreciate feedback but will merge to unblock @yipingliao (sorry about the inconvenience!)

kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
```diff
@@ -41,11 +42,34 @@ trait UserActions { self: Controller =>
...
+    }
+  }
+
+  private def maybeAugmentCORS[A](res: Result)(implicit request: Request[A]): Result = {
+    getOrigin(request) map { h =>
+      res.withHeaders(
+        "Access-Control-Allow-Origin" -> h,
+        "Access-Control-Allow-Credentials" -> "true"
+      )
+    } getOrElse res
```
2is10, line 58: This looks safe, but I was falsely alarmed for a while because I read `maybeAugmentCORS` first and it doesn’t appear to be doing any verification of the request origin. I’d suggest either folding `getOrigin` into `maybeAugmentCORS` for better locality (as we do elsewhere) or renaming `getOrigin` to reflect more accurately what it does (maybe `getOriginIfTrusted`?). I’d be surprised if we ever reuse that method.

raymondng, line 58: Thanks @2is10. Agreed that `getOrigin` can be named better. Let me take care of it.

PR #10971, opened 2014-09-17
refreshed kifi-black emails for everybody

joshm1:
@stkem 

cc @ahsu1230  to see the email senders


joshm1:
I've got 2 more pull requests to finalize the refreshed emails. b/c this one is big already I'm going to submit them separately

kifi-backend/modules/shoebox/app/com/keepit/commanders/UserCommander.scala
```diff
@@ -373,11 +372,11 @@ class UserCommander @Inject() (
               category = NotificationCategory.User.CONTACT_JOINED
             )
 
-            toNotify
+            Future.sequence(emailsF.toSeq) map (_ => toNotify)
```
eishay, line 375: what does this line do?

joshm1, line 375: Wraps the existing return value in a future that allows me to unit test the email this method sends (await).  The previous emails in shoebox were queued synchronously.  I use this technique in other places too for the same reason.

PR #10972, opened 2014-09-17
Attributing Recos to a Library you follow

stkem:
@yingjieMiao 
@yipingliao (FE part)


yipingliao:
:+1:

kifi-backend/modules/curator/app/com/keepit/curator/model/CuratorLibraryMembershipInfoRepo.scala
```diff
@@ -5,12 +5,15 @@ import com.keepit.common.db.Id
...
+import com.keepit.model.{ LibraryAccess, LibraryKind, Library, User, NormalizedURI }
+
+import scala.slick.jdbc.StaticQuery
 
 @ImplementedBy(classOf[CuratorLibraryMembershipInfoRepoImpl])
 trait CuratorLibraryMembershipInfoRepo extends DbRepo[CuratorLibraryMembershipInfo] {
   def getByUserAndLibraryId(userId: Id[User], libId: Id[Library])(implicit session: RSession): Option[CuratorLibraryMembershipInfo]
   def getLibrariesByUserId(userId: Id[User])(implicit session: RSession): Seq[Id[Library]]
+  def getFollowedLibrariesWithUri(userId: Id[User], uriId: Id[NormalizedURI])(implicit session: RSession): Set[Id[Library]]
 }
```
yingjieMiao, line 17: why `Set`? You converted `list` to `Set`, and never used set properties. The only usage I see is to construct a `seq` in the attribution, which requires another conversion.

kifi-backend/modules/curator/app/com/keepit/curator/model/CuratorLibraryMembershipInfoRepo.scala
```diff
@@ -42,4 +45,25 @@ class CuratorLibraryMembershipInfoRepoImpl @Inject() (
...
+        AND
+        libs.user_id=$userId
+        AND
+        keeps.state='active'
+        AND
+        libs.state='active'
+        AND
+        libs.library_access='read_only';
+    """.as[Id[Library]].list.toSet
+
```
yingjieMiao, line 67: nice format

yingjieMiao:
one minor comment. lgtm overall.

kifi-backend/modules/curator/app/com/keepit/curator/model/CuratorLibraryMembershipInfoRepo.scala
stkem, line 17: I used `Set` to make it clear from the type signature of the function that the returned library Ids are unique.

stkem, line 17: i.e. while I'm not really using the set properties/methods, the result is semantically a set.

yingjieMiao, line 17: I see, but that can be achieved by `select distinct`. Performance concern?

stkem, line 17: Oh, they are already unique due to the unique constraint on `(user_id, library_id)` in the table. The `Set` is just to communicate that fact through the type, i.e. a user of the method can rely on the results being unique without having to know about the implementation.

kifi-backend/modules/curator/app/com/keepit/curator/model/CuratorLibraryMembershipInfoRepo.scala
2is10, line 67: could be just as readable, I think, without operators on their own lines:

``` sql
 SELECT libs.library_id
 FROM curator_keep_info keeps, curator_library_membership_info libs
 WHERE keeps.library_id = libs.library_id
   AND keeps.uri_id = $uriId
   AND libs.user_id = $userId
   AND keeps.state = 'active'
   AND libs.state = 'active'
   AND libs.library_access = 'read_only';
```

PR #10977, opened 2014-09-17
library ingestions into graph

yingjieMiao:
@stkem @leogrim @andrewconner 
- create/remove keep-lib, user-lib edges
- created `NeighborQuerier` trait to enable easy nb query of the graph
- farily comprehensive test for graph data ingestion


leogrim:
:+1: for the ingestion part
Check out GraphPrimitives for the query part, it'd be nice to build primitives in there rather than tying them to random walks. We may want to use them in a different context :)

yingjieMiao:
oh, nice. I didn't know it was there.

kifi-backend/modules/graph/app/com/keepit/graph/manager/GraphUpdater.scala
```diff
@@ -112,8 +117,11 @@ class GraphUpdaterImpl @Inject() () extends GraphUpdater {
         writer.saveEdge(keepVertexId, userVertexId, EmptyEdgeData)
         writer.saveEdge(keepVertexId, uriVertexId, EmptyEdgeData)
 
-      case _ =>
-        writer.removeVertexIfExists(keepVertexId)
+        writer.saveEdge(libVertexId, keepVertexId, EmptyEdgeData)
+        writer.saveEdge(keepVertexId, libVertexId, EmptyEdgeData)
+
+      case _ => // do nothing
```
leogrim, line 124: We need to remove the keep vertex when it's not active don't we?

leogrim:
We may also want to ingest library vertex from the library table, so that vertices corresponding to inactive libraries can be cleared out.

kifi-backend/modules/graph/app/com/keepit/graph/manager/GraphUpdater.scala
yingjieMiao, line 124: it's removed before the `match`

leogrim, line 124: Oh right, sorry :)

yingjieMiao:
@leogrim  I think you are right. will ingest that part.

PR #10978, opened 2014-09-18
Implement more library share interaction functionality

yipingliao:
Implemented more library share interaction functionality:
- Share menu displayed only when share button is pressed, and share button is displayed only if the user has the right permissions
- Autofocus on input; auto select the first user
- Add backfill items for new email contact and importing Gmail contacts
- Existing contacts get a share button
- Contacts who are already following the library are displayed with "Following" and cannot be invited
- Send share invitations with optional message

The menu is still very ugly and unstyled, and some detailed user interactions are still wonky. Making this pull since we're just internally testing right now.

@andrewconner @2is10


andrewconner:
Nice, looks good.

kifi-backend/modules/shoebox/angular/src/common/util.js
```diff
@@ -15,6 +15,8 @@ angular.module('util', [])
       trimInput: function (input) {
         return input ? input.trim().replace(/\s+/g, ' ') : '';
       },
+      // TODO(yiping): conform this to the test being used by the extension.
+      // This one is valid for abc@example and does not need the '.' while that for the extension does.
```
2is10, line 19: I just changed the last `*` in the regular expression to a `+` for the extension. I did that because the extension validates as the user types, and it looks dumb if it suggests sending to `jaredjacobs@gm`. Looks better to say “keep typing”. If the angular app is doing something similar, feel free to use the same approach as the extension. But if not, feel free to leave this regular expression as is (the HTML5 standard).

kifi-backend/modules/shoebox/angular/src/libraries/library.js
```diff
@@ -3,8 +3,16 @@
...
+  '$scope',
+  '$rootScope',
+  '$location',
+  '$routeParams',
+  'keepDecoratorService',
+  'libraryService',
+  'profileService',
+  'util', 
+  function ($scope, $rootScope, $location, $routeParams,
+            keepDecoratorService, libraryService, profileService, util) {
```
2is10: May not be that big a deal, but having one list as a vertical column and one in horizontal rows makes it a little harder to visually verify that the lists match up.

kifi-backend/modules/shoebox/angular/src/common/util.js
yipingliao, line 19: Yes, I am trying to conform the user experience to the extension's (per Danny). Will be borrowing styles as well. :)

kifi-backend/modules/shoebox/angular/src/libraries/library.js
yipingliao: Good point. It's just that they get so long... But having two vertical columns looks weird too. I'll start migrating back to the two horizontal rows style.

andrewconner: When we get some time, we can get ng-annotate (https://github.com/olov/ng-annotate) to work, automatically doing the duplication for us.

yipingliao: Nice!

PR #10979, opened 2014-09-18
shoebox service client for canViewLibrary() and hashing on invite pass c...

ahsu1230:
...odes (on controller levels)

ShoeboxServiceClient now has a `canViewLibrary()` function - which may or may not include a userId
uses the `LibraryCommander.canViewLibrarY()` function which checks if user can at least view library based on...
1) library is published
2) correct universal link is provided
3) user has membership to library
4) user provides correct authentication code AND hashed pass phrase
@ymatsuda 

Also added in hashing on passcodes
A few changes on leaveLibrary lol
@andrewconner


ahsu1230:
do not merge yet - still sorting through tests

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -397,6 +413,7 @@ class LibraryCommander @Inject() (
     db.readWrite { implicit s =>
       libraryMembershipRepo.getWithLibraryIdAndUserId(libraryId, userId) match {
         case None => Left(LibraryFail("membership_not_found"))
+        case Some(mem) if mem.access == LibraryAccess.OWNER => Left(LibraryFail("cannot_leave_own_library"))
```
andrewconner, line 422: Nice

ahsu1230:
still sorting through tests

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -207,6 +207,28 @@ class LibraryCommander @Inject() (
...
+    passCode: Option[HashedPassPhrase] = None): Boolean = {
+
+    library.visibility == LibraryVisibility.PUBLISHED || // published library
+      db.readOnlyMaster { implicit s =>
+        userId match {
+          case Some(id) =>
+            libraryMembershipRepo.getOpt(userId = id, libraryId = library.id.get).nonEmpty ||
+              libraryInviteRepo.getWithLibraryIdAndUserId(userId = id, libraryId = library.id.get, excludeState = Some(LibraryInviteStates.INACTIVE)).nonEmpty
+          case None if (passCode.nonEmpty && inviteToken.nonEmpty) =>
+            libraryInviteRepo.getWithLibraryId(libraryId = library.id.get)
```
andrewconner, line 221: Query by user.

ahsu1230, line 221: this is the case that no userId is given. soo I don't think you can query by user?

andrewconner, line 221: I see. Perhaps we should store the hash then? This could get quite expensive. i.e., instead of the plaintext

ymatsuda:
authToken, accessToken, accessCode, are these the same thing as the auth code?
also there are passCode and passPhrase.
Confusing. Please decide the official terminology and stick with it.
@andrewconner @ahsu1230

ymatsuda:
I assume I can get hashedPassPhrase from a cookie. Correct?
What's the name of the cookie? Have you decided?
@andrewconner @ahsu1230

ahsu1230:
@ymatsuda no we haven't decided yet. Any suggestions?
Also we'll stick with authToken & passCode from now on. Sorry about the confusion

andrewconner:
@ymatsuda 

> I assume I can get hashedPassPhrase from a cookie. Correct?

Yes. I am using this to finalize my custom Action that handles this.

> What's the name of the cookie? Have you decided?

It's in the Play session cookie. Jared had thoughts regarding the naming, but this will be abstracted away from the consumer.

> authToken, accessToken, accessCode, are these the same thing as the auth code

Yes. It should be "authToken" everywhere.

> also there are passCode and passPhrase

Should be "passPhrase" everywhere.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -207,6 +207,28 @@ class LibraryCommander @Inject() (
...
   }
 
+  def canViewLibrary(userId: Option[Id[User]], library: Library,
+    inviteToken: Option[String] = None,
+    passCode: Option[HashedPassPhrase] = None): Boolean = {
+
+    library.visibility == LibraryVisibility.PUBLISHED || // published library
+      db.readOnlyMaster { implicit s =>
+        userId match {
+          case Some(id) =>
```
andrewconner, line 217: This branch does not check the authToken and passPhrase.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -207,6 +207,28 @@ class LibraryCommander @Inject() (
...
+            libraryMembershipRepo.getOpt(userId = id, libraryId = library.id.get).nonEmpty ||
```
andrewconner, line 218: Need to check state.

andrewconner, line 218: And should be dealing with a list here (so getOpt is probably not the best choice). If you invited me, I accepted, I accidentally left, and you re-invited me, this may not find the 2nd one.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -207,6 +207,28 @@ class LibraryCommander @Inject() (
...
+              libraryInviteRepo.getWithLibraryIdAndUserId(userId = id, libraryId = library.id.get, excludeState = Some(LibraryInviteStates.INACTIVE)).nonEmpty
```
andrewconner, line 219: Needs to filter out `ACCEPTED` and `DECLINED` too. i.e., only allow `ACTIVE`

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -207,6 +207,28 @@ class LibraryCommander @Inject() (
...
+          case None if (passCode.nonEmpty && inviteToken.nonEmpty) =>
+            libraryInviteRepo.getWithLibraryId(libraryId = library.id.get)
+              .exists { i =>
+                HashedPassPhrase.generateHashedPhrase(i.passCode) == passCode.get &&
+                  i.authToken == inviteToken.get
```
andrewconner, line 224: Alternatively, have this condition first, and most of the performance considerations are fixed for now.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
andrewconner, line 221: Decided: query by auth token + libraryId instead.

ymatsuda:
So, you are going to give me an API to get HashedPassPhrase from a request for a given library. Correct?
@andrewconner

andrewconner:
@2is10 @ymatsuda @ahsu1230 

This is how auth is handled from the client's perspective:
1. User "logs in" to library, if necessary. In their Play session cookie, they get this added: ("library_access", "$libraryPublicId/$hashedPassCode")
2. All requests that should be authenticated will use a `LibraryViewAction`, built on Ray's Play 2.3 Action refactor. This action automatically handles the authentication. If authentication fails, it immediately replies with a Forbidden.

Internally, `LibraryViewAction` will extract the libraryId from the request, the authCode from the query string, and the hashedPassCode from the cookie. It will make a request to shoebox to verify access.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
ahsu1230, line 218: definitely need to check state
however I don't think we intended a user to have multiple memberships to same library (only one membership per (user, library) pair)
if accepted, left, re-invited, the membership is still inactive
but if the user accepted the re-invite, the membership will be changed again.

andrewconner:
The jury has decided: authToken and passPhrase

ahsu1230:
:+1:

PR #10981, opened 2014-09-18
There are few cases where eliza is calling shoebox for a normalized uri ...

eishay:
...where the url is clearly not valid.

This pull request comes to guard the inputs and outputs so we can catch it earlier and figure the source.
@stkem @andrewconner

### DIFF
--- a/kifi-backend/modules/eliza/app/com/keepit/eliza/commanders/MessagingCommander.scala
+++ b/kifi-backend/modules/eliza/app/com/keepit/eliza/commanders/MessagingCommander.scala
@@ -3,6 +3,7 @@ package com.keepit.eliza.commanders
 import com.google.inject.Inject
 
 import com.keepit.abook.{ ABookServiceClient }
+import com.keepit.common.net.URI
 import com.keepit.eliza.model._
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/eliza/app/com/keepit/eliza/commanders/MessagingCommander.scala
```diff
@@ -226,7 +229,13 @@ class MessagingCommander @Inject() (
     val userParticipants = (from +: userRecipients).distinct
     val urlOpt = (urls \ "url").asOpt[String]
     val tStart = currentDateTime
-    val nUriOpt = urlOpt.map { url: String => Await.result(shoebox.internNormalizedURI(url, scrapeWanted = true), 10 seconds) } // todo: Remove Await
+    val uri = urlOpt.map { url: String =>
+      URI.parse(url) match {
+        case Success(uri) => uri
+        case Failure(e) => throw new Exception(s"Scan't send message for bad URL: [$url] from $from with title $titleOpt and source $source")
```
stkem: extra "S"

stkem:
lgtm

PR #10983, opened 2014-09-18
docs and timing for fetchFromEmbedly

eishay:
@raymondng @andrewconner @stkem

### DIFF
--- a/kifi-backend/modules/scraper/app/com/keepit/commanders/ScraperURISummaryCommander.scala
+++ b/kifi-backend/modules/scraper/app/com/keepit/commanders/ScraperURISummaryCommander.scala
@@ -1,6 +1,10 @@
 package com.keepit.commanders
 
+import com.keepit.common.performance._
+
 import java.awt.image.BufferedImage
+
+import com.keepit.common.logging.Logging
 
 import scala.concurrent.Future
 import scala.util.{ Failure, Success }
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/scraper/app/com/keepit/commanders/ScraperURISummaryCommander.scala
```diff
@@ -74,34 +78,48 @@ class ScraperURISummaryCommanderImpl @Inject() (
...
+
       val summary = for {
         nUriId <- nUri.id
         embedlyInfo <- embedlyInfoOpt
       } yield {
 
-        callback.syncSavePageInfo(embedlyInfo.toPageInfo(nUriId))
+        //here we call back shoebox, in the middle of shoebox calling us, right?
+        timing(s"[embedly] syncSavePageInfo for uri ${nUriId.id}") {
+          callback.syncSavePageInfo(embedlyInfo.toPageInfo(nUriId))
```
andrewconner, line 93: This is blocking, but the signature of this method is `Future` — should be trivial to swap around? That could be an issue if we're assuming it's non-blocking (for at least part of the time). Thoughts?

kifi-backend/modules/scraper/app/com/keepit/commanders/ScraperURISummaryCommander.scala
```diff
@@ -74,34 +78,48 @@ class ScraperURISummaryCommanderImpl @Inject() (
...
           val nonBlankImages = images.filter { image =>
-            image.url map (ScraperURISummaryCommander.filterImageByUrl(_)) getOrElse false
+            image.url map ScraperURISummaryCommander.filterImageByUrl getOrElse false
           }
           val (smallImages, selectedImageOpt) = partitionImages(nonBlankImages, minSize)
 
-          smallImages.foreach { fetchAndInternImage(nUri, _) }
+          timing(s"[embedly] fetching ${smallImages.size} small images of ${nUriId.id}") {
+            //Why are we fetching all those small images? Why do we do in in sync?
+            smallImages.foreach { fetchAndInternImage(nUri, _) }
```
andrewconner, line 107: This seems like it'll do the fetching asynchronously, but the saving to shoebox is blocking again (`syncSaveImageInfo` in `internImage`)

PR #10986, opened 2014-09-18
new welcome email for all

joshm1:
@stkem 

I took a lot out from the existing welcome email.  can you check I didn't overlook something?


stkem:
:+1:

PR #10987, opened 2014-09-18
new Kifi  friend from Facebook gets different email

joshm1:
@stkem 

this changes the "connection made" email if the auto-friended user is < 24 hours old and the connection is made from FB/LinkedIn


stkem:
lgtm

PR #10995, opened 2014-09-18
Sending all events from authed users through our servers (instead of directly to mixpanel)

stkem:
This is needed in order to augment the events with campaign information for the user and should make our lives a lot easier in the future as all (authed) events now go through one place (pending a change on iOS).

One commit in this PR per component changed (backend, angular site, extension)

Please review @andrewconner @2is10 @leogrim @yipingliao


2is10:
Mixpanel knows about users. Isn’t there a way to tell mixpanel a user’s campaign information just once?

https://mixpanel.com/docs/managing-users/managing-user_specific-properties
https://mixpanel.com/help/reference/javascript#storing-user-profiles

Would be sad to have to repeat it in every event.

kifi-backend/modules/shoebox/app/com/keepit/controllers/tracking/EventProxyController.scala
```diff
@@ -0,0 +1,55 @@
...
+    actionAuthenticator: ActionAuthenticator,
+    heimdal: HeimdalServiceClient,
+    heimdalContextBuilderFactoryBean: HeimdalContextBuilderFactory) extends WebsiteController(actionAuthenticator) with ShoeboxServiceController {
+
+  private def jsValue2SimpleContextData(datum: JsValue): SimpleContextData = {
+    datum match {
+      case JsNumber(x) => ContextDoubleData(x.toDouble)
+      case JsBoolean(b) => ContextBoolean(b)
+      case JsString(s) => ContextStringData(s)
+      case huh => ContextStringData(huh.toString) //unexpected, trying our best not to loose data
```
andrewconner, line 23: JsArray?

stkem:
Unfortunately you can't filter events easily by user properties when doing segmentation/funnels :-(

kifi-backend/modules/shoebox/conf/shoebox.routes
```diff
@@ -59,6 +59,14 @@ GET     /admin/unsubscribe          @com.keepit.controllers.website.EmailOptOutC
 GET     /extmsg/email/mute          @com.keepit.controllers.website.EmailNonUserMuteController.optOut(publicId: String)
 POST    /extmsg/email/mute          @com.keepit.controllers.website.EmailNonUserMuteController.optOutAction(publicId: String)
 
+
+##########################################
+# Tracking
+##########################################
+
+POST   /ev                          @com.keepit.controllers.tracking.EventProxyController.track()
```
2is10: As I mentioned in person, let’s use origin-specific routes, like `/ext/events` for the extension, for flexibility, as with every other API endpoint. They can all point to the same controller for now.

kifi-backend/modules/shoebox/app/com/keepit/controllers/tracking/EventProxyController.scala
stkem, line 23: Top level arrays are dealt with in the function that calls this. Nested arrays are not supported.

packrat/main.js
```diff
@@ -213,15 +213,13 @@ function onGetFail(uri, done, failures, req) {
...
+var tracker = {
   enabled: true,
   queue: [],
   batch: [],
   sendBatch: function () {
     if (this.batch.length > 0) {
-      var json = JSON.stringify(this.batch);
-      var dataString = 'data=' + api.util.btoa(unescape(encodeURIComponent(json)));
-      api.postRawAsForm("https://api.mixpanel.com/track/", dataString);
+      ajax('POST', '/ev', this.batch.slice(0));
```
2is10: no need to `.slice`, `ajax` stringifies synchronously

packrat/main.js
```diff
@@ -213,15 +213,13 @@ function onGetFail(uri, done, failures, req) {
...
-var mixpanel = {
+var tracker = {
   enabled: true,
   queue: [],
   batch: [],
   sendBatch: function () {
     if (this.batch.length > 0) {
-      var json = JSON.stringify(this.batch);
-      var dataString = 'data=' + api.util.btoa(unescape(encodeURIComponent(json)));
-      api.postRawAsForm("https://api.mixpanel.com/track/", dataString);
```
2is10, line 224: the `api.postRawAsForm` implementations can be removed too

kifi-backend/modules/shoebox/conf/shoebox.routes
stkem: done

packrat/main.js
stkem: ah, good to know. removing.

kifi-backend/modules/shoebox/angular/src/common/services/mixpanelService.js
```diff
@@ -38,6 +40,13 @@
     helpRankReKeeps: /^\/helprank\/rekeeps?$/
   };
 
+  function trackEventThroughProxy(event, properties)  {
+    return $http.post(env.origin + '/ev', [{
```
2is10: You shouldn’t need to specify `env.origin`. The browser resolves relative ajax URIs automatically.

2is10:
Minor comments. Otherwise looks good.

PR #11003, opened 2014-09-18
renames (passPhrase, authToken) and changes to canViewLibrary()

ahsu1230:
- rename database columns (everything should be consistent now) @ymatsuda 
- add unique constraint on libraryMemberships (user, library) is unique per membership
- when re-joining an invite, update membership (state, access and "createdAt" fields)
- add LibraryInvite.query by authentication token & library Id

@andrewconner


kifi-backend/modules/common/conf/evolutions/shoebox/250.sql
```diff
@@ -0,0 +1,15 @@
+# SHOEBOX
+
+# --- !Ups
+
+ALTER TABLE library_invite
+    DROP passcode;
+ALTER TABLE library_invite
+    ADD COLUMN pass_phrase varchar(30) NOT NULL;
```
2is10: It would be good to end your `ADD COLUMN` statement with `AFTER auth_token`.

PR #11008, opened 2014-09-18
supporting "dev" certificates for push notifications 

stkem:
When registering, simply send `dev: true` in the json object along with the token.
@eishay @thassss @jaredpetker


kifi-backend/modules/common/conf/prod/eliza.conf
```diff
@@ -40,6 +40,8 @@ db.shoebox {
 urban-airship {
   key = "PCP1hTNTSA6YwXSjT1cSbA"
   secret = "3iQ62IA8Te6m9vXxyLe3Yw"
+  dev_key = ""
```
stkem: @thassss @jaredpetker I'm going to need the actual key and secret for the dev project in urban airship once you have it.

PR #11010, opened 2014-09-19
adding notifications to invites

ahsu1230:
another big one

This is to add kifi notifications & emails to invites
emails go to both users & non-users
notifications only go to users

Thanks @joshm1 for the help on EmailSenders (it's called `LibraryInviteEmailSender`). Please take a look =)
Thanks @andrewconner @stkem @raymondng for the Q&A's

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/emails/LibraryInviteEmailSender.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/emails/LibraryInviteEmailSender.scala
@@ -1,0 +1,41 @@
+package com.keepit.commanders.emails
+
+import com.google.inject.Inject
+import com.keepit.common.db.Id
+import com.keepit.common.db.slick.Database
+import com.keepit.common.healthcheck.AirbrakeNotifier
+import com.keepit.common.logging.Logging
+import com.keepit.common.mail.{ EmailAddress, SystemEmailAddress, ElectronicMail }
+import com.keepit.common.mail.template.EmailToSend
+import com.keepit.common.social.BasicUserRepo
+import com.keepit.model._
+
+import scala.concurrent.Future
+
+class LibraryInviteEmailSender @Inject() (
+    db: Database,
+    emailTemplateSender: EmailTemplateSender,
+    basicUserRepo: BasicUserRepo,
+    libraryRepo: LibraryRepo,
+    protected val airbrake: AirbrakeNotifier) extends Logging {
+
+  def inviteUserToLibrary(toUserRecipient: Either[Id[User], EmailAddress], fromUserId: Id[User], libraryId: Id[Library]): Future[ElectronicMail] = {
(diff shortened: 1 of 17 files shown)


ahsu1230:
merging now to check out library invite emails
happy to address comments in future PR

kifi-backend/modules/shoebox/app/com/keepit/commanders/emails/LibraryInviteEmailSender.scala
```diff
@@ -0,0 +1,41 @@
...
+      (user, library, libOwner)
+    }
+    val libLink = s"""www.kifi.com${Library.formatLibraryPath(libraryOwner.username, libraryOwner.externalId, library.slug)}?auth=${library.universalLink}"""
+    val emailToSend = EmailToSend(
+      fromName = Some(Left(fromUserId)),
+      from = SystemEmailAddress.NOTIFICATIONS,
+      subject = s"${requestingUser.firstName} ${requestingUser.lastName} invited you to follow ${library.name}.",
+      to = toUserRecipient,
+      category = NotificationCategory.User.LIBRARY_INVITATION,
+      htmlTemplate = views.html.email.libraryInvitation(toUserRecipient.left.toOption, fromUserId, inviteMsg, library.name, library.description, libLink),
```
joshm1, line 36: it's not required, but there is a textTemplate that sends a text-only version of the email (for non-html clients)

kifi-backend/modules/shoebox/app/com/keepit/commanders/emails/LibraryInviteEmailSender.scala
```diff
@@ -0,0 +1,41 @@
...
+      val user = basicUserRepo.load(fromUserId)
+      val library = libraryRepo.get(libraryId)
+      val libOwner = basicUserRepo.load(library.ownerId)
+      (user, library, libOwner)
+    }
+    val libLink = s"""www.kifi.com${Library.formatLibraryPath(libraryOwner.username, libraryOwner.externalId, library.slug)}?auth=${library.universalLink}"""
+    val emailToSend = EmailToSend(
+      fromName = Some(Left(fromUserId)),
+      from = SystemEmailAddress.NOTIFICATIONS,
+      subject = s"${requestingUser.firstName} ${requestingUser.lastName} invited you to follow ${library.name}.",
```
joshm1, line 33: subject lines can use the template placeholders are well  `s"{@fullName(fromUserId)} invited you...`.  Therefore you can remove the db request for `requestingUser` (which will be duplicated by the email template processor to generate the email's `fromName` value)

kifi-backend/modules/shoebox/app/com/keepit/commanders/emails/LibraryInviteEmailSender.scala
stkem, line 36: Good point, we should include the textTemplate whenever possible. It can help with passing spam filters.

PR #11011, opened 2014-09-19
an endpoint for loading keep details to show in the keeper

2is10:
It just returns the keep title now, but will soon return the image URL, comment(s), and possibly tags.

@ahsu1230

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
@@ -20,6 +20,7 @@ import com.keepit.model._
 import com.keepit.search.SearchServiceClient
 import com.keepit.social.BasicUser
 
+import play.api.http.Status.{ FORBIDDEN, NOT_FOUND }
 import play.api.libs.concurrent.Execution.Implicits.defaultContext
 import play.api.libs.functional.syntax._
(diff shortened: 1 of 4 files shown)


2is10:
I’m going to merge now so I can start developing with it. Happy to address any comments in a follow-up.

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
```diff
@@ -162,6 +163,19 @@ class KeepsCommander @Inject() (
...
 
+  def getKeep(libraryId: Id[Library], keepExtId: ExternalId[Keep], userId: Id[User]): Either[(Int, String), Keep] = {
+    db.readOnlyMaster { implicit session =>
+      if (libraryMembershipRepo.getWithLibraryIdAndUserId(libraryId, userId).isDefined) {
+        keepRepo.getByExtIdandLibraryId(keepExtId, libraryId) match {
+          case Some(k) => Right(k)
+          case None => Left(NOT_FOUND, "keep_not_found")
+        }
+      } else {
+        Left(FORBIDDEN, "library_access_denied")
```
ahsu1230, line 174: cool! I like this idea!

ahsu1230:
:cake: :+1:
look good, you're totally right about the separate functions in the tests, they really do look much cleaner

PR #11012, opened 2014-09-19
send digest email on force refresh user recos

joshm1:
@stkem

### DIFF
--- a/kifi-backend/modules/curator/app/com/keepit/curator/commanders/email/FeedDigestEmailSender.scala
+++ b/kifi-backend/modules/curator/app/com/keepit/curator/commanders/email/FeedDigestEmailSender.scala
@@ -50,6 +50,33 @@ object DigestEmail {
   val RECO_THRESHOLD = 8
 }
 
+trait RecoRankStrategy {
+  def recommendationsToQuery: Int
+  val minRecommendationsToDeliver: Int = 2
+  val maxRecommendationsToDeliver: Int = 3
+  def ordering: Ordering[UriRecommendation]
+}
+
+object GeneralFeedDigestStrategy extends RecoRankStrategy {
+  val recommendationsToQuery = 100
+
+  object ordering extends Ordering[UriRecommendation] {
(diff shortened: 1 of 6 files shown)


kifi-backend/modules/curator/app/com/keepit/curator/commanders/email/FeedDigestEmailSender.scala
```diff
@@ -50,6 +50,33 @@ object DigestEmail {
...
 }
 
+trait RecoRankStrategy {
+  def recommendationsToQuery: Int
+  val minRecommendationsToDeliver: Int = 2
+  val maxRecommendationsToDeliver: Int = 3
+  def ordering: Ordering[UriRecommendation]
+}
+
+object GeneralFeedDigestStrategy extends RecoRankStrategy {
```
stkem, line 60: Nice!

kifi-backend/modules/curator/app/com/keepit/curator/controllers/internal/CuratorController.scala
```diff
@@ -76,6 +83,44 @@ class CuratorController @Inject() (
 
   def refreshUserRecos(userId: Id[User]) = Action { request =>
     SafeFuture(seedCommander.forceIngestGraphData(userId), Some("Force ingesting Graph Data to refresh Recos"))
+
+    val sendEmailF = shoebox.getExperimentsByUserIds(Seq(userId)) flatMap { usersExperiments =>
```
stkem, line 91: could you have all of this in it's own private method (and just call it here), to keep the actual action simpler?

stkem:
one small structural comment, lgtm otherwise

kifi-backend/modules/curator/app/com/keepit/curator/controllers/internal/CuratorController.scala
stkem, line 91: Also a comment saying that this is a prototype for internal testing would be good, before someone complains about the unusual use of the scheduler :-)

PR #11013, opened 2014-09-19
disable PYMK in welcome email

joshm1:
@stkem 

doesn't include the other changes Jen talked about, she didn't finish/save the asana task with the exact copy changes


stkem:
:+1: 
We should find a good way of turning Tips on and off without having to change tests around too much (but without removing tests)

PR #11016, opened 2014-09-19
Library action auth, LibraryViewAction

andrewconner:
@ahsu1230 @raymondng

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
@@ -229,7 +229,7 @@ class LibraryCommander @Inject() (
       }
   }
 
-  def canViewLibrary(userId: Option[Id[User]], libraryId: Id[Library], accessToken: Option[String], passCode: Option[HashedPassPhrase]) = {
+  def canViewLibrary(userId: Option[Id[User]], libraryId: Id[Library], accessToken: Option[String], passCode: Option[HashedPassPhrase]): Boolean = {
     val library = db.readOnlyReplica { implicit session =>
       libraryRepo.get(libraryId)
     }
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
@@ -229,6 +229,13 @@ class LibraryCommander @Inject() (
       }
   }
 
+  def canViewLibrary(userId: Option[Id[User]], libraryId: Id[Library], accessToken: Option[String], passCode: Option[HashedPassPhrase]) = {
+    val library = db.readOnlyReplica { implicit session =>
+      libraryRepo.get(libraryId)
+    }
+    canViewLibrary(userId, library, accessToken, passCode)
+  }
+
   def copyKeepsFromCollectionToLibrary(libraryId: Id[Library], tagName: Hashtag): Either[LibraryFail, Seq[(Keep, LibraryError)]] = {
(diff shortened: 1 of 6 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -238,6 +238,13 @@ class LibraryCommander @Inject() (
       }
   }
 
+  def canViewLibrary(userId: Option[Id[User]], libraryId: Id[Library], accessToken: Option[String], passCode: Option[HashedPassPhrase]): Boolean = {
```
andrewconner, line 241: @ahsu1230 I added this because it became surprisingly common to want library access, and if the `Library` object is required, then we need to inject a Database everywhere that calls this. That okay?

kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtKeepImageController.scala
```diff
@@ -50,18 +51,36 @@ class ExtKeepImageController @Inject() (
     }
   }
 
-  def loadPrevImageForKeep(libraryId: Id[Library], take: Int, drop: Int) = Action.async { request =>
-    val keeps = db.readOnlyReplica { implicit session =>
-      keepRepo.getByLibrary(libraryId, take, drop)
-    }
+  def loadPrevImageForKeep(startUserId: Long, endUserId: Long, take: Int, drop: Int) = Action.async { request =>
```
andrewconner, line 54: This is migration code. I'll be running it over the next few days (it will take several days to finish). I may throw in a `SystemValue` to keep track of progress.

andrewconner:
Merging so I can run it in prod, reviews appreciated :)

kifi-backend/modules/shoebox/app/com/keepit/shoebox/controllers/LibraryViewActions.scala
```diff
@@ -0,0 +1,83 @@
...
+import com.keepit.common.controller.{ MaybeUserRequest, UserActions, UserRequest }
+import com.keepit.common.crypto.PublicId
+import com.keepit.common.db.Id
+import com.keepit.model.{ HashedPassPhrase, Library, User }
+import play.api.mvc.{ ActionFilter, Controller, Result }
+
+import scala.concurrent.Future
+import scala.util.Success
+
+trait LibraryViewActions { self: UserActions with Controller =>
```
andrewconner, line 16: @raymondng Would you prefer an abstract class or trait with self type?

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
ahsu1230, line 241: Yup! It might just be better to use this function signature instead since any place with a Library already will have the LibraryId. I'll take a look later on

kifi-backend/modules/shoebox/app/com/keepit/shoebox/controllers/LibraryViewActions.scala
raymondng, line 16: No strong preference, but this looks good to me.

andrewconner, line 16: Let's tend towards traits then, so that we can mix several in (since you can only extend one abstract class). In this trait, you can see how I handled DI (instead of passing a parameter, just force that the implementation class defines it).

raymondng, line 16: Agreed. Was thinking the same.

PR #11017, opened 2014-09-19
Following for libraries

andrewconner:
@yipingliao This is a bridge until backend supports telling the client if the user can follow. It's so we can demo this morning the flow.

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/library.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/library.js
@@ -2,7 +2,7 @@
 
 angular.module('kifi')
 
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/library.js
```diff
@@ -10,7 +10,7 @@ angular.module('kifi')
     var username = $routeParams.username;
     var librarySlug = $routeParams.librarySlug;
     var selectedCount = 0;
-
```
andrewconner, line 13: This diff is because there was space on this line. A while back we agreed we would trim empty lines.

In Sublime, you can do this automatically with this in your preferences:

`"trim_trailing_white_space_on_save": true`

yipingliao, line 13: Thanks! Will do.

yipingliao:
Looks good!

PR #11022, opened 2014-09-19
Style (basic) of library share dropdown menu.

yipingliao:
@andrewconner cc @2is10 
An incomplete styling of the library share dropdown menu. Wanted to get this in for the demo. :) Still a bit of style refactoring and clean up to do, which I plan to do when I finish all the styling.

https://app.asana.com/0/15523651812135/15523601193412


kifi-backend/modules/shoebox/angular/src/common/contenteditable.js
```diff
@@ -0,0 +1,44 @@
...
+      }
+
+      // Write data to the model
+      function read() {
+        var html = element.html();
+        // When we clear the content editable the browser leaves a <br> behind
+        // If strip-br attribute is provided then we strip this out
+        if ( attrs.stripBr && html === '<br>' ) {
+          html = '';
+        }
```
2is10, line 26: Different browsers create different HTML in contenteditable elements. Some use `<br>`. Some use `<div></div>` for everything. Better would be to check if `el.textContent.trim()` is non-empty.

By the way, in general, within our own code, I try to avoid introducing options until we have a specific use case for each value of the option.

kifi-backend/modules/shoebox/angular/src/libraries/library.styl
```diff
@@ -1,28 +1,209 @@
...
+  // add the vendor prefixes because the stylus preprocessor is not generating
+  // them correctly for the filter property.
+  -webkit-filter: drop-shadow(0px 0px 7px #afafaf)
+  filter: drop-shadow(0px 0px 7px #afafaf)
+
+  z-index: 2
+
+  .library-share-input
+    box-sizing: border-box
+    border-top: 1px solid rgb(210, 210, 210)
```
2is10, line 51: let’s standardize on hex codes when there’s no alpha transparency

2is10, line 51: Tip: You can Option-click a color in the Chrome dev inspector to toggle between the various color formats.

yipingliao, line 51: Thanks! Will do.

kifi-backend/modules/shoebox/angular/src/common/common.styl
```diff
@@ -0,0 +1,3 @@
+[contenteditable=true]:empty:before{
```
2is10, line 1: We can use the current standard format for pseudo-elements (`::before` with two colons).

Also, use `[contenteditable]`.

> The values "true" and "false" are not allowed on boolean attributes. To represent a false value, the attribute has to be omitted altogether.
> https://html.spec.whatwg.org#boolean-attributes

2is10:
Nice work! Can’t wait to try this out.

Regarding the Gmail logo. Here’s a high quality SVG version of their newer logo that I’m using in the extension, which I got from [here](http://commons.wikimedia.org/wiki/File:New_Logo_Gmail.svg).

Here it is as a data URI in case you want to put it directly in the stylesheet. Pretty small.

```
data:image/svg+xml,<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 45555 33097"><path fill="%23e75a4d" d="M41608 250c2030,0 3691,1661 3691,3691l0 1070 -5128 3663 -17458 12026 -17458 -12148 0 24295 -1314 0c-2030,0 -3691,-1661 -3691,-3691l0 -25214c0,-2030 1661,-3691 3691,-3691l18772 13918 18894 -13918z"/><polygon fill="%23e7e4d7" points="5255,8552 5255,32847 40172,32847 40172,8674 22713,20700"/><polygon fill="%23b8b7ae" points="5255,32847 22713,20700 22634,20644 5255,32580"/><polygon fill="%23b7b6ad" points="40172,8674 40182,32847 22713,20700"/><path fill="%23b2392f" d="M45299 5011l6 24148c-88,2684 -1251,3642 -5123,3687l-10 -24173 5128 -3663z"/><polygon fill="%23f7f5ed" points="3941,250 22713,14168 41608,250"/></svg>
```

yipingliao:
Thanks! I am going to merge this for demo, but will go back and address all your comments.

PR #11026, opened 2014-09-19
Actions: impersonate

raymondng:
Added basic impersonation support (identity/securesocial pending)
Lazy fetch User, Experiments, etc. via thunks
Wire-in `AdminAuthController` for impersonation
Removed `UserAPIController` (just use `UserActions` trait)
Added `ProdRemoteUserActionsHelperModule`

@andrewconner @ymatsuda


raymondng:
Will merge so @ymatsuda can pick it up. Reviews are appreciated!

kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
```diff
@@ -15,23 +17,31 @@ case class NonUserRequest[T](request: Request[T]) extends WrappedRequest(request
...
     request: Request[T],
     userId: Id[User],
-    userF: Future[User],
-    experimentsF: Future[Set[ExperimentType]],
-    kifiInstallationId: Option[ExternalId[KifiInstallation]] = None,
-    adminUserId: Option[Id[User]] = None) extends WrappedRequest(request) with MaybeUserRequest[T] {
-  lazy val user = Await.result(userF, 5 seconds)
-  lazy val experiments = Await.result(experimentsF, 5 seconds)
+    adminUserId: Option[Id[User]],
+    userF: () => Future[User],
```
andrewconner, line 21: Bit worried about this. It re-queries for the user.

Should be either a private field (without accessors) or we legitimize getting the user object that way, and do something like:

``` scala
private lazy val _userF = userActionsHelper.getUserOpt(userId).map(_.get)
def userF = _userF
lazy val user = Await.result(_userF, 5 seconds)
```

Thoughts?

raymondng, line 21: Thanks Andrew. Interesting idea. Agreed that the code (this older version) suffer from the issue you described.
For simplicity, right now I pick the former route (please checkout latest code) as the trait `UserRequest` does not currently expose `userF`. `user` is obviously required for now, for backward compatibility -- will continue to investigate and incorporate your input.

andrewconner, line 21: Also, I noticed that `kifiInstallationId` is not backwards compatible now (it requires adding a `()`). Up to you if you think it's a big enough issue.

raymondng, line 21: Thanks -- good catch! 
Admittedly haven't really looked into `kifiInstallationId` yet -- in fact, it's currently also not exposed (yup still catching up  with old code but thus far not an issue)

PR #11044, opened 2014-09-20
Actions: identity

raymondng:
Added support for `SecureSocial`'s `Identity` (for backward compatibility to aid migration to new framework)
`UserRequest`, etc. are now modeled as traits
`buildNonUserRequest` factored out to `UserActionsHelper` to aid injection of different implementations
Added HTML (i.e. `Page`) support
Wired up `OAuth2Controller` (abook import); validated locally

@andrewconner

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
+++ b/kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
@@ -25,11 +25,12 @@ trait UserRequest[T] extends MaybeUserRequest[T] {
 }
 
 trait SecureSocialIdentityAccess[T] { self: MaybeUserRequest[T] =>
-  def identityOptF: () => Future[Option[Identity]] = () => Future.successful(None)
-  lazy val identityOpt = Await.result(identityOptF.apply, 5 seconds)
+  def identityOpt: Option[Identity]
 }
 
-case class SimpleNonUserRequest[T](request: Request[T]) extends WrappedRequest(request) with NonUserRequest[T] with SecureSocialIdentityAccess[T]
+case class SimpleNonUserRequest[T](request: Request[T]) extends WrappedRequest(request) with NonUserRequest[T] with SecureSocialIdentityAccess[T] {
+  def identityOpt: Option[Identity] = None
+}
 
 case class SimpleUserRequest[T](
     request: Request[T],
@@ -37,10 +38,11 @@ case class SimpleUserRequest[T](
     adminUserId: Option[Id[User]],
     userF: () => Future[User],
     experimentsF: () => Future[Set[ExperimentType]],
-    override val identityOptF: () => Future[Option[Identity]],
+    identityOptF: () => Future[Option[Identity]],
     kifiInstallationId: () => Option[ExternalId[KifiInstallation]]) extends WrappedRequest(request) with UserRequest[T] with SecureSocialIdentityAccess[T] {
   lazy val user = Await.result(userF.apply, 5 seconds)
   lazy val experiments = Await.result(experimentsF.apply, 5 seconds)
+  lazy val identityOpt = Await.result(identityOptF.apply, 5 seconds)
 }
 
 trait UserActionsRequirements {
@@ -137,9 +139,9 @@ trait UserActions extends Logging { self: Controller =>
     resF.map(maybeAugmentCORS(_))
   }
 
-  private def pageAction[P[_]] = new ActionFunction[P, P] {
+  private def PageAction[P[_]] = new ActionFunction[P, P] {
     def invokeBlock[A](request: P[A], block: (P[A]) => Future[Result]): Future[Result] = {
-      block(request).map(_.withHeaders(CONTENT_TYPE -> HTML))
+      block(request).map(_.withHeaders(CONTENT_TYPE -> HTML)) // todo(ray): handle error with redirects
     }
   }
 
@@ -152,7 +154,7 @@ trait UserActions extends Logging { self: Controller =>
       }
     }
   }
-  val UserPage = (UserAction andThen pageAction)
+  val UserPage = (UserAction andThen PageAction)
 
   object MaybeUserAction extends ActionBuilder[MaybeUserRequest] {
     def invokeBlock[A](request: Request[A], block: (MaybeUserRequest[A]) => Future[Result]): Future[Result] = {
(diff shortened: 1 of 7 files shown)


kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
```diff
@@ -5,28 +5,48 @@ import com.keepit.common.db.{ ExternalId, Id }
...
+trait UserRequest[T] extends MaybeUserRequest[T] {
+  def request: Request[T]
+  def userId: Id[User]
+  def adminUserId: Option[Id[User]]
+  def user: User
+  def experiments: Set[ExperimentType]
+}
+
+trait SecureSocialIdentityAccess[T] { self: MaybeUserRequest[T] =>
+  def identityOptF: () => Future[Option[Identity]] = () => Future.successful(None)
```
andrewconner: I'm not sure this is a good pattern. Apologies for not catching it earlier.

``` scala
scala> def t: () => String = () => {println("called"); "done"}
t: () => String

scala> t()
called
res2: String = done

scala> t()
called
res3: String = done

scala> t()
called
res4: String = done
```

Seems like the Future will accidentally get re-generated quite a bit.

kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
```diff
@@ -5,28 +5,48 @@ import com.keepit.common.db.{ ExternalId, Id }
...
+  lazy val identityOpt = Await.result(identityOptF.apply, 5 seconds)
+}
+
+case class SimpleNonUserRequest[T](request: Request[T]) extends WrappedRequest(request) with NonUserRequest[T] with SecureSocialIdentityAccess[T]
+
+case class SimpleUserRequest[T](
```
andrewconner, line 36: Why `Simple` again?

kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
raymondng: Ah good point. In general the lazy val takes care of caching -- but in this case I didn't separate it out. Will take care of it.

kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
raymondng, line 36: Yes I reverted back to modeling using traits (extending `Request`) -- so this is now an implementation detail that most clients shouldn't be concerned with (and in fact I might just hide this from them).

kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
```diff
@@ -116,6 +137,12 @@ trait UserActions extends Logging { self: Controller =>
     resF.map(maybeAugmentCORS(_))
   }
 
+  private def pageAction[P[_]] = new ActionFunction[P, P] {
```
andrewconner: Lowercase `pageAction` (others are uppercase)?

kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
```diff
@@ -116,6 +137,12 @@ trait UserActions extends Logging { self: Controller =>
...
+    def invokeBlock[A](request: P[A], block: (P[A]) => Future[Result]): Future[Result] = {
+      block(request).map(_.withHeaders(CONTENT_TYPE -> HTML))
```
andrewconner: This should be fairly automatic based on what's passed to `Ok` (and etc). Is this a stub for more logic later?

kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
andrewconner, line 36: Will we have non-simple UserRequests?

kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
raymondng: Nothing other than the fact that it's a `def` not an `object` -- do we name functions with upper-case?

kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
raymondng: Actually I find that (e.g. in `TypeaheadAdminController`) it's not 100% automatic so I added explicit support.

kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
andrewconner: Not usually, but it's used like an object. In the Play doc examples, they use uppercase, so the implementation doesn't matter. May be confusing over time to mix it up.

kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
andrewconner: I only see one HTML action there, `index`. Am I missing them?

kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
```diff
@@ -5,28 +5,48 @@ import com.keepit.common.db.{ ExternalId, Id }
...
+  lazy val identityOpt = Await.result(identityOptF.apply, 5 seconds)
+}
+
+case class SimpleNonUserRequest[T](request: Request[T]) extends WrappedRequest(request) with NonUserRequest[T] with SecureSocialIdentityAccess[T]
+
+case class SimpleUserRequest[T](
     request: Request[T],
     userId: Id[User],
     adminUserId: Option[Id[User]],
     userF: () => Future[User],
```
andrewconner, line 40: Will we expose the `Future[User]`, or encourage people to just call `request.user`?

kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
raymondng: Sure I was a bit hesitant about this one as well. Will take care of it.

kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
raymondng: See the last method (`search`) -- this method is hacky though so I don't blame Play ...

kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
raymondng, line 40: We will, but I also want to see if we can discourage even that.

kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
raymondng, line 36: Right now we have non-simple `NonUserRequest` and I like the flexibility afforded by the trait-based approach.

PR #11046, opened 2014-09-20
Introducing DistributedSearchServiceClient

leogrim:
@ymatsuda @eishay 
This is a separate service client to handle intra-cluster communication in Search. This should help keep a lot of search specific logic / classes out of Common. This is an intermediary (working) state because some calls in SearchServiceClient still require the DistributedSearchRouter (e.g. sharingUserInfo). I'll wrap this up after we have migrated to the new search engine / indexes.


eishay:
:+1:  less is common is better!

PR #11050, opened 2014-09-22
create library endpoint

ahsu1230:
@2is10 @andrewconner 

create library endpoint
takes in name & visibility
returns LibraryData

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
@@ -92,47 +92,43 @@ class LibraryCommander @Inject() (
     badMessage match {
       case Some(x) => Left(LibraryFail(x))
       case _ => {
-        val myLibs = db.readOnlyReplica { implicit s =>
-          for (t <- libraryRepo.getByUser(ownerId) if t._1 == LibraryAccess.OWNER) yield {
-            t._2
-          }
-        }
         val validSlug = LibrarySlug(libAddReq.slug)
-        if (myLibs.exists(lib => lib.name == libAddReq.name || lib.slug == validSlug)) {
-          Left(LibraryFail("library_name_or_slug_exists"))
-        } else {
-          val (collaboratorIds, followerIds) = db.readOnlyReplica { implicit s =>
-            val collabs = libAddReq.collaborators.getOrElse(Seq()).map { x =>
-              val inviteeIdOpt = userRepo.getOpt(x) collect { case user => user.id.get }
-              inviteeIdOpt.get
+        db.readOnlyReplica { implicit s => libraryRepo.getByNameOrSlug(ownerId, libAddReq.name, validSlug) } match {
+          case Some(lib) =>
+            Left(LibraryFail("library_name_or_slug_exists"))
+          case None =>
+            val (collaboratorIds, followerIds) = db.readOnlyReplica { implicit s =>
+              val collabs = libAddReq.collaborators.getOrElse(Seq()).map { x =>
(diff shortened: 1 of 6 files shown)


ahsu1230:
right now, takes in a slug as parameter
-> removing that and will use the same "auto-create" slug as on the website

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -97,8 +97,8 @@ class LibraryCommander @Inject() (
             t._2
           }
         }
-        val exists = myLibs.exists(lib => lib.name == libAddReq.name || lib.slug == libAddReq.slug)
-        if (exists) {
+        val validSlug = LibrarySlug(libAddReq.slug)
+        if (myLibs.exists(lib => lib.name == libAddReq.name || lib.slug == validSlug)) {
```
2is10: This check could be done in the database pretty easily, no?

kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtLibraryController.scala
```diff
@@ -97,6 +97,24 @@ class ExtLibraryController @Inject() (
     }
   }
 
+  def addLibrary = JsonAction.authenticatedParseJson { request =>
```
2is10: Okay if we keep these action methods in the same order they appear in the routes file? This would go just under `getLibraries`.

kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtLibraryController.scala
```diff
@@ -97,6 +97,24 @@ class ExtLibraryController @Inject() (
...
+    val body = request.body.as[JsObject]
+    val name = (body \ "name").as[String]
+    val visibility = (body \ "visibility").as[LibraryVisibility]
+    val slug = (body \ "slug").as[String]
+    val addRequest = LibraryAddRequest(name, visibility, description = None, slug, collaborators = None, followers = None)
+    libraryCommander.addLibrary(addRequest, request.userId) match {
+      case Left(fail) => BadRequest(Json.obj("error" -> fail.message))
+      case Right(lib) =>
+        val owner = db.readOnlyMaster { implicit s => basicUserRepo.load(request.userId) }
```
2is10: The `username` and `externalId` should be available directly on `request.user`. No need for a `BasicUser`.

kifi-backend/modules/shoebox/test/com/keepit/controllers/ext/ExtLibraryControllerTest.scala
```diff
@@ -290,6 +290,39 @@ class ExtLibraryControllerTest extends Specification with ShoeboxTestInjector wi
...
+        status(addLibrary(user1, Json.obj("name" -> "Lib/\" 3", "visibility" -> "secret", "slug" -> "lib3"))) === BAD_REQUEST
+
+        // invalid slug
+        status(addLibrary(user1, Json.obj("name" -> "Lib 3", "visibility" -> "secret", "slug" -> "lib  3"))) === BAD_REQUEST
+
+        db.readOnlyMaster { implicit s =>
+          libraryRepo.count === 1
+        }
+      }
+    }
```
2is10: nice test cases. okay if we move them up just under the `getLibraries` tests?

2is10:
Thanks, Aaron! Minor comments. Looks great.

kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtLibraryController.scala
ahsu1230: right I forgot about request.user

PR #11051, opened 2014-09-22
more flexible cortex model version control

yingjieMiao:
@stkem 

I initially wanted to have a more flexible control, e.g. view topic words of different versions. Turns out it's far more complicated than I thought. For example, in the admin LDA page, we can query word's topic and doc topic, but this can only be done for "the active" model. (unless we load several possible models together into memory). Things like user/uri interest score, lda configurations, etc, also significantly adds complexity. 

What's in this PR:
- In `LDACommander` and `LDAInfoCommander`, the ldaVersion is not from injected `wordRepresenter`. It's 'hard coded' to the PublishingLDAVersion. (Decouple of the retrieval version and updating version)
- take `wordRep` and `docRep` out of LDACommander, mainly because they could have different underlying versions than other components in the `LDACommander`. Put them in the same context can be confusing. They now form a `RepresenterCommander`, used for admin only. 
- some clean up/ code reordering. No actual logic change.

### DIFF
--- a/kifi-backend/modules/cortex/app/com/keepit/common/commanders/FeatureRetrievalCommander.scala
+++ b/kifi-backend/modules/cortex/app/com/keepit/common/commanders/FeatureRetrievalCommander.scala
@@ -1,21 +1,24 @@
 package com.keepit.common.commanders
 
 import com.google.inject.{ Inject, Singleton }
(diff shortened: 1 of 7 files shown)


stkem:
lgtm
good start

PR #11055, opened 2014-09-22
Lib share ui part deux

yipingliao:
@andrewconner @2is10 

Finishing up UI styling and interactions for library share search.

Still left to do:
(1) Assets for the Send and Share icons (paper airplane and envelope, respectively). @andrewconner Any luck with cutting assets? I know you were working on it last Friday.
(2) Scroll bar should use our custom scrollbar.
(3) Move common colors and mixins into a global `.styl` file. I need to figure out how to make that file be compiled first so that the rest of the styles can depend on it.

Will circle back to (2) and (3) after working on the rest of this week's tasks since they are a little lower priority.

Thanks!


kifi-backend/modules/shoebox/angular/img/gmail_logo.svg
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" version="1.1" viewBox="0 0 45555 33097">
```
yipingliao: @2is10 I tried the data URI you gave me but it didn't work. Now that I am looking at this, I think that the one I used didn't have the `<g>` tag. I'll try it again.

kifi-backend/modules/shoebox/angular/src/common/util.js
```diff
@@ -9,14 +9,15 @@ angular.module('util', [])
...
+      startsWithCaseInsensitive: function (str, prefix) {
+        return this.startsWith(str.toLowerCase(), prefix.toLowerCase());
+      },
       endsWith: function (str, suffix) {
         return str === suffix || str.indexOf(suffix, str.length - suffix.length) !== -1;
       },
       trimInput: function (input) {
         return input ? input.trim().replace(/\s+/g, ' ') : '';
       },
-      // TODO(yiping): conform this to the test being used by the extension.
```
yipingliao, line 18: Already done. Forgot to take out comment.

kifi-backend/modules/shoebox/angular/src/recos/recos.styl
```diff
@@ -101,14 +101,14 @@
 
 .kf-reco.ng-leave
   opacity: 1
-  transition: 0.5s ease all
+  transition: opacity 0.5s ease
```
yipingliao, line 104: Per earlier code review comment from @2is10, making it clear what property is being transitioned.

kifi-backend/modules/shoebox/angular/src/libraries/library.styl
```diff
@@ -35,51 +55,42 @@
...
-  // them correctly for the filter property.
-  -webkit-filter: drop-shadow(0px 0px 7px #afafaf)
-  filter: drop-shadow(0px 0px 7px #afafaf)
+.library-share-menu-content
+  overflow: hidden
+  border: 1px solid #b5b5b5
+  border-top: none
+  border-bottom-left-radius: 5px
+  border-bottom-right-radius: 5px
+  box-shadow: 0px 4px 8px 0px #afafaf
```
2is10, line 87: nice!

kifi-backend/modules/shoebox/angular/src/libraries/library.styl
```diff
@@ -35,51 +55,42 @@
...
 
-  z-index: 2
+  & > div
+    border-bottom: 1px solid #d2d2d2
```
2is10, line 90: let’s try to avoid just `div` in descendant/child selectors

kifi-backend/modules/shoebox/angular/src/libraries/libraryShareSearch.js
```diff
@@ -66,6 +68,23 @@ angular.module('kifi')
           });
         }
 
+        function emphasizeMatchedPrefix (text, prefix) {
+          if (util.startsWithCaseInsensitive(text, prefix)) {
+            return '<b>' + text.substr(0, prefix.length) + '</b>' + text.substr(prefix.length);
```
2is10, line 73: We always need to HTML-escape user-supplied (unsafe) text when formatting it as HTML. Otherwise we may be opening the door to XSS attacks.

kifi-backend/modules/shoebox/angular/src/libraries/libraryShareSearch.js
```diff
@@ -66,6 +68,23 @@ angular.module('kifi')
           });
         }
 
+        function emphasizeMatchedPrefix (text, prefix) {
+          if (util.startsWithCaseInsensitive(text, prefix)) {
```
2is10, line 72: @andrewconner @stkem We really ought to be returning match indices with typeahead results. Now we have two different JS attempts to identify the matches—one in the extension and one in the website—and I’m afraid neither does exactly what the server is doing. I requested that @raymondng add this 2-3 months ago, but it hasn’t been a priority. Could we make it one now?

2is10:
Nice work!

kifi-backend/modules/shoebox/angular/src/libraries/libraryShareSearch.js
andrewconner, line 72: Yes. Do you want HTML or markers for formatting purposes?

kifi-backend/modules/shoebox/angular/src/libraries/library.styl
yipingliao, line 90: Will do.

kifi-backend/modules/shoebox/angular/src/libraries/libraryShareSearch.js
yipingliao, line 73: I am using `ngBindHtml` to output this text in the template, which sanitizes the string. I _think_ that should be okay.
https://docs.angularjs.org/api/ng/directive/ngBindHtml
https://docs.angularjs.org/api/ngSanitize/service/$sanitize

2is10, line 73: Oh, cool. Most templating systems don’t have something so fancy.

PR #11056, opened 2014-09-23
Implement library search similarly to kifi search

leogrim:
@ymatsuda 
I'll likely have to refine the scoring logic once I get the endpoint to work. It's also not super DRY, I think I can probably reorganize it a bit to share more code with keep search.


ymatsuda:
lgtm!

PR #11057, opened 2014-09-23
factor out AdminUserActions

raymondng:
Factored out `AdminUserActions` to make `UserActions` a bit more light weight, and to facilitate migration from old framework

@andrewconner


andrewconner:
Nice

PR #11059, opened 2014-09-23
increase ttl for user cache to one week so it would survive the weekend and few inactivity days

eishay:
@andrewconner


eishay:
now we're at "hits 391909 , misses 72541, sets 967" seems to be a "not enough sets" problem. added some `readFromMaster` calls

eishay:
@ahsu1230

PR #11060, opened 2014-09-23
Fix staggering of connection cards on /invite.

yipingliao:
@2is10 Hi! Can you take a look at this small PR? It fixes a display issue Eishay brought up yesterday.

Thanks!
Screenshot in https://app.asana.com/0/2295914802884/16460795118057

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/invite/connectionCard.styl
+++ b/kifi-backend/modules/shoebox/angular/src/invite/connectionCard.styl
@@ -83,8 +83,8 @@
 
   .kf-connection-card-who
 
+    float: left
     padding-top: 20px
-    display: inline-block
     width: 60%
 
     .kf-main-label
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/angular/src/invite/connectionCard.tpl.html
```diff
@@ -1,11 +1,11 @@
 <div class="kf-connection-card" ng-class="{'kf-connection-invited': invited}" ng-hide="hidden"> <!-- Should really be ng-if but that causes some error for some reason -->
   <img class="kf-connection-card-image" onerror="this.src='https://www.kifi.com/assets/img/ghost.100.png'" ng-src="{{mainImage}}" ng-class="{'kf-connection-card-email-image': email}">
-  <span class="kf-connection-card-who">
+  <div class="kf-connection-card-who">
```
yipingliao, line 3: This doesn't help with the bug, but it makes more sense to me to have the spans be contained in a div, since the spans are stacked on top of each other.

kifi-backend/modules/shoebox/angular/src/invite/connectionCard.styl
```diff
@@ -83,8 +83,8 @@
 
   .kf-connection-card-who
 
+    float: left
     padding-top: 20px
-    display: inline-block
```
yipingliao, line 87: `inline-block` was causing problems because the baseline of the text moves up and down depending on which lines in `.kf-connection-card-who` are being displayed, causing a staggering effect.

2is10, line 87: adding `vertical-align: top` to this rule should fix it without floating

kifi-backend/modules/shoebox/angular/src/invite/connectionCard.tpl.html
2is10, line 3: `div` sounds good

kifi-backend/modules/shoebox/angular/src/invite/connectionCard.styl
yipingliao, line 87: Yes, thank you! :)

PR #11061, opened 2014-09-23
changing create library route to POST /ext/libraries

2is10:
@ahsu1230 

This is the dominant convention for creating a resource when PUT is not applicable. That is, the resource is not fully defined by the request (e.g. ID will be server generated). To verify this, you can search the web for “REST conventions”, visit any of the top results, and then search the page for “creat”.


2is10:
merging so I can use it, comments welcome

ahsu1230:
:+1:

PR #11066, opened 2014-09-23
removing userScoreCache usage from Graph (no cache hits)

ahsu1230:
report has been showing up as 100% cache misses on GRAPH for `user_connection_score`.
But really, this is probably a reporting error on GRAPH's side.

@yingjieMiao found that even though there are no cache hits on GRAPH, there are `user_connection_score` usages in CURATOR (which only has a 5% miss). When CURATOR can't find the user score it wants in the cache, it will call a method from GRAPH to retrieve that info.
However, currently in GRAPH, it checks the `user_connection_score` cache (which doesn't have it anyway) and will always cause cache misses and never hits.

So, it might be best to keep the cache reporting on CURATOR (who is really the only one using it anyway) and take the cache gets/sets out of GRAPH.

@yingjieMiao did I miss anything?
@eishay

### DIFF
--- a/kifi-backend/modules/graph/app/com/keepit/graph/commanders/GraphCommander.scala
+++ b/kifi-backend/modules/graph/app/com/keepit/graph/commanders/GraphCommander.scala
@@ -38,8 +38,6 @@ class GraphCommander @Inject() (
     val urisList = getUriScoreList(startingVertexId, journal, avoidFirstDegreeConnection)
     val usersList = getUsersScoreList(startingVertexId, journal, avoidFirstDegreeConnection)
     uriScoreCache.set(ConnectedUriScoreCacheKey(userId, avoidFirstDegreeConnection), urisList)
-    userScoreCache.set(ConnectedUserScoreCacheKey(userId, avoidFirstDegreeConnection), usersList)
-    log.error(s"Cache Set on UserScoreCache Set ($userId, $avoidFirstDegreeConnection)")
     (urisList, usersList)
   }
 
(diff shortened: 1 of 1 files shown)


yingjieMiao:
you want to modify two methods:
`getConnectedUriScores` and `getConnectedUserScores`. 
You want to remove the read-from-cache step from graph. 
You need to keep the set-cache on graph.

kifi-backend/modules/graph/app/com/keepit/graph/commanders/GraphCommander.scala
```diff
@@ -38,8 +38,6 @@ class GraphCommander @Inject() (
     val urisList = getUriScoreList(startingVertexId, journal, avoidFirstDegreeConnection)
     val usersList = getUsersScoreList(startingVertexId, journal, avoidFirstDegreeConnection)
     uriScoreCache.set(ConnectedUriScoreCacheKey(userId, avoidFirstDegreeConnection), urisList)
-    userScoreCache.set(ConnectedUserScoreCacheKey(userId, avoidFirstDegreeConnection), usersList)
```
yingjieMiao, line 41: better to set this? I understand it will be set on the curator side. I think it's still graph's responsibility to keep the cache fresh and up.

eishay, line 41: thanks!

ahsu1230:
okay, wasn't sure on getConnectedUriScores since the cache key `user_feed_score` hasn't shown up on any error reports so didn't want to touch it =)

ahsu1230:
thanks for the help!

yingjieMiao:
ah, you have a point. let's see if anything changes.

PR #11068, opened 2014-09-23
Utility function to show/hide a recommendations score breakdown

stkem:
to be called from the console only for debugging
(specifically, running `$(".recos-view").scope().toggleExplain(); $(".recos-view").scope().$digest();` in the bowser console while on the recommendations page will toggle between showing the page description and the score breakdown in the card)

@yipingliao @yingjieMiao


andrewconner:
Mind throwing in a comment with that in the source?

stkem:
Good idea; will do that.

yingjieMiao:
lgtm

PR #11071, opened 2014-09-23
use less steps

yingjieMiao:
@stkem


stkem:
lgtm, but why the change?

yingjieMiao:
reduce email noise. Lots of long wait errors b/c of calling these endpoints. If you think these scores need to be "simulated with high confidence", then we can stick with 100K. In some sense, 100K is a rule of thumb, and the actual probabilities over "dominating vertexes" seem to converge faster.

stkem:
no, the reduction is fine, I was just wandering what prompted it. go ahead and merge it.

PR #11073, opened 2014-09-23
Actions: AdminControllers

raymondng:
Migrated all `AdminController`s to the new framework
Mostly mechanical -- a few required additional work due to explicit dependencies on `AuthenticatedRequest`, etc. (fortunately as `UserRequest` is in general backward compatible, it's not too bad)
All tests passed!

Would appreciate a quick review. I have another pending PR waiting for this going in.

@andrewconner 
cc: @eishay


andrewconner:
Awesome, will go through it.

kifi-backend/modules/shoebox/test/com/keepit/controllers/admin/AdminDashboardControllerTest.scala
```diff
@@ -69,9 +72,11 @@ class AdminDashboardControllerTest extends Specification with ShoeboxApplication
           u1
         }
 
+        val userActionsHelper = inject[UserActionsHelper].asInstanceOf[FakeUserActionsHelper]
+        userActionsHelper.setUser(u1, Set(ADMIN))
         val cookie = Authenticator.create(su).right.get.toCookie
         val fakeRequest = FakeRequest().withCookies(cookie)
-        val authRequest = AuthenticatedRequest(null, u1.id.get, u1, fakeRequest)
+        val authRequest = SimpleUserRequest(fakeRequest, u1.id.get, None, () => Future.successful(u1), () => Future.successful(Set(ADMIN)), () => Future.successful(None), () => None)
```
andrewconner, line 79: Why is this exposed? Mind walking me through why `Simple*` was used to begin with? I'm worried it expands the mental model people will have to have when using the actions, since it's not an implementation detail.

raymondng, line 79: Ha you caught this one :) Yes ideally I agree this shouldn't be exposed but if it's okay I'll fix this one later -- knee deep in another bigger PR that deals with removal of `AdminControllers`.

raymondng:
Will merge and build upon this. Review feedback appreciated!

PR #11074, opened 2014-09-23
KeepImageRequest

andrewconner:
This lets the extension pass in an `imageUrl` when keeping, and wires up `KeepImage` when fetching the keep. Furthermore, adds a request tracking table so we can have a record of image fetches and what happened.

@2is10 @ahsu1230


andrewconner:
Moves to new Actions as well.

raymondng:
:+1:

andrewconner:
Merging to get this in, but feedback appreciated.

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepImageHelper.scala
```diff
@@ -149,8 +152,17 @@ object ImageProcessState {
...
-  case class ExistingStoredImagesFound(images: Seq[KeepImage]) extends ImageProcessState with ImageProcessDone
+  case object StoreSuccess extends ImageProcessState with ImageProcessSuccess
+  case class ExistingStoredImagesFound(images: Seq[KeepImage]) extends ImageProcessState with ImageProcessSuccess
+
+  def stateToResponse(state: ImageProcessDone) = {
+    state match {
+      case fail: KeepImageStoreFailure =>
+        InternalServerError(Json.obj("error" -> fail.reason))
+      case success: ImageProcessSuccess =>
+        Ok(JsString("success"))
```
2is10, line 163: Okay if we move this state-to-response translation to the controller? As I understand it, response creation/formatting is a controller concern.

kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtLibraryController.scala
```diff
@@ -68,30 +73,66 @@ class ExtLibraryController @Inject() (
             hcb += ("guided", true)
           }
           implicit val context = hcb.build
+
           val rawBookmark = info.as[RawBookmarkRepresentation]
           val keepInfo = keepsCommander.keepOne(rawBookmark, request.userId, libraryId, request.kifiInstallationId, source)
+
+          // Determine image choice.
+          val imageStatus = (info \ "imageUrl").asOpt[String] match {
```
2is10, line 81: Would be good to use either `"imageUrl"` or `"image"` consistently in the extension APIs. `LateLoadKeepData` uses `"image"`. I could go with either—I like both brevity and precision. I guess it should be clear that `"image"` will be a URL because there’s no other good way to represent an image in JSON.

kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtLibraryController.scala
```diff
@@ -68,30 +73,66 @@ class ExtLibraryController @Inject() (
...
+            case Some(imageUrl) if imageUrl.startsWith("http") =>
+              val (keep, keepImageRequest) = db.readWrite { implicit session =>
+                val keep = keepRepo.getOpt(keepInfo.id.get).get // Weird pattern, but this should always exist.
+                val keepImageRequest = keepImageRequestRepo.save(KeepImageRequest(keepId = keep.id.get, source = KeepImageSource.UserPicked))
+                (keep, keepImageRequest)
+              }
+              keepImageCommander.setKeepImageFromUrl(imageUrl, keep.id.get, KeepImageSource.UserPicked, Some(keepImageRequest.id.get))
+              Json.obj("imageStatusPath" -> keepImageRequest.token) // url
```
2is10, line 89: TODO: path instead of just token?

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepImageHelper.scala
andrewconner, line 163: Sure. I put it here since it will be likely commonly used across clients. Can factor it out somewhere else when needed.

kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtLibraryController.scala
andrewconner, line 81: Sure.

kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtLibraryController.scala
andrewconner, line 89: Yep. Missed that.

kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtLibraryController.scala
```diff
@@ -1,19 +1,19 @@
 package com.keepit.controllers.ext
 
 import com.google.inject.Inject
-import com.keepit.commanders.{ KeepData, KeepsCommander, RawBookmarkRepresentation, LibraryCommander, LibraryData, LibraryAddRequest }
-import com.keepit.common.controller.{ ShoeboxServiceController, BrowserExtensionController, ActionAuthenticator }
+import com.keepit.commanders.{ KeepData, KeepsCommander, LibraryAddRequest, LibraryCommander, LibraryData, RawBookmarkRepresentation, _ }
```
2is10, line 4: Did you mean to add a `_` at the end?

2is10:
Any chance we could get a `ExtKeepImageControllerTest`?

`ExtLibraryControllerTest` has been quite helpful, describing and validating the other extension library endpoints.

PR #11077, opened 2014-09-23
counting others

ymatsuda:
@leogrim 
- added otherDiscoverableKeeps to AugmentedItem and AugmentationInfo
- this is needed for the backward compatibility support...


leogrim:
:+1: 
We need to check with Mark what we're supposed to do with this unicity issue - counting unique keepers is hard. I think it's also relevant for @2is10

PR #11091, opened 2014-09-24
user attribution with library

yingjieMiao:
@stkem  @leogrim


stkem:
lgtm, sorry for the delay

PR #11092, opened 2014-09-24
adding numFollowers to LibraryInfo

ahsu1230:
@andrewconner 

@stkem once deployed, we should be able to have "number of followers" per library for the front-end

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/model/LibraryRepo.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/model/LibraryRepo.scala
@@ -21,7 +21,7 @@ trait LibraryRepo extends Repo[Library] with SeqNumberFunction[Library] {
   def getBySlugAndUserId(userId: Id[User], slug: LibrarySlug, excludeState: Option[State[Library]] = Some(LibraryStates.INACTIVE))(implicit session: RSession): Option[Library]
   def getByNameOrSlug(userId: Id[User], name: String, slug: LibrarySlug, excludeState: Option[State[Library]] = Some(LibraryStates.INACTIVE))(implicit session: RSession): Option[Library]
   def getOpt(ownerId: Id[User], slug: LibrarySlug)(implicit session: RSession): Option[Library]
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/app/com/keepit/model/LibraryRepo.scala
```diff
@@ -119,9 +119,11 @@ class LibraryRepoImpl @Inject() (
     getOptCompiled(ownerId, slug).firstOption
   }
 
-  def updateMemberCount(libraryId: Id[Library])(implicit session: RWSession): Unit = {
-    sqlu"UPDATE library SET member_count = (SELECT COUNT(id) FROM library_membership WHERE library_id = ${libraryId} and state='active') WHERE id = ${libraryId} AND state='active'".execute()
+  def updateMemberCount(libraryId: Id[Library])(implicit session: RWSession): Int = {
+    val newCnt = sql"""SELECT COUNT(id) FROM library_membership WHERE library_id = ${libraryId} and state='active'""".as[Int].first
```
andrewconner, line 123: Selecting first, then updating?

ahsu1230, line 123: yea I wanted the function to return an integer value
but it doesn't seem UPDATE returns an int
so I thought it would hopefully be the same thing if I selected, then updated. Does that work?

ahsu1230, line 123: woops sorry didn't see your comment before merging

2is10, line 123: If we’re going to do two queries, seems like it would be cleaner to just eliminate this method and instead use `LibraryRepo.save` and a simple count method in `LibraryMembershipRepo`.

``` scala
val updatedLib = libraryRepo.save(lib.copy(memberCount = membershipRepo.count(libraryId)))
```

ahsu1230, line 123: yea I think so too
will be coming in next PR

PR #11103, opened 2014-09-24
Actions: misc improvements

raymondng:
Internal house-cleaning -- clients not affected
Overall aimed for a simpler design with fewer moving parts based on Andrew's feedback
Current design continue to focus on backward-compatibility requirements 
Follow-up PR will (hopefully) do better.

@andrewconner


raymondng:
Merging to build upon it. Feedback appreciated!

PR #11106, opened 2014-09-24
fix library left pane bar - correct num followers & fix on site & couple...

ahsu1230:
... of library SVGs used

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/layout/nav/navLibs.styl
+++ b/kifi-backend/modules/shoebox/angular/src/layout/nav/navLibs.styl
@@ -34,3 +34,10 @@
   
 .kf-lib-input-box
   padding: 14px 0 10px 0
+
+.library-card
+  width: 27px
+  height: 35px
+  display: inline-block
+  vertical-align: top
+  background-repeat: no-repeat
--- a/kifi-backend/modules/shoebox/angular/src/tags/tags.js
+++ b/kifi-backend/modules/shoebox/angular/src/tags/tags.js
@@ -74,7 +74,9 @@ angular.module('kifi')
         });
 
         function setTagListHeight() {
-          scrollableTagList.height(w.height() - (scrollableTagList.offset().top - w[0].pageYOffset));
+          if (scrollableTagList.offset() === {}) {
+            scrollableTagList.height(w.height() - (scrollableTagList.offset().top - w[0].pageYOffset));
+          }
         }
         $timeout(setTagListHeight);
 
@@ -288,7 +290,7 @@ angular.module('kifi')
 
         scope.$watch('filter.name', function () {
           $timeout(scope.refreshHighlight);
-          scope.refreshScroll();
+          scope.refreshScroll && scope.refreshScroll();
         });
 
         tagService.fetchAll();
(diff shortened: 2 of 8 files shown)


kifi-backend/modules/shoebox/angular/src/tags/tags.js
```diff
@@ -74,7 +74,9 @@ angular.module('kifi')
         });
 
         function setTagListHeight() {
-          scrollableTagList.height(w.height() - (scrollableTagList.offset().top - w[0].pageYOffset));
+          if (scrollableTagList.offset() === {}) {
```
yipingliao, line 77: If `scrollableTagList.offset()` is an empty object, how will it have a `top` property?

kifi-backend/modules/shoebox/angular/src/layout/nav/navLibs.styl
```diff
@@ -34,3 +34,10 @@
   
 .kf-lib-input-box
   padding: 14px 0 10px 0
+
+.library-card
+  width: 27px
+  height: 35px
+  display: inline-block
```
yipingliao, line 41: List positioning and display properties first.

kifi-backend/modules/shoebox/angular/src/tags/tags.js
ahsu1230, line 77: ah crap mistype - but the errors don't show anymore

kifi-backend/modules/shoebox/angular/src/layout/nav/navLibs.styl
ahsu1230, line 41: sure,
I guess display takes precedence over width&height even though they are all part of the BOX property https://smacss.com/book/formatting#grouping

kifi-backend/modules/shoebox/angular/src/tags/tags.js
yipingliao, line 77: I wonder why it's working...

2is10, line 77: that will always evaluate to `false`

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryMembershipRepo.scala
```diff
@@ -24,6 +24,7 @@ trait LibraryMembershipRepo extends Repo[LibraryMembership] with RepoWithDelete[
     excludeState: Option[State[LibraryMembership]] = Some(LibraryMembershipStates.INACTIVE))(implicit session: RSession): Seq[LibraryMembership]
   def countWithLibraryIdAndAccess(libraryId: Id[Library], accessSet: Set[LibraryAccess],
     excludeState: Option[State[LibraryMembership]] = Some(LibraryMembershipStates.INACTIVE))(implicit session: RSession): Int
+  def countMemberships(libraryId: Id[Library], excludeState: Option[State[LibraryMembership]] = Some(LibraryMembershipStates.INACTIVE))(implicit session: RSession): Int
```
2is10, line 27: `Memberships` is redundant, implied by the repo type. Consider just `count` (argument makes it clear that it’s for a particular library) or `countWithLibraryId` (like the method above).

PR #11110, opened 2014-09-25
few changes (lib card color corrections, basic-scrollbar)

ahsu1230:
- adding the correct SVGs for Library Cards
- have a basic scrollbar for now - will definitely refine later
- some small refactoring
  @andrewconner @yipingliao @2is10


kifi-backend/modules/shoebox/app/com/keepit/model/LibraryMembershipRepo.scala
```diff
@@ -24,7 +24,7 @@ trait LibraryMembershipRepo extends Repo[LibraryMembership] with RepoWithDelete[
     excludeState: Option[State[LibraryMembership]] = Some(LibraryMembershipStates.INACTIVE))(implicit session: RSession): Seq[LibraryMembership]
   def countWithLibraryIdAndAccess(libraryId: Id[Library], accessSet: Set[LibraryAccess],
     excludeState: Option[State[LibraryMembership]] = Some(LibraryMembershipStates.INACTIVE))(implicit session: RSession): Int
-  def countMemberships(libraryId: Id[Library], excludeState: Option[State[LibraryMembership]] = Some(LibraryMembershipStates.INACTIVE))(implicit session: RSession): Int
+  def countWithLibraryId(libraryId: Id[Library], excludeState: Option[State[LibraryMembership]] = Some(LibraryMembershipStates.INACTIVE))(implicit session: RSession): Int
```
2is10, line 27: Thanks! Minor thing, I probably should have said nothing.

ahsu1230, line 27: No, they're useful, Jared! Thanks for the feedback - really. Having good practice is important.
I was thinking of renaming to `count` but would override the original DB `count` function.
It's definitely smaller & more concise, but I thought it would be clearer to keep both functions separate names

andrewconner, line 27: You can make it count without overriding. It has a different signature.

PR #11112, opened 2014-09-25
Some tests for ExtKeepImageController. More to come.

andrewconner:
@2is10 Sorry about the bug, it's fixed.

I'm still working on full test coverage for ExtKeepImageController — specifically, tests that fetch remote images. Right now we're using Play's WS client, which doesn't have a Fake version (so I can't give it a file directly). They allow you to create your own, but I need to make sure we have backwards compatibility with our current configs.


kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtLibraryController.scala
```diff
@@ -13,6 +13,7 @@ import com.keepit.model._
 import play.api.libs.json._
 import play.api.mvc.Result
 
+import scala.concurrent.Future
```
2is10, line 16: let’s see your invitation, buddy

kifi-backend/modules/shoebox/conf/shoebox.routes
```diff
@@ -123,7 +123,8 @@ POST    /ext/libraries              @com.keepit.controllers.ext.ExtLibraryContro
 POST    /ext/libraries/:id/keeps    @com.keepit.controllers.ext.ExtLibraryController.addKeep(id: PublicId[Library])
 GET     /ext/libraries/:id/keeps/:k @com.keepit.controllers.ext.ExtLibraryController.getKeep(id: PublicId[Library], k: ExternalId[Keep], imgSize: Option[String] = None)
 GET     /ext/libraries/:id/keeps/:k/image/:token @com.keepit.controllers.ext.ExtKeepImageController.checkImageStatus(id: PublicId[Library], k: ExternalId[Keep], token: String)
-POST    /ext/libraries/:id/keeps/:k/image @com.keepit.controllers.ext.ExtKeepImageController.uploadKeepImage(id: PublicId[Library], k: ExternalId[Keep])
+POST    /ext/libraries/:id/keeps/:k/imageFile @com.keepit.controllers.ext.ExtKeepImageController.uploadKeepImage(id: PublicId[Library], k: ExternalId[Keep])
```
2is10, line 126: What if this route were instead:
`POST  /ext/libraries/:id/keeps/:k/image/:token`
This would allow the server to associate an uploaded image with an earlier `addKeep` (or `changeKeepImage`) request and a particular URL. Conceivably the user could set a keep image twice quickly, and the uploads could finish out of order. The order of the original requests (`KeepImageRequest.createdAt`?) would resolve the race condition.

2is10, line 126: Nevermind that, but for the route, what do you think of:
`POST    /ext/libraries/:id/keeps/:k/image/upload`

kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtKeepImageController.scala
```diff
@@ -34,9 +34,9 @@ class ExtKeepImageController @Inject() (
       (k, lib)
```
2is10, line 34: Why load the keep twice, and why call it `lib` instead of `keep` the second time?

kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtKeepImageController.scala
```diff
@@ -90,6 +88,36 @@ class ExtKeepImageController @Inject() (
...
 
+  def changeKeepImage(libraryPubId: PublicId[Library], keepExtId: ExternalId[Keep]) = UserAction.async(parse.tolerantJson) { request =>
+    val (keepOpt, libOpt) = db.readOnlyMaster { implicit session =>
+      val k = keepRepo.getOpt(keepExtId)
+      val lib = Library.decodePublicId(libraryPubId).toOption.flatMap(libId => keepRepo.getByExtIdandLibraryId(keepExtId, libId))
+      (k, lib)
+    }
+    val imageJson = request.body \ "image"
+
+    (keepOpt, libOpt, imageJson) match {
```
2is10, line 99: Same issue in this method. The value currently in `libOpt` should replace the value in `keepOpt`.

kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtKeepImageController.scala
```diff
@@ -90,6 +88,36 @@ class ExtKeepImageController @Inject() (
     checkStatus().map(Future.successful).getOrElse(pollCheck())
   }
 
+  def changeKeepImage(libraryPubId: PublicId[Library], keepExtId: ExternalId[Keep]) = UserAction.async(parse.tolerantJson) { request =>
+    val (keepOpt, libOpt) = db.readOnlyMaster { implicit session =>
+      val k = keepRepo.getOpt(keepExtId)
+      val lib = Library.decodePublicId(libraryPubId).toOption.flatMap(libId => keepRepo.getByExtIdandLibraryId(keepExtId, libId))
+      (k, lib)
+    }
+    val imageJson = request.body \ "image"
```
2is10, line 97: Probably a good idea to parse the URL here, and change `KeepImageCommander.setKeepImageFromUrl` to take a `com.keepit.common.net.URI`. This way this action will respond to bad URLs with a 400 and a clear explanation. We can Airbrake on those if we want to detect bad extension behavior or potentially improve our `URI` parser.

kifi-backend/modules/shoebox/test/com/keepit/controllers/ext/ExtKeepImageControllerTest.scala
```diff
@@ -0,0 +1,222 @@
...
+          (user1, user2, lib, mem1, mem2)
+        }
+        implicit val config = inject[PublicIdConfiguration]
+        val libPubId = Library.publicId(lib.id.get)
+
+        val keepResp = addKeep(user1, libPubId, Json.obj("url" -> "http://www.foo.com", "title" -> "Foo"))
+        status(keepResp) === OK
+        val keepJson = contentAsJson(keepResp)
+        val keepExtId = (keepJson \ "id").as[ExternalId[Keep]]
+        val libraryPubId = (keepJson \ "libraryId").as[PublicId[Library]]
```
2is10, line 67: This is the same as `libPubId` above, no?

kifi-backend/modules/shoebox/test/com/keepit/controllers/ext/ExtKeepImageControllerTest.scala
```diff
@@ -0,0 +1,222 @@
...
+
+        db.readOnlyMaster { implicit session =>
+          keepImageRepo.all().length === 0
```
2is10, line 70: Small changes in our assertions can sometimes result in much more helpful failure messages.

``` scala
keepImageRepo.all() === Seq.empty
```

This will print any unexpected objects that are in the repo in a failure message, instead of just `"1 is not 0"`.

kifi-backend/modules/shoebox/test/com/keepit/controllers/ext/ExtKeepImageControllerTest.scala
```diff
@@ -0,0 +1,222 @@
...
+    inject[ExtKeepImageController].checkImageStatus(libraryId, keepExtId, token)(request(routes.ExtKeepImageController.checkImageStatus(libraryId, keepExtId, token)))
+  }
+
+  private def request(route: Call) = FakeRequest(route.method, route.url)
+
+  private def fakeFile = {
+    val tf = TemporaryFile(new File("test/data/image1-" + Math.random() + ".png"))
+    tf.file.deleteOnExit()
+    FileUtils.copyFile(new File("test/data/image1.png"), tf.file)
+    tf
```
2is10, line 220: Could we make the upload action a thin wrapper around a helper method that takes an `Iteratee[Byte]` (or something like that) instead of directly accessing `request.body` as a file? Then if we just test the helper method, we could avoid the disk accesses here, for speed.

2is10:
Nice work. Merge whenever you like.​

kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtKeepImageController.scala
```diff
@@ -90,6 +88,36 @@ class ExtKeepImageController @Inject() (
     checkStatus().map(Future.successful).getOrElse(pollCheck())
```
2is10, line 88: We probably only need to check that the keep exists in the library once.

2is10:
I’m going to merge for the fix.

kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtKeepImageController.scala
```diff
@@ -90,6 +88,36 @@ class ExtKeepImageController @Inject() (
...
+    }
+    val imageJson = request.body \ "image"
+
+    (keepOpt, libOpt, imageJson) match {
+      case (None, _, _) | (_, None, _) =>
+        Future.successful(NotFound(Json.obj("error" -> "keep_not_found")))
+      case (Some(keep), _, JsNull) =>
+        keepImageCommander.removeKeepImageForKeep(keep.id.get)
+        Future.successful(NoContent)
+      case (Some(keep), _, JsString(imageUrl)) if imageUrl.startsWith("http") =>
```
2is10, line 105: We should probably also detect here whether `imageUrl` is already one or our keep image URLs, and if so, bypass downloading it from ourselves, calculating MD5, etc. This can happen if, for example, the user views an existing keep, changes the keep image, then changes it back.

PR #11119, opened 2014-09-25
Actions: lazily

raymondng:
Added simple helper class `Lazily` which (lazily) caches both `Future` and value

Wired-up `Lazily` to consistently expose various attributes (`user`, `experiments`, etc.)

This brings the implementation closer to the spec

@andrewconner


andrewconner:
Cool, nice

PR #11122, opened 2014-09-25
Keep image improvements

andrewconner:
@2is10 Based on your feedback. Still haven't removed the need for a File — it's tricky.


2is10:
:+1:

PR #11125, opened 2014-09-25
trimming user names from the login form auth request

eishay:
@raymondng @thassss @jaredpetker @andrewconner @2is10

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/social/providers/UsernamePasswordProvider.scala
+++ b/kifi-backend/modules/common/app/com/keepit/social/providers/UsernamePasswordProvider.scala
@@ -27,12 +27,13 @@ class UsernamePasswordProvider(app: Application)
     UPP.loginForm.bindFromRequest().fold(
       errors => Left(error("bad_form")),
       credentials => {
-        UserService.find(IdentityId(credentials._1, id)) match {
+        val (username, password) = credentials
+        UserService.find(IdentityId(username.trim, id)) match {
           case Some(identity) =>
             identity match {
               case userIdentity: UserIdentity =>
                 log.info(s"[doAuth] userIdentity=$userIdentity")
-                if (passwordAuth.authenticate(userIdentity.userId.get, credentials._2)) Right(userIdentity) else Left(error("wrong_password"))
+                if (passwordAuth.authenticate(userIdentity.userId.get, password)) Right(userIdentity) else Left(error("wrong_password"))
               case _ =>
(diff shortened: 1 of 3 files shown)


andrewconner:
Cool. Do we trim when they register too?

raymondng:
lgtm! 

I think at this point we are pretty much only dealing with emails ... how about making this assumption explicit and use `EmailAddress` here?

eishay:
@raymondng Sure, good idea and it would help us catch some problems.
@andrewconner no idea, I'll check

2is10:
:+1: I know the JS on the website trims and validates email address before sending. Good idea to also do both on the server and in mobile apps.

eishay:
@raymondng added.
@andrewconner yes, in `SocialFinalizeInfo`

kifi-backend/modules/common/app/com/keepit/social/providers/UsernamePasswordProvider.scala
```diff
@@ -27,12 +28,14 @@ class UsernamePasswordProvider(app: Application)
     UPP.loginForm.bindFromRequest().fold(
       errors => Left(error("bad_form")),
       credentials => {
-        UserService.find(IdentityId(credentials._1, id)) match {
+        val (emailString, password) = credentials
+        val identityId = IdentityId(EmailAddress.validate(emailString.trim).get.address, id)
```
raymondng, line 32: It might be better to do trimming inside `EmailAddress`?

2is10, line 32: Better to trim here, I think, although I don’t feel strongly. The decision to trim is related to the specific source of the email address (user-typed, in this case).

@eishay, `EmailAddress.validate(emailString.trim).get` will throw if the address is invalid. This will result in a 500, no? 400 with explanation would be better. Come to think of it, I’m not convinced we ought to be validating the email address on login. I was thinking of signup before. If our email address validation ever becomes more strict, this code could block legitimate logins. Trimming should always be okay though. Just something to consider.

raymondng, line 32: @2is10 If we do trim inside `EmailAddress` (and consistently use it instead of `String`), then we'll not need to worry it about sprinkling `trim` in our codebase prior to creating an `EmailAddress`. It seems to be a small/simple win here. WDYT?

eishay, line 32: @2is10 i think the email must be valid within our system, otherwise its granted that it will be a  failed attempt
I'll do the 400

PR #11126, opened 2014-09-25
Typeahead: always honor limit

raymondng:
Fixing a bug there Typeahead does not honor limit when there are not enough good/zero hits -- a silly bug that somehow got in there. Thanks @2is10 for reporting the issue.

Added rudimentary sanity check. Will add more.

@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/TypeaheadCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/TypeaheadCommander.scala
@@ -284,9 +284,7 @@ class TypeaheadCommander @Inject() (
       zHits ++= ordered.takeWhile { case (_, hit) => hit.score == 0 }
       (zHits.length < limit) tap { res => if (res) allHits ++= ordered }
     }) map { _ =>
-      if (zHits.length >= limit) zHits.take(limit) else {
-        allHits.sorted(hitOrd)
-      }
+      (if (zHits.length >= limit) zHits else allHits.sorted(hitOrd)).take(limit)
     }
   }
 
--- a/kifi-backend/modules/shoebox/test/com/keepit/controllers/website/TypeaheadControllerTest.scala
+++ b/kifi-backend/modules/shoebox/test/com/keepit/controllers/website/TypeaheadControllerTest.scala
@@ -48,7 +48,7 @@ class TypeaheadControllerTest extends Specification with ShoeboxTestInjector {
 
   "TypeaheadController" should {
 
-    "query & sort results" in {
+    "query & sort & limit results" in {
       withDb(modules: _*) { implicit injector =>
         val u1 = inject[Database].readWrite { implicit session =>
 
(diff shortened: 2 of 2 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/TypeaheadCommander.scala
```diff
@@ -284,9 +284,7 @@ class TypeaheadCommander @Inject() (
       zHits ++= ordered.takeWhile { case (_, hit) => hit.score == 0 }
       (zHits.length < limit) tap { res => if (res) allHits ++= ordered }
     }) map { _ =>
-      if (zHits.length >= limit) zHits.take(limit) else {
-        allHits.sorted(hitOrd)
-      }
+      (if (zHits.length >= limit) zHits else allHits.sorted(hitOrd)).take(limit)
```
2is10, line 287: what does `score == 0` signify?

2is10:
thanks for fixing

kifi-backend/modules/shoebox/app/com/keepit/commanders/TypeaheadCommander.scala
raymondng, line 287: that means it's a good/perfect hit -- i.e. the first token startsWith the prefix

raymondng:
Will merge and push to prod. Feedback appreciated!

kifi-backend/modules/shoebox/app/com/keepit/commanders/TypeaheadCommander.scala
2is10, line 287: I see. Might consider renaming `zHits` to `bestHits` or something for added clarity.

2is10, line 287: It looks like we’re using the `score == 0` rule inconsistently. Here’s a sequence I just saw as I typed `g` `a` `l`:
image
image
image
It feels weird that my Kifi friend Ben Galbraith is there, then gone, then back. Personally, I’d prefer to always see matching Kifi users ahead of email addresses.

raymondng, line 287: Actually this is deterministic and expected, but there are some "product rules" that are meant to help, but does look confusing in this case. 

1) When you typed "a" -- email addresses were not returned; that's due to a specific "product rule" to exclude email addresses when there is only 1 character in the query. So your friend's hit is on the list.

2) When you typed "ga" there were a lot of good email matches and as we sort strictly by scoring so the (score == 0) hits (emails) went to the top. Your friend's hit missed the cutoff as it's not a perfect match.

3) When you typed "gal" it eliminated a bunch of "ga" matches and then your friend's hit went to the top. 

If you typed "Ben" instead, then chances are your friend's hit will show more consistently. Does it make sense?

There are some product specific rules involved which has an undeniable impact on the code and results. There are always trade-offs involved and I'd be happy to talk to @JRuff42 about this together with you if you feel the current rules are confusing.

Also: will rename `zHits` for better clarity

2is10, line 287: I see. I guess it’s the product rule in (1) that introduces the inconsistency.

I don’t think email addresses should ever be ranked above matching kifi users, since there’s such a large prior probability that kifi users will be using kifi to message friends already on kifi. Happy to take it up with product.

raymondng, line 287: Sounds good to me. Thanks Jared.

PR #11128, opened 2014-09-25
fixing spacing for non-library experimenters & other things

ahsu1230:
@andrewconner @yipingliao 

and other things:
- library name wrap-around for first line
- then truncate with elipsis for second line
- styled up "filter" box for "Find a Library..." input

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
+++ b/kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
@@ -38,6 +38,18 @@ angular.module('kifi')
                 return lib.kind === 'user_created';
               });
               scope.invitedLibs = libraryService.invitedSummaries;
+
+              var lines;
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
```diff
@@ -69,7 +81,74 @@ angular.module('kifi')
...
+                secondLine = fullName.substr(0, remainingLen);
+              } else {
+                secondLine = fullName.substr(0, maxLength-3) + '...';
+              }
+            }
+          }
+          return [firstLine, secondLine];
+        }
+
+        // Filter Box Stuff
```
ahsu1230, line 113: A lot filtering stuff taken from tags.js

ahsu1230:
Happy to address feedback on future PR

PR #11132, opened 2014-09-26
require callers to be slightly more explicit when requesting INACTIVE rows

2is10:
@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepImageCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepImageCommander.scala
@@ -247,7 +247,7 @@ class KeepImageCommanderImpl @Inject() (
           ki
       }
       db.readWrite(attempts = 3) { implicit session => // because of request consolidator, this can be very race-conditiony
-        val existingImagesForKeep = keepImageRepo.getAllForKeepId(keepId)
+        val existingImagesForKeep = keepImageRepo.getForKeepId(keepId, excludeState = None)
         if (existingImagesForKeep.isEmpty) {
           keepImages.map(keepImageRepo.save)
         } else {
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepImageCommander.scala
```diff
@@ -247,7 +247,7 @@ class KeepImageCommanderImpl @Inject() (
           ki
       }
       db.readWrite(attempts = 3) { implicit session => // because of request consolidator, this can be very race-conditiony
-        val existingImagesForKeep = keepImageRepo.getAllForKeepId(keepId)
+        val existingImagesForKeep = keepImageRepo.getForKeepId(keepId, excludeState = None)
```
2is10, line 250: It’s not much longer, and I think it’s much clearer and less likely to be used accidentally. What do you think?

kifi-backend/modules/shoebox/app/com/keepit/model/KeepImageRepo.scala
```diff
@@ -49,18 +48,11 @@ class KeepImageRepoImpl @Inject() (
 
   override def deleteCache(model: KeepImage)(implicit session: RSession): Unit = {}
 
-  private val getForKeepIdCompiled = Compiled { keepId: Column[Id[Keep]] =>
-    for (r <- rows if r.keepId === keepId && r.state === KeepImageStates.ACTIVE) yield r
+  private def getForKeepIdCompiled(keepId: Column[Id[Keep]], excludeState: Option[State[KeepImage]]) = Compiled {
```
andrewconner: This is problematic, it will recompile the query on every call (defeating the purpose of pre-compiled queries). The original was written in that fixed manner (i.e., `getForKeepId` and `getAllForKeepId`) so we could have them compiled.

We could call `getAllForKeepId` `getAllForKeepIdIncludingInactive` if you want

andrewconner: Or, just drop trying to pre-compile it.

2is10: I’ve updated it with a `val` compiled query. All tests pass, but I’m not 100% confident it’s correct. What do you think?

2is10:
nevermind. not all tests pass. closing without merging.

PR #11136, opened 2014-09-26
get filtering box on left hand Library Pane to work!

ahsu1230:
can filter out libraries as you type in letters in the input box! woohoo!
Main & Secret libraries never show up under the search box too

@andrewconner @yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
+++ b/kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
@@ -18,8 +18,8 @@ angular.module('kifi')
         scope.librariesEnabled = false;
         scope.mainLib = {};
         scope.secretLib = {};
-        scope.userLibs = [];
-        scope.invitedLibs = [];
+        scope.userLibs = libraryService.userLibsToShow;
+        scope.invitedLibs = libraryService.invitedLibsToShow;
 
         scope.$watch(function () {
           return libraryService.isAllowed();
@@ -34,22 +34,6 @@ angular.module('kifi')
               scope.secretLib = _.find(libraries, function (lib) {
                   return lib.kind === 'system_secret';
               });
-              scope.userLibs = _.filter(libraries, function (lib) {
-                return lib.kind === 'user_created';
-              });
-              scope.invitedLibs = libraryService.invitedSummaries;
-
-              var lines;
-              for (var i = 0; i < scope.userLibs.length; i++) {
-                lines = shortenName(scope.userLibs[i].name);
-                scope.userLibs[i].firstLine = lines[0];
-                scope.userLibs[i].secondLine = lines[1];
-              }
-              for (i = 0; i < scope.invitedLibs.length; i++) {
-                lines = shortenName(scope.invitedLibs[i].name);
-                scope.invitedLibs[i].firstLine = lines[0];
-                scope.invitedLibs[i].secondLine = lines[1];
-              }
             });
           }
         });
@@ -80,35 +64,6 @@ angular.module('kifi')
         scope.inRecoExperiment = function () {
           return profileService.me && profileService.me.experiments && profileService.me.experiments.indexOf('recos_beta') >= 0;
         };
-
-        var maxLength = 22;
-        function shortenName (fullName) {
-          var firstLine = fullName;
-          var secondLine = '';
-          if (fullName.length > maxLength) {
-            var full = false;
-            var line = '';
-            while (!full) {
-              var pos = fullName.indexOf(' ');
-              if (line.length + pos < maxLength) {
-                line = line + fullName.substr(0, pos+1);
-                fullName = fullName.slice(pos+1);
-              } else {
-                full = true;
-              }
-            }
-            firstLine = line;
-            var remainingLen = fullName.length;
-            if (remainingLen > 0) {
-              if (remainingLen < maxLength) {
-                secondLine = fullName.substr(0, remainingLen);
-              } else {
-                secondLine = fullName.substr(0, maxLength-3) + '...';
-              }
-            }
-          }
-          return [firstLine, secondLine];
-        }
 
         // Filter Box Stuff
         scope.filter = {};
@@ -145,7 +100,7 @@ angular.module('kifi')
         }*/
 
         scope.onFilterChange = function () {
-          //libraryService.filterList(scope.filter.name);
+          libraryService.filterLibraries(scope.filter.name);
         };
 
         // Scroll-Bar Stuff
--- a/kifi-backend/modules/shoebox/angular/src/libraries/libraryService.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/libraryService.js
@@ -7,19 +7,38 @@ angular.module('kifi')
   function ($http, util, profileService, routeService, Clutch, $q) {
     var librarySummaries = [],
         invitedSummaries = [],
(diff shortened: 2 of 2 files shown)


kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
```diff
@@ -34,22 +34,6 @@ angular.module('kifi')
               scope.secretLib = _.find(libraries, function (lib) {
                   return lib.kind === 'system_secret';
               });
-              scope.userLibs = _.filter(libraries, function (lib) {
-                return lib.kind === 'user_created';
-              });
-              scope.invitedLibs = libraryService.invitedSummaries;
-
-              var lines;
```
ahsu1230, line 42: naming wrapping & stuff moved to libraryService so the firstLine / secondLine fields are added right when libraries are retrieved from servers. No need to do it again anywhere else.

andrewconner:
Merging for demo, will go through more soon.

ahsu1230:
thanks

kifi-backend/modules/shoebox/angular/src/libraries/libraryService.js
```diff
@@ -192,6 +243,23 @@ angular.module('kifi')
...
+
+      filterLibraries: function(term) {
+        var newMyLibs = librarySummaries;
+        var newMyInvited = invitedSummaries;
+        if (term.length) {
+          newMyLibs = fuseSearch1.search(term);
+          newMyInvited = fuseSearch2.search(term);
+        }
+        newMyLibs = _.filter(newMyLibs, function (lib) {
+          return lib.kind === 'user_created';
```
yipingliao, line 256: This filtering to get to the user created library happens in a couple of places. Maybe it'd be convenient to have a userCreatedLibrarySummaries that is created whenever we update librarySummaries, and the rest of the code can just refer to that.

kifi-backend/modules/shoebox/angular/src/libraries/libraryService.js
```diff
@@ -192,6 +243,23 @@ angular.module('kifi')
             return library.id === libraryId;
           });
         });
+      },
+
+      filterLibraries: function(term) {
```
yipingliao, line 248: Add space after `function`.

kifi-backend/modules/shoebox/angular/src/libraries/libraryService.js
```diff
@@ -7,19 +7,38 @@ angular.module('kifi')
...
 
     var librarySummariesService = new Clutch(function () {
       return $http.get(routeService.getLibrarySummaries).then(function (res) {
-        util.replaceArrayInPlace(librarySummaries, res.data.libraries || []);
-        util.replaceArrayInPlace(invitedSummaries, res.data.invited || []);
+        var libs = res.data.libraries || [];
+        var invites = res.data.invited || [];
+        var i, lines;
+        for (i = 0; i < libs.length; i++) {
+          lines = shortenLibName(libs[i].name);
```
yipingliao, line 27: This code can be DRYed up with the code in the next for-loop.

kifi-backend/modules/shoebox/angular/src/libraries/libraryService.js
```diff
@@ -7,19 +7,38 @@ angular.module('kifi')
...
+    var fuseSearch2 = new Fuse(invitedSummaries, fuseOptions);
 
     var librarySummariesService = new Clutch(function () {
       return $http.get(routeService.getLibrarySummaries).then(function (res) {
-        util.replaceArrayInPlace(librarySummaries, res.data.libraries || []);
-        util.replaceArrayInPlace(invitedSummaries, res.data.invited || []);
+        var libs = res.data.libraries || [];
+        var invites = res.data.invited || [];
+        var i, lines;
+        for (i = 0; i < libs.length; i++) {
```
yipingliao, line 26: Try `forEach` for better clarity.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach

kifi-backend/modules/shoebox/angular/src/libraries/libraryService.js
```diff
@@ -62,10 +81,42 @@ angular.module('kifi')
       return candidate;
     };
 
+    var maxLength = 22;
+    
+    function shortenLibName (fullName) {
```
yipingliao, line 86: Remove space after `shortenLibName`.

kifi-backend/modules/shoebox/angular/src/libraries/libraryService.js
```diff
@@ -7,19 +7,38 @@ angular.module('kifi')
...
     var librarySummaries = [],
         invitedSummaries = [],
+        userLibsToShow = [],
+        invitedLibsToShow = [],
         libraryState = {}; // This is a variable to pass state from different library components of the system
-    /*
+    
     var fuseOptions = {
        keys: ['name'],
        threshold: 0.3
```
yipingliao, line 16: Any explanation as to why the threshold is set to 0.3 and what that means?

yipingliao:
Just a bunch of formatting nits and cleanup suggestions. The infinite loop definitely escaped me when I glanced at this yesterday. :(

kifi-backend/modules/shoebox/angular/src/libraries/libraryService.js
andrewconner, line 16: Fuse allows for fuzzy search. Threshold is how far to go. Their docs are pretty sparse about meanings, but 0 means "exact match" and 1 means "any order of the letters"

kifi-backend/modules/shoebox/angular/src/libraries/libraryService.js
```diff
@@ -7,19 +7,38 @@ angular.module('kifi')
...
     };
-    var fuseSearch = new Fuse(libraries, fuseOptions);
-    */
+    var fuseSearch1 = new Fuse(librarySummaries, fuseOptions);
```
andrewconner, line 18: Better naming of this? `librarySummarySearch` and `invitedSummarySearch`?

kifi-backend/modules/shoebox/angular/src/libraries/libraryService.js
andrewconner, line 86: Is this the best way to do it? What about just using CSS?

2is10, line 86: +1 for CSS. Trailing `...` isn’t critical, in my opinion. Library names longer than two lines will probably be rare, and the users will likely already be familiar with them. If someone insists on a truncation indicator, we could use the second-line fade effect that we use in the extension.

andrewconner, line 86: Multi-line ellipsis http://codepen.io/martinwolf/pen/qlFdp

2is10, line 86: Cool, but that’s webkit-only, using an obsolete property.

The best CSS-only approach I’ve seen uses a fade and, optionally, an
ellipsis aligned to the box’s right edge. I find an ellipsis not aligned
with to the end of the text to be a bit gimmicky, but if it were really
faint, I could live with it.
http://www.cssmojo.com/line-clamp_for_non_webkit-based_browsers/

PR #11138, opened 2014-09-26
return 400 if the email address format is not valid

eishay:
@andrewconner @raymondng @2is10

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/social/providers/UsernamePasswordProvider.scala
+++ b/kifi-backend/modules/common/app/com/keepit/social/providers/UsernamePasswordProvider.scala
@@ -9,7 +9,7 @@ import com.keepit.social.{ UserIdentity, UserIdentityProvider }
 import play.api.Application
 import play.api.libs.json.Json
 import play.api.mvc.{ Request, Result }
-import play.api.mvc.Results.Forbidden
+import play.api.mvc.Results.{ Forbidden, BadRequest }
 
 import securesocial.core.providers.{ UsernamePasswordProvider => UPP }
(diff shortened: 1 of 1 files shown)


andrewconner:
Looks like `error` returns a `Forbidden`

raymondng:
Just checking -- how do we handle this on the client side? Any changes required?

eishay:
@andrewconner changed to BadRequest
@raymondng I don't think that they handle the different error checks now, though it will be good to do so

kifi-backend/modules/common/app/com/keepit/social/providers/UsernamePasswordProvider.scala
```diff
@@ -29,19 +31,25 @@ class UsernamePasswordProvider(app: Application)
...
-            identity match {
-              case userIdentity: UserIdentity =>
-                log.info(s"[doAuth] userIdentity=$userIdentity")
-                if (passwordAuth.authenticate(userIdentity.userId.get, password)) Right(userIdentity) else Left(error("wrong_password"))
-              case _ =>
-                log.error(s"[doAuth] identity passed in is not of type <UserIdentity>; class=${identity.getClass}; obj=$identity")
-                Left(error("wrong_password")) // wrong_password for compatibility; auth_failure/internal_error more accurate
+        EmailAddress.validate(emailString.trim) match {
+          case Failure(e) =>
+            log.error(s"bad email format $emailString used for login", e)
```
2is10: A stack trace may not be particularly helpful in the log in this case.

2is10:
:+1:

PR #11144, opened 2014-09-26
returning our new keep image URL after changing it

2is10:
Also validating early that we can parse keep image URIs.
@andrewconner


andrewconner:
Works for me. Your requested tests are still on their way :)

PR #11150, opened 2014-09-26
fixing infinite loop

2is10:
@ahsu1230 cc @andrewconner


kifi-backend/modules/shoebox/angular/src/libraries/libraryService.js
```diff
@@ -91,7 +91,7 @@ angular.module('kifi')
         var line = '';
         while (!full) {
           var pos = fullName.indexOf(' ');
-          if (line.length + pos < maxLength) {
+          if (pos >= 0 && line.length + pos <= maxLength) {
```
2is10, line 94: changed to `<=` because a trailing space isn’t rendered and so shouldn’t count

the fix was checking to make sure `pos` wasn’t `-1`, which resulted in appending an empty string to `line` and stripping nothing from `fullName`, hence the infinite loop

PR #11152, opened 2014-09-26
infrastructure for more detailed analytics from email clicks

joshm1:
@stkem 

the tracking isn't used yet, but if you have any feedback for improvement I'd love to hear it before I start using it to generate the query param in emails


kifi-backend/modules/common/app/com/keepit/common/mail/template/EmailTrackingParam.scala
```diff
@@ -0,0 +1,47 @@
...
+  subAction: Option[String] = None,
+  variableComponents: Seq[String] = Seq.empty,
+  tips: Seq[EmailTip] = Seq.empty,
+  auxiliaryData: Option[HeimdalContext] = None)
+
+object EmailTrackingParam extends Logging {
+  val paramName = "dat"
+
+  /*
+    ***** README *****
```
stkem, line 20: :+1:

kifi-backend/modules/shoebox/app/com/keepit/commanders/SendgridCommander.scala
```diff
@@ -1,14 +1,15 @@
...
-import com.keepit.common.healthcheck.SystemAdminMailSender
+import com.keepit.common.healthcheck.{ AirbrakeNotifier, SystemAdminMailSender }
 import com.keepit.common.logging.Logging
+import com.keepit.common.mail.template.EmailTrackingParam
+import com.keepit.common.mail.template.EmailTip.toContextData
 import com.keepit.common.mail.{ ElectronicMail, ElectronicMailRepo, EmailAddress, SystemEmailAddress }
 import com.keepit.common.time.{ DEFAULT_DATE_TIME_ZONE, currentDateTime }
-import com.keepit.heimdal.{ HeimdalContextBuilderFactory, HeimdalServiceClient, NonUserEvent, NonUserEventTypes, UserEvent, UserEventTypes }
-import com.keepit.model.{ NormalizedURI, UriRecommendationFeedback, EmailOptOutRepo, NotificationCategory, UserEmailAddressRepo, UserEmailAddressStates }
+import com.keepit.heimdal._
```
stkem: I don't like `._` imports. There is a lot being imported here, so it's kind of borderline. Just registering my mild displeasure :-)

stkem:
lgtm

kifi-backend/modules/shoebox/app/com/keepit/commanders/SendgridCommander.scala
joshm1: I agree with you.  My IntelliJ is setup to automatically the wildcard when there are 7+ from the same package.  I've updated it to not do this anymore.

PR #11153, opened 2014-09-26
fix impersonation

raymondng:
This should do it. Thanks @andrewconner for catching the issue.


raymondng:
Will merge + push + test. Feedback appreciated!

PR #11161, opened 2014-09-27
Remove empty collections and clean up related code

leogrim:
@andrewconner @2is10 
I guess we can also remove all the code related to manual deletion of collections once the UI is gone.
Feel free to merge it if I don't :)


2is10:
sorry, reverting to fix build. looks like there are some assumptions about how `collectionChanged` is supposed to work that no longer hold. it might be better to do this check for whether the collection is empty in the several places where we’re specifically unkeeping instead of just trying to do it once in `collectionChanged`.

PR #11165, opened 2014-09-28
Modal refactor + create/manage library modals

yipingliao:
DO NOT MERGE - I plan to do a bit of git commit squashing. 

@andrewconner cc @2is10 (1) Refactor modals to use a modal service and (2) implement changes to finish functionality and styling for create and manage library modals.

I apologize for yet another huge PR; I know they are difficult to review. Unfortunately, in this case, once I started to refactor the modals I had to refactor them all or some of them will be broken. :(

Most of the changes are for the modal refactor. I will add comments on the files that are actually related to the create/manage library modals.

The new modal usage uses `modalService`, which gets a HTML template from `$templateCache`, and compiles it and adds it to the DOM. Previously, all of the modals were embedded in the DOM (they were hidden).

To open a modal, `modalService.open` is called with a set of options. The options must contain a path to the template file of the modal. Optionally, data can be passed into the modal. Also optionally, a scope can be passed into the modal so that the modal is compiled with a passed-in scope and code in the modal can access the calling scope.

The modals in our code base use the modals in many different ways. Some are completely static, some use the `kfBasicModalContent` directive which already comes with cancel and submit buttons, some are so complicated that they contain their own directives, and sometimes there is a sequence of modals for a particular action. I'd like document this in the wiki after (and if!) this PR is approved.

Thanks!
Yiping


kifi-backend/modules/shoebox/angular/src/common/build-css/common.styl
```diff
@@ -1,3 +1,6 @@
+//
```
yipingliao, line 1: I found that the global stylus rules can go in here!  So I moved some recently created global rules in here.

kifi-backend/modules/shoebox/angular/src/common/build-css/constants.styl
```diff
@@ -1,25 +1,41 @@
+//
```
yipingliao, line 1: Also found this handy file for constants. Alpha-sorted the colors so that it's easy to tell if we're about to make a color with a duplicate name.

kifi-backend/modules/shoebox/angular/src/common/services/routeService.js
```diff
@@ -127,6 +127,9 @@ angular.module('kifi')
       },
       leaveLibrary: function (libraryId) {
         return route('/libraries/' + libraryId + '/leave');
+      },
+      deleteLibrary: function (libraryId) {
```
yipingliao, line 131: Unrelated to modals: for deleting a library.

kifi-backend/modules/shoebox/angular/src/friends/savePymkService.js
```diff
@@ -1,17 +0,0 @@
-'use strict';
```
yipingliao, line 1: The original impetus for the modal refactor. :) Removed this service that is a global variable in disguise.

kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
```diff
@@ -38,8 +38,18 @@ angular.module('kifi')
           }
         });
 
+        $rootScope.$on('changedLibrary', function () {
```
yipingliao, line 41: Update the libraries listed on the left side when a library has changed.

kifi-backend/modules/shoebox/angular/src/libraries/library.js
```diff
@@ -50,18 +50,23 @@ angular.module('kifi')
...
-        library: $scope.library,
-        returnAction: function () {
-          libraryService.getLibraryById($scope.library.id, true).then(function (data) {
-            libraryService.getLibraryByUserSlug(username, data.library.slug, true);
-            if (data.library.slug !== librarySlug) {
-              $location.path('/' + username + '/' + data.library.slug);
-            }
-          });
+      modalService.open({
+        template: 'libraries/manageLibraryModal.tpl.html',
```
yipingliao, line 54: `libraryState` was also a "global" variable in `libraryService` that held data to be passed into the modal, but the data can now be passed directly into the modal.

kifi-backend/modules/shoebox/angular/src/libraries/libraryShareSearch.js
```diff
@@ -2,13 +2,14 @@
 
 angular.module('kifi')
 
-.directive('kfLibraryShareSearch', ['$document', 'friendService', 'keyIndices', 'libraryService', 'socialService', 'util',
-  function ($document, friendService, keyIndices, libraryService, socialService, util) {
+.directive('kfLibraryShareSearch', ['$document', '$timeout', 'friendService', 'keyIndices', 'libraryService', 'socialService', 'util',
```
yipingliao, line 5: This is the directive for the library-share dropdown menu where you can find people to invite to view libraries. These changes are made to accommodate using this dropdown within the manage-library modal, which requires a couple of different behaviors.

kifi-backend/modules/shoebox/angular/src/libraries/libraryShareSearch.tpl.html
```diff
@@ -1,12 +1,12 @@
 <div class="kf-library-share-search">
-  <button class="library-share-button" ng-click="toggleMenu()">Share</button>
+  <button ng-if="!manageLibInvite" class="library-share-button" ng-click="toggleMenu()">Share</button>
```
yipingliao, line 2: Some headers in the library share dropdown are not shown when the dropdown is inside the manage-library modal.

kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.js
```diff
@@ -2,100 +2,175 @@
...
-        'description': '',
-        'slug': ''
-      };
-      $scope.modalTitle = 'Create a library';
-    }
-
-    var generateSlug = function (name) {
-      return name.toLowerCase().replace(/[^\w\s]|_/g, '').replace(/\s+/g, '-').replace(/^-/, '');
-    };
+.directive('kfManageLibrary', ['$location', '$rootScope', 'friendService', 'libraryService', 'profileService',
```
yipingliao, line 5: Converted the manage-library controller for the manage-library modal into a directive for better control over the modal, but the "meat" of the code is the same.

kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.styl
```diff
@@ -0,0 +1,320 @@
...
+
+.manage-lib-follower-status
+  position: absolute
+  top: 22px
+  right: 25px
+  color: green
+
+  .manage-lib-owner &
+    color: kifiBlack
+
```
yipingliao, line 317: TODO(yiping): remove extra newlines.

kifi-backend/modules/shoebox/angular/src/libraries/manageLibraryModal.styl
```diff
@@ -1,52 +0,0 @@
-.manage-library
```
yipingliao, line 1: Migrated to `manageLibrary.styl`.

kifi-backend/modules/shoebox/angular/src/libraries/manageLibraryModal.tpl.html
```diff
@@ -1,43 +1,3 @@
-<div kf-modal show="data.showManageLibraryModal" kf-width="600px">
```
yipingliao, line 1: The meat of this template file has been migrated to `manageLibrary.tpl.html`.

kifi-backend/modules/shoebox/angular/src/common/build-css/constants.styl
andrewconner, line 1: Brilliant!

kifi-backend/modules/shoebox/angular/src/common/services/modalService.js
```diff
@@ -0,0 +1,38 @@
...
+      var $modal = angular.element($templateCache.get(template));
+
+      var scope = opts.scope || $rootScope.$new();
+      if (opts.modalData) {
+        scope.modalData = opts.modalData;
+      }
+
+      $timeout(function () {
+        $compile($modal)(scope);
+        angular.element(document.body).append($modal);
```
andrewconner: (Assuming it does since you put it there, but...) does this need to be in a `$timeout`?

kifi-backend/modules/shoebox/angular/src/common/services/modalService.js
```diff
@@ -0,0 +1,38 @@
...
+      });
+    }
+
+    function close () {
```
andrewconner, line 25: It may make sense to add a reference to `close()` to the modal's scope (unless there's another mechanism I haven't gotten to yet)?

andrewconner, line 25: On second thought, this may not be a good idea since it'll pollute the parent scope (since the scope may be the parent scope). Perhaps we should create a new child scope?

yipingliao, line 25: The `close()` function is actually called from `modals.js`, by the `kfModal` directive. The idea is that something in the modal template will call this function when it's closing. In our case all our modals use the `kfModal` directive, which wraps this `close()` function within its own `close` function that does some other stuff.

I am not completely happy with the API provided by `modalService`; it's asymmetric and unintuitive. I can't come up with anything better at the moment, but suggestions are very much welcomed.

kifi-backend/modules/shoebox/angular/src/common/services/modalService.js
yipingliao: I'll double-check to make sure.

2is10: if so, it’d be good to leave a tiny comment out to the side mentioning what goes wrong without it

kifi-backend/modules/shoebox/angular/src/friends/seeMutualFriends.js
```diff
@@ -74,19 +65,6 @@ angular.module('kifi')
...
-        // to respond to the 'showGlobalModal' event and display the
-        // modal (we cannot scroll hidden elements).
-        var mutualFriendsContainer = element.find('.kf-mutual-friends-friends');
-        $rootScope.$on('showGlobalModal', function (e, modal) {
-          if (modal === 'seeMutualFriends') {
-            $timeout(function () {
-              mutualFriendsContainer.scrollTop(0);
-            }, 100);
-          }
-        });
```
2is10, line 89: yay! so long, ugly code ;)

kifi-backend/modules/shoebox/angular/src/common/services/modalService.js
andrewconner, line 25: I haven't fully parsed how this is used, but what about creating a new child scope always, based on what's passed in? So something like:

``` javascript
var scope = (opts.scope && opts.scope.$new()) || $rootScope.$new();
```

kifi-backend/modules/shoebox/angular/src/layout/main/MainCtrl.js
```diff
@@ -93,100 +93,73 @@ angular.module('kifi')
...
-          break;
       }
     });
 
     $scope.importBookmarks = function (makePublic) {
-      $scope.data.showImportModal = false;
+      $scope.forceClose = true;
 
-      var kifiVersion = $window.document.getElementsByTagName('html')[0].getAttribute('data-kifi-ext');
+      $timeout(function () {
```
2is10: small comment with reason for `$timeout`?

kifi-backend/modules/shoebox/angular/src/layout/main/MainCtrl.js
```diff
@@ -261,16 +234,21 @@ angular.module('kifi')
...
-            } else { // hrmph.
-              $scope.modal = 'import_bookmarks_error';
-              $scope.data.showBookmarkFileModal1 = false;
-              $scope.data.showBookmarkFileError = true;
-              $scope.modal = 'import_bookmark_error';
-            }
+
+            $scope.forceClose = true;
+
+            $timeout(function () {
```
2is10: another `$timeout`

kifi-backend/modules/shoebox/angular/src/libraries/libraryService.js
```diff
@@ -260,6 +258,14 @@ angular.module('kifi')
         util.replaceArrayInPlace(invitedLibsToShow, newMyInvited);
 
         return userLibsToShow.concat(invitedLibsToShow);
+      },
+
+      deleteLibrary: function (libraryId) {
+        return $http.post(routeService.deleteLibrary(libraryId)).then( function () {
+          _.remove(librarySummaries, function (library) {
+            return library.id === libraryId;
+          });
```
2is10, line 267: looks like lodash [supports](http://lodash.com/docs#remove) a declarative shorthand:

``` js
_.remove(librarySummaries, {id: libraryId});
```

kifi-backend/modules/shoebox/angular/src/libraries/libraryShareSearch.js
```diff
@@ -43,8 +44,16 @@ angular.module('kifi')
...
-          populateDropDown();
+
+          if (!scope.manageLibInvite) {
+            // When we test this conditional, Angular thinks that we're trying to
+            // enter an already-in-progress digest loop. To get around this, use
+            // $timeout to schedule the following code in a future call stack.
+            $timeout(function () {
+              searchInput.focus();
+              populateDropDown();
+            }, 0);
```
2is10, line 55: Is there a way to pass a “fixed” (constant, non-live) parameter to an Angular template? It seems this value doesn’t actually need to be watched, and if it weren’t, this problem might be avoided.

andrewconner, line 55: Easiest answer right now: https://github.com/Pasvaz/bindonce

In 1.3 (coming out soon!), there's built in syntax for binding once (no watches after the data is loaded).

kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.js
```diff
@@ -2,100 +2,175 @@
...
+        // Scope data.
+        //
+        scope.username = profileService.me.username;
+        scope.userHasEditedSlug = false;
+        scope.$error = {};
 
+        scope.visibility = {
+          'published': {
+            'title': 'Published Library',
+            'content': 'This library is available for everyone to see. It also generates a synamic public page to share with non-Kifi users.'
```
2is10: “synamic” → “dynamic”

kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.js
```diff
@@ -2,100 +2,175 @@
...
+
+        //
+        // Scope methods.
+        //
+        scope.close = function () {
+          kfModalCtrl.close();
+        };
+
+        scope.editSlug = function () {
+          scope.userHasEditedSlug = scope.library.slug? true : false;
```
2is10: space before `?` or maybe just use `!!` instead

kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.js
```diff
@@ -2,100 +2,175 @@
...
+        };
+
+
+        //
+        // Watches and listeners.
+        //
+        scope.$watch(function () {
+          return scope.library.name;
+        }, function (newVal, oldVal) {
+          scope.userHasEditedSlug = scope.library.name ? true : false;
```
2is10: Could this line go instead the `if` block below? Code might be clearer if this came right after an assignment to `.slug`.

kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.tpl.html
```diff
@@ -0,0 +1,78 @@
...
+              <div class="manage-lib-choice-card">
+                <div class="manage-lib-choice-card-img svg-red-card-with-lock"></div>
+                <input type="radio" id="manage-lib-radio-secret" class="manage-lib-choice-input" ng-model="library.visibility" value="secret">
+                <label for="manage-lib-radio-secret" class="manage-lib-custom-radio"><span class="manage-lib-custom-radio-check"></span>L</label>
+                <span class="manage-lib-choice-card-text">Secret</span>
+              </div>
+              <div class="manage-lib-choice-card">
+                <div class="manage-lib-choice-card-img svg-yellow-card-with-globe"></div>
+                <input type="radio" id="manage-lib-radio-discoverable" class="manage-lib-choice-input" ng-model="library.visibility" value="discoverable">
+                <label for="manage-lib-radio-discoverable" class="manage-lib-custom-radio"><span class="manage-lib-custom-radio-check"></span>L</label>
```
2is10: FYI: You can avoid using `for` and `id` if you enclose a control element in its `<label>`:

``` html
<label class="...">
  <input type="radio" class="..." value="..."> label text here
</label>

```

kifi-backend/modules/shoebox/angular/src/profile/profileImage.js
```diff
@@ -222,39 +222,51 @@ angular.module('kifi')
...
+            if (upload) {
+              upload.promise.then(function (result) {
+                var scaling = positioning.imageWidth / positioning.currentWidth;
+                var data = {
+                  picToken: result && result.token,
+                  picWidth: positioning.imageWidth,
+                  picHeight: positioning.imageHeight,
+                  cropX: Math.floor(scaling * (maskOffset - positioning.currentLeft)),
+                  cropY: Math.floor(scaling * (maskOffset - positioning.currentTop)),
+                  cropSize: Math.floor(Math.min(scaling * maskSize, scaling * maskSize))
```
2is10: `min` of `x` and `x` is usually `x`

kifi-backend/modules/shoebox/angular/src/profile/profileInput.js
```diff
@@ -44,19 +44,25 @@ angular.module('kifi')
 
         function onClickOutsideInput(event) {
           // When user clicks outside the input and save button, cancel edits.
-          if (!event.target.classList.contains('profile-email-input') &&
-            !event.target.classList.contains('profile-input-save')) {
+          if (!event.target.classList ||
+              (!event.target.classList.contains('profile-email-input') &&
+               !event.target.classList.contains('profile-input-save'))) {
```
2is10: Looks like IE 9 doesn’t support `.classList`. Maybe:

``` js
if (!$(event.target).is('.profile-email-input,.profile-input-save')) {
```

kifi-backend/modules/shoebox/angular/src/profile/profileNameInput.js
```diff
@@ -46,19 +46,25 @@ angular.module('kifi')
...
-          if (!event.target.classList.contains('profile-name-input') &&
-            !event.target.classList.contains('profile-input-save')) {
+          if (!event.target.classList ||
+              (!event.target.classList.contains('profile-name-input') &&
+               !event.target.classList.contains('profile-input-save'))) {
             $document.off('mousedown', onClickOutsideInput);
-            scope.$apply(cancel);
+            $timeout(function () {
+              cancel();
+            }, 0);
```
2is10: `cancel` can be passed by name to `$timeout`:

``` js
$timeout(cancel, 0);
```

2is10:
Huge improvement! Very nice! Thanks for sticking with this ball of yarn change!
:+1: :+10:

kifi-backend/modules/shoebox/angular/src/common/services/modalService.js
yipingliao: Actually this particular $timeout is not needed. :) Removed. I also went back and checked all the timeouts: some I removed after restructuring the `close` function in `kfModal` a little bit (it now closes the current modal before executing the passed in function, instead of executing the passed in function before closing), some I converted to `$scope.$evalAsync`, which executes earlier than `$timeout` but achieves what we need in those situations. Remaining timeouts are needed and have been commented.

kifi-backend/modules/shoebox/angular/src/libraries/libraryShareSearch.js
yipingliao, line 55: Nice! I'd rather wait for 1.3 for this instead of adding another library to our code right now. Please let me know if you feel otherwise.

kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.js
yipingliao: Thanks! Done.

kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.js
yipingliao: Thanks! Done.

kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.js
yipingliao: Yup. Done.

kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.tpl.html
yipingliao: Thanks! Done.

kifi-backend/modules/shoebox/angular/src/profile/profileImage.js
yipingliao: Lol, fixed.

kifi-backend/modules/shoebox/angular/src/profile/profileNameInput.js
yipingliao: Done. Thanks!

kifi-backend/modules/shoebox/angular/src/profile/profileInput.js
yipingliao: Done. Thanks!

PR #11171, opened 2014-09-29
simplifying replaceArrayInPlace

2is10:
@yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/common/util.js
+++ b/kifi-backend/modules/shoebox/angular/src/common/util.js
@@ -24,11 +24,8 @@ angular.module('util', [])
       },
       replaceArrayInPlace: function (oldArray, newArray) {
         // empties oldArray, loads newArray values into it, keeping the same reference.
-        oldArray = oldArray || [];
         oldArray.length = 0;
-        // returning the array in case it was undefined before
-        Array.prototype.push.apply(oldArray, newArray);
-        return oldArray;
+        oldArray.push.apply(oldArray, newArray);
       },
       completeObjectInPlace: function (oldObj, newObj) {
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/angular/src/common/util.js
```diff
@@ -24,11 +24,8 @@ angular.module('util', [])
       },
       replaceArrayInPlace: function (oldArray, newArray) {
         // empties oldArray, loads newArray values into it, keeping the same reference.
-        oldArray = oldArray || [];
         oldArray.length = 0;
-        // returning the array in case it was undefined before
-        Array.prototype.push.apply(oldArray, newArray);
-        return oldArray;
+        oldArray.push.apply(oldArray, newArray);
```
2is10, line 28: I think it’s better to expect sane inputs (that correspond with the method name) than to try to recover somehow if they aren’t sane.

kifi-backend/modules/shoebox/angular/src/common/util.js
```diff
@@ -98,14 +95,17 @@ angular.module('util', [])
...
         }, {});
 
-        var that = this;
         _.forEach(keeps, function (keep) {
           var newTagList = _.map(_.union(keep.collections, keep.tags), function (tagId) {
             return idMap[tagId] || null;
           }).filter(function (tag) {
             return tag != null;
           });
-          keep.tagList = that.replaceArrayInPlace(keep.tagList, newTagList);
```
2is10, line 108: This was the only place that the return value from `.replaceArrayInPlace` was used.

yipingliao:
Looks great. Thanks!

PR #11177, opened 2014-09-29
moving joinTags out of util

2is10:
@yipingliao 
(ran lint this time)

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/common/util.js
+++ b/kifi-backend/modules/shoebox/angular/src/common/util.js
@@ -29,10 +29,10 @@ angular.module('util', [])
       },
       replaceObjectInPlace: function (oldObj, newObj) {
         // empties oldObj, loads newObj key/values into it, keeping the same reference.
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/angular/src/common/util.js
```diff
@@ -29,10 +29,10 @@ angular.module('util', [])
       },
       replaceObjectInPlace: function (oldObj, newObj) {
         // empties oldObj, loads newObj key/values into it, keeping the same reference.
-        _.forOwn(oldObj || {}, function (num, key) {
+        _.forOwn(oldObj, function (num, key) {
           delete oldObj[key];
         });
-        _.forOwn(newObj || {}, function (num, key) {
+        _.forOwn(newObj, function (num, key) {
```
2is10, line 35: `_.forOwn` just returns if first argument is `null` or `undefined`

yipingliao:
Thanks for doing all this cleanup!

PR #11178, opened 2014-09-29
Modal refactor squashed

yipingliao:
Squashed version of the modal refactoring pull you reviewed this morning. Also had to change some karma tests. :)

@andrewconner @2is10


kifi-backend/modules/shoebox/angular/test/unit/friends/seeMutualFriends.spec.js
```diff
@@ -99,18 +76,8 @@ describe('kifi.friends.seeMutualFriends', function () {
 
     });
 
-    it('should update modal content when saved person changes', function () {
```
yipingliao, line 102: This test is no longer needed because the modal is now opened with a new person, which will not change while the modal is open.

2is10:
:+1:

PR #11179, opened 2014-09-29
Make empty collection behavior consistent with other collections when ke...

leogrim:
...eping / unkeeping

@2is10 @andrewconner 
I'm not removing an empty collection after the "removeFromCollection" action since it only applies to the old UI (which allows manual deletion of tags).

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/model/CollectionRepo.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/model/CollectionRepo.scala
@@ -26,7 +26,7 @@ trait CollectionRepo extends Repo[Collection] with ExternalIdColumnFunction[Coll
   def count(userId: Id[User])(implicit session: RSession): Int
   def getBookmarkCounts(collIds: Set[Id[Collection]])(implicit session: RSession): Map[Id[Collection], Int]
   def getCollectionsChanged(num: SequenceNumber[Collection], limit: Int)(implicit session: RSession): Seq[Collection]
-  def collectionChanged(collectionId: Id[Collection], isNewKeep: Boolean = false)(implicit session: RWSession): Collection
+  def collectionChanged(collectionId: Id[Collection], isNewKeep: Boolean = false, inactivateIfEmpty: Boolean = false)(implicit session: RWSession): Collection
   def getTagsByKeepId(keepId: Id[Keep])(implicit session: RSession): Set[Hashtag]
 }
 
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/app/com/keepit/model/CollectionRepo.scala
```diff
@@ -132,14 +132,11 @@ class CollectionRepoImpl @Inject() (
...
     val collection = get(collectionId)
-    if (isNewKeep) {
-      save(collection withLastKeptTo clock.now())
-    } else {
-      val keepCount = getBookmarkCount(collectionId)
-      if (keepCount > 0) save(collection) else save(collection.copy(state = CollectionStates.INACTIVE))
-    }
+    if (isNewKeep) { save(collection withLastKeptTo clock.now()) }
+    else if (inactivateIfEmpty && getBookmarkCount(collectionId) == 0) { save(collection.copy(state = CollectionStates.INACTIVE)) }
+    else { save(collection) }
```
2is10, line 139: if/else-if/else blocks on the same line are hard to read

kifi-backend/modules/shoebox/app/com/keepit/model/CollectionRepo.scala
```diff
@@ -132,14 +132,11 @@ class CollectionRepoImpl @Inject() (
...
+  def collectionChanged(collectionId: Id[Collection], isNewKeep: Boolean = false, inactivateIfEmpty: Boolean = false)(implicit session: RWSession): Collection = {
     val collection = get(collectionId)
-    if (isNewKeep) {
-      save(collection withLastKeptTo clock.now())
-    } else {
-      val keepCount = getBookmarkCount(collectionId)
-      if (keepCount > 0) save(collection) else save(collection.copy(state = CollectionStates.INACTIVE))
-    }
+    if (isNewKeep) { save(collection withLastKeptTo clock.now()) }
+    else if (inactivateIfEmpty && getBookmarkCount(collectionId) == 0) { save(collection.copy(state = CollectionStates.INACTIVE)) }
```
2is10, line 138: might consider optimizing `getBookmarkCount` to `hasAnyBookmarks` (so the database can return after finding the first record instead of all records)

2is10:
:+1:

kifi-backend/modules/shoebox/app/com/keepit/model/CollectionRepo.scala
leogrim, line 138: That's true, though we're likely going to move away from this anyway or revisit it properly when we get everyone on the new UI, so I don't want to spend much more time on this.

2is10, line 138: agreed​

PR #11182, opened 2014-09-29
more robust detection of backup machine

yingjieMiao:
@ymatsuda @leogrim 

checking status inside `indexModule` is not right. (the status there could be `startingup`)


leogrim:
:+1:

ymatsuda:
:+1:

PR #11183, opened 2014-09-29
Have current SharingUserInfo rely on AugmentationCommander

leogrim:
@ymatsuda will try this in a bit, if all is well, I'll start cleaning things up and move the router into search.


ymatsuda:
nice!

PR #11187, opened 2014-09-29
Library card and library page header

yipingliao:
Basic library card and library page header. Same as the earlier pull that @2is10 has reviewed, but this one has pulled in the changes from modal refactor that have since been merged to master, and has cleaned up per review comments.


2is10:
:+1:

PR #11191, opened 2014-09-29
tweak in recent interest computation

yingjieMiao:
@stkem 

the 'normalization' process is a bit vulnerable to noisy classification.


stkem:
that looks like scaling is now always off, is that intended?

yingjieMiao:
yes, intended. We only need the recent interest score from the cosine distance. This change will change the overall_cosine_interest score as well, however, that is discarded anyway.

stkem:
ok, lgtm

PR #11197, opened 2014-09-30
Implement follower pics on library header.

yipingliao:
@andrewconner @2is10 cc @ahsu1230 

Display follower pics on the library card. See design at:
http://content.screencast.com/users/Dannyblumenfeld/folders/Snagit/media/c6fe325a-8eb8-4187-8f72-2c63e6ac4835/09.17.2014-10.png

Thanks!

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/library.styl
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/library.styl
@@ -48,12 +48,15 @@
   .kf-keep-lib-owner-name
     color: #939393
 
+  .kf-keep-lib-description-container
+    padding-bottom: 5px
+
   .kf-keep-lib-description
     margin: 10px 0
 
-  .kf-keep-lib-stats
-    padding-top: 10px
+  .kf-keep-lib-stats-and-followers
     border-top: 1px solid #dadae2
+    padding-top: 15px
 
   .kf-keep-lib-keeps-followers-stats
     display: inline-block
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/library.styl
```diff
@@ -101,6 +126,9 @@
   font-size: 13px
   color: #666
 
+  a
```
yipingliao, line 129: For social follow (not fully implemented yet).

kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
```diff
@@ -48,6 +80,7 @@ angular.module('kifi')
             scope.clippedDescription = true;
           }
 
+          scope.library.shareUrl = 'http://www.kifi.com' + scope.library.url;
```
yipingliao: For social follow (not fully implemented yet).

2is10: `https:`, but better not to hardcode the URL. Alternatives I know about:
- `location.origin + ...` (in modern browsers only)
- `location.protocol + '//' + location.host + ...`
- `$location.protocol() + '//' + $location.host() + $location.port() + ...` (omit port if 80 or 443)
- `a.href` (put the path in an anchor element’s `href` attribute and then read its `.href` property)

2is10:
nice

kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
andrewconner: Or `env.origin`

yipingliao: What's `env.origin`? (what's `env`? It's not a property on `window`).

kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
```diff
@@ -48,6 +80,7 @@ angular.module('kifi')
             scope.clippedDescription = true;
           }
 
+          scope.library.shareUrl = $location.protocol() + '://' + $location.host() + $location.path();
```
2is10, line 83: The path should be `scope.library.url`, yes?
Also, I believe Angular’s `$location.host()` omits a custom port, behaving like `location.hostname` rather than `location.host`, which will result in a broken URL in dev where we use ports 8080 and 9000.
A quick code search shows `env` is something you can inject, a factory we define in app.js. +1 for `env.origin`.

yipingliao, line 83: Thanks! Will fix in next pull.

PR #11198, opened 2014-09-30
reco computation for a few more users

stkem:
@yingjieMiao


yingjieMiao:
minor comments on the rate: the system will have much higher load when we approach to the long tail

stkem:
yeah, we'll have to lower the number in smaller and smaller steps (going from 500 -> 400 added about a hundred users)

PR #11199, opened 2014-09-30
tag/untag endpoints for extension

2is10:
@ahsu1230 @andrewconner 
Most of the logic can probably be refactored in a commander for reuse later.


andrewconner:
Great, thanks.

PR #11203, opened 2014-09-30
preventing library collaborators from tagging for the time being

2is10:
The previous code could have potentially let collaborators create their own personal tags on someone else’s keeps in someone else’s library. I’m disabling that to keep our current schema consistent until we properly support library-specific tags.

@andrewconner 
cc @leogrim


andrewconner:
Looks good.

PR #11214, opened 2014-09-30
More work on library card (header)

yipingliao:
Hi! Here are some more UI additions to the library card/header. Just a bunch of small changes; the commits list them.

@andrewconner @2is10 @ahsu1230 

Thanks!
Yiping


andrewconner:
Great :)

kifi-backend/modules/shoebox/angular/src/app.js
```diff
@@ -51,13 +51,16 @@ angular.module('kifi', [
     var host = $location.host(),
       dev = /^dev\.ezkeep\.com|localhost$/.test(host),
       local = $location.port() === 9000,
-      origin = local ? $location.protocol() + '://' + host  + ':' + $location.port() : 'https://www.kifi.com';
+      origin = local ? $location.protocol() + '://' + host  + ':' + $location.port() : 'https://www.kifi.com',
+      localWeb = $location.port() === 8080,
```
yipingliao: I added `localWeb` and `originWeb` to indicate if the app is running its Angular client locally (since `local` and `origin` indicate whether the server is running locally). I need this to get the Facebook URL sharing to work (you can only share a link from the same web site). This may go away when I work on the FB share dialog and use the full FB API.

ahsu1230:
:+1:

kifi-backend/modules/shoebox/angular/src/app.js
andrewconner: Mind checking for `$location.port() === 9000` as well here, since that uses `dev.ezkeep.com` as well?

2is10: We have a test facebook app for dev.ezkeep.com:9000 and a local play server can handle the facebook return path. I think Yiping is correctly only having `originWeb` override `origin` when there’s no play server.

kifi-backend/modules/shoebox/angular/src/app.js
```diff
@@ -51,13 +51,16 @@ angular.module('kifi', [
...
+      originWeb = localWeb ? $location.protocol() + '://' + host  + ':' + $location.port() : 'https://www.kifi.com';
```
2is10, line 56: I think we can eliminate `localWeb` and shorten these lines like this:

``` js
  origin = $location.protocol() + '://' + host  + (dev ? ':' + $location.port() : ''),
  originWeb = dev && !local ? 'https://www.kifi.com' : origin;
```

yipingliao, line 56: Done. Thanks for the shortened code!

yipingliao:
Closing this pull since we got into an unexpected conversation about `env.origin`. I will make two new pulls: one for the UI part, one for the `origin`part, so that we don't have a big mixed-up pull.

PR #11217, opened 2014-09-30
Light TypeAhead Refactoring (preparing for Hashtags)

leogrim:
@stkem @andrewconner @raymondng

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/typeahead/PrefixFilter.scala
+++ b/kifi-backend/modules/common/app/com/keepit/typeahead/PrefixFilter.scala
@@ -5,6 +5,8 @@ import scala.collection.mutable.ArrayBuffer
 import scala.math.min
 import java.text.Normalizer
 import com.keepit.common.logging.Logging
+import java.nio.ByteBuffer
+import com.keepit.serializer.{ BinaryFormat, ArrayBinaryFormat }
 
 object PrefixFilter {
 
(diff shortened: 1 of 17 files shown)


kifi-backend/modules/common/app/com/keepit/typeahead/PrefixFilter.scala
```diff
@@ -56,6 +58,23 @@ object PrefixFilter {
...
+  def fromByteArray[T](bytes: Array[Byte]): PrefixFilter[T] = {
+    require(bytes.length % 8 == 0, "Invalid byte array: cannot be converted to PrefixFilter")
+    val intBuffer = ByteBuffer.wrap(bytes).asLongBuffer
+    val outArray = new Array[Long](bytes.length / 8)
+    intBuffer.get(outArray)
+    new PrefixFilter(outArray)
+  }
+
+  implicit def binaryFormat[T]: BinaryFormat[PrefixFilter[T]] = ArrayBinaryFormat.longArrayFormat.map(_.data, new PrefixFilter(_))
 }
```
leogrim, line 78: @ymatsuda Ideally, I'd like to dry this out by modifying BinaryFormat so that the logic to encode a value A (no prefix) and an Option[A] (with a prefix) can be shared and made available altogether through the BinaryFormat API

stkem:
lgtm, but somebody more knowledgeable on these code path should probably have a quick read too.

PR #11218, opened 2014-09-30
Lib main or secret

raymondng:
WIP
Mainly data-level changes
Replace user_id constraint with lib_id
Replace `getPrimaryByUriAndUser` with `getPrimaryByUriAndLibrary`
Code uptake in `KeepInterner` and `UriIntegrityPluginImpl` (plus some minor refactoring)
Add main_or_secret column (data only; no code uptake yet -- looks like `LibraryCommander` would be next)

Please take a look and let me know if I have missed anything.

@andrewconner @ymatsuda


ymatsuda:
so far, so good

kifi-backend/modules/common/conf/evolutions/shoebox/253.sql
```diff
@@ -0,0 +1,17 @@
+#Shoebox
+
+# --- !Ups
+ALTER TABLE bookmark ADD COLUMN main_or_secret BOOLEAN DEFAULT TRUE AFTER is_primary;
```
2is10: Mark was talking about changing “secret” back to “private” today, and possibly using “main” to describe both the main library and the main secret (private) library. I think it would be clearer to avoid these terms.

Take a look at `LibraryKind` in Library.scala. The main and secret libraries are designated as “system generated”. It might be clearer and more durable to name this column `in_generated_library` or `in_system_library` or something along those lines.

raymondng: Yes that'd be an issue. Thanks @2is10 for the heads-up!
Calling this `in_(generated|system)_library` might be an issue if we have more than 2 types of system generated libraries (esp. if they don't share any uniqueness constraints) ... how about `user_uniq` as a designator?

2is10: `user_uniq` could be slightly misleading, since a `true` value doesn’t mean the keep is actually unique to the user.
`in_disjoint_library` is another possibility—directly describes its purpose/function.

kifi-backend/modules/common/app/com/keepit/model/Keep.scala
```diff
@@ -19,6 +19,7 @@ case class Keep(
     title: Option[String] = None,
     uriId: Id[NormalizedURI],
     isPrimary: Boolean = true, // trick to let us have multiple inactive Keeps while keeping integrity constraints
+    mainOrSecret: Boolean = true,
```
yingjieMiao: I prefer this not to have default value

andrewconner: +1 Default values in constructors have introduced quite a few bugs.

raymondng: Hmm ... can you elaborate?

andrewconner: This gets set as true everywhere an object is created, whether it should be true or not. So old code “just works”, even though we _don't_ want it to.

andrewconner: `isPrimary` is a _better_ candidate for default value, because under normal operation, it's `true`. `mainOrSecret` doesn't have that quality — it's a denormalization.

kifi-backend/modules/common/conf/evolutions/shoebox/253.sql
yingjieMiao: Default to true doesn't seem to be correct for all keeps, right? Have a plan to correct the data? 
How about first default to NULL (without adding the constraint), then apply a patch to fill the column with correct data, then apply the constrain, then remove the old constrain? 

Just throwing some thoughts. Would love to hear about a more detailed migration/patch plan, specifically the sequence of operations.

raymondng: Thanks @2is10. So far I like `in_disjoint_library` -- I will proceed with this unless others object.

@yingjieMiao I think you have cross-posted :)

To answer your question(s): yes this is mentioned in the spec (https://docs.google.com/a/42go.com/document/d/1gzLKkK3nRTNtyVFXXOhvP6Kn--ZvpCh9c8UQY7gYOwc/edit) -- take a look at the IMPACT section. Would love to hear your thoughts.

kifi-backend/modules/common/app/com/keepit/model/Keep.scala
raymondng: Agreed. Thanks @yingjieMiao and @andrewconner -- will take care of it.

kifi-backend/modules/common/conf/evolutions/shoebox/253.sql
yingjieMiao: Thanks @raymondng for the link. 

Do you think all current database rows (will) satisfy the main_or_secret constrain? 

for example

```

select bm1.id, bm2.id, bm1.user_id, bm1.uri_id, bm1.is_primary, bm1.library_id from bookmark as bm1 inner join bookmark as bm2 on bm1.user_id = bm2.user_id where bm1.uri_id = bm2.uri_id and bm1.id <> bm2.id limit 20

select * from bookmark where user_id = 1 and uri_id = 3484

```

Am I missing anything?

raymondng: No it won't. This is also stated in the spec -- patch-up is required before we apply the constraint. 
Also, github might not be the best place to discuss what's in the spec -- I'd direct your comments there. Thanks.

raymondng:
Thanks for all the feedback. Assuming there are no major objections, I will start making the incremental db-level changes tomorrow morning.

kifi-backend/modules/common/app/com/keepit/model/Library.scala
```diff
@@ -32,6 +32,10 @@ case class Library(
   def withId(id: Id[Library]) = this.copy(id = Some(id))
   def withUpdateTime(now: DateTime) = this.copy(updatedAt = now)
   def withState(myState: State[Library]) = this.copy(state = myState)
+  val isDisjoint: Boolean = kind match {
```
raymondng, line 36: @andrewconner @2is10 new `isDisjoint` field added to `Library`

kifi-backend/modules/common/conf/evolutions/shoebox/253.sql
```diff
@@ -0,0 +1,17 @@
+#Shoebox
+
+# --- !Ups
+ALTER TABLE bookmark ADD COLUMN in_disjoint_lib BOOLEAN DEFAULT TRUE AFTER is_primary;
+
+ALTER TABLE bookmark ADD CONSTRAINT lib_primary_uri UNIQUE (library_id, is_primary, uri_id);
+ALTER TABLE bookmark ADD CONSTRAINT user_disjoint_primary_uri UNIQUE (user_id, in_disjoint_lib, is_primary, uri_id);
```
raymondng: @ymatsuda @andrewconner @2is10 added `is_primary` to this constraint to handle dup keeps -- thanks @yingjieMiao for identifying the issue!

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -526,14 +526,14 @@ class LibraryCommander @Inject() (
       case Some(_) =>
         def saveKeep(k: Keep, s: RWSession): Unit = {
           implicit val session = s
-          val newKeep = keepRepo.save(Keep(title = k.title, uriId = k.uriId, url = k.url, urlId = k.urlId, visibility = library.visibility,
-            userId = k.userId, source = k.source, libraryId = Some(toLibraryId)))
+          val newKeep = keepRepo.save(Keep(title = k.title, uriId = k.uriId, url = k.url, urlId = k.urlId, visibility = toLibrary.visibility,
+            userId = k.userId, source = k.source, libraryId = Some(toLibraryId), inDisjointLib = toLibrary.isDisjoint))
```
raymondng, line 537: @andrewconner FYI

PR #11219, opened 2014-09-30
drop down for library sorting & owner images

ahsu1230:
@andrewconner @yipingliao 

few things:
- left side panel: drop down menu for library sorting (blank at first)
  - can sort libraries by "A-Z", "Z-A", "# Keeps"
- put library owner images over library-card icons
  - for invites, it's not owner of library but owner of invite
- few test changes to backend (orphan cleaner being fixed @leogrim)
- FullLibraryInfo also gets owner (BasicUser) instead of just owner externalId

next steps:
- anti-scroll / scrollbar stuff
- adding last_kept_to field to `library` table
- will address '...' elipsis for library names


kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
```diff
@@ -9,15 +9,15 @@ angular.module('kifi')
       //replace: true,
       restrict: 'A',
       templateUrl: 'layout/nav/nav.tpl.html',
-      link: function (scope /*, element, attrs*/ ) {
+      link: function (scope) {
```
yipingliao: I think that the practice in the code base is to always list all three arguments into the `link` function, and comment out the unused ones. I don't know why that is, and can definitely be convinced otherwise, but if you're going to change this, then you'll want to change them everywhere. :)

kifi-backend/modules/shoebox/angular/src/layout/nav/nav.tpl.html
```diff
@@ -21,74 +21,100 @@
...
-  </div>
-
-  <div class="kf-scrollable-libs" ng-click="scrollAround()">
-    <ul class="kf-nav-lib-users">
-      <li ng-repeat="library in userLibs" class="kf-nav-lib-item" ng-class="{ active: isActive(library.url) }">
-        <a class="kf-nav-link" ng-href="{{library.url}}">
-          <div class="kf-nav-library-card svg-red-card-with-lock" ng-if="library.visibility=='secret'"></div>
-          <div class="kf-nav-library-card svg-yellow-card-with-globe" ng-if="library.visibility=='discoverable'"></div>
-          <div class="kf-nav-library-card svg-green-card-with-globe" ng-if="library.visibility=='published'"></div>
+  <div ng-show="mainLib.numKeeps >= 0">
```
yipingliao: So, we don't list libraries if there are no keeps in the main library?

kifi-backend/modules/shoebox/angular/src/libraries/libraryService.js
```diff
@@ -3,40 +3,62 @@
...
         userLibsToShow = [],
-        invitedLibsToShow = [];
+        invitedLibsToShow = [],
+        mainLib = {},
+        secretLib = {};
 
     var fuseOptions = {
        keys: ['name'],
-       threshold: 0.3
+       threshold: 0.3 // 0 means exact match, 1 means match with anything
```
yipingliao: Thanks!

kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
ahsu1230: oh I see. I'll leave them in then lol

kifi-backend/modules/shoebox/angular/src/layout/nav/nav.tpl.html
ahsu1230: This is for loading the entire library pane at once. Before, the angular app would drop random things like the Main/Secret cards and "keeps". And things would gradually start popping up.
With this, everything under Recommendations gets loaded together.
so even if the number of keeps is 0 (>=0), we'll still display the libraries. The comparison is really to see if mainLib is defined yet or not

kifi-backend/modules/shoebox/angular/src/libraries/libraryService.js
```diff
@@ -3,40 +3,62 @@
...
-    var fuseSearch2 = new Fuse(invitedSummaries, fuseOptions);
+    var librarySummarySearch = new Fuse(librarySummaries, fuseOptions);
+    var invitedSummarySearch = new Fuse(invitedSummaries, fuseOptions);
 
     var librarySummariesService = new Clutch(function () {
       return $http.get(routeService.getLibrarySummaries).then(function (res) {
         var libs = res.data.libraries || [];
+
+        var main = _.find(libs, function (lib) {
+            return lib.kind === 'system_main';
```
yipingliao: This can be rewritten as `_.find(libs, { 'kind': 'system_main' })`. Same goes for line 29. `_.filter` on line 34 has a similar "shorthand". Jared pointed these out to me earlier this week. :)

kifi-backend/modules/shoebox/angular/src/libraries/libraryService.js
```diff
@@ -3,40 +3,62 @@
...
+            return lib.kind === 'system_secret';
+        });
+        util.replaceObjectInPlace(mainLib, main);
+        util.replaceObjectInPlace(secretLib, secret);
+        libs = _.filter(libs, function (lib) {
+          return lib.kind === 'user_created';
+        });
+        
+        var lines;
+        libs.forEach( function(lib) {
```
yipingliao: No need for space after `(` here. Same goes for line 50.

kifi-backend/modules/shoebox/angular/src/libraries/libraryService.js
```diff
@@ -3,40 +3,62 @@
...
+          lines = shortenLibName(lib.name);
+          lib.firstLine = lines[0];
+          lib.secondLine = lines[1];
+          if (lib.owner) {
+            lib.owner.image = friendService.getPictureUrlForUser(lib.owner);
```
yipingliao, line 22: I do a lot of this `owner.image` and `isMine` stuff in `libraryCard.js` too. I think we're just repeating stuff. What do you think about having all these data augmentation done in `libraryService.js`? If that gets too big we can split it into two files: `libraryActionService.js` and `libraryDecoratorService.js`. We do this for keeps (see `keepDecoratorService.js` and now that we're filling up library code maybe it's time to do that for libraries).

kifi-backend/modules/shoebox/angular/src/libraries/libraryService.js
```diff
@@ -243,16 +267,58 @@ angular.module('kifi')
         });
       },
 
-      filterLibraries: function(term) {
+      deleteLibrary: function (libraryId) {
```
yipingliao, line 226: There is already a `deleteLibrary` function in `libraryService.js` that you can just call.

yipingliao, line 226: Oops, I am in `libraryService.js` right now. I thought I was in `nav.js`. nvm.

kifi-backend/modules/shoebox/angular/src/libraries/libraryService.js
```diff
@@ -243,16 +267,58 @@ angular.module('kifi')
...
+        return $http.post(routeService.deleteLibrary(libraryId)).then( function () {
+          _.remove(librarySummaries, function (library) {
+            return library.id === libraryId;
+          });
+        });
+      },
+
+      sortLibrariesByName: function (order) {
```
yipingliao: Both `userLibsToShow` and `invitedLibsToShow` are only used by `nav.js`, as are these sorting functions. What do you think about moving these into `nav.js`?

kifi-backend/modules/shoebox/angular/src/libraries/libraryService.js
```diff
@@ -243,16 +267,58 @@ angular.module('kifi')
...
+        util.replaceArrayInPlace(invitedLibsToShow, invited);
+      },
+
+      filterLibraries: function (term) {
         var newMyLibs = librarySummaries;
         var newMyInvited = invitedSummaries;
         if (term.length) {
-          newMyLibs = fuseSearch1.search(term);
-          newMyInvited = fuseSearch2.search(term);
+          newMyLibs = librarySummarySearch.search(term);
```
yipingliao: Yay for clear names!

kifi-backend/modules/shoebox/angular/src/layout/nav/nav.tpl.html
yipingliao: Ah I see! My brain skipped over the `=` part. If we're using the existence of `numKeeps` as a proxy for detecting when `mainLib` is defined, we can just do `ng-show="mainLib.numKeeps`? Or, can we even get away with `ng-show="mainLib"`?!

kifi-backend/modules/shoebox/angular/src/libraries/libraryService.js
ahsu1230: yea I suppose we could.

kifi-backend/modules/shoebox/angular/src/libraries/libraryService.js
ahsu1230, line 22: Yes, let's at least put data augmentation in libraryService.js so it can be used in other places (no repeated calculations)
As for splitting into two files, it might not be too urgent but it definitely wouldl be good down the road =)

yipingliao:
Looks good (front-end part). We can talk about `libraryDecoratorService.js` later. Thanks!

PR #11225, opened 2014-10-01
Modify Typeahead Id generation to allow for non model typeaahead

leogrim:
@raymondng @ymatsuda 
This is a minor change to the API, but hiding the extractId operation within getAllInfos and changing the getInfos API to take the owner id as input make it possible for the info ids not to be model ids (they can be generated from whatever id space that belongs to owner id, e.g. a sequence stored in S3)


andrewconner:
Looks good

raymondng:
Sorry I missed this. lgtm!

PR #11226, opened 2014-10-01
LibraryActions work — i.e., public library API should work!

andrewconner:
Bit more to do, but this is the bulk of the backend work to make public libraries work. @ahsu1230 @yipingliao 

@raymondng Actually want to bounce an idea off you about the new Request API


kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
```diff
@@ -44,79 +45,56 @@ class LibraryController @Inject() (
       case Left(LibraryFail(message)) =>
         Future.successful(BadRequest(Json.obj("error" -> message)))
       case Right(newLibrary) =>
-        libraryCommander.createFullLibraryInfo(request.userId, newLibrary).map { lib =>
+        libraryCommander.createFullLibraryInfo(Some(request.userId), newLibrary).map { lib =>
```
ahsu1230, line 48: small thing, but could use request.userIdOpt?

ahsu1230:
really liking the UserAction andThen LibraryActions - makes the controller API methods even thinner! :+1: good job!

raymondng:
Sure just grab me whenever -- would love to hear it.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -50,7 +50,7 @@ class LibraryCommander @Inject() (
     implicit val publicIdConfig: PublicIdConfiguration,
     clock: Clock) extends Logging {
 
-  def createFullLibraryInfo(userId: Id[User], library: Library): Future[FullLibraryInfo] = {
+  def createFullLibraryInfo(userIdOpt: Option[Id[User]], library: Library): Future[FullLibraryInfo] = {
```
2is10: might help readers if we include the user’s role in the name of this argument, like `viewerIdOpt` or `viewerUserIdOpt`

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
andrewconner, line 48: No need. Other uses are using an implicit conversion to pattern match on the request to extract the `userId` if it's a `UserRequest`. In this case, it's always there. imho, better not to confuse future readers about a possibility of `None` here.

kifi-backend/modules/shoebox/app/com/keepit/shoebox/controllers/LibraryAccessActions.scala
```diff
@@ -0,0 +1,94 @@
...
+  implicit private val implicitPublicId = publicIdConfig
+  val libraryCommander: LibraryCommander
+
+  def LibraryViewAction(id: PublicId[Library]) = new ActionFilter[MaybeUserRequest] {
+    def filter[A](input: MaybeUserRequest[A]): Future[Option[Result]] = Future.successful(lookupViewAccess(id, input))
+  }
+
+  def LibraryWriteAction(id: PublicId[Library]): ActionFilter[UserRequest] = new ActionFilter[UserRequest] {
+    def filter[A](input: UserRequest[A]): Future[Option[Result]] = Future.successful(lookupWriteAccess(id, input))
+  }
```
2is10, line 30: Cool. I guess these are package-specific? I may put a copy of these in `controllers.ext` without the passphrase stuff.

andrewconner, line 30: They can be used anywhere on shoebox. The auth token / passphrase stuff shouldn't cause problems, but may introduce a very slight performance hit. Your call.

PR #11231, opened 2014-10-01
Update env.origin so that it is always referring to the origin URL of the page.

yipingliao:
@2is10 @andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/app.js
+++ b/kifi-backend/modules/shoebox/angular/src/app.js
@@ -49,7 +49,7 @@ angular.module('kifi', [
   '$location',
   function ($location) {
     var host = $location.host();
-    var dev = /^dev\.ezkeep\.com|localhost$/.test(host);
+    var dev = /^dev\.ezkeep\.com|localhost|^protractor\.kifi\.com$/.test(host);
     var origin = $location.protocol() + '://' + host  + (dev ? ':' + $location.port() : '');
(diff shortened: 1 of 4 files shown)


andrewconner:
Looks good

kifi-backend/modules/shoebox/angular/src/app.js
```diff
@@ -48,22 +48,21 @@ angular.module('kifi', [
...
       production: !dev,
       origin: origin,
-      originWeb: originWeb,
-      xhrBase: origin + '/site',
-      xhrBaseEliza: origin.replace('www', 'eliza') + '/eliza/site',
-      xhrBaseSearch: origin.replace('www', 'search'),
+      navBase: navBase,
+      xhrBase: navBase + '/site',
+      xhrBaseEliza: navBase.replace('www', 'eliza') + '/eliza/site',
+      xhrBaseSearch: navBase.replace('www', 'search'),
```
2is10: This pull request looks great. Only one minor suggestion: might consider renaming the `navBase` local variable to `navOrigin` or `dataOrigin`. The term “origin” is slightly more specific than “base”—it guarantees no path.

To be clear, I’m happy with `env.navBase` being named `env.navBase`—just talking about the name of the local variable.

yipingliao: Will do. Thanks!

PR #11233, opened 2014-10-01
Replace ownerId with owner.id.

yipingliao:
To self: use grep next time!

@andrewconner @ahsu1230


2is10:
Don’t feel too bad. I think that just recently changed in the JSON the server sends back.

PR #11236, opened 2014-10-01
verifying format of SurrogateExternalId and better error messages

eishay:


kifi-backend/modules/common/app/com/keepit/common/db/ExternalId.scala
```diff
@@ -86,9 +87,9 @@ abstract class SurrogateExternalIdCompanion[T <: SurrogateExternalId] {
         case Right(id) =>
           asOpt(id) match {
             case Some(extId) => Right(extId)
-            case None => Left(s"Unable to bind an ExternalId with $id")
+            case None => Left(s"Unable to bind aa SurrogateExternalId with $id")
```
2is10, line 90: aa → a

2is10:
:+1:

PR #11237, opened 2014-10-01
Introduce PersonalTypeahead

leogrim:
@andrewconner @stkem @ymatsuda @raymondng 

So the idea is to be able to manipulate a PrefixFilter and the way the infos I that this filter indexes together, so that we can be sure they are in sync (not a problem when these infos are models with long lasting ids, but a potential problem if infos are hashtags for which ids are derived on the fly).

Mostly, the API and consolidators that used to deal with PrefixFilter now deal with PersonalTypeahead. 

I also moved the logic dealing with cache / storage invalidation into the main trait.

### DIFF
--- a/kifi-backend/modules/abook/app/com/keepit/abook/typeahead/EContactTypeahead.scala
+++ b/kifi-backend/modules/abook/app/com/keepit/abook/typeahead/EContactTypeahead.scala
@@ -36,28 +36,33 @@ class EContactTypeahead @Inject() (
     s"$name ${info.email.address}"
   }
 
-  def refresh(userId: Id[User]): Future[PrefixFilter[EContact]] = {
-    build(userId).map { filter =>
-      cache.set(EContactTypeaheadKey(userId), filter)
-      store += (userId -> filter)
-      log.info(s"[rebuild($userId)] cache/store updated; filter=$filter")
-      filter
+  protected def create(userId: Id[User]) = {
+    getAllInfos(userId).map { allInfos =>
+      val filter = buildFilter(userId, allInfos)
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/abook/app/com/keepit/abook/typeahead/EContactTypeahead.scala
```diff
@@ -36,28 +36,33 @@ class EContactTypeahead @Inject() (
...
-            refresh(userId) // async
+            refresh(userId)
           }
-          Future.successful(filter) // return curr one
-        case None => refresh(userId)
+          filter
       }
     }
+    Future.successful(filterOpt.map(makeTypeahead(userId, _)))
   }
```
leogrim, line 66: @raymondng I'm a bit bugged by the async refresh that is triggered within the cache lookup here, since both methods will try to invalidate the cache in parallel. However I couldn't think of another way to do it since "meta" is only available within the closure. I have left it for now.

I have removed the getOrElseFuture since the cache was already invalidated in refresh (which is now called in the top level trait)

raymondng, line 66: Thanks @leogrim. Can you elaborate on your concerns? Is there a specific scenario you had in mind that'd be problematic?

leogrim, line 66: Not really, the race condition windows is really small so we should be ok. We could also break down the getOrElse into an explicit get / set but it's probably not worth it :)

PR #11243, opened 2014-10-01
Switching keep card header to Meriweather

andrewconner:
@yipingliao I cleaned up this a bit, it was using `::after`, which wasn't what we wanted all the time.

Feel free to use this font anywhere else.

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/app.styl
+++ b/kifi-backend/modules/shoebox/angular/src/app.styl
@@ -96,3 +96,16 @@ a[href]
   padding: 2px 0 0
   border-bottom: separatorBorder
   margin: 10px 0
+
+
+@font-face
+  font-family: 'Merriweather'
+  font-style: normal
+  font-weight: 400
+  src: local('Merriweather'), url(//djty7jcqog9qu.cloudfront.net/fonts/mw400.woff) format('woff')
+
+@font-face
+  font-family: 'Merriweather'
+  font-style: normal
+  font-weight: 700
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/angular/src/app.styl
```diff
@@ -96,3 +96,16 @@ a[href]
   padding: 2px 0 0
   border-bottom: separatorBorder
   margin: 10px 0
+
+
+@font-face
+  font-family: 'Merriweather'
+  font-style: normal
+  font-weight: 400
+  src: local('Merriweather'), url(//djty7jcqog9qu.cloudfront.net/fonts/mw400.woff) format('woff')
```
andrewconner, line 105: This is our CDN by the way. Cache headers are set to forever.

kifi-backend/modules/shoebox/angular/src/keep/keep.styl
```diff
@@ -272,13 +262,23 @@ keepBorderRadius = 3px
...
+  top: 2px;
+  position: relative
+  padding: 0 3px
 
 .kf-keep-description
   font-size: 16px
   line-height: 23px
   margin-top: 5px
-  font-family: "Georgia", Arial, sans-serif
+  font-family: "Helvetica", Arial, sans-serif
```
2is10, line 281: @andrewconner I must know, does your heart twinge a little every time we switch away from Georgia?

andrewconner, line 281: Nope, it's referring to the country, not the state.

yipingliao:
Thanks! And thanks for the clean-up too.

PR #11249, opened 2014-10-02
Libraries web UI: a few more small changes

yipingliao:
Here are a bunch of small UI changes for the web client. Need to be more productive tomorrow! :)

Thanks,
Yiping
@ahsu1230 @andrewconner cc @2is10


kifi-backend/modules/shoebox/angular/src/app.styl
```diff
@@ -14,6 +14,9 @@ body
   font-family: Helvetica, Arial, sans-serif
   min-width: 800px
 
+  &.modal-open
```
yipingliao, line 17: @2is10 We had a discussion about this two months ago. I had closed that pull request to wait for modals to be refactored. So here it is again.

kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
```diff
@@ -159,9 +160,9 @@ angular.module('kifi')
 
         // Wait until library is ready.
         scope.$watch(function () {
-          return scope.library.id;
+          return scope.loading;
```
yipingliao: Waiting for `library.id` is not enough; at that point in time we only have the most basic information about libraries. Now we wait for the full library information to be loaded before using that information.

andrewconner: This can be rewritten as `scope.$watch('loading')`...

kifi-backend/modules/shoebox/angular/src/app.styl
2is10, line 17: Oh, yes. This prevents the body from scrolling while modal is open, yes? Thanks for remembering.

kifi-backend/modules/shoebox/angular/src/common/modal/modal.js
```diff
@@ -14,8 +14,24 @@ angular.module('kifi')
...
-        $scope.close = this.close = function (closeAction) {
+        function escapeModal (event) {
+          if (event.which === 27) {  // Escape key
+            $scope.close();
+          }
+        }
+
+        function onOpen() {
+          $document.on('keydown', escapeModal);
+          $rootElement.find('body').addClass('modal-open');
```
2is10: `$document.find('body')` should work just as well

2is10:
LGTM

andrewconner:
Thanks, feel free to merge when ready.

kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
yipingliao: Nice. Thanks!

kifi-backend/modules/shoebox/angular/src/common/modal/modal.js
yipingliao: Will do. Thanks!

PR #11253, opened 2014-10-02
Refresh HashtagTypeahead (maybe)

leogrim:
@andrewconner this is where the decision to stick with user-owned collections are a bit problematic. Since the typeahead is owned by a library, we want to refresh it every time a tag appears for the first time or disappears from a library. But there's no easy way to track this since the Collection object can be shared across libraries (so its lifecycle doesn't align with the required typeahead invalidation). So for now I'm just trying to be aggressive, but that's probably a lot of unnecessary work.


andrewconner:
I see. We can talk about when we'll migrate the data again if you want.

PR #11255, opened 2014-10-02
on change Library (update library pane)

ahsu1230:
Sorry! @yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
+++ b/kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
@@ -49,6 +49,15 @@ angular.module('kifi')
           }
         });
 
+        $rootScope.$on('changedLibrary', function () {
+          if (scope.librariesEnabled) {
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
```diff
@@ -49,6 +49,15 @@ angular.module('kifi')
...
         });
 
+        $rootScope.$on('changedLibrary', function () {
+          if (scope.librariesEnabled) {
+            scope.mainLib = _.find(scope.librarySummaries, { 'kind' : 'system_main' });
+            scope.secretLib = _.find(scope.librarySummaries, { 'kind' : 'system_secret' });
+            scope.allUserLibs = _.filter(scope.librarySummaries, { 'kind' : 'user_created' });
+            util.replaceArrayInPlace(scope.userLibsToShow, scope.allUserLibs);
+          }
+        });
```
ahsu1230, line 57: Because librarySummaries is retrieved again, it contains mainLib & secretLib so we need to filter those out again and replace the list of libraries to show

yipingliao:
:+1:

PR #11256, opened 2014-10-02
last_kept & last_viewed to libraries

ahsu1230:
@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepInterner.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepInterner.scala
@@ -17,6 +17,7 @@ import com.keepit.heimdal.{ HeimdalContext, HeimdalServiceClient }
 import com.keepit.model._
 import com.keepit.normalizer.{ NormalizationCandidate, NormalizedURIInterner }
 import com.keepit.scraper.ScrapeScheduler
+import org.joda.time.DateTime
 import play.api.libs.json.Json
 
 import scala.util.{ Failure, Random, Success, Try }
@@ -29,6 +30,7 @@ class KeepInterner @Inject() (
   normalizedURIInterner: NormalizedURIInterner,
   scraper: ScrapeScheduler,
   keepRepo: KeepRepo,
+  libraryRepo: LibraryRepo,
   keepToCollectionRepo: KeepToCollectionRepo,
   collectionRepo: CollectionRepo,
   urlRepo: URLRepo,
@@ -216,6 +218,7 @@ class KeepInterner @Inject() (
       // A inactive keep may have had tags already. Index them if any.
       keepToCollectionRepo.getCollectionsForKeep(internedKeep.id.get) foreach { cid => collectionRepo.collectionChanged(cid) }
     }
+    libraryRepo.save(library.copy(lastKept = Some(DateTime.now())))
 
     (isNewKeep, wasInactiveKeep, internedKeep)
   }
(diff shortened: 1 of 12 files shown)


kifi-backend/modules/shoebox/app/com/keepit/model/LibraryRepo.scala
```diff
@@ -103,11 +105,11 @@ class LibraryRepoImpl @Inject() (
     getByNameOrSlugCompiled(userId, name, slug, excludeState).firstOption
   }
 
-  def getByUser(userId: Id[User], excludeState: Option[State[Library]], excludeAccess: Option[LibraryAccess])(implicit session: RSession): Seq[(LibraryAccess, Library)] = {
+  def getByUser(userId: Id[User], excludeState: Option[State[Library]], excludeAccess: Option[LibraryAccess])(implicit session: RSession): Seq[(LibraryMembership, Library)] = {
     val q = for {
       lib <- rows if lib.state =!= excludeState.orNull
       lm <- libraryMembershipRepo.rows if lm.libraryId === lib.id && lm.userId === userId && lm.access =!= excludeAccess.orNull && lm.state === LibraryMembershipStates.ACTIVE
-    } yield (lm.access, lib)
+    } yield (lm, lib)
```
ahsu1230, line 113: getByUser now returns LibraryMembership instead of just LibraryAccess so we can also have `lastViewed` field
Also will give us more flexibility in the future if we'd like to include other things from LibraryMembership

andrewconner, line 113: Sounds good.

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepInterner.scala
```diff
@@ -216,6 +218,7 @@ class KeepInterner @Inject() (
       // A inactive keep may have had tags already. Index them if any.
       keepToCollectionRepo.getCollectionsForKeep(internedKeep.id.get) foreach { cid => collectionRepo.collectionChanged(cid) }
     }
+    libraryRepo.save(library.copy(lastKept = Some(DateTime.now())))
```
andrewconner: Mind moving this to `internRawBookmarksWithStatus` and `internRawBookmark`? It would be nice to only have one public "gateway" method, but looks like there's two. Right now, the library will be updated many times if there's a batch operation.

Also, use an injected `Clock` instead of `DateTime.now()`, so that it's easier to test (in tests, we can set specific values that the injected clock will return, so we can have deterministic outcomes and specifically check edge cases).

andrewconner:
Looks good. Minor comments, feel free to merge when ready.

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepInterner.scala
yingjieMiao: is it enough to just UPDATE the `lastKept` field, instead of SAVE the entire object, hence bump up seqNum? Services like search, cortex, curator are all listening on this seqNum. For example, `getLibrariesAndMembershipsChanged` in `ShoeboxDataPipeController`. When this seqNum changed, that method also query `libMemshipRepo`. It's expensive. If we can avoid unnecessary seqNum bump up, that would be great. 

@leogrim  @ymatsuda  @stkem

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
```diff
@@ -107,24 +108,31 @@ class LibraryController @Inject() (
...
+          libraryRepo.getBySlugAndUserId(userId = owner.id.get, slug = LibrarySlug(slugStr))
+        } match {
+          case None =>
+            Future.successful(BadRequest(Json.obj("error" -> "no_library_found")))
+          case Some(lib) =>
+            if (libraryCommander.canViewLibrary(request.userIdOpt, lib, authToken, passPhrase)) {
+              request.userIdOpt.map { userId =>
+                db.readWrite { implicit s =>
+                  libraryMembershipRepo.getWithLibraryIdAndUserId(lib.id.get, userId).map { mem =>
+                    libraryMembershipRepo.save(mem.copy(lastViewed = Some(DateTime.now)))
```
yingjieMiao: again, is it necessary to save the entire object here? I am not feeling strongly against this.

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepInterner.scala
andrewconner: Agreed. There should be a special `updateLastKeptTo` method that specifically does not update the seqNum. Good catch.

andrewconner: Make sure to add a comment saying that we're purposely not incrementing the seqNum, so someone later doesn't think it was accidentally left out.

PR #11258, opened 2014-10-02
Rough cut at public libraries.

andrewconner:
@yipingliao  Initial run at making public libraries work when logged out.

The main remaining issue is that throughout the system, we immediately request data (like friend requests) when a controller initializes. This causes a lot of 403s for non-users.

Changes coming soon:
- `env` will store logged in state — this is a better place for it. So, when `profileService` realizes the client is logged out, it will set the property appropriately in `env`. 
- The server will preload `me` when the page is delivered. This is a production-only optimization, though, but should free us to feel comfortable having a lot of `if (env.userLoggedIn) { … }` throughout the code.
- We should talk about tradeoffs about when to check for logged in state. If we let the controller initialize but have a `ng-show` to actually draw it, users will get a faster render, but non-users will be slower.


kifi-backend/modules/shoebox/angular/src/app.js
```diff
@@ -123,8 +123,8 @@ angular.module('kifi', [
 ])
 
 .controller('AppCtrl', [
-  'profileService', '$window', '$rootScope', 'friendService', '$timeout', '$log',
-  function (profileService, $window, $rootScope, friendService, $timeout, $log) {
+  'profileService', '$window', '$rootScope', 'friendService', '$timeout', '$log', '$scope',
+  function (profileService, $window, $rootScope, friendService, $timeout, $log, $scope) {
```
yipingliao: Let's move `$scope` to the head of the list.

kifi-backend/modules/shoebox/angular/src/layout/header/header.tpl.html
```diff
@@ -1,20 +1,22 @@
...
-  </div>
-  <div class="kf-header-right">
-    <div class="sprite sprite-add-link" ng-click="addKeeps()">Add a link</div>
-    <div class="kf-header-icon-hover-container">
-      <a href="/friends" class="kf-header-icon-friends sprite sprite-friends">Friends</a>
-      <menu class="kf-dropdown-menu">
-        <a class="kf-dropdown-menu-item" ng-class="{'active': isActive('/friends')}" href="javascript:" ng-click="navigateToFriends()">Your Kifi Friends</a>
-        <a class="kf-dropdown-menu-item" ng-class="{'active': isActive('/invite')}" href="javascript:" ng-click="navigateToInvite()">Invite a friend to Kifi</a>
-      </menu>
+<div class="kf-header" ng-controller="HeaderCtrl" ng-show="showHeader">
```
yipingliao, line 1: So, we're not going to show any kind of header to a non-logged-in user? That means that a non-logged-in user will not be able to search since the search bar is in the header. Do we want non-logged-in users to have search functionality? Maybe something to consider?

kifi-backend/modules/shoebox/angular/src/layout/rightCol/RightColCtrl.js
```diff
@@ -19,7 +19,7 @@ angular.module('kifi')
     }, 1200);
 
     $scope.readyToDraw = function () {
-      return profileService.me.seqNum > 0 && friendsReady;
+      return profileService.userLoggedIn() === true && profileService.me.seqNum > 0 && friendsReady;
```
yipingliao, line 22: Will `profileService.me.seqNum` ever be > 0 if the user is not logged in? Can we enforce that within `profileService` instead of requiring the client of `profileService.me.seqNum` to know that this is not valid unless the `profileService.userLoggedIn()` is true?

kifi-backend/modules/shoebox/angular/src/layout/sidebar/sidebar.tpl.html
```diff
@@ -1,4 +1,4 @@
-<div kf-sidebar class="kf-col-inner kf-sidebar-inner">
+<div kf-sidebar ng-controller="SidebarCtrl" class="kf-col-inner kf-sidebar-inner" ng-show="showUserSidebar">
```
yipingliao, line 1: The left sidebar is not shown if we have a non-logged-in user. But there is no such check for the right sidebar. Are we showing the right sidebar?

kifi-backend/modules/shoebox/angular/src/profile/profileService.js
```diff
@@ -241,7 +247,12 @@ angular.module('kifi')
       return $http.post(routeService.userCloseAccount, data);
     }
 
+    function getUserLoggedIn() {
```
yipingliao, line 250: Sometimes I see getter functions in the service APIs, and sometimes the data is exposed directly as a field in the returned object. What is the rule of thumb for choosing one or the other?

kifi-backend/modules/shoebox/angular/src/layout/header/header.tpl.html
andrewconner, line 1: I think there's a different header design for logged out users. I haven't implemented this yet, so for now, hiding the user header (so that it doesn't show user-related UI elements).

kifi-backend/modules/shoebox/angular/src/profile/profileService.js
```diff
@@ -18,12 +19,17 @@ angular.module('kifi')
...
+        if (err.status === 403) {
+          userLoggedIn = false;
+        }
       });
     }, {
       cacheDuration: 5000
     });
 
     function updateMe(data) {
+      userLoggedIn = true;
```
yipingliao: Move this to above line 21. `userLoggedIn` being set to true is not dependent on `data`, and it will be clearer and more symmetric if this statement is at the same level as line 24.

kifi-backend/modules/shoebox/angular/src/layout/rightCol/RightColCtrl.js
andrewconner, line 22: I was being safe, but you're probably right.

kifi-backend/modules/shoebox/angular/src/layout/sidebar/sidebar.tpl.html
andrewconner, line 1: Right now, it handles itself because one of the required calls fails with a forbidden. Lots more to do there.

kifi-backend/modules/shoebox/angular/src/profile/profileService.js
andrewconner, line 250: This is a primitive, so we need a getter function (else, updates won't be propagated). In general I prefer exposing the object.

kifi-backend/modules/shoebox/angular/src/layout/sidebar/sidebar.js
```diff
@@ -0,0 +1,12 @@
+'use strict';
+
+angular.module('kifi')
+
+.controller('SidebarCtrl', [
+  'profileService', '$scope',
+  function (profileService, $scope) {
+    profileService.getMe().then(function () {
+      $scope.showUserSidebar = profileService.userLoggedIn();
```
yipingliao: This is different from the code in `HeaderCtrl.js`, where `showHeader` is set to `true` only if `profileService.userLoggedIn()` is really true, not just truthy. Why the difference? Related: does the app behave differently between `userLoggedIn()` being `false` vs. `undefined`?

kifi-backend/modules/shoebox/angular/src/layout/sidebar/sidebar.js
```diff
@@ -0,0 +1,12 @@
+'use strict';
+
+angular.module('kifi')
+
+.controller('SidebarCtrl', [
+  'profileService', '$scope',
```
yipingliao: Move `$scope` to head of list.

kifi-backend/modules/shoebox/angular/src/layout/header/header.tpl.html
yipingliao, line 1: Got it. Another new design!

kifi-backend/modules/shoebox/angular/src/layout/sidebar/sidebar.tpl.html
yipingliao, line 1: Got it. Thanks!

kifi-backend/modules/shoebox/angular/src/profile/profileService.js
yipingliao, line 250: Ah, thanks!

kifi-backend/modules/shoebox/angular/src/layout/sidebar/sidebar.tpl.html
andrewconner, line 1: Actually, I was wrong. There is a check.

yipingliao:
:+1:

PR #11265, opened 2014-10-02
Fixing heisentest

andrewconner:
@ahsu1230 @joshm1


andrewconner:
@ymatsuda

andrewconner:
Looks like there was an implicit contract that saving the email would be immediate, but that's no longer the case.

PR #11272, opened 2014-10-02
Add highlighting information to search results

leogrim:
@ymatsuda highlighting info added to the first chunk per request of @2is10


ymatsuda:
:+1:

PR #11274, opened 2014-10-02
adding text template for email confirmation and updating email test controller

eishay:
@joshm1

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/model/LibraryRepo.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/model/LibraryRepo.scala
@@ -1,6 +1,6 @@
 package com.keepit.model
 
-import com.google.inject.{ ImplementedBy, Inject, Singleton }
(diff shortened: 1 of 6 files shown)


kifi-backend/modules/shoebox/app/com/keepit/model/LibraryRepo.scala
```diff
@@ -29,7 +29,7 @@ trait LibraryRepo extends Repo[Library] with SeqNumberFunction[Library] {
 class LibraryRepoImpl @Inject() (
   val db: DataBaseComponent,
   val clock: Clock,
-  val libraryMembershipRepo: LibraryMembershipRepoImpl,
+  val libraryMembershipRepo: Provider[LibraryMembershipRepoImpl],
```
eishay, line 32: @ahsu1230 @andrewconner fyi

ahsu1230, line 32: gotcha

PR #11275, opened 2014-10-02
De-duping images we already know about by hash

andrewconner:
@2is10

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepImageCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepImageCommander.scala
@@ -6,8 +6,10 @@ import java.net.URLConnection
 import java.sql.SQLException
 
 import com.google.inject.{ Singleton, ImplementedBy, Inject }
+import com.keepit.common.db.slick.DBSession.RWSession
 import com.keepit.common.db.{ State, Id }
 import com.keepit.common.db.slick.Database
+import com.keepit.common.healthcheck.AirbrakeNotifier
 import com.keepit.common.logging.Logging
 import com.keepit.common.service.RequestConsolidator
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepImageCommander.scala
```diff
@@ -425,5 +436,32 @@ class KeepImageCommanderImpl @Inject() (
     }
   }
 
+  private val cdnUrl = {
+    s3ImageConfig.cdnBase.drop(s3ImageConfig.cdnBase.indexOf("//"))
+  }
+  private val ourOwnImageUrl = s"(https?\\:)?$cdnUrl/keep/([0-9a-f]{32})_\\d+x\\d+.*".r
```
2is10, line 442: Embedding `cdnUrl` in the regular expression is a little unfortunate. The controller already parses the url into our `URI` type. If it passed that through to the commander, it might be easier to verify that its `.host == cdnUrl`, and the regular expression could be simpler and would only need to be matched on `.path`.

2is10:
Thanks!

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepImageCommander.scala
andrewconner, line 442: I actually noticed that we're inconsistent with what `cdnBase` includes. Sometimes it's something like "//host.com", other times it's "http://host.com". So I'm pushing that fix further down the road for now.

2is10, line 442: That’s not what I was talking about, but okay. My suggestion wasn’t anything critical.

PR #11277, opened 2014-10-02
More aggressive HashtagTypeahead refresh

leogrim:
@2is10

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/typeahead/HashtagTypeahead.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/typeahead/HashtagTypeahead.scala
@@ -28,10 +28,16 @@ class HashtagTypeahead @Inject() (
     collectionRepo: CollectionRepo,
     db: Database) extends Typeahead[Library, Hashtag, Hashtag, LibraryHashtagTypeahead] {
 
+  protected val refreshRequestConsolidationWindow = 0 seconds
+
+  protected val fetchRequestConsolidationWindow = 15 seconds
+
+  private val refreshIfOlderThan = 15 minutes
+
   protected def get(libraryId: Id[Library]): Future[Option[LibraryHashtagTypeahead]] = Future.successful {
     import com.keepit.common.cache.TransactionalCaching.Implicits.directCacheAccess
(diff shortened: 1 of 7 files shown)


kifi-backend/modules/shoebox/app/com/keepit/typeahead/HashtagTypeahead.scala
```diff
@@ -28,10 +28,16 @@ class HashtagTypeahead @Inject() (
     collectionRepo: CollectionRepo,
     db: Database) extends Typeahead[Library, Hashtag, Hashtag, LibraryHashtagTypeahead] {
 
+  protected val refreshRequestConsolidationWindow = 0 seconds
```
2is10, line 31: Should we perhaps have a separate test configuration? It looks like maybe our tests are constraining how we configure this in production. Or could tests manually trigger a refresh?

2is10:
:+1: thanks for fixing

kifi-backend/modules/shoebox/app/com/keepit/typeahead/HashtagTypeahead.scala
leogrim, line 31: Well it's probably a good thing that this test fails, this is the behavior we want anyway don't we?

2is10, line 31: I don’t fully understand what `refreshRequestConsolidationWindow` means. Your call.

2is10, line 31: In production, I think we should be free to refresh asynchronously within a reasonable small amount of time. In tests, we may want to refresh synchronously. (I don’t know the details of how refreshing is done.) I’m just saying that I think it would be okay to require tests that rely on synchronous refresh to explicitly trigger it.

leogrim, line 31: I was already thinking of changing this API to be able to force a refresh
(that is to bypass the request consolidator). I'll work on it.

2014-10-02 16:34 GMT-07:00 Jared Jacobs notifications@github.com:

> done

PR #11279, opened 2014-10-02
'repeat' a few events in user used_kifi event

yingjieMiao:
@stkem 

c.f. asana task `Expand events rolling into user_used_kifi`


kifi-backend/modules/common/app/com/keepit/heimdal/HeimdalContext.scala
```diff
@@ -103,6 +103,10 @@ class HeimdalContextBuilder {
   def +=[T](key: String, values: Seq[T])(implicit toSimpleContextData: T => SimpleContextData): Unit = data(key) = ContextList(values.map(toSimpleContextData))
   def build: HeimdalContext = HeimdalContext(data.toMap)
 
+  def addExistingContexts(context: HeimdalContext): Unit = {
```
stkem: `addExistingContext` (no `s`)?

stkem:
lgtm

kifi-backend/modules/common/app/com/keepit/heimdal/HeimdalContext.scala
yingjieMiao: ah, good catch.

PR #11281, opened 2014-10-02
fix keep into disjoint

raymondng:
This should do it. Will merge, validate && add test as appropriate.

@andrewconner @2is10


raymondng:
Will merge and validate. Comments appreciated.

PR #11283, opened 2014-10-02
hiding LibraryMembership serialization and using more of LibraryMembersh...

eishay:
...ipView

@andrewconner @ahsu1230

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/model/view/LibraryMembershipView.scala
+++ b/kifi-backend/modules/common/app/com/keepit/model/view/LibraryMembershipView.scala
@@ -7,7 +7,7 @@ import org.joda.time.DateTime
 import play.api.libs.json.Json
 
 case class LibraryMembershipView(
-  id: Option[Id[LibraryMembership]],
+  id: Id[LibraryMembership],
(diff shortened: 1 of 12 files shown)


kifi-backend/modules/common/app/com/keepit/model/view/LibraryMembershipView.scala
```diff
@@ -0,0 +1,21 @@
...
+
+case class LibraryMembershipView(
+  id: Option[Id[LibraryMembership]],
+  libraryId: Id[Library],
+  userId: Id[User],
+  access: LibraryAccess,
+  createdAt: DateTime = currentDateTime,
+  state: State[LibraryMembership],
+  seq: SequenceNumber[LibraryMembership],
+  showInSearch: Boolean)
```
ahsu1230, line 17: was `lastViewed` left out on purpose?

eishay, line 17: yes, we do not use it outside shoebox so there's no need to send it around. 
For the same reason I'm changing the Id to not be an option since its always required when serializing.

PR #11288, opened 2014-10-03
antiscroll

ahsu1230:
not complete but should be working

small scrollbar on right sometimes appears (race condition) and doesn't appear in the correct position (also a rendering race condition)

also new icons for Main/Secret/Create Library

@yipingliao and I tried looking at what's going on, but couldn't really fix it.
@andrewconner


andrewconner:
I'll take a look. Thanks.

PR #11289, opened 2014-10-03
I think this is cool.

leogrim:
@ymatsuda @andrewconner @2is10

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/typeahead/PrefixMatching.scala
+++ b/kifi-backend/modules/common/app/com/keepit/typeahead/PrefixMatching.scala
@@ -2,6 +2,7 @@ package com.keepit.typeahead
 
 import scala.math.min
 import com.keepit.common.logging.Logging
+import scala.util.matching.Regex
 
 object PrefixMatching extends Logging {
   private[this] def initDistance(numTerms: Int): Array[Int] = {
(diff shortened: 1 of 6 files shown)


andrewconner:
Nice

kifi-backend/modules/shoebox/test/com/keepit/typeahead/PrefixMatchingTest.scala
```diff
@@ -59,5 +59,24 @@ class PrefixMatchingTest extends Specification {
       )
       prefixMatch(names, "s") === Seq("Sam Sasaki", "Satoshi Ueno", "Eishay Smith")
     }
+
+    "highlight the longest matches" in {
```
eishay, line 63: cool!

kifi-backend/modules/common/app/com/keepit/typeahead/PrefixMatching.scala
```diff
@@ -66,5 +68,30 @@ object PrefixMatching {
...
+  private def doTheyAlign(firstString: String, secondString: String): Boolean = {
+    val firstTokens = PrefixFilter.tokenizeNormalizedName(firstString)
+    val secondTokens = PrefixFilter.tokenizeNormalizedName(secondString)
+    (firstTokens.length == secondTokens.length) &&
+      (firstTokens zip secondTokens).forall { case (firstToken, secondToken) => firstToken.length == secondToken.length }
+  }
+
+  def getHighlightingRegex(query: String): Regex = {
+    val queryTerms = PrefixFilter.tokenize(query)
+    ("(^|" + PrefixFilter.tokenBoundary + ")(" + queryTerms.sortBy(-_.length).mkString("|") + ")").r
```
2is10, line 81: Probably worth testing how this behaves when the query contains characters with a special meaning in regular expressions. Might be safest to use [`Pattern.quote`](http://docs.oracle.com/javase/7/docs/api/java/util/regex/Pattern.html#quote%28java.lang.String%29) on each query term.

leogrim, line 81: Good catch, fixing.

2is10:
Very nice! thanks.

Should we add comments reminding people to keep the highlighting logic in sync with the indexing/matching logic? Maybe that’s obvious.

ymatsuda:
cool!

leogrim:
This is painting the target after the fact, so it's more or less independent of the matching logic (which needs to be light / fast, while this is used on the top results only). However it relies on the normalization / tokenization methods from PrefixFilter so will stay in sync from this perspective. I think the main way it could break is if the normalization changes and breaks the alignment between original and normalization name, as the comment mentions.

PR #11290, opened 2014-10-03
Keep to libraries from keep card

yipingliao:
Basic implementation of keeping to libraries to keep card. UI is barebones, but can keep to multiple libraries.

@andrewconner @ahsu1230


andrewconner:
Thanks, looks good.

PR #11292, opened 2014-10-03
reorganize email files

joshm1:
- moved the layout files from common to shoebox
- changed names of the base layout files to default to allow for other layouts to be specified

@stkem


eishay:
:+1:

PR #11304, opened 2014-10-03
library invite email

joshm1:
this is not complete (ready for production) - it is a wired-up version of the design/HTML
there are still edge cases we still need to account for:
- improve design when empty message from inviter
- the "Follow Library" button is a full image, and it's not possible to actually follow the library from email, so maybe it should be renamed
- non users need the passphrase
- the image is a static placeholder

@stkem 
cc @ahsu1230

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/emails/LibraryInviteEmailSender.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/emails/LibraryInviteEmailSender.scala
@@ -1,12 +1,15 @@
 package com.keepit.commanders.emails
 
 import com.google.inject.Inject
+import com.keepit.commanders.LibraryInfo
+import com.keepit.common.crypto.PublicIdConfiguration
 import com.keepit.common.db.Id
 import com.keepit.common.db.slick.Database
 import com.keepit.common.healthcheck.AirbrakeNotifier
 import com.keepit.common.logging.Logging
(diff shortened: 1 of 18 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/emails/LibraryInviteEmailSender.scala
```diff
@@ -17,24 +20,28 @@ class LibraryInviteEmailSender @Inject() (
...
       fromName = Some(Left(fromUserId)),
       from = SystemEmailAddress.NOTIFICATIONS,
       subject = s"${fullName(fromUserId)} invited you to follow ${library.name}!",
       to = toUserRecipient,
       category = NotificationCategory.User.LIBRARY_INVITATION,
-      htmlTemplate = views.html.email.libraryInvitation(toUserRecipient.left.toOption, fromUserId, inviteMsg, library.name, library.description, libLink),
-      campaign = Some("libraryInvite")
+      htmlTemplate = views.html.email.libraryInvitation(toUserRecipient.left.toOption, fromUserId, inviteMsg, libraryInfo),
+      textTemplate = Some(views.html.email.libraryInvitationText(toUserRecipient.left.toOption, fromUserId, inviteMsg, libraryInfo)),
+      templateOptions = Seq(CustomLayout).toMap
```
ahsu1230: :+1:

PR #11307, opened 2014-10-03
detect disjoint

raymondng:
Same fix as in `KeepInterner` yesterday. Should have double-checked this one -- tricky to get all moving parts aligned. My bad.

Thanks @yingjieMiao for alerting me to the errors.

@andrewconner @ymatsuda @yingjieMiao


yingjieMiao:
:+1:

andrewconner:
You think this sort of logic will be common? If so, we can centralize it to a keepCommander method if you think that would help.

raymondng:
I was thinking along the same lines. That's definitely something we should look into.

PR #11310, opened 2014-10-03
Changing fonts to always use HTTPS + new filename

andrewconner:
@yipingliao Also re-uploaded them so that CloudFront re-caches everything.

I changed CloudFront to forward the Origin header also.

Headers seem good, but my browser doesn't actually fetch the font anymore since I now have it locally. Let me know if this fixes it.


yipingliao:
Thanks!

PR #11313, opened 2014-10-04
adding lastEmailSent field to the library memebership with relevant repo...

eishay:
... methods and db patch
[DO NOT MERGE UNTIL DB PATCH IS APPLIED]
@andrewconner @joshm1


eishay:
actually, i need to add an index, hold off

eishay:
applied db patch

PR #11314, opened 2014-10-04
backend support for pymk mutual friend count in emails

joshm1:
@stkem


joshm1:
@stkem I believe this is all the backend work required for the PYMK redesign. when I get the design from Jen I'll update the HTML

stkem:
:+1:

PR #11320, opened 2014-10-04
cookie fix

raymondng:
This should fix the long-standing cookie issue wrt mobile.

@andrewconner 
cc: @thassss @jaredpetker @2is10

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/core/AuthHelper.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/core/AuthHelper.scala
@@ -33,7 +33,7 @@ import scala.util.Success
 import play.api.mvc.Cookie
 import com.keepit.common.mail.EmailAddress
 import com.keepit.social.SocialId
-import com.keepit.common.controller.AuthenticatedRequest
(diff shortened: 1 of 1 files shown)


andrewconner:
Great, thanks. Does the new actions support their existing cookie by itself?

raymondng:
The new actions currently only support the play session cookie (i.e. there's no fallback). I'm a bit hesitant to add the fallback, but if it's needed, let me know.

andrewconner:
Since mobile clients don't send the session cookie, it is needed.

raymondng:
Sure. Just to confirm -- even if we set it, the mobile clients won't return it, right?

andrewconner:
Yes, otherwise, this will never be fixed.

jaredpetker:
To try and help, if this is about the kifi_session, older clients probably
don't send it, however we can make sure newer clients will be once it is
working properly.

raymondng:
Sounds great. I'll add it then (the fallback)

raymondng:
Merging this to make the play cookie available. Will add fallback in next PR.

raymondng:
Pushed. @jaredpetker @thassss -- please let me know if you see any issues. Thanks!

PR #11322, opened 2014-10-06
track more client info in invite event

yingjieMiao:
@stkem  FYI

to get client info, we need to pass in an AuthRequest.


stkem:
lgtm

PR #11323, opened 2014-10-06
Fixing anti scroll on libraries and tag list

andrewconner:
@ahsu1230 @yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/common/directives/antiscroll.js
+++ b/kifi-backend/modules/shoebox/angular/src/common/directives/antiscroll.js
@@ -23,17 +23,6 @@ angular.module('antiscroll', ['kifi'])
           });
         };
 
-        // Antiscroll doesn't refresh automatically if nodes are added to its container element.
-        // This is an issue in Angular because this happens quite often. This bit below tries
-        // to fix potential race condition issues. It's not 100%: if a scrollbar did exist,
-        // then it won't refresh.
-        element.on('mouseenter', function () {
-          if (scope.scroller && (!scope.scroller.vertical || !scope.scroller.horizontal)) {
-            scope.refreshScroll();
-          }
-        });
-
-
         scope.refreshScroll();
 
         // http://stackoverflow.com/questions/986937/how-can-i-get-the-browsers-scrollbar-sizes
--- a/kifi-backend/modules/shoebox/angular/src/common/directives/antiscroll.js
+++ b/kifi-backend/modules/shoebox/angular/src/common/directives/antiscroll.js
@@ -23,6 +23,17 @@ angular.module('antiscroll', ['kifi'])
           });
         };
 
+        // Antiscroll doesn't refresh automatically if nodes are added to its container element.
+        // This is an issue in Angular because this happens quite often. This bit below tries
+        // to fix potential race condition issues. It's not 100%: if a scrollbar did exist,
+        // then it won't refresh.
+        element.on('mouseenter', function () {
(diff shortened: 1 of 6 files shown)


kifi-backend/modules/shoebox/angular/src/common/directives/antiscroll.js
```diff
@@ -23,6 +23,17 @@ angular.module('antiscroll', ['kifi'])
           });
         };
 
+        // Antiscroll doesn't refresh automatically if nodes are added to its container element.
+        // This is an issue in Angular because this happens quite often. This bit below tries
+        // to fix potential race condition issues. It's not 100%: if a scrollbar did exist,
+        // then it won't refresh.
+        element.on('mouseenter', function () {
```
andrewconner: This was added because the libraries list renders before the list is ready. It's probably a safe thing to have anyways to protect against updates.

kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
```diff
@@ -3,9 +3,9 @@
 angular.module('kifi')
 
 .directive('kfNav', [
-  '$location', '$window', '$rootScope', '$timeout', '$document', 'util', 
+  '$location', '$window', '$rootScope', '$timeout', '$document', 'util',
```
andrewconner, line 6: Sublime has a setting that auto-trims lines when you save. May be helpful to enable so that others don't cause strange diffs.

kifi-backend/modules/shoebox/angular/src/layout/nav/navLibs.styl
```diff
@@ -66,6 +66,7 @@
 
 .kf-scrollable-libs
   width: calc(100% + 40px)
+  position: relative
```
andrewconner, line 69: This fixed the scrollbar being at the top of the window.

kifi-backend/modules/shoebox/angular/src/tags/tags.tpl.html
```diff
@@ -1,4 +1,4 @@
-<div class="kf-tags" ng-show="!librariesEnabled">
+<div class="kf-tags" ng-if="!librariesEnabled">
```
andrewconner, line 1: This wasn't strictly needed, but saves rendering a potentially large list when the user shouldn't be showing this anyways.

kifi-backend/modules/shoebox/angular/src/common/directives/antiscroll.js
2is10: Would it be feasible to trigger antiscroll refreshing based on data changes that alter the list DOM instead of basing it on user behavior?

andrewconner: Yeah, that's probably better.

kifi-backend/modules/shoebox/angular/src/tags/tags.tpl.html
ahsu1230, line 1: right, I think we had to change from `ng-if` to `ng-show` before because there were errors showing up regarding antiscroll not finding certain objects

kifi-backend/modules/shoebox/angular/src/layout/nav/navLibs.styl
ahsu1230, line 69: oh.

andrewconner:
@2is10 I removed the user-initiated hack — it shouldn't be necessary anymore if the directives give antiscroll a head's up that the data has changed.

2is10:
:+1:

PR #11326, opened 2014-10-06
user_created event

yingjieMiao:
@stkem FYI


kifi-backend/modules/common/app/com/keepit/heimdal/EventType.scala
```diff
@@ -15,6 +15,7 @@ object UserEventTypes {
   // Growth
   val JOINED = EventType("joined")
   val INVITED = EventType("invited")
+  val CREATED = EventType("created")
```
2is10: The difference between `JOINED` and `CREATED` may not be immediately obvious to people (e.g. me). Probably worth commenting.

yingjieMiao: thanks @2is10  will add some comments

PR #11330, opened 2014-10-06
Library web front-end bug fixes

yipingliao:
Fixes:
- Edit keeps should not show up if there are no keeps (https://app.asana.com/0/15523651812135/17179575596235
- Name of library while on the library page should not be clickable.
  (https://app.asana.com/0/15523651812135/17179575596238)
- When accept invite, left pane doesn’t update
  (https://app.asana.com/0/15523651812135/17177092390144)
- When a keep has been unkept, the number of keeps is not updated in the left nav bar. (https://app.asana.com/0/15523651812135/17258517427096)

@andrewconner @ahsu1230 cc @2is10

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/keep/keep.js
+++ b/kifi-backend/modules/shoebox/angular/src/keep/keep.js
@@ -278,9 +278,6 @@ angular.module('kifi')
               keepOne(keep);
             });
 
-            // TODO(yiping): investigate why this is not working.
-            // Maybe because we need an unkeep-from-library endpoint?
-            libraryService.fetchLibrarySummaries(true);
             libraryService.addToLibraryCount(keep.libraryId, -1);
             tagService.addToKeepCount(-1);
           });
--- a/kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
+++ b/kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
@@ -12,30 +12,34 @@ angular.module('kifi')
       restrict: 'A',
       templateUrl: 'layout/nav/nav.tpl.html',
       link: function (scope , element /*, attrs*/) {
-        scope.counts = {
-          friendsCount: friendService.totalFriends(),
-          friendsNotifCount: friendService.requests.length
-        };
-
-        scope.librariesEnabled = false;
-        scope.mainLib = {};
-        scope.secretLib = {};
-
+        //
+        // Internal data.
+        //
         var allUserLibs = [];
-        scope.userLibsToShow = [];
-        scope.invitedLibsToShow = [];
 
         var w = angular.element($window);
         var scrollableLibList = element.find('.kf-scrollable-libs');
         var dropDownMenu = element.find('.kf-sort-libs-button');
 
-        // on resizing window -> trigger new turn -> reset library list height
-        w.bind('resize', function () {
-          scope.$apply(function () {
-            setLibListHeight();
-          });
-        });
 
+        //
+        // Scope data.
+        //
+        scope.librariesEnabled = false;
+        scope.mainLib = {};
+        scope.secretLib = {};
+        scope.userLibsToShow = [];
+        scope.invitedLibsToShow = [];
+
+        scope.counts = {
+          friendsCount: friendService.totalFriends(),
+          friendsNotifCount: friendService.requests.length
+        };
+
+
+        //
+        // Internal methods.
+        //
         function setLibListHeight() {
           if (scrollableLibList.offset()) {
             scrollableLibList.height(w.height() - (scrollableLibList.offset().top - w[0].pageYOffset));
@@ -44,14 +48,6 @@ angular.module('kifi')
             scope.refreshScroll();
           }
         }
-
-        var fuseOptions = {
-           keys: ['name'],
-           threshold: 0.3 // 0 means exact match, 1 means match with anything
-        };
-        var librarySummarySearch = {};
-        var invitedSummarySearch = {};
-
 
         function updateNavLibs() {
           scope.mainLib = _.find(libraryService.librarySummaries, { 'kind' : 'system_main' });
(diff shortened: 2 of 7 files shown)


kifi-backend/modules/shoebox/angular/src/keep/keep.js
```diff
@@ -278,9 +278,6 @@ angular.module('kifi')
               keepOne(keep);
             });
 
-            // TODO(yiping): investigate why this is not working.
-            // Maybe because we need an unkeep-from-library endpoint?
-            libraryService.fetchLibrarySummaries(true);
```
yipingliao, line 283: I think I wrote this because unkeeping was not updating in the left pane, which has been fixed by other code changes in this PR.

kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
```diff
@@ -12,33 +12,34 @@ angular.module('kifi')
       restrict: 'A',
       templateUrl: 'layout/nav/nav.tpl.html',
       link: function (scope , element /*, attrs*/) {
-        scope.counts = {
-          friendsCount: friendService.totalFriends(),
-          friendsNotifCount: friendService.requests.length
-        };
+        //
```
yipingliao, line 15: @ahsu1230 I cleaned up/moved things around in `nav.js` a bit. Please take a look and let me know what you think. Generally I've started to divide code into sections for: internal data, scope data, internal methods, scope methods, and watches/listeners to make the code more organized. In this case, I also left the three big sections on filtering, sorting, and scrolling together down at the bottom. Please free feel to organize code another way, but when these link functions get big, it's nice to have _some_ organization in them.

kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
```diff
@@ -12,33 +12,34 @@ angular.module('kifi')
...
+        // Internal data.
+        //
+        var allUserLibs = [];
+
+        var w = angular.element($window);
+        var scrollableLibList = element.find('.kf-scrollable-libs');
+        var dropDownMenu = element.find('.kf-sort-libs-button');
 
-        scope.librariesEnabled = false;
-        scope.librarySummaries = libraryService.librarySummaries;
```
yipingliao, line 21: I've removed this scope variable and just used `libraryService.librarySummaries` instead. This variable was never exposed on the template anyway. In general, please make sure that everything that lives inside `scope` really needs to be there.

kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
```diff
@@ -12,33 +12,34 @@ angular.module('kifi')
...
       templateUrl: 'layout/nav/nav.tpl.html',
       link: function (scope , element /*, attrs*/) {
-        scope.counts = {
-          friendsCount: friendService.totalFriends(),
-          friendsNotifCount: friendService.requests.length
-        };
+        //
+        // Internal data.
+        //
+        var allUserLibs = [];
```
yipingliao, line 18: Another case where this was a scope variable. It's actually only used internally so I made it an internal variable.

kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
```diff
@@ -48,61 +49,53 @@ angular.module('kifi')
...
-            setLibListHeight();
-            if (lastHeight === scrollableLibList.height()) {
-              $interval.cancel(promiseLibList);
-            }
-            lastHeight = scrollableLibList.height(); // probably a better way to do this - sometimes scrollbar is buggy but this secures the height
-          }
-        }, 100);
+        $rootScope.$on('librarySummariesChanged', updateNavLibs);
 
+        // Is this still necessary? Should we do librarySummariesChanged instead?
```
yipingliao: Oops, yes. I need to remove this comment.

kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
```diff
@@ -117,23 +110,51 @@ angular.module('kifi')
...
-          return loc === path || util.startsWith(loc, path + '/');
-        };
+        // on resizing window -> trigger new turn -> reset library list height
+        w.bind('resize', function () {
+          scope.$apply(function () {
+            setLibListHeight();
+          });
+        });
 
-        scope.inRecoExperiment = function () {
```
yipingliao, line 125: I removed this because it's no longer used anywhere.

kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
andrewconner, line 21: Whoops! Thanks.

kifi-backend/modules/shoebox/angular/src/libraries/library.styl
```diff
@@ -19,7 +19,7 @@
 
   .kf-keep-lib-name-container
     display: inline-block
-    width: calc(100% - 230px)
+    width: calc(100% - 80px)
```
yipingliao, line 22: Re-increased width of library name because we had previously moved the edit-lib button up there but Danny has moved it back down.

kifi-backend/modules/shoebox/angular/src/libraries/libraryService.js
```diff
@@ -162,6 +162,9 @@ angular.module('kifi')
           return librarySummary.id === libraryId;
         });
         lib.numKeeps += val;
+
+        $rootScope.$emit('libraryUpdated', lib);
```
yipingliao, line 166: So the synchronization between `libraryService`, the library page, and the left nav pane whenever library objects are mutated (e.g., followers added, keeps deleted, etc.) are getting kind of spaghetti. Currently we have various paths for doing these updates and synchronizing between the local state kept in nav, library, libraryService. @andrewconner @ahsu1230 maybe we can find some time to talk about this a little more.

andrewconner, line 166: We used to always share the same object, so it happened automatically. Can we not do this?

yipingliao:
@andrewconner Thanks for your help in thinking through the synchronization issues. I am going to merge this PR as is right now because:
- `$scope.watchCollection` is still too shallow for us since `librarySummaries` is an array and we need to watch for property changes on the array, whereas `watchCollection` only watches for changes in the array items.
- I think that we can separate out library augmentation and library state maintenance into a different service that keeps one single good and updated copy of the library data. This should reduce the amount of message passing (by object sharing) and make whatever remaining message passing that is still needed cleaner. What do you think? I made an Asana task (https://app.asana.com/0/15523651812135/17258517427106).  Right now I want to put these fixes in and do more bug fixes and feature development before circling back to make this service.

PR #11331, opened 2014-10-06
scalability enhancements to feed digest sender

joshm1:
- does not rely on a calls to shoebox to determine if a user has a primaryEmail (the sender handles this)
- uses reactive lock to ensure a a limited # of threads are consumed when queuing messages

I created an endpoint getPrimaryEmailAddressForUsers that I was originally using (before I realized I didn't need it).  Leaving it in it could be useful somewhere else.

@stkem


stkem:
lgtm

PR #11337, opened 2014-10-06
Cookie fallback

raymondng:
Add cookie fallback via secure social (for older mobile clients)
Some refactoring due to `Futur[e]`-izing `getUserIdOpt` and bringing `SecureSocialHelper` to `common`

@andrewconner 
cc: @thassss @jaredpetker


raymondng:

PR #11340, opened 2014-10-06
assign kcids to organic traffic as well

stkem:
The rules are:
-Explicit tracking always wins (i.e. if there is a link with explicit
tracking it overrides any possibly existing kcid)
-Traffic form other places on kifi.com always looses
(All of this only applies pre signup, once you sign up your icid is
permanent)

@yingjieMiao @andrewconner @raymondng


yingjieMiao:
lgtm

PR #11343, opened 2014-10-06
Bumping up the min image size.

andrewconner:
@yipingliao Danny doesn't want the smallest images to show.


yipingliao:
Looks good!

PR #11349, opened 2014-10-07
assertion => log

raymondng:
As discussed.
@andrewconner


raymondng:

PR #11350, opened 2014-10-07
Website Search API

leogrim:
@ymatsuda @stkem 
Will likely refactor this to dry it out once the other search APIs are better defined.

### DIFF
--- a/kifi-backend/modules/search/app/com/keepit/controllers/website/WebsiteSearchController.scala
+++ b/kifi-backend/modules/search/app/com/keepit/controllers/website/WebsiteSearchController.scala
@@ -1,7 +1,7 @@
 package com.keepit.controllers.website
 
 import com.keepit.common.concurrent.ExecutionContext._
-import com.keepit.common.crypto.{ PublicIdConfiguration }
+import com.keepit.common.crypto.{ PublicId, PublicIdConfiguration }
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/search/app/com/keepit/controllers/website/WebsiteSearchController.scala
```diff
@@ -71,26 +74,36 @@ class WebsiteSearchController @Inject() (
...
+        }
+      }
     }
-
-    val decorationEnumerator = reactiveEnumerator(decorationFutures)
-
-    val resultEnumerator = Enumerator("[").andThen(plainResultEnumerator).andThen(decorationEnumerator).andThen(Enumerator("]")).andThen(Enumerator.eof)
-
-    Ok.chunked(resultEnumerator).withHeaders("Cache-Control" -> "private, max-age=10")
+    futureResult.map(result => Ok(Json.toJson(result)))
```
ymatsuda, line 106: Personally I don't like to use case classes to format Json. It's inflexible and wasteful.

leogrim, line 106: I don't have a strong opinion as long as the data is not supposed to be deserialized again, so I'll rewrite this to build the Json explicitly :)

PR #11351, opened 2014-10-07
tag paginations & unkeep multiple for website

ahsu1230:
tag paginations support sorting (by name, by user's previous ordering, or by num_keeps)
unkeep multiple from a single library endpoint for website

@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/CollectionCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/CollectionCommander.scala
@@ -66,6 +66,26 @@ class CollectionCommander @Inject() (
     collections
   }
 
+  def pageCollections(sort: String, offset: Int, pageSize: Int, userId: Id[User]) = {
+    log.info(s"Getting all collections for $userId (sort $sort)")
+    val unsortedCollections = db.readOnlyMaster { implicit s =>
+      val colls = collectionRepo.getUnfortunatelyIncompleteTagSummariesByUser(userId)
+      val bmCounts = collectionRepo.getBookmarkCounts(colls.map(_.id).toSet)
+      colls.map { c => BasicCollection.fromCollection(c, bmCounts.get(c.id).orElse(Some(0))) }
+    }
+    val collections = sort match {
+      case "user" =>
+        userSort(userId, unsortedCollections)
+      case "num_keeps" =>
+        unsortedCollections.sortBy(_.keeps)(Ordering[Option[Int]].reverse)
+      case "name" =>
+        unsortedCollections.sortBy(_.name.toLowerCase)(Ordering[String])
+      case _ => // default is "last_kept"
+        unsortedCollections
+    }
+    collections.drop(offset * pageSize).take(pageSize)
+  }
+
   def updateCollectionOrdering(uid: Id[User])(implicit s: RWSession): Seq[ExternalId[Collection]] = {
     setCollectionOrdering(uid, getCollectionOrdering(uid))
   }
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/CollectionCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/CollectionCommander.scala
@@ -68,22 +68,22 @@ class CollectionCommander @Inject() (
 
   def pageCollections(sort: String, offset: Int, pageSize: Int, userId: Id[User]) = {
     log.info(s"Getting all collections for $userId (sort $sort)")
-    val unsortedCollections = db.readOnlyMaster { implicit s =>
-      val colls = collectionRepo.getUnfortunatelyIncompleteTagSummariesByUser(userId)
-      val bmCounts = collectionRepo.getBookmarkCounts(colls.map(_.id).toSet)
-      colls.map { c => BasicCollection.fromCollection(c, bmCounts.get(c.id).orElse(Some(0))) }
+    db.readOnlyMaster { implicit s =>
+      sort match {
+        case "num_keeps" =>
+          collectionRepo.getByUserSortedByNumKeeps(userId, offset, pageSize).map { summary =>
+            BasicCollection.fromCollection(summary._1, Some(summary._2))
+          }
+        case "name" =>
+          val sortedTags = collectionRepo.getByUserSortedByName(userId, offset, pageSize)
+          val bmCounts = collectionRepo.getBookmarkCounts(sortedTags.map(_.id.get).toSet)
+          sortedTags.map { c => BasicCollection.fromCollection(c.summary, bmCounts.get(c.id.get).orElse(Some(0))) }
+        case _ => // default is "last_kept"
+          val sortedTags = collectionRepo.getByUserSortedByLastKept(userId, offset, pageSize)
+          val bmCounts = collectionRepo.getBookmarkCounts(sortedTags.map(_.id.get).toSet)
+          sortedTags.map { c => BasicCollection.fromCollection(c.summary, bmCounts.get(c.id.get).orElse(Some(0))) }
+      }
     }
-    val collections = sort match {
-      case "user" =>
-        userSort(userId, unsortedCollections)
-      case "num_keeps" =>
-        unsortedCollections.sortBy(_.keeps)(Ordering[Option[Int]].reverse)
-      case "name" =>
-        unsortedCollections.sortBy(_.name.toLowerCase)(Ordering[String])
-      case _ => // default is "last_kept"
-        unsortedCollections
-    }
-    collections.drop(offset * pageSize).take(pageSize)
   }
 
   def updateCollectionOrdering(uid: Id[User])(implicit s: RWSession): Seq[ExternalId[Collection]] = {
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/KeepsController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/KeepsController.scala
@@ -372,6 +372,15 @@ class KeepsController @Inject() (
     }
   }
 
+  def page(sort: String, offset: Int, pageSize: Int) = UserAction.async { request =>
+    val collectionsFuture = SafeFuture { collectionCommander.pageCollections(sort, offset, pageSize, request.userId) }
+    for {
+      collections <- collectionsFuture
+    } yield {
+      Ok(Json.obj("collections" -> collections))
+    }
+  }
+
   def saveCollection() = UserAction { request =>
     implicit val context = heimdalContextBuilder.withRequestInfoAndSource(request, KeepSource.site).build
     collectionCommander.saveCollection(request.userId, request.body.asJson.flatMap(Json.fromJson[BasicCollection](_).asOpt)) match {
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/KeepsController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/KeepsController.scala
@@ -372,13 +372,9 @@ class KeepsController @Inject() (
     }
   }
 
-  def page(sort: String, offset: Int, pageSize: Int) = UserAction.async { request =>
-    val collectionsFuture = SafeFuture { collectionCommander.pageCollections(sort, offset, pageSize, request.userId) }
-    for {
-      collections <- collectionsFuture
-    } yield {
-      Ok(Json.obj("collections" -> collections))
-    }
+  def page(sort: String, offset: Int, pageSize: Int) = UserAction { request =>
+    val tags = collectionCommander.pageCollections(sort, offset, pageSize, request.userId)
+    Ok(Json.obj("tags" -> tags))
   }
 
   def saveCollection() = UserAction { request =>
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
@@ -371,6 +371,16 @@ class LibraryController @Inject() (
     }
   }
 
+  def removeKeep(pubId: PublicId[Library], extId: ExternalId[Keep]) = (UserAction andThen LibraryWriteAction(pubId))(parse.tolerantJson) { request =>
+    val libraryId = Library.decodePublicId(pubId).get
+    val source = KeepSource.site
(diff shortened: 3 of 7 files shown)


kifi-backend/modules/shoebox/test/com/keepit/controllers/website/LibraryControllerTest.scala
```diff
@@ -962,5 +962,66 @@ class LibraryControllerTest extends Specification with ShoeboxTestInjector {
...
+        val k4Id: ExternalId[Keep] = ExternalId()
+        val json2 = Json.obj(
+          "ids" -> Json.toJson(Seq(k1.externalId, k2.externalId, k3.externalId, k4Id))
+        )
+        val request2 = FakeRequest("POST", testPathRemove).withBody(json2)
+        val result2 = libraryController.removeKeeps(pubId1)(request2)
+        status(result2) must equalTo(OK)
+        contentType(result2) must beSome("application/json")
+
+        Json.parse(contentAsString(result2)) must equalTo(Json.parse(
```
ahsu1230, line 1009: @yipingliao 
usually for tests try to look for these `Json.parse(contentAsString(...))` for endpoint responses

kifi-backend/modules/shoebox/test/com/keepit/controllers/website/LibraryControllerTest.scala
```diff
@@ -962,5 +962,66 @@ class LibraryControllerTest extends Specification with ShoeboxTestInjector {
...
+          keeps.length === 3
+          (keeps(0), keeps(1), keeps(2))
+        }
+
+        val testPathRemove = com.keepit.controllers.website.routes.LibraryController.removeKeeps(pubId1).url
+        val k4Id: ExternalId[Keep] = ExternalId()
+        val json2 = Json.obj(
+          "ids" -> Json.toJson(Seq(k1.externalId, k2.externalId, k3.externalId, k4Id))
+        )
+        val request2 = FakeRequest("POST", testPathRemove).withBody(json2)
```
ahsu1230: @yipingliao 
look for `.withBody(json)` and then look for the json variable content to see REQUIRED input parameters for endpoints
optional parameters may not appear in tests, so then you'll have to look at the actual Controller source code

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/KeepsController.scala
```diff
@@ -372,6 +372,15 @@ class KeepsController @Inject() (
     }
   }
 
+  def page(sort: String, offset: Int, pageSize: Int) = UserAction.async { request =>
+    val collectionsFuture = SafeFuture { collectionCommander.pageCollections(sort, offset, pageSize, request.userId) }
```
andrewconner: Why in a future?

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
```diff
@@ -371,6 +371,21 @@ class LibraryController @Inject() (
     }
   }
 
+  def removeKeeps(pubId: PublicId[Library]) = (UserAction andThen LibraryWriteAction(pubId))(parse.tolerantJson) { request =>
+    Library.decodePublicId(pubId) match {
+      case Failure(ex) =>
```
andrewconner: This is a bit weird, but at this point, we actually know the `pubId` can be decoded (because of  `LibraryWriteAction`). However, it makes for interesting looking code (ie, `val libId = Library.decodePublicId(pubId).get`), so keep this if you want (it just shouldn't ever run).

kifi-backend/modules/shoebox/conf/shoebox.routes
```diff
@@ -296,6 +297,7 @@ POST    /site/libraries/:id/decline @com.keepit.controllers.website.LibraryContr
 POST    /site/libraries/:id/leave   @com.keepit.controllers.website.LibraryController.leaveLibrary(id: PublicId[Library])
 GET     /site/libraries/:id/keeps   @com.keepit.controllers.website.LibraryController.getKeeps(id: PublicId[Library], count: Int, offset: Int ?= 0, auth:Option[String] ?= None, passCode:Option[HashedPassPhrase] ?= None)
 POST    /site/libraries/:id/keeps   @com.keepit.controllers.website.LibraryController.addKeeps(id: PublicId[Library])
+DELETE  /site/libraries/:id/keeps   @com.keepit.controllers.website.LibraryController.removeKeeps(id: PublicId[Library])
```
andrewconner: Seems a bit strange to require a body for a DELETE method (especially when you're not deleting a "keeps" object, but a list of keeps in the body). The spec looks like it doesn't forbid it, but [rfc2616 4.3](http://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.3) ("if the request method does not include defined semantics for an entity-body, then the message-body SHOULD be ignored when handling the request") says the server should ignore it.

Elsewhere, we use POST to an endpoint like `/site/libraries/:id/keeps/delete`

kifi-backend/modules/shoebox/app/com/keepit/commanders/CollectionCommander.scala
```diff
@@ -66,6 +66,26 @@ class CollectionCommander @Inject() (
     collections
   }
 
+  def pageCollections(sort: String, offset: Int, pageSize: Int, userId: Id[User]) = {
+    log.info(s"Getting all collections for $userId (sort $sort)")
+    val unsortedCollections = db.readOnlyMaster { implicit s =>
+      val colls = collectionRepo.getUnfortunatelyIncompleteTagSummariesByUser(userId)
```
andrewconner: `getUnfortunatelyIncompleteTagSummariesByUser` :). You likely want a separate repo method that isn't incomplete.

Also, below, `drop(...).take(...)` means they were all loaded into memory to begin with. Pagination was to avoid loading them into memory to begin with. As-is, this is basically doing what `tagService.js` does.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/KeepsController.scala
```diff
@@ -372,6 +372,15 @@ class KeepsController @Inject() (
     }
   }
 
+  def page(sort: String, offset: Int, pageSize: Int) = UserAction.async { request =>
```
andrewconner: Naming? `KeepsController#page` doesn't seem to be related to tags.

ahsu1230: Hmm we don't have a `CollectionController` though =/

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/KeepsController.scala
```diff
@@ -372,6 +372,15 @@ class KeepsController @Inject() (
...
+    val collectionsFuture = SafeFuture { collectionCommander.pageCollections(sort, offset, pageSize, request.userId) }
+    for {
+      collections <- collectionsFuture
+    } yield {
+      Ok(Json.obj("collections" -> collections))
```
andrewconner: Might be nice to kill this naming, since nothing client-side should reference a "collection" anymore (that naming is just baggage from tags being named "collections" at one point).

kifi-backend/modules/shoebox/app/com/keepit/commanders/CollectionCommander.scala
ahsu1230: Ok yea I was thinking about that - but if we want this to support sorting & pagination together, you kind of need to sort the entire list. Or can slick do the sorting for you?

kifi-backend/modules/shoebox/conf/shoebox.routes
ahsu1230: Ok sounds good

kifi-backend/modules/shoebox/app/com/keepit/commanders/CollectionCommander.scala
andrewconner: Yeah, you can do it in the database. Sorting by number of keeps may be a bit expensive if you have quite a bit, but a _lot_ cheaper than iterating for each tag and fetching the count.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
ahsu1230: I see :+1:

kifi-backend/modules/shoebox/app/com/keepit/commanders/CollectionCommander.scala
andrewconner: It's not too bad. For our worst user, the query is less than a second:

``` sql
select c.*,count(kc.id) as keep_count from collection c, keep_to_collection kc where kc.collection_id = c.id and user_id=<userid> and kc.state = "active" and c.state="active" group by c.id order by keep_count desc limit 10;
```

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/KeepsController.scala
ahsu1230: ah okay good point.

kifi-backend/modules/shoebox/conf/shoebox.routes
2is10: Yeah, URL should fully specify what’s being deleted when using the DELETE method.

kifi-backend/modules/shoebox/test/com/keepit/controllers/website/LibraryControllerTest.scala
```diff
@@ -962,5 +962,66 @@ class LibraryControllerTest extends Specification with ShoeboxTestInjector {
...
+        val keepsToAdd =
+          RawBookmarkRepresentation(title = Some("title 11"), url = "http://www.hi.com11", isPrivate = None) ::
+            RawBookmarkRepresentation(title = Some("title 21"), url = "http://www.hi.com21", isPrivate = None) ::
+            RawBookmarkRepresentation(title = Some("title 31"), url = "http://www.hi.com31", isPrivate = None) ::
+            Nil
+
+        inject[FakeUserActionsHelper].setUser(user1)
+
+        val json = Json.obj(
+          "keeps" -> JsArray(keepsToAdd map { k => Json.toJson(k) })
```
2is10: By the way, this can be shortened to:
`"keeps" -> keepsToAdd`
With `Json.obj`, the coercion to `JsValue` types is automatic.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
```diff
@@ -371,6 +371,21 @@ class LibraryController @Inject() (
...
+        BadRequest(Json.obj("error" -> "invalid_id"))
+      case Success(libId) =>
+        val keepExtIds = (request.body \ "ids").as[Seq[ExternalId[Keep]]]
+        val source = KeepSource.site
+        implicit val context = heimdalContextBuilder.withRequestInfoAndSource(request, source).build
+        keepsCommander.unkeepManyFromLibrary(keepExtIds, libId, request.userId) match {
```
yipingliao: The `keepsCommander` also has a `unkeepOneFromLibrary` function. Should we also expose that on the controller? Currently there is a `/site/keeps/:id/delete` endpoint that calls `KeepsController.unkeep()`, but I am not sure if that does what it's supposed to do or whether going forward we should use `unkeepOneFromLibrary` anyway. Can you look into that? Thanks!

ahsu1230: You're right. This `unkeep()` will only unkeep the first instance of a keep externalId & userId. While it should still work (even if the parameters are not library aware), but we should probably be more explicit & make things library aware. But... should we replace the endpoint now?

2is10: This is a good point. I think we should use simple endpoints for simple actions (which are more common) whenever possible.

@ahsu1230 Using the `DELETE` method is appropriate for deleting a single keep, with the keep ID in the URL.

kifi-backend/modules/shoebox/conf/shoebox.routes
```diff
@@ -296,6 +297,8 @@ POST    /site/libraries/:id/decline @com.keepit.controllers.website.LibraryContr
 POST    /site/libraries/:id/leave   @com.keepit.controllers.website.LibraryController.leaveLibrary(id: PublicId[Library])
 GET     /site/libraries/:id/keeps   @com.keepit.controllers.website.LibraryController.getKeeps(id: PublicId[Library], count: Int, offset: Int ?= 0, auth:Option[String] ?= None, passCode:Option[HashedPassPhrase] ?= None)
 POST    /site/libraries/:id/keeps   @com.keepit.controllers.website.LibraryController.addKeeps(id: PublicId[Library])
+POST    /site/libraries/:id/keeps/delete   @com.keepit.controllers.website.LibraryController.removeKeeps(id: PublicId[Library])
+DELETE  /site/libraries/:id/keeps/:keepId   @com.keepit.controllers.website.LibraryController.removeKeep(id: PublicId[Library], keepId: ExternalId[Keep])
```
ahsu1230, line 301: unkeep one from a library endpoint provided (with test) @yipingliao @2is10 
does this work?

yipingliao, line 301: I thought that per @2is10 we don't need these, if we're sure that the current keeps delete endpoint works correctly w/r to libraries?

2is10:
:+1: great, thanks

kifi-backend/modules/shoebox/conf/shoebox.routes
2is10, line 301: It’s better to specify the library ID in the URL.

ahsu1230, line 301: Oh I was under the impression that because the original unkeeping doesn't clearly specify which library it is unkeeping from, we should add it in and use this new endpoint instead.
Down the road (maybe after our release) we should be able to clean this API up and consolidate the keeping/unkeeping functions

edit: ok looks like we're in agreement lol

andrewconner, line 301: I agree with Jared, moving to new APIs that specify libraries is better.

andrewconner, line 301: (ie, this is a good change)

PR #11357, opened 2014-10-07
Non-user state rework

andrewconner:
This is the very start of the split. Users and non-users now have different templates. We should do a better job separating the code (in terms of files) into separate places if it's user-only, non-user-only, and shared.

The logged in state is exposed in four ways:
- profileService.userLoggedIn
- $rootScope.userLoggedIn
- `.kf-logged-in` and `.kf-logged-out` classes added to `<html>` element
- `userLoggedInStateChange` `$broadcast` message.

We can probably consolidate these a bit over time.

Not super happy with the CSS special-casing. Will work on that further.


andrewconner:
@yipingliao

yipingliao:
Looks good!

PR #11365, opened 2014-10-07
upgrading to play 2.3.5

stkem:
mostly to fix a security vulnerability (see https://www.playframework.com/changelog)
thanks @andrewconner (or should I say typesafe subscription) for discovering
@raymondng @eishay


andrewconner:
Of note, we're not exposed, but it's good to not have open holes around.

raymondng:
Thanks. Should eng@42go.com be subscribed to the subscription mailing list?

andrewconner:
It's an alias, so can't be a member of a google group. If we changed it, we could.

andrewconner:
This is a good thing to subscribe to: https://groups.google.com/forum/#!forum/play-framework-security

PR #11366, opened 2014-10-07
pageinfo re-org & fix bug

ahsu1230:
addresses this bug: 
https://app.asana.com/0/13310647066703/17367049029076

should hopefully be fixed. Using Maps before might have caused an issue @leogrim @2is10 @ymatsuda 
And some re-org of code to return keepers info from search (so kept in which libraries is not used for now)
and use the database to figure out which keeps have correct permissions
Please take a look =)
@andrewconner @raymondng


kifi-backend/modules/shoebox/app/com/keepit/model/KeepRepo.scala
```diff
@@ -15,6 +15,7 @@ import com.keepit.common.core._
 trait KeepRepo extends Repo[Keep] with ExternalIdColumnFunction[Keep] with SeqNumberFunction[Keep] {
   def page(page: Int, size: Int, includePrivate: Boolean, excludeStates: Set[State[Keep]])(implicit session: RSession): Seq[Keep]
   def getByUriAndUser(uriId: Id[NormalizedURI], userId: Id[User])(implicit session: RSession): Option[Keep]
+  def getAllByUriAndUser(uriId: Id[NormalizedURI], userId: Id[User])(implicit session: RSession): Seq[Keep]
```
ahsu1230, line 18: `getByUriAndUser` is already used in many places and bank on the fact that there should be only one URI per user
Since we just removed this constraint, is it better to make a new repo function and consolidate later or change the other places as well (just add .headOption)

kifi-backend/modules/shoebox/app/com/keepit/commanders/PageCommander.scala
```diff
@@ -10,7 +10,7 @@ import com.keepit.common.net.URI
 import com.keepit.common.social._
 import com.keepit.model._
 import com.keepit.normalizer.{ NormalizedURIInterner, NormalizationService }
-import com.keepit.search.{ AugmentableItem, ItemAugmentationRequest, SearchServiceClient }
+import com.keepit.search.{RestrictedKeepInfo, AugmentableItem, ItemAugmentationRequest, SearchServiceClient}
```
ymatsuda: This line is missing space chars after '{' and before '}'. This indicates that this file was not compiled when you created this pull request. Please compile and run tests before making a pull request.

kifi-backend/modules/shoebox/app/com/keepit/commanders/PageCommander.scala
```diff
@@ -120,33 +121,40 @@ class PageCommander @Inject() (
     nUriOpt.map { normUri =>
       val item = AugmentableItem(normUri.id.get)
       val request = ItemAugmentationRequest.uniform(userId, item)
-      searchClient.augmentation(request).map { response =>
+
+      // get all keepers from search (read_only data)
+      val getKeepersFuture = searchClient.augmentation(request).map { response =>
```
ymatsuda, line 126: still need to call search?

ahsu1230, line 126: we need a way to retrieve all keepers for a particular URI which is hard on database, so I think so?

ymatsuda, line 126: got it.

kifi-backend/modules/shoebox/app/com/keepit/commanders/PageCommander.scala
```diff
@@ -120,33 +121,41 @@ class PageCommander @Inject() (
...
+        val userKeeps = keepRepo.getAllByUriAndUser(normUri.id.get, userId)
+        userKeeps.map { keep =>
+          val keeperId = keep.userId
+          val mine = userId == keeperId
+          keep.libraryId.map { libraryId =>
+            val lib = libraryRepo.get(libraryId)
+            val removable = libraryMembershipRepo.getWithLibraryIdAndUserId(libraryId, userId) match {
+              case Some(mem) => mem.hasWriteAccess
+              case _ => false
             }
```
2is10: `.exists(_.hasWriteAccess)` would be more concise than the match

kifi-backend/modules/shoebox/app/com/keepit/commanders/PageCommander.scala
```diff
@@ -120,33 +121,41 @@ class PageCommander @Inject() (
...
+        }
+      }
+
+      // find all keeps in database (with uri) (read_write actions)
+      val keepsData = db.readOnlyMaster { implicit session =>
+        val userKeeps = keepRepo.getAllByUriAndUser(normUri.id.get, userId)
+        userKeeps.map { keep =>
+          val keeperId = keep.userId
+          val mine = userId == keeperId
+          keep.libraryId.map { libraryId =>
```
2is10: This `.map` means we ignore keeps not in a library. All keeps should be in libraries now. I’d rather just use `keep.libraryId.get`, since it’s shorter and makes our expectations clearer. I believe we could change `Keep.libraryId` to no longer be an `Option` to avoid the issue altogether.

2is10:
Looks great, thanks Aaron!

kifi-backend/modules/shoebox/app/com/keepit/commanders/PageCommander.scala
```diff
@@ -120,33 +121,41 @@ class PageCommander @Inject() (
...
-              KeepData(
-                id = keepId,
-                mine = mine,
-                removable = mine || userId == lib.ownerId, // TODO: also make removable true if user has RW lib access
-                secret = lib.visibility == LibraryVisibility.SECRET,
-                libraryId = Library.publicId(lib.id.get))
+        db.readOnlyMaster { implicit session =>
+          restrictedKeeps.map {
+            case RestrictedKeepInfo(keepId, libId, keeperId, _) =>
+              keeperId.map(basicUserRepo.load) // get keeper info (if exists, otherwise just None)
```
leogrim: There's a bulk load action for this, you may save some roundtrips.

kifi-backend/modules/shoebox/app/com/keepit/model/KeepRepo.scala
```diff
@@ -174,6 +175,10 @@ class KeepRepoImpl @Inject() (
       bookmarks.find(_.inDisjointLib).orElse(bookmarks.headOption) // order: disjoint, custom
     }
 
+  def getAllByUriAndUser(uriId: Id[NormalizedURI], userId: Id[User])(implicit session: RSession): Seq[Keep] = {
+    (for (b <- rows if b.uriId === uriId && b.userId === userId && b.isPrimary === true && b.state === KeepStates.ACTIVE) yield b).list
+  }
```
leogrim, line 180: This only returns a user's own keeps, not the keeps from the libraries he / she follows. Is this what we want?

ahsu1230, line 180: I think this is what we'd like at least for now. And in the future, if we need it, we may want to retrieve keeps from followed libraries @2is10

2is10, line 180: When the time comes, I believe keeps in libraries the user follows can be taken from the `augmentation` response.

PR #11367, opened 2014-10-07
Keep2lib widget (very very basic)

yipingliao:
A very basic keep to library widget with zero styling. Just pops up the widget on top of the element that includes its directive and keeps track of which library has been selected (click only right now). Also contains search box to filter libraries. Tried it with add keep modal and it's okay.

@andrewconner @ahsu1230


kifi-backend/modules/shoebox/angular/src/keep/addKeep.js
```diff
@@ -110,7 +110,7 @@ angular.module('kifi')
         scope.keepToLibrary = function () {
           var url = (scope.state.input) || '';
           if (url && util.validateUrl(url)) {
-            return keepActionService.keepToLibrary([url], scope.data.selectedLibraryId).then(function (result) {
+            return keepActionService.keepToLibrary([url], scope.selection.library.id).then(function (result) {
```
yipingliao, line 113: Renaming `data` to `selection` so that when the widget is used in other scopes it's clear what this contains.

kifi-backend/modules/shoebox/angular/src/keep/addKeep.tpl.html
```diff
@@ -10,11 +10,8 @@
         <div class="kf-add-keep-private-container" ng-if="!librariesEnabled" tabindex="0"><span class="kf-add-keep-private-checkbox sprite" ng-class="{ 'sprite-checkbox-icon': !state.checkedPrivate, 'sprite-checkbox-icon-active': state.checkedPrivate }" ng-click="togglePrivate()"></span><span class="kf-add-keep-private-label">Private</span></div>
 
         <div class="kf-add-keep-to-library" ng-if="librariesEnabled">
-          <span class="sprite sprite-cards-small"></span>&nbsp;&nbsp;
-          <select ng-model="data.selectedLibraryId">
-            <option ng-repeat="lib in libraries" 
-                    value="{{lib.id}}">{{lib.name}}</option>
-          </select>
+          <span kf-keep-to-library-widget ng-click="showWidget()" class="sprite sprite-cards-small"></span>&nbsp;&nbsp;
```
yipingliao, line 13: How the widget is used - as an attribute directive.

kifi-backend/modules/shoebox/angular/src/keep/keepToLibraryWidget.js
```diff
@@ -0,0 +1,95 @@
...
+          }
+        }
+
+
+        //
+        // Scope methods.
+        //
+        scope.showWidget = function () {
+          // Create widget.
+          widget = angular.element($templateCache.get('keep/keepToLibraryWidget.tpl.html'));
```
yipingliao, line 67: I made this widget analogous to the one used for the hover friend card that appears when you hover over friends, because it's more of a widget than a modal.

kifi-backend/modules/shoebox/angular/src/keep/keepToLibraryWidget.tpl.html
```diff
@@ -0,0 +1,6 @@
+<div class="keep-to-library-widget">
+  <input ng-model="search.name">
+  <ul class="library-select-list">
+    <li ng-repeat="library in libraries | filter:search" class="library-select-option" ng-click="selectLibrary(library)">{{library.name}}</li>
```
yipingliao, line 4: Currently using Angular's `filter` to search for library names. Any reason to use the more sophisticated fuzzy search library that is being used in the left nav? Should we use it for consistency?

kifi-backend/modules/shoebox/angular/src/keep/keepToLibraryWidget.js
```diff
@@ -0,0 +1,95 @@
...
+            return;
+          }
+
+          // Clicked on a selection option? Also exit, but after a small timeout because
+          // otherwise the widget disappears too quickly for the user to see a nice
+          // highlight on their selection.
+          if (angular.element(event.target).closest('.library-select-option').length) {
+            $timeout(removeWidget, 200);
+            return;
+          }
```
ahsu1230, line 58: oh cool nice! I'll be using this same thing for the left pane sidebar =P

kifi-backend/modules/shoebox/angular/src/keep/keepToLibraryWidget.tpl.html
ahsu1230, line 4: yea maybe consistency would be good. It would be weird if one way you search is different from another on the same product lol

yipingliao, line 4: Ok. I used `filter` because it's quick. I'll move the search over and probably ask you for help if I get stuck.

PR #11370, opened 2014-10-08
fixing test compilation and adding porn filter and keeps i kept myself

eishay:


eishay:
@joshm1

eishay:
should also make the email generation be faster

PR #11371, opened 2014-10-08
Use proper endpoints for unkeeping and bulk unkeeping.

yipingliao:
DO NOT MERGE - we're having a small problem on the server with the DELETE method expecting a payload, but that should not affect client-side code (too much).

@ahsu1230 @andrewconner Using the correct endpoints for unkeeping and bulk unkeeping while libraries are enabled.

https://app.asana.com/0/15523651812135/17258517427134

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/keep/keep.js
+++ b/kifi-backend/modules/shoebox/angular/src/keep/keep.js
@@ -285,19 +285,27 @@ angular.module('kifi')
         };
 
         scope.unkeep = function (keep) {
-          keepActionService.unkeepOne(keep).then(function () {
-            keep.makeUnkept();
+          if (scope.librariesEnabled) {
+            keepActionService.unkeepFromLibrary(keep.libraryId, keep.id).then(function () {
+              keep.makeUnkept();
 
-            undoService.add('Keep deleted.', function () {
-              keepOne(keep);
-            });
+              undoService.add('Keep deleted.', function () {
+                keepOne(keep);
+              });
 
-            if (scope.librariesEnabled) {
               libraryService.addToLibraryCount(keep.libraryId, -1);
-            } else {
+            });
+          } else {
+            keepActionService.unkeepOne(keep).then(function () {
+              keep.makeUnkept();
+
+              undoService.add('Keep deleted.', function () {
+                keepOne(keep);
+              });
+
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/shoebox/angular/src/keep/keep.js
```diff
@@ -285,19 +285,27 @@ angular.module('kifi')
         };
 
         scope.unkeep = function (keep) {
-          keepActionService.unkeepOne(keep).then(function () {
-            keep.makeUnkept();
+          if (scope.librariesEnabled) {
```
yipingliao, line 288: Use different endpoints based on whether libraries are enabled. Looking forward to taking away all these hooks at the end of the month!

kifi-backend/modules/shoebox/angular/src/keeps/keepActionService.js
```diff
@@ -162,7 +162,8 @@ angular.module('kifi')
       };
 
       $log.log('keepActionService.keepToLibrary()', data);
-      var url = env.xhrBase + '/libraries/' + libraryId + '/keeps';
+
+      var url = routeService.addKeepsToLibrary(libraryId);
```
yipingliao, line 166: Unrelated to unkeeping; now using `routeService` to give us the route to adding keeps to a library.

kifi-backend/modules/shoebox/angular/src/keeps/keeps.js
```diff
@@ -181,20 +184,42 @@ angular.module('kifi')
         scope.unkeep = function (keeps) {
           var selectedKeeps = scope.selection.getSelected(keeps);
 
-          keepActionService.unkeepMany(selectedKeeps).then(function () {
-            _.forEach(selectedKeeps, function (selectedKeep) {
-              selectedKeep.makeUnkept();
-            });
+          if (scope.librariesEnabled) {
```
yipingliao, line 187: Another big if-conditional for libraries. This one is for bulk unkeeping.

andrewconner:
Hm, what's this DELETE problem?

ahsu1230:
I accidentally left a (parse.tolerantJson) in the `LibraryController.removeKeeps()` method
so the DELETE required a body, which it shouldn't.
Addressing the problem in this next PR #11366

PR #11378, opened 2014-10-08
removing library query parameter from new extension search endpoint, new route

2is10:
@leogrim 

(I think we should migrate the other extension routes to `/ext/...` too.)


leogrim:
:+1: let's make it consistent

PR #11380, opened 2014-10-08
moving non-extension admin action to admin controller

2is10:
@leogrim

### DIFF
--- a/kifi-backend/modules/search/conf/search.routes
+++ b/kifi-backend/modules/search/conf/search.routes
@@ -1,7 +1,7 @@
 ##########################################
 # Search Admin
 ##########################################
-GET         /search/instance      @com.keepit.controllers.search.SearchAdminController.instance()
+GET         /admin/search/instance         @com.keepit.controllers.search.SearchAdminController.instance()
 
 ##########################################
 # Search Extension API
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/search/conf/search.routes
```diff
@@ -1,10 +1,14 @@
 ##########################################
+# Search Admin
+##########################################
+GET         /search/instance      @com.keepit.controllers.search.SearchAdminController.instance()
```
2is10: we may want to put `/admin/` in the routes here, but I didn’t want to change the route, not knowing how it’s used today

ymatsuda: I agree. I put this for my convenience.

ymatsuda:
lgtm

kifi-backend/modules/search/conf/search.routes
```diff
@@ -1,10 +1,14 @@
 ##########################################
+# Search Admin
+##########################################
+GET         /admin/search/instance         @com.keepit.controllers.search.SearchAdminController.instance()
```
2is10, line 4: @ymatsuda wasn’t sure whether to use `/admin/search` or `/search/admin/` — feel free to switch it

PR #11382, opened 2014-10-08
removing unused user search endpoints

2is10:
@yingjieMiao I believe the extension hasn’t been using these endpoints for some time.


yingjieMiao:
lgtm

PR #11385, opened 2014-10-08
Use simpler svg icons for libraries

yipingliao:
Tossing these tiny fixes in because I need them for the keep-to-library widget anyway. :)

Using the simpler icons for libraries. Addresses:
https://app.asana.com/0/15523651812135/17179575596247
https://app.asana.com/0/15523651812135/17359884207542

@andrewconner @ahsu1230


kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.tpl.html
```diff
@@ -1,10 +1,16 @@
 <div class="kf-keep library-card">
   <div class="kf-keep-body">
     <div class="kf-keep-lib-header">
-      <div ng-switch="library.visibility" class="kf-keep-lib-icon-container">
-        <div ng-switch-when="published" class="kf-keep-lib-icon svg-green-card-with-globe"></div>
-        <div ng-switch-when="secret" class="kf-keep-lib-icon svg-red-card-with-lock"></div>
-        <div ng-switch-when="discoverable" class="kf-keep-lib-icon svg-yellow-card-with-globe"></div>
+      <div ng-switch="library.kind" class="kf-keep-lib-icon-container">
+        <div ng-switch-when="system_main" class="kf-keep-lib-icon svg-main-library-icon"></div>
```
yipingliao, line 5: Also updated the library card to use the white library icons for the main and secret libraries.

PR #11388, opened 2014-10-08
Fixing tags scroll

andrewconner:
@ahsu1230 You were right :) Needs to be `ng-show` not `ng-if`.


ahsu1230:
was the tag list scrollbar not working again?

PR #11389, opened 2014-10-08
Auth work

andrewconner:
@ahsu1230 More consistent use of LibraryViewAction, fixing a few access bugs.


kifi-backend/modules/shoebox/angular/src/app.js
```diff
@@ -78,11 +78,6 @@ angular.module('kifi', [
       _.forOwn($location.search(), function (value, key) {
         state[key] = value;
       });
-
-      if ($location.path() !== '/find') {
```
andrewconner, line 82: This was here mostly for prettiness. We can't have it anymore because we use the query param for authTokens

kifi-backend/modules/shoebox/angular/src/invite/inviteService.js
```diff
@@ -10,8 +10,6 @@ angular.module('kifi')
      *  - Kifi User is a user of Kifi, may not be a friend.
      */
 
-    $FB.getLoginStatus(); //This causes the Facebook SDK to initialize properly. Don't remove!
```
andrewconner, line 13: Interestingly, I put the comment there. It's actually fine without (I hope?)

kifi-backend/modules/shoebox/angular/src/libraries/library.js
```diff
@@ -12,6 +12,8 @@ angular.module('kifi')
     //
     var selectedCount = 0;
     var prePopulated = false;
+    var authToken = $location.search().authToken || $location.search().authCode || $location.search().accessToken || '';
```
ahsu1230, line 15: what is authCode & accessToken used for here?

andrewconner, line 15: There was still some inconsistency elsewhere regarding what we call authToken. So this is just temporary fallbacks until it's fixed everywhere.

kifi-backend/modules/shoebox/angular/src/libraries/library.tpl.html
```diff
@@ -1,24 +1,48 @@
...
-    scroll-disabled="!hasMore"
-    scroll-distance="scrollDistance"
-    scroll-next="getNextKeeps()"
-    edit-mode="editMode"
-    toggle-edit="toggleEdit"
-    update-selected-count="updateSelectedCount(numSelected)"
-  ></div>
+
+  <!-- Permission Denied -->
+  <div class="kf-main-library-keeps" ng-if="page === 'permission_denied'">
```
ahsu1230, line 30: small thing, but maybe camelCase instead?

andrewconner, line 30: Typically, we've used snake case for values (this isn't a variable)

ahsu1230, line 30: :ok_hand:

ahsu1230:
looks good to me

PR #11390, opened 2014-10-08
Test left pane (bug fixes)

ahsu1230:
Hope to fix the following bugs:
- hovering weirdness over sort library button -> only toggle dropdown menu on/off
- can hide dropdown menu by clicking anywhere in page or by clicking on an option
- prevent losing text after focus
- can type in filter term then sort the filtered list or vice-versa
- hopefully lastKept ordering should be fixed after this
- default ordering is 'last_kept'

@andrewconner @yipingliao


kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
```diff
@@ -180,110 +166,133 @@ angular.module('kifi')
         };
 
         scope.onFilterChange = function () {
```
andrewconner, line 177: You could use a `watch` for this, which could basically remove the need to have all your calls to `changeList` elsewhere as well.

PR #11391, opened 2014-10-08
pymk redesign + updates to all "kifi black" emails

joshm1:
- all emails were reviewed and updated to look better in mobile (particularly any client that supports CSS media queries)
- the "closingLines" field in EmailToSend is removed since these lines will go with the main email content instead of below the tip

@stkem


stkem:
lgtm

PR #11399, opened 2014-10-09
Ext search api

leogrim:
@ymatsuda @2is10 @stkem 
Please mostly review what's going on in AugmentedItem :)


leogrim:
Danny wants the number of unique keepers for a specific result, and though he agrees we could fake it, I'm not very satisfied by that. I think we should just go and build the set when we traverse all the keeps in the augmentation step. Currently it's not a problem and later on we can have some cap or use a probabilistic counting method. I'll add this tomorrow morning.

kifi-backend/modules/search/app/com/keepit/controllers/website/WebsiteSearchController.scala
```diff
@@ -148,59 +148,31 @@ class WebsiteSearchController @Inject() (
...
-    val myKeeps = myRestrictedKeeps.flatMap(_.keptIn).sortBy(scores.byLibrary).map(libraries(_))
+  def toWebsiteSearchHit(kifiShardHit: KifiShardHit, userId: Id[User], summary: URISummary, augmentedItem: AugmentedItem, allUsers: Map[Id[User], BasicUser], allLibraries: Map[Id[Library], BasicLibrary]): JsObject = {
 
+    val myKeeps: Seq[BasicLibrary] = augmentedItem.myKeeps.map(myKeep => allLibraries(myKeep.keptIn.get))
     val moreKeeps = {
-      var uniqueKeepers = mutable.HashSet[Id[User]]()
-      uniqueKeepers += userId
-      moreRestrictedKeeps.sortBy(keep => (keep.keptBy.map(scores.byUser), keep.keptIn.map(scores.byLibrary))).collect {
+      import scala.collection.mutable
+      var uniqueKeepers = mutable.Set[Id[User]]()
```
2is10, line 161: can be `val`

leogrim, line 161: Good catch! Btw, this is pretty arbitrary, we'll see how the design evolves on the website.

kifi-backend/modules/search/app/com/keepit/search/AugmentationCommander.scala
```diff
@@ -174,3 +192,78 @@ class AugmentationCommanderImpl @Inject() (
...
+      val keepCategory = {
+        if (keep.keptBy.exists(_ == userId)) myKeeps
+        else if (keep.keptIn.exists(libraries.contains)) keepsFromMyLibraries
+        else if (keep.keptBy.exists(friends.contains)) keepsFromMyFriends
+        else otherKeeps
+      }
+      keepCategory += keep
+    }
+    val moreKeeps = keepsFromMyLibraries ++ keepsFromMyFriends ++ otherKeeps
+    (myKeeps.toList, moreKeeps.toList)
```
2is10, line 257: Is there a reason not to use `.partition`? 5 lines instead of 15:

``` scala
  sortedKeeps.partition { keep =>
    keep.keptBy.exists(_ == userId) ||         // my keeps
    keep.keptIn.exists(libraries.contains) ||  // keeps from my libraries
    keep.keptBy.exists(friends.contains)       // keeps from my friends
  }
```

leogrim, line 257: This would lead to the same sets of keeps, but the ordering would be different. This code is meant to reorder the keeps (already sorted by user / library relevance) by the corresponding blocks.

2is10, line 257: I see. `.groupBy` would probably result in slightly less code, but may not be worth rewriting.

2is10:
LGTM

ymatsuda:
The number of UNIQUE keepers? Computing it EXACTLY can become very expensive quickly. I don't like it. No user knows the exact number of unique keepers. Why would we want to pay a cost for such a thing? Another approach is approximation like HyperLogLog. I did a similar thing for LInkedIn Signal search to estimate sharing user counts per link.

kifi-backend/modules/search/app/com/keepit/search/AugmentationCommander.scala
leogrim, line 257: Actually I had that breakdown available in AugmentedItem, but so far it's not used so hiding it in this function for now.

leogrim:
I knew you would have a cool approximation algorithm in store :) And that's exactly what I was hoping for in a "probabilistic counting method". 

I'm just wondering how quickly the exact computation will become too expensive, I actually don't care so much about it being exact as it being trivial to implement. My guess is that at our size, we can afford it (the most kept page has been kept 1377 times) and save the development time for now. 

When we grow and some page breaks a certain threshold, we can first cap the count to defend the system (and derive a fake number from the total number of keeps - this would only affect the count of a few popular pages, since it will most likely follow a power-law) and at the same time work on implementing something like HyperLogLog. 

I just think we should provide the best possible experience to our early users as long as we can afford it, and make compromises later on. Am I missing something that would break the current experience?

ymatsuda:
OK. Fair enough.

PR #11401, opened 2014-10-09
adding access to individual library infos

ahsu1230:
@jaredpetker & @yipingliao 
would like someway to figure out if a user can tell if he or she is following a particular library

Currently right now, in order to figure that out, they would go to an individual library (by path or by libraryId), and hope to see if the userID is among the `followers` field.
However, this `followers` field is finite and won't always load all the particular followers, so it's not good to rely on this. Alternately, you would have to get all library summaries and find the library within that list.

Instead, we can include an 'access' field.
{ library: {...}, access: "read_only" }
If the user is signed in & visits a library, the endpoint will give back what access the user has (i.e. `owner`, `read_only`)
If the user is not following the library, no membership will be available, so the access field will return `none`
If the user is not a kifi user, access field will return `none`

@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
@@ -90,8 +90,15 @@ class LibraryController @Inject() (
   def getLibraryById(pubId: PublicId[Library]) = (MaybeUserAction andThen LibraryViewAction(pubId)).async { request =>
     val id = Library.decodePublicId(pubId).get
     val lib = db.readOnlyMaster { implicit s => libraryRepo.get(id) }
-    libraryCommander.createFullLibraryInfo(request.userIdOpt, lib).map { library =>
-      Ok(Json.obj("library" -> Json.toJson(library)))
+    libraryCommander.createFullLibraryInfo(request.userIdOpt, lib).map { libInfo =>
+      val accessStr = request.userIdOpt.map { userId =>
+        db.readOnlyMaster { implicit s =>
+          libraryMembershipRepo.getWithLibraryIdAndUserId(id, userId)
+        }.map(_.access.value)
+      }.flatten.getOrElse {
+        "none"
+      }
+      Ok(Json.obj("library" -> Json.toJson(libInfo), "access" -> accessStr))
     }
   }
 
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
```diff
@@ -90,8 +90,15 @@ class LibraryController @Inject() (
...
-      Ok(Json.obj("library" -> Json.toJson(library)))
+    libraryCommander.createFullLibraryInfo(request.userIdOpt, lib).map { libInfo =>
+      val accessStr = request.userIdOpt.map { userId =>
+        db.readOnlyMaster { implicit s =>
+          libraryMembershipRepo.getWithLibraryIdAndUserId(id, userId)
+        }.map(_.access.value)
+      }.flatten.getOrElse {
+        "none"
+      }
+      Ok(Json.obj("library" -> Json.toJson(libInfo), "access" -> accessStr))
```
ahsu1230, line 101: Decided not to include `access` in `FullLibraryInfo` because it's dependent on the user as opposed to the library itself, y'know?

andrewconner:
Sounds good

kifi-backend/modules/shoebox/test/com/keepit/controllers/website/LibraryControllerTest.scala
```diff
@@ -213,39 +213,75 @@ class LibraryControllerTest extends Specification with ShoeboxTestInjector {
...
+             |  "username":"${basicUser1.username.get.value}"
+             |  },
+             |"collaborators":[],
+             |"followers":[],
+             |"keeps":[],
+             |"numKeeps":0,
+             |"numCollaborators":0,
+             |"numFollowers":0
+           |},
+           |"access":"owner"
```
yipingliao, line 247: What is the `access` string when the user is a follower?

yipingliao:
Thanks for your help!

kifi-backend/modules/shoebox/test/com/keepit/controllers/website/LibraryControllerTest.scala
ahsu1230, line 247: if a user is a follower the access can be `owner`, `read_write`, `read_insert` or `read_only`.
So you can also differentiate whether a user is the owner of the library or not

if the user does not have access (not signed in, has invite but not following, etc.) the string should be `none`

yipingliao, line 247: Does access of `owner` mean that the use is the owner of the library? So is it correct to say that `read_write`, `read_insert`, or `read_only` equate to follower status?

ahsu1230, line 247: yup for both questions :smiley_cat:

yipingliao, line 247: Thanks for the clarification!

PR #11404, opened 2014-10-09
Reapply "simplifying the new extension search endpoint a bit + shorter query parameter names"

leogrim:


leogrim:
Double revert, because why not.

PR #11406, opened 2014-10-09
Logging in to libraries

andrewconner:
@yipingliao


kifi-backend/modules/shoebox/angular/src/common/services/clutch.js
```diff
@@ -30,10 +30,10 @@ angular.module('kifi')
         var key = stringize(arguments);
         var hit = this._cache[key];
 
-        if (!hit) {
+        if (!hit || (!hit.value && !hit.activeRequest)) {
```
andrewconner, line 33: This was a bug in Clutch. If the first request failed (ie, 403 forbidden), it would be stuck in a bad state.

yipingliao:
:+1:

PR #11408, opened 2014-10-09
adding /ext/ prefix to last remaining extension search route

2is10:
@leogrim


leogrim:
lgtm, but I won't merge it :)

PR #11410, opened 2014-10-09
token-based endpoints (site)

raymondng:
Right now it's the same as mobile counterparts. Would be happy to make any changes required for site.

`OAuth2TokenInfo` replaces use of `securesocial.core.OAuth2Info` in the new endpoints. That further removes dependency on securesocial, and affords additional flexibility (should we choose to add signature fields, etc.)

@andrewconner


raymondng:
Minor change. Merging. Comments appreciated.

PR #11411, opened 2014-10-09
Count unique keepers (naive)

leogrim:
@ymatsuda @2is10

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/search/AugmentableItem.scala
+++ b/kifi-backend/modules/common/app/com/keepit/search/AugmentableItem.scala
@@ -24,7 +24,7 @@ object RestrictedKeepInfo {
   implicit val format = Json.format[RestrictedKeepInfo]
 }
 
-case class AugmentationInfo(keeps: Seq[RestrictedKeepInfo], otherPublishedKeeps: Int, otherDiscoverableKeeps: Int)
+case class AugmentationInfo(keeps: Seq[RestrictedKeepInfo], otherPublishedKeeps: Int, otherDiscoverableKeeps: Int, uniqueKeepers: Int = -1)
 object AugmentationInfo {
   implicit val format = Json.format[AugmentationInfo]
 }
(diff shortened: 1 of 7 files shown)


kifi-backend/modules/common/app/com/keepit/search/AugmentableItem.scala
```diff
@@ -24,7 +24,7 @@ object RestrictedKeepInfo {
   implicit val format = Json.format[RestrictedKeepInfo]
 }
 
-case class AugmentationInfo(keeps: Seq[RestrictedKeepInfo], otherPublishedKeeps: Int, otherDiscoverableKeeps: Int)
+case class AugmentationInfo(keeps: Seq[RestrictedKeepInfo], otherPublishedKeeps: Int, otherDiscoverableKeeps: Int, uniqueKeepers: Int = -1)
```
leogrim, line 27: I'll remove the -1 after it's deployed.

2is10:
LGTM

ymatsuda:
lgtm

kifi-backend/modules/search/app/com/keepit/search/graph/library/LibraryIndexer.scala
```diff
@@ -44,6 +44,14 @@ class LibraryIndexer(indexDirectory: IndexDirectory, shoebox: ShoeboxServiceClie
...
   }
+
+  def isSecret(lib: Id[Library]): Boolean = {
+    getSearcher.getLongDocValue(LibraryFields.visibilityField, lib.id).exists(_ == LibraryFields.Visibility.SECRET)
+  }
+
+  def getRecord(libraryId: Id[Library]): Option[LibraryRecord] = {
+    getSearcher.getDecodedDocValue(LibraryFields.recordField, libraryId.id)
+  }
 }
```
leogrim, line 55: @ymatsuda I have encapsulated these basic retrieval methods into the LibraryIndexer itself. We could move some of the methods in SearchFactory here (e.g. getLibraries) when they are specific to a single index. What do you think?
I'm afraid I might be missing a Lucene pitfall, may the Searcher change on you if you call "getSearcher" repeatedly instead of holding to the same searcher instance?

PR #11418, opened 2014-10-09
User signup

raymondng:
Endpoint for email/pwd signup (one-step)
Refactored code to allow for re-use
No auto-library follow yet -- that's next

@andrewconner


raymondng:
Merging to build on it. Comments appreciated.

PR #11421, opened 2014-10-09
email to keep and bookmark import tips

joshm1:
@stkem 

Cannot merge b/c waiting for advice from Jen on copy and which emails to include these tips with.


kifi-backend/modules/shoebox/app/views/email/tips/bookmarkImport.scala.html
```diff
@@ -0,0 +1,7 @@
+@()
+@import com.keepit.common.mail.template.helpers._
+@email.tips.didYouKnow() {
+  You can can import your links into kifi from Pocket, Delicious, Kippt, Instapaper, or pinboard...Bla bla
```
stkem: I assume this is no final :-)

stkem:
:+1: 
We'll need to make sure that any link in the bookmark import tip (which it should probably have, at least eventually) works on mobile as well.

joshm1:
@stkem somehow some non-tip-related style updates found their way into this branch instead of others.  since the tips are disabled anyway, I'm merging this. I'll get with product to help finalize the copy.

PR #11422, opened 2014-10-09
Complete Ext Search API (but I want to make this code cleaner)

leogrim:
@2is10 @ymatsuda

### DIFF
--- a/kifi-backend/modules/search/app/com/keepit/controllers/ext/ExtSearchController.scala
+++ b/kifi-backend/modules/search/app/com/keepit/controllers/ext/ExtSearchController.scala
@@ -9,12 +9,29 @@ import com.keepit.controllers.util.SearchControllerUtil
 import com.keepit.model._
 import com.keepit.model.ExperimentType.ADMIN
 import com.keepit.search.graph.library.LibraryIndexer
-import com.keepit.search.{ AugmentationCommander, SearchCommander }
+import com.keepit.search.{ RestrictedKeepInfo, AugmentationCommander, SearchCommander }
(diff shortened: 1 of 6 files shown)


kifi-backend/modules/search/app/com/keepit/controllers/ext/ExtSearchController.scala
```diff
@@ -9,12 +9,29 @@ import com.keepit.controllers.util.SearchControllerUtil
...
+import com.keepit.common.crypto.PublicIdConfiguration
+import play.api.libs.json.JsBoolean
+import play.api.libs.json.JsObject
+
+import java.math.BigDecimal
+
+object ExtSearchController {
+  private[ExtSearchController] val maxKeepersShown = 15
+  private[ExtSearchController] val maxLibrariesShown = 15
+  private[ExtSearchController] val maxTagsShown = 15
```
2is10, line 26: I’d recommend 20, 10, 15 as the initial values for these. I think they should be based on how much screen space each thing requires. Users can be represented quite compactly (just photo circles—or a list of names if you hover), tags are medium-sized (the tag text), libraries are large (picture and library name). We may be able to lower the library limit even further, depending on how we end up showing them when you hover over “+N”.

We might also consider allowing the extension to supply these limits in a query parameter (not necessary now).

2is10:
:+1: Works for me. Thanks!

kifi-backend/modules/search/app/com/keepit/controllers/ext/ExtSearchController.scala
leogrim, line 26: Done!

PR #11423, opened 2014-10-09
auto-join lib: baby steps

raymondng:
Add auto-join for signup (email/pwd only)
No auto-invites (yet)

@andrewconner


raymondng:
Merging to build on it. Feedback appreciated.

andrewconner:
Looks good. We may want to do some architecture planning about the general case of post-signup actions based on pre-existing actions. Elsewhere we auto-connect users who were invited. Here we're auto-following the library you clicked on when joining. Elsewhere we will auto-invite to libraries you were invited to before. Possible we should add you (Kifi user) to conversations that were with your email address when you join. Etc, etc. Would love to hear your thoughts.

PR #11424, opened 2014-10-09
Manage Tags page

ahsu1230:
@andrewconner @yipingliao 
- additional header icon (only if in libraries experiment)

Displays all tags
Smart-scroll with paginated tags
Sort by Name or Most Keeps
Can filter but uses Fuse.search, which only works if ALL the tags are presently in the library page (may need to use real tag search later)
Create Library link uses the library widget to keep to library
Can delete tag. But undo tag deletion may need some extra work


kifi-backend/modules/shoebox/angular/src/tagManage/manageTagService.js
```diff
@@ -0,0 +1,60 @@
...
+
+angular.module('kifi')
+
+.factory('manageTagService', [
+  '$http', 'routeService', 'Clutch', '$q',
+  function ($http, routeService, Clutch, $q) {
+    var list = [];
+    var more = true;
+    var offset = 0;
+    var pageSize = 10;
```
ahsu1230, line 11: whoops need to change. 30? 40?

kifi-backend/modules/shoebox/angular/src/tagManage/manageTagService.js
```diff
@@ -0,0 +1,60 @@
+'use strict';
```
andrewconner, line 1: Thanks for creating a new service for this!

kifi-backend/modules/shoebox/angular/src/tagManage/manageTagService.js
andrewconner, line 11: 100? They're all cheap.

andrewconner, line 11: We just don't want someone getting 10,000 at once.

kifi-backend/modules/shoebox/angular/src/tagManage/manageTagService.js
```diff
@@ -0,0 +1,60 @@
...
+
+angular.module('kifi')
+
+.factory('manageTagService', [
+  '$http', 'routeService', 'Clutch', '$q',
+  function ($http, routeService, Clutch, $q) {
+    var list = [];
```
andrewconner, line 8: It's probably better to store state in the Manage tag controller, and not the service. Thoughts, @yipingliao?

kifi-backend/modules/shoebox/angular/src/tagManage/manageTagService.js
```diff
@@ -0,0 +1,60 @@
...
+        } else {
+          offset += 1;
+          list.push.apply(list, res.data.tags);
+        }
+        return res.data;
+      });
+    });
+
+    var api = {
+      reset: function () {
```
andrewconner, line 27: If state was in the controller, this wouldn't be needed.

kifi-backend/modules/shoebox/angular/src/tagManage/manageTagService.js
```diff
@@ -0,0 +1,60 @@
...
+        list.length = 0;
+        offset = 0;
+        more = true;
+        manageTagRemoteService.expireAll();
+      },
+      getMore: function (sort) {
```
andrewconner, line 33: This is hard to test, since every time you call it, it will return something else with the same inputs. A possibly better signature may be: `getPage(offset, pageSize, sort)`

kifi-backend/modules/shoebox/angular/src/tagManage/tagManage.js
```diff
@@ -0,0 +1,111 @@
...
+        return $scope.selectedSort;
+      }, function () {
+      manageTagService.reset();
+      manageTagService.getMore($scope.selectedSort);
+    });
+
+    //
+    // Keeping to Library
+    //
+    $scope.librariesEnabled = false;
```
andrewconner, line 45: Honestly, I wouldn't worry about this one too much. If they find this page, so be it. Non-library users won't see a link to it, right?

kifi-backend/modules/shoebox/angular/src/tagManage/tagManage.js
```diff
@@ -0,0 +1,111 @@
...
+      return libraryService.isAllowed();
+    }, function (newVal) {
+      $scope.librariesEnabled = newVal;
+      if ($scope.librariesEnabled) {
+        libraryService.fetchLibrarySummaries().then(function () {
+          $scope.libraries = _.filter(libraryService.librarySummaries, function(lib) {
+            return lib.access !== 'read_only';
+          });
+          $scope.selection = $scope.selection || {};
+          $scope.selection.library = _.find($scope.libraries, { 'name': 'Main Library' });
```
andrewconner, line 58: Hm, I think we need a better way to find the system libraries. The name may change as product wants.

kifi-backend/modules/shoebox/angular/src/tagManage/tagManage.js
```diff
@@ -0,0 +1,111 @@
...
+      $scope.isFilterFocused = false;
+    };
+
+    $scope.clearFilter = function () {
+      $scope.filter.name = '';
+      $scope.onFilterChange();
+    };
+
+    $scope.onFilterChange = function () {
+      tagSearch = new Fuse($scope.tagList, fuseOptions);
```
andrewconner, line 100: Do we need to create a new one every time the filters change? Expensive.

kifi-backend/modules/shoebox/angular/src/tagManage/tagManage.tpl.html
```diff
@@ -0,0 +1,52 @@
...
+      	</div>
+
+      	<div class="manage-tag-content" smart-scroll scroll-distance="tagsScrollDistance" scroll-disabled="isTagScrollDisabled()" scroll-next="manageTagScrollNext()">
+      		<ul class="manage-tag-list">
+      			<li class="manage-tag-row" ng-repeat="tag in tagsToShow">
+      				<div class="manage-tag-row-name-section">
+                                    <div class="manage-tag-row-name">{{tag.name}}</div>
+                              </div>
+                              <span class="manage-tag-row-num-keeps-section">
+                                    <ng-pluralize count="tag.keeps | number:0"
```
andrewconner: I notice this is used this way (as a custom tag) in a couple places. Is there a reason? I only ask because we don't invoke directives like that anywhere else. @ahsu1230 @yipingliao

kifi-backend/modules/shoebox/angular/src/layout/header/header.tpl.html
```diff
@@ -9,6 +9,7 @@
     </div>
     <div class="kf-header-right">
       <div class="sprite sprite-add-link" ng-click="addKeeps()">Add a link</div>
+      <a href="/tags" class="kf-header-manage-tags sprite svg-tag-black" ng-class="{'svg-tag-hover': hover}" ng-mouseenter="hover = true" ng-mouseleave="hover = false" ng-if="librariesEnabled">Manage your tags</a>
```
yipingliao: nit: I like to put `ng-if`, `ng-show`, and `ng-switch` stuff at the beginning of a tag so we can quickly see what affects the html element being displayed.

kifi-backend/modules/shoebox/angular/src/layout/header/header.tpl.html
```diff
@@ -16,7 +17,7 @@
           <a class="kf-dropdown-menu-item" ng-class="{'active': isActive('/invite')}" href="javascript:" ng-click="navigateToInvite()">Invite a friend to Kifi</a>
         </menu>
       </div>
-      <a href="/profile" class="sprite sprite-settings">Settings</div>
+      <a href="/profile" class="sprite sprite-settings">Settings</a>
```
yipingliao, line 20: Good catch!

kifi-backend/modules/shoebox/angular/src/tagManage/manageTagService.js
yipingliao, line 8: Yes, that's consistent with what we've been doing.

kifi-backend/modules/shoebox/angular/src/tagManage/tagManage.js
yipingliao, line 58: I've seen elsewhere that we use `library.kind` and compare it to `system_main`. Is that a bit more robust?

kifi-backend/modules/shoebox/angular/src/tagManage/tagManage.tpl.html
```diff
@@ -0,0 +1,52 @@
+<div ng-controller="ManageTagCtrl">
```
yipingliao, line 1: Indentation and white space in this file looks super weird.

kifi-backend/modules/shoebox/angular/src/tagManage/tagManage.tpl.html
yipingliao: Are you talking about `ng-pluralize`?

kifi-backend/modules/shoebox/angular/src/tagManage/tagManage.js
andrewconner, line 58: Much more-so. That's a static system name for it, so it won't change any time soon.

kifi-backend/modules/shoebox/angular/src/tagManage/tagManage.tpl.html
andrewconner, line 1: Ah, it's tabs. We use two spaces everywhere.

kifi-backend/modules/shoebox/angular/src/keep/keepToLibraryWidget.js
```diff
@@ -67,6 +69,9 @@ angular.module('kifi')
           // highlight on their selection.
           if (angular.element(event.target).closest('.library-select-option').length) {
             removeTimeout = $timeout(removeWidget, 200);
+            if (_.isFunction(scope.clickAction)) {
```
yipingliao, line 72: :) You'll probably also want to let the user press ENTER on a library to do the action too, since we allow up and down key pressing.

kifi-backend/modules/shoebox/angular/src/tagManage/tagManage.tpl.html
andrewconner: Yeah

kifi-backend/modules/shoebox/angular/src/tagManage/manageTagService.js
```diff
@@ -0,0 +1,60 @@
...
+          return $q.when(list);
+        }
+        return first;
+      },
+      list: list
+    };
+
+    return api;
+  }
+]);
```
yipingliao: Need newline.

kifi-backend/modules/shoebox/angular/src/tagManage/tagManage.js
```diff
@@ -0,0 +1,124 @@
...
+    $scope.tagsScrollDistance = '100%';
+    $scope.isTagScrollDisabled = function () {
+      return !manageTagService.hasMore();
+    };
+    $scope.manageTagScrollNext = function () {
+      manageTagService.getMore($scope.selectedSort);
+    };
+
+    $scope.$watch(function () {
+        return $scope.selectedSort;
```
yipingliao: This and next line needs to be unindented one stop.

kifi-backend/modules/shoebox/angular/src/tagManage/tagManage.js
```diff
@@ -0,0 +1,124 @@
...
+      }, function () {
+      manageTagService.reset();
```
yipingliao, line 38: These lines need indentation.

kifi-backend/modules/shoebox/angular/src/tagManage/tagManage.js
```diff
@@ -0,0 +1,124 @@
...
+            return lib.access !== 'read_only';
+          });
+          $scope.selection = $scope.selection || {};
+          $scope.selection.library = _.find($scope.libraries, { 'name': 'Main Library' });
+        });
+      }
+    });
+
+    $scope.clickAction = function () {
+      if ($scope.selectedTagName !== '') {
```
yipingliao, line 64: Curious about when will this be an empty string...

kifi-backend/modules/shoebox/angular/src/tagManage/tagManage.js
```diff
@@ -0,0 +1,124 @@
...
+
+      return $scope.tagsToShow;
+    };
+
+
+    //
+    // Manage Tags
+    //
+    $scope.removeTag = function (tag) {
+      var choice = $window.confirm('Are you sure you want to delete '+ tag.name + '?');
```
yipingliao, line 117: :) I use this in deleting libraries and need to upgrade it too. Can you add a TODO here?

yipingliao:
Nice! By the way it's also quite easy to custom position that library selection widget if it's not popping up where you want it on the tags page. I've extended it on my branch and can show you quickly.

kifi-backend/modules/shoebox/angular/src/tagManage/tagManage.tpl.html
yipingliao: It's used like this a lot in the code base. But we could convert it to an attribute instead.

kifi-backend/modules/shoebox/angular/src/tagManage/tagManage.tpl.html
ahsu1230, line 1: woops

andrewconner:
Great job. The state thing can be fixed later if you want — it's a newer pattern that we like better than keeping state in a service (which was the old pattern and caused problems). Either way, minor things.

ahsu1230:
getting this in for demo. Can refactor stuff after

PR #11426, opened 2014-10-10
short term fix for tracking experiments

yingjieMiao:
@stkem  as discussed.


kifi-backend/modules/common/app/com/keepit/heimdal/HeimdalContext.scala
```diff
@@ -135,6 +135,7 @@ class HeimdalContextBuilder {
   def addExperiments(experiments: Set[ExperimentType]): Unit = {
     this += ("experiments", experiments.map(_.value).toSeq)
     this += ("userStatus", ExperimentType.getUserStatus(experiments))
+    experiments.foreach { ex => this += (ex.value, true) }
```
stkem, line 138: This doesn't really cover it:
1) The experiments should be prefixed to avoid collisions with other event properties and to make it clear it's an experiment, i.e. `this += ("exp_" + ex.value, true)`
2) This is not the only place that experiment are added to a context and we don't really need them exploded in out own data storage, but just in mixpanel, so the better place to do this would be when the event is forwarded to mixpanel. That way we can't miss any.

Please make at least the prefix change before this get's deployed, otherwise the unprefixed properties will be in mixpanel quasi forever.

PR #11435, opened 2014-10-10
adding small things to manage tags page

ahsu1230:
refactor + reorg coming later

@andrewconner @yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
@@ -515,6 +515,8 @@ class LibraryCommander @Inject() (
           badKeeps += keep -> LibraryError.AlreadyExistsInDest
         }
       }
+      if (badKeeps.size != keeps.size)
+        libraryRepo.updateLastKept(library.id.get)
     }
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -515,6 +515,8 @@ class LibraryCommander @Inject() (
           badKeeps += keep -> LibraryError.AlreadyExistsInDest
         }
       }
+      if (badKeeps.size != keeps.size)
+        libraryRepo.updateLastKept(library.id.get)
```
ahsu1230, line 519: update Last Kept for copy / move Keeps between Libraries & copy keeps from tag to library

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -515,6 +515,8 @@ class LibraryCommander @Inject() (
           badKeeps += keep -> LibraryError.AlreadyExistsInDest
         }
       }
+      if (badKeeps.size != keeps.size)
```
yipingliao, line 518: Use triple equal.

PR #11437, opened 2014-10-10
Fix BigDecimal comparison

leogrim:
@2is10

### DIFF
--- a/kifi-backend/modules/search/app/com/keepit/controllers/ext/ExtSearchController.scala
+++ b/kifi-backend/modules/search/app/com/keepit/controllers/ext/ExtSearchController.scala
@@ -9,16 +9,13 @@ import com.keepit.controllers.util.SearchControllerUtil
 import com.keepit.model._
 import com.keepit.model.ExperimentType.ADMIN
 import com.keepit.search.graph.library.LibraryIndexer
-import com.keepit.search.{ RestrictedKeepInfo, AugmentationCommander, SearchCommander }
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/search/app/com/keepit/controllers/ext/ExtSearchController.scala
```diff
@@ -9,24 +9,22 @@ import com.keepit.controllers.util.SearchControllerUtil
...
-
 object ExtSearchController {
   private[ExtSearchController] val maxKeepersShown = 20
   private[ExtSearchController] val maxLibrariesShown = 10
   private[ExtSearchController] val maxTagsShown = 15
 
   @inline private def canBeOmitted(value: JsValue): Boolean = value match {
-    case JsNull | JsBoolean(false) | JsString("") | JsArray(Seq()) | JsNumber(BigDecimal.ZERO) => true
+    case JsNull | JsBoolean(false) | JsString("") | JsArray(Seq()) => true
+    case JsNumber(zero) if zero.equals(0) => true
```
2is10, line 27: I believe you need to use `.compareTo(ZERO) == 0`.

```
scala> new java.math.BigDecimal("0.000").equals(0)
res1: Boolean = false

scala> new java.math.BigDecimal("0").equals(0)
res2: Boolean = false

scala> new java.math.BigDecimal(0).equals(0)
res3: Boolean = false
```

PR #11440, opened 2014-10-10
Fix incorrect number of followers problem; also fix rendering problem on following pics

yipingliao:
@andrewconner 

See:
https://app.asana.com/0/15523651812135/17663547464681

Also fixes a problem that @raymondng noticed where we sometimes get strange configuration of follower pics, like pics dropping down (permanently) when the window is small.

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
@@ -40,24 +40,27 @@ angular.module('kifi')
 
           // 250px needed for other stuff in the parent's width.
           var maxFollowersToShow = Math.floor((parentWidth - 250) / widthPerFollowerPic);
-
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
```diff
@@ -40,24 +40,27 @@ angular.module('kifi')
...
-            }
 
-            if (maxFollowersToShow >= scope.library.followers.length) {
+          // If we only have one additional follower that we can't fit in, then we can fit that one
+          // in if we don't show the additional-number-of-followers circle.
+          if (maxFollowersToShow === scope.library.numFollowers - 1) {
+            maxFollowersToShow++;
+          }
+
+          scope.$evalAsync(function () {
```
yipingliao, line 51: Strange rendering problem fixed when we use `$evalAsync`.

PR #11443, opened 2014-10-10
Fix libraries dropping off.

yipingliao:
Small fix for:
https://app.asana.com/0/15523651812135/17473201402543

@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/library.styl
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/library.styl
@@ -31,6 +31,7 @@
 
   .kf-keep-lib-name
     display: inline-block
+    width: 97%
     font-size: 26px
     font-weight: bold
     vertical-align: middle


kifi-backend/modules/shoebox/angular/src/libraries/library.styl
```diff
@@ -31,6 +31,7 @@
 
   .kf-keep-lib-name
     display: inline-block
+    width: 97%
```
yipingliao, line 34: I _think_ it's because I've got another empty inline-block element as a sibling and that's taking up some extra width. Maybe.

andrewconner, line 34: Could you do something like `calc(100% - 5px)`? 97% is flexible

yipingliao, line 34: Will do.

PR #11450, opened 2014-10-11
Website Search API + DRYing 

leogrim:
@2is10 
I think I found the problem, JsNumber takes a scala.math.BigDecimal, not a java.math.BigDecimal
From testing in the REPL, this should work.


kifi-backend/modules/search/app/com/keepit/controllers/ext/ExtSearchController.scala
```diff
@@ -24,7 +24,7 @@ object ExtSearchController {
 
   @inline private def canBeOmitted(value: JsValue): Boolean = value match {
     case JsNull | JsBoolean(false) | JsString("") | JsArray(Seq()) => true
-    case JsNumber(zero) if zero.equals(0) => true
+    case JsNumber(zero) if zero == 0 => true
```
2is10: cool. could this just be `case JsNumber(0) => true`?

leogrim: For some reason the implicit conversion is not applied. Initially I had case JsNumber(BigDecimal.zero) but was using the Java class. I'll make a few more tests in the REPL to see if this can be further simplified.

leogrim:
@2is10 @stkem

2is10:
fine by me

stkem:
:+1:

PR #11453, opened 2014-10-13
being more clear about external id exception and printing out more conte...

eishay:
...xt in recent case of parsing error.


2is10:
FYI: The search result data is arriving in a completely new format in the dev build (for libraries). We’ve made no corresponding change in the JSON format for reporting search events. On Friday afternoon I attempted to adapt the new data to the old search event JSON format in the dev extension code. If the errors are from a dev build prior to 3.2.16.23, then we probably don’t need to worry about them. If they’re still occurring in 3.2.16.23 or are occurring in the production extension, then we still have an issue to address.

PR #11455, opened 2014-10-13
reducing curator concurrency

stkem:
@eishay @yingjieMiao


eishay:
like the reactive lock :-)

PR #11456, opened 2014-10-13
Profile bugs fix

yipingliao:
@andrewconner I found a few bugs in `/profile` while doing modal refresh. They are:
- Cannot change email address from one verified email to another. Based on git blame, this bug seems really old...
- Occasionally, some `scope.$apply` are still giving me `$digest already in progress` errors. I switched to `$evalAsync` for these. That seems to work really well. :)
- There was a bug where re-editing a saved name will re-populate the name with the old name w/o page refresh (though the name was saved correctly to the db).

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/profile/profile.js
+++ b/kifi-backend/modules/shoebox/angular/src/profile/profile.js
@@ -123,7 +123,7 @@ angular.module('kifi')
         return;
       }
       if (emailInfo.isVerified) {
-        return profileService.setNewPrimaryEmail($scope.me, emailInfo.address);
+        return profileService.setNewPrimaryEmail(emailInfo.address);
       }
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/angular/src/profile/profile.js
```diff
@@ -123,7 +123,7 @@ angular.module('kifi')
         return;
       }
       if (emailInfo.isVerified) {
-        return profileService.setNewPrimaryEmail($scope.me, emailInfo.address);
+        return profileService.setNewPrimaryEmail(emailInfo.address);
```
yipingliao, line 126: `setNewPrimaryEmail` expects just one argument, an email, and the `me` object was passed in instead, resulting in a server error. :(

andrewconner:
Thanks!

PR #11457, opened 2014-10-13
simplify

ymatsuda:
@eishay 
- Switched from scala.util.Random to java.util.Random scala.util.Random is just  wrapper, and we are not using anything in it.
- LoggingStatsdClient is created lazily. There is not point creating Random lazily inside LoggingStatsdClient.
- We should use "private[this]" rather than "private" when possible.


eishay:
:+1:

PR #11458, opened 2014-10-13
Modal refresh

yipingliao:
These commits:
- Style add keep to library modal per design
  (http://content.screencast.com/users/Dannyblumenfeld/folders/Snagit/media/5b38d74d-259c-472b-9bd3-65ba108ac810/10.06.2014-17.png)
- Style import bookmarks and import bookmark files to library per design (same as above). Note that the backend support for importing to libraries is still missing
- Updates all modals to be stylistically consistent with the new design
- Note that create/manage modals are minimally "not too ugly"; they will be refreshed per design in the next pull.

Thanks!
Yiping
@andrewconner @ahsu1230


andrewconner:
Excellent!

PR #11461, opened 2014-10-13
retry with a sanitized url

ymatsuda:
@raymondng @eishay 
To mitigate an error like below, this change catches an exception and retry with a sanitized version of URL.

Cause:
java.lang.IllegalArgumentException: Illegal character in path at index 46: http://happysmilebeauty.blog.shinobi.jp/ホームメイド　編み物編/できるようになった編み方特集%20part%201 
java.net.URI.create(URI.java:859)


ymatsuda:
The example is Japanese. However, this is not Japanese specific.

eishay:
like

raymondng:
lgtm! Thanks!

PR #11462, opened 2014-10-13
supporting either ExternalId[Collection] or Hashtag (String) in "tags" field of KifiHitContext

2is10:
@ymatsuda @eishay


2is10:
merging to silence errors, feedback appreciated

PR #11463, opened 2014-10-13
action migration: non-sb controllers

raymondng:
All non-shoebox Controllers migrated to using new `[User]Actions` framework

(Some Controllers on shoebox are more complex; hence deal with the low hanging fruit first)

95%++ mechanical changes (i.e. search-and-replace)

Tests passed!

@andrewconner 

cc: @42eng


raymondng:
Will merge to build upon it. Feedback appreciated.

PR #11467, opened 2014-10-13
temporarily removing links from the recommendation email that break on mobile

stkem:
@joshm1 @eishay


eishay:
Thanks!should deploy today

stkem:
Yup, waiting for the build.

PR #11471, opened 2014-10-14
Refresh create-library modal.

yipingliao:
Manage library will come tomorrow. :)
https://app.asana.com/0/15523651812135/17247218437907

@andrewconner


andrewconner:
Nice!

kifi-backend/modules/shoebox/angular/src/libraries/libraryVisibilityCheckbox.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
```
andrewconner, line 1: Whoa, modular :)

andrewconner:
Looks good to me.

PR #11474, opened 2014-10-14
Possible fix for unicode library path

leogrim:
@andrewconner hope this fixes it. Looks like we could bake the normalization into the type itself?
(like EmailAddress?)


eishay:
will it work with non western chars as well? e.g. japanese

leogrim:
Not doing anything specific to "é" :)

PR #11476, opened 2014-10-14
Possible fix for unicode library path [V2]

leogrim:
@andrewconner merging to test whether the fix works and we can talk about UsernameOps being moved to Username (which is in common)


andrewconner:
Hm, couldn't the repo just do this? I see the reason you did this, but can't we just make sure we normalize before persisting? This doesn't protect you — `copy` is probably more likely to be used than `withUsername`

leogrim:
Yes this is just convenience. I was thinking of having validation in the Format, like we do for email addresses, so that we can fail early. What do you think?

2is10:
When possible, bug fixes should be accompanied by a simple test case (or several) that demonstrate that they work. This practice helps us avoid costly commit-push-merge-wait-deploy-check cycles and helps prevent regressions going forward. Seems like this particular fix could be pretty easily verified in a test.

leogrim:
Fair enough.

andrewconner:
@leogrim Well, having `withUsername` means the `Username` / `UsernameOps` object needs to be in common — the wrong direction. It's actually a large class because of the word lists, so was in shoebox on purpose.

leogrim:
Yeah but withUsername was only for convenience. Having UsernameOps available with Username makes it possible to validate usernames within Formatters. Anyway I have reverted the whole thing, it doesn't fix anything anyway.

andrewconner:
Interesting, weird. The 404 is coming from AngularRouter. Weird it doesn't think "léo" is a valid username.

PR #11478, opened 2014-10-14
using different cache for the "global keep count" and using it (was not ...

eishay:
...used before). also extracting the count fetch logic to a a controller so we can reuse it for the new counter on the site
@andrewconner @yingjieMiao


andrewconner:
Looks good.

yingjieMiao:
lgtm

PR #11479, opened 2014-10-14
Public ids threadsafe fix

andrewconner:
@2is10 Guava / Scala has this weird bug when storing `java.util.Long`s (Scala tries to optimize it and unbox them, but that blows up the cache at runtime). Without the cache seems fast enough, but I'll keep an eye on it.

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/common/crypto/ModelWithPublicId.scala
+++ b/kifi-backend/modules/common/app/com/keepit/common/crypto/ModelWithPublicId.scala
@@ -63,25 +63,13 @@ trait ModelWithPublicIdCompanion[T <: ModelWithPublicId[T]] {
   */
   protected[this] val publicIdIvSpec: IvParameterSpec
 
-  //private val decryptCache: Cache[java.lang.Long, Try[java.lang.Long]] = CacheBuilder.newBuilder().concurrencyLevel(1).initialCapacity(1024).maximumSize(4096).build()
   private def threadSafeDecrypt(id: Long, config: PublicIdConfiguration): Try[Long] = {
-    //    decryptCache.get(id, new Callable[Try[java.lang.Long]] {
-    //      def call() = synchronized {
-    //        Try(config.aes64bit(publicIdIvSpec).decrypt(id))
-    //      }
-    //    }).map(Long2long)
     synchronized {
       Try(config.aes64bit(publicIdIvSpec).decrypt(id))
     }
   }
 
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/common/app/com/keepit/common/crypto/ModelWithPublicId.scala
```diff
@@ -58,9 +59,21 @@ trait ModelWithPublicIdCompanion[T <: ModelWithPublicId[T]] {
   */
   protected[this] val publicIdIvSpec: IvParameterSpec
 
+  private def threadSafeDecrypt(id: Long, config: PublicIdConfiguration): Try[Long] = {
+    synchronized {
```
ymatsuda: Which part is thread-unsafe? If it is Cipher, why don't you synchronized just on cipher.doFinal in Aes64BitCipher?

kifi-backend/modules/common/app/com/keepit/common/crypto/ModelWithPublicId.scala
```diff
@@ -58,9 +59,21 @@ trait ModelWithPublicIdCompanion[T <: ModelWithPublicId[T]] {
...
+      Try(config.aes64bit(publicIdIvSpec).decrypt(id))
+    }
```
2is10: This looks correct, but synchronizing on the `ModelWithPublicIdCompanion` relies on knowledge of how `PublicIdConfiguration` manages `Aes64BitCipher` instance lifetimes.

We could localize that knowledge and the solution to the thread-safety concern better by making `PublicIdConfiguration` handle both creation/reuse _and_ synchronization. The `PublicIdConfiguration` API might then look like:

``` scala
config.encode(publicIdIvSpec, id)  // or .encrypt
config.decode(publicIdIvSpec, id)  // or .decrypt
```

kifi-backend/modules/common/app/com/keepit/common/crypto/ModelWithPublicId.scala
andrewconner: True, thanks

kifi-backend/modules/common/app/com/keepit/common/crypto/ModelWithPublicId.scala
andrewconner: That's better, will do

andrewconner:
@ymatsuda Simplified based on your feedback. If this is a bottleneck later, I can complicate it to make it faster

2is10:
:+1:

PR #11488, opened 2014-10-14
Fixing FB invites when the script was blocked from the page

andrewconner:
@yipingliao Old bug — if facebook's script was blocked from loading, we ended up in a weird state where users wouldn't get good error messages. This should fix it.


kifi-backend/modules/shoebox/angular/src/common/services/angular-facebook-api.js
```diff
@@ -123,6 +124,18 @@ angular.module('jun.facebook', [])
 				});
 			};
 
+			var failedToLoadHander = function () {
```
yipingliao, line 127: `s/failedToLoadHander/failedToLoadHandler` :)

PR #11492, opened 2014-10-14
action migration: simple sb controllers

raymondng:
On to shoebox. Focusing on the simpler controllers where mostly mechanical changes would do the trick.

Interestingly there are a few places where the tests do not set user at all and it still passed with old `FakeActionAuthenticator`. I suspect that's not a feature. In any case -- added setUser where needed.

Tests passed!

@andrewconner


raymondng:
Merging to build upon it. Feedback appreciated.

andrewconner:
I think we default to userId 1 in tests if it's not provided. We probably should remove it to force testers to set the username (and/or, be able to check no user set)

raymondng:
Agreed. The new `FakeUserActions` module requires explicit call to `setUser`. That results in some test breakage (and require fixing) but IMO it's worth it.

PR #11493, opened 2014-10-14
type ahead endpoint for searching user tags

ahsu1230:
Thanks @leogrim 

@andrewconner


ahsu1230:
merging to use on front-end side

PR #11500, opened 2014-10-14
real tag searching (type ahead) for managing tags

ahsu1230:
Stop using Fuse search & use type ahead searching instead
Still keep sort order & if filter is cleared, return back to regular tag pagination in sorted order.
@andrewconner @yipingliao 
- 2 bug fixes
  - Library Invitation Notification added to "triggered" event
  - Library sidebar ordering gets jumbled up sometimes when library list is updated (used to be database ordering). Now re-sorts after update.


kifi-backend/modules/common/app/com/keepit/model/NotificationCategory.scala
```diff
@@ -54,7 +54,7 @@ object NotificationCategory {
     val aboutFriends = Set(CONTACT_JOINED, CONNECTION_MADE, SOCIAL_FRIEND_JOINED)
 
     // Formatting Categories used in the extension
-    val triggered = Set(FRIEND_ACCEPTED, FRIEND_REQUEST, WHO_KEPT_MY_KEEP, CONTACT_JOINED, CONNECTION_MADE, SOCIAL_FRIEND_JOINED)
+    val triggered = Set(FRIEND_ACCEPTED, FRIEND_REQUEST, WHO_KEPT_MY_KEEP, CONTACT_JOINED, CONNECTION_MADE, SOCIAL_FRIEND_JOINED, LIBRARY_INVITATION)
```
ahsu1230, line 57: @2is10 @stkem quick fix!

kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
```diff
@@ -27,14 +27,13 @@ angular.module('kifi')
         scope.secretLib = {};
         scope.userLibsToShow = [];
         scope.invitedLibsToShow = [];
-        scope.sortingMenu = { show : false, option : '' };
+        scope.sortingMenu = { show : false, option : 'last_kept' };
```
yipingliao, line 30: No space before `:`. Here and elsewhere in the file.

kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
```diff
@@ -52,9 +51,9 @@ angular.module('kifi')
           scope.secretLib = _.find(libraryService.librarySummaries, { 'kind' : 'system_secret' });
           allUserLibs = _.filter(libraryService.librarySummaries, { 'kind' : 'user_created' });
 
-          scope.userLibsToShow = allUserLibs;
-          scope.invitedLibsToShow = libraryService.invitedSummaries;
-          scope.sortingMenu.option = 'last_kept';
+          var newList = sortLibraries(allUserLibs, libraryService.invitedSummaries);
```
yipingliao, line 54: Maybe a name like `sortedLibraries` would be more descriptive than `newList`?

kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
```diff
@@ -190,7 +189,18 @@ angular.module('kifi')
...
+          newLibs = newList[0];
+          newInvited = newList[1];
+
+          scope.userLibsToShow = newLibs;
+          scope.invitedLibsToShow = newInvited;
+          return scope.userLibsToShow.concat(scope.invitedLibsToShow);
+        };
 
+        function sortLibraries(userLibs, invitedLibs) {
+          var newLibs = userLibs;
```
yipingliao, line 202: Is this assignment necessary? In the switch statement below, if all `newLibs` and `newInvited` on the RHS are replaced with `userLibs` and `invitedLibs`, that should work as well.

kifi-backend/modules/shoebox/angular/src/tagManage/tagManage.js
```diff
@@ -35,11 +37,21 @@ angular.module('kifi')
...
-      manageTagService.reset();
-      getPage();
+      if ($scope.filter.name === '') {
+        $scope.tagList.length = 0;
+        $scope.offset = 0;
+        $scope.more = true;
+        manageTagService.reset();
+        getPage();
+      } else {
+        var sortedTags = $scope.tagsToShow;
```
yipingliao, line 47: This can be DRYed up with the code starting at line 136 below.

PR #11506, opened 2014-10-14
Refresh manage library modal and make delete library confirmation modal.

yipingliao:
@andrewconner 
Asana: https://app.asana.com/0/15523651812135/17247218437907
Design: http://content.screencast.com/users/Dannyblumenfeld/folders/Snagit/media/5b38d74d-259c-472b-9bd3-65ba108ac810/10.06.2014-17.png (towards bottom)

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/common/services/modalService.js
+++ b/kifi-backend/modules/shoebox/angular/src/common/services/modalService.js
@@ -4,6 +4,8 @@ angular.module('kifi')
 
 .factory('modalService', ['$compile', '$rootScope', '$templateCache',
   function ($compile, $rootScope, $templateCache) {
+    var modals = [];
+
     function open (opts) {
       opts = opts || {};
 
@@ -20,10 +22,11 @@ angular.module('kifi')
 
       $compile($modal)(scope);
       angular.element(document.body).append($modal);
+      modals.push($modal);
     }
 
     function close () {
-      var $modal = angular.element(document.getElementById('kf-modal'));
+      var $modal = modals.pop();
       $modal.scope().$destroy();
       $modal.remove();
     }
--- a/kifi-backend/modules/shoebox/angular/src/common/services/modalService.js
+++ b/kifi-backend/modules/shoebox/angular/src/common/services/modalService.js
@@ -15,13 +15,14 @@ angular.module('kifi')
       }
       var $modal = angular.element($templateCache.get(template));
 
-      var scope = opts.scope || $rootScope.$new();
+      var scope = (opts.scope && opts.scope.$new()) || $rootScope.$new();
       if (opts.modalData) {
         scope.modalData = opts.modalData;
       }
 
       $compile($modal)(scope);
-      angular.element(document.body).append($modal);
+      angular.element(document.body).find('.kf-cols').append($modal);
+
       modals.push($modal);
     }
 
--- a/kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.js
@@ -3,8 +3,8 @@
 angular.module('kifi')
 
 .directive('kfManageLibrary', [
-  '$location', '$window', '$rootScope', 'friendService', 'libraryService', 'profileService', 'util',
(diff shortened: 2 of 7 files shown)


kifi-backend/modules/shoebox/angular/src/common/services/modalService.js
```diff
@@ -4,6 +4,8 @@ angular.module('kifi')
 
 .factory('modalService', ['$compile', '$rootScope', '$templateCache',
   function ($compile, $rootScope, $templateCache) {
+    var modals = [];
```
yipingliao, line 7: Simple extension to include multiple modals. `modals` is a stack so that the last opened modal is closed first.

kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.js
```diff
@@ -92,28 +93,39 @@ angular.module('kifi')
...
+            submitting = false;
+            $rootScope.$emit('changedLibrary');
+            scope.close();
+            $location.path('/');
+          });
+        };
 
-          if (confirmDelete) {
-            submitting = true;
+        scope.showFollowersPanel = function () {
```
yipingliao, line 123: The "followers" panel is the second screen of the manage library modal. It is shown only when the user clicks on a CTA button on the first screen.

kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.styl
```diff
@@ -226,5 +305,34 @@ manage-lib-button =
...
+.manage-lib-content
+  position: relative
+  height: 398px
+  overflow: hidden
+
+.manage-lib-default-content
+  position: absolute
+  left: 15px
+  background-color: #fff
+  transition: margin ease 0.5s
```
yipingliao, line 317: The first screen slides away to the left to show the second screen. :)

kifi-backend/modules/shoebox/angular/src/common/services/modalService.js
```diff
@@ -20,10 +22,11 @@ angular.module('kifi')
 
       $compile($modal)(scope);
       angular.element(document.body).append($modal);
```
yipingliao: Related to our discussion re `rootScope`, we may want to append the `$modal` like this instead:
`angular.element(document.body).find('.kf-cols').append($modal);`

andrewconner: Interesting, we can if you want.

kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.styl
andrewconner, line 317: Nice!

kifi-backend/modules/shoebox/angular/src/common/services/modalService.js
yipingliao: This would be in lieu of the storing a reference pointer solution you proposed earlier. If the problem is that an element gets `$rootScope` when appended to `body`, then it's not clear whether destroying the reference pointer to the original scope cleanly cleans up the modal's "scope".

andrewconner: Ah I see. I'm fine either way, your call. I imagine the parent scope being destroyed does _not_ destroy the modal scope (just based on what I'm seeing in the console). However, I think it's pretty rare for a controller to spawn a modal, get destroyed, and the modal stay open for a long time. Thoughts?

kifi-backend/modules/shoebox/angular/src/common/services/modalService.js
```diff
@@ -13,17 +15,19 @@ angular.module('kifi')
       }
       var $modal = angular.element($templateCache.get(template));
 
-      var scope = opts.scope || $rootScope.$new();
+      var scope = (opts.scope && opts.scope.$new()) || $rootScope.$new();
```
yipingliao, line 18: @andrewconner I added this fix that you showed me earlier, and below I am appending to `.kf-cols`. What do you think? (not sure if I entirely understood your last comment; let me know your concerns? thanks!)

yipingliao:
Merging this; will address any additional comments when they come in.

PR #11508, opened 2014-10-15
Wire up front end to import bookmark file to library endpoint.

yipingliao:
@andrewconner This is just for importing files; still awaiting the endpoint for importing browser bookmarks.
https://app.asana.com/0/15523651812135/17247218437903

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/common/modal/importBookmarkFileLibraryModal.tpl.html
+++ b/kifi-backend/modules/shoebox/angular/src/common/modal/importBookmarkFileLibraryModal.tpl.html
@@ -38,7 +38,6 @@
           <td class="import-library-how-to-row-heading">Step 3</td>
           <td class="import-library-how-to-row-content">Choose a library to import to</td>
         </tr>
-
       </table>
     </div>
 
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/angular/src/common/modal/importBookmarkFileLibraryModal.tpl.html
```diff
@@ -38,7 +38,6 @@
           <td class="import-library-how-to-row-heading">Step 3</td>
           <td class="import-library-how-to-row-content">Choose a library to import to</td>
         </tr>
-
```
yipingliao, line 41: Just clean up in this file.

kifi-backend/modules/shoebox/angular/src/layout/main/MainCtrl.js
```diff
@@ -263,6 +263,31 @@ angular.module('kifi')
       return deferred.promise;
     }
 
+    function uploadBookmarkFileToLibraryHelper(file, libraryId) {
```
yipingliao, line 266: 99% of the code added to this file is just copying what was there for non-libraries. It's not DRY but we're going to rip out the non-libraries stuff in a few more weeks anyway.

andrewconner:
Importing browser bookmarks: let's talk with Jared tomorrow about it. Most of the work happens in the extension itself, so we need to make sure the extension can deal with it. @2is10 (context: the site will let you pick a library to import your bookmarks to. extension needs to appropriately pass this to the background page / server).

Should be pretty simple

PR #11510, opened 2014-10-15
3.2.18: extension support for importing browser bookmarks to a library

2is10:
@andrewconner 
cc @yipingliao 

The site will need to pass `{type: 'import_bookmarks', libraryId: 'lxxxxxxxxxx'}` to `window.postMessage`.

The changes I’ve made here should be backwards compatible.

### DIFF
--- a/packrat/scripts/installed.js
+++ b/packrat/scripts/installed.js
@@ -39,10 +39,10 @@
         api.port.emit('import_bookmarks_declined');
         break;
       case 'import_bookmarks':
-        api.port.emit('import_bookmarks');
+        api.port.emit('import_bookmarks', data.libraryId);
         break;
-      case 'import_bookmarks_public':
-        api.port.emit('import_bookmarks_public');
+      case 'import_bookmarks_public':  // deprecated
+        api.port.emit('import_bookmarks', 'main');
         break;
       case 'close_tab':
         api.port.emit('close_tab');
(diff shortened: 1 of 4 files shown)


packrat/scripts/installed.js
```diff
@@ -39,10 +39,10 @@
         api.port.emit('import_bookmarks_declined');
         break;
       case 'import_bookmarks':
-        api.port.emit('import_bookmarks');
+        api.port.emit('import_bookmarks', data.libraryId);
         break;
-      case 'import_bookmarks_public':
-        api.port.emit('import_bookmarks_public');
+      case 'import_bookmarks_public':  // deprecated
+        api.port.emit('import_bookmarks', 'main');
```
yipingliao, line 45: For non-libraries-enabled users, we are currently still presenting them the option of importing publicly or privately. In the latter case we should be importing into `'secret'`. Should we also special-case that? Alternatively, we can remove that option in the import UI.

andrewconner:
Wow, you're awesome. I'll add the endpoint.

packrat/scripts/installed.js
2is10, line 45: This extension code will work properly if the site continues to behave as it does today for non-library enabled users. Today it passes `.postMessage` either `'import_bookmarks'` or `'import_bookmarks_public'`. It doesn’t pass a JSON object.

The site should continue to behave this way for a while for non-library users so as to avoid breaking if users don’t have the latest extension.

PR #11511, opened 2014-10-15
Fixing multi-word tags, making them real links.

andrewconner:
@ahsu1230 When clicking tags with a space, it wasn't properly quoting them. So "Fun plans" was turning into `tag:Fun plans`, which means something else. This should fix it.

While I was at it, I made it a real link (vs a click handler).

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/tagManage/manageTagService.js
+++ b/kifi-backend/modules/shoebox/angular/src/tagManage/manageTagService.js
@@ -8,10 +8,24 @@ angular.module('kifi')
     var pageSize = 100;
     var searchLimit = 30;
 
+    var decorate = function (tags) {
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/angular/src/tagManage/manageTagService.js
```diff
@@ -8,10 +8,24 @@ angular.module('kifi')
     var pageSize = 100;
     var searchLimit = 30;
 
+    var decorate = function (tags) {
+      return _.map(tags || [], function (t) {
+        var tagPath;
+        var n = t.name;
+        if (n.indexOf(' ') !== -1) {
+          tagPath = '"' + n + '"';
```
ahsu1230, line 16: I see gotcha :+1: Thanks for this!

PR #11513, opened 2014-10-15
adding library mobile endpoints - no tests yet

eishay:
@thassss @jaredpetker @andrewconner 
The endpoints are there, there are no tests yet (working on it) but they can start to get used now.
Please let me know if there are any non mobile endpoints that mobile is using


jaredpetker:
looks like that's what is currently all being used.

PR #11515, opened 2014-10-15
Add scrolling to up/down on library share search.

yipingliao:
@andrewconner 

https://app.asana.com/0/17575799252940/17469187460768


kifi-backend/modules/shoebox/angular/src/libraries/libraryShareSearch.js
```diff
@@ -181,15 +185,58 @@ angular.module('kifi')
...
-          // For dev testing.
-          // scope.library.followers.push({ id: 'dc6cb121-2a69-47c7-898b-bc2b9356054c' });
-
           return _.some(scope.library.followers, function (follower) {
             return follower.id === user.id;
           });
         }
 
 
+        // TODO(yiping): make a directive for displaying a list of items where up and down
```
yipingliao, line 194: This will be fun with `ng-transclude`. Something for post-launch.

yipingliao:
Merging this; doesn't seem to be too controversial.

PR #11516, opened 2014-10-15
clean up in UriIntergrity Plugin

yingjieMiao:
@ymatsuda @andrewconner @raymondng 

THIS PR SHOULD HAVE NO LOGIC CHANGE. Please review carefully.

In the old code, we have a method

```
def save(duplicate: Keep, primary: Keep, libraryId: Option[Id[Library]], visibility: LibraryVisibility)
```

where `libraryId` and `visibility` are only used as

```
primary.copy(visibility = visibility, libraryId = libraryId, uriId = newUriId, isPrimary = true, state = KeepStates.ACTIVE)
```

However, all call to the `save()` method are of form:

```
save(duplicate = oldBm, primary = currentPrimary, libraryId = currentPrimary.libraryId, visibility = currentPrimary.visibility)
save(duplicate = currentPrimary, primary = oldBm, libraryId = oldBm.libraryId, visibility = oldBm.visibility)
```

where `libraryId` and `visibility` are from the parameter variable `primary`. Thus, these are redundant and a bit confusing. 

I removed those two paramters from the `save()` method.

When migrate keeps between system_main and system_secret, we have a more subtle issue (not addressed in this PR):
In the old code, we have a comment "// if one of them is private, make the surviving keep private to true to be safe". This logic is not covered anymore. Translating this logic to the library jargon: If one of the keep is in system_secret library, make the surviving keep has visibility `secret`. But the surviving keep's visibility is tied to its library's visibility. Additional constraint!

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/integrity/UriIntegrityPlugin.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/integrity/UriIntegrityPlugin.scala
@@ -78,36 +78,41 @@ class UriIntegrityActor @Inject() (
             (Some(oldBm), None)
           }
           case Some(currentPrimary) => {
-            def save(duplicate: Keep, primary: Keep, libraryId: Option[Id[Library]], visibility: LibraryVisibility): (Option[Keep], Option[Keep]) = {
+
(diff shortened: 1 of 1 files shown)


ymatsuda:
:+1:  lgtm

PR #11517, opened 2014-10-15
Remove LibraryHashtagTypeahead, use UserHashtagTypeahead instead

leogrim:
@andrewconner @2is10 the library specific API doesn't make sense anymore, but I haven't touched it so that the change is transparent for you. We can change it / make it support tag suggestions later on.


kifi-backend/modules/shoebox/test/com/keepit/controllers/ext/ExtLibraryControllerTest.scala
```diff
@@ -441,9 +441,11 @@ class ExtLibraryControllerTest extends Specification with ShoeboxTestInjector wi
         contentAsString(searchTags(user1, libPubId, "a", 2)) === """[{"tag":"aardvark","matches":[[0,1]]},{"tag":"animal","matches":[[0,1]]}]"""
         contentAsString(searchTags(user1, libPubId, "s", 2)) === """[]"""
 
+        /* todo(Léo): reconsider when tags have been figured out from a product perspective
         // other user with read access to library can search tags
         contentAsString(searchTags(user2, libPubId, "a", 3)) === """[{"tag":"aardvark","matches":[[0,1]]},{"tag":"animal","matches":[[0,1]]},{"tag":"Awesome","matches":[[0,1]]}]"""
         contentAsString(searchTags(user2, libPubId, "s", 3)) === """[]"""
+        */
```
leogrim, line 448: @andrewconner we don't support this, we'll have to figure it out when we do collaborative libraries.

2is10:
:+1:

PR #11519, opened 2014-10-15
bug fixes for libraries side bar & manage tags

ahsu1230:
@andrewconner @yipingliao 

Fix the following bugs:
- [BE] fix inconsistency between keeps count when sorting. Using a new repo retrieval
- sidebar input text color is dark gray
- unusual tag names (#) prevented tags from being kept to library (and clickable)
- change up dialogue box for copying keeps from tags


yipingliao:
:+1:

kifi-backend/modules/shoebox/angular/src/tagManage/tagToLibMsg.tpl.html
```diff
@@ -1,13 +1,15 @@
 <div class="dialog-content">
-  <a class="dialog-x sprite-close-x-icon" ng-click="hide()"></a>
+  <a class="dialog-x" ng-click="hide()"></a>
   <div class="dialog-body-wrap">
     <div class="dialog-body dialog-centered">
-	    <div>We're copying your tagged keeps over, but it might take a second...</div>
-	    <div>Feel free to continue managing your tags or head straight to the library!</div>
+    	<h1 class="dialog-header ng-binding">Copying Keeps</h1><br>
```
andrewconner: `ng-binding`?

PR #11522, opened 2014-10-15
auth controllers!

raymondng:
Almost there. Only did the requisite to migrate things over. Will need a bit more work later on (including sending extra info back, etc.)

No attempt to do any major re-write at this point. I tried to preserve original code/structure as much as possible (to minimize risks/bugs). Nonetheless I think the result is simpler that's easier to understand and reason about.

`MaybeUserAction` is pretty much now the new base `Action` -- it's probably a bit too heavy for that. Should revisit it later for simpler/lighter code (e.g. `CORSAction` as suggested, something along those lines)

Yes tests passed. Doing some ad hoc testing as well.

@andrewconner please take a look.

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/common/controller/ActionAuthenticator.scala
+++ b/kifi-backend/modules/common/app/com/keepit/common/controller/ActionAuthenticator.scala
@@ -32,22 +32,6 @@ case class AuthenticatedRequest[T](
 
 object ActionAuthenticator {
   val FORTYTWO_USER_ID = "fortytwo_user_id"
-
-  implicit class MaybeAuthenticatedRequest(val request: Request[_]) extends AnyVal {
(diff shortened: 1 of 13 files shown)


andrewconner:
Looks good, my eyeball compiler didn't see anything bad :)

raymondng:
Awesome. That's good enough for me :)

andrewconner:
Feel free to merge when ready. I can help test everything.

raymondng:
Thanks. Will do.

raymondng:
Looks like I am not done. There are some indirect dependencies on identityOpt that's hidden by `MaybeAuthenticatedRequest`. I was thinking I'd remove code after migration -- will need to remove this one now.

raymondng:
AFAIK issue fixed. Will merge and push.
Please let me know if you spot any issues.

PR #11523, opened 2014-10-15
library invites abuse monitor

ahsu1230:
Sorry @andrewconner, I was going to submit this much earlier before
but we were testing & thus spamming library invites at each other. Which would eventually stop us from continue testing.

Now that we've tested a good amount, this should be here. Sorry about the delay!


kifi-backend/modules/shoebox/app/com/keepit/shoebox/AbuseControlModule.scala
```diff
@@ -22,5 +22,16 @@ case class AbuseControlModule() extends ScalaModule {
       db = db,
       airbrake = airbrake)
 
+  @Provides
+  @Singleton
+  def libraryInvitesAbuseMonitor(libraryInviteRepo: LibraryInviteRepo, db: Database, airbrake: AirbrakeNotifier): LibraryInvitesAbuseMonitor =
+    new LibraryInvitesAbuseMonitor(
+      absoluteWarn = 20,
```
andrewconner, line 29: Haha, based on recent usage, I bet we'll have a few people trip this. We'll see.

andrewconner:
Looks good.

PR #11524, opened 2014-10-15
New tag search route also suggesting tags

leogrim:
@2is10 @andrewconner


kifi-backend/modules/shoebox/conf/shoebox.routes
```diff
@@ -138,10 +138,11 @@ POST    /ext/libraries/:id/keeps/:k @com.keepit.controllers.ext.ExtLibraryContro
 DELETE  /ext/libraries/:id/keeps/:k @com.keepit.controllers.ext.ExtLibraryController.removeKeep(id: PublicId[Library], k: ExternalId[Keep])
 POST    /ext/libraries/:id/keeps/:k/tags/:tag    @com.keepit.controllers.ext.ExtLibraryController.tagKeep(id: PublicId[Library], k: ExternalId[Keep], tag: String)
 DELETE  /ext/libraries/:id/keeps/:k/tags/:tag    @com.keepit.controllers.ext.ExtLibraryController.untagKeep(id: PublicId[Library], k: ExternalId[Keep], tag: String)
+GET     /ext/libraries/:id/keeps/:k/tags/search  @com.keepit.controllers.ext.ExtLibraryController.searchTags(id: PublicId[Library], k: ExternalId[Keep],  q: String, n: Option[Int] ?= None)
```
2is10, line 141: I may change this to `.../suggest`. As is, it could easily be misinterpreted as “search a keep’s tags”.

2is10:
:+1:

kifi-backend/modules/shoebox/conf/shoebox.routes
leogrim, line 141: Sure!

PR #11525, opened 2014-10-15
keep left library pane consistent at all times

ahsu1230:
sometimes, the left pane bar doesn't keep its library sorting correctly
and that's because on a certain signal `changedLibrary` it keeps the database sorting.

@yipingliao is going to re-org some of this stuff so will leave this small change here for now. 
@andrewconner


yipingliao:
LGTM. Thanks!

PR #11532, opened 2014-10-15
Call extension code to import browser bookmarks for importing to library

yipingliao:
@andrewconner 

Thank you @2is10 for the extension support and explanation.

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/layout/main/MainCtrl.js
+++ b/kifi-backend/modules/shoebox/angular/src/layout/main/MainCtrl.js
@@ -158,7 +158,7 @@ angular.module('kifi')
       // Use $evalAsync to wait for forceClose to close the currently open modal before opening
       // the next modal.
       $scope.$evalAsync(function () {
-        var kifiVersion = $window.document.getElementsByTagName('html')[0].getAttribute('data-kifi-ext');
+        var kifiVersion = $window.document.documentElement.getAttribute('data-kifi-ext');
 
         if (!kifiVersion) {
           modalService.open({
@@ -167,19 +167,20 @@ angular.module('kifi')
           return;
         }
 
-        $analytics.eventTrack('user_clicked_page', {
-          'type': 'browserImport',
-          'action': 'ImportToLibrary'  // TODO: update this when we have the full tracking spec.
-        });
+        // TODO(yiping): update this when we have the full tracking spec.
+        // $analytics.eventTrack('user_clicked_page', {
+        //   'type': 'browserImport',
+        //   'action': '???'
+        // });
 
-        var event = $scope.msgEvent && $scope.msgEvent.origin && $scope.msgEvent.source && $scope.msgEvent;
+        var event = $scope.msgEvent;
         var message = {
           type: 'import_bookmarks',
           libraryId: library.id
         };
 
         if (event) {
-          event.source.postMessage(message, $scope.msgEvent.origin);
+          event.source.postMessage(message, event.origin);
         } else {
           $window.postMessage(message, '*');
         }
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/angular/src/layout/main/MainCtrl.js
```diff
@@ -158,17 +158,35 @@ angular.module('kifi')
...
+        if (!kifiVersion) {
           modalService.open({
             template: 'common/modal/importBookmarksErrorModal.tpl.html'
           });
+          return;
+        }
+
+        $analytics.eventTrack('user_clicked_page', {
+          'type': 'browserImport',
+          'action': 'ImportToLibrary'  // TODO: update this when we have the full tracking spec.
```
2is10: We need to be careful about sending event data to mixpanel that hasn’t been finalized/approved by product (Jen). Any strings we use even once will forever appear in dropdowns and (sometimes) in charts, which can be really annoying for people who use our mixpanel account a lot.

kifi-backend/modules/shoebox/angular/src/layout/main/MainCtrl.js
```diff
@@ -158,17 +158,35 @@ angular.module('kifi')
...
+        });
+
+        var event = $scope.msgEvent && $scope.msgEvent.origin && $scope.msgEvent.source && $scope.msgEvent;
+        var message = {
+          type: 'import_bookmarks',
+          libraryId: library.id
+        };
+
+        if (event) {
+          event.source.postMessage(message, $scope.msgEvent.origin);
```
2is10: `$scope.msgEvent` → `event`

kifi-backend/modules/shoebox/angular/src/layout/main/MainCtrl.js
```diff
@@ -158,17 +158,35 @@ angular.module('kifi')
...
           });
+          return;
+        }
+
+        $analytics.eventTrack('user_clicked_page', {
+          'type': 'browserImport',
+          'action': 'ImportToLibrary'  // TODO: update this when we have the full tracking spec.
+        });
+
+        var event = $scope.msgEvent && $scope.msgEvent.origin && $scope.msgEvent.source && $scope.msgEvent;
```
2is10: The sanity checking doesn’t appear to be necessary. `$scope.msgEvent` should be falsy if not a valid message event.

kifi-backend/modules/shoebox/angular/src/layout/main/MainCtrl.js
```diff
@@ -158,17 +158,35 @@ angular.module('kifi')
...
       // the next modal.
       $scope.$evalAsync(function () {
-        // Fake check. To do: use real endpoints to determine which modal to open.
-        if (library) {
-          modalService.open({
-            template: 'common/modal/importBookmarksLibraryInProgressModal.tpl.html',
-            scope: $scope
-          });
-        } else {
+        var kifiVersion = $window.document.getElementsByTagName('html')[0].getAttribute('data-kifi-ext');
```
2is10: `.documentElement` is short for `.getElementsByTagName('html')[0]`

2is10:
noted a few minor style suggestions... looks good!

kifi-backend/modules/shoebox/angular/src/layout/main/MainCtrl.js
yipingliao: Ugh, that's true. I'll comment out this whole section.

kifi-backend/modules/shoebox/angular/src/layout/main/MainCtrl.js
yipingliao: Will update. Thanks!

kifi-backend/modules/shoebox/angular/src/layout/main/MainCtrl.js
yipingliao: Will update. Thanks!

kifi-backend/modules/shoebox/angular/src/layout/main/MainCtrl.js
yipingliao: Just to make sure: a simple `var event = $scope.msgEvent;` should do?

2is10: Yes.​

PR #11533, opened 2014-10-15
delete tag modal

ahsu1230:
@andrewconner @yipingliao 
looks very similar to the delete library modal popup

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/tagManage/deleteTag.js
+++ b/kifi-backend/modules/shoebox/angular/src/tagManage/deleteTag.js
@@ -1,0 +1,22 @@
+'use strict';
+
+angular.module('kifi')
+
+.directive('kfDeleteTag', ['$location', 'tagService',
+  function ($location, tagService) {
+    return {
+      restrict: 'A',
+      require: '^kfModal',
+      templateUrl: 'tagManage/deleteTagMsg.tpl.html',
+      link: function (scope, element, attrs, kfModalCtrl) {
+
+        scope.deleteTag = function () {
+          tagService.remove(scope.modalData.tag);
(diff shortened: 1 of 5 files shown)


andrewconner:
Thanks

kifi-backend/modules/shoebox/angular/src/tagManage/deleteTag.js
```diff
@@ -0,0 +1,22 @@
...
+      restrict: 'A',
+      require: '^kfModal',
+      templateUrl: 'tagManage/deleteTagMsg.tpl.html',
+      link: function (scope, element, attrs, kfModalCtrl) {
+
+        scope.deleteTag = function () {
+          tagService.remove(scope.modalData.tag);
+          _.remove(scope.modalData.tagsToShow, function(t) { return t === scope.modalData.tag; });
+          kfModalCtrl.close();
+          return scope.modalData.tagsToShow;
```
yipingliao, line 17: Why does this need its own directive? And who is looking at this returned data?

kifi-backend/modules/shoebox/angular/src/tagManage/deleteTag.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
+
+angular.module('kifi')
+
+.directive('kfDeleteTag', ['$location', 'tagService',
```
yipingliao, line 5: `$location` is not being used.

kifi-backend/modules/shoebox/angular/src/tagManage/deleteTagMsg.tpl.html
```diff
@@ -0,0 +1,5 @@
+<div kf-modal class="kf-delete-tag-confirm-modal">
```
yipingliao, line 1: You have two `kf-modal`s nested now.

PR #11534, opened 2014-10-15
Better suggestTags API

leogrim:
@2is10 per your request + making the returned json consistent in both scenarios, makes sense?


2is10:
:+1:

leogrim:
@stkem let me know when you need a similar API for the website (keep cards)

PR #11536, opened 2014-10-15
library invite email updates

joshm1:
- removes bottom image
- removes about 3-4 layers of nested tables
- replace image-based rounded corners with CSS border
- misc spacing tweaks
- removed unused css classes
- removed duplicate copy-pasted css in the <head> and formatted css to be readable
- content width is 100% in mobile w/ some padding on the side

@stkem


stkem:
:+1:

PR #11541, opened 2014-10-16
Add spacing between delighted survey and library header card.

yipingliao:
Asana: https://app.asana.com/0/15523651812135/17379239993717

If the Delighted survey is visible, there needs to be a margin between it and the library header card. If not, there is no margin.

There was a bit of clean-up here in the code; that's why it looks like so many lines.

Thanks!
@andrewconner


yipingliao:
Merging this, since it's a small style change. I've double-checked for non-libraries-enabled users and made sure that this does not change their experience.

PR #11542, opened 2014-10-16
Add restful endpoint to add recipients to a message thread and allow injectable execution context...

eishay:
... that we can wait on when testing code that spawns threads/futures
Thanks to @ymatsuda for guideline and help with the execution and @stkem for the idea.

This should mark the end of `Thread.sleep` in test!

@2is10 we may want to use restful endpoint for extension as well.
/cc @leogrim @raymondng @andrewconner @joshm1 @ahsu1230 please check out use of 

``` scala
inject[WatchableExecutionContext].drain()
```

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/shoebox/ShoeboxServiceClient.scala
+++ b/kifi-backend/modules/common/app/com/keepit/shoebox/ShoeboxServiceClient.scala
@@ -1,12 +1,12 @@
 package com.keepit.shoebox
 
+import com.keepit.common.concurrent.ExecutionContext
 import com.keepit.common.mail.template.EmailToSend
 import com.keepit.model.cache.{ UserSessionViewExternalIdKey, UserSessionViewExternalIdCache }
 import com.keepit.shoebox.model.ids.UserSessionExternalId
 import com.keepit.model.view.{ LibraryMembershipView, UserSessionView }
 
-import scala.concurrent.Future
-import scala.concurrent.Promise
+import scala.concurrent.{ ExecutionContext => ScalaExecutionContext, Future, Promise }
 import scala.concurrent.duration._
 import com.google.inject.Inject
 import com.keepit.common.db.{ ExternalId, Id, SequenceNumber }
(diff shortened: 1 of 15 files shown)


kifi-backend/modules/eliza/test/com/keepit/eliza/controllers/mobile/MobileMessagingControllerTest.scala
```diff
@@ -51,7 +53,56 @@ class MobileMessagingControllerTest extends Specification with ElizaTestInjector
...
+
+        inject[FakeUserActionsHelper].setUser(shanee)
+
+        val result = inject[MobileMessagingController].addParticipantsToThread(threadId = thread.externalId, users = eishay.externalId.toString, emailContacts = "joe@smith.com,jack@smith.com")(FakeRequest().withHeaders("user-agent" -> "iKeefee/1.0.12823 (Device-Type: iPhone, OS: iOS 7.0.6)"))
+
+        status(result) must equalTo(OK)
+        contentType(result) must beSome("text/plain")
+        val runners = inject[WatchableExecutionContext].drain()
+        //yes! 41 futures where created to get this test to pass! You can remove or edit the next line if it breaks test, just a nice to know
+        runners === 41
```
andrewconner: When this breaks, this will likely break _only_ after clean, since it's measuring a side effect not taken into account by sbt. I don't think we should include it in this.

eishay: sure

kifi-backend/modules/common/app/com/keepit/shoebox/ShoeboxServiceClient.scala
```diff
@@ -1,12 +1,12 @@
...
 
+import com.keepit.common.concurrent.ExecutionContext
 import com.keepit.common.mail.template.EmailToSend
 import com.keepit.model.cache.{ UserSessionViewExternalIdKey, UserSessionViewExternalIdCache }
 import com.keepit.shoebox.model.ids.UserSessionExternalId
 import com.keepit.model.view.{ LibraryMembershipView, UserSessionView }
 
-import scala.concurrent.Future
-import scala.concurrent.Promise
+import scala.concurrent.{ ExecutionContext => ScalaExecutionContext, Future, Promise }
```
2is10, line 9: Why do we define our `ExecutionContext` class with the same name as Scala’s? Now every time we see `ExecutionContext` in our code, we have to check the imports section to know which one it is.

Also, looks like the naming results in at laest one scalac warning:

```
[warn] modules/common/app/com/keepit/common/concurrent/ExecutionContextModule.scala:6: imported `ExecutionContext' is permanently hidden by definition of object ExecutionContext in package concurrent
[warn] import scala.concurrent.ExecutionContext
[warn]                         ^
```

PR #11553, opened 2014-10-16
fix delete tag modal & fix undo remove

ahsu1230:
fix up delete tag modal from #11533 
and when "undo" pops up, show unremove accordingly

@yipingliao @andrewconner


yipingliao:
:+1:

andrewconner:
Works for me

PR #11557, opened 2014-10-16
Search Bar Header

ahsu1230:
Search bar on top in header! (only for library-experimenters)
- search bar should be in header on ALL pages
- Chip (library name & owner pic) appears when visiting a library
- pressing delete (when no query) should delete the chip
- unfortunately, right now, we're using the old searching through a re-direct. We'll have library-specific searching soon

@andrewconner @yipingliao


kifi-backend/modules/shoebox/angular/src/friends/friends.js
```diff
@@ -41,6 +41,8 @@ angular.module('kifi')
 
     friendService.getKifiFriends();
     friendService.getRequests();
+
+    $rootScope.$emit('libraryUrl', {});
```
andrewconner, line 45: Hm, I imagine this is to clear the chip? If so, pretty easy to forget.

What if you listen to `$locationChangeSuccess` in header and clear out if it's not a library?

```
$scope.$on('$locationChangeSuccess', function(event) {
  // ....
});
```

kifi-backend/modules/shoebox/angular/src/layout/header/HeaderCtrl.js
```diff
@@ -20,6 +20,74 @@ angular.module('kifi')
       }
     });
 
+    $scope.isFocused = false;
```
andrewconner, line 23: I can see this increasing in complexity over time. How about a separate directive?

ahsu1230:
merging now to not block @stkem 
Will move things to a separate directive & $locationChange later =)

PR #11559, opened 2014-10-16
refreshed invite email

joshm1:
- updated bookmark import and keep from email tips
- unified green color on all buttons

@stkem


stkem:
Yay!

PR #11560, opened 2014-10-16
remove MainSearcherFactory from user search

ymatsuda:
@yingjieMiao @leogrim 

I am getting ready to remove MainSearcher and MainSearcherFactory. I needed to create a new commander, UserSearchCommander to make user search independent from MainSearcherFactory.
There are two controllers that do user search. Their codes are similar but not identical. So, I make the new UserSearchCommander to support both use cases. 
I think I caught subtle differences of user search between SearchController and MobileUserSearchController.
Please review carefully. Thanks!


yingjieMiao:
lgtm!

PR #11562, opened 2014-10-16
RIP ActionAuthenticator

raymondng:
This should do it.

@andrewconner


raymondng:
Festive :) Alright merging!

PR #11564, opened 2014-10-16
do not send user social connection made notifications/emails for 5 minutes

joshm1:
@stkem 

all this just to add a 5 minute delay to sending the notifications...

in retrospect i should've just added an `Option[Duration]` to `updateUserConnections` and passed None from my tests. :angry:


stkem:
lgtm. We'll have to come up with a permanent solution for this in the near future though. I don't have a good idea, but will think about it.

PR #11567, opened 2014-10-16
Fix issues with followers

yipingliao:
First, the number of followers is not updated when the user leaves the library.
https://app.asana.com/0/15523651812135/18098869304217

Next, the "Follow" button was incorrect in that it thinks the user is a follower if the user is in the set of 10 followers initially returned by the server. When there are > 10 followers and the user is not in the first 10, the following status would be wrong.

Third, clean up synchronization signals by removing one of them. @ahsu1230 I am really happy about this. :)

@andrewconner @ahsu1230

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/libraryService.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/libraryService.js
@@ -46,7 +46,13 @@ angular.module('kifi')
 
     var libraryByUserSlugService = new Clutch(function (username, slug, authToken) {
       return $http.get(routeService.getLibraryByUserSlug(username, slug, authToken)).then(function (res) {
-        return res.data && res.data.library;
+        // TODO: Take this manual check out when the backend endpoint properly has an 'access' property
+        //       on the library object.
+        if (res.data && res.data.library) {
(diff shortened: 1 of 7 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/libraryService.js
```diff
@@ -225,6 +232,13 @@ angular.module('kifi')
         return $http.post(routeService.leaveLibrary(libraryId)).then(function () {
           _.remove(librarySummaries, { id: libraryId });
           $rootScope.$emit('librarySummariesChanged');
+
+          return api.getLibraryById(libraryId, true).then(function (data) {
+            // TODO: Take this manual check out when the backend endpoint properly has an 'access' property
```
yipingliao, line 237: @ahsu1230 Please remember to make Asana task; this is really ugly.

ahsu1230, line 237: noted

yipingliao:
Gonna merge this. Please let me know if you guys notice anything funny with the left nav or the followers. Thanks!

PR #11570, opened 2014-10-17
New icons

andrewconner:
@yipingliao


yipingliao:
:+1: Thanks for doing this!

2is10:
Will the extension need updates too?

andrewconner:
@2is10 Sounds like it. This was unplanned. I have the SVGs, but they're optimized for exactly the size we're using (shapes/paths are pixel aligned at the target size). Grab Danny and me, we'll figure out what you need.

PR #11572, opened 2014-10-17
Add invited people to FullLibraryInfo and paginated getLibraryMembers en...

leogrim:
...dpoint
@andrewconner


kifi-backend/modules/shoebox/app/com/keepit/commanders/RecommendationsCommander.scala
```diff
@@ -219,7 +219,7 @@ class RecommendationsCommander @Inject() (
               url = libInfo.url,
               description = libInfo.description,
               owner = db.readOnlyReplica { implicit session => basicUserRepo.load(ownerId) },
-              followers = libInfo.followers,
+              followers = libInfo.followers.takeWhile(_.lastInvitedAt.isEmpty).map(_.member.left.get),
```
leogrim: @stkem @yingjieMiao

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
```diff
@@ -229,44 +229,29 @@ class LibraryController @Inject() (
...
+        val limit = Math.min(count, 30)
         val numKeepsF = libraryCommander.getKeepsCount(libraryId)
         for {
-          keeps <- libraryCommander.getKeeps(libraryId, take, offset)
+          keeps <- libraryCommander.getKeeps(libraryId, offset, limit)
           keepInfos <- keepsCommander.decorateKeepsIntoKeepInfos(request.userIdOpt, keeps)
           numKeeps <- numKeepsF
         } yield {
-          Ok(Json.obj("keeps" -> Json.toJson(keepInfos), "count" -> Math.min(take, keepInfos.length), "offset" -> offset, "numKeeps" -> numKeeps))
+          Ok(Json.obj("keeps" -> Json.toJson(keepInfos), "count" -> Math.min(limit, keepInfos.length), "offset" -> offset, "numKeeps" -> numKeeps))
```
leogrim, line 245: @andrewconner I know it's probably low on your checklist right now, but I'd like to make sure these APIs make sense. They have not been pushed to any mobile build yet, so we can still change them safely, it could prevent maintenance work later on. Typically, if this is meant to defend against abuse, why not just return an error?

kifi-backend/modules/shoebox/app/com/keepit/commanders/RecommendationsCommander.scala
yingjieMiao: not clear to me why `left.get` won't throw exception here. I might missed some logic.

leogrim: followers = collaborators + followers + invited users + invited email addresses
Only invited people have lastInvitedAt set, so this gives you collaborators and followers, who must be users.

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryInviteRepo.scala
```diff
@@ -115,4 +117,27 @@ class LibraryInviteRepoImpl @Inject() (
     }
   }
 
+  def pageInviteesByLibraryId(libraryId: Id[Library], offset: Int, limit: Int, includeStates: Set[State[LibraryInvite]])(implicit session: RSession): Seq[(Either[Id[User], EmailAddress], Set[LibraryInvite])] = {
+    val invitees = {
+      val invitesGroupedByInvitee = (for (r <- rows if r.libraryId === libraryId && r.state.inSet(includeStates)) yield r).groupBy(r => (r.userId, r.emailAddress))
+      val inviteesWithLastInvitedAt = invitesGroupedByInvitee.map { case ((userId, emailAddress), invites) => (userId, emailAddress, invites.map(_.createdAt).min) }
```
yingjieMiao, line 123: named as `LastInvitedAt`, but seems we are taking `min` of time stamp.

leogrim, line 123: This is just for sorting the results (this way the ordering is stable, independent of the user's actions e.g. resend). This field is not returned by the query.

kifi-backend/modules/shoebox/app/com/keepit/commanders/RecommendationsCommander.scala
yingjieMiao: Got you. The `takeWhile` method relies on the fact that `getLibraryMembers` packs library members in a certain order. This looks brittle. I understand this is an optimization.

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryInviteRepo.scala
```diff
@@ -115,4 +117,27 @@ class LibraryInviteRepoImpl @Inject() (
...
+    }
+
+    val invitesByUserId = {
+      (for (r <- rows if r.libraryId === libraryId && r.state.inSet(includeStates) && r.userId.inSet(userIds)) yield r).list
+    }.groupBy(_.userId.get).mapValues(_.toSet)
+
+    val invitesByEmailAddress = {
+      (for (r <- rows if r.libraryId === libraryId && r.state.inSet(includeStates) && r.emailAddress.inSet(emailAddresses)) yield r).list
+    }.groupBy(_.emailAddress.get).mapValues(_.toSet)
+
```
yingjieMiao, line 140: running 3 db queries here? Looks like the last 2 queries are just filtered version of the 1st query (ignoring the groupBy part). Wondering if we can do query only one time and do logics in scala code?

leogrim, line 140: The trick is that the sort / drop / take applies to invited people, not to invites. So the first query figures out who to show, the second and third query returns the corresponding invitations. I struggled with Slick to reuse the first query as a subquery for the second / third (which could be merged). So since I struggled, I went for better readability. Maybe I should rewrite this in pure SQL. I'll take another shot at it for pure performance, but at least this is correct and gets the website team what they need.

Even without this, the second and third query could indeed be merged, but I optimized for readability.

andrewconner:
This appears to be a breaking change (followers -> members). We need to corrdinate the frontend for this before merging

leogrim:
The only endpoint that uses "members" is getMembers, which is a new one that replaces getCollaborators (unused per our discussion yesterday)

andrewconner:
Ah I see, my mistake. So the library json is changed (just `followers` has more data now). Makes sense, carry on ;)

kifi-backend/modules/shoebox/app/com/keepit/commanders/RecommendationsCommander.scala
leogrim: Well, I'm not sure it's such a good idea in the long run to mix in actual members and invites, but this is the current design. Maybe we should have two distinct fields in FullLibraryInfo, the front-end could concatenate them. What do you think @andrewconner ?

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryInviteRepo.scala
yingjieMiao, line 140: I meant alternatively we can run 

```
(for (r <- rows if r.libraryId === libraryId && r.state.inSet(includeStates)) yield r).list
```

and do scala logic from there on. I don't feel very strong about this.

leogrim, line 140: I'd rather paginate in the database, a library could end up with many followers.

leogrim:
@andrewconner I'm going to merge this with the changes we discussed. I'll send a different pull-request this afternoon with clean / consistent APIs for both the website and mobile.

andrewconner:
Works for me

PR #11573, opened 2014-10-17
setting username to all users with no username

eishay:
please review @andrewconner


kifi-backend/modules/shoebox/app/com/keepit/commanders/UserCommander.scala
```diff
@@ -611,4 +613,24 @@ class UserCommander @Inject() (
...
+        if (user.username.isDefined) throw new Exception(s"user already has a user name: $user")
+        val username = autoSetUsername(user, readOnly)
+        if (username.isEmpty) throw new Exception(s"could not set a username for $user")
+        log.info(s"[readOnly = $readOnly] [$counter] setting user ${user.id.get} ${user.fullName} with username ${username.get}")
+      }
+      batch = db.readOnlyMaster { implicit s =>
+        userRepo.getUsersWithNoUsername(batchSize)
+      }
+    }
+    counter
```
andrewconner, line 634: Never increments?

kifi-backend/modules/shoebox/app/com/keepit/commanders/UserCommander.scala
```diff
@@ -611,4 +613,24 @@ class UserCommander @Inject() (
...
+  def updateUsersWithNoUserName(readOnly: Boolean): Int = {
+    var counter = 0
+    val batchSize = 50
+    var batch = db.readOnlyMaster { implicit s =>
+      userRepo.getUsersWithNoUsername(batchSize)
+    }
+    while (batch.nonEmpty) {
+      batch foreach { user =>
+        if (user.username.isDefined) throw new Exception(s"user already has a user name: $user")
+        val username = autoSetUsername(user, readOnly)
```
andrewconner, line 625: We'll want this to return a `Username` (not Option) obviously, but it _should_ work for all since we try 30 times.

andrewconner:
Thanks

andrewconner:
Let me get https://github.com/kifi/keepit/pull/11568 in first, because otherwise, we'll normalize with the bug ;)

PR #11574, opened 2014-10-17
fixing user invitations #acceptInvite

eishay:


eishay:
@andrewconner @raymondng 
User invitation was broken for three days in production!!!

eishay:
apparently it didn't fix it

andrewconner:
Just tested, seems to work?

PR #11577, opened 2014-10-17
more RESTful route for mobile add participants endpoint

2is10:
@eishay We try to use RESTful public-facing route names. Using the controller method name in the path is a convention for our internal routes only, since those routes are much easier to update.

cc @thassss @jaredpetker @stkem 
Is it too late for this?


2is10:
Interpreting lack of response as lack of support. Maybe let’s just do better with our public routes going forward: 1) favor simple nouns over phrases, 2) don’t use a verb if the HTTP method suffices (e.g. `POST` to a plural resource means add/create).

jaredpetker:
I can change the endpoint anytime before release, straightforward for me.

PR #11578, opened 2014-10-17
Fixing redirects

andrewconner:
@stkem Shouldn't prepend `"/"` in front of redirects (makes redirects to a different domain impossible).


stkem:
ah, good point

PR #11579, opened 2014-10-17
Fixing url escaping issue

andrewconner:
@leogrim


andrewconner:
(Not introduced by you, just tagging you)

PR #11583, opened 2014-10-17
Mini lib card followup

yipingliao:
Fix a couple of small bugs with the mini hover card:
- It didn't show up on invited libraries
- It slides under the main content when the browser width is small

@andrewconner @stkem

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
@@ -159,7 +159,6 @@ angular.module('kifi')
               return;
             }
 
-            // Let the user be the first one listed. :)
             library.followers.push({
               id: profileService.me.id,
               firstName: profileService.me.firstName,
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
```diff
@@ -159,7 +160,7 @@ angular.module('kifi')
             }
 
             // Let the user be the first one listed. :)
-            library.followers.unshift({
+            library.followers.push({
```
yipingliao, line 162: When a user follows from a mini hover card, there is no way from the client side to unshift the user to be first. So I am changing this to be consistent with that behavior.

kifi-backend/modules/shoebox/angular/src/libraries/libraryMiniCard.tpl.html
```diff
@@ -15,14 +15,13 @@
       </span>
     </div>
   </div>
-  <div class="kf-library-mini-card-body">
```
yipingliao, line 18: Just whitespace changes in this file.

stkem:
Thanks! lgtm

PR #11594, opened 2014-10-17
no longer need to handle extension specifying image at keep time

2is10:
@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtLibraryController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtLibraryController.scala
@@ -102,46 +102,37 @@ class ExtLibraryController @Inject() (
       db.readOnlyMaster { implicit s =>
         libraryMembershipRepo.getOpt(request.userId, libraryId)
       } match {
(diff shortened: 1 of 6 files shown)


kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtLibraryController.scala
```diff
@@ -102,46 +102,37 @@ class ExtLibraryController @Inject() (
       db.readOnlyMaster { implicit s =>
         libraryMembershipRepo.getOpt(request.userId, libraryId)
       } match {
-        case Some(mem) if mem.access != LibraryAccess.READ_ONLY =>
-          val info = request.body
+        case Some(mem) if mem.canWrite => // TODO: also allow keep if mem.canInsert and keep is not already in library
```
2is10, line 105: @andrewconner It seems that `KeepInterner` doesn’t yet have a method with "keep only if not already kept" semantics, which I think would be required for `READ_INSERT` access to work properly.

2is10:
merging to use it, review comments appreciated

andrewconner:
Works for me

PR #11598, opened 2014-10-17
Library getKeeps API cleanup

leogrim:
@andrewconner @thassss @jaredpetker 
Please confirm you're good with these changes in input ("count" query parameter is deprecated and should be replaced with "limit") and output ("count" and "offset" are no longer returned)

It still seems weird to me that we're returning the total keep count when fetching each page, but this needs more discussion.


jaredpetker:
Could you ping us once this is deployed? thanks!

PR #11599, opened 2014-10-17
smart scroll on followers list in manage library modal

ahsu1230:
as well as a few style fixes in header

Thanks @andrewconner for the tip! @yipingliao


ahsu1230:
woops messed up

PR #11600, opened 2014-10-17
smart scroll on follower List for libraries & header fix

ahsu1230:
header search bar had bugs in it when you made the window smaller

and now smart scroll in manageLib modal

Thanks @andrewconner for the tips!
@yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.js
@@ -128,6 +128,38 @@ angular.module('kifi')
           scope.showFollowers = false;
         };
 
+        scope.moreFollowers = true;
+        scope.followerList = [];
+        scope.followerScrollDistance = '100%';
+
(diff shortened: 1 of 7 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.js
```diff
@@ -128,6 +128,38 @@ angular.module('kifi')
           scope.showFollowers = false;
         };
 
+        scope.moreFollowers = true;
```
yipingliao, line 131: Can you add a section header for this whole segment of code, like "Followers scroll"? Right now it's all under the "Scope methods" section.

yipingliao:
Thanks for doing this!

kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.js
ahsu1230, line 131: sure will do

PR #11601, opened 2014-10-17
Add ownerId, librarySlug to LibraryRecord

leogrim:
@ymatsuda @yingjieMiao 
Following a design change, this is so that we can include library urls in search results and make the library cartridges clickables. This still requires fetching the owner's basic user to get his / her username. Ideally we could denormalize the owner's username into the library table, but that requires making sure it stays in sync when the user changes it, so we'll get into that later.


yingjieMiao:
lgtm

ymatsuda:
lgtm

PR #11604, opened 2014-10-17
Style public (logged out) library header (prelimary).

yipingliao:
DO NOT MERGE Awaiting consultation with Danny and new Kifi logo asset.

@andrewconner I've still got a couple of things to check with Danny and need the new new Kifi logos, but wanted to put this in as a pull right now.


kifi-backend/modules/shoebox/angular/src/layout/main/main.styl
```diff
@@ -10,6 +10,9 @@
     //border-right: 1px solid #dbdee3
     box-sizing: border-box
 
+.kf-logged-out .kf-main
+  width: 100%
```
yipingliao, line 14: Adding this allows the rendering to get the wider width in the beginning. That was causing problems with displaying the followers' pictures (with an initial narrow width, no follower pics were ever being displayed).

kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
```diff
@@ -21,6 +21,7 @@ angular.module('kifi')
         //
         // Scope data.
         //
+        scope.isUserLoggedOut = $rootScope.userLoggedIn === false;
```
yipingliao, line 24: I copied this from `dev.html` - using triple equals to be consistent; not sure if truthy/falsy test is adequate.

kifi-backend/modules/shoebox/angular/src/layout/header/publicHeader.tpl.html
```diff
@@ -0,0 +1,17 @@
...
+    <div class="kf-header-left">
+      <a href="/">
+        <span class="reco-icon svg-reco-icon"></span>
+        <img class="kf-header-logo" src="/img/kifi-logo-color.png">
+        <span class="public-header-tagline">Connecting people with knowledge</span>
+      </a>
+    </div>
+
+    <div class="kf-header-right">
+      <a href="//www.kifi.com/signup" class="public-header-link">Join Kifi</a>
```
yipingliao: Keep the `kifi.com` hostname, or use `/signup` only? I needed the hostname for dev, but am not sure if this good practice for prod.

andrewconner:
Wahoo!

kifi-backend/modules/shoebox/angular/src/layout/header/publicHeader.tpl.html
andrewconner: Yeah, I've been torn. For signup/auth, we actually do need the kifi.com domain.

2is10: Could we use `env.navBase` instead? It’s intended specifically for cases when we need to navigate to a Play server route (on either dev.ezkeep.com:9000 or else www.kifi.com).

yipingliao: Yes, thank you. Will do.

andrewconner: Thanks @2is10. Yeah that's best.

yipingliao:
Closing to resolve merge conflicts.

PR #11606, opened 2014-10-17
hooking up new search endpoint with library attribution

stkem:
this is backwards compatible, i.e. should work correctly for people without the libraries experiment, although the code can be simplified a bit when everyone is on libraries. 
Also includes a few minor fixed to the library hover card.

@andrewconner @yipingliao @ahsu1230 @joshm1 
cc @leogrim


kifi-backend/modules/shoebox/angular/src/common/services/routeService.js
```diff
@@ -81,6 +81,7 @@ angular.module('kifi')
         return route('/user/' + id + '/hide');
       },
       search: searchRoute('/site/search'),
```
andrewconner, line 83: Remove, rename `search2` → `search`

andrewconner:
I approve.

leogrim:
:+1:

PR #11607, opened 2014-10-17
Fix bug where the status of a library-share contact is cached even for the next library share.

yipingliao:
@andrewconner 
https://app.asana.com/0/15523651812135/18170273343523


yipingliao:
Merging b/c it's a small change. Will address comments later. :)

PR #11620, opened 2014-10-18
Add invites info to users page (admin)

raymondng:
Rudimentary invite stats. Only query for ACTIVE and ACCEPTED for now.

@andrewconner @eishay


raymondng:

PR #11621, opened 2014-10-18
For libraries-enabled users, remove mass edit privacy and add copy to library.

yipingliao:
@andrewconner Hi! I haven't hooked up the move-to-library endpoint (will do that over the weekend), but it will be very similar to what I got for copy, so am tossing this pull in for your review earlier rather than later.

https://app.asana.com/0/15523651812135/18173349899168

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/common/modal/importBookmarkFileLibraryModal.tpl.html
+++ b/kifi-backend/modules/shoebox/angular/src/common/modal/importBookmarkFileLibraryModal.tpl.html
@@ -43,10 +43,10 @@
 
     <div class="import-library-selection">
       <div class="add-keep-to-library-dropdown" ng-if="librariesEnabled" kf-keep-to-library-widget ng-click="showWidget()">
-        <span ng-switch="selection.library.kind" class="keep-to-library-lib-icon-container">
+        <span ng-switch="librarySelection.library.kind" class="keep-to-library-lib-icon-container">
           <span ng-switch-when="system_main" class="keep-to-library-lib-icon svg-main-library-icon"></span>
           <span ng-switch-when="system_secret" class="keep-to-library-lib-icon svg-secret-library-icon"></span>
-          <span ng-switch-default ng-switch="selection.library.visibility">
+          <span ng-switch-default ng-switch="librarySelection.library.visibility">
             <span ng-switch-when="published" class="keep-to-library-lib-icon svg-public-library-icon"></span>
             <span ng-switch-when="secret" class="keep-to-library-lib-icon svg-private-library-icon"></span>
 
@@ -54,13 +54,13 @@
             <span ng-switch-when="discoverable" class="keep-to-library-lib-icon svg-yellow-card-with-globe"></span>
           </span>
         </span>
-        {{selection.library.name}}
+        {{librarySelection.library.name}}
         <span class="add-keep-to-library-dropdown-arrow"></span>
       </div>
     </div>
 
     <div class="dialog-buttons">
-      <button class="import-button" ng-class="{'disabled': disableBookmarkImport || !importFilename}" ng-click="importBookmarkFileToLibrary(selection.library)">Import</button>
+      <button class="import-button" ng-class="{'disabled': disableBookmarkImport || !importFilename}" ng-click="importBookmarkFileToLibrary(librarySelection.library)">Import</button>
     </div>
   </div>
 </div>
--- a/kifi-backend/modules/shoebox/angular/src/common/services/routeService.js
+++ b/kifi-backend/modules/shoebox/angular/src/common/services/routeService.js
@@ -111,6 +111,9 @@ angular.module('kifi')
       addKeepsToLibrary: function (libraryId) {
         return route('/libraries/' + libraryId + '/keeps');
       },
+      moveKeepsToLibrary: function () {
+        return route('/libraries/move');
+      },
       removeKeepFromLibrary: function (libraryId, keepId) {
         return route('/libraries/' + libraryId + '/keeps/' + keepId);
       },
--- a/kifi-backend/modules/shoebox/angular/src/common/services/routeService.js
+++ b/kifi-backend/modules/shoebox/angular/src/common/services/routeService.js
@@ -111,6 +111,9 @@ angular.module('kifi')
       addKeepsToLibrary: function (libraryId) {
         return route('/libraries/' + libraryId + '/keeps');
       },
+      copyKeepsToLibrary: function () {
+        return route('/libraries/copy');
+      },
       moveKeepsToLibrary: function () {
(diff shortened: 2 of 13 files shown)


kifi-backend/modules/shoebox/angular/src/common/modal/importBookmarkFileLibraryModal.tpl.html
```diff
@@ -43,10 +43,10 @@
 
     <div class="import-library-selection">
       <div class="add-keep-to-library-dropdown" ng-if="librariesEnabled" kf-keep-to-library-widget ng-click="showWidget()">
-        <span ng-switch="selection.library.kind" class="keep-to-library-lib-icon-container">
+        <span ng-switch="librarySelection.library.kind" class="keep-to-library-lib-icon-container">
```
yipingliao, line 46: Global rename of `selection` to `librarySelection`. Originally, the library widget expects its parent scope to have a `selection` property. However, in `kfKeeps`, the scope already has a `selection` property (it tracks which keeps were selected in the edit-keeps mode), so I've changed the contract for the widget to expect `librarySelection` instead. Note: isolate scope for this widget is not possible in this case. Long story; can discuss in person. :)

kifi-backend/modules/shoebox/angular/src/keeps/keeps.js
```diff
@@ -247,6 +263,28 @@ angular.module('kifi')
...
+        scope.$watch('keepsLoading', function (newVal) {
+          if (!newVal) {
+            if (scope.librariesEnabled) {
+              libraryService.fetchLibrarySummaries().then(function () {
+                scope.libraries = _.filter(libraryService.librarySummaries, function (library) {
+                  return (library.access !== 'read_only') && (library.id !== scope.library.id);
+                });
+
+                scope.librarySelection = {};
+                scope.libSelectTopOffset = 300;  // This needs to be dynamic.
```
yipingliao, line 290: Oops, still need to do this too.

kifi-backend/modules/shoebox/angular/src/keeps/keeps.tpl.html
```diff
@@ -7,7 +7,15 @@
...
       <div class="kf-keep-selection-options" ng-class="{disabled: editingTags || selection.getSelected(keeps).length === 0}">
-        <div class="kf-keep-selection-option"><span ng-click="togglePrivate(keeps)">Make {{selectionPrivacyState(keeps)}}</span></div>
+        <div ng-if="!librariesEnabled" class="kf-keep-selection-option"><span ng-click="togglePrivate(keeps)">Make {{selectionPrivacyState(keeps)}}</span></div>
+
+        <div ng-if="librariesEnabled" class="kf-keep-selection-option">
+          <span kf-keep-to-library-widget ng-click="showWidget()" libraries="libraries" selection="librarySelection">
+            Copy to library
+          </span>
+        </div>
+        <!-- <div ng-if="librariesEnabled" class="kf-keep-selection-option"><span>Move to library</span></div> -->
```
yipingliao: Moving to come in a bit.

andrewconner:
I'll go through more thoroughly later, but make sure this doesn't break current (non-library) users.

Edit: saw the title of this, you probably already checked :)

yipingliao:
Thanks for the reminder! I am always worried that I forget about the non-libraries user cases. Here, I placed the different kinds of links in `ng-if`s (in `keeps.tpl.html`) conditioned on whether libraries are enabled. Also manually checked for both types of users. Just pushed changes for moving keeps to a library.

kifi-backend/modules/shoebox/angular/src/common/services/routeService.js
```diff
@@ -111,6 +111,9 @@ angular.module('kifi')
       addKeepsToLibrary: function (libraryId) {
```
yipingliao, line 111: @ahsu1230 For copying keeps to a library, I am still using the route with the `libraryId` inside it (i.e., the `addKeepsToLibrary` route). It works fine. Do we need the `copy` route? (Does it do something different in the database? Adding keeps to a library _is_ copying them).

kifi-backend/modules/shoebox/angular/src/keeps/keeps.js
```diff
@@ -113,6 +114,34 @@ angular.module('kifi')
           }
         }
 
+        function copyToLibrary () {
+          // Copies the keeps that are selected into the library that is selected.
+          var selectedKeeps = scope.selection.getSelected(scope.keeps);
+          var selectedLibrary = scope.librarySelection.library;
+
+          keepActionService.keepToLibrary(_.pluck(selectedKeeps, 'url'), selectedLibrary.id).then(function () {
+            // TODO: look at result and flag errors. Right now, even a partial error is flagged so that's
```
yipingliao, line 123: @ahsu1230 I need some help with figuring out what error results the server might pass back. At first I was just checking for `result.error`, but that's not good in the context of moving/copy keeps in the library, because that is true even if the action was partially successful (e.g., most keeps were copied/moved but one keep was already in the library).

yipingliao:
@ahsu1230 After I hooked up this move endpoint and tested it, I found that the keeps were successfully removed from the source library and added to the destination library. :) However, the `numKeeps` property on the library object of the source library was not updated. Note that the `numKeeps` property on the destination library does get correctly updated.

kifi-backend/modules/shoebox/angular/src/common/services/routeService.js
ahsu1230, line 111: The main difference between the two endpoints is that copy will actually try to transfer tags over to the new copied keeps. Yes, the two endpoints are similar, but let's try to use the copy keeps endpoint so it's more explicit in what we're trying to do =)

kifi-backend/modules/shoebox/angular/src/keeps/keeps.js
ahsu1230, line 123: right I think we need to talk about this. Currently, for moveKeeps, when a keep URI is already in a library, trying to move it will do nothing and the original keep will stay in its current library. For better UI it would make sense to remove that keep from the original library, and leave the keep in the target library as is. Thoughts? @andrewconner

ahsu1230:
I don't think the moveKeeps function() is supposed to report back a `numKeeps` field. It's really just a `success` field or a `failures` field. So, after a moveKeeps, you may need to call another get summaries. Or am I misunderstanding the question?

kifi-backend/modules/shoebox/angular/src/common/services/routeService.js
yipingliao, line 111: Got it. Will do.

yipingliao:
After a moveKeeps, when I re-fetch the summaries, the `numKeeps` on the source library did not decrease.

PR #11623, opened 2014-10-18
library name truncate & rework library follower smart scrolling

ahsu1230:
@andrewconner @yipingliao 
@2is10 using the multi-line ellipsis! http://www.mobify.com/blog/multiline-ellipsis-in-pure-css/

normal-case:
library-card-title-normal

bad-case:
library card title worst


andrewconner:
Let's clear this by product before merging, since it's unfortunately visually different. Looks like there's a script solution: http://dotdotdot.frebsite.nl/

yipingliao:
Thanks for fixing these two issues. :+1: 

Looks like dotdotdot is almost 700 lines of code!

2is10:
Looks like that script solution uses binary search, like I suggested to Aaron/Yiping on Friday.
We could write just the functionality we need in 15-20 lines, I bet.

andrewconner:
I vote for that. This implementation is a lot simpler, but doesn't look close to the designs. We should drop it entirely (ie, hard cut off) or try to get it close to the design.

ahsu1230:
closing PR since stuff needs to get reworked anyway

PR #11624, opened 2014-10-18
add max for updating users with no username so we can do some small scal...

eishay:
...e tests and base logic for update all normalized username


eishay:
@andrewconner

PR #11627, opened 2014-10-19
add renormalization code, dealing with empty first/last names and trimmi...

eishay:
...ng very long first/last names as some are killing the validator and just don't make sense.

The max in the db is 64, but 33 chars should do for now :-)

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/UserCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/UserCommander.scala
@@ -542,7 +542,13 @@ class UserCommander @Inject() (
   }
 
   def autoSetUsername(user: User, readOnly: Boolean): Option[Username] = {
-    val name = s"${UsernameOps.lettersOnly(user.firstName)}-${UsernameOps.lettersOnly(user.lastName)}".toLowerCase
+    val firstName = UsernameOps.lettersOnly(user.firstName.trim).take(15).toLowerCase
+    val lastName = UsernameOps.lettersOnly(user.lastName.trim).take(15).toLowerCase
+    val name = if (firstName.isEmpty || lastName.isEmpty) {
+      if (firstName.isEmpty) lastName else firstName
+    } else {
+      s"$firstName-$lastName"
+    }
     val seed = if (name.length < 4) {
       val filler = Seq.fill(4 - name.length)(0)
       s"$name-$filler"
@@ -661,7 +667,7 @@ class UserCommander @Inject() (
     counter
   }
 
-  def reNormalizedUsername(readOnly: Boolean): Int = {
+  def reNormalizedUsername(readOnly: Boolean, max: Int): Int = {
     var counter = 0
     val batchSize = 50
     var page = 0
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/UserCommander.scala
```diff
@@ -542,7 +542,13 @@ class UserCommander @Inject() (
   }
 
   def autoSetUsername(user: User, readOnly: Boolean): Option[Username] = {
-    val name = s"${UsernameOps.lettersOnly(user.firstName)}-${UsernameOps.lettersOnly(user.lastName)}".toLowerCase
+    val firstName = UsernameOps.lettersOnly(user.firstName.trim).take(15).toLowerCase
+    val lastName = UsernameOps.lettersOnly(user.lastName.trim).take(15).toLowerCase
+    val name = if (firstName.isEmpty || lastName.isEmpty) {
```
yingjieMiao, line 547: what happens when firstName and lastName both empty?

eishay, line 547: then a random four letter character will be the name (see line 553)

kifi-backend/modules/shoebox/test/com/keepit/model/UserTest.scala
```diff
@@ -20,6 +20,9 @@ class UserTest extends Specification with ShoeboxTestInjector {
     "valid" in {
       UsernameOps.isValid("eishay-kifi") === false
       UsernameOps.isValid("nada-boutros") === false
+      //the following line takes more then a minute to run!
+      //UsernameOps.isValid("eishaytestwithaveryveryveryveryveryveryveryverylongmailgocom-") === false
```
yingjieMiao, line 24: really? takes 1 minute to run? 
I see, we have a long list of strings in the source code to construct hashsets. The source code's width is 46210 characters ...

yingjieMiao, line 24: no way..... it's the regex matching takes too long...

```
val usernameRegex = """([\p{L}\d][\p{L}\d\.\_\-]?){2,30}[\p{L}\d]""".r.pattern
val x = "eishaytestwithaveryveryveryveryveryveryveryverylongmailgocom-"
usernameRegex.matcher(x).matches
```

my laptop burns.  can we improve the performance here? 

It fails on any long string! this will tear down shoebox. 

try also 

```
val x = "abddkdsldlafklskfldsajflashgasdhglahsdghadshgasdhglasdlgdddddddddddddddddddddddsssss""
```

or any thing long.

eishay, line 24: I set a limit on the size that would be checked (the generated usernames),
its few milliseconds now and I already created usernames to all users in
our db.

yingjieMiao, line 24: Too much backtracking is the real problem, even size fits. 

consider 

```
val re = """([\p{L}\d][\p{L}\d\.\_\-]?){2,30}[\p{L}\d]""".r.pattern

val x = "0000111122223333444455556666777700001111222233334444555566667"
re.matcher(x).matches     // returns true immediately

val x = "000011112222333344445555666677770000111122223333444455556666-"
re.matcher(x).matches     // same length, only modified last char, returns false after a LONG TIME. 
```

In fact, just set

```
val x = "000011112222333344445555666677778888-"
```

can cause enough trouble.

PR #11628, opened 2014-10-19
Registration from Libraries

andrewconner:
It's a big one. There's some minor design changes made after I started, but functionally it should be good. Still need to handle some error cases (backend bugs filed).

I actually got re-load-less signups working, so when you register, it never reloads the page (unless there was an error, in which case we liberally fall back to the existing signup flow).

@yipingliao Would appreciate your feedback. I made initializing a registration session possible by using a service (so it can be called anywhere). The service then creates an instance of a controller, which handles all the modal flows. Does this seem reasonable?

Handling error cases seems a bit messy, but it's getting there. I actually think it would be cool for modalService.open to return a promise, that's resolved as successful on the success case, and rejected if the user closes it.


andrewconner:
Apologies for the ugly commits, I'll squash.

raymondng:
Whoohoo!

andrewconner:
Squashing and reopening

PR #11629, opened 2014-10-19
Library error page

andrewconner:


andrewconner:
@yipingliao

kifi-backend/modules/shoebox/angular/src/errors/notFound.tpl.html
```diff
@@ -0,0 +1,7 @@
+<div class="kf-error-body-wrap">
+  <div class="kf-error-header">Whomp. We can’t find the page you’re looking for.</div>
```
yipingliao, line 2: Since you're using `<footer>`, use `<header>` here?

yipingliao:
Nice!

PR #11631, opened 2014-10-20
username and normalized username are not optional enymore

eishay:
@andrewconner


eishay:
@andrewconner please review, don't merge yet. need to do another db run and add constraints

andrewconner:
Looks good. We should invest in testing infrustructure. As you saw, adding or removing a field to `User` (or `Keep`, etc) is _quite_ painful due to tests.

A builder pattern DSL would be pretty cool.

andrewconner:
I notice we have a user with a null username. Remember to run the migration again before merging this.

andrewconner:
Also, for a while, I introduced a bug that assigned usernames too quickly. For email signups, it used their email address (since that is their firstName until they finish registration).

`select id,username from user where id > 5600 and id < 8770;`

We should probably clear out those usernames and auto-assign.

2is10:
> For email signups, it used their email address (since that is their firstName until they finish registration).

The underlying issue sounds like technical debt. If we can create a `User` without knowing the name, then `firstName` and `lastName` should be optional. Alternatively, we might want to consider using a different model (e.g. `IncompleteEmailSignup`) to store information about email address signups until the user specifies his/her name.

PR #11632, opened 2014-10-20
Library registration

andrewconner:
It's a big one. There's some minor design changes made after I started, but functionally it should be good. Still need to handle some error cases (backend bugs filed).

I actually got re-load-less signups working, so when you register, it never reloads the page (unless there was an error, in which case we liberally fall back to the existing signup flow).

@yipingliao Would appreciate your feedback. I made initializing a registration session possible by using a service (so it can be called anywhere). The service then creates an instance of a controller, which handles all the modal flows. Does this seem reasonable?

Handling error cases seems a bit messy, but it's getting there. I actually think it would be cool for modalService.open to return a promise, that's resolved as successful on the success case, and rejected if the user closes it.


kifi-backend/modules/shoebox/angular/src/common/services/modalService.js
```diff
@@ -23,13 +23,20 @@ angular.module('kifi')
       $compile($modal)(scope);
       angular.element(document.body).find('.kf-cols').append($modal);
 
-      modals.push($modal);
+      // We need to save a reference to the scope, because .append() appears to set the scope of the element
+      // as .kf-cols's scope. Clearly we're missing something trivial (using a scope in $compile and then
+      // assigning it to the DOM node's scope), so if someone can fix it, go for it :).
+      modals.push([$modal, scope]);
```
andrewconner, line 29: @yipingliao this was my modal change to save the scope. Let me know what you think.

kifi-backend/modules/shoebox/angular/src/signup/signupService.js
```diff
@@ -0,0 +1,245 @@
...
+        setModalScope(modalService.open({
+          template: 'signup/thanksForRegisteringModal.tpl.html',
+          scope: $scope
+        }), function onClose() {
+          $scope.onSuccess();
+        });
+      } else {
+        $scope.onSuccess();
+      }
+      $rootScope.$emit('appStart');
```
andrewconner, line 240: If you ever need to simulate a page reload, this is the new way to do it. It should reload any data that users get.

andrewconner:
Merging to avoid merge conflicts with modals, feedback appreciated!

kifi-backend/modules/shoebox/angular/src/common/services/installService.js
```diff
@@ -17,6 +17,11 @@ angular.module('kifi')
       other.parentNode.insertBefore(elem, other);
     }
 
+    var detectIfIsInstalled = function () {
+      var kifiAttr = $window.document.children[0].attributes['data-kifi-ext'];
```
yipingliao, line 21: Can be rewritten to be consistent with line 26.

kifi-backend/modules/shoebox/angular/src/common/services/modalService.js
yipingliao, line 29: Looks good!

kifi-backend/modules/shoebox/angular/src/signup/signupService.js
```diff
@@ -0,0 +1,245 @@
...
+        var scope = $rootScope.$new();
+
+        userData = userData || {};
+        handlers = handlers || {};
+
+        scope.userData = userData;
+
+        scope.onError = function (reason) {
+          (handlers.onError || function (reason) {
+            $window.location.href = (reason && reason.redirect) || 'https://www.kifi.com/signup';
```
yipingliao, line 23: Use `env.navBase'.

kifi-backend/modules/shoebox/angular/src/signup/signupService.js
```diff
@@ -0,0 +1,245 @@
...
+          })(reason);
+        };
+
+
+        scope.onSuccess = function () {
+          (handlers.onSuccess || function () {
+            return true;
+            //$window.location.reload();
```
yipingliao, line 31: Remove comment?

yipingliao:
I just tried this, and it looks really good. :) Modal code makes sense to me; not sure how having `modalService.open` return a promise can be used?

PR #11636, opened 2014-10-20
Move/Copy keeps to other libraries

yipingliao:
Closed last pull because of merge conflicts. An update today is that I am now using the 'copy' endpoint. I am also always refetching library info from the network.

Follow-up:
- Do not always flush cache for library info (made Asana task: https://app.asana.com/0/15523651812135/18279074736905)
- Dynamically relocate the widget (to be done soon!)
- Per discussions with @ahsu1230 , there are also a few backend followup items to this feature.

@andrewconner @ahsu1230


ahsu1230:
had discussion with @andrewconner about clarifications with move/copy (for now). Working on it now so you can test later.

Copying/Moving keeps from Library A -> B when URI already exists in B.
- if URI is inactive, reactivate the keep in B (inactivate keep A for move) and then combine all tag infos onto keep B
- if URI is active, simply combine all tag infos onto keep B

I think later down the road, we may have like a user popup that will allow them to decide which keeps they want to replace/not replace in the destination library.

PR #11637, opened 2014-10-20
A few test cases for Hashtag sensitivity

leogrim:
@2is10 I guess you expect test cases for the commander methods, they're harder to put together because the logic relies on the search index, which is async and needs to be up-to-date. I'll create a task and will get to it when I'm done with more urgent things that other people depend on.


2is10:
Thanks. This is a decent start. I think any code using regular expressions deserves tests that exercise their main features. `"abüse"` would use the diacritical mark logic, yes?

PR #1292, opened 2013-04-01
2.2.46 legible text on windows (bye Lato)

2is10:

PR #11640, opened 2014-10-20
Style tweaks

andrewconner:
The CSS can be cleaned up since the fonts are now unified, but I'll do that in a later PR.


andrewconner:
Oops, need to handle the webfonts

2is10:
And we’ve come full circle. #1292

andrewconner:
@2is10 Hahaha

PR #11648, opened 2014-10-20
remove unnecessary timing in DbRepo

yingjieMiao:
@raymondng @ymatsuda @stkem @eishay 

we already track running time when session closes. In addition, within each db session, we are timing every operation (get, insert, delete, update), using _session starting time_. These are redundant. 

Whenever we have a batch operation like

```
bigArray.foreach{ repo.save(_) }
```

once we start timing out at an item, all subsequent items would throw `java.lang.IllegalStateException`.

This 1) could slow down the application 2) sends too much data to the log.


stkem:
Thanks! lgtm

PR #11651, opened 2014-10-20
Homepage doesn't have a library in scope

andrewconner:
@yipingliao


kifi-backend/modules/shoebox/angular/src/keeps/keeps.js
```diff
@@ -283,7 +283,7 @@ angular.module('kifi')
             if (scope.librariesEnabled) {
               libraryService.fetchLibrarySummaries().then(function () {
                 scope.libraries = _.filter(libraryService.librarySummaries, function (library) {
-                  return (library.access !== 'read_only') && (library.id !== scope.library.id);
+                  return scope.library && library.access !== 'read_only' && library.id !== scope.library.id;
```
yipingliao, line 286: `scope.libraries = !scope.library ? [] : _.filter ...`? Or am I misunderstanding the code?

PR #11652, opened 2014-10-20
bug fixing in getCandidateURIs, and throttle.

yingjieMiao:
@stkem

### DIFF
--- a/kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
+++ b/kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
@@ -50,7 +50,7 @@ class RecommendationGenerationCommander @Inject() (
     serviceDiscovery: ServiceDiscovery) extends Logging {
 
   val defaultScore = 0.0f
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
```diff
@@ -50,7 +50,7 @@ class RecommendationGenerationCommander @Inject() (
     serviceDiscovery: ServiceDiscovery) extends Logging {
 
   val defaultScore = 0.0f
-  val recommendationGenerationLock = new ReactiveLock(8)
+  val recommendationGenerationLock = new ReactiveLock(4)
```
stkem: This will reduce the entire concurrency of the recommendation generation to 4, which is really low in general. I was thinking of a separate reactive lock to reduce concurrency of just the `getCandidateUris` call to 4 (It's a very light call compared to some of the others, so that should no affect overall throughput significantly)

yingjieMiao:
@stkem  updated

yingjieMiao:
@stkem  updated

stkem:
:+1:

PR #11654, opened 2014-10-20
login/logout -- add/rm fortytwo_id

raymondng:
Explicitly setting/removing kifi user_id, replacing `SecureSocial`-based mechanism, which does not seem reliable (or is no longer reliable)

We are moving away from `SecureSocial` and in this particular case (kifi user id) we probably shouldn't depend on it in the first place.

Reviewed with @andrewconner 

Thanks @jaredpetker and @2is10 for reporting the problem. Validated locally. Will push to prod to confirm.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/AuthCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/AuthCommander.scala
@@ -1,6 +1,7 @@
 package com.keepit.commanders
 
 import com.google.inject.Inject
+import com.keepit.common.controller.KifiSession
 
 import com.keepit.common.core._
 import com.keepit.common.akka.SafeFuture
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/AuthCommander.scala
```diff
@@ -320,7 +321,7 @@ class AuthCommander @Inject() (
           error => throw error,
           authenticator =>
             Ok(Json.obj("code" -> "user_logged_in", "sessionId" -> authenticator.id))
-              .withSession(newSession - SecureSocial.OriginalUrlKey - IdentityProvider.SessionId - OAuth1Provider.CacheKey)
+              .withSession(newSession - SecureSocial.OriginalUrlKey - IdentityProvider.SessionId - OAuth1Provider.CacheKey + (KifiSession.FORTYTWO_USER_ID -> userId.toString))
```
eishay, line 324: userId is an `Option[Id[User]]` leading the the exceptions in the email
@andrewconner @raymondng

raymondng, line 324: yes. fixed.

PR #11656, opened 2014-10-20
makes wander return option of journal

yingjieMiao:
@leogrim  @stkem 

Old code: when vertex not in graph, we skip random walk, return an 'uninitialized journal'. Such journal maybe later called (getStartingVertexId), and thus throw exceptions.

New code: return an option


leogrim:
Or maybe a Try instead of an Option?

yingjieMiao:
semantically `Try` might be better. Functionally they are equivalent. Will merge this to fix issues in production.

PR #11659, opened 2014-10-20
redo copy/move keeps between libraries - also handles tag info

ahsu1230:
@andrewconner 
functions are more explicit with what they do regarding URIs in a library that do not exist yet, already exist but keep is inactive, or already exist & keep is active.

and then with tests


yipingliao:
Thanks for doing this so quickly!

PR #11665, opened 2014-10-21
Do not publicly share any library that's not public.

yipingliao:
https://app.asana.com/0/15523651812135/18301952819592

@andrewconner


andrewconner:
Great, thanks.

PR #11666, opened 2014-10-21
fix count

yingjieMiao:
@stkem @andrewconner 

this is (only?) used on the admin bookmark page to show keeps within 24 hours.


stkem:
This is definitely a bug, but I don't thing it will help with the count being inflated due to people making tags into libraries, since that creates actually new entries in the bookmark table, with new `createdAt`.

yingjieMiao:
as discussed, this fixes an existing bug, though not entirely fix the problem

PR #11667, opened 2014-10-21
fix typo/bug

raymondng:
Overlooked this one. My bad.
@andrewconner @eishay


andrewconner:
Would you please write a test for this, verifying cookies / redirects etc for login / logout methods?

raymondng:
Sure

PR #11669, opened 2014-10-21
userId => userIdOpt

raymondng:
Minor refactor: in this case, better/clearer naming would help

@andrewconner @eishay @2is10


eishay:
lgtm

PR #11670, opened 2014-10-21
adding a openInAppIfMobile flag to links. if the flag exist and its a mo...

eishay:
...bile browser, a new script should be sent that should open the app store or the app if exist on both iOS and Android

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/KifiSiteRouter.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/KifiSiteRouter.scala
@@ -1,6 +1,7 @@
 package com.keepit.controllers.website
 
 import com.google.inject.{ Provider, Inject, Singleton }
+import com.keepit.common.http._
 import com.keepit.common.controller._
 import com.keepit.common.db.slick.DBSession.RSession
 import com.keepit.common.db.slick.Database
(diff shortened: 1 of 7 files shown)


eishay:
will test on mobile device before merging.
@andrewconner @jaredpetker @joshm1 please review

eishay:
got inspiration for the redirect from http://fokkezb.nl/2013/09/20/url-schemes-for-ios-and-android-2/

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/KifiSiteRouter.scala
```diff
@@ -51,12 +52,16 @@ class KifiSiteRouter @Inject() (
...
-  def routeRequest[T](request: MaybeUserRequest[T]) = {
+  def routeRequest[T](request: MaybeUserRequest[T]): Result = {
     // Short-circuit for landing pages
+    val userAgentOpt = request.userAgentOpt
     if (request.host.contains("42go")) {
       MovedPermanently(applicationConfig.applicationBaseUrl + "/about/mission")
     } else if (request.path == "/" && request.userIdOpt.isEmpty) {
       landingPage(request)
+    } else if (userAgentOpt.exists(_.isMobile) && request.queryString.get("kma").exists(_.contains("1"))) {
+      val uriNoProto = applicationConfig.applicationBaseUrl.replaceFirst("https?:", "") + request.uri
```
joshm1, line 63: Should we want to remove the baseUrl from the URI call for mobile and the urls can be like `kifi:${request.uri}` (e.g. `kifi:/invite=?friend=...`)

PR #11671, opened 2014-10-21
a bit more robust

raymondng:
Add try-catch; fallback to secure social
@andrewconner @eishay @2is10

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
+++ b/kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
@@ -112,14 +112,19 @@ trait SecureSocialHelper extends Logging {
   }
 }
 
-trait UserActionsHelper extends UserActionsRequirements {
+trait UserActionsHelper extends UserActionsRequirements with Logging {
 
   protected def getUserIdFromRequest(implicit request: Request[_]): Option[Id[User]] = {
     request.session.get(KifiSession.FORTYTWO_USER_ID).map(id => Id[User](id.toLong))
   }
 
   def getUserIdOpt(implicit request: Request[_]): Future[Option[Id[User]]] = {
-    getUserIdFromRequest match {
+    val kifiIdOpt = try { getUserIdFromRequest } catch {
+      case t: Throwable =>
+        log.error(s"[getUserIdFromRequest] Caught exception $t while retrieving userId from request; cause=${t.getCause}", t)
+        None
+    }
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
```diff
@@ -112,14 +112,19 @@ trait SecureSocialHelper extends Logging {
...
-trait UserActionsHelper extends UserActionsRequirements {
+trait UserActionsHelper extends UserActionsRequirements with Logging {
 
   protected def getUserIdFromRequest(implicit request: Request[_]): Option[Id[User]] = {
     request.session.get(KifiSession.FORTYTWO_USER_ID).map(id => Id[User](id.toLong))
   }
 
   def getUserIdOpt(implicit request: Request[_]): Future[Option[Id[User]]] = {
-    getUserIdFromRequest match {
+    val kifiIdOpt = try { getUserIdFromRequest } catch {
```
andrewconner, line 122: I assume this is to protect against `FORTYTWO_USER_ID` not being a long? What about just returning `None` from `getUserIdFromRequest`?

kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
```diff
@@ -112,14 +112,19 @@ trait SecureSocialHelper extends Logging {
...
+      case t: Throwable =>
+        log.error(s"[getUserIdFromRequest] Caught exception $t while retrieving userId from request; cause=${t.getCause}", t)
+        None
```
andrewconner, line 125: This should Airbrake. It's quite serious and we don't want to not know it's happening.

raymondng, line 125: Agreed. Airbrake is not yet available in these helpers -- it's on the TODO list.

andrewconner, line 125: Hm, that's unfortunate.

kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
raymondng, line 122: Yes that's also a possibility that I considered, but that'd hide it from the caller. Airbrake would have been indeed handy but it's not available (yet). To make it available, I'll need to rework things a bit.

PR #11674, opened 2014-10-21
optimization and hardening

ymatsuda:
@yingjieMiao @eishay

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/common/zookeeper/ServiceDiscovery.scala
+++ b/kifi-backend/modules/common/app/com/keepit/common/zookeeper/ServiceDiscovery.scala
@@ -53,25 +53,25 @@ class ServiceDiscoveryImpl(
   servicesToListenOn: Set[ServiceType])
     extends ServiceDiscovery with Logging {
 
-  @volatile private var lastStatusChangeTime = System.currentTimeMillis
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/common/app/com/keepit/common/zookeeper/ServiceDiscovery.scala
```diff
@@ -146,17 +146,20 @@ class ServiceDiscoveryImpl(
...
-  @volatile private var forceUpdateInProgress = false
+  @volatile private[this] var forceUpdateInProgress = false
   def forceUpdate(): Unit = if (!forceUpdateInProgress) synchronized {
     if (!forceUpdateInProgress) {
       forceUpdateInProgress = true
-      zkClient.session { zk =>
-        for (cluster <- clusters.values) {
-          val children = zk.getChildren(cluster.servicePath).map(child => (child, zk.getData[String](child).get))
-          cluster.update(zk, children)
+      try {
```
ymatsuda, line 153: added try ~ finally otherwise forceUpdateInProgress will never set to false once zkClient.session{ ... } failed and threw an exception.

yingjieMiao, line 153: :+1:

yingjieMiao:
lgtm

PR #11676, opened 2014-10-21
replace println by log

yingjieMiao:
@ymatsuda @eishay   FYI


eishay:
We had `println` since at that time the logging system may not be initialized and these messages will not be shown

yingjieMiao:
huh, so let me revert?

PR #11679, opened 2014-10-21
Add keep-to-library widget to keep card.

yipingliao:
@joshm1 

Adding the keep-to-library widget to the master keep button on the keep card.

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/keep/addKeep.js
+++ b/kifi-backend/modules/shoebox/angular/src/keep/addKeep.js
@@ -145,6 +145,7 @@ angular.module('kifi')
             return lib.access !== 'read_only';
           });
           scope.librarySelection = {};
+          scope.librarySelection.library = _.find(scope.libraries, { 'kind': 'system_main' });
           scope.libSelectTopOffset = 110;
         }
 
--- a/kifi-backend/modules/shoebox/angular/src/keep/keep.js
+++ b/kifi-backend/modules/shoebox/angular/src/keep/keep.js
@@ -42,6 +42,7 @@ angular.module('kifi')
       restrict: 'A',
       scope: {
         keep: '=',
+        libraries: '=',
         editMode: '=',
         toggleSelect: '&',
         isSelected: '&',
@@ -884,13 +885,15 @@ angular.module('kifi')
   }
 ])
 
-.directive('kfKeepMasterButton', [
-  '$log',
-  function ($log) {
+.directive('kfKeepMasterButton', ['keepActionService', 'keepDecoratorService', 'libraryService', 'tagService',
+  function (keepActionService, keepDecoratorService, libraryService, tagService) {
     return {
       restrict: 'A',
       scope: {
-        keep: '='
+        keep: '=',
+
+        // This is the libraries set on the scope of kfKeeps (see 'keeps.js').
+        libraries: '='
       },
       replace: false,
       templateUrl: 'keep/keepMasterButton.tpl.html',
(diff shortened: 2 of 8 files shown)


kifi-backend/modules/shoebox/angular/src/keep/addKeep.js
```diff
@@ -145,6 +145,7 @@ angular.module('kifi')
             return lib.access !== 'read_only';
           });
           scope.librarySelection = {};
+          scope.librarySelection.library = _.find(scope.libraries, { 'kind': 'system_main' });
```
yipingliao, line 148: This line used to be in the widget code, but it doesn't work that well that way, so I am refactoring it back out. Not directly related to keep card.

kifi-backend/modules/shoebox/angular/src/keep/keepToLibraryWidget.js
```diff
@@ -278,12 +278,6 @@ angular.module('kifi')
 
 
         //
-        // On link.
-        //
-        scope.librarySelection.library = _.find(scope.libraries, { 'kind': 'system_main' });
```
yipingliao, line 283: Refactoring stuff.

kifi-backend/modules/shoebox/angular/src/keeps/keeps.js
```diff
@@ -124,9 +124,6 @@ angular.module('kifi')
         // 'selection' keeps track of which keeps have been selected.
         scope.selection = new selectionService.Selection();
 
-        // Not used in template right now but will be when we hide bulk privacy changes.
```
yipingliao, line 127: This is just clean up. This scope variable is set elsewhere by a watch.

kifi-backend/modules/shoebox/angular/src/keeps/keeps.js
```diff
@@ -500,6 +497,7 @@ angular.module('kifi')
                 });
 
                 scope.librarySelection = {};
+                scope.librarySelection.library = _.find(scope.libraries, { 'kind': 'system_main' });
```
yipingliao, line 500: Again, refactoring.

kifi-backend/modules/shoebox/angular/src/tagManage/tagManage.js
```diff
@@ -78,6 +78,7 @@ angular.module('kifi')
             return lib.access !== 'read_only';
           });
           $scope.librarySelection = {};
+          $scope.librarySelection.library = _.find($scope.libraries, { 'kind': 'system_main' });
```
yipingliao, line 81: Refactoring.

kifi-backend/modules/shoebox/angular/src/keep/keep.js
```diff
@@ -900,11 +903,31 @@ angular.module('kifi')
         scope.isKeptPrivate = keep.isMyBookmark && keep.isPrivate;
         scope.isKeptPublic = keep.isMyBookmark && !keep.isPrivate;
         scope.isNotKept = !keep.isMyBookmark;
-        $log.info(keep);
 
-        scope.keepAction = function () {
-          // TODO(josh)
+        scope.librarySelection = {};
```
yipingliao, line 907: The widget expects the user's libraries to be on `scope.libraries`. Then, the library that the user selects in the widget is set in `scope.librarySelection`, and the widget calls `scope.clickAction`. In this case, selecting a library in the widget results in `scope.keep` being kept to the selected library.

joshm1:
lgtm, thanks!

PR #11682, opened 2014-10-21
Cookie check

raymondng:
Migrated 2 old endpoints to new actions
Decoupling fallback logic while setting userId
Very simple cookie sanity check in unit test (`PasswordTest` as it does explicit u/p authentication)
Obvious gap is testing with `SecureSocial` and FB/LNKD ... need to look into it a bit more

@andrewconner @eishay @2is10

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
+++ b/kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
@@ -191,16 +191,18 @@ trait UserActions extends Logging { self: Controller =>
   }
 
   private def maybeSetUserIdInSession[A](userId: Id[User], res: Result)(implicit request: Request[A]): Result = {
-    userActionsHelper.getUserIdFromSession match {
-      case Success(userIdOpt) =>
-        userIdOpt match {
-          case Some(id) if id == userId => res
-          case _ => res.withSession(request.session + (KifiSession.FORTYTWO_USER_ID -> userId.toString))
-        }
-      case Failure(t) =>
-        log.error(s"[maybeSetUserIdInSession($userId)] Caught exception while retrieving userId from kifi cookie", t)
-        res.withSession(request.session + (KifiSession.FORTYTWO_USER_ID -> userId.toString))
-    }
+    Play.maybeApplication.map { app =>
+      userActionsHelper.getUserIdFromSession match {
+        case Success(userIdOpt) =>
+          userIdOpt match {
+            case Some(id) if id == userId => res
+            case _ => res.withSession(request.session + (KifiSession.FORTYTWO_USER_ID -> userId.toString))
+          }
+        case Failure(t) =>
+          log.error(s"[maybeSetUserIdInSession($userId)] Caught exception while retrieving userId from kifi cookie", t)
+          res.withSession(request.session + (KifiSession.FORTYTWO_USER_ID -> userId.toString))
+      }
+    } getOrElse res
   }
 
   private def impersonate[A](adminUserId: Id[User], impersonateExtId: ExternalId[User])(implicit request: Request[A]): Future[UserRequest[A]] = {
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
```diff
@@ -185,13 +190,19 @@ trait UserActions extends Logging { self: Controller =>
...
-        case Some(id) if id == userId => res
-        case _ => res.withSession(request.session + (KifiSession.FORTYTWO_USER_ID -> userId.toString))
+  private def maybeSetUserIdInSession[A](userId: Id[User], res: Result)(implicit request: Request[A]): Result = {
+    Play.maybeApplication.map { app =>
+      userActionsHelper.getUserIdFromSession match {
+        case Success(userIdOpt) =>
+          userIdOpt match {
+            case Some(id) if id == userId => res
+            case _ => res.withSession(request.session + (KifiSession.FORTYTWO_USER_ID -> userId.toString))
+          }
```
2is10: probably cleaner to collapse these two `match` blocks into one

``` scala
case Success(Some(id)) if id == userId => ...
case Success(_) => ...
case Failure(t) => ...
```

raymondng: Thanks. I actually started with that but settled on the current one. Not married to it though -- let me take another look.

kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
```diff
@@ -116,14 +117,18 @@ trait SecureSocialHelper extends Logging {
...
+      Play.maybeApplication.flatMap { _ => request.session.get(KifiSession.FORTYTWO_USER_ID).map(id => Id[User](id.toLong)) }
+    }
+
+  def getUserIdOptWithFallback(implicit request: Request[_]): Future[Option[Id[User]]] = {
+    val kifiIdOpt = getUserIdFromSession recover {
       case t: Throwable =>
         airbrake.notify(s"[getUserIdOpt] Caught exception $t while retrieving userId from request; cause=${t.getCause}", t)
         None
-    }
+    } get
```
2is10, line 130: This `get` is a `None.get` in the `Failure` case, correct? Then why not just rethrow the original exception? The `recover` is a bit misleading.

raymondng, line 130: Actually, it's a `Try.get` which results in `None` in the failure case.

2is10, line 130: Oh, `recover` wraps the `None` in a `Success`, so it’s a `Success(None).get`. I see. A little confusing.

raymondng, line 130: Yes I had to look it up as well. I find `map for exception` helpful.

2is10, line 130: I think the method that returns `Try[Option[...]]` is the source of the complexity. It seems like a parse error in the cookie is a concern that could be handled locally, rather than propagating it up a level to the caller. Both callers simply recover, one after airbraking and one after logging. It’s not immediately obvious why they need to react in different ways.

Also, it appears that `getUserIdOptWithFallback` can always return synchronously. Why use `Future` in the return type?

I guess I’m not seeing the payoff for having two methods that essentially do the same thing with different return types here.

2is10, line 130: If you’re happy with this change, feel free to merge. I’m done commenting. Just some feedback you might consider.

raymondng, line 130: Thanks for the feedback!

Agreed it's a bit complex due to error handling. I'll look into simplifying it.

raymondng:
Will merge first and build upon it. Feedback appreciated.

PR #11683, opened 2014-10-21
bug fix in leaderPriority

yingjieMiao:
@ymatsuda  @raymondng


ymatsuda:
lgtm

raymondng:
Should we also check whether the instance is healthy or not?

yingjieMiao:
it has a very strict filter on leader (only accept Up status). Selection of other instances uses various heuristics.

PR #11684, opened 2014-10-21
library name with ellipsis

ahsu1230:
@andrewconner @yipingliao @2is10 

Javascript implementation (ellipsis directive) that can truncate long text
Would probably need to add to it in the future, so we can use this directive in other places


kifi-backend/modules/shoebox/angular/src/common/directives/ellipsis/ellipsis.js
```diff
@@ -0,0 +1,124 @@
...
+          * Execute ellipsis truncate on ngBind update
+          */
+          scope.$watch('ellipsisAppend', function () {
+            buildEllipsis();
+          });
+
+           /**
+          * When window width or height changes - re-init truncation
+          */
+          angular.element($window).bind('resize', function () {
```
yipingliao, line 107: There is a lot of '$window.addEventListener' in the code base. Saves the jQuery wrapping. Also, you may want to debounce on resizing.

kifi-backend/modules/shoebox/angular/src/common/directives/ellipsis/ellipsis.js
```diff
@@ -0,0 +1,124 @@
...
+    scope   : {
+      ngBind        : '=',
+      ellipsisAppend    : '@',
+      ellipsisSymbol    : '@'
+    },
+    compile : function(/*elem, attr, linker*/) {
+
+      return function(scope, element, attributes) {
+        /* Window Resize Variables */
+          attributes.lastWindowResizeTime = 0;
```
yipingliao, line 30: These seem to be internal variables? Why are they hung on `attributes`?

kifi-backend/modules/shoebox/angular/src/common/directives/ellipsis/ellipsis.js
```diff
@@ -0,0 +1,124 @@
...
+          });
+
+           /**
+          * When window width or height changes - re-init truncation
+          */
+          angular.element($window).bind('resize', function () {
+            $timeout.cancel(attributes.lastWindowTimeoutEvent);
+
+            attributes.lastWindowTimeoutEvent = $timeout(function() {
+              if (attributes.lastWindowResizeWidth !== window.innerWidth || attributes.lastWindowResizeHeight !== window.innerHeight) {
```
yipingliao, line 111: Replace `window` with `$window`.

yipingliao, line 111: If we're already resizing, why do we need to check that the current width and height are indeed different from the previous width and height?

kifi-backend/modules/shoebox/angular/src/common/directives/ellipsis/ellipsis.js
```diff
@@ -0,0 +1,124 @@
...
+                buildEllipsis();
+              }
+
+              attributes.lastWindowResizeWidth = window.innerWidth;
+              attributes.lastWindowResizeHeight = window.innerHeight;
+            }, 75);
+          });
+
```
yipingliao, line 119: Too many newlines.

kifi-backend/modules/shoebox/angular/src/common/directives/ellipsis/ellipsis.js
```diff
@@ -0,0 +1,124 @@
...
+           /**
+          * Execute ellipsis truncate on ngBind update
+          */
+          scope.$watch('ellipsisAppend', function () {
+            buildEllipsis();
+          });
+
+           /**
```
yipingliao, line 104: Weird indent (here and elsewhere).

kifi-backend/modules/shoebox/angular/src/common/directives/ellipsis/ellipsis.js
```diff
@@ -0,0 +1,124 @@
...
+            var currentHeight = element.height();
+            var currentNumLines = currentHeight / heightPerLine;
+            var maxIndex = scope.ngBind.length;
+
+            if (currentNumLines <= 2) { // entire name fits
+              return;
+            }
+
+            var ellipsisSymbol =
+              (typeof(attributes.ellipsisSymbol) !== 'undefined') ?
```
yipingliao, line 48: Will a check for falsy suffice here?

kifi-backend/modules/shoebox/angular/src/common/directives/ellipsis/ellipsis.js
```diff
@@ -0,0 +1,124 @@
...
+                attributes.ellipsisSymbol : '&hellip;';
+            var appendString =
+              (typeof(scope.ellipsisAppend) !== 'undefined' && scope.ellipsisAppend !== '') ?
```
yipingliao, line 51: Same: will a check for falsy suffice here?

andrewconner:
Could you make the height dynamic instead of hard coded? This breaks for public libraries, since the font is larger.

kifi-backend/modules/shoebox/angular/src/common/directives/ellipsis/ellipsis.js
```diff
@@ -0,0 +1,124 @@
...
+            if (numIntervals < 3) {
+              $timeout(interval, 1000);
+            }
+          }
+          $timeout(interval, 200);
+
+           /**
+          * Execute ellipsis truncate on ngBind update
+          */
+          scope.$watch('ngBind', function () {
```
andrewconner, line 93: as mentioned, this is the same as `scope.$watch('ngBind', buildEllipsis)`

kifi-backend/modules/shoebox/angular/src/common/directives/ellipsis/ellipsis.js
```diff
@@ -0,0 +1,124 @@
...
+            var currentNumLines = currentHeight / heightPerLine;
+            var maxIndex = scope.ngBind.length;
+
+            if (currentNumLines <= 2) { // entire name fits
+              return;
+            }
+
+            var ellipsisSymbol =
+              (typeof(attributes.ellipsisSymbol) !== 'undefined') ?
+                attributes.ellipsisSymbol : '&hellip;';
```
yipingliao, line 49: Add newline after `:` in ternary operator.

kifi-backend/modules/shoebox/angular/src/common/directives/ellipsis/ellipsis.js
```diff
@@ -0,0 +1,124 @@
...
+            var appendString =
+              (typeof(scope.ellipsisAppend) !== 'undefined' && scope.ellipsisAppend !== '') ?
+                ellipsisSymbol + '<span>' + scope.ellipsisAppend + '</span>' : ellipsisSymbol;
```
yipingliao, line 52: Add newline after `:`.

kifi-backend/modules/shoebox/angular/src/common/directives/ellipsis/ellipsis.js
```diff
@@ -0,0 +1,124 @@
...
+
+            // binary search for correct maxIndex
+            var hi = scope.ngBind.length;
+            var lo = scope.ngBind.length / 2;
```
yipingliao, line 56: Based on just reading the code, if the max number of chars we can fit within 2 lines is less than half of the text length, then we still display half of the text. Please correct if I am misunderstanding!

kifi-backend/modules/shoebox/angular/src/common/directives/ellipsis/ellipsis.js
```diff
@@ -0,0 +1,124 @@
...
+ *  how to use:
+ *  <p data-ellipsis data-ng-bind="longName"></p>
+ *  <p data-ellipsis data-ng-bind="longName" data-ellipsis-symbol="---"></p>
+ *  <p data-ellipsis data-ng-bind="longName" data-ellipsis-append="read more"></p>
+ *
+ */
+'use strict';
+angular.module('kifi')
+
+.directive('kfEllipsis', ['$timeout', '$window', function($timeout, $window) {
```
yipingliao, line 17: Add space after `function`.

kifi-backend/modules/shoebox/angular/src/common/directives/ellipsis/ellipsis.js
```diff
@@ -0,0 +1,124 @@
...
+
+  return {
+    restrict  : 'A',
```
yipingliao, line 20: No space after `:` in JavaScript objects (here and below).

kifi-backend/modules/shoebox/angular/src/common/directives/ellipsis/ellipsis.js
```diff
@@ -0,0 +1,124 @@
+/**
+ *  Angular directive to truncate multi-line text to visible height
+ *  inputs:
+ *  (required) ng-bind
```
yipingliao, line 4: Can you explain to me (maybe in person) why `ng-bind` is being used? `ngBind` is a native Angular directive, so I am wondering about how we're using its special powers here.

kifi-backend/modules/shoebox/angular/src/common/directives/ellipsis/ellipsis.js
```diff
@@ -0,0 +1,124 @@
...
+.directive('kfEllipsis', ['$timeout', '$window', function($timeout, $window) {
+
+  return {
+    restrict  : 'A',
+    scope   : {
+      ngBind        : '=',
+      ellipsisAppend    : '@',
+      ellipsisSymbol    : '@'
+    },
+    compile : function(/*elem, attr, linker*/) {
```
yipingliao, line 26: Maybe using `ngBind` is related to using `compile`? I don't quite understand how and when `compile` should be used, but would like to learn. :)

kifi-backend/modules/shoebox/angular/src/common/directives/ellipsis/ellipsis.js
ahsu1230, line 56: yeah you're right. I had the assumption that ellipsis would come at the end of a library name.
But since this is going to be a general directive, we shouldn't assume that.

kifi-backend/modules/shoebox/angular/src/common/directives/ellipsis/ellipsis.js
```diff
@@ -0,0 +1,124 @@
...
+
+        function buildEllipsis() {
+          if (typeof(scope.ngBind) !== 'undefined') {
+            element.html(scope.ngBind);
+            var heightPerLine = 30;
+            var currentHeight = element.height();
+            var currentNumLines = currentHeight / heightPerLine;
+            var maxIndex = scope.ngBind.length;
+
+            if (currentNumLines <= 2) { // entire name fits
```
yipingliao, line 43: Oops, one more thing: I just realized that we will need this directive to truncate keep titles. I think Mark said to 3 lines. Can we parameterize the max number of lines (defaulting to 2, perhaps?) by adding it as another optional scope variable, like `ellipsisSymbol`?

andrewconner, line 43: +1, this should be as generic as possible.

kifi-backend/modules/shoebox/angular/src/common/directives/ellipsis/ellipsis.js
2is10, line 111: I think it’s good to check to potentially avoid costly work. When the `resize` events are properly debounced, there’s no guarantee the window will be a different size than the last time this code executed.

kifi-backend/modules/shoebox/angular/src/common/directives/ellipsis/ellipsis.js
2is10, line 107: Definitely debounce or throttle. I’d probably throttle to every 200ms, at least.

kifi-backend/modules/shoebox/angular/src/common/directives/ellipsis/ellipsis.js
```diff
@@ -0,0 +1,124 @@
...
+          // todo (aaron): find something better than this hack
+          var numIntervals = 0;
+          function interval() {
+            buildEllipsis();
+            numIntervals++;
+            if (numIntervals < 3) {
+              $timeout(interval, 1000);
+            }
+          }
+          $timeout(interval, 200);
```
2is10, line 88: The word “interval” has a very specific meaning in browser JavaScript (see [setInterval](https://developer.mozilla.org/en-US/docs/Web/API/WindowTimers.setInterval), clearInterval). It’s a little confusing to see “interval” used in names in conjunction with timeouts.

kifi-backend/modules/shoebox/angular/src/common/directives/ellipsis/ellipsis.js
2is10, line 52: Might consider just having one optional parameter for the HTML string to be appended. Let the caller embed whatever tags/classes/text/symbols desired, in any order. It’d be both simpler and more flexible.

kifi-backend/modules/shoebox/angular/src/common/directives/ellipsis/ellipsis.js
```diff
@@ -0,0 +1,124 @@
...
+          }
+        }
+
+         /**
+        * Watchers
+        */
+          // font-changes will affect sizing, but ng-bind applies first!
+          // todo (aaron): find something better than this hack
+          var numIntervals = 0;
+          function interval() {
```
2is10, line 81: This function’s name doesn’t need to be scoped to the entire enclosing function. It could be a named function expression, which limits the scope of the function’s name to its own body:

``` js
var numBuilds = 0;
$timeout(function build() {
  buildEllipsis();
  if (++numBuilds < 3) {
    $timeout(build, 1000);
  }  
}, 200);
```

2is10:
Very nice work, @ahsu1230! A tricky problem.

PR #11686, opened 2014-10-21
trying to resolve java.net.URISyntaxException (scraper)

ymatsuda:
Some of the annoying java.net.URISyntaxException are caused by 8 bit control characters in URL.
URISanitizer was not encoding 8 bit control characters properly.

@eishay @raymondng


raymondng:
lgtm

PR #11687, opened 2014-10-21
Kill send unread messages

eishay:
@joshm1


kifi-backend/modules/shoebox/app/views/email/libraryInvitation.scala.html
```diff
@@ -5,7 +5,7 @@
   passPhrase: String)
 @import com.keepit.common.mail.template.helpers._
 @libraryAssetUrl(path: String) = @{assetUrl("library/" + path)}
-@libraryUrl = {@{baseUrl}@libraryInfo.url}
+@libraryUrl = {@{baseUrl}@{libraryInfo.url}@{KifiMobileAppLinkFlag.arg}}
```
joshm1, line 8: we should make this a function that returns a accepts a `content` argument and calls `htmlUrl(libraryInfo.url, content, true)` so it will include other tracking params as well and then call it anywhere `libraryUrl` current is used in this view

kifi-backend/modules/common/app/com/keepit/common/mail/package.scala
```diff
@@ -6,6 +6,12 @@ import com.keepit.heimdal.HeimdalContext
 import com.keepit.model.{ Library, User }
 import play.twirl.api.Html
 
+object KifiMobileAppLinkFlag {
+  val key = "kma"
+  val value = "1"
+  val arg = s"&$key=$value"
```
joshm1, line 12: one problem with hard-coding the '&' here is that we assume the URL you want to append this to always has another query parameter before it: `url?query=value${KifiMobileAppLinkFlag.arg}` .  I think we should require the callers to put the '?' or '&' there

eishay, line 12: fixed

PR #11688, opened 2014-10-21
Hooking up library attribution for recos

stkem:
@yingjieMiao 
@yipingliao @joshm1


stkem:
Note that this is still gated by the library experiment in the FE

yingjieMiao:
lgtm

PR #11690, opened 2014-10-21
Style tweaks

andrewconner:
@yipingliao


kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.js
```diff
@@ -208,7 +208,7 @@ angular.module('kifi')
             'description': '',
             'slug': '',
 
-            // By default, the create library form selects the "discoverable" visibility for a new library.
+            // By default, the create library form selects the "published" visibility for a new library.
```
yipingliao, line 211: Thank you.

yipingliao:
Looks good!

PR #11692, opened 2014-10-21
activates reco computation when following a lib (without own keeps) + better empty state test libs

stkem:
@yingjieMiao

### DIFF
--- a/kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
+++ b/kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
@@ -117,7 +117,7 @@ class RecommendationGenerationCommander @Inject() (
   }
 
   private def shouldInclude(scores: UriScores): Boolean = {
-    if ((scores.overallInterestScore > 0.4 || scores.recentInterestScore > 0 || scores.libraryInducedScore.isDefined) && computeMasterScore(scores) > 6) {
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
```diff
@@ -125,6 +125,7 @@ class RecommendationGenerationCommander @Inject() (
         scores.recentInterestScore > 0.15 ||
         scores.rekeepScore > 0.3 ||
         scores.discoveryScore > 0.3 ||
+        scores.libraryInducedScore.isDefined ||
         (scores.curationScore.isDefined && (scores.overallInterestScore > 0.5 || scores.recentInterestScore > 0.2))
     } else { //Yes, this could be expressed purly with a logic expression, but I think this is clearer -Stephen
```
yingjieMiao, line 130: wah, the decision tree has grown quite a bit, a little hard to visualize now.

kifi-backend/modules/shoebox/app/com/keepit/commanders/RecommendationsCommander.scala
```diff
@@ -205,8 +205,10 @@ class RecommendationsCommander @Inject() (
   }
 
   def topPublicLibraryRecos(userId: Id[User]): Future[Seq[FullRecoInfo]] = {
+    val curatedLibs: Seq[Id[Library]] = Seq(25957L, 25537L, 25353L, 25368L, 25171L, 25116L, 25341L).map(Id[Library])
+
```
yingjieMiao, line 209: I thought we had more. Anyway, a good start.

PR #11693, opened 2014-10-22
privatize FORTYTWO_USER_ID

raymondng:
Made `FORTYTWO_USER_ID` private
Added implicit class with type-safe operations on `userId`
DRY-er code

Relying on `toString` is unsafe/dangerous ... this should help prevent/eliminate this type of errors.

@andrewconner @eishay @2is10


andrewconner:
Works for me. Are we also going to (locally, in Play tests) test normal user flows? i.e., register 1st step API endpoint → register 2nd step → check cookies?

raymondng:
Yup that is my goal.

PR #11695, opened 2014-10-22
After user edits a slug, changes to the library name input no longer affects the slug

yipingliao:
Also:
- Do not squash dashes in the slug
- Do not update slug for library name change on existing libraries.
  https://app.asana.com/0/15523651812135/18355508515843

@andrewconner Should we do validation on slug? Currently we only validate name.


kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.js
```diff
@@ -23,6 +23,7 @@ angular.module('kifi')
         //
         scope.username = profileService.me.username;
         scope.userHasEditedSlug = false;
+        scope.emptySlug = true;
```
yipingliao, line 26: Previously, there was a variable called `scope.userHasEditedSlug` (see line 25), but that was a misnomer because it actually represented whether the slug was empty and was used for styling only (adding a dotted border under "new library" in the slug field). I've added `scope.emptySlug` to replace that, and now `scope.userHasEditedSlug` properly reflects what its name says it does, and is used to determine whether library name changes should update the library slug.

kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.styl
```diff
@@ -121,10 +121,10 @@
 
     .manage-lib-slug-input
       border-bottom: 1px dotted #c7c7c7
-      width: 70px
+      width: 75px
```
yipingliao, line 124: This is the width on the border under "new-library" in the slug field, and needed to be increased for the recent font changes.

kifi-backend/modules/shoebox/angular/src/common/util.js
```diff
@@ -91,7 +91,7 @@ angular.module('util', [])
...
         // .replace(nonLetters, '-').replace(/-(?=-)/g, '').toLowerCase();
-        return name.toLowerCase().replace(/[^\w\s]|_/g, '').replace(/\s+/g, '-').replace(/^-/, '');
+        return name.toLowerCase().replace(/[^\w\s-]|_/g, '').replace(/\s+/g, '-').replace(/^-/, '');
```
yipingliao, line 90: Why this works is a mystery to me! Previously, this first regex will squash all dashes because dashes are not in the `\w` set. I didn't think that adding the `-` inside this group will work because I thought the logic might short-circuit on `\w`. But this works in practice! I am explaining it to myself as `[^xyz]` is being expanded out logically into `!x && !y && !z` during the regex matching, giving the last dash a chance to not match. Super fragile?

andrewconner:
Thanks

PR #11701, opened 2014-10-22
open messages with the app if the recipient is a user and the app exist;

eishay:
cleanup unused emails
Also:
- Assume we work on all iOS platforms (ipad included)
- open /email/verified on email verification


eishay:
@andrewconner @joshm1

joshm1:
:+1:

PR #11707, opened 2014-10-22
fix user status tracking

yingjieMiao:
@joshm1 @stkem

### DIFF
--- a/kifi-backend/modules/common/test/com/keepit/common/concurrent/WatchableExecutionContext.scala
+++ b/kifi-backend/modules/common/test/com/keepit/common/concurrent/WatchableExecutionContext.scala
@@ -7,6 +7,7 @@ import scala.concurrent.{ ExecutionContext => ScalaExecutionContext }
 class WatchableExecutionContext extends ScalaExecutionContext {
   private[this] val internalContext: ScalaExecutionContext = scala.concurrent.ExecutionContext.fromExecutorService(Executors.newFixedThreadPool(2))
   private[this] val lock = new Object()
+  private[this] var addedAnything = false
   private[this] var counter: Int = 0
   private[this] var maxExeutionCount: Int = 0
 
@@ -15,6 +16,7 @@ class WatchableExecutionContext extends ScalaExecutionContext {
   def execute(runnable: Runnable): Unit = lock.synchronized {
     counter += 1
     maxExeutionCount += 1
+    addedAnything = true
     val wrapper = new Runnable {
       override def run(): Unit = {
         runnable.run()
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/shoebox/test/com/keepit/commanders/SendgridCommanderTest.scala
```diff
@@ -102,6 +102,7 @@ class SendgridCommanderTest extends Specification with ShoeboxTestInjector {
             event = Some(SendgridEventTypes.CLICK))
 
           commander.processNewEvents(Seq(sgEvent))
+          Thread.sleep(500)
```
joshm1: can we either use a single thread execution context or update the method(s) to return a Future so we don't do this?

stkem: Yeah, check out `FakeExecutionContextModule.scala`. It will let you inject a `WatchableExecutionContext` which will let you wait until it's done (instead of an arbitrary amount of time). If this turn out to be a major pain to integrate right now, leave it as is for this week and file in Asana for next.

yingjieMiao:
@joshm1 @stkem @eishay   updated test with `WatchableExecutionContext`.

@eishay : @joshm1  and I suspect there is a bug/feature with `WatchableExecutionContext`. `drain()` not necessarily wait for all jobs to complete. In fact, if none of the job is added yet, drain will just pass through.

kifi-backend/modules/common/test/com/keepit/common/concurrent/WatchableExecutionContext.scala
```diff
@@ -30,7 +32,7 @@ class WatchableExecutionContext extends ScalaExecutionContext {
 
   def drain(waitTime: Long = 1000): Int = {
     lock.synchronized {
-      while (counter > 0) {
+      if (!addedAnything || (addedAnything && counter > 0)) {
```
stkem, line 35: I think you still need the `while` here, to actually wait till it's empty.

yingjieMiao, line 35: it's notified only when `counter` goes down to zero, right?

yingjieMiao, line 35: As discussed in person, this doesn't eliminate all race conditions, e.g. (counter goes down to 0 when there are new runnables adding to the context). This should be OK for test purposes.

PR #11711, opened 2014-10-22
Style tweaks

andrewconner:
More small changes
- Title (as in, `<title>`) is set for libraries now. SEO!
- Bolding selected title, fixing active bug
- Making slimmer library user pages look better

@yipingliao @ahsu1230


ahsu1230:
:+1:

kifi-backend/modules/shoebox/angular/src/libraries/library.js
```diff
@@ -99,6 +99,10 @@ angular.module('kifi')
       }
     });
 
+    var setTitle = function (lib) {
+      $window.document.title = lib.name + ' by ' + lib.owner.firstName + ' ' + lib.owner.lastName + ' • Kifi' ;
```
yipingliao, line 103: Yay!

yipingliao:
:+1:

PR #11715, opened 2014-10-22
Implement invite header for library card.

yipingliao:
https://app.asana.com/0/15523651812135/18412051444297

@andrewconner @ahsu1230 

Note that although calls to the "decline" endpoint succeed (200, with "success" in the response), when I reload the page the declined library still appears in my list of invited libraries. Not sure if this is intended behavior.


ahsu1230:
I looked at the backend - it's a backend bug. When we get library & invite summaries, we forget to exclude `DECLINED` invites. Will fix it in next PR =)

ahsu1230:
Thanks for this! It looked great :cake:

andrewconner:
Great, awesome

yipingliao:
Thanks for looking into the backend, @ahsu1230 :)

PR #11716, opened 2014-10-22
manage tags support moving keeps

ahsu1230:
- fixed bug with ellipsis... truncating too early
- changed font color of right column to a new blue
- and fixed bug of header search bar having chip when going to profile
- get library/invite summaries excludes library invites with DECLINED state

@andrewconner @yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/layout/rightCol/footer.styl
+++ b/kifi-backend/modules/shoebox/angular/src/layout/rightCol/footer.styl
@@ -17,9 +17,6 @@
   .sprite-import-from-other-services-icon
     margin-right: 12px
 
-  a
-    color: #1990ea
-
   a:hover
     text-decoration: underline
 
@@ -62,6 +59,9 @@
   vertical-align: top
   padding: 5px 5px 5px 5px
 
+  a
+    color: linkBlueColor
+
 .kf-footer-help-rank-number
   font-size: 28px
   margin-right: 14px
@@ -80,6 +80,9 @@
 
 .kf-footer-more-list
   line-height: 32px
+
+  a
+    color: linkBlueColor
 
 .kf-footer-item-link
   display: inline-block
--- a/kifi-backend/modules/shoebox/angular/src/layout/rightCol/footer.styl
+++ b/kifi-backend/modules/shoebox/angular/src/layout/rightCol/footer.styl
@@ -17,6 +17,9 @@
   .sprite-import-from-other-services-icon
     margin-right: 12px
 
+  a
+    color: #1990ea
+
   a:hover
     text-decoration: underline
 
--- a/kifi-backend/modules/shoebox/angular/src/tagManage/tagManage.js
+++ b/kifi-backend/modules/shoebox/angular/src/tagManage/tagManage.js
@@ -83,6 +83,18 @@ angular.module('kifi')
       }
     });
 
+    $scope.$watch(function() {
+      return $window.innerWidth;
+    }, function (width) {
+      if (width < 1220) {
+        $scope.copyToLibraryDisplay = 'Copy';
+        $scope.moveToLibraryDisplay = 'Move';
+      } else {
+        $scope.copyToLibraryDisplay = 'Copy to Library';
+        $scope.moveToLibraryDisplay = 'Move to Library';
+      }
+    });
+
     $rootScope.$on('librarySummariesChanged', function () {
       $scope.libraries = _.filter(libraryService.librarySummaries, function (lib) {
         return lib.access !== 'read_only';
@@ -91,24 +103,6 @@ angular.module('kifi')
 
     $rootScope.$emit('libraryUrl', {});
 
-    $scope.clickAction = function () {
-      var tagName = encodeURIComponent($scope.selectedTag.name);
-      libraryService.copyKeepsFromTagToLibrary($scope.librarySelection.library.id, tagName).then(function () {
-        modalService.open({
-          template: 'tagManage/tagToLibModal.tpl.html',
-          modalData: { library : $scope.librarySelection.library }
-        });
-        libraryService.addToLibraryCount($scope.librarySelection.library.id, $scope.selectedTag.keeps);
-      })['catch'](function () {
-        modalService.open({
-          template: 'common/modal/genericErrorModal.tpl.html'
-        });
-      });
-    };
-
-    $scope.changeSelection = function (tag) {
-      $scope.selectedTag = tag;
-    };
 
     //
     // Filtering Stuff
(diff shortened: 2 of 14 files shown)


andrewconner:
Great thanks

kifi-backend/modules/shoebox/angular/src/layout/rightCol/footer.styl
```diff
@@ -17,6 +17,9 @@
   .sprite-import-from-other-services-icon
     margin-right: 12px
 
+  a
```
andrewconner: We try to avoid selectors like this that select an entire tag. Browsers match CSS selectors right to left, so this will be a large set.

kifi-backend/modules/shoebox/angular/src/layout/rightCol/footer.styl
```diff
@@ -17,6 +17,9 @@
...
+    color: #1990ea
```
andrewconner: Would be great to add a variable name for this color. See `constants.styl`

kifi-backend/modules/shoebox/angular/src/tagManage/tagManage.js
```diff
@@ -162,6 +156,47 @@ angular.module('kifi')
...
+          });
+          libraryService.addToLibraryCount($scope.librarySelection.library.id, $scope.selectedTag.keeps);
+        })['catch'](function () {
+          modalService.open({
+            template: 'common/modal/genericErrorModal.tpl.html'
+          });
+        });
+      } else {
+        libraryService.moveKeepsFromTagToLibrary($scope.librarySelection.library.id, tagName).then(function () {
+          modalService.open({
```
andrewconner, line 183: This code below is repeated for both cases. It's more maintainable to combine it.

One way is doing something like:

``` javascript
var tagResult;
if ($scope.actionToLibrary === 'copy') {
 tagResult = libraryService.copyKeepsFromTagToLibrary(...)
else {
  tagResult = libraryService.addToLibraryCount(...)
}
tagResult.then(function () { ... })
```

kifi-backend/modules/shoebox/angular/src/tagManage/tagManage.tpl.html
```diff
@@ -32,8 +32,11 @@
                 </ng-pluralize>
               </span>
       				<span class="manage-tag-row-links-section">
```
andrewconner: Spacing is still weird in this file. We use two spaces (no tabs).

andrewconner: Your editor can do this automatically. In Sublime, you can add this to your User prefs:

```
    "tab_size": 2,
    "translate_tabs_to_spaces": true,
```

kifi-backend/modules/shoebox/angular/src/tagManage/tagToLibMsg.tpl.html
```diff
@@ -1,6 +1,6 @@
 <div kf-modal class="kf-copy-tag-to-library-modal" kf-width="480px">
-  <div kf-basic-modal-content title="Copying Keeps" action-text="Go to Library" action="navigateToLibrary()" >
-    <div>We're copying your tagged keeps over, but it might take a second...<br>
+  <div kf-basic-modal-content title="{{action}} Keeps" action-text="Go to Library" action="navigateToLibrary()" >
+    <div>We're {{action.toLowerCase()}} your tagged keeps over, but it might take a second...<br>
```
andrewconner, line 3: Missing 'ing'?

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -425,7 +425,7 @@ class LibraryCommander @Inject() (
   def getLibrariesByUser(userId: Id[User]): (Seq[(LibraryMembership, Library)], Seq[(LibraryInvite, Library)]) = {
     db.readOnlyMaster { implicit s =>
       val myLibraries = libraryRepo.getByUser(userId)
-      val myInvites = libraryInviteRepo.getByUser(userId, Set(LibraryInviteStates.ACCEPTED, LibraryInviteStates.INACTIVE))
+      val myInvites = libraryInviteRepo.getByUser(userId, Set(LibraryInviteStates.ACCEPTED, LibraryInviteStates.INACTIVE, LibraryInviteStates.DECLINED))
```
andrewconner, line 428: Thanks! Is there a test for this case?

kifi-backend/modules/shoebox/angular/src/tagManage/tagToLibMsg.tpl.html
ahsu1230, line 3: see line 23 in `tagToLib.js` - adds 'ing' and makes sure the first letter is capitalized for the title

kifi-backend/modules/shoebox/angular/src/tagManage/tagManage.js
```diff
@@ -162,6 +156,47 @@ angular.module('kifi')
...
+    $scope.clickAction = function () {
+      var tagName = encodeURIComponent($scope.selectedTag.name);
+
+      if ($scope.actionToLibrary === 'copy') {
+        libraryService.copyKeepsFromTagToLibrary($scope.librarySelection.library.id, tagName).then(function () {
+          modalService.open({
+            template: 'tagManage/tagToLibModal.tpl.html',
+            modalData: { library : $scope.librarySelection.library, action: $scope.actionToLibrary }
+          });
+          libraryService.addToLibraryCount($scope.librarySelection.library.id, $scope.selectedTag.keeps);
```
yipingliao: What is `$scope.selectedTag.keeps`, an array of keeps or the number of keeps? If it's an array, use `keeps.length`? If it's an actual number, can we rename it to `numKeeps` instead? Thanks!

kifi-backend/modules/shoebox/angular/src/tagManage/tagManage.js
```diff
@@ -162,6 +156,47 @@ angular.module('kifi')
...
+            template: 'common/modal/genericErrorModal.tpl.html'
+          });
+        });
+      } else {
+        libraryService.moveKeepsFromTagToLibrary($scope.librarySelection.library.id, tagName).then(function () {
+          modalService.open({
+            template: 'tagManage/tagToLibModal.tpl.html',
+            modalData: { library : $scope.librarySelection.library, action: $scope.actionToLibrary }
+          });
+          libraryService.addToLibraryCount($scope.librarySelection.library.id, $scope.selectedTag.keeps);
```
yipingliao: If we're moving, are we removing the keeps from the tag? Do we need to subtract the number of keeps from the tags count? Or, maybe whenever we move tags we move all the tags so the tag is just gone (and that's why there is no subtraction here)?

yipingliao:
:+1:

kifi-backend/modules/shoebox/angular/src/tagManage/tagManage.js
ahsu1230: when we move keeps, the tags still stick with the keeps. So tag count should not be affected.
But, the backend doesn't support which libraries lose how many keeps - so I'll add it in next PR request

kifi-backend/modules/shoebox/angular/src/tagManage/tagManage.js
ahsu1230: will change the API so we get a number of keeps affected field =)

kifi-backend/modules/shoebox/angular/src/tagManage/tagManage.js
```diff
@@ -162,6 +156,44 @@ angular.module('kifi')
...
+          });
+          // todo (aaron): call addToLibraryCount accordingly (make sure source libraries do NOT lose keep counts)
+        });
+      } else {
+        tagActionResult = libraryService.moveKeepsFromTagToLibrary($scope.librarySelection.library.id, tagName).then(function () {
+          modalService.open({
+            template: 'tagManage/tagToLibModal.tpl.html',
+            modalData: { library : $scope.librarySelection.library, action: $scope.actionToLibrary }
+          });
+          // todo (aaron): call addToLibraryCount accordingly (make sure source libraries lose keep counts)
```
ahsu1230, line 187: this & above will address later today (first do backend)

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
ahsu1230, line 428: added!

kifi-backend/modules/shoebox/angular/src/tagManage/tagManage.js
andrewconner, line 187: I probably wasn't clear. The `then` block above is duplicated.

kifi-backend/modules/shoebox/angular/src/tagManage/tagToLibMsg.tpl.html
andrewconner, line 3: I see. Using `action` for both meanings confused me.

kifi-backend/modules/shoebox/angular/src/tagManage/tagManage.js
```diff
@@ -83,6 +83,18 @@ angular.module('kifi')
       }
     });
 
+    $scope.$watch(function() {
+      return $window.innerWidth;
```
andrewconner, line 87: You could instead just listen to window's resize event. However, better is probably handling this in CSS with media queries. Happy to help if you need anything.

kifi-backend/modules/shoebox/angular/src/tagManage/tagManage.js
ahsu1230, line 187: Yea, I was thinking of changing it now, but will do it once we finalize what the copy.move keeps server API will return - since copy/move keeps might return different results. See PR #11722

PR #11720, opened 2014-10-22
Style tweaks

andrewconner:


kifi-backend/modules/shoebox/angular/src/app.styl
```diff
@@ -109,13 +109,6 @@ a[href]
   .public-body
     position: relative
 
-  .body-upper-white-background
```
andrewconner, line 112: @yipingliao Removing this — we need this functionality, but it has been causing z-index issues. What we need is white to extend the entire width of the screen for the height of the library card.

kifi-backend/modules/shoebox/angular/src/layout/header/header.styl
```diff
@@ -65,7 +65,7 @@
     .kf-header-search-bar
       width: calc(100% - 380px)
 
-  .kf-header-left
+  .kf-logged-in & .kf-header-left
```
andrewconner, line 68: We should probably split the CSS for logged in and out into different classes. Fixed several issues when one is tweaked and the other breaks.

yipingliao, line 68: Sounds good.

kifi-backend/modules/shoebox/angular/src/app.styl
yipingliao, line 112: Okay. To extend to the width of the screen, will we have to have a different layout for the non-logged-in page? Library card is restricted to its width by its parent element.

andrewconner, line 112: Yeah — it's unfortunate. We'll come up with a way. We may want a separate wrapper around the library card than the keeps, and the wrapper can provide the white (and restrict the library card to be the width of the keeps).

PR #11722, opened 2014-10-22
new copy/move keeps API - returns back number of keeps moved and which l...

ahsu1230:
...ibraries they come from

@andrewconner 

The copy keeps returns 
{"successes" : 10, "failures" : "..." }

The move keeps returns
{"successes": [ { "library" : "sourceLibPublicId", "numKeeps": 10 }, {...}, ... ], "failures": "..."}


ahsu1230:
merging so front-end can use this updated endpoint

ahsu1230:
happy to address comments after

PR #11723, opened 2014-10-22
Switch to narrow styles for buttons when library card is narrow.

yipingliao:
https://app.asana.com/0/15523651812135/18431947315171
@ahsu1230 cc @andrewconner 

Screenshot attached.
screen shot 2014-10-22 at 2 57 15 pm

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/library.styl
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/library.styl
@@ -36,6 +36,20 @@ invite-header-orange = #ff7c3c
 
 
 .library-card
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/library.styl
```diff
@@ -187,6 +200,9 @@ invite-header-orange = #ff7c3c
     padding: 15px 20px
     background-color: #fff
 
+  .kf-library-share-search
```
yipingliao, line 203: From here on down in this file are style tweaks, not related to the Asana issue.

PR #11725, opened 2014-10-22
PasswordSignupTest

raymondng:
This covers the signup flow (u/p)

@andrewconner please take a look -- don't merge yet.


andrewconner:
Awesome. Should be easy to add `/auth/email-signup` too. Thanks!

raymondng:
Yes that'd actually be simpler!

raymondng:
Alright merging. Will take care of email-signup in a follow-up PR (hopefully as simple as I thought)

PR #11728, opened 2014-10-22
Tiny bug fixes

yipingliao:
@ahsu1230 

This fixes keeps not moving from the library (thanks for noticing this!)

Also, remove an inadvertently added file that I committed about 6 hours ago.


kifi-backend/modules/shoebox/angular/src/keeps/keeps.js
```diff
@@ -218,7 +222,8 @@ angular.module('kifi')
                 selectedKeep.makeUnkept();
               });
 
-              undoService.add(selectedKeeps.length + ' keeps deleted.', function () {
+              var keepsDeletedText = selectedKeeps.length > 1 ? ' keeps deleted' : ' keep deleted';
```
yipingliao, line 225: I was seeing "1 keeps deleted", so this fixes that.

PR #11737, opened 2014-10-23
Keep Image Stuff

leogrim:
@andrewconner @stkem 
Ignore the first commit, it's just an opportunistic cleanup in Scraper. 
The relevant changes are in KeepsCommander, if I'm right this kills both remaining tasks and matches all the "conflicting goals" mentioned in Asana.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
@@ -298,10 +298,7 @@ class KeepsCommander @Inject() (
     futureSummary.map { summary =>
       keepImageUrlOpt match {
         case Some(keepImageUrl) => summary.copy(imageUrl = Some(keepImageUrl))
-        case None => {
-          SafeFuture { keepImageCommander.autoSetKeepImage(keep.id.get, localOnly = true, overwriteExistingChoice = false) }
-          summary
-        }
+        case None => summary
       }
     }
   }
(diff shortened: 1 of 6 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
```diff
@@ -291,8 +292,18 @@ class KeepsCommander @Inject() (
     }
   }
 
-  private def getKeepSummary(keep: Keep, waiting: Boolean = false): Future[URISummary] = {
-    uriSummaryCommander.getDefaultURISummary(keep.uriId, waiting)
+  private def getKeepSummary(keep: Keep, idealImageSize: ImageSize, waiting: Boolean = false): Future[URISummary] = {
```
andrewconner, line 295: How will the scraper call this?

leogrim, line 295: Why does it have to?

andrewconner, line 295: 1) When the user reopens the keeper, there's an image (if we have one). 2) When the user goes to the site, they see the image.

As it stands now, as I understand (correct me if I'm wrong), after a successful scrape, scraper populates the uri summary proactively. We still want that _if_ it's the first time we've made this call.

leogrim, line 295: By my understanding, this calls Scraper if necessary, not the other way around. So I guess point 1) is valid since the extension doesn't call this - I'm looking where else we can hook in - but it seems to me that this will handle point 2.

leogrim:
@andrewconner I have removed the async autoset in getKeepSummary, it won't work with waiting = false. So this covers the augmentation task, and I'm investigating other solutions for actually setting keep images.

PR #11747, opened 2014-10-23
Hover card fixes

yipingliao:
@stkem 

Update styles a little for the mini hover card and the attribution in the keep card.
Asana: https://app.asana.com/0/15523651812135/18370961969861

Made new Asana for the tooltip issue we discussed yesterday:
https://app.asana.com/0/15523651812135/18528554791003


kifi-backend/modules/shoebox/angular/src/common/directives/tooltip/tooltip.tpl.html
```diff
@@ -1,5 +1,5 @@
 <div class="kf-tooltip">
-  <div ng-if="showing()">
+  <div ng-show="showing()">
```
stkem, line 2: As discussed, back to ng-if

PR #11753, opened 2014-10-23
Make library hover card in left nav consistent with following state.

yipingliao:
@ahsu1230 
https://app.asana.com/0/15523651812135/18585164429705

Actually, I couldn't reproduce Jen's bug. I think that it may have been fixed by https://github.com/kifi/keepit/pull/11749.

However, I did see an issue where the left nav was not being updated after following or unfollowing from a mini hover card. This fixes _that_.

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/libraryMiniCard.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/libraryMiniCard.js
@@ -2,8 +2,9 @@
 
 angular.module('kifi')
 
-.directive('kfLibraryMiniCard', ['libraryService', 'profileService', 'friendService', 'modalService', '$location',
-  function (libraryService, profileService, friendService, modalService, $location) {
+.directive('kfLibraryMiniCard', [
+  '$rootScope', 'libraryService', 'profileService', 'friendService', 'modalService', '$location',
+  function ($rootScope, libraryService, profileService, friendService, modalService, $location) {
     return {
       restrict: 'A',
       replace: true,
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/libraryMiniCard.js
```diff
@@ -62,6 +67,10 @@ angular.module('kifi')
             libraryService.leaveLibrary(scope.libraryId).then(function () {
               lib.membership = 'none';
               lib.numFollowers = lib.numFollowers - 1;
+
+              libraryService.fetchLibrarySummaries(true).then(function () {
```
andrewconner, line 71: Shouildn't fetching the summaries emit the message? Harder to forget to do it.

kifi-backend/modules/shoebox/angular/src/libraries/libraryMiniCard.js
```diff
@@ -62,6 +67,10 @@ angular.module('kifi')
             libraryService.leaveLibrary(scope.libraryId).then(function () {
```
andrewconner, line 67: Could leaving a library auto-refresh library summaries? + joining

kifi-backend/modules/shoebox/angular/src/libraries/libraryMiniCard.js
yipingliao, line 71: Will look into this. Yes, it's a little messy right now. Will clean up.

PR #11759, opened 2014-10-23
Sign-up test improvements ++

raymondng:
`PasswordSignupTest` now covers the following:
- initial sign-up
- finalize 
- logout
- login
  And does cookie check in each phase (among others).
  Note that this is similar to what the current E2E Selenium test already does today.

Refactored `AuthController` to [1] reduce dependency on `SecureSocial` and [2] set userId explicitly upon login. 
Associate `request.session` with result explicitly

More work required.

@andrewconner


andrewconner:
Great thanks

PR #11764, opened 2014-10-23
bulk add tags to keep

eishay:
@andrewconner @joshm1 @ahsu1230

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
@@ -11,7 +11,6 @@ import com.keepit.common.core._
 import com.keepit.common.crypto.{ PublicIdConfiguration, PublicId }
 import com.keepit.common.db._
 import com.keepit.common.db.slick._
-import com.keepit.common.net.URISanitizer
 import com.keepit.common.time._
 import com.keepit.common.akka.SafeFuture
 import com.keepit.common.logging.Logging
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
```diff
@@ -651,6 +651,16 @@ class KeepsCommander @Inject() (
     addToCollection(tag.id.get, bookmarks) // why doesn't this update search?
   }
 
+  def tagKeeps(tag: Collection, userId: Id[User], keepIds: Seq[ExternalId[Keep]])(implicit context: HeimdalContext): (Seq[Keep], Seq[Keep]) = {
+    val (canEditKeep, cantEditKeeps) = db.readOnlyMaster { implicit s =>
+      keepIds map keepRepo.get partition { keep =>
+        keep.libraryId.exists(libraryCommander.canModifyLibrary(_, userId))
```
andrewconner, line 657: Expensive. Should do a `groupBy(_.libraryId)` first.

eishay, line 657: sure

ahsu1230:
:+1: thanks Eishay. I think we have multiple tag multiple keeps (1 keep -> multiple tags) endpoints...
/site/collections/:id/addKeeps
/site/keeps/tag

both in KeepsController & both take in collectionId (one in URL, other in POSt body) so we should probably remove some endpoints that overlap in the future =)

eishay:
can you please create an asana task for those for post launch?

PR #11766, opened 2014-10-23
In not overwriting mode, do not replace inactive KeepImage

leogrim:
@andrewconner does this make sense?
This way that method is idempotent. This overloads the "inactive" state though, I think a cleaner solution would be to save an active dummy "no image" row in there when users explicitly don't want one. Good enough for now though?


andrewconner:
Previously, `ExistingStoredImagesFound(existingImagesForKeep)` means active, valid images for the keep. Would need to think through the cases again to make sure this isn't an issue.

andrewconner:
I think it's fine. It changes semantics a bit — before, when `overwriteExistingImage` was false, it meant to not overwrite an active selection. Your change is a valid interpretation also, and I think may be a better way to handle it anyways.

PR #11767, opened 2014-10-23
New keep card

joshm1:
@stkem @andrewconner @yipingliao 

apologies for the long pull request. It never got to a point where I was comfortable merging, especially after forking files (see deprecated files/directives that are copies from the non-deprecated version before this pull request). so it's not as large as it looks

I started by renaming css classes for the old card keep to have "-deprecated" suffix, but that was getting too complicated b/c many classes were getting changed, so I started adding "-new" to CSS classes updated for the new keep card.

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/common/directives/keepWho/keepWhoPics.js
+++ b/kifi-backend/modules/shoebox/angular/src/common/directives/keepWho/keepWhoPics.js
@@ -26,11 +26,23 @@ angular.module('kifi')
         keep: '='
       },
       link: function (scope) {
+        var keep = scope.keep;
+
         scope.me = profileService.me;
         scope.getPicUrl = keepWhoService.getPicUrl;
         scope.getName = keepWhoService.getName;
         scope.isMyBookmark = scope.keep && scope.keep.isMyBookmark;
         scope.librariesEnabled = libraryService.isAllowed();
+        
+        if (scope.librariesEnabled) {
+          scope.visibleKeepLibaries = _.union(keep.libraries, keep.myLibraries);
+          scope.$emit('getCurrentLibrary', { callback: function (currentLibrary) {
+            var currentLibraryIdx = _.findIndex(scope.visibleKeepLibaries, { id: currentLibrary.id });
+            if (currentLibraryIdx > -1) {
+              scope.visibleKeepLibaries.splice(currentLibraryIdx, 1);
+            }
+          }});
+        }
       }
     };
   }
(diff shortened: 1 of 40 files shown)


joshm1:
Migration plan from old to new keep card is to delete the "deprecated" directives and templates. and remove the "-new"  suffix from the new files.

joshm1:
still working on:
- ability to unkeep from the library dialog (prompted by keeping)

andrewconner:
I'll post bugs I find here.

Don't show dropdown at all if there's no suggestions:


--- done (josh)

andrewconner:
“x” is oddly positioned. Maybe missing from design?



--- done (josh)

andrewconner:
Perhaps not your bug, but the UX involving the “kept” button is confusing. I click it, unsure of what will happen. Library selection opens — I'm still unsure of what I'm doing. I click a library, unsure of what it will do (move to that library? copy to that library). Selection closes, no confirmation.

What it does: copies the keep to another library

@yipingliao How can we fix this?

andrewconner:
Removing the last tag leaves card in weird state. See: http://acon.s3.amazonaws.com/Screen%20Recording-1.mov

--- done (josh)

andrewconner:
z-index / layer issue with tag list: http://acon.s3.amazonaws.com/Screen%20Recording-2.mov

--- nab/wnf

andrewconner:
Eishay has a new batch tags edit endpoint: https://github.com/kifi/keepit/pull/11764/files

--- i'll update in another PR

andrewconner:
Loading library pages initiates a LOT of outgoing new requests. I don't think this is necessary — seems to be related to the attribution perhaps? http://acon.s3.amazonaws.com/Screen%20Recording-3.mov

--- not a bug from this PR

kifi-backend/modules/shoebox/angular/src/common/directives/keepWho/keepWhoPics.js
```diff
@@ -26,11 +26,23 @@ angular.module('kifi')
...
+
         scope.me = profileService.me;
         scope.getPicUrl = keepWhoService.getPicUrl;
         scope.getName = keepWhoService.getName;
         scope.isMyBookmark = scope.keep && scope.keep.isMyBookmark;
         scope.librariesEnabled = libraryService.isAllowed();
+        
+        if (scope.librariesEnabled) {
+          scope.visibleKeepLibaries = _.union(keep.libraries, keep.myLibraries);
+          scope.$emit('getCurrentLibrary', { callback: function (currentLibrary) {
```
andrewconner, line 39: @ahsu1230 By the way, this is a new way for you to get the library from elsewhere in the app. May be useful for the search bar?

ahsu1230, line 39: hmm, I see. Just for clarification if we're on the same page, the search bar could emit a signal to LibraryCtrl, passing a callback function that takes the LibraryCtrl's current library, and sets the search bar's scope library.
This signal could be emitted on all route changes & even page refreshes.
From my understanding?

PR #11769, opened 2014-10-23
wire-up connectOptionView

raymondng:
2 cases (when email matches some kifi user):
- sui cannot be found 
- sui found but no associated user 

we then present the same page the old endpoint does (`connectOptionView`)

@andrewconner please take a look. don't merge yet.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/core/AuthHelper.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/core/AuthHelper.scala
@@ -63,9 +63,26 @@ class AuthHelper @Inject() (
     libraryCommander: LibraryCommander,
     userCommander: UserCommander,
     heimdalContextBuilder: HeimdalContextBuilderFactory,
-    secureSocialClientIds: SecureSocialClientIds,
+    implicit val secureSocialClientIds: SecureSocialClientIds,
     resetPasswordEmailSender: ResetPasswordEmailSender,
     fortytwoConfig: FortyTwoConfig) extends HeaderNames with Results with Status with Logging {
+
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/app/com/keepit/controllers/core/AuthHelper.scala
```diff
@@ -453,27 +470,33 @@ class AuthHelper @Inject() (
...
-                  val saved = UserService.save(UserIdentity(None, filledUser.copy(oAuth2Info = Some(oauth2InfoNew)), allowSignup = false))
-                  log.info(s"[accessTokenSignup($providerName)] created social user: $saved")
-                  Authenticator.create(saved).fold(
-                    error => throw error,
-                    authenticator =>
-                      Ok(Json.obj("code" -> "continue_signup", "sessionId" -> authenticator.id))
-                        .withSession(request.session - SecureSocial.OriginalUrlKey - IdentityProvider.SessionId - OAuth1Provider.CacheKey)
-                        .withCookies(authenticator.toCookie)
-                  )
+                  if (emailAddressMatchesSomeKifiUser(filledUser)) Ok(connectOptionView(filledUser))
```
andrewconner, line 473: This is a JSON API. HTML will just confuse the parser.

raymondng, line 473: Yeah I was wondering whether that'd work. Should I just send a code and let client handle it?

raymondng, line 473: Ok. code + uri for now. May need a new route. Closing.

PR #11770, opened 2014-10-23
Fix issues with mini hover cards and refactor fetch and follow/unfollow events

yipingliao:
@andrewconner @stkem cc @ahsu1230 

I started working on the mini-hover card following and ended up with a medium-sized pull that fixes the following things:
(1) Address code review comments in https://github.com/kifi/keepit/pull/11753
(2) Actually fix https://app.asana.com/0/15523651812135/18585164429705 (the behavior was inconsistent due to a race condition and I couldn't reproduce earlier but now this should fix it)
(3) Refactor mini-hover-card code to not initiate network requests for individual library summaries when it's not needed (this is not needed when the hover card is in the left nav pane, but needed when the hover card is in the keep card)
(4) Hacky fix for the problem where the mini hover card has its bottom half hidden when we're hovering over a library in lower left part of the nav pane).

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/common/directives/tooltip/tooltip.js
+++ b/kifi-backend/modules/shoebox/angular/src/common/directives/tooltip/tooltip.js
@@ -50,7 +50,7 @@ angular.module('kifi')
               setCssPos(parentPos.top - 0.5*el.outerHeight() + 0.5*container.outerHeight(), parentPos.left - el.outerWidth());
               break;
             case 'right':
-              setCssPos(parentPos.top - 0.5*el.outerHeight() + 0.5*container.outerHeight(), parentPos.left + container.outerWidth());
+              setCssPos(parentPos.top - 0.5*el.outerHeight() + 0.5*container.outerHeight(), parentPos.left + 0.5*container.outerWidth());
               break;
           }
         }
--- a/kifi-backend/modules/shoebox/angular/src/keep/keepToLibraryWidget.js
+++ b/kifi-backend/modules/shoebox/angular/src/keep/keepToLibraryWidget.js
@@ -261,8 +261,6 @@ angular.module('kifi')
 
           libraryService.createLibrary(library).then(function () {
             libraryService.fetchLibrarySummaries(true).then(function () {
-              $rootScope.$emit('librarySummariesChanged');
-
               scope.$evalAsync(function () {
                 scope.librarySelection.library = _.find(scope.libraries, { 'name': library.name });
                 if (_.isFunction(scope.clickAction)) {
--- a/kifi-backend/modules/shoebox/angular/src/layout/nav/nav.tpl.html
+++ b/kifi-backend/modules/shoebox/angular/src/layout/nav/nav.tpl.html
@@ -120,11 +120,10 @@
                                      'other':' {} followers'}"></ng-pluralize>
               </div>
             </div>
-            <div kf-tooltip position="right" padding="3">
-              <div kf-library-mini-card library-id="library.id"></div>
-            </div>
           </a>
-
+          <div kf-tooltip position="top" padding="3">
+            <div kf-library-mini-card library="library"></div>
+          </div>
         </li>
         <li ng-repeat="library in invitedLibsToShow" class="kf-nav-lib-item" ng-class="{ active: isActive(library.url) }">
           <a class="kf-nav-link" ng-href="{{library.url}}">
@@ -137,10 +136,10 @@
               <div class="kf-nav-lib-name">{{library.secondLine}}</div>
               <div class="kf-nav-lib-text">invitation pending</div>
             </div>
-            <div kf-tooltip position="right" padding="3">
-              <div kf-library-mini-card library-id="library.id"></div>
-            </div>
           </a>
+          <div kf-tooltip position="top" padding="3">
+            <div kf-library-mini-card library="library" invite="true"></div>
+          </div>
         </li>
       </ul>
     </div>
--- a/kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
@@ -239,23 +239,13 @@ angular.module('kifi')
               pictureName: profileService.me.pictureName
             });
 
-            libraryService.fetchLibrarySummaries(true).then(function () {
-              $rootScope.$emit('librarySummariesChanged');
-              augmentData();
-              adjustFollowerPicsSize();
-            });
+            augmentData();
+            adjustFollowerPicsSize();
           });
         };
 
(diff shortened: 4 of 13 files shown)


kifi-backend/modules/shoebox/angular/src/common/directives/tooltip/tooltip.js
```diff
@@ -50,7 +50,7 @@ angular.module('kifi')
               setCssPos(parentPos.top - 0.5*el.outerHeight() + 0.5*container.outerHeight(), parentPos.left - el.outerWidth());
               break;
             case 'right':
-              setCssPos(parentPos.top - 0.5*el.outerHeight() + 0.5*container.outerHeight(), parentPos.left + container.outerWidth());
+              setCssPos(parentPos.top - 0.5*el.outerHeight() + 0.5*container.outerHeight(), parentPos.left + 0.5*container.outerWidth());
```
yipingliao, line 53: @stkem This results in the hover card covering the right half of its parent container. The reason is that after I moved the hover card out of the `a` tag, it was too far to the right because its new parent container has a much larger width. Post-launch and for generality, this directive should let the user specify what the offset should be - for a wide container, we probably want the container to be mostly covered, but not so for a narrow container.

kifi-backend/modules/shoebox/angular/src/keep/keepToLibraryWidget.js
```diff
@@ -261,8 +261,6 @@ angular.module('kifi')
 
           libraryService.createLibrary(library).then(function () {
             libraryService.fetchLibrarySummaries(true).then(function () {
-              $rootScope.$emit('librarySummariesChanged');
```
yipingliao, line 264: I've removed all calls to `$rootScope.$emit('librarySummariesChanged1')` in the callback for `fetchLibrarySummaries`. Instead, `fetchLibrarySummaries` itself emits this. Thanks for the suggestion, @andrewconner . Much cleaner now.

kifi-backend/modules/shoebox/angular/src/layout/nav/nav.tpl.html
```diff
@@ -120,11 +120,10 @@
                                      'other':' {} followers'}"></ng-pluralize>
               </div>
             </div>
-            <div kf-tooltip position="right" padding="3">
-              <div kf-library-mini-card library-id="library.id"></div>
-            </div>
           </a>
-
+          <div kf-tooltip position="top" padding="3">
```
yipingliao, line 124: Moved out of `a` tag. :)

kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
```diff
@@ -239,23 +239,13 @@ angular.module('kifi')
...
         scope.unfollowLibrary = function (library) {
-          libraryService.leaveLibrary(library.id).then(function () {
-            libraryService.fetchLibrarySummaries(true).then(function () {
-              $rootScope.$emit('librarySummariesChanged');
-
-              // Note: no need to augmentData for unfollowed library.
-              adjustFollowerPicsSize();
-            });
-          });
+          libraryService.leaveLibrary(library.id);
```
yipingliao, line 248: `joinLibrary` and `leaveLibrary` were already emitting `librarySummariesChanged.`, so no need to do this extra stuff.

kifi-backend/modules/shoebox/angular/src/libraries/libraryMiniCard.js
```diff
@@ -3,98 +3,89 @@
...
+          libraryService.joinLibrary(scope.library.id).then(function (result) {
+            if (result === 'already_joined') {
+              scope.genericErrorMessage = 'You are already following this library!';
+              modalService.open({
+                template: 'common/modal/genericErrorModal.tpl.html',
+                scope: scope
+              });
+              return;
+            } else {
+              if ($location.path() === scope.library.url) {
```
yipingliao, line 30: When we follow a library from the mini hover card, and we're already on that library branch, we need to reload the page so that the changes will be reflected in the library card.

kifi-backend/modules/shoebox/angular/src/keep/keepToLibraryWidget.js
ahsu1230, line 264: yay!

kifi-backend/modules/shoebox/angular/src/layout/nav/nav.tpl.html
yipingliao, line 124: Also, making the default position top, which fixes things :), but in a brittle way. Hopefully good enough for launch.

kifi-backend/modules/shoebox/angular/src/tagManage/tagManage.js
```diff
@@ -188,9 +186,7 @@ angular.module('kifi')
             modalData: { library : $scope.librarySelection.library, action: $scope.actionToLibrary }
           });
           // todo (aaron): call addToLibraryCount accordingly (make sure source libraries lose keep counts)
-          libraryService.fetchLibrarySummaries(true).then(function () {
-            $rootScope.$emit('librarySummariesChanged');
-          });
+          libraryService.fetchLibrarySummaries(true);
```
ahsu1230, line 189: :+1: really liking the clean up :ghost:

yipingliao:
I'm gonna try merging this so that we can see what are the real follow/hover bugs left.

kifi-backend/modules/shoebox/angular/src/common/directives/tooltip/tooltip.js
stkem, line 53: Why did you move it out of the `a` tag?

yipingliao, line 53: We talked about this before. When you click "Follow" or "Unfollow", the page routes to that library page because of the `a` tag, not because of the buttons themselves.

This is causing an issue because sometimes when the user follows a library from the hover card, that library page loads with the old library data, which doesn't reflect the fact that the user has started to follow the library. So, the code now routes to the library page only after `joinLibrary` or `leaveLibrary` has succeeded, ensuring that when we land on the page we have the updated data.

kifi-backend/modules/shoebox/angular/src/layout/nav/nav.tpl.html
stkem, line 124: As you mentioned in another comment, this introduces positioning problems. A better way would be to wrap the a-tag and the tooltip in a common (new) parent, that will have the same size as the `a` tag.

stkem, line 124: That way you can avoid the additional shift by half the parent size you added, which reduces generic usability of the directive.

PR #11780, opened 2014-10-24
Autoset KeepImage asynchronously + request consolidation of URISummary

leogrim:
@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/URISummaryCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/URISummaryCommander.scala
@@ -24,6 +24,9 @@ import com.keepit.cortex.CortexServiceClient
 import com.keepit.search.ArticleStore
 import com.keepit.scraper.embedly.EmbedlyKeyword
 import com.keepit.normalizer.NormalizedURIInterner
+import com.keepit.common.service.RequestConsolidator
+import scala.concurrent.duration._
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/URISummaryCommander.scala
```diff
@@ -90,35 +93,24 @@ class URISummaryCommander @Inject() (
     }
   }
```
leogrim, line 95: @ymatsuda

kifi-backend/modules/shoebox/app/com/keepit/commanders/URISummaryCommander.scala
```diff
@@ -227,7 +227,7 @@ class URISummaryCommander @Inject() (
           imageInfoRepo.save(ImageInfo(uriId = nUri.id.get, url = Some(nUri.url), provider = Some(ImageProvider.EMBEDLY), format = Some(ImageFormat.UNKNOWN)))
         }
         airbrake.notify(s"Could not fetch from embedly because of timeout, persisting a tombstone for the image in $failImageInfo", timeout)
-        Future.successful(Some(URISummary(title = nUri.title)))
+        Future.successful(None)
```
leogrim, line 230: @eishay returning None here to avoid this being persisted / cached.

andrewconner:
Thanks, I think this will work as we need

PR #11783, opened 2014-10-24
user preferences front end -> fix on JsString

ahsu1230:
@andrewconner @yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
+++ b/kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
@@ -27,7 +27,7 @@ angular.module('kifi')
         scope.secretLib = {};
         scope.userLibsToShow = [];
         scope.invitedLibsToShow = [];
-        scope.sortingMenu = { show : false, option : 'last_kept' };
+        scope.sortingMenu = { show : false, option : '' };
 
         scope.counts = {
           friendsCount: friendService.totalFriends(),
@@ -51,12 +51,17 @@ angular.module('kifi')
           scope.secretLib = _.find(libraryService.librarySummaries, { 'kind' : 'system_secret' });
           allUserLibs = _.filter(libraryService.librarySummaries, { 'kind' : 'user_created' });
 
-          var newList = sortLibraries(allUserLibs, libraryService.invitedSummaries);
-          scope.userLibsToShow = newList[0];
-          scope.invitedLibsToShow  = newList[1];
+          if (typeof(scope.sortingMenu.option) === 'undefined') {
+            scope.userLibsToShow = [];
+            scope.invitedLibsToShow  = [];
+          } else {
+            var newList = sortLibraries(allUserLibs, libraryService.invitedSummaries);
+            scope.userLibsToShow = newList[0];
+            scope.invitedLibsToShow  = newList[1];
+            librarySummarySearch = new Fuse(allUserLibs, fuseOptions);
+            invitedSummarySearch = new Fuse(libraryService.invitedSummaries, fuseOptions);
+          }
 
-          librarySummarySearch = new Fuse(allUserLibs, fuseOptions);
-          invitedSummarySearch = new Fuse(libraryService.invitedSummaries, fuseOptions);
           scope.$broadcast('refreshScroll');
         }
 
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
```diff
@@ -89,6 +94,11 @@ angular.module('kifi')
 
         $rootScope.$on('librarySummariesChanged', updateNavLibs);
 
+        $rootScope.$on('changedLibrarySorting', function() {
+          scope.sortingMenu.option = profileService.prefs.library_sorting_pref.replace(/[\\\"]/g, '');
```
ahsu1230, line 98: this is needed because for some reason the endpoint returns embedded quotations like: "\"name\""
this fix should get rid of this, so will remove the replace after confirming

ahsu1230:
merging now for deploy. Can address comments later =)

kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
andrewconner, line 98: Hm, not sure why it's doing this. I see the `JsString` below, interesting. Normally, `JsString("hi")` works as expected.

ahsu1230, line 98: For some reason when you JsString("hi") it returns as expected
val a = "hi"
but JsString(a) gives "\"hi\""
_SHRUG_

andrewconner, line 98: By the way, `prefs.library_sorting_pref.replace` will have problems if library_sorting_pref is undefined.

kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
```diff
@@ -106,8 +116,11 @@ angular.module('kifi')
         scope.$watch(function () {
             return scope.sortingMenu.option;
           }, function () {
-          scope.changeList();
-          scope.turnDropdownOff();
+          if (scope.sortingMenu.option !== '' && typeof(scope.sortingMenu.option) !== 'undefined') {
```
2is10, line 119: This is newbie JS code. :smirk: We very rarely need to differentiate between the different falsy values (`false`,`null`,`undefined`,`''`,`0`,`NaN`). There are six; memorize them. More idiomatic (and fewer bytes) would be:

``` js
  if (scope.sortingMenu.option) {
```

ahsu1230, line 119: woops. Is {} a falsy value?

andrewconner, line 119: :) Of note, a mistake I've made, `[]` is not falsey

kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
2is10, line 98: > For some reason when you JsString("hi") it returns as expected
> val a = "hi"
> but JsString(a) gives "\"hi\""

I’m not seeing that:

``` scala
scala> val a = "hi"
a: String = hi

scala> JsString(a) 
res0: play.api.libs.json.JsString = "hi"

scala> JsString("hi")
res1: play.api.libs.json.JsString = "hi"
```

I’m afraid this code may be writing quotes into the database.

kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
2is10, line 119: You can easily check in your browser console like this:

``` js
Boolean([])
> true
Boolean({})
> true
Boolean('')
> false
```

kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
ahsu1230, line 98: yea I think it's in `savePrefs()` in `UserCommander` (line 500)
The value being saved is a JsValue type
I've been testing the JsString/Json stuff through a string checking test... but I haven't found what would work.

2is10, line 98: I’ve merged a fix. #11785

PR #11784, opened 2014-10-24
"fixing" tooltip positioning

stkem:
@yipingliao 
There are now 8 different positions the tooltip can take (added diagonal ones) so that should cover all cases.


yipingliao:
Eight! Nice. :)

PR #11787, opened 2014-10-24
user preferences for library sorting - removing newb code

ahsu1230:
@andrewconner @2is10 

Thanks for the fixes before =) @2is10

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
+++ b/kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
@@ -91,9 +91,10 @@ angular.module('kifi')
 
         $rootScope.$on('changedLibrarySorting', function() {
           scope.sortingMenu.option = profileService.prefs.library_sorting_pref;
-          if (scope.sortingMenu.option) {
-            updateNavLibs();
+          if (!scope.sortingMenu.option) {
+            scope.sortingMenu.option = 'last_kept';
           }
+          updateNavLibs();
         });
 
         scope.$watch(function () {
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
```diff
@@ -95,7 +90,10 @@ angular.module('kifi')
         $rootScope.$on('librarySummariesChanged', updateNavLibs);
 
         $rootScope.$on('changedLibrarySorting', function() {
-          scope.sortingMenu.option = profileService.prefs.library_sorting_pref.replace(/[\\\"]/g, '');
+          scope.sortingMenu.option = profileService.prefs.library_sorting_pref;
+          if (!scope.sortingMenu.option) {
+            scope.sortingMenu.option = 'last_kept';
```
ahsu1230, line 95: if fetch preferences, and the sorting pref is undefined, the user has never sorted before. So we give the default value of 'last-kept'. And it should be saved to the database after

kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
```diff
@@ -95,7 +90,10 @@ angular.module('kifi')
         $rootScope.$on('librarySummariesChanged', updateNavLibs);
 
         $rootScope.$on('changedLibrarySorting', function() {
-          scope.sortingMenu.option = profileService.prefs.library_sorting_pref.replace(/[\\\"]/g, '');
+          scope.sortingMenu.option = profileService.prefs.library_sorting_pref;
```
andrewconner, line 93: This can be written:

``` javascript
scope.sortingMenu.option = profileService.prefs.library_sorting_pref || 'last_kept';
```

andrewconner:
Thanks

andrewconner:
Merging because this fixes a problem in production. Any time `library_sorting_pref` is undefined (ie, for al library users now until they set it), there's an exception.

ahsu1230:
Ok, thanks!

PR #11789, opened 2014-10-24
connectOption

raymondng:
code + uri, as discussed. email also added.
@andrewconner please take a look. don't merge yet.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/core/AuthController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/core/AuthController.scala
@@ -15,6 +15,8 @@ import com.kifi.macros.json
 import com.keepit.common.controller.KifiSession._
 
 import play.api.Play._
+import play.api.data._
+import play.api.data.Forms._
 import play.api.i18n.Messages
 import play.api.libs.json.{ JsValue, JsNumber, Json }
 import play.api.mvc._
@@ -41,9 +43,7 @@ object AuthController {
   def obscureEmailAddress(address: String) = obscureRegex.replaceFirstIn(address, """$1...@""")
 }
 
-@json case class ConnectOptionInfo(private val email: String, provider: String) {
-  val emailAddress = EmailAddress(email)
-}
+case class ConnectOptionInfo(val emailAddress: EmailAddress, provider: String)
 
 @json case class EmailSignupInfo(email: String)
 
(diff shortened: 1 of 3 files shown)


andrewconner:
Thanks, I think this will work. Only unsure about having to redirect the user to a POST, but I think I know how.

kifi-backend/modules/shoebox/app/com/keepit/controllers/core/AuthController.scala
```diff
@@ -252,6 +256,18 @@ class AuthController @Inject() (
     }
   }
 
+  def connectOption() = MaybeUserAction(parse.tolerantJson) { implicit request =>
```
andrewconner: Can the body be form URL encoded? I'm sending the user here, not using a json API here.

kifi-backend/modules/shoebox/app/com/keepit/controllers/core/AuthController.scala
```diff
@@ -41,6 +41,10 @@ object AuthController {
   def obscureEmailAddress(address: String) = obscureRegex.replaceFirstIn(address, """$1...@""")
 }
 
+@json case class ConnectOptionInfo(private val email: String, provider: String) {
```
andrewconner: `provider` is formatted how? `facebook` and `email`?

kifi-backend/modules/shoebox/app/com/keepit/controllers/core/AuthController.scala
raymondng: Ah good point. Let's do that.

kifi-backend/modules/shoebox/app/com/keepit/controllers/core/AuthController.scala
raymondng: It's used to construct a `SocialNetworkType` so I'd say yes -- definitely with `facebook`. I am not too familiar with the `email` type but based on limited testing looks like it should work.

raymondng: Actually given we only support FB in this flow, and we trigger it via social -- `facebook` looks like the only legitimate option here for now until we support LNKD or other providers. WDYT?

kifi-backend/modules/shoebox/app/com/keepit/controllers/core/AuthController.scala
```diff
@@ -252,6 +256,26 @@ class AuthController @Inject() (
...
 
+  val connectOptionForm = Form[ConnectOptionInfo](
+    mapping(
+      "email" -> EmailAddress.formMapping,
+      "provider" -> text
+    )(ConnectOptionInfo.apply)(ConnectOptionInfo.unapply)
+  )
+  def connectOption() = MaybeUserAction { implicit request =>
+    connectOptionForm.bindFromRequest.fold(
+      hasErrors = formWithErrors => BadRequest(Json.obj("error" -> formWithErrors.errors.head.message)),
```
andrewconner, line 267: Returning JSON to a user-facing request?

kifi-backend/modules/shoebox/app/com/keepit/controllers/core/AuthController.scala
```diff
@@ -252,6 +256,26 @@ class AuthController @Inject() (
...
+      success = {
+        case info: ConnectOptionInfo =>
+          if (authHelper.emailAddressMatchesSomeKifiUser(info.emailAddress)) { // sanity check
+            Ok(authHelper.connectOptionView(info.emailAddress, info.provider))
+          } else {
+            BadRequest(Json.obj("error" -> "not_found")) // shouldn't happen
```
andrewconner, line 273: Same

kifi-backend/modules/shoebox/app/com/keepit/controllers/core/AuthController.scala
raymondng, line 267: Yes I was going to ask you about this. That does look like a common pattern for form-handling (see `AuthHelper`) -- I thought maybe we have a way to handle this, but probably assumed too much.

raymondng, line 267: Looks like we need another endpoint or another way to handle this edge case. Will look into it.

PR #11793, opened 2014-10-24
stop indexing main/private library names

ymatsuda:
@leogrim 
This stops indexing of system created library names (main and private).
They create noises. (and the name of private libraries are "Secret Library"!)

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/model/Library.scala
+++ b/kifi-backend/modules/common/app/com/keepit/model/Library.scala
@@ -131,10 +131,8 @@ object LibraryVisibility {
 sealed abstract class LibraryKind(val value: String)
 
 object LibraryKind {
-  trait SystemGenerated extends LibraryKind
-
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/common/app/com/keepit/model/Library.scala
```diff
@@ -131,10 +131,8 @@ object LibraryVisibility {
 sealed abstract class LibraryKind(val value: String)
 
 object LibraryKind {
-  trait SystemGenerated extends LibraryKind
```
ymatsuda, line 134: @andrewconner 
I removed SystemGenerated trait to eliminate "match not exhaustive" warning.
SystemGenerated looks like a marker trait, but it is not used.

andrewconner, line 134: Oh interesting, I would have thought it wouldn't have affected matching negatively. Yep, just a marker trait.

PR #11794, opened 2014-10-24
Lib title height fix, white bar, android location

andrewconner:


andrewconner:
@yipingliao Merging to get it out

PR #11798, opened 2014-10-24
Bug on empty library state (links below didn't work)

ahsu1230:
@andrewconner @yipingliao 

two bugs on empty library state
Links at the bottom didn't work & "user sign up" message showed for signed-in users

Links should now work (but currently a small bug with importing bookmarks from browser)

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/library.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/library.js
@@ -172,6 +172,19 @@ angular.module('kifi')
       });
     };
 
+    $scope.callAddKeep = function () {
+      $rootScope.$emit('triggerAddKeep', $scope.library);
+    };
+    $scope.callImportBookmarks = function () {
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/library.js
```diff
@@ -172,6 +172,19 @@ angular.module('kifi')
...
     };
 
+    $scope.callAddKeep = function () {
+      $rootScope.$emit('triggerAddKeep', $scope.library);
+    };
+    $scope.callImportBookmarks = function () {
+      $rootScope.$emit('showGlobalModal', 'importBookmarks');
+    };
+    $scope.callImportBookmarkFile = function () {
+      $rootScope.$emit('showGlobalModal', 'importBookmarkFile');
```
ahsu1230, line 182: importing bookmarks used signaling before, so no need for new signals to be called

andrewconner:
Great! Super fast, nice job.

PR #11799, opened 2014-10-24
User sign-up with different social

raymondng:
As discussed, looks like redirecting to current signup endpoint should do the trick, as long as a secure social identity exists. It will then detect that it's a `NonUserRequest` but `identityOpt` is defined. 

Hopefully, this simplifies things. Do let me know if I have missed or misunderstood anything.

@andrewconner


raymondng:
Reviewed in person with @andrewconner. Merging.

PR #11800, opened 2014-10-24
Fix problem where import modals are not being populated with an initial ...

yipingliao:
...library in the widget.

@ahsu1230


ahsu1230:
thanks!

PR #11802, opened 2014-10-24
mobile installation tracking

yingjieMiao:
@leogrim  @eishay

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileAuthController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileAuthController.scala
@@ -10,12 +10,13 @@ import com.keepit.common.logging.Logging
 import com.keepit.common.net.UserAgent
 import com.keepit.common.time.Clock
 import com.keepit.controllers.core.{ AuthController, AuthHelper }
+import com.keepit.heimdal.{ ContextDoubleData, HeimdalContext, HeimdalContextBuilderFactory, HeimdalServiceClient, UserEvent, UserEventTypes }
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileAuthController.scala
```diff
@@ -90,12 +95,30 @@ class MobileAuthController @Inject() (
...
 
+    reportMobileInstallation(userId, installation, newInstallation, context)
+
     Ok(Json.obj(
       "installation" -> installation.externalId.toString,
       "newInstallation" -> newInstallation
     ))
   }
 
+  private def reportMobileInstallation(userId: Id[User], installation: KifiInstallation, isNewInstall: Boolean, context: HeimdalContext): Unit = {
```
leogrim, line 106: Seems like you could just pass the builder instead of a context, or even add this as a method on the builder?

yingjieMiao, line 106: passing `context` is a pattern in our code base.

leogrim, line 106: Yes, when you pass it up the chain of controller / commander / analytics. But here you call context.build just to pass it to this (private) helper, seems safe just to add directly to the builder at hand before passing it up.

yingjieMiao, line 106: Ah, got you.

PR #11804, opened 2014-10-24
color changes & move to constants for library coloring

ahsu1230:
@andrewconner @yipingliao @stkem 

keep card attribution to libraries was always green -> now consistent with their respective visibility
and constants =)


kifi-backend/modules/shoebox/angular/src/common/directives/keepWho/keepWhoLib.tpl.html
```diff
@@ -1,5 +1,5 @@
 <span>
-  <a class="kf-keep-who-pic-box" ng-href="{{library.path}}">
+  <a class="kf-keep-who-pic-box" ng-href="{{library.path}}" ng-class="{ 'secret': library.secret == true}">
```
yipingliao: Using `==` is the same as just `library.secret`, which tests for truthy. Is truthy good enough, or did you mean `===`?

yipingliao:
Yay!

stkem:
Thanks!

kifi-backend/modules/shoebox/angular/src/common/directives/keepWho/keepWhoLib.tpl.html
andrewconner: You didn't add this in this PR, but why is there a `secret` field on library anyways. Can we just check visibility?

kifi-backend/modules/shoebox/angular/src/common/build-css/constants.styl
```diff
@@ -13,6 +13,8 @@ darkBlueColor = #478eeb
 greenColor = #41c482
 kifiBlack = #333
 kifiHeadColor = #99ccf2
+libraryRed = #f06463
+libraryGreen = #73c785
```
andrewconner, line 17: Thanks!

kifi-backend/modules/shoebox/angular/src/common/directives/keepWho/keepWhoLib.tpl.html
ahsu1230: So for some reason, there isn't a visibility field at this level. It must've gotten taken out somewhere & substituted with secret boolean field

andrewconner: I see.

PR #11810, opened 2014-10-24
Sending passphrase only to non users

leogrim:


kifi-backend/modules/shoebox/app/com/keepit/commanders/emails/LibraryInviteEmailSender.scala
```diff
@@ -42,14 +42,18 @@ class LibraryInviteEmailSender @Inject() (
 
       val trimmedInviteMsg = invite.message map (_.trim) filter (_.nonEmpty)
       val fromUserId = invite.inviterId
+      val passPhrase = toRecipient match {
+        case Left(_: Id[User]) => None
+        case Right(_: EmailAddress) => Some(invite.passPhrase)
+      }
```
leogrim, line 48: @eishay @stkem 
Sneaky one, it was initially:

```
val passPhrase = toRecipient match { 
   case Right(userId) => None 
   case Left(emailAddress) => Some(invite.passPhrase) 
}
```

eishay, line 48: :-)

eishay:
thanks!

eishay:
pelase test it in prod

leogrim:
Tested in prod, all good.

stkem:
:+1:

PR #11816, opened 2014-10-24
Better type ahead behavior

andrewconner:
This tiny change actually makes it so the backend doesn't send duplicates. Also fixes an issue where two requests are made: one for `query === undefined` and one where `query === ''`


andrewconner:
@yipingliao @ahsu1230

PR #11818, opened 2014-10-24
Keep card fixes

joshm1:
@stkem


kifi-backend/modules/shoebox/angular/src/recos/recos.styl
```diff
@@ -229,6 +229,29 @@
   width: 18px
   height: 18px
 
+.svg-reco-keep-icon
```
andrewconner, line 232: Mark said this is wrong, it was correct before this PR. Was there a new design? + @stkem

joshm1, line 232: before this the reco cards looked like this:

https://www.dropbox.com/s/2e09dzxaov5c8as/Screenshot%202014-10-27%2009.10.46.png?dl=0

This was updated per the latest design that I saw.

PR #11821, opened 2014-10-24
Rename System Libraries

raymondng:
The names are converted when we read from Db. AFAIK this should do the trick. 

@andrewconner @ymatsuda -- please take a look.


andrewconner:
Great. One sec, will get the name.

ymatsuda:
lgtm

andrewconner:
Keep it, you got the right ones.

raymondng:
Cool. Thanks!

PR #11840, opened 2014-10-25
rootScope $on signal handlers destroyed on scope's '$destroy'

ahsu1230:
@yipingliao @andrewconner @joshm1 

am I missing anything?


andrewconner:
Great, thanks. You have a merge conflict.

kifi-backend/modules/shoebox/angular/src/tagManage/tagManage.js
```diff
@@ -215,13 +215,14 @@ angular.module('kifi')
       $scope.tagsToShow.splice(removedIndex, 1);
     };
 
-    $rootScope.$on('undoRemoveTag', function () {
+    var undoRemoveTagHandler = $rootScope.$on('undoRemoveTag', function () {
```
yipingliao: In general, all these variable are misnomers, because they are not the handlers themselves, but rather functions that deregister the handlers. What do you think about prefixing them with "deregister"?

yipingliao:
Thank you @ahsu1230 for doing this. :)

kifi-backend/modules/shoebox/angular/src/tagManage/tagManage.js
ahsu1230: Sure, they would be clearer. like `undoRemoveTagDeregister`

ahsu1230:
thanks for the reviewing!

PR #11852, opened 2014-10-25
keep button visibility and remove user's own library attribution

joshm1:
@stkem


yipingliao:
We're removing the user's own library attribution, really?

stkem:
As per product, attribution to your own libraries (with the chip) will only be shown in search results but not when browsing libraries. The data source should thus set the `libraries` attribute of the keep to include whatever should be displayed in the given scenario, the attribution logic won't need to know. I expect this to be experimented with soon.

yipingliao:
Nice! I think this is a good user experience.

PR #11861, opened 2014-10-25
fixes

andrewconner:
@stkem


kifi-backend/modules/shoebox/angular/src/search/searchActionService.js
```diff
@@ -45,12 +45,8 @@ angular.module('kifi')
         var user;
         if (idxUser !== -1) {
           user = users[idxUser];
-          decompressedLibraries.push({
-            id: lib.id,
-            name: lib.name,
-            keeperPic: friendService.getPictureUrlForUser(user),
-            path: lib.path
-          });
+          lib.keeperPic = friendService.getPictureUrlForUser(user);
```
andrewconner, line 48: This is the big part. The rest is just icing

PR #11864, opened 2014-10-25
fixes issue where user libraries didn't consistently show up in keeper

joshm1:
also fixes the inability to unkeep from recommendations

breaks some instances where if you open the keeper back up it doesn't show the library you just kept to, but this was inconsistent before this

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/keep/keepToLibraryWidget.js
+++ b/kifi-backend/modules/shoebox/angular/src/keep/keepToLibraryWidget.js
@@ -174,18 +174,23 @@ angular.module('kifi')
           scope.newLibrary = {};
           newLibraryNameInput = widget.find('.keep-to-library-create-name-input');
 
-          scope.widgetLibraries = _.filter(scope.libraries, function (library) {
-            return !_.find(scope.excludeLibraries, { 'id': library.id });
-          });
+          libraryService.fetchLibrarySummaries(false).then(function (data) {
+            var libraries = _.filter(data.libraries, { access: 'owner' });
 
-          scope.widgetLibraries.forEach(function (library) {
-            library.keptTo = false;
-            if (scope.keptToLibraries && _.find(scope.keptToLibraries, { 'id': library.id })) {
-              library.keptTo = true;
-            }
-          });
+            libraries = _.filter(libraries, function (library) {
+              return !_.find(scope.excludeLibraries, { 'id': library.id });
+            });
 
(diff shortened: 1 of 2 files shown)


joshm1:
@yipingliao please take a look, it is related to the keeper and maybe some other bugs

@stkem @andrewconner

kifi-backend/modules/shoebox/angular/src/keep/keepToLibraryWidget.js
```diff
@@ -174,18 +174,23 @@ angular.module('kifi')
           scope.newLibrary = {};
           newLibraryNameInput = widget.find('.keep-to-library-create-name-input');
 
-          scope.widgetLibraries = _.filter(scope.libraries, function (library) {
-            return !_.find(scope.excludeLibraries, { 'id': library.id });
-          });
+          libraryService.fetchLibrarySummaries(false).then(function (data) {
```
yipingliao, line 177: Why is this needed? The contract of the widget is that the parent scope of the widget provides a list of libraries on the scope (see line 14). The widget shouldn't be fetching its own libraries. Each parent that calls the widget may want to provide a different list of libraries.

andrewconner, line 177: Also, any caller should be able to use libraryService.librarySummaries. No fetch needed (unless there's worry that the list is out of date?)

joshm1, line 177: This was to fix broken usages of this this directive.  It's fine for the parent to provide the list of libraries, but it was broken and inconsistent in multiple places (search and recommendations), and thus difficult to track down.  

In general, I think we need to be more descriptive with our variable/scope names, and more restrictive between sharing state between scopes.  This directive depends on a parent scope setting this variable, but `scope.libraries` is used a lot and means different things in different places.  I've noticed a lot of bugs that were related to both of these issues. I think we should use more isolate scopes and be explicit about passing state between directives and only use 2-way binding when it's necessary.

yipingliao, line 177: Agreed. `scope.libraries` is a bad name. Also, isolate scope would be great and I really wanted to use it but couldn't figure out how to do it in this case because we needed to have the `showWidget` function be callable. Argh. I've moved the Asana bug to Fast Follow and hope to rework it soon.

PR #11869, opened 2014-10-25
Frontend tweak + return an image when we likely already know what it is.

andrewconner:
@2is10

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/libraryShareSearch.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/libraryShareSearch.js
@@ -126,7 +126,7 @@ angular.module('kifi')
                 if (result.id) {
                   result.image = friendService.getPictureUrlForUser(result);
                   result.isFollowing = !!result.membership;
-                  result.name = result.firstName + ' ' + result.lastName;
+                  result.name = (result.firstName || '') + (result.lastName ? ' ' + result.lastName : '');
                 }
 
                 if (opt_query) {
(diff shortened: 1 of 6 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/libraryShareSearch.js
```diff
@@ -126,7 +126,7 @@ angular.module('kifi')
                 if (result.id) {
                   result.image = friendService.getPictureUrlForUser(result);
                   result.isFollowing = !!result.membership;
-                  result.name = result.firstName + ' ' + result.lastName;
+                  result.name = (result.firstName || '') + (result.lastName ? ' ' + result.lastName : '');
```
2is10, line 129: This will end up with a leading space if there’s a last name and no first name. Not sure if that can happen. Might be simplest to just `trim` the original expression.

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepImageCommander.scala
```diff
@@ -70,6 +71,13 @@ class KeepImageCommanderImpl @Inject() (
     KeepImageSize.pickBest(idealSize, validKeepImages)
   }
 
+  def getExistingImageUrlForKeepUri(nUriId: Id[NormalizedURI])(implicit session: RSession): Option[String] = {
+    imageInfoRepo.getLargestByUriWithPriority(nUriId).flatMap { imageInfo =>
+      val nuri = normalizedUriRepo.get(nUriId)
```
2is10, line 76: `nUri`

kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtLibraryController.scala
```diff
@@ -113,12 +113,15 @@ class ExtLibraryController @Inject() (
 
           val (keep, isNewKeep) = keepsCommander.keepOne(body.as[RawBookmarkRepresentation], request.userId, libraryId, request.kifiInstallationId, source)
           val (tags, image) = if (isNewKeep) {
```
2is10, line 115: `image` → `imageUri` for clarity

2is10:
great, thanks

PR #11870, opened 2014-10-26
correct inviter name in orange accept/ignore banner

stkem:
@andrewconner @yipingliao 
was showing the library owner name


yipingliao:
Thanks! It didn't occur to me that someone other than the owner could be inviting.

PR #11871, opened 2014-10-26
Use keeps not my libraries

yipingliao:
Using `keep.keeps` as indicator of which other libraries a particular keep is in. This is more accurate than the previous `keep.libraries` field (which was filtered into `keep.myLibraries` for the user's libraries) because `keeps` is pulled from the db, whereas `libraries` was pulled form search index (which may take some time to update).

Moreover, we can now unkeep from libraries other than the library being viewed because `keeps` gives us those keep ids!

@andrewconner cc @joshm1

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/common/directives/keepWho/keepWhoPics.js
+++ b/kifi-backend/modules/shoebox/angular/src/common/directives/keepWho/keepWhoPics.js
@@ -32,16 +32,11 @@ angular.module('kifi')
         scope.isMyBookmark = scope.keep && scope.keep.isMyBookmark;
         scope.librariesEnabled = libraryService.isAllowed();
 
-        function updateVisibleKeepLibraries() {
-          if (scope.librariesEnabled) {
-            scope.visibleKeepLibraries = scope.keep.libraries;
-          } else {
-            scope.visibleKeepLibraries = [];
-          }
+        if (scope.librariesEnabled) {
+          scope.visibleKeepLibraries = scope.keep.libraries;
+        } else {
+          scope.visibleKeepLibraries = [];
         }
-
-        scope.$watch('keep.myLibraries.length', updateVisibleKeepLibraries);
-        updateVisibleKeepLibraries();
       }
     };
   }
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/shoebox/angular/src/common/directives/keepWho/keepWhoPics.js
```diff
@@ -32,16 +32,11 @@ angular.module('kifi')
         scope.isMyBookmark = scope.keep && scope.keep.isMyBookmark;
         scope.librariesEnabled = libraryService.isAllowed();
 
-        function updateVisibleKeepLibraries() {
```
yipingliao, line 35: Changes to this file are follow-up to: https://github.com/kifi/keepit/pull/11852
If we don't show user libraries anymore, then there is no need to keep a watch, since the user cannot influence other people's libraries during a session.

kifi-backend/modules/shoebox/angular/src/keep/keep.js
```diff
@@ -988,7 +982,7 @@ angular.module('kifi')
         // Watches and listeners.
         //
         scope.$watchCollection(function () {
-          return _.pluck(scope.keep.myLibraries, 'secret');
+          return _.pluck(scope.keep.keeps, 'visibility');
```
yipingliao, line 985: This is also a follow-up to: https://github.com/kifi/keepit/pull/11852
We're now using `keeps` to determine keep status, not `myLibraries`.

kifi-backend/modules/shoebox/angular/src/keep/keepDecoratorService.js
```diff
@@ -5,8 +5,7 @@ angular.module('kifi')
 .factory('keepDecoratorService', ['tagService', 'util', 'friendService', 'profileService',
   function (tagService, util, friendService, profileService) {
     function processLibraries(item) {
-      item.myLibraries = [];
```
yipingliao, line 8: We no longer need to filter out a set of `myLibraries`.

kifi-backend/modules/shoebox/angular/src/keep/keepToLibraryWidget.js
```diff
@@ -174,6 +174,7 @@ angular.module('kifi')
           scope.newLibrary = {};
           newLibraryNameInput = widget.find('.keep-to-library-create-name-input');
 
+          // May remove this fetch; awaiting discussion with Josh.
           libraryService.fetchLibrarySummaries(false).then(function (data) {
```
yipingliao, line 178: Per earlier discussion, I am thinking about removing this.

PR #11872, opened 2014-10-26
Bringing back keeps view — non-libraries users need it.

andrewconner:
@stkem They have no way to view their keeps.


stkem:
ah, good point

PR #11873, opened 2014-10-26
Template, not redirect

andrewconner:


kifi-backend/modules/shoebox/angular/src/route.js
```diff
@@ -32,7 +32,7 @@ angular.module('kifi')
     }).when('/recommendations', {
       redirectTo: '/'
```
yipingliao, line 33: Oh wow (didn't see it this when it first flew by).

andrewconner, line 33: Yep we rolled it out last night.

PR #11874, opened 2014-10-26
Remove fetching of library summaries in library widget; let all keeps get library summaries

yipingliao:
This addresses:
https://app.asana.com/0/15523651812135/18673889383145

@joshm1 @andrewconner 

When @joshm1 added the fetching of library summaries in the initialization code of the library widget, the above bug is fixed. However, when that fetching is removed, I see the bug.

The problem is that the `kfKeeps` directive, a parent of the widget, was not populating `scope.libraries` when there is no `scope.library`. Detailed explanation inline.

This commit populates `scope.libraries` to fix the bug, and removes the fetching from the widget. Since this does not "fix" anything that's actually broken on the master branch, I am not going to merge this until this is actually approved.

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/keeps/keeps.js
+++ b/kifi-backend/modules/shoebox/angular/src/keeps/keeps.js
@@ -297,6 +297,7 @@ angular.module('kifi')
         var lastSizedAt = $window.innerWidth;
         scope.availableKeeps = scope.keeps;
 
+
         //
         // Internal methods.
         //
@@ -502,7 +503,7 @@ angular.module('kifi')
           if (!newVal) {
             if (scope.librariesEnabled) {
               libraryService.fetchLibrarySummaries().then(function () {
-                scope.libraries = !scope.library ? [] : _.filter(libraryService.librarySummaries, function (library) {
+                scope.libraries = _.filter(libraryService.librarySummaries, function (library) {
                   return library.access !== 'read_only';
                 });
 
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/angular/src/keeps/keeps.js
```diff
@@ -502,7 +503,7 @@ angular.module('kifi')
           if (!newVal) {
             if (scope.librariesEnabled) {
               libraryService.fetchLibrarySummaries().then(function () {
-                scope.libraries = !scope.library ? [] : _.filter(libraryService.librarySummaries, function (library) {
```
yipingliao, line 505: @andrewconner This null check for `scope.library` before populating `scope.libraries` was added here:
https://github.com/kifi/keepit/commit/a3573cd8e5195e450f3bac87d67eece79f53952c

As far as I can tell there is no harm in populating `scope.libraries` when there is no library in scope. This is the case for search, helprank, and any other page where we may have keeps. However, I may be missing a reason.

We can definitely grab libraries from `libraryService` in the widget, as @joshm1 did, and still maintain flexibility by having the parent scope optionally provide a filter function to filter the libraries. All of this is post-launch - as far as I can tell the widget is working on master right now. :)

andrewconner, line 505: That commit was fixing an exception in prod. We could always have a `scope.libraries`, but is it necessary? Seems like it could introduce outdated data, or components modifying a shared global libraries object without realizing.

Slightly unrelated: Several of the bugs fixed on Friday were related to sharing an object across scopes, and one scope modifying the object in ways the other scope didn't expect. Shared objects are great to get updated data, but shouldn't be modified for local purposes.

yipingliao, line 505: Absolutely. I plan to clean up this widget, hopefully today.

PR #11875, opened 2014-10-26
When ENTER key is pressed in the library widget with no library search results, go to the create panel.

yipingliao:
https://app.asana.com/0/15523651812135/18655297498303

@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/keep/keepToLibraryWidget.js
+++ b/kifi-backend/modules/shoebox/angular/src/keep/keepToLibraryWidget.js
@@ -247,7 +247,7 @@ angular.module('kifi')
               // Prevent any open modals from processing this.
               $event.stopPropagation();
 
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/angular/src/keep/keepToLibraryWidget.js
```diff
@@ -160,6 +160,11 @@ angular.module('kifi')
             var left = element.offset().left;
             widget.css({top: top + 'px', left: left + 'px'});
             widget.show();
+
+            // Wait for the widget to be shown, then focus on the search input.
+            $timeout(function () {
```
yipingliao, line 165: This is unrelated to the Asana bug in the pull description. Rather, it was broken from a time where I hide the widget to do some calculations before showing it. Now, the search input in the widget will get focus when the widget is opened.

PR #11911, opened 2014-10-27
Less misleading subtitle

raymondng:
The current helprank page is subtitled "Showing your latest <num> keeps" and it's a bit confusing.

Just showing "Your Helpful Keeps" for now -- will make the number of helpful keeps available and show something like "showing x out of y helpful keeps"

@andrewconner @eishay


raymondng:
Discussed in person with @eishay

PR #11916, opened 2014-10-27
Fix broken bulk selection on the search page ("/find").

yipingliao:
https://app.asana.com/0/17575799252940/18788833524418

@andrewconner 
cc @ahsu1230 Aaron, FYI. Since we filed this bug together.

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/common/selectionService.js
+++ b/kifi-backend/modules/shoebox/angular/src/common/selectionService.js
@@ -10,16 +10,16 @@ angular.module('kifi')
     }
 
     Selection.prototype.select = function (keep) {
-      if (keep && keep.id) {
-        this.selected[keep.id] = keep;
+      if (keep && keep.url) {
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/angular/src/common/selectionService.js
```diff
@@ -10,16 +10,16 @@ angular.module('kifi')
     }
 
     Selection.prototype.select = function (keep) {
-      if (keep && keep.id) {
-        this.selected[keep.id] = keep;
+      if (keep && keep.url) {
```
yipingliao, line 13: `selectionService` used to keep track of which keeps have been selected using the keep's id as a key into a hash. On the search page, keeps no longer have ids because they are/could be in multiple libraries. Switched to using the keep's url instead.

ahsu1230, line 13: I see, no wonder. thanks!

PR #11917, opened 2014-10-27
Fixing not being able to set anything to the session

andrewconner:
@raymondng 

We were also setting a new session for every request (ie, returning a new cookie).


raymondng:
lgtm. Thanks!

PR #11921, opened 2014-10-27
adding 301 redirects from removed /about/... pages to /about

2is10:
@eishay


2is10:
merging, comments welcome

PR #11922, opened 2014-10-27
fixing tag links on keep cards

stkem:
didn’t work for tags with spaces in them
@yipingliao


kifi-backend/modules/shoebox/angular/src/tags/tagList.js
```diff
@@ -26,7 +26,14 @@ angular.module('kifi')
         scope.shouldGiveFocus = false;
 
         scope.searchTagUrl = function (tag) {
-          return '/find?q=tag:' + encodeURIComponent(tag);
+          var tagPath;
```
yipingliao, line 29: `var tagPath = (tag.indexOf(' ') === -1) ? tag : '"' + tag + '"';`

PR #11925, opened 2014-10-27
alerting us whenever a username changes. At this point it should not hap...

eishay:
...pen unless its an admin call


eishay:
@andrewconner fyi

PR #11926, opened 2014-10-27
Display email address when there is no affiliated name in library share widget.

yipingliao:
@andrewconner Found this when I went into that share widget.

https://app.asana.com/0/17575799252940/18783354695223


kifi-backend/modules/shoebox/angular/src/libraries/libraryShareSearch.js
```diff
@@ -130,13 +131,12 @@ angular.module('kifi')
                   result.name = (result.firstName || '') + (result.lastName ? ' ' + result.lastName : '');
                 }
 
-                if (opt_query) {
```
yipingliao, line 133: `result.emailFormatted` was `undefined` if there is no search query, resulting in the bug. :( Now we are formatting all emails; the formatted email will just be unchanged if there is no search query.

PR #11927, opened 2014-10-27
Refactor / optimize createFullLibraryInfo

leogrim:
@stkem @andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
@@ -268,70 +268,73 @@ class KeepsCommander @Inject() (
   }
 
   def decorateKeepsIntoKeepInfos(perspectiveUserIdOpt: Option[Id[User]], keeps: Seq[Keep], idealImageSize: ImageSize = KeepImageSize.Large.idealSize): Future[Seq[KeepInfo]] = {
-    val augmentationFuture = {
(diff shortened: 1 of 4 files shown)


stkem:
lgtm

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
```diff
@@ -268,70 +268,73 @@ class KeepsCommander @Inject() (
...
+            val others = augmentationInfoForKeep.keepersTotal - augmentationInfoForKeep.keepers.size - augmentationInfoForKeep.keepersOmitted
+            val keepers = perspectiveUserIdOpt.map { userId => augmentationInfoForKeep.keepers.filterNot(_ == userId) } getOrElse augmentationInfoForKeep.keepers
+            KeepInfo(
+              id = Some(keep.externalId),
+              title = keep.title,
+              url = keep.url,
+              isPrivate = keep.isPrivate,
+              createdAt = Some(keep.createdAt),
+              others = Some(others),
+              keeps = allMyKeeps.get(keep.uriId),
```
yingjieMiao, line 319: I understand this is used in the frontend to determine keep's "global" visibility. Does this field have other use cases? Just curious.

leogrim, line 319: I believe it's also used to display the library drop down

PR #11931, opened 2014-10-28
fix bug for keep titles not working when kept from Keep Cards

ahsu1230:
@andrewconner @yipingliao 

When keeping from Keep Card, we only pass in keep URL. Title was not included.

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/keeps/keepActionService.js
+++ b/kifi-backend/modules/shoebox/angular/src/keeps/keepActionService.js
@@ -145,7 +145,7 @@ angular.module('kifi')
       });
     }
 
-    function keepToLibrary(keepUrls, libraryId) {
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/shoebox/angular/src/keeps/keepActionService.js
```diff
@@ -154,10 +154,17 @@ angular.module('kifi')
       });
 
       var data = {
-        keeps: keepUrls.map(function (keepUrl) {
-          return {
-            url: sanitizeUrl(keepUrl)
-          };
+        keeps: keepInfos.map(function (keep) {
```
yipingliao: `var keepData = { url: sanitizeUrl(keep.url) };`
`if (keep.title) { keepData.title = keep.title };`
`return keepData;`

yipingliao:
:+1:

PR #11936, opened 2014-10-28
alerting for username change only if the user creation time is more then...

eishay:
... one hour ago


eishay:
@andrewconner

andrewconner:
Thanks

PR #11940, opened 2014-10-28
quick fix for keep to library widget

ahsu1230:
@yipingliao @andrewconner 

There was a bug for keeping to a newly created library using the library widget (from a keep card) reported by Danny that needed to be quickly fix

The `scope.librarySelection` couldn't find the new library id and just had an error & hung.

This is a quick fix - happy to do things better in a future PR!!!!!!!!

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/keep/keep.js
+++ b/kifi-backend/modules/shoebox/angular/src/keep/keep.js
@@ -944,7 +944,7 @@ angular.module('kifi')
         // Scope methods.
         //
         scope.clickAction = function () {
-          if (scope.librarySelection.library.keptTo) {
+          if (scope.librarySelection.library && scope.librarySelection.library.keptTo) {
             var keepToUnkeep = _.find(scope.keep.keeps, { libraryId: scope.librarySelection.library.id });
             keepActionService.unkeepFromLibrary(scope.librarySelection.library.id, keepToUnkeep.id).then(function () {
               if (scope.librarySelection.library.id === scope.keep.libraryId) {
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/angular/src/keep/keep.js
```diff
@@ -944,7 +944,7 @@ angular.module('kifi')
         // Scope methods.
         //
         scope.clickAction = function () {
-          if (scope.librarySelection.library.keptTo) {
+          if (scope.librarySelection.library && scope.librarySelection.library.keptTo) {
```
ahsu1230, line 947: before, library field was undefined, so this part would crash
but now that I think about it, I think we don't want this to fail silently, so I think I'll change it back later

kifi-backend/modules/shoebox/angular/src/keep/keepToLibraryWidget.js
```diff
@@ -298,8 +298,23 @@ angular.module('kifi')
...
+
+                libraries.forEach(function (library) {
+                  library.keptTo = false;
+                  if (_.indexOf(scope.keptToLibraries, library.id) !== -1) {
+                    library.keptTo = true;
+                  }
+                });
+
+                libraries[selectedIndex].selected = true;
+                scope.widgetLibraries = libraries;
```
ahsu1230, line 317: this code is used in another place in the file. Will create a new function for this

yipingliao:
Thanks for working on this! I have a couple of questions and will come by tomorrow to ask.

PR #11943, opened 2014-10-28
tag-keep cache invalidation and prevent returning inactive tags for a keep

joshm1:
@andrewconner @leogrim

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/model/KeepToCollection.scala
+++ b/kifi-backend/modules/common/app/com/keepit/model/KeepToCollection.scala
@@ -22,7 +22,7 @@ case class KeepToCollection(
 }
 
 case class CollectionsForKeepKey(keepId: Id[Keep]) extends Key[Seq[Id[Collection]]] {
-  override val version = 3
+  override val version = 4
   val namespace = "collections_for_bookmark"
   def toKey(): String = keepId.toString
 }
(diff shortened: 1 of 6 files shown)


kifi-backend/modules/common/app/com/keepit/model/KeepToCollection.scala
```diff
@@ -22,7 +22,7 @@ case class KeepToCollection(
 }
 
 case class CollectionsForKeepKey(keepId: Id[Keep]) extends Key[Seq[Id[Collection]]] {
-  override val version = 3
+  override val version = 4
```
joshm1, line 25: to invalidate caches that include inactive tags

kifi-backend/modules/shoebox/test/com/keepit/model/CollectionTest.scala
```diff
@@ -123,7 +123,9 @@ class CollectionTest extends Specification with CommonTestInjector with DbInject
         }
 
         db.readWrite { implicit s =>
-          keepToCollectionRepo.getCollectionsForKeep(bookmark1.id.get).foreach(collectionRepo.collectionChanged(_, inactivateIfEmpty = false))
+          keepToCollectionRepo.getByKeep(bookmark1.id.get).groupBy(_.collectionId).foreach {
+            case (collectionId, _) => collectionRepo.collectionChanged(collectionId, inactivateIfEmpty = false)
+          }
```
joshm1, line 128: the previous version actually depended on the CollectionsForKeepCache being stale for this test to work. this change only bypasses the cache to invalidate the collection caches related to this keep (bookmark1).

The problem is that setting a keep to inactive does not automatically invalidate the CollectionsForKeepCaches, and that has to be done manually. We could do this on keepRepo.save, but I don't want to bloat the repo and there is certainly a performance tradeoff to invalidate all related caches at this low of a level (for instance, the cache invalidates are unnecessary if we change anything other than the `state` column).

andrewconner:
Looks good

PR #11948, opened 2014-10-28
update state token handling

raymondng:
To keep up with latest changes from Google
1. Add state token (short TTL) cache for tokens
2. Removed dependency on session state
3. Simplified token handling

Validated locally (yay dev mode!) Need to figure out a way to automate this (like social login ...)

Code can use some streamlining. Refactor PR is next.

@andrewconner


raymondng:
Merging to build on it. Feedback appreciated.

PR #11954, opened 2014-10-28
revert last change with endpoint & change front-end to call getMemberLis...

ahsu1230:
...t from offset 0
- will now include email addresses & invitees (even for libraries with < 10 followers)

@andrewconner @yipingliao


andrewconner:
Awesome thanks!

kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.js
```diff
@@ -194,7 +194,7 @@ angular.module('kifi')
 
         function setMemberStatus (member) {
           if (member.lastInvitedAt) {
-            return 'Inivation Pending';
+            return 'Invitation Pending';
```
ahsu1230, line 197: couldn't spell before lol >.<

PR #11956, opened 2014-10-28
Correct usage of "keep.keepersTotal" in the web site code.

yipingliao:
@joshm1 Thanks for pointing this out. I'd like to also get a little help in reviewing this code in action to make sure that it works as expected.
https://app.asana.com/0/17575799252940/18870880724767

For reference, also see the commit that replaced "keep.others" with "keepersTotal" last week:
https://github.com/kifi/keepit/commit/266a885a1048ff75d31efdeacea9ab89fbaf71a1#diff-a08839942899da9f448d8d67c39afe47

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/search/searchActionService.js
+++ b/kifi-backend/modules/shoebox/angular/src/search/searchActionService.js
@@ -17,7 +17,7 @@ angular.module('kifi')
       _.extend(hit, hit.bookmark); //still need this??
 
       hit.isPrivate = hit.secret || false;
-      hit.others = hit.keepersTotal - hit.keepers.length;
+      hit.others = hit.keepersTotal - hit.keepers.length - hit.keepersOmitted;
       hit.isProtected = !hit.isMyBookmark; // will not be hidden if user keeps then unkeeps
     }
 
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/shoebox/angular/src/search/searchActionService.js
```diff
@@ -17,7 +17,7 @@ angular.module('kifi')
       _.extend(hit, hit.bookmark); //still need this??
 
       hit.isPrivate = hit.secret || false;
-      hit.others = hit.keepersTotal - hit.keepers.length;
+      hit.others = hit.keepersTotal - hit.keepers.length - hit.keepersOmitted;
```
yipingliao: Looks like search already calculates its own `others` field, but did not take into account `keepersOmitted`.

joshm1: you need to take into account whether the current user kept this as well, which is not included in hit.keepers.length.  Do you may need to subtract 1 if this user has kept this.

yipingliao:
Updated per @joshm1's comments, and restored `keep.keepers` per discussion with @stkem

PR #11958, opened 2014-10-28
less airbrake messages on indexing errors

ymatsuda:
- increment errorCount on error
- reset errorCount after successful indexing
- ignore an error when errorCount is zero

@leogrim @yingjieMiao @eishay


ymatsuda:
So, we send airbrake messages only when we have two or more consecutive indexing errors (per index).

yingjieMiao:
cool

ymatsuda:
Having false alarms for every shoebox deployment is not cool, so this may reduce them.

PR #11959, opened 2014-10-28
making sure slug generation code always returns a valid slug

2is10:
@andrewconner


andrewconner:
Thanks Jared, works for me

PR #11960, opened 2014-10-28
Search refactor

andrewconner:
Fixes a few issues we've been having with the header. SearchCtrl now stays around as you type your search query. The UX should be a bit faster — keeps are only replaced when new results are available.

@stkem @ahsu1230 @yingjieMiao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/layout/header/HeaderCtrl.js
+++ b/kifi-backend/modules/shoebox/angular/src/layout/header/HeaderCtrl.js
@@ -53,9 +53,7 @@ angular.module('kifi')
     $scope.$on('$destroy', deregisterLibraryChip);
 
     $scope.$on('$routeUpdate', function (event, current) {
-      console.log('called6')
       if (current.params.q) {
-        console.log('called7')
         $scope.search.text = current.params.q;
       }
     });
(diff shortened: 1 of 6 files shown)


kifi-backend/modules/shoebox/angular/src/search/search.tpl.html
```diff
@@ -25,16 +25,4 @@
     toggle-edit="toggleEdit"
     update-selected-count="updateSelectedCount(numSelected)"
   ></div>
-  <div ng-if="!librariesEnabled"
```
yipingliao, line 28: Should we start taking out these as we encounter them in our bug fixes?

kifi-backend/modules/shoebox/angular/src/layout/header/HeaderCtrl.js
```diff
@@ -52,12 +52,13 @@ angular.module('kifi')
     });
     $scope.$on('$destroy', deregisterLibraryChip);
 
-    $scope.$on('$routeChangeSuccess', function (event, current) {
+    $scope.$on('$routeUpdate', function (event, current) {
```
ahsu1230, line 55: Oh, so this calls even for page refreshes?

ahsu1230:
:+1: thanks!

kifi-backend/modules/shoebox/angular/src/search/search.tpl.html
stkem, line 28: Yes, I think we should

stkem:
lgtm.
merging now to reduce merge conflict

kifi-backend/modules/shoebox/angular/src/search/search.tpl.html
andrewconner, line 28: Let's do it.

ahsu1230, line 28: :ok_hand:

PR #11963, opened 2014-10-28
[Do not merge] Introduce UsernameAlias and LibraryAlias

leogrim:
@andrewconner @eishay 
Models + Repos for library redirections
I need to investigate a search bug, we'll get on the tracking logic when I'm done.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/model/LibraryAlias.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/model/LibraryAlias.scala
@@ -51,12 +51,22 @@ class LibraryAliasRepoImpl @Inject() (
 
   override def deleteCache(slugAlias: LibraryAlias)(implicit session: RSession): Unit = {}
 
-  private def compiledGetByUserIdAndSlug(ownerId: Id[User], slug: LibrarySlug, excludeState: Option[State[LibraryAlias]]) = Compiled {
-    for (row <- rows if row.ownerId === ownerId && row.slug === slug && row.state =!= excludeState.orNull) yield row
+  private val compiledGetByUserIdAndSlug = Compiled {
+    (ownerId: Column[Id[User]], slug: Column[LibrarySlug]) =>
+      for (row <- rows if row.ownerId === ownerId && row.slug === slug) yield row
+  }
+
+  private val compiledGetByUserIdAndSlugAndExcludedState = Compiled {
+    (ownerId: Column[Id[User]], slug: Column[LibrarySlug], excludedState: Column[State[LibraryAlias]]) =>
+      for (row <- rows if row.ownerId === ownerId && row.slug === slug && row.state =!= excludedState) yield row
   }
 
   def getByOwnerIdAndSlug(ownerId: Id[User], slug: LibrarySlug, excludeState: Option[State[LibraryAlias]] = Some(LibraryAliasStates.INACTIVE))(implicit session: RSession): Option[LibraryAlias] = {
-    compiledGetByUserIdAndSlug(ownerId, slug, excludeState).firstOption
+    val q = excludeState match {
+      case Some(state) => compiledGetByUserIdAndSlugAndExcludedState(ownerId, slug, state)
+      case None => compiledGetByUserIdAndSlug(ownerId, slug)
+    }
+    q.firstOption
   }
 
   def alias(ownerId: Id[User], slug: LibrarySlug, libraryId: Id[Library])(implicit session: RWSession): LibraryAlias = {
--- a/kifi-backend/modules/shoebox/app/com/keepit/model/LibraryAlias.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/model/LibraryAlias.scala
@@ -26,7 +26,7 @@ object LibraryAliasStates extends States[LibraryAlias]
 @ImplementedBy(classOf[LibraryAliasRepoImpl])
 trait LibraryAliasRepo extends Repo[LibraryAlias] {
   def getByOwnerIdAndSlug(ownerId: Id[User], slug: LibrarySlug, excludeState: Option[State[LibraryAlias]] = Some(LibraryAliasStates.INACTIVE))(implicit session: RSession): Option[LibraryAlias]
-  def intern(ownerId: Id[User], slug: LibrarySlug, libraryId: Id[Library])(implicit session: RWSession): LibraryAlias
+  def alias(ownerId: Id[User], slug: LibrarySlug, libraryId: Id[Library])(implicit session: RWSession): LibraryAlias
 }
 
 @Singleton
@@ -59,14 +59,17 @@ class LibraryAliasRepoImpl @Inject() (
     compiledGetByUserIdAndSlug(ownerId, slug, excludeState).firstOption
   }
 
-  def intern(ownerId: Id[User], slug: LibrarySlug, libraryId: Id[Library])(implicit session: RWSession): LibraryAlias = {
+  def alias(ownerId: Id[User], slug: LibrarySlug, libraryId: Id[Library])(implicit session: RWSession): LibraryAlias = {
     getByOwnerIdAndSlug(ownerId, slug, excludeState = None) match {
       case None => save(LibraryAlias(ownerId = ownerId, slug = slug, libraryId = libraryId))
-      case Some(inactiveAlias) if inactiveAlias.state == LibraryAliasStates.INACTIVE => {
-        save(inactiveAlias.copy(createdAt = clock.now, updatedAt = clock.now, state = LibraryAliasStates.ACTIVE, libraryId = libraryId))
+      case Some(alias) if alias.state == LibraryAliasStates.INACTIVE => {
+        val requestedAlias = alias.copy(createdAt = clock.now, updatedAt = clock.now, state = LibraryAliasStates.ACTIVE, slug = slug, libraryId = libraryId)
+        save(requestedAlias)
       }
-      case Some(validAlias) if validAlias.libraryId == libraryId => validAlias
-      case Some(invalidAlias) => save(invalidAlias.copy(libraryId = libraryId))
+      case Some(activeAlias) => {
+        val requestedAlias = activeAlias.copy(slug = slug, libraryId = libraryId)
+        if (requestedAlias == activeAlias) requestedAlias else save(requestedAlias)
+      }
     }
   }
 
--- a/kifi-backend/modules/shoebox/app/com/keepit/model/LibraryAlias.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/model/LibraryAlias.scala
@@ -62,13 +62,9 @@ class LibraryAliasRepoImpl @Inject() (
   def alias(ownerId: Id[User], slug: LibrarySlug, libraryId: Id[Library])(implicit session: RWSession): LibraryAlias = {
     getByOwnerIdAndSlug(ownerId, slug, excludeState = None) match {
       case None => save(LibraryAlias(ownerId = ownerId, slug = slug, libraryId = libraryId))
-      case Some(alias) if alias.state == LibraryAliasStates.INACTIVE => {
-        val requestedAlias = alias.copy(createdAt = clock.now, updatedAt = clock.now, state = LibraryAliasStates.ACTIVE, slug = slug, libraryId = libraryId)
-        save(requestedAlias)
-      }
-      case Some(activeAlias) => {
-        val requestedAlias = activeAlias.copy(slug = slug, libraryId = libraryId)
-        if (requestedAlias == activeAlias) requestedAlias else save(requestedAlias)
+      case Some(existingAlias) => {
+        val requestedAlias = existingAlias.copy(state = LibraryAliasStates.ACTIVE, slug = slug, libraryId = libraryId)
+        if (requestedAlias == existingAlias) requestedAlias else save(requestedAlias)
       }
     }
   }
--- a/kifi-backend/modules/shoebox/app/com/keepit/model/LibraryAlias.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/model/LibraryAlias.scala
@@ -1,0 +1,73 @@
+package com.keepit.model
+
+import com.keepit.common.db._
+import org.joda.time.DateTime
+import com.keepit.common.time._
+import com.google.inject.{ ImplementedBy, Inject, Singleton }
+import com.keepit.common.db.slick._
+import com.keepit.common.logging.Logging
+import com.keepit.common.db.slick.DBSession.{ RWSession, RSession }
+
+case class LibraryAlias(
+    id: Option[Id[LibraryAlias]] = None,
+    createdAt: DateTime = currentDateTime,
+    updatedAt: DateTime = currentDateTime,
+    state: State[LibraryAlias] = LibraryAliasStates.ACTIVE,
+    ownerId: Id[User],
+    slug: LibrarySlug,
+    libraryId: Id[Library]) extends ModelWithState[LibraryAlias] {
+  def withId(id: Id[LibraryAlias]) = this.copy(id = Some(id))
+  def withUpdateTime(now: DateTime) = this.copy(updatedAt = now)
+  def isActive = (state == LibraryAliasStates.ACTIVE)
+}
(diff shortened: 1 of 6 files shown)


kifi-backend/modules/shoebox/app/com/keepit/model/UsernameAlias.scala
```diff
@@ -0,0 +1,104 @@
...
+  def intern(username: Username, userId: Id[User])(implicit session: RWSession): UsernameAlias = {
+    val normalizedUsername = normalize(username)
+    getByNormalizedUsername(normalizedUsername, excludeState = None) match {
+      case None => save(UsernameAlias(username = normalizedUsername, userId = userId))
+      case Some(inactiveAlias) if inactiveAlias.state == UsernameAliasStates.INACTIVE => {
+        save(inactiveAlias.copy(createdAt = clock.now, updatedAt = clock.now, state = UsernameAliasStates.ACTIVE, username = normalizedUsername, userId = userId))
+      }
+      case Some(validAlias) if validAlias.userId == userId => validAlias
+      case Some(reservedAlias) if reservedAlias.isReserved => throw ReservedUsernameException(reservedAlias)
+      case Some(unreservedAlias) => save(unreservedAlias.copy(userId = userId))
```
2is10: I found this naming convention a little confusing, since the names are applied before they’re known to be appropriate. For example, consider the clause `if reservedAlias.isReserved` or `if inactiveAlias.state == INACTIVE`. I’d suggest just naming each one `alias`. They’re very narrowly scoped.

kifi-backend/modules/shoebox/app/com/keepit/model/UsernameAlias.scala
```diff
@@ -0,0 +1,104 @@
...
+    state: State[UsernameAlias] = UsernameAliasStates.ACTIVE,
+    username: Username, // normalized
+    userId: Id[User]) extends ModelWithState[UsernameAlias] {
+  def withId(id: Id[UsernameAlias]) = this.copy(id = Some(id))
+  def withUpdateTime(now: DateTime) = this.copy(updatedAt = now)
+  def isReserved = (state == UsernameAliasStates.RESERVED)
+}
+
+object UsernameAliasStates extends States[UsernameAlias] {
+  val RESERVED = State[UsernameAlias]("reserved")
```
2is10: What’s the difference between `ACTIVE` and `RESERVED`? A brief why comment here would be very helpful.

kifi-backend/modules/shoebox/app/com/keepit/model/UsernameAlias.scala
```diff
@@ -0,0 +1,104 @@
...
+
+  private def normalize(username: Username): Username = {
+    Username(UsernameOps.normalize(username.value))
+  }
+
+  def getByUsername(username: Username, excludeState: Option[State[UsernameAlias]] = Some(UsernameAliasStates.INACTIVE))(implicit session: RSession): Option[UsernameAlias] = {
+    getByNormalizedUsername(normalize(username), excludeState)
+  }
+
+  def intern(username: Username, userId: Id[User])(implicit session: RWSession): UsernameAlias = {
```
2is10: I found the semantics of this method difficult to guess based on its name and type signature alone. A comment explaining that this will return an alias with state either `ACTIVE` or `RESERVED`, depending on whether the row already existed in one of those states, would be helpful. Alternatively, consider taking desired state as an argument.
(I suppose I was surprised or found it complex because it’s unusual to have a method that both writes and tries to read and preserve preexisting information.)

2is10:
Nice work. This is a really important feature.

kifi-backend/modules/shoebox/app/com/keepit/model/UsernameAlias.scala
yingjieMiao: +1 for @2is10  's naming suggestions.

When we find `Some(alias)`, then based on `userId` and `state`, there are `2 x 3 = 6` case branches. I understand that some branches yield same action. It might be helpful to reorder the matching sequence (if no logic change). Right now, it's not super straightforward to navigate the decision tree. For example, in the last match,
what is the state of the `unreservedAlias`?  `Active`? Then we are overwriting it with a different `userId`?

ymatsuda:
What are aliases? How are they used?
Is there any implication for search?

kifi-backend/modules/shoebox/app/com/keepit/model/UsernameAlias.scala
```diff
@@ -0,0 +1,104 @@
...
+  val RESERVED = State[UsernameAlias]("reserved")
+}
+
+@ImplementedBy(classOf[UsernameAliasRepoImpl])
+trait UsernameAliasRepo extends Repo[UsernameAlias] {
+  def getByUsername(username: Username, excludeState: Option[State[UsernameAlias]] = Some(UsernameAliasStates.INACTIVE))(implicit session: RSession): Option[UsernameAlias]
+  def intern(username: Username, userId: Id[User])(implicit session: RWSession): UsernameAlias
+  def reserve(username: Username, userId: Id[User])(implicit session: RWSession): Unit
+  def release(username: Username)(implicit session: RWSession): Boolean
+}
```
yingjieMiao, line 39: from an API user's perspective, it's not clear what these 3 methods do (without digging into the source code). Are there special ordering to call these methods? If so,  what kind of exceptions can happen if they are not called in correct sequence? Some documentation could be helpful for future maintenance.

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryAlias.scala
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  def intern(ownerId: Id[User], slug: LibrarySlug, libraryId: Id[Library])(implicit session: RWSession): LibraryAlias = {
+    getByOwnerIdAndSlug(ownerId, slug, excludeState = None) match {
+      case None => save(LibraryAlias(ownerId = ownerId, slug = slug, libraryId = libraryId))
+      case Some(inactiveAlias) if inactiveAlias.state == LibraryAliasStates.INACTIVE => {
+        save(inactiveAlias.copy(createdAt = clock.now, updatedAt = clock.now, state = LibraryAliasStates.ACTIVE, libraryId = libraryId))
+      }
+      case Some(validAlias) if validAlias.libraryId == libraryId => validAlias
+      case Some(invalidAlias) => save(invalidAlias.copy(libraryId = libraryId))
```
yingjieMiao: in what case we end up in this last case match? We lose the track of `libraryId` here?

leogrim:
@2is10 @yingjieMiao I have refactored both the code and corresponding tests with your feedback, I hope this is clearer.

@ymatsuda see task https://app.asana.com/0/5388476470807/18778727426533
To my understanding, we want to be robust (= respond with a 301 redirect) against:
- a change in user username 
- a change in library slug
- a change in library owner
  This is about routing to libraries in Shoebox, I don't think it impacts Search.

This has implications in terms of username policy, which hasn't been defined. I tried to make the best out of it with UsernameAlias, which can handle locked / permanent / hard aliases ("reserved") or soft aliases (which can be reclaimed, possibly after some period of time so that existing 301 can be detected by crawlers)

2is10:
This is definitely clearer. Thanks.

In general, for functionality that requires domain model changes, I think it’s a good idea to press product to get clear requirements before coding. It can be a collaborate process, where we propose ideas. This “reserved” mechanism seems useful, but I imagine it will only be directly exposed to admins, and we may eventually need cron jobs to flip it. I can imagine us having a policy that old usernames are “reserved” for a period of N days, after which they become fair game for others. In that case, we might choose to store a last-activated date instead of using a third state.

kifi-backend/modules/shoebox/app/com/keepit/model/UsernameAlias.scala
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  def alias(username: Username, userId: Id[User], reserve: Boolean = false)(implicit session: RWSession): UsernameAlias = {
+    val normalizedUsername = normalize(username)
+    val requestedState = if (reserve) RESERVED else ACTIVE
+    getByNormalizedUsername(normalizedUsername, excludeState = None) match {
+      case Some(alias) if alias.isReserved => if (alias.userId == userId) alias else throw ReservedUsernameException(alias) // reserved aliases must be explicitly released
+      case Some(availableAlias) if availableAlias.state == INACTIVE => {
+        val requestedAlias = availableAlias.copy(createdAt = clock.now, updatedAt = clock.now, state = requestedState, username = normalizedUsername, userId = userId)
+        save(requestedAlias)
```
yingjieMiao: 1. do we need to set `username` here? Seems that it won't change. 
2. no need to explicitly set `updatedAt`. It will be overwritten again in the save method.
3. Why here we set `createdAt`, but in Line 88 we don't?

kifi-backend/modules/shoebox/app/com/keepit/model/UsernameAlias.scala
```diff
@@ -0,0 +1,103 @@
...
+    username: Username, // normalized
+    userId: Id[User]) extends ModelWithState[UsernameAlias] {
+  def withId(id: Id[UsernameAlias]) = this.copy(id = Some(id))
+  def withUpdateTime(now: DateTime) = this.copy(updatedAt = now)
+  def isReserved = (state == UsernameAliasStates.RESERVED)
+}
+
+object UsernameAliasStates extends States[UsernameAlias] {
+  val RESERVED = State[UsernameAlias]("reserved") // while an 'active' alias can be claimed by another user (soft alias), a 'reserved' alias cannot (hard alias)
+}
```
yingjieMiao, line 32: Minor: from the methods below, looks like `inactive` alias can be claimed by another user as well. It maybe helpful to explicitly have a `isAvailable` method defined in the `UsernameAlias` class.

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryAlias.scala
```diff
@@ -0,0 +1,76 @@
...
+      case None => save(LibraryAlias(ownerId = ownerId, slug = slug, libraryId = libraryId))
+      case Some(alias) if alias.state == LibraryAliasStates.INACTIVE => {
+        val requestedAlias = alias.copy(createdAt = clock.now, updatedAt = clock.now, state = LibraryAliasStates.ACTIVE, slug = slug, libraryId = libraryId)
+        save(requestedAlias)
+      }
+      case Some(activeAlias) => {
+        val requestedAlias = activeAlias.copy(slug = slug, libraryId = libraryId)
+        if (requestedAlias == activeAlias) requestedAlias else save(requestedAlias)
+      }
+    }
```
yingjieMiao, line 79: I imagine for the same user, he could overwrite the `slug` and `libraryId` arbitrarily many times in arbitrary ways? Seems that we do not (want to) track the whole history?

yingjieMiao:
some minor comments. Looks great overall!

kifi-backend/modules/shoebox/app/com/keepit/model/UsernameAlias.scala
leogrim: 1.2.3. "inactive" should be semantically equivalent to non existent, so this is to make sure that the case Some(inactiveAlias) is analog to case None.
In particular, resetting the username is meant to cope with possible case-sensitivity changes in the normalization process (since the database is case-insensitive).

kifi-backend/modules/shoebox/app/com/keepit/model/UsernameAlias.scala
leogrim, line 32: Well, isAvailable = !isReserved

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryAlias.scala
leogrim, line 79: Well if the same path ever points to two or more different libraries (e.g. the same user changes a library's slug and then creates a new library with the former slug), how would we know which one is requested by the caller? It seems like the reasonable thing to do it to direct towards the most recent library associated with that path.

leogrim:
@2is10 I asked for product's involvement on Monday, and didn't attract much interest. Per discussion with @eishay and @andrewconner , we came up with some requirements that I tried to formalize in Asana. 

Specifically, we want to be able to lock an alias for an arbitrary period of time (e.g. for an influencer), hence the state. Now for unreserved / soft aliases, I was actually thinking of using "updatedAt" to protect them for a small given period of time before actually making them fair game (so that crawlers can update). An alternative could be a `lastActivatedAt` DateTime field + a `reservation` Duration that could potentially be be infinity.

2is10:
I like the `lastActivatedAt` idea. Our conventions are that `createdAt` should never be updated and `updatedAt` should not be relied upon for business logic.

kifi-backend/modules/shoebox/app/com/keepit/model/UsernameAlias.scala
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  def table(tag: Tag) = new UsernameAliasTable(tag)
+  initTable
+
+  override def invalidateCache(usernameAlias: UsernameAlias)(implicit session: RSession): Unit = {}
+
+  override def deleteCache(usernameAlias: UsernameAlias)(implicit session: RSession): Unit = {}
+
+  private def compiledGetByNormalizedUsername(normalizedUsername: Username, excludeState: Option[State[UsernameAlias]]) = Compiled {
```
andrewconner: Somewhere this pattern leaked in our codebase, but this doesn't compile the query once. Needs to be a `val` not `def`.

kifi-backend/modules/shoebox/app/com/keepit/model/UsernameAlias.scala
```diff
@@ -0,0 +1,103 @@
...
+
+  def getByUsername(username: Username, excludeState: Option[State[UsernameAlias]] = Some(INACTIVE))(implicit session: RSession): Option[UsernameAlias] = {
+    getByNormalizedUsername(normalize(username), excludeState)
+  }
+
+  def alias(username: Username, userId: Id[User], reserve: Boolean = false)(implicit session: RWSession): UsernameAlias = {
+    val normalizedUsername = normalize(username)
+    val requestedState = if (reserve) RESERVED else ACTIVE
+    getByNormalizedUsername(normalizedUsername, excludeState = None) match {
+      case Some(alias) if alias.isReserved => if (alias.userId == userId) alias else throw ReservedUsernameException(alias) // reserved aliases must be explicitly released
```
andrewconner: Would be nice not to throw an exception, but return a failure state. Perhaps an `Either` or `Try`?

kifi-backend/modules/shoebox/app/com/keepit/model/UsernameAlias.scala
leogrim: Oh, makes sense, will fix.

kifi-backend/modules/shoebox/app/com/keepit/model/UsernameAlias.scala
leogrim: Sure, I can wrap this in a Try.

andrewconner: Better to return a `Failure` vs letting Try catch the exception.

leogrim:
@2is10 @andrewconner 
Wow, I was not aware of these conventions - better late than never :) 

My understanding: an `INACTIVE` row was to behave like a deleted row. Then the state of the database after reactivating an inactive row should be the same as the state of the database after inserting that same row, hence the resetting of `createdAt` and `updatedAt`. 

Why this understanding: for instance, we're relying on 'createdAt' to tell users when they kept a page. If you unkeep a page, and then keep it again months later, we wouldn't want to show the old date. So I guess we don't respect the conventions in some places already. In this instance, we should have an explicit `lastKeptAt` DateTime field in `Keep`. 

So in the case at hand, I want a `lastActivatedAt` (so that I don't rely on createdAt or updatedAt for the expiration business logic) _and_ a `RESERVED` state (to exclude certain aliases from the expiration policy)

leogrim:
Thanks for your comments, I'll close the pull request and send a one with the rest of the logic later on.

PR #11964, opened 2014-10-28
when keeping to main vs. private library, keep count of src library is o...

ahsu1230:
...ff (caching issue)

If Main Library had 5 keeps and Private Library had 8 keeps...
Let's say I kept a keep in Main Library (now has 6 keeps)
Then I kept the same uri in Private Library
The count of main library should be back to 5, while Private Library has 9
Even after refresh, the destination library count does go up, but the source library does not decrement & will only go back to it's correct number when you do an operation that will change the count cache.

@andrewconner @2is10

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepInterner.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepInterner.scala
@@ -202,8 +202,12 @@ class KeepInterner @Inject() (
     val (isNewKeep, wasInactiveKeep, internedKeep) = currentBookmarkOpt match {
       case Some(bookmark) =>
         val wasInactiveKeep = !bookmark.isActive
-        if (bookmark.isActive && bookmark.inDisjointLib) // invalidate if keep URI is active in a system library
+        if (bookmark.isActive && bookmark.inDisjointLib && bookmark.libraryId.get != library.id.get) {
+          // invalidate if keep URI is active in a system library
           countByLibraryCache.remove(CountByLibraryKey(bookmark.libraryId.get))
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepInterner.scala
```diff
@@ -201,6 +202,8 @@ class KeepInterner @Inject() (
     val (isNewKeep, wasInactiveKeep, internedKeep) = currentBookmarkOpt match {
       case Some(bookmark) =>
         val wasInactiveKeep = !bookmark.isActive
+        if (bookmark.isActive && bookmark.inDisjointLib) // invalidate if keep URI is active in a system library
+          countByLibraryCache.remove(CountByLibraryKey(bookmark.libraryId.get))
```
ahsu1230, line 207: placed it here in `KeepInterner` since this is the part that deals with `disjointLib` field

2is10, line 207: Is the `CountByLibraryCache` entry for the kept-to library also invalidated? Might be clearer if this were done near where that’s done.

Also, it might be a good idea to check that the destination library is different from the current library before invalidating anything. For various reasons, sometimes we get duplicate (redundant, idempotent) action requests.

ahsu1230, line 207: put in an additional check to make sure dst library & src library are different.
The destination library cache will also get invalidated, but I think the best place is still here since here is where adding keeps involves taking a keep out of another one

PR #11968, opened 2014-10-29
Mobile libraries look pretty now

andrewconner:
@yipingliao Mark and I sat down to do a CSS-only change to mobile public libraries. This "works" by overriding existing styles; it's hacky and may break when we do routine changes to the site. However, it's a short term fix until we can split templates.

I built everything inside of an auto-added `.kf-mobile` class, and used as high specificity as reasonably possible. A few places, I had to do some unclean things to end up with higher specificity. We'll improve as we go.


eishay:
yey!

yipingliao:
Yay!

PR #11974, opened 2014-10-29
bugs with share library search & keep to library from keep card

ahsu1230:
@andrewconner @yipingliao 

Share Library Search showed users who were invited (but not accepted) as Following
Keep to Library (from Keep Card) now persists tags - uses copyKEeps instead of addKeeps

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/keep/keep.js
+++ b/kifi-backend/modules/shoebox/angular/src/keep/keep.js
@@ -958,9 +958,9 @@ angular.module('kifi')
             });
           } else {
             var keepInfo = { title: scope.keep.title, url: scope.keep.url };
-            keepActionService.keepToLibrary([keepInfo], scope.librarySelection.library.id).then(function (result) {
-              if ((!result.failures || !result.failures.length) && result.alreadyKept.length === 0) {
-                return keepActionService.fetchFullKeepInfo(result.keeps[0]).then(function (fullKeep) {
+            keepActionService.copyToLibrary([scope.keep.id], scope.librarySelection.library.id).then(function (result) {
+              if (result.successes > 0) {
+                return keepActionService.fetchFullKeepInfo(scope.keep).then(function (fullKeep) {
                   libraryService.fetchLibrarySummaries(true);
                   libraryService.addToLibraryCount(scope.librarySelection.library.id, 1);
                   tagService.addToKeepCount(1);
--- a/kifi-backend/modules/shoebox/angular/src/keep/keep.js
+++ b/kifi-backend/modules/shoebox/angular/src/keep/keep.js
@@ -957,7 +957,6 @@ angular.module('kifi')
               libraryService.addToLibraryCount(scope.librarySelection.library.id, -1);
             });
           } else {
-            var keepInfo = { title: scope.keep.title, url: scope.keep.url };
             keepActionService.copyToLibrary([scope.keep.id], scope.librarySelection.library.id).then(function (result) {
               if (result.successes > 0) {
                 return keepActionService.fetchFullKeepInfo(scope.keep).then(function (fullKeep) {
--- a/kifi-backend/modules/shoebox/angular/src/libraries/libraryShareSearch.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/libraryShareSearch.js
@@ -129,7 +129,7 @@ angular.module('kifi')
                 if (result.id) {
                   result.image = friendService.getPictureUrlForUser(result);
                   result.isInvited = !!result.lastInvitedAt;
-                  result.isFollowing = !!result.membership & !result.isInvited;
+                  result.isFollowing = !!result.membership && !result.isInvited;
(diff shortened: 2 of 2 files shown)


kifi-backend/modules/shoebox/angular/src/keep/keep.js
```diff
@@ -957,10 +957,9 @@ angular.module('kifi')
               libraryService.addToLibraryCount(scope.librarySelection.library.id, -1);
             });
           } else {
-            var keepInfo = { title: scope.keep.title, url: scope.keep.url };
```
yipingliao, line 960: Are we losing the title information as a result of moving to this endpoint?

kifi-backend/modules/shoebox/angular/src/libraries/libraryShareSearch.js
```diff
@@ -128,7 +128,8 @@ angular.module('kifi')
               newResults.forEach(function (result) {
                 if (result.id) {
                   result.image = friendService.getPictureUrlForUser(result);
-                  result.isFollowing = !!result.membership;
+                  result.isInvited = !!result.lastInvitedAt;
```
yipingliao, line 131: Do we need an extra property on `result` that is just the boolean cast of `result.lastInvitedAt`? Also, a quick grep shows that `result.isInvited` is not used anywhere else in the codebase, so we can remove line 131 and just use `result.lastInvitedAt` in line 132. Finally, are there any indications that a user has been invited on the UI itself? Or does that user just show up as not following?

kifi-backend/modules/shoebox/angular/src/keep/keep.js
```diff
@@ -957,10 +957,9 @@ angular.module('kifi')
...
-            keepActionService.keepToLibrary([keepInfo], scope.librarySelection.library.id).then(function (result) {
-              if ((!result.failures || !result.failures.length) && result.alreadyKept.length === 0) {
-                return keepActionService.fetchFullKeepInfo(result.keeps[0]).then(function (fullKeep) {
+            keepActionService.copyToLibrary([scope.keep.id], scope.librarySelection.library.id).then(function (result) {
+              if (result.successes > 0) {
+                return keepActionService.fetchFullKeepInfo(scope.keep).then(function (fullKeep) {
```
yipingliao, line 962: With the old endpoint, `fetchFullKeepInfo` was called with the new keep (`result.keeps[0])` but now it looks like it's being called with the current keep (`scope.keep`).

kifi-backend/modules/shoebox/angular/src/keep/keep.js
ahsu1230, line 960: nope, copying keeps copies the URI id, URL, title, etc. etc.
And makes sure its libraryId & visibility is aligned with the destination library

kifi-backend/modules/shoebox/angular/src/keep/keep.js
ahsu1230, line 962: so the endpoint doesn't return keep Ids or anything, only number of successes & "failures". So yea, I guess fetching the current keep info doesn't make too much sense. Did it used to fetch the newly added keep? What did we use that information for?

kifi-backend/modules/shoebox/angular/src/keep/keep.js
yipingliao, line 960: Nice!

kifi-backend/modules/shoebox/angular/src/libraries/libraryShareSearch.js
ahsu1230, line 131: your right, we dont' need to add another field
But I kept it as following or not following, so we have a way to support multiple invites?
For now it's kinda bleh, but with collaborators with different access levels (invite first with read_only then invite for read_write), it makes more sense.

ahsu1230, line 131: what do you think?

kifi-backend/modules/shoebox/angular/src/keep/keep.js
yipingliao, line 962: I think that it's used to build a "full" keep with enough information to render the entire keep card. I am not sure if we still need it in this case. We obviously have a keep card already. Let's try taking it out entirely!

kifi-backend/modules/shoebox/angular/src/libraries/libraryShareSearch.js
yipingliao, line 131: Yes, I was also thinking that this way the user can invite again, which would be useful even today so that the invitee will get another email if they lost the first email or something.

kifi-backend/modules/shoebox/angular/src/keep/keep.js
yipingliao, line 962: Ugh, actually, we have a `keepAdded` event that is emitted. So we _will_ need the new keep.

PR #11982, opened 2014-10-29
Do not emit 'updateLoggedInState' event on initial page load.

yipingliao:
https://app.asana.com/0/17575799252940/18797430281236

@andrewconner


andrewconner:
Thanks for working with me to find the fix.

yipingliao:
Thank you for tracking this down! I am going to check that logging in still works when this is deployed.

PR #11983, opened 2014-10-29
Only use mobile styles on logged out clients

andrewconner:
@yipingliao If a user happens to be logged in, don't give them the new styles. They break the logged in site.


yipingliao:
:+1:

PR #11984, opened 2014-10-29
copy to library support for keep cards

joshm1:
@stkem 

cc @ahsu1230 since you're fixing bug(s) related to this

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/keeps/keeps.js
+++ b/kifi-backend/modules/shoebox/angular/src/keeps/keeps.js
@@ -287,7 +287,8 @@ angular.module('kifi')
         editMode: '=',
         editOptions: '&',
         toggleEdit: '=',
-        updateSelectedCount: '&'
+        updateSelectedCount: '&',
+        selectedKeepsFilter: '&'
       },
       controller: 'KeepsCtrl',
       templateUrl: 'keeps/keeps.tpl.html',
@@ -325,9 +326,15 @@ angular.module('kifi')
           }
         }
 
+        function getSelectedKeeps() {
+          var selectedKeeps = scope.selection.getSelected(scope.availableKeeps);
+          var filter = scope.selectedKeepsFilter();
+          return _.isFunction(filter) ? filter(selectedKeeps) : selectedKeeps;
+        }
+
         function copyToLibrary () {
           // Copies the keeps that are selected into the library that is selected.
-          var selectedKeeps = scope.selection.getSelected(scope.availableKeeps);
+          var selectedKeeps = getSelectedKeeps();
           var selectedLibrary = scope.librarySelection.library;
 
           keepActionService.copyToLibrary(_.pluck(selectedKeeps, 'id'), selectedLibrary.id).then(function () {
--- a/kifi-backend/modules/shoebox/angular/src/keeps/keeps.js
+++ b/kifi-backend/modules/shoebox/angular/src/keeps/keeps.js
@@ -285,6 +285,7 @@ angular.module('kifi')
         scrollDisabled: '=',
         scrollNext: '&',
         editMode: '=',
+        editOptions: '&',
         toggleEdit: '=',
         updateSelectedCount: '&'
       },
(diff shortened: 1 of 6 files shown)


kifi-backend/modules/shoebox/angular/src/keeps/keeps.js
```diff
@@ -324,9 +326,15 @@ angular.module('kifi')
           }
         }
 
+        function getSelectedKeeps() {
+          var selectedKeeps = scope.selection.getSelected(scope.availableKeeps);
+          var filter = scope.selectedKeepsFilter();
```
joshm1, line 331: `selectedKeepsFilter` is an optional function that can be added to the scope by the parent, this is useful for search since each card can represent multiple keeps to take action on

kifi-backend/modules/shoebox/angular/src/search/search.js
```diff
@@ -69,6 +70,10 @@ angular.module('kifi')
     //
     // Scope methods.
     //
+    $scope.selectedKeepsFilter = function (hits) {
```
yipingliao, line 73: Why is this function in `search.js` instead of `keeps.js`? It's only used in `keeps.js`. If it's only used to determine whether the keeps are on a search page, we can pass a boolean value like `onSearchPage` from `search` to `keeps`.

kifi-backend/modules/shoebox/angular/src/search/search.js
```diff
@@ -91,6 +96,9 @@ angular.module('kifi')
         var hits = result.hits;
 
         hits.forEach(function (hit) {
+          // hack a unique ID for each "hit" (URI) used in bulk edit for the hit (result) doesn't have one
+          hit.id = hit.id || 'h-' + (++hitCounter);
```
yipingliao: Do we still need this? I think you said that this was for selecting keeps because the old `selectionService` used ids instead of urls?

kifi-backend/modules/shoebox/angular/src/search/search.js
yipingliao, line 73: Changed my mind; I like the function in `search` better after I thought about it some more. :)

kifi-backend/modules/shoebox/angular/src/keeps/keeps.js
yipingliao, line 331: In general, try to keep down the number of local variables used. In this case it's more readable to not have the `filter` variable.

kifi-backend/modules/shoebox/angular/src/search/search.js
joshm1, line 73: This is an abstraction. I'd rather use this pattern than bloat keeps.js with an additional conditional & branch of execution. This will allow other callers to provider their own keep-selection filters as well.

kifi-backend/modules/shoebox/angular/src/keeps/keeps.js
```diff
@@ -324,9 +326,15 @@ angular.module('kifi')
...
+          return _.isFunction(filter) ? filter(selectedKeeps) : selectedKeeps;
```
yipingliao, line 332: `return _.isFunction(scope.selectedKeepsFilter) ? scope.selectKeepsFilter(selectedKeeps): selectedKeeps`
On line 331 `scope.selectedKeepsFilter` was actually invoked; not sure if that was intended. Or is it that because it's on Angular scope we have to put `()` after it just to get at the function itself?! (confused)

kifi-backend/modules/shoebox/angular/src/keeps/keeps.js
joshm1, line 331: I think we have a difference of opinion. I think this is more readable than:

``` javascript
return _.isFunction(scope.selectedKeepsFilter()) ? scope.selectedKeepsFilter()(selectedKeeps) : selectedKeeps;
```

edit: I also don't think having 2 local variables is a problem.  I agree with limiting # of local variables used, but not when that # is 2.

yipingliao, line 331: Okay. It looks like we _do_ need those parens to get the function on the scope. How confusing.

joshm1, line 331: Yep, the result of passing a function in a one-way binding. :)

kifi-backend/modules/shoebox/angular/src/search/search.js
```diff
@@ -12,6 +12,7 @@ angular.module('kifi')
     var filter = $routeParams.f || 'm';
     var lastResult = null;
     var selectedCount = 0;
+    var hitCounter = 0;
```
yipingliao, line 15: This is no longer used.

PR #11985, opened 2014-10-29
parse android user agent

yingjieMiao:
@2is10


kifi-backend/modules/common/app/com/keepit/common/net/UserAgent.scala
```diff
@@ -35,6 +35,7 @@ object UserAgent extends Logging {
   private val MAX_USER_AGENT_LENGTH = 512
   lazy val parser = UADetectorServiceFactory.getResourceModuleParser()
   lazy val iosAppRe = """^(iKeefee)/(\d+\.\d+)(\.\d+) \(Device-Type: (.+), OS: (iOS) (.+)\)$""".r("appName", "appVersion", "buildSuffix", "device", "os", "osVersion")
+  lazy val androidAppRe = """(\w+)\/\d+\.\d+\.\d+ \(Linux; U; (Android) (\d+\.\d+\.\d+); (.+?) Build\/.+\)""".r("appName", "os", "osVersion", "device")
```
2is10: This regular expression looks pretty specific, considering that it needs to match strings that we don’t control. How did you determine that this will match our app on all Android devices? Might be worth adding the URL of a page about it, if you found one that was helpful, or else some indication of your confidence level about it.

A quick search turned up a few Dalvik user agent strings, which the regular expression above properly matches:

```
Dalvik/1.6.0 (Linux; U; Android 4.0.3; A500 Build/IML74K; BrowserMob RBU)
Dalvik/1.6.0 (Linux; U; Android 4.2.2; Galaxy Nexus Build/JDQ39)
```

And a few Android browser user-agent strings which the regular expression above properly does _not_ match.

```
Mozilla/5.0 (Linux; U; Android 4.0.4; nl-nl; GT-N8010 Build/IMM76D) AppleWebKit/534.30 (KHTML, like Gecko) Version/4.0 Safari/534.30
Mozilla/5.0 (Android; Mobile; rv:26.0) Gecko/26.0 Firefox/26.0
Mozilla/5.0 (Android; Tablet; rv:26.0) Gecko/26.0 Firefox/26.0
```

In short, I have no reason to question this code. I’d just like a little more evidence that it’s correct, if possible.

2is10:
:+1:

kifi-backend/modules/common/app/com/keepit/common/net/UserAgent.scala
yingjieMiao: That regex is mainly used for mobile app. I come up with this regex based on existing userAgents found in `kifi_installation` table (where platform = android) 

For things don't match the regex, we default to
`UserAgent.parser.parse(userAgent)` where parser is from 

```
UADetectorServiceFactory.getResourceModuleParser()
```

We use this parser as a fall back at a few places. 

I can remove the `Linux; U;` part, it does look very specific.

2is10: Thanks for the explanation! Glad it was based on our real data.​

PR #11986, opened 2014-10-29
fix copying keeps between Main & Private + add in error modals

ahsu1230:

problem with keeping from Keep Card or copying/moving keeps in between Main & Private Libraries
Fix in endpoint + Tests
Add in error modals so we have UI of when things break!!!!


ahsu1230:
merging to fix in production - comments still welcome!

PR #11987, opened 2014-10-29
fix keeping recommendations

joshm1:
keeping from recommendations broke when keepToLibrary was changed to copyToLibrary

@ahsu1230 @stkem

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/keep/keep.js
+++ b/kifi-backend/modules/shoebox/angular/src/keep/keep.js
@@ -962,23 +962,38 @@ angular.module('kifi')
 
           // Keep.
           } else {
-            keepActionService.copyToLibrary([scope.keep.id], scope.librarySelection.library.id).then(function (result) {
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/angular/src/keep/keep.js
```diff
@@ -962,23 +962,38 @@ angular.module('kifi')
...
+
+            var keepToLibrary;
+            if (scope.keep && scope.keep.id) {
+              keepToLibrary = keepActionService.copyToLibrary([scope.keep.id], scope.librarySelection.library.id).then(function (result) {
+                if (result.successes > 0) {
+                  return keepActionService.fetchFullKeepInfo(scope.keep).then(fetchKeepInfoCallback);
+                }
+              });
+            } else {
+              // used for keeping recommendations
```
ahsu1230, line 987: Thanks for the fix!

yipingliao:
:+1:

PR #11995, opened 2014-10-29
Position keep-to-library widget so that it's always visible.

yipingliao:
@andrewconner Merge conflicts resolved version. Please feel free to comment on the original pull and I'll address them post-merge.

Also, add more customizable up/down offsets.
Highlight tag when a widget is open for that tag.
https://app.asana.com/0/18079737520750/18785877576576


andrewconner:
Cool. If we're ever in the situation where it's hard to fit an element in a viewport because of the position of the page, we also have `dom.scrollIntoViewLazy`, which may help.

yipingliao:
@andrewconner Yes, `dom.scrollIntoViewLazy` would be nice in that case. Thanks!

PR #12000, opened 2014-10-30
Fixing /install white screen, testing for UserActions

andrewconner:
A user reported that `/install` was returning a white page for non-users; this was because it was `UserAction`. It turns out that `UserPage` has never worked as intended (ie, wrapping `UserAction` calls and handling errors in a user friendly way).

It appears that `ActionA andThen ActionB` _only_ passes the result from `ActionA` to `ActionB` if it wasn't filtered. We can probably rewrite things to work as originally expected, but I didn't immediately see how based on the signatures of `ActionFunction`. This PR has a less-than-minimalistic implementation that works.

There's no tests on the new actions. `UserAction` / `MaybeUserAction` get called in other tests, but none dedicated to unit testing the core functionality. The bulk of this PR are the beginning tests. We need more, and we need to remove the application dependency on the session. Tests are critical for core infrastructure, because subtle bugs can spread to everywhere.

@stkem @eishay @raymondng


raymondng:
lgtm. Thanks!

PR #12002, opened 2014-10-30
Public header fix

andrewconner:
@yipingliao This just wires up the public header to the existing signup flow, and launches the app if appropriate to log in.


yipingliao:
Awesome.

PR #12003, opened 2014-10-30
sends a notification when someone follows your library

stkem:
@andrewconner @ahsu1230


kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -722,6 +738,7 @@ class LibraryCommander @Inject() (
         libraryMembershipRepo.getOpt(userId, libraryId) match {
           case None =>
             libraryMembershipRepo.save(LibraryMembership(libraryId = libraryId, userId = userId, access = maxAccess, showInSearch = true))
+            notifyOwnerOfNewFollower(userId, lib)
```
ahsu1230, line 741: woohoo!

PR #12004, opened 2014-10-30
During a search session, replace the URL instead of pushing a new one in the history

andrewconner:
Thanks @ymatsuda for the bug report. This fixes the history issue where searching would create many browser history entries.

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/layout/header/HeaderCtrl.js
+++ b/kifi-backend/modules/shoebox/angular/src/layout/header/HeaderCtrl.js
@@ -98,7 +98,7 @@ angular.module('kifi')
               $location.url('/find?q=' + $scope.search.text + '&f=' + 'm');
             }
           } else {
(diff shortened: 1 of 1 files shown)


andrewconner:
@yipingliao Turns out it was a very simple fix. [See "Replace method"](https://docs.angularjs.org/guide/$location)... interesting.

kifi-backend/modules/shoebox/angular/src/layout/header/HeaderCtrl.js
```diff
@@ -98,7 +98,7 @@ angular.module('kifi')
               $location.url('/find?q=' + $scope.search.text + '&f=' + 'm');
             }
           } else {
-            $location.search('q', $scope.search.text); // this keeps any existing URL params
+            $location.search('q', $scope.search.text).replace(); // this keeps any existing URL params
```
yipingliao, line 101: Nice!

ymatsuda:
cool!

PR #12005, opened 2014-10-30
bulk keep to library from search

joshm1:
@stkem 

cc @yipingliao 

another issue I need to fix is that bulk keeping does not mark the library as "kept" for the URLs individual keeper

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/keep/keep.js
+++ b/kifi-backend/modules/shoebox/angular/src/keep/keep.js
@@ -948,6 +948,11 @@ angular.module('kifi')
         // Scope methods.
         //
         scope.clickAction = function () {
+          if (scope.librarySelection.library.isReadOnly) {
+            // do nothing; library is set as ready only (URL is already kept)
+            return;
+          }
+
           // Unkeep.
           if (scope.librarySelection.library && scope.librarySelection.library.keptTo) {
             var keepToUnkeep = _.find(scope.keep.keeps, { libraryId: scope.librarySelection.library.id });
@@ -976,7 +981,7 @@ angular.module('kifi')
               var keep = new keepDecoratorService.Keep(fullKeep);
               keep.buildKeep(keep);
               keep.makeKept();
-              scope.$emit('keepAdded', libraryService.getSlugById(scope.librarySelection.library.id), keep, scope.librarySelection.library);
+              scope.$emit('keepAdded', libraryService.getSlugById(scope.librarySelection.library.id), [keep], scope.librarySelection.library);
             };
 
             var keepToLibrary;
--- a/kifi-backend/modules/shoebox/angular/src/keep/keepToLibraryWidget.js
+++ b/kifi-backend/modules/shoebox/angular/src/keep/keepToLibraryWidget.js
@@ -15,7 +15,7 @@ angular.module('kifi')
        *  librarySelection - an object whose 'library' property will be the selected library.
        *
        *  Optional properties on parent scope:
-       *   excludeLibraries - an array of libraries to exclude from libraries when populating the widget.
+       *   readOnlyLibraries - an array of libraries that the user cannot take action against
        *   keptToLibraries - an array of library ids that are already keeping the keep.
        *   clickAction() - a function that can be called once a library is selected;
        *                   called with the element that this widget is on.
@@ -49,8 +49,6 @@ angular.module('kifi')
         scope.showCreate = false;
         scope.newLibrary = {};
         scope.widgetLibraries = [];
-        scope.excludeLibraries = scope.excludeLibraries || [];
-        scope.keptToLibraries = scope.keptToLibraries || [];
 
 
         //
@@ -139,11 +137,20 @@ angular.module('kifi')
           }
         }
 
+        function mutateReadOnlyLibraries(libraries, readOnlyLibraries) {
+          libraries.forEach(function (library) {
+            library.isReadOnly = !!_.find(readOnlyLibraries, { 'id': library.id });
+          });
+        }
+
 
         //
         // Scope methods.
         //
         scope.showWidget = function () {
+          var readOnlyLibraries = scope.readOnlyLibraries || [];
+          var keptToLibraries = scope.keptToLibraries || [];
+
           // Create widget.
           widget = angular.element($templateCache.get('keep/keepToLibraryWidget.tpl.html'));
           $rootElement.find('html').append(widget);
@@ -229,14 +236,11 @@ angular.module('kifi')
           newLibraryNameInput = widget.find('.keep-to-library-create-name-input');
 
           var libraries = _.filter(libraryService.librarySummaries, { access: 'owner' });
-
-          libraries = _.filter(libraries, function (library) {
-            return !_.find(scope.excludeLibraries, { 'id': library.id });
-          });
+          mutateReadOnlyLibraries(libraries, readOnlyLibraries);
 
           libraries.forEach(function (library) {
             library.keptTo = false;
-            if (_.indexOf(scope.keptToLibraries, library.id) !== -1) {
+            if (_.indexOf(keptToLibraries, library.id) !== -1) {
               library.keptTo = true;
             }
           });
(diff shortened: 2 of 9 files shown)


kifi-backend/modules/shoebox/angular/src/keep/keepToLibraryWidget.tpl.html
```diff
@@ -6,7 +6,7 @@
     </div>
     <ul class="library-select-list">
       <li ng-repeat="library in widgetLibraries | filter:search" class="library-select-option"
-        ng-class="{'selected': library.selected, 'keptTo': library.keptTo}"
+        ng-class="{'selected': library.selected, 'keptTo': library.keptTo, 'readOnly': library.isReadOnly}"
```
yipingliao, line 9: Any CSS styles based on the `readOnly` class? I only saw JavaScript checks for `library.isReadOnly`. How is a library's read-only attribute indicated to the user?

kifi-backend/modules/shoebox/angular/src/keeps/keeps.tpl.html
```diff
@@ -15,6 +15,9 @@
         <div ng-if="librariesEnabled && editOptions().actions.moveToLibrary" class="kf-keep-selection-option move-to-library">
           <span kf-keep-to-library-widget ng-click="showWidget()">Move to library</span>
         </div>
+        <div ng-if="librariesEnabled && editOptions().actions.keepToLibrary" class="kf-keep-selection-option copy-to-library">
```
yipingliao, line 18: Why do we want an extra action called "Keep to library" that does the exact same thing as "Copy to library"?

kifi-backend/modules/shoebox/angular/src/keep/keep.js
```diff
@@ -976,7 +981,7 @@ angular.module('kifi')
               var keep = new keepDecoratorService.Keep(fullKeep);
               keep.buildKeep(keep);
               keep.makeKept();
-              scope.$emit('keepAdded', libraryService.getSlugById(scope.librarySelection.library.id), keep, scope.librarySelection.library);
+              scope.$emit('keepAdded', libraryService.getSlugById(scope.librarySelection.library.id), [keep], scope.librarySelection.library);
```
yipingliao, line 984: Since we are now passing in an array of keeps, what do you think about renaming the event to `keepsAdded`?

kifi-backend/modules/shoebox/angular/src/keeps/keeps.tpl.html
joshm1, line 18: it used to be different behavior before a refactor, now the only difference is the text.  I'm fine with reusing Copy To Library and I think it makes sense

kifi-backend/modules/shoebox/angular/src/keep/keepToLibraryWidget.tpl.html
joshm1, line 9: No CSS at this time. It's not indicated to the user. So this CSS class does nothing.

kifi-backend/modules/shoebox/angular/src/keep/keep.js
joshm1, line 984: Agreed, better name

kifi-backend/modules/shoebox/angular/src/keep/keepToLibraryWidget.js
```diff
@@ -347,14 +351,14 @@ angular.module('kifi')
             libraryService.fetchLibrarySummaries(true).then(function (data) {
               scope.$evalAsync(function () {
                 var libraries = _.filter(data.libraries, { access: 'owner' });
+                var readOnlyLibraries = scope.readOnlyLibraries || [];
+                var keptToLibraries = scope.keptToLibraries || [];
 
-                libraries = _.filter(libraries, function (library) {
-                  return !_.find(scope.excludeLibraries, { 'id': library.id });
-                });
+                mutateReadOnlyLibraries(libraries, readOnlyLibraries);
```
yipingliao, line 357: Just a suggestion: make a `augmentLibraries` function that adds both `.isReadOnly` and `.keptTo` fields, if applicable, to the libraries.

kifi-backend/modules/shoebox/angular/src/keeps/keeps.js
```diff
@@ -341,6 +341,10 @@ angular.module('kifi')
             // TODO: look at result and flag errors. Right now, even a partial error is flagged so that's
             //       not good.
             libraryService.fetchLibrarySummaries(true);
+
+            var addedKeeps = _.map(_.groupBy(selectedKeeps, 'url'), function (keeps, url) { return { 'url': url }; });
```
yipingliao, line 345: Will `_.map(_.pluck(selectedKeeps, 'url'), function (url) { return { 'url': url }; })` work?

kifi-backend/modules/shoebox/angular/src/keeps/keeps.js
```diff
@@ -542,9 +547,13 @@ angular.module('kifi')
...
                 scope.librarySelection = {};
                 scope.librarySelection.library = _.find(scope.libraries, { 'kind': 'system_main' });
-                scope.excludeLibraries = !scope.library ? [] : [scope.library];
+                scope.readOnlyLibraries = !scope.library ? [] : [scope.library];
                 scope.clickAction = function (widgetElement) {
-                  if (widgetElement.closest('.copy-to-library').length) {
+                  if (scope.librarySelection.library.isReadOnly) {
+                    return; // do nothing
+
+                    // FIXME refactor to not use CSS for business logic
```
yipingliao, line 555: Yes, that would be great! We've been using `TODO` for tasks to do later.

kifi-backend/modules/shoebox/angular/src/search/search.js
```diff
@@ -72,7 +72,11 @@ angular.module('kifi')
     // Scope methods.
     //
     $scope.selectedKeepsFilter = function (hits) {
-      return _.flatten(_.pluck(hits, 'keeps'));
+      return _.flatten(_.map(hits, function (hit) {
```
yipingliao, line 75: Can you add a comment here about what this is doing? Maybe an example data set, like the examples on the lodash documentation page, with input and output? Four lodash calls in the same line is a little hard to figure out. :)

kifi-backend/modules/shoebox/angular/src/keeps/keeps.js
```diff
@@ -560,6 +569,28 @@ angular.module('kifi')
         }, function (numSelected) {
           scope.disableEditTags();
           scope.updateSelectedCount({ numSelected: numSelected });
+
+          if (numSelected > 0) {
+            // creates an array of objects that identify the selected keeps and libraries they are in
+            var urlToLibraryPairs = _.flatten(_.map(scope.selection.getSelected(scope.keeps), function (keep) {
```
yipingliao, line 575: This is another suggestion; you may disagree.
When there are three nested lodash functions, it's very difficult to figure out what is going on. This block of code involves only 2 arrays - `scope.selection.getSelected(scope.keeps)` scope.keeps`and`keep.keeps`. If we can break at least part of this up into a forEach or for loop, it may be much more readable.

kifi-backend/modules/shoebox/angular/src/keeps/keeps.js
```diff
@@ -542,9 +547,13 @@ angular.module('kifi')
 
                 scope.librarySelection = {};
                 scope.librarySelection.library = _.find(scope.libraries, { 'kind': 'system_main' });
-                scope.excludeLibraries = !scope.library ? [] : [scope.library];
+                scope.readOnlyLibraries = !scope.library ? [] : [scope.library];
```
yipingliao, line 550: I don't think we need this line anymore. Rather, maybe we can just set `scope.readOnlyLibraries` to `[]`. When we talked about this yesterday, we found that we don't have the case where we want to "exclude" `scope.library` anymore.

yipingliao:
Thanks!

kifi-backend/modules/shoebox/angular/src/keep/keepToLibraryWidget.tpl.html
yipingliao, line 9: Let's remove if it doesn't do anything.

kifi-backend/modules/shoebox/angular/src/keeps/keeps.tpl.html
yipingliao, line 18: Is that the UI we want to present? That is, do we want to present to users two different actions that actually do the same thing?

kifi-backend/modules/shoebox/angular/src/keeps/keeps.js
joshm1, line 345: No because in search multiple keeps can have the same URL and that could return a collection of duplicate objects with the same url.

PR #12006, opened 2014-10-30
caching keep images

ahsu1230:
There are two repo functions (getForKeepId & getAllForKeepId)
- getForKeepId only checks for active states
- getAllForKeepId checks for any state <- I put the cache here because retrieving images for `getBestImageForKeep()` uses only this repo function. If there are no active keep images for whatever reason, I believe `autoSetKeepImage` will persist a new image.

What do you think? @andrewconner


andrewconner:
Are you seeing slow queries for this? This should be indexed — I tried manually running the queries and it was less than 1ms.

ahsu1230:
I see. I guess they are just called often, so maybe caching is not the way to go.The longer queries are `createFullLibraryInfo()` and `getLibraryMembers()`

ahsu1230:
I checked `getLibraryMembers()` and it turns out we only have a cache for LibraryMembershipId -> LibraryMembership, which actually isn't used much (only for `LibraryMembershipRepo.get()`)
I think it would be better to cache operations within LibraryMembership - specifically maybe `pageWithLibraryIdAndAccess`
what do you think?

andrewconner:
Could be, all depends on what queries are taking a while. Well indexed queries on smaller tables should be fast enough. Though, we may want to cache sets of query results.

PR #12010, opened 2014-10-30
dead code

yingjieMiao:
@ymatsuda


ymatsuda:
Sorry that the search blind test was not used extensively. I think it was implemented nicely, though.
If you think it should go, I am ok.

yingjieMiao:
we can fairly easily get it back if needed.

PR #12014, opened 2014-10-30
Limiting init as well

andrewconner:
`init` in `LibraryCtrl` didn't respect `$scope.loading`, so _if_ you somehow had it called twice before it returned for the first time, it could double load. @yipingliao


yipingliao:
Nice.

PR #12018, opened 2014-10-30
CollectionHelpers (rename)

raymondng:
Thanks @leogrim for collecting some helper functions into `Collection`. 

I was looking for `dedupBy` and Idea got confused with `Collection` -- renaming to align with `FutureHelpers` and maybe help save a few seconds here and there.

@leogrim


raymondng:
Merging as I'll be using this. Hopefully renaming ok.

leogrim:
:+1:

PR #12019, opened 2014-10-30
Use result.session in UserPage

raymondng:
Small change. To be consistent and to prevent any loss of session state that might be added for non-users (e.g. kcid)

@andrewconner


raymondng:
Discussed in person with @andrewconner

PR #12030, opened 2014-10-30
refactors and bug fixes when keeping from search

joshm1:
- bulk "copy to library" actions now add the library to as kept in the individual keepers
- removed the "already kept" check from bulk copy to library

@stkem @yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/search/search.js
+++ b/kifi-backend/modules/shoebox/angular/src/search/search.js
@@ -3,8 +3,8 @@
 angular.module('kifi')
 
 .controller('SearchCtrl', [
-  '$scope', '$location', '$routeParams', '$window', 'keepDecoratorService', 'searchActionService',
-  function ($scope, $location, $routeParams, $window, keepDecoratorService, searchActionService) {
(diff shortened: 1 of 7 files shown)


kifi-backend/modules/shoebox/angular/src/search/search.js
```diff
@@ -196,6 +206,23 @@ angular.module('kifi')
...
+          searchKeep.keeps.push({
+            id: keep.id,
+            isMine: true,
+            libraryId: library.id,
+            mine: true,
+            visibility: library.visibility
+          });
+        }
+      });
+    });
```
joshm1, line 229: this event listener in the search controller allows us to update the data model based on actions performed on keep card(s), the UI will automatically update as long as the proper watches are in place. I think this is a better/safer pattern over manually updating the "view only" objects (i.e. `keptToLibraries`, `visibleKeepLibraries`), which can cause problems if view models is out of sync with business models.

yipingliao, line 229: Nice!

PR #12035, opened 2014-10-30
Report hits on urls thought to have moved permanently

leogrim:
@2is10 @eishay @stkem @andrewconner 
Let's see how often this happens.


2is10:
:+1:

As you know, “permanently” in the 301 sense only means that the publisher doesn’t plan to move the thing back. It doesn’t mean that nothing new will ever be published at the old location.

PR #12036, opened 2014-10-30
fix bug which gets more keeps after unkeeping / moving

ahsu1230:
also a quick fix for copy keeps @joshm1 

@andrewconner @yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/keeps/keeps.js
+++ b/kifi-backend/modules/shoebox/angular/src/keeps/keeps.js
@@ -369,6 +369,12 @@ angular.module('kifi')
             libraryService.addToLibraryCount(currentLibraryId, -1 * selectedKeeps.length);
             scope.availableKeeps = _.difference(scope.availableKeeps, selectedKeeps);
             scope.selection.unselectAll();
+
+            if (scope.availableKeeps.length < 10) {
+              scope.scrollNext()(scope.availableKeeps.length).then(function () {
+                scope.availableKeeps = _.filter(scope.keeps, function(k) { return k.unkept === false; });
+              });
+            }
           })['catch'](function () {
             modalService.open({
               template: 'common/modal/genericErrorModal.tpl.html'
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/shoebox/angular/src/keeps/keeps.js
```diff
@@ -458,53 +464,41 @@ angular.module('kifi')
 
         scope.unkeep = function (keeps) {
           var selectedKeeps = scope.selection.getSelected(keeps);
+          var libraryId = scope.library.id;
 
-          if (scope.librariesEnabled) {
```
ahsu1230, line 462: removed librariesEnabled here

kifi-backend/modules/shoebox/angular/src/keeps/keeps.tpl.html
```diff
@@ -29,7 +29,7 @@
       </div>
     </div>
   </div>
-  <ol class="kf-my-keeps" smart-scroll scroll-distance="scrollDistance" scroll-disabled="isScrollDisabled()" scroll-next="scrollNext()">
+  <ol class="kf-my-keeps" smart-scroll scroll-distance="scrollDistance" scroll-disabled="isScrollDisabled()" scroll-next="scrollNext()(keeps.length)">
```
ahsu1230, line 32: thanks @joshm1 for the help!

kifi-backend/modules/shoebox/angular/src/libraries/library.tpl.html
```diff
@@ -32,7 +32,7 @@
       keeps-has-more="hasMore"
       scroll-disabled="!hasMore"
       scroll-distance="scrollDistance"
-      scroll-next="getNextKeeps()"
+      scroll-next="getNextKeeps"
```
yipingliao, line 35: I think that if you kept the original parens, then in the .tpl file you would be able to do `scroll-next="scrollNext(keeps.length)"` instead of the strange looking `scroll-next="scrollNext()(keeps.length)"`.

kifi-backend/modules/shoebox/angular/src/keeps/keeps.js
```diff
@@ -369,6 +369,12 @@ angular.module('kifi')
             libraryService.addToLibraryCount(currentLibraryId, -1 * selectedKeeps.length);
             scope.availableKeeps = _.difference(scope.availableKeeps, selectedKeeps);
             scope.selection.unselectAll();
+
+            if (scope.availableKeeps.length < 10) {
+              scope.scrollNext()(scope.availableKeeps.length).then(function () {
```
yipingliao, line 374: Similarly, you won't need the extra parens here. I just remembered that in general, when passing functions into an isolate scope, we should include the parens in the attributes so that the actual usage of them does not require extra parens.

kifi-backend/modules/shoebox/angular/src/keeps/keeps.js
yipingliao, line 462: Yay!

kifi-backend/modules/shoebox/angular/src/keeps/keeps.js
```diff
@@ -369,6 +369,12 @@ angular.module('kifi')
...
+                scope.availableKeeps = _.filter(scope.keeps, function(k) { return k.unkept === false; });
```
yipingliao, line 375: `_.reject(scope.keeps, 'unkept')`?

yipingliao:
Nice improvement to the user experience. :)

kifi-backend/modules/shoebox/angular/src/keeps/keeps.js
ahsu1230, line 375: ah okay. that would work better lol

kifi-backend/modules/shoebox/angular/src/libraries/library.tpl.html
ahsu1230, line 35: For some reason, if I try that, the offset parameter doesn't get passed correctly (comes out undefined).

joshm1, line 35: if the parens are there is actually calls the function and passes to the child the return value of that method call.  He is trying to pass a function without calling it.  The double parens are because of 1-way binding, where the first function call actually returns the function he needs.  If there is a cleaner way to do this, I'd love to hear it.

edit: If the method call always used the same scope objects as the function parameter, then what yiping suggested will work, but it looks like `scrollNext` is being called differently in the same directive ( `keeps.length` and `scope.availableKeeps.length`)

@yipingliao @ahsu1230

kifi-backend/modules/shoebox/angular/src/keeps/keeps.js
joshm1, line 374: ditto my comment below

PR #12041, opened 2014-10-31
sitemap (prototype)

raymondng:
A prototype/silver-bullet for quick feedback/learning. Does not scale and may not work.

The plugin currently is no-op, but the admin endpoint should be functional, though it functions in a way that (again) does not scale. You should consider it a temporary endpoint.

@andrewconner please take a look. don't merge yet. tested locally -- will need to add test, etc.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/common/seo/SiteMapGenerator.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/common/seo/SiteMapGenerator.scala
@@ -1,0 +1,98 @@
+package com.keepit.common.seo
+
+import com.google.inject.{ Singleton, Inject }
+import com.keepit.commanders.LibraryCommander
+import com.keepit.common.CollectionHelpers
+import com.keepit.common.actor.ActorInstance
+import com.keepit.common.akka.FortyTwoActor
+import com.keepit.common.db.slick.Database
+import com.keepit.common.healthcheck.AirbrakeNotifier
+import com.keepit.common.logging.Logging
+import com.keepit.common.plugin.{ SchedulerPlugin, SchedulingProperties }
+import com.keepit.model.{ UserRepo, Library, LibraryRepo }
+import play.api.{ Play, Plugin }
+import Play.current
+import scala.concurrent.Future
+import scala.concurrent.duration._
+import scala.xml.Elem
+
+trait SiteMapGeneratorPlugin extends Plugin {
+  def generate()
+}
+
+class SiteMapGeneratorPluginImpl @Inject() (
+    actor: ActorInstance[SiteMapGeneratorActor],
+    val scheduling: SchedulingProperties) extends Logging with SiteMapGeneratorPlugin with SchedulerPlugin {
+
+  // plugin lifecycle methods
+  override def enabled: Boolean = true
+  override def onStart() {
+    for (app <- Play.maybeApplication) {
+      val (initDelay, freq) = if (Play.isDev) (30 seconds, 10 minutes) else (15 minutes, 60 minutes)
+      log.info(s"[onStart] SiteMapGeneratorPlugin started with initDelay=$initDelay freq=$freq")
+      scheduleTaskOnLeader(actor.system, initDelay, freq, actor.ref, Generate)
+    }
+  }
+
+  override def generate() { actor.ref ! Generate }
+
+}
+
+case class Generate()
+
+// library-only for now
+class SiteMapGeneratorActor @Inject() (
+    airbrake: AirbrakeNotifier,
+    db: Database,
+    generator: SiteMapGenerator,
+    userRepo: UserRepo,
+    libraryRepo: LibraryRepo,
+    libraryCommander: LibraryCommander) extends FortyTwoActor(airbrake) with Logging {
+
+  def receive() = {
+    case Generate =>
+    // todo: generate and save to disk, maybe
+  }
+
+}
+
+@Singleton
+class SiteMapGenerator @Inject() (
+    airbrake: AirbrakeNotifier,
+    db: Database,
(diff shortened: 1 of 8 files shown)


andrewconner:
Wahoo! + @eishay

kifi-backend/modules/shoebox/app/com/keepit/common/seo/SiteMapGenerator.scala
```diff
@@ -0,0 +1,98 @@
...
+class SiteMapGenerator @Inject() (
+    airbrake: AirbrakeNotifier,
+    db: Database,
+    userRepo: UserRepo,
+    libraryRepo: LibraryRepo,
+    libraryCommander: LibraryCommander) extends Logging {
+
+  // prototype/silver-bullet implementation -- will learn and improve
+  def generate(): Future[Elem] = Future.successful {
+    val libraries = db.readOnlyReplica { implicit ro => libraryRepo.getAllPublishedLibraries().take(50000) } // ahem does not scale
```
eishay, line 70: paginate?

kifi-backend/modules/shoebox/app/com/keepit/common/seo/SiteMapGenerator.scala
```diff
@@ -0,0 +1,98 @@
...
+    // may want to add warning when > 40K
+
+    // batch, optimize
+    val ownerIds = CollectionHelpers.dedupBy(libraries.map(_.ownerId))(id => id)
+    val owners = db.readOnlyMaster { implicit ro => userRepo.getUsers(ownerIds) } // cached
+    val paths = libraries.filter { lib => owners.get(lib.ownerId).isDefined } map { lib =>
```
eishay: we should not put libraries with three or less keeps in them. lets filter them out by count

eishay:
i assume we don't have an endpoint yet and there's no ping to google, right?

raymondng:
Right. Nothing in prod right now. No worries.

kifi-backend/modules/shoebox/app/com/keepit/common/seo/SiteMapGenerator.scala
andrewconner: Agreed.

kifi-backend/modules/shoebox/app/com/keepit/common/seo/SiteMapGenerator.scala
```diff
@@ -0,0 +1,98 @@
...
+      val owner = owners(lib.ownerId)
+      s"https://kifi.com${Library.formatLibraryPath(owner.username, owner.externalId, lib.slug)}"
```
andrewconner: `https://www.kifi.com`

eishay:
also, it seems like the sitemap should be persisted (s3?) and cached for very fast retrieval.

eishay:
Note that "The limit is 50,000 URLs or a maximum size of 10MB uncompressed".
http://googlewebmastercentral.blogspot.com/2014/10/best-practices-for-xml-sitemaps-rssatom.html
so we need to check for size as we grow it or put a safe distance from it though not make it too small "A common mistake is to put only a handful of URLs into each XML sitemap file, which usually makes it harder for Google to download all of these XML sitemaps in a reasonable time".

eishay:
also please look at the "XML Sitemap Index" http://www.sitemaps.org/protocol.html for the chunked sitemaps. I think we should start with one sitemap asap (for quick user growth, super important to do it quick) and rss for each library, then soon after have the chunked sitemap with the index so we can scale it.

raymondng:
Thanks for the feedback! Yes that's the idea -- to do a quick prototype, and to (benefit &) learn from it.
I am aware of the 50K limit (and we currently have only 500 custom published libraries -- lots and lots of headroom), so no real need for chunking right now ... and you may have noticed comments like `may want to add warning when > 40K` and `generate and save to disk, maybe` on the list of things to do. In reality I think the code will be "ready" way before we hit 50K limit, but if we grow faster than that, I'd be happy to jump in with both feet and take care of things.

kifi-backend/modules/shoebox/app/com/keepit/common/seo/SiteMapGenerator.scala
raymondng: :)

kifi-backend/modules/shoebox/app/com/keepit/common/seo/SiteMapGenerator.scala
raymondng: what's the downside? 3 keeps can also be good/useful, right?

kifi-backend/modules/shoebox/app/com/keepit/common/seo/SiteMapGenerator.scala
raymondng, line 70: yes we'll need it going forward

kifi-backend/modules/shoebox/app/com/keepit/common/seo/SiteMapGenerator.scala
andrewconner: Rough proxy for quality. Less than 3 likely doesn't have useful content to extract anyways.

raymondng: Thanks for the clarifications. No worries.

For the record, I do have a slightly different perspective -- I don't know if we are able to judge (or in the position to judge) quality, and I don't know if we can say for sure that a library with 3 keeps is not useful. Nonetheless let's do this and we'll iterate.

raymondng:
Thanks for feedback. Will merge and take it out for a spin.

PR #12046, opened 2014-10-31
trying to avoid org.apache.lucene.store.AlreadyClosedException

ymatsuda:
An indexer sometimes throws AlreadyClosedException (trying to write to an already closed IndexWriter) when a search instance is shutting down. There is no real harm, but airbrake messages are annoying. So, when shutting down, isClosing flag is turned and stops indexing in indexing actors.

@leogrim @yingjieMiao @eishay


yingjieMiao:
:+1:

PR #12052, opened 2014-10-31
Putting sitemap.xml in assets. can't figure out where to upload it to Google directly.

andrewconner:
@raymondng


raymondng:
:+1:

raymondng:
Should we put `robots.txt` as assets as well? If nothing else, there is a `Sitemap` directive we can use in `robots.txt` (to be doubly sure that google crawler knows about it) Thoughts?

2is10:
Yes, we should have a robots.txt. And [specify the sitemap](http://www.sitemaps.org/protocol.html#submit_robots) there.

And we should be sure to have our marketing pages in our sitemap too, preferably [prioritized](http://www.sitemaps.org/protocol.html#prioritydef).

PR #12054, opened 2014-10-31
optimize library pages (a little bit) -> save counting on libraryMembers...

ahsu1230:
...hipRepo

Right now, the bulk of library pages comes up `createFullLibraryInfo`
and a part of that is `getLibraryMembers` which pages memberships and counts the total number of collaborators, followers, etc. It would also look through the table repo twice (one for collaborators and one for followers)

`createFullLibraryInfo` ends up calling this counting twice (since `getLibraryMembers` already does it)
so the two main changes were:
- have createFullLibraryInfo only search the library_membership_repo once
- the repo function returns a map (access -> count) on every library access so you won't have to look through the table again for a different access
  but `getLibraryMembers` now also returns a map (access -> count)
  @andrewconner @leogrim

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/model/LibraryMembership.scala
+++ b/kifi-backend/modules/common/app/com/keepit/model/LibraryMembership.scala
@@ -58,4 +58,6 @@ object LibraryAccess {
       case OWNER.value => OWNER
     }
   }
+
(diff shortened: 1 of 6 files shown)


andrewconner:
Great thanks

eishay:
Thanks!
Did we do an index review on all calls to make sure we're always using an index?

ahsu1230:
I can try taking a look now that I know how =)

PR #12055, opened 2014-10-31
redirect /path.html to /path if /path has an existing matching file. if ...

eishay:
...not, 404
@andrewconner


kifi-backend/modules/shoebox/app/com/keepit/controllers/website/MarketingSiteRouter.scala
```diff
@@ -37,7 +37,12 @@ object MarketingSiteRouter extends AssetsBuilder with Controller with Logging {
   def marketingSite(path: String = "index") = {
     val file = if (path.isEmpty) "index" else path
     if (file.contains(".html")) {
-      NotFound(views.html.error.notFound(s"request.path (try to remove the .html)"))
+      val noHtml = file.replace(".html", "")
+      if (maybeCachedIndex(noHtml).nonEmpty) {
+        Redirect(s"https://www.kifi.com/$noHtml")
+      } else {
+        NotFound(views.html.error.notFound(s"request.path (try to remove the .html)"))
```
andrewconner, line 44: `s"${request.path} (try to remove the .html)")`

PR #12057, opened 2014-10-31
a few kifi.com search fixes/polish (see contained commits)

stkem:
@andrewconner @yipingliao


andrewconner:
Great thanks

stkem:
sorry this took so long. the change needed in library.js was rather elusive for a while

PR #12062, opened 2014-10-31
library membership repo change -> unique constraint

ahsu1230:
@andrewconner @eishay 

Adding unique constraint onto library_membership repo so there won't be duplicate entries of (user_id, library_id)

also changed a couple of libraryMembershipRepo functions to be pre-compiled & get rid of `getOpt()` since it doesn't check for state, which is confusing - especially if we already have `getByLibraryIdAndUserId()`


kifi-backend/modules/shoebox/app/com/keepit/model/LibraryMembershipRepo.scala
```diff
@@ -77,27 +76,32 @@ class LibraryMembershipRepoImpl @Inject() (
   def pageWithLibraryIdAndAccess(libraryId: Id[Library], offset: Int, limit: Int, accessSet: Set[LibraryAccess], excludeState: Option[State[LibraryMembership]] = Some(LibraryMembershipStates.INACTIVE))(implicit session: RSession): Seq[LibraryMembership] = {
     (for (b <- rows if b.libraryId === libraryId && b.access.inSet(accessSet) && b.state =!= excludeState.orNull) yield b).sortBy(r => (r.access, r.createdAt)).drop(offset).take(limit).list
   }
+
+  private def getWithLibraryIdCompiled(libraryId: Column[Id[Library]], excludeState: Option[State[LibraryMembership]]) = Compiled {
```
andrewconner: This is probably my fault, but compiled queries shouldn't be `def`s (defeats the purpose of pre-compiling it if it has to compile every time). See `UsernameAlias.scala:80` for an example of a good way to do it.

andrewconner:
Nice, thanks!

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryMembershipRepo.scala
ahsu1230: ohh I see. whoops. gotcha

PR #12064, opened 2014-10-31
Do not include passphrase with invites to published libraries

leogrim:
@andrewconner @stkem

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/emails/LibraryInviteEmailSender.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/emails/LibraryInviteEmailSender.scala
@@ -12,7 +12,7 @@ import com.keepit.common.mail.template.EmailToSend
 import com.keepit.common.mail.template.TemplateOptions._
 import com.keepit.common.social.BasicUserRepo
 import com.keepit.common.mail.template.helpers.fullName
-import com.keepit.model.{ UserEmailAddressRepo, UserRepo, NotificationCategory, User, LibraryInvite, LibraryRepo, KeepRepo }
+import com.keepit.model._
 import com.keepit.social.BasicUser
 
 import play.api.libs.concurrent.Execution.Implicits.defaultContext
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/emails/LibraryInviteEmailSender.scala
```diff
@@ -12,7 +12,7 @@ import com.keepit.common.mail.template.EmailToSend
 import com.keepit.common.mail.template.TemplateOptions._
 import com.keepit.common.social.BasicUserRepo
 import com.keepit.common.mail.template.helpers.fullName
-import com.keepit.model.{ UserEmailAddressRepo, UserRepo, NotificationCategory, User, LibraryInvite, LibraryRepo, KeepRepo }
+import com.keepit.model._
```
stkem, line 15: IntelliJ?

andrewconner, line 15: It was up to 8 imports. Can that be okay? It makes for massive headers, and especially for `com.keepit.model`, it's very clear what you're using.

yingjieMiao, line 15: we need a consensus on this ...

andrewconner, line 15: Fight!

leogrim, line 15: My other pull request brought back the list. I'll let you fight though.

yingjieMiao, line 15: not sure if this works:

http://stackoverflow.com/questions/3348816/intellij-never-use-wildcard-imports

PR #12066, opened 2014-10-31
Fixing join_us.html link

andrewconner:
@eishay Previous fix didn't work because `MarketingSiteRouter` isn't a wildcard router. Manually added the offending page.


eishay:
thanks, was just adding it as well and got a conflict :-)

PR #12076, opened 2014-10-31
an attempt to stop sending 0.jpg when a user has no picture

2is10:
@andrewconner 
cc @stkem

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
@@ -126,7 +126,7 @@ angular.module('kifi')
           scope.library.shareTwitterUrl = scope.library.shareUrl +
             '?utm_medium=vf_twitter&utm_source=library_invite&utm_content=lid_' + scope.library.id +
             '&kcid=na-vf_twitter-library_invite-lid_' + scope.library.id;
-          scope.library.shareText = 'Check out this Kifi library about ' + scope.library.name + '!';
+          scope.library.shareText = 'Discover this amazing @Kifi library about ' + scope.library.name + '!';
 
           // Figure out whether this library is a library that the user has been invited to.
           // If so, display an invite header.


2is10:
Some tests are still failing, but I’d appreciate hearing your opinions.

andrewconner:
What about removing `pictureName` and add `pic` / `picUrl` or something, and provide a full URL. That way we can stop uploading a ghost image for every user, and return something user specific, or random, etc.

How the URL is determined becomes an abstraction, and we can return a consistent single ghost for users with no images. Thoughts?

andrewconner:
I don't think this solution is safe — it's possible BasicUser is serialized before the image uploads (so before it's assigned in the `user` table).

2is10:
Yeah, I don’t think this solution is safe either. We could make `UserRepo` more aware of pictures and never write to the BasicUser cache before there’s a picture.

I actually intend for some stuff to break in our system as a result of whatever protective change we make. As long as the signup flow still works, I’d argue that anything else is better broken hard than completed with a broken user picture URL—especially notifications, whose picture URLs will never be repaired.

I’d be open to the full URL idea, but it really blows up the size of a list of `BasicUser` JSONs, which we pass around a lot. Also, I’d like to start generating custom ghost images for each user using some randomization of colors and shapes, and maybe an initial or two. I think it’d be best to continue uploading these for each user, even if we only initially chose a fixed set of N random images, because we probably want each user’s picture to be stable even if/when we change the random images in the future, but I may be convinced otherwise.

andrewconner:
The other option I can think of is having an identifier in `pictureName` that means "we have nothing". Perhaps undefined or an empty string.

Using the whole URL does cost those extra bytes, but future proofs us a bit against changes. Right now, we're pretty solidly locked in to the current architecture.

2is10:
I’m not going to spend any more time on this now, but I think someone needs to fix it this week, and I’ll be happy to discuss strategies in person. I think we’ll want to avoid changing the semantics of `BasicUser` for clients.

PR #12083, opened 2014-11-01
[DO NOT MERGE] landing a/b

raymondng:
Added simple A/B landing based on the assumption that the files will be named "index1" and "index2"

Don't merge. Want to send this out first because I have been losing internet connectivity regularly these days.

For hashing I picked murmur3 but feel free to suggest other alternatives. For our simple case probably doesn't matter too much -- might even be an overkill. Will monitor. Also I chose not to convert to `InetAddress` (just hash the string) to avoid overhead.

Tested locally. No unit tests yet. I will look into adding one as well.

@andrewconner @eishay

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/MarketingSiteRouter.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/MarketingSiteRouter.scala
@@ -1,13 +1,18 @@
 package com.keepit.controllers.website
 
 import java.io.StringWriter
+import java.net.InetAddress
 
+import com.google.common.base.Charsets
+import com.google.common.hash.Hashing
 import com.keepit.common.logging.Logging
 import controllers.AssetsBuilder
 import org.apache.commons.io.IOUtils
 import play.api.Play.current
-import play.api.mvc.{ Action, Controller }
+import play.api.mvc.{ Request, Controller }
 import play.api.{ Mode, Play }
+
+import scala.util.Try
 
 object MarketingSiteRouter extends AssetsBuilder with Controller with Logging {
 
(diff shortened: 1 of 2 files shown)


eishay:
its `index.1.html` and `index.2.html`

raymondng:
Oh will fix. Thanks.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/MarketingSiteRouter.scala
```diff
@@ -34,8 +39,42 @@ object MarketingSiteRouter extends AssetsBuilder with Controller with Logging {
...
+        val ip = Try {
+          request.remoteAddress // remoteAddress looks up 'X-Forwarded-For'
+        } recover {
+          case t: Throwable =>
+            InetAddress.getLocalHost.toString // with affinity this might be sufficient
+        } get
+        val hasher = hashFunction.newHasher()
+        val hc = hasher.putString(ip, Charsets.UTF_8).hash()
+        val hash = (Math.abs(hc.asInt()) % 100) // rough
+        val winner = if (hash < 50) "index1" else "index2"
```
eishay: wow, that's hard core :-)
thought that `ip.hashCode` is enough

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/MarketingSiteRouter.scala
```diff
@@ -34,8 +39,42 @@ object MarketingSiteRouter extends AssetsBuilder with Controller with Logging {
...
+        log.error(s"[landing] Caught exception $t while parsing queryParam(v):${request.queryString("v")}")
+        None
+    } get
+
+    pickOpt match {
+      case Some(idx) if (idx == 1 || idx == 2) =>
+        "index." + idx
+      case None =>
+        val ip = Try {
+          request.remoteAddress // remoteAddress looks up 'X-Forwarded-For'
```
andrewconner: Is this known to throw exceptions?

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/MarketingSiteRouter.scala
```diff
@@ -34,8 +39,42 @@ object MarketingSiteRouter extends AssetsBuilder with Controller with Logging {
...
+      request.getQueryString("v").map(_.toInt)
+    } recover {
+      case t: Throwable =>
+        log.error(s"[landing] Caught exception $t while parsing queryParam(v):${request.queryString("v")}")
+        None
+    } get
+
+    pickOpt match {
+      case Some(idx) if (idx == 1 || idx == 2) =>
+        "index" + idx
```
eishay:
``` scala
 "index." + idx
```

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/MarketingSiteRouter.scala
raymondng: Oh good catch. I was using `InetAddress` so I wanted to be on the safe side. Now there is no need.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/MarketingSiteRouter.scala
raymondng: Not really :) I started out with `hashCode` and in fact that probably should work fine. Bouncing JVM might be an issue though.

eishay: HashCode is using the string chars, it's consistent across vms

raymondng: Yes you're right. My comment applies in general but does not apply to `String`, as it's a special case.

joshm1:
tested on dev with latest marketing pages. lgtm! thanks!

raymondng:
Cool. Thanks!

PR #12084, opened 2014-11-01
Angular 1.3

andrewconner:
The update was quite simple.

Cool new features:
- Many bug fixes
- Some huge performance gains in the watch loop
- Form improvements
- `$scope.$watchGroup`: watch several values together
- [`ngModelOptions`](https://docs.angularjs.org/api/ng/directive/ngModelOptions): configure debounces, etc in models
- [Lazy one-time binding](https://docs.angularjs.org/guide/expression#one-time-binding): `{{ ::yourVariable }}` gets evaluated until it is defined, and then never again. Great for templates where the data won't update (this is very common: two-way binding needs are fairly limited).

I've tested in dev mode and fixed the issues I saw. Will continue to test, and we can roll it out Monday.

@yipingliao @joshm1 @ahsu1230 @eishay @stkem @2is10


joshm1:
awesome!

eishay:
Cool :-)

yipingliao:
Yay! Am really happy about lazy one-time binding.

PR #12097, opened 2014-11-03
Fixing white bar, which messed up `overflow`

andrewconner:
@ahsu1230 

I'm not happy with this solution, but it seems to work.

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/app.styl
+++ b/kifi-backend/modules/shoebox/angular/src/app.styl
@@ -77,7 +77,6 @@ a[href]
 
 .kf-cols
   text-align: start
-  background-color: appBackgroundColor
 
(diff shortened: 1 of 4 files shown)


ahsu1230:
thanks a lot! such a tricky bug... =P
I think you should get to sound the gong lol

kifi-backend/modules/shoebox/angular/src/app.styl
```diff
@@ -105,10 +104,22 @@ a[href]
   border-bottom: separatorBorder
   margin: 10px 0
 
+.white-background
+  display: none
+
 .kf-logged-out
+  .white-background
+    display: block
+    content: ' '
```
2is10, line 113: `content` is irrelevant

2is10:
nicely done

PR #12098, opened 2014-11-03
do not show current library in attribution list

joshm1:
@stkem


stkem:
As discussed, this _may_ have a problem elsewhere due to modification of the keep objects libraries field.

PR #12100, opened 2014-11-03
remove library repo access from getBaicKeeps

ymatsuda:
@leogrim 

Correct me if I am wrong. keep.visibility and library.visibility are same, thus we don't have to access library repo in this code.


leogrim:
Indeed, good catch!

PR #12101, opened 2014-11-03
Dynamically size keep cards on first load.

yipingliao:
https://app.asana.com/0/17575799252940/19282420960489
@andrewconner Please let me know if keep cards are supposed to be triggered into resizing some other way. What I saw was that in the resizing code in `keep.js`, the initial card width variable was set to 0 because at the time the card sizing code was executed, the keep card wasn't visible yet. This bug fix forces the resizing code to execute after the cards are visible, but I am not sure if this is the best way. Thanks!


andrewconner:
Oh weird, this is unexpected. Perhaps related to new things, because it certainly worked before.

yipingliao:
I am going to merge now to fix the behavior, and will address any comments about where this should be done.

PR #12112, opened 2014-11-04
Fix collection table

leogrim:
@joshm1 @stkem

### DIFF
--- a/kifi-backend/modules/common/conf/evolutions/shoebox/260.sql
+++ b/kifi-backend/modules/common/conf/evolutions/shoebox/260.sql
@@ -1,0 +1,9 @@
+#Shoebox
+
+# --- !Ups
+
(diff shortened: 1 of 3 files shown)


joshm1:
:+1:

kifi-backend/modules/common/conf/evolutions/shoebox/260.sql
```diff
@@ -0,0 +1,9 @@
+#Shoebox
+
+# --- !Ups
+
+ALTER TABLE collection ADD CONSTRAINT collection_u_user_id_name UNIQUE (user_id, name);
```
stkem, line 5: The table already has a (non-unique) index on `(user_id, name)`. It would be better to just alter that to be be unique.

leogrim, line 5: Yep you're right, + I forgot the route

PR #12114, opened 2014-11-04
sending X-Kifi-Client: BE X.X.X[.X] HTTP header in all server requests

2is10:
cc @yingjieMiao @thassss @jaredpetker @andrewconner


yingjieMiao:
thanks!

PR #12115, opened 2014-11-04
fixed bulk keeping adding library attributions; can now keep on all search tags

joshm1:
@stkem @yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/keep/keep.js
+++ b/kifi-backend/modules/shoebox/angular/src/keep/keep.js
@@ -474,7 +474,6 @@ angular.module('kifi')
         // Internal methods.
         //
         function init() {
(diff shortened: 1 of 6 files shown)


kifi-backend/modules/shoebox/angular/src/keep/keep.js
```diff
@@ -474,7 +474,6 @@ angular.module('kifi')
         // Internal methods.
         //
         function init() {
-          scope.keptToLibraryIds = _.pluck(scope.keep.keeps, 'libraryId');
```
joshm1, line 477: removed this because it's redundant with a $watchCollections in this scope, which always gets called at least once when the directive is linked

yipingliao, line 477: Thanks for catching this!! I noticed the watch and took out a couple of these when merging but missed this one.

PR #12116, opened 2014-11-04
followers list from clicking on +number icon in library card

ahsu1230:
On Library Card, black +number can bring up a list of only followers (no invitees)

@andrewconner
@yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
@@ -259,6 +259,7 @@ angular.module('kifi')
           modalService.open({
             template: 'libraries/manageLibraryModal.tpl.html',
             modalData: {
+              followersOnly: false,
               library: scope.library,
               returnAction: function () {
                 libraryService.getLibraryById(scope.library.id, true).then(function (data) {
@@ -282,6 +283,15 @@ angular.module('kifi')
           scope.editKeepsText = scope.editKeepsText === 'Edit Keeps' ? 'Done Editing' : 'Edit Keeps';
         };
 
+        scope.showFollowers = function () {
+          modalService.open({
+            template: 'libraries/manageLibraryModal.tpl.html',
+            modalData: {
+              followersOnly: true,
+              library: scope.library
+            }
+          });
+        };
 
         //
         // Watches and listeners.
--- a/kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
@@ -259,7 +259,7 @@ angular.module('kifi')
           modalService.open({
             template: 'libraries/manageLibraryModal.tpl.html',
             modalData: {
-              followersOnly: false,
+              startOnFollowers: false,
               library: scope.library,
               returnAction: function () {
                 libraryService.getLibraryById(scope.library.id, true).then(function (data) {
@@ -287,7 +287,7 @@ angular.module('kifi')
           modalService.open({
             template: 'libraries/manageLibraryModal.tpl.html',
             modalData: {
-              followersOnly: true,
+              startOnFollowers: true,
               library: scope.library
             }
           });
--- a/kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
@@ -259,7 +259,7 @@ angular.module('kifi')
           modalService.open({
             template: 'libraries/manageLibraryModal.tpl.html',
             modalData: {
-              startOnFollowers: false,
+              pane: 'manage',
               library: scope.library,
               returnAction: function () {
                 libraryService.getLibraryById(scope.library.id, true).then(function (data) {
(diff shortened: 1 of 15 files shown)


kifi-backend/modules/shoebox/conf/shoebox.routes
```diff
@@ -344,7 +344,7 @@ POST    /site/libraries/:id/decline @com.keepit.controllers.website.LibraryContr
 POST    /site/libraries/:id/leave   @com.keepit.controllers.website.LibraryController.leaveLibrary(id: PublicId[Library])
 GET     /site/libraries/:id/keeps   @com.keepit.controllers.website.LibraryController.getKeeps(id: PublicId[Library], offset: Int ?= 0, limit: Int ?= 10)
 POST    /site/libraries/:id/keeps   @com.keepit.controllers.website.LibraryController.addKeeps(id: PublicId[Library])
-GET     /site/libraries/:id/members @com.keepit.controllers.website.LibraryController.getLibraryMembers(id: PublicId[Library], offset: Int ?= 0, limit: Int ?= 10)
+GET     /site/libraries/:id/members @com.keepit.controllers.website.LibraryController.getLibraryMembers(id: PublicId[Library], offset: Int ?= 0, limit: Int ?= 10, showInvites: Boolean ?= false)
```
andrewconner: We don't trust clients — only the owner should be able to get the invite list right now. No flag needed.

kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.tpl.html
```diff
@@ -3,25 +3,25 @@
 
   <div class="dialog-body-wrap">
     <h1 class="dialog-header">
-      <button ng-if="showFollowers" class="go-back-button" ng-click="hideFollowersPanel()">
+      <button ng-if="showFollowers && library.isMine" class="go-back-button" ng-click="hideFollowersPanel()">
```
andrewconner: Right now, since this is done, we can stick with it — but we'll want separate modals for non-owners. Non-owners JUST need the list, not everything else. As-is, we'll need to do a lot of `ng-if="library.isMine"` every time we add to this modal.

kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
```diff
@@ -259,6 +259,7 @@ angular.module('kifi')
           modalService.open({
             template: 'libraries/manageLibraryModal.tpl.html',
             modalData: {
+              startOnFollowers: false,
```
andrewconner: Perhaps `pane: 'invite'` and also allow `pane: 'members'` or something.

kifi-backend/modules/shoebox/conf/shoebox.routes
ahsu1230: I see. taking away the flag, and putting the logic in the controller

kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.tpl.html
ahsu1230: I don't mind making a separate modal, I figured it might be easier to just add a few things to the current manage lib modal & re-use code. But as you mentioned, it's probably not so sustainable for the future.

kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.js
```diff
@@ -174,7 +174,8 @@ angular.module('kifi')
           if (loading) { return; }
           if (scope.library.id) {
             loading = true;
-            libraryService.getMoreMembers(scope.library.id, pageSize, scope.offset).then(function (resp) {
+            var showInvites = scope.library.isMine;
```
yipingliao: Suggestion: not have an extra variable (`showInvites`). It's only used once and is the same value as `scope.library.isMine`.

kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.tpl.html
yipingliao: My bad. I suggested to @ahsu1230  that it may be a good idea to reuse the current modal so that owners can easily invite more people; didn't thoroughly consider the non-owner experience.

yipingliao:
:+1:

kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.tpl.html
andrewconner: Yeah, re-using for owners is great.

kifi-backend/modules/shoebox/app/com/keepit/model/KeepRepo.scala
```diff
@@ -400,22 +400,32 @@ class KeepRepoImpl @Inject() (
     (for (b <- rows if b.libraryId === libraryId && b.state =!= excludeState.orNull) yield b).sortBy(_.createdAt desc).drop(offset).take(limit).list
   }
 
-  private def getCountByLibraryCompiled(libraryId: Column[Id[Library]], excludeState: Option[State[Keep]]) = Compiled {
```
ahsu1230, line 403: some repo changes @ymatsuda thanks for the tip!

PR #12117, opened 2014-11-04
process keep attribution in batches

raymondng:
process keeps in batches
@stkem


stkem:
lgtm. If it remains a problem we may have to reduce the batch size a bit.

raymondng:
Sounds reasonable.

PR #12118, opened 2014-11-04
minor changes to marketing site router

2is10:
@eishay 

using 301 for .html redirects, letting Play! add the host, and only looking at the end of a path for .html


2is10:
Nevermind. I don’t have my local repo set up to properly try this out.

eishay:
thansk

PR #12121, opened 2014-11-04
towards multi-versioned models

yingjieMiao:
@stkem 

one model, wrapped in a new interface `MultiVersionedLDAURIRepresenter`


stkem:
lgtm

PR #12125, opened 2014-11-04
library indexing request should be made after commit

ymatsuda:
@andrewconner @ahsu1230 

Do not send indexing requests until the transaction is committed. Otherwise, it is very likely that the indexer is unable to see the new values when the request reaches it. The change won't be picked up until indexing is invoked later either by another request or by the 1-min periodical indexing cycle.


andrewconner:
Thanks

ahsu1230:
thanks

PR #12127, opened 2014-11-04
fix race condition in search with showing system library attribution

joshm1:
@stkem

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/common/directives/keepWho/keepWhoLibsAndPics.js
+++ b/kifi-backend/modules/shoebox/angular/src/common/directives/keepWho/keepWhoLibsAndPics.js
@@ -3,8 +3,8 @@
 angular.module('kifi')
 
 .directive('kfKeepWhoLibsAndPics', [
-  '$rootScope', 'friendService', 'keepWhoService', 'profileService',
-  function ($rootScope, friendService, keepWhoService, profileService) {
+  '$rootScope', 'friendService', 'keepWhoService', 'profileService', 'libraryService',
+  function ($rootScope, friendService, keepWhoService, profileService, libraryService) {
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/shoebox/angular/src/common/directives/keepWho/keepWhoLibsAndPics.js
```diff
@@ -36,7 +36,7 @@ angular.module('kifi')
           _.each(keeps, function (keep) {
             if (!scope.keep.id &&                      // No scope.keep.id if the keep is not on a library page.
                 scope.keep.url === keep.url &&
-                library.kind === 'user_created' &&     // Do not show system libraries as attributions.
+                !libraryService.isSystemLibrary(library) &&     // Do not show system libraries as attributions.
```
yipingliao, line 39: Nice! Should we/can we replace all existing uses of `library.kind === 'user_created'` with calls to this new method?

stkem, line 39: Gradually, as we come across it, I think.

kifi-backend/modules/shoebox/app/com/keepit/controllers/internal/ShoeboxController.scala
```diff
@@ -121,7 +121,7 @@ class ShoeboxController @Inject() (
     Ok("true")
   }
 
-  def processAndSendMail = Action.async(parse.tolerantJson) { request =>
+  def processAndSendMail = Action.async(parse.tolerantJson(maxLength = 1024 * 500)) { request =>
```
stkem, line 124: ahh, good one :-)

PR #12129, opened 2014-11-04
Add null check for followers on a library.

yipingliao:
https://app.asana.com/0/17575799252940/18870880724776

This null check should help with things looking really bad, but the root cause of why followers became undefined is still unclear. Checking this in to move on to other tasks, and will circle back to it.

@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
@@ -97,11 +97,10 @@ angular.module('kifi')
             scope.library.owner.picUrl = friendService.getPictureUrlForUser(scope.library.owner);
           }
 
-          if (_.isArray(scope.library.followers)) {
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
```diff
@@ -97,11 +97,10 @@ angular.module('kifi')
             scope.library.owner.picUrl = friendService.getPictureUrlForUser(scope.library.owner);
           }
 
-          if (_.isArray(scope.library.followers)) {
-            scope.library.followers.forEach(function (follower) {
-              follower.picUrl = friendService.getPictureUrlForUser(follower);
-            });
-          }
+          scope.library.followers = scope.library.followers || [];
```
yipingliao, line 100: When the bug is manifested, `scope.library.followers` is `undefined`.

andrewconner:
Great thanks

PR #12136, opened 2014-11-04
Fixing JS error when destroying during $digest cycle

andrewconner:
I believe this is the best fix for https://app.asana.com/0/17575799252940/19091436963747

We were getting a JS error that is caused by calling scope.$destroy _during_ a digest. Not sure if this is the best solution, but it seems to stop the error. @yipingliao @joshm1 Thoughts appreciated.


yipingliao:
Makes sense to me! Thanks for tracking this down.

joshm1:
thanks!

PR #12138, opened 2014-11-04
be more stingy

ymatsuda:
@andrewconner @ahsu1230 @stkem @eishay 
- added a different flavor of getCollectionsForKeep to KeepToCollectionRepo. It takes Keep instead of Id[Keep], so we don't have to access the bookmark table when we already have a Keep in memory. As the result, we do 2-way join instead of 3-way join.
- I noticed BasicCollectionCache is not invalidated when collections change. Invalidation is added to CollectionRepo and KeepToCollectionRepo.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/model/KeepToCollectionRepo.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/model/KeepToCollectionRepo.scala
@@ -1,6 +1,7 @@
 package com.keepit.model
 
 import com.google.inject.{ Provider, Inject, Singleton, ImplementedBy }
+import com.keepit.commanders.{ BasicCollectionByIdKey, BasicCollectionByIdCache }
 import com.keepit.common.db.slick.DBSession.{ RSession, RWSession }
 import com.keepit.common.db.slick._
 import com.keepit.common.db.{ State, Id }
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/app/com/keepit/model/KeepToCollectionRepo.scala
```diff
@@ -68,6 +72,21 @@ class KeepToCollectionRepoImpl @Inject() (
...
 
+  def getCollectionsForKeep(keep: Keep)(implicit session: RSession): Seq[Id[Collection]] = {
+    if (keep.isActive) {
+      collectionsForKeepCache.getOrElse(CollectionsForKeepKey(keep.id.get)) {
+        val query = for {
+          kc <- rows if kc.bookmarkId === keep.id.get && kc.state === KeepToCollectionStates.ACTIVE
+          c <- collectionRepo.rows if c.id === kc.collectionId && c.state === CollectionStates.ACTIVE
+        } yield kc
+
+        query.sortBy(_.updatedAt).map(_.collectionId).list // todo(martin): we should add a column for explicit ordering of tags
```
andrewconner, line 83: I don't think Martin will ever get to that :(

andrewconner:
Thanks, nice!

kifi-backend/modules/shoebox/app/com/keepit/model/KeepToCollectionRepo.scala
stkem, line 83: :-P

stkem:
:+1:

eishay:
good find. its a strange merge to start with, i would expect that the caller would not call it if the keep is not active.

PR #12139, opened 2014-11-04
fix on scrollbar of doom / death / despair

ahsu1230:
@andrewconner @yipingliao


yipingliao:
Thanks for your help, @ahsu1230 !

PR #12141, opened 2014-11-04
removing keywords from library meta tags

ymatsuda:
@eishay @andrewconner 

After a discussion with Eishay, we decided to remove keywords meta tag from library meta tags.
Keywords meta tag is not very effective for SEO. Some even says that it should not be used at all.


eishay:
:+1:  thanks!

eishay:
http://www.metatags.org/google_ignores_meta_tags_in_ranking

andrewconner:
Works for me

PR #12146, opened 2014-11-05
Fixing massive inefficency in our CSS output

andrewconner:
Whoops, my fault. Anything in `build-css/` get included in every Stylus file. This means this snippet was included quite a few times. The uncompressed CSS output was ~700kb (it _did_ compress quite well, though).

@yipingliao


yipingliao:
My fault too. I actually noticed this a couple of weeks back and always meant to ask you about it, but I only seem to remember this issue outside the office. :)  Thanks so much for fixing it!

PR #12147, opened 2014-11-05
library share (typeahead) result

raymondng:
This should address the library share typeahead issue of email contact name not being displayed.
Added bonus: no changes to frontend required.

Note that I didn't change `MaybeLibraryUser` as there are other dependencies.

@andrewconner @leogrim


leogrim:
Why do we need to introduce yet another class from `MaybeLibraryMember`, can't we just modify to use a `BasicContact` instead of an `EmailAddress`?

raymondng:
As discussed, there are other dependencies ... but maybe it's not too bad. Will take another look.

raymondng:
Thanks for the feedback. Will merge to build on it. Looks like we need an additional API for `BasicContact`

leogrim:
Awesome thanks!

PR #12149, opened 2014-11-05
3rd step towards multi-versioned models

yingjieMiao:
@stkem

Most methods in `ldaCommander` now take an implicit parameter `version`.
`ldaCommander` is called in `ldaController`, where I placed an `implicit val version`. 

Upshot: 1) no clients need to change APIs yet. 2) Since there is only one version, no ambiguous implicit yet.


stkem:
:+1:

PR #12153, opened 2014-11-05
Add errors to create library panel in library widget

yipingliao:
@andrewconner 

Actually, addresses the following:
- https://app.asana.com/0/18079737520750/18688718389292
- https://app.asana.com/0/18079737520750/18998731396271
- https://app.asana.com/0/18079737520750/19402371355029

Also:
- Fixes small scroll issues
- Fixes logical error in Fuse search where no search results for a non-empty search query results in the entire list of user libraries being shown.

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/keep/keepToLibraryWidget.js
+++ b/kifi-backend/modules/shoebox/angular/src/keep/keepToLibraryWidget.js
@@ -46,13 +46,14 @@ angular.module('kifi')
         // Internal data.
         //
         var widget = null;
-        var selectedIndex = 0;
         var searchInput = null;
         var libraryList = null;
+        var newLibraryNameInput = null;
+
+        var selectedIndex = 0;
         var justScrolled = false;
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/shoebox/angular/src/keep/keepToLibraryWidget.js
```diff
@@ -346,7 +343,7 @@ angular.module('kifi')
 
         scope.onSearchInputChange = function (name) {
           var matchedLibraries = libraryNameSearch.search(name);
-          groupWidgetLibraries(matchedLibraries.length ? matchedLibraries : allLibraries);
+          groupWidgetLibraries(!name ? allLibraries : matchedLibraries);
```
yipingliao, line 346: This is the fix for the Fuse search problem where no matches result in all libraries being shown. What I really wanted was for all libraries to be shown when the search query is empty.

andrewconner:
Wahoo, awesome!

PR #12155, opened 2014-11-05
Fix occurrences of KifiHitContext in HelpRankEventTrackingCommander

leogrim:
@eishay fix build


kifi-backend/modules/heimdal/app/com/keepit/commander/HelpRankEventTrackingCommander.scala
```diff
@@ -66,7 +66,7 @@ class HelpRankEventTrackingCommander @Inject() (
               // todo(ray): remove dependency on KifiHitContext
```
leogrim, line 66: @raymondng could you please take action on this todo when you have some time? 
SearchAnalytics shouldn't be in Common.

PR #12156, opened 2014-11-05
less lookups of library memberships

ymatsuda:
@andrewconner @leogrim @eishay  
- getBasicKeeps
  - trying to reduce lookups of library memberships
- decorateKeepsIntoKeepInfos
  - moved creation of pageInfosFuture after creation of basicInfosFuture because creation of pageInfosFuture is expensive. It is better to set up the future chain (augmentationFuture->basicInfosFuture) earlier


andrewconner:
Awesome!

PR #12172, opened 2014-11-05
Fix bug where invited header is not showing up on library card.

yipingliao:
@ahsu1230 Can you take a quick look? Thanks!
https://app.asana.com/0/17575799252940/19398903825426

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
@@ -130,7 +130,7 @@ angular.module('kifi')
           // Figure out whether this library is a library that the user has been invited to.
           // If so, display an invite header.
           var promise = null;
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
```diff
@@ -130,7 +130,7 @@ angular.module('kifi')
           // Figure out whether this library is a library that the user has been invited to.
           // If so, display an invite header.
           var promise = null;
-          if (libraryService.invitedSummaries) {
+          if (libraryService.invitedSummaries.length) {
```
yipingliao, line 133: The previous code was a bad check because `libraryService.invitedSummaries` is `[]` when empty, which is truthy. :)

ahsu1230:
yea, that might actually be it. I have a few questions later, so I'll roll by when you're ready =)

PR #12178, opened 2014-11-05
Fix tag counts

leogrim:
@ahsu1230 @andrewconner


ahsu1230:
Thanks a lot! This makes so many things much cleaner! =)

PR #12182, opened 2014-11-06
send page social tags only if we suspect the receiving side is a scraping bot of some sort

eishay:


eishay:
@andrewconner

eishay:
@ymatsuda

PR #12183, opened 2014-11-06
fix last_viewed sorting (doesn't update as users visit libraries)

ahsu1230:
Last Viewed sorting doesn't dynamically resort whenever you visit your libraries. You'd have to refresh in order to get the correct ordering. This uses angular's `Date.now()` to dynamically update the nav libs list.

@andrewconner @yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
+++ b/kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
@@ -98,6 +98,19 @@ angular.module('kifi')
         });
         scope.$on('$destroy', deregisterChangedLibrary);
 
+        var deregisterLibraryVisited = $rootScope.$on('lastViewedLib', function (e, lib) {
+          if (lib.name) {
+            var ind = _.findIndex(scope.userLibsToShow, function(l) { return l.name === lib.name; });
+            if (scope.userLibsToShow[ind]) {
+              scope.userLibsToShow[ind].lastViewed = Date.now();
+            }
+            if (scope.sortingMenu.option === 'last_viewed') {
+              updateNavLibs();
+            }
+          }
+        });
+        scope.$on('$destroy', deregisterLibraryVisited);
+
         scope.$watch(function () {
           return friendService.requests.length;
         }, function (value) {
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
```diff
@@ -98,6 +98,19 @@ angular.module('kifi')
...
         scope.$on('$destroy', deregisterChangedLibrary);
 
+        var deregisterLibraryVisited = $rootScope.$on('lastViewedLib', function (e, lib) {
+          if (lib.name) {
+            var ind = _.findIndex(scope.userLibsToShow, function(l) { return l.name === lib.name; });
+            if (scope.userLibsToShow[ind]) {
+              scope.userLibsToShow[ind].lastViewed = Date.now();
+            }
+            if (scope.sortingMenu.option === 'last_viewed') {
+              updateNavLibs();
```
ahsu1230, line 108: only update the list if you're on `last_viewed` sorting option. But, still record times of which libraries you're visiting, so correct sorting persists whether you're in `last_viewed` sorting or not

kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
```diff
@@ -98,6 +98,19 @@ angular.module('kifi')
         });
         scope.$on('$destroy', deregisterChangedLibrary);
 
+        var deregisterLibraryVisited = $rootScope.$on('lastViewedLib', function (e, lib) {
+          if (lib.name) {
+            var ind = _.findIndex(scope.userLibsToShow, function(l) { return l.name === lib.name; });
```
yipingliao: `var index = _.findIndex(scope.userLibsToShow, { 'name': lib.name })`
If you don't like `index` try `idx` (more idiomatic than `ind`?)

yipingliao:
Nice!

kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
ahsu1230: :+1: dohhh I keep forgetting that notation. Thanks

PR #12185, opened 2014-11-06
and then there are two

raymondng:
Removed a bunch of sync\* calls (now there are two left).
Aside from 'async-ify' -- no major behavior/functional change. 
Note that alongside with the sync calls, the locks also go away. Given @ymatsuda's new async seqNum update scheme we should be ok.
In general the changes are rather mechanical in nature -- as mechanical as this sort of change allows. 

@ymatsuda @leogrim @yingjieMiao @andrewconner


ymatsuda:
:+1:

raymondng:
Thanks for the feedback! Will merge and build upon it.

PR #12188, opened 2014-11-06
2 chunks in extension search response instead of 4

2is10:
@leogrim

### DIFF
--- a/kifi-backend/modules/search/app/com/keepit/controllers/ext/ExtSearchController.scala
+++ b/kifi-backend/modules/search/app/com/keepit/controllers/ext/ExtSearchController.scala
@@ -104,10 +104,10 @@ class ExtSearchController @Inject() (
       Ok.chunked(plainResultEnumerator
         .andThen(augmentationEnumerator)
         .andThen(Enumerator.eof)).as(TEXT)
-    } else {
+    } else {  // JSON format last used by extension 3.3.10
       Ok.chunked(Enumerator("[").andThen(plainResultEnumerator)
         .andThen(Enumerator(",")).andThen(augmentationEnumerator)
-        .andThen(Enumerator("]").andThen(Enumerator.eof)))
+        .andThen(Enumerator("]").andThen(Enumerator.eof))).as(JSON)
     } withHeaders ("Cache-Control" -> "private, max-age=10")
   }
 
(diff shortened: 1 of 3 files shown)


2is10:
In the browser I’m seeing the XHR progress event fire specifically for the final closing bracket `]` in the response, which seems wasteful.

leogrim:
:+1:

ymatsuda:
I intensionally made brackets and commas separate chunks by themselves to avoid making copies of large strings.

ymatsuda:
How bad is this for browsers? @2is10

2is10:
@ymatsuda I’m aware of that expense, but I figured it was the lesser of two evils. The [XHR spec says](http://www.w3.org/TR/XMLHttpRequest2/#make-progress-notifications):

> When it is said to make progress notifications, while the download is progressing, queue a task to fire a progress event named progress about every 50ms or for every byte received, whichever is least frequent.

The latency the string concatenation adds on the server is negligible compared with up to a 50ms delay in the browser (between the opening `[` and the end of the first chunk, for example). And today we’re already doing a string concatenation to prepend the comma to the second chunk, so I figured a concatenation on the first chunk wouldn’t be too much worse. Plus, the responses aren’t huge—my first chunk when I search for "scala" (5 hits) is 1211 chars (2422 bytes in JVM).

That being said, I don’t see any evidence of a delay due to browser chunk buffering. (The developer tools don’t expose chunk arrival times. I’d need to run a custom build of each browser or use wireshark to measure delays.) So this may not be solving a real problem.

I think our best bet would be to stop inserting the `[` and `,` and `]` and just serve the response as `text/plain` (two adjacent JSON values). That would break existing extensions, so we’d need a new endpoint for it, or to temporarily accept an additional query parameter or a specific `Accept` header value or something to specifically request a `text/plain` response.

2is10:
In that commit, I added a code path to omit `[` and `,` and `]` if the `Accept` header is `text/plain`. Also modified the existing code path not to string-concatenate the comma with the augmentation chunk.

There is currently no coverage of this endpoint in `ExtSearchControllerTest`. I’ve spent some time on adding test coverage tonight, but it’s not ready yet. I’ll finish that tomorrow before merging this. In the meantime, comments welcome.

ymatsuda:
2KB is not small for me. It stresses JVM memory system, more chances of GCs, more chances of promotion to old gen, wastes CPU cycles, pollutes CPU cache... I know I am obsessed. But we will have more data and more users soon (hopefully). I want to cut every corner.

Anyways, I like your idea of omitting brackets and commas. It totally makes sense to me.

2is10:
Added basic tests for the two response formats. Anyone can feel free to flesh them out a little further.

PR #12194, opened 2014-11-06
send the first byte to the client asap

eishay:
Thanks to @andrewconner


eishay:
@andrewconner please review

andrewconner:
Looks good.

PR #12195, opened 2014-11-06
Better logging when there's an error parsing user id

andrewconner:
@eishay @raymondng 

The logging on UserActions.scala:236 is hard to connect with this error.

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
+++ b/kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
@@ -126,11 +126,11 @@ trait UserActionsHelper extends UserActionsRequirements with Logging {
     }
 
   def getUserIdOptWithFallback(implicit request: Request[_]): Future[Option[Id[User]]] = {
-    val kifiIdOpt = getUserIdFromSession recover {
+    val kifiIdOpt = getUserIdFromSession.recover {
       case t: Throwable =>
-        airbrake.notify(s"[getUserIdOpt] Caught exception $t while retrieving userId from request; cause=${t.getCause}", t)
+        airbrake.notify(s"[getUserIdOpt] Caught exception $t while retrieving userId from request; cause=${t.getCause}. Path: ${request.path}, headers: ${request.headers.toMap}", t)
         None
-    } get
+    }.get
 
     kifiIdOpt match {
       case Some(userId) =>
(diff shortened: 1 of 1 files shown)


eishay:
lgtm

kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
```diff
@@ -297,7 +297,7 @@ object KifiSession {
   val FORTYTWO_USER_ID = "fortytwo_user_id"
 
   implicit class HttpSessionWrapper(val underlying: Session) extends AnyVal {
-    def setUserId(userId: Id[User]): Session = underlying + (FORTYTWO_USER_ID -> userId.toString)
+    def setUserId(userId: Id[User]): Session = underlying + (FORTYTWO_USER_ID -> userId.id.toString)
```
raymondng, line 300: Given the implementation, this `.id` shouldn't be necessary. Also we have unit tests to cover this ... am I missing something?

kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
```diff
@@ -126,11 +126,11 @@ trait UserActionsHelper extends UserActionsRequirements with Logging {
     }
 
   def getUserIdOptWithFallback(implicit request: Request[_]): Future[Option[Id[User]]] = {
-    val kifiIdOpt = getUserIdFromSession recover {
+    val kifiIdOpt = getUserIdFromSession.recover {
       case t: Throwable =>
-        airbrake.notify(s"[getUserIdOpt] Caught exception $t while retrieving userId from request; cause=${t.getCause}", t)
+        airbrake.notify(s"[getUserIdOpt] Caught exception $t while retrieving userId from request; cause=${t.getCause}. Path: ${request.path}, headers: ${request.headers.toMap}", t)
         None
```
eishay, line 132: @andrewconner @ray So if there's an exception while parsing the session, are we going to remove that cookie/session?

raymondng:
lgtm. Thanks!

kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
raymondng, line 132: We should re-set it to the proper value.

kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
andrewconner, line 300: Correct, this was just cleanup I noticed. Using an object's `toString` can be dangerous because someone may change that (thinking they're being helpful and making it "Id(123)" when it used to be "123"). And yep, there's a test for this — in general, though, better to not have to rely on that. If someone did indeed change `Id`'s `toString`, the fix would be this.

raymondng, line 300: Yup sounds good to me. Better be extra careful. Thanks!

kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
eishay, line 132: @raymondng by resetting do you mean wipe it out?
there could be many reason for corrupted session, don't know if we can know the proper value.
Would you take care of it?

raymondng, line 132: By reset I mean re-set to the user we looked up using `SecureSocial` -- which should be fine as we relied on `SecureSocial` cookie exclusively for the longest time.
Certainly I will see if we can make it even more robust.

PR #12197, opened 2014-11-06
Fixing broken endpoint

andrewconner:


kifi-backend/modules/shoebox/angular/src/common/services/routeService.js
```diff
@@ -89,7 +89,7 @@ angular.module('kifi')
         return env.xhrBase + '/user/' + id + '/friend';
       },
       libraryShareSuggest: function (libId, opt_query) {
-        return route('/libraries/' + libId + '/members/suggest?n=100' + (opt_query ? '&q=' + opt_query : ''));
+        return route('/libraries/' + libId + '/members/suggest?n=10' + (opt_query ? '&q=' + opt_query : ''));
```
andrewconner, line 92: @yipingliao Why was this changed to 100? Backend returns an error if you request more than 30. If there's a good reason, let's talk — backend can be changed.

PR #12201, opened 2014-11-06
Invalidate cache when refetching library information after library has been modified.

yipingliao:
@andrewconner 
https://app.asana.com/0/17575799252940/19507268943159

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
@@ -21,6 +21,12 @@ angular.module('kifi')
       templateUrl: 'libraries/libraryCard.tpl.html',
       link: function (scope, element/*, attrs*/) {
         //
+        // Internal data.
+        //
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
```diff
@@ -268,7 +274,7 @@ angular.module('kifi')
               library: scope.library,
               returnAction: function () {
                 libraryService.getLibraryById(scope.library.id, true).then(function (data) {
-                  libraryService.getLibraryByUserSlug(scope.username, data.library.slug, true).then(function (library) {
+                  libraryService.getLibraryByUserSlug(scope.username, data.library.slug, authToken, true).then(function (library) {
```
yipingliao, line 277: The problem was that the arguments were not updated after this change:
https://github.com/kifi/keepit/commit/be04ac89ba2e2f8fb1f5af64b7c22ee638dfb228#diff-a611a1e98f78c4aae99dc9f0f6d47a6fR128

andrewconner:
Thanks

ahsu1230:
thanks

PR #12202, opened 2014-11-06
Adding scrollbar to modals if window size is too small

ahsu1230:
@andrewconner @yipingliao


kifi-backend/modules/shoebox/angular/src/tagManage/tagManage.js
```diff
@@ -151,7 +151,7 @@ angular.module('kifi')
       libraryService.copyKeepsFromTagToLibrary(clickedLibrary.id, tagName).then(function () {
         modalService.open({
           template: 'tagManage/tagToLibModal.tpl.html',
-          modalData: { library : clickedLibrary, action: $scope.actionToLibrary }
+          modalData: { library : clickedLibrary, action: 'copy' }
```
ahsu1230, line 154: this only affects the message in the pop-up modal after copying/keeping tags
turns out this was never set lol

yipingliao, line 154: Oops! Thanks for the catch.

PR #12204, opened 2014-11-06
Take out event listener in basicModal

ahsu1230:
the window resize event listener is in kfModal directive anyway - and basicModal should always be used within kfModal.

@yipingliao @andrewconner


yipingliao:
Thanks for cleaning this up!

PR #12205, opened 2014-11-06
RIP SyncShoeboxDbCallbacks

raymondng:
Did away with the final 2 synchronous db callbacks, and with that `SyncShoeboxDbCallbacks` can be removed!

Work was done yesterday after the 'almost async' PR. 

@leogrim @ymatsuda @yingjieMiao @andrewconner


raymondng:
Will merge and build on it. Feedback appreciated.

eishay:
removing the await is (almost) always a good idea. if we leave this pull request open too long it will rot

raymondng:
Good idea. I held off due to issues over weekend and other priorities. Will
take care it.

kifi-backend/modules/scraper/app/com/keepit/scraper/ScrapeWorker.scala
```diff
@@ -333,20 +332,27 @@ class ScrapeWorkerImpl @Inject() (
...
+      }
+    }
+
+    val filtered = redirects.dropWhile(!_.isLocatedAt(uri.url))
+    if (filtered.headOption.exists(redirect => !redirect.isPermanent)) {
+      Future.successful(updateRedirectRestriction(uri, filtered.head))
+    } else {
+      hasFishy301(uri) flatMap { isFishy =>
+        if (isFishy) {
+          Future.successful(updateRedirectRestriction(uri, filtered.head))
```
leogrim, line 353: Here `filtered.head` might throw an exception. In the previous code that case was under `case Seq(redirect, _*)` which assumes that there is indeed a `head` element. 

Well done making this async otherwise!

leogrim:
See comment, otherwise :+1: thanks!

kifi-backend/modules/scraper/app/com/keepit/scraper/ScrapeWorker.scala
raymondng, line 353: Thanks Léo. Note that this line is guarded by a `filtered.headOption.exists` which means this should be ok.
However I do feel that (while working on it) the code is not as clean as I'd have liked -- any suggestions welcomed!

raymondng, line 353: Oh you meant the second one. Sorry I mis-read -- let me take a closer look.

raymondng:
Thanks! Appreciate the review

PR #12208, opened 2014-11-06
join / leave library backend endpoints for Extension

ahsu1230:
@2is10 

Endpoints only receive a public Library Id & will either receive a No Content (204) or Bad Request with a message.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtLibraryController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtLibraryController.scala
@@ -98,6 +98,28 @@ class ExtLibraryController @Inject() (
     }
   }
 
+  def joinLibrary(libraryPubId: PublicId[Library]) = UserAction { request =>
+    decode(libraryPubId) { libraryId =>
+      implicit val context = heimdalContextBuilder.withRequestInfoAndSource(request, KeepSource.keeper).build
+      libraryCommander.joinLibrary(request.userId, libraryId) match {
+        case Left(fail) =>
+          BadRequest(Json.obj("error" -> fail.message))
+        case Right(lib) =>
+          NoContent
+      }
+    }
+  }
+
+  def leaveLibrary(libraryPubId: PublicId[Library]) = UserAction { request =>
+    decode(libraryPubId) { libraryId =>
+      implicit val context = heimdalContextBuilder.withRequestInfoAndSource(request, KeepSource.keeper).build
+      libraryCommander.leaveLibrary(libraryId, request.userId) match {
+        case Left(fail) => BadRequest(Json.obj("error" -> fail.message))
+        case Right(_) => NoContent
+      }
+    }
+  }
+
   def addKeep(libraryPubId: PublicId[Library]) = UserAction(parse.tolerantJson) { request =>
     decode(libraryPubId) { libraryId =>
       db.readOnlyMaster { implicit s =>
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtLibraryController.scala
```diff
@@ -143,6 +143,28 @@ class ExtLibraryController @Inject() (
...
+          NoContent
+      }
+    }
+  }
+
+  def leaveLibrary(libraryPubId: PublicId[Library]) = UserAction { request =>
+    decode(libraryPubId) { libraryId =>
+      implicit val context = heimdalContextBuilder.withRequestInfoAndSource(request, KeepSource.keeper).build
+      libraryCommander.leaveLibrary(libraryId, request.userId) match {
+        case Left(fail) => BadRequest(Json.obj("error" -> fail.message))
```
ahsu1230: I really like having response statuses (2xx, 4xx) be returned from the Commander.
Right now, the class `LibraryFail` (`fail` in this line) is just a wrapper around a string. I'd like to add in an Int field to the class so I can specify failure responses better with different 4xx statuses. However, the site & mobile may depend on Bad Request statuses (hopefully they depend on error tokens instead), so I'll double check with everybody before making any changes =)

kifi-backend/modules/shoebox/test/com/keepit/controllers/ext/ExtLibraryControllerTest.scala
```diff
@@ -479,6 +562,16 @@ class ExtLibraryControllerTest extends Specification with ShoeboxTestInjector wi
     controller.deleteLibrary(libraryId)(request(routes.ExtLibraryController.deleteLibrary(libraryId)))
   }
 
+  private def joinLibrary(user: User, libraryId: PublicId[Library])(implicit injector: Injector): Future[Result] = {
+    inject[FakeUserActionsHelper].setUser(user)
+    controller.joinLibrary(libraryId)(request(routes.ExtLibraryController.joinLibrary(libraryId)))
+  }
```
2is10, line 582: I like the ordering you chose in this file, putting `joinLibrary` and `leaveLibrary` right after the `get`/`create`/`delete` methods. Before merging, I’d appreciate it if you’d move the methods in the controller and the routes in the routes file to be consistent with the ordering here. Thanks.

2is10:
:+1: Thanks!

kifi-backend/modules/shoebox/test/com/keepit/controllers/ext/ExtLibraryControllerTest.scala
```diff
@@ -182,6 +182,89 @@ class ExtLibraryControllerTest extends Specification with ShoeboxTestInjector wi
...
+
+        // leave a library
+        status(leaveLibrary(user2, libPubId1)) === NO_CONTENT
+        db.readOnlyMaster { implicit s =>
+          libraryMembershipRepo.getWithLibraryIdAndUserId(lib1.id.get, user2.id.get, None).get.state === LibraryMembershipStates.INACTIVE
+          libraryMembershipRepo.countWithLibraryId(lib1.id.get) === 1
+        }
+
+        // leave a library (no membership)
+        status(leaveLibrary(user2, libPubId2)) === BAD_REQUEST
```
2is10: In general, I like actions to be idempotent whenever possible, because they make for a better user experience in the event of network errors or data synchronization issues that can result in the state of the UI not accurately reflecting state on the server. Okay with you if we return 204 in this case? (Or, if you prefer, a 200 with a body indicating whether anything changed, perhaps just `true` or `false`.)

The idempotent mentality corresponds to thinking of these requests more broadly as “make sure I’m following (or not following) this library”.

It’d be great to see an assertion that joining a library is idempotent too. Thanks.

kifi-backend/modules/shoebox/test/com/keepit/controllers/ext/ExtLibraryControllerTest.scala
ahsu1230, line 582: :+1:

ahsu1230:
now with some assertions to guarantee idempotency

PR #12209, opened 2014-11-06
3.3.11: better timing numbers for Kifi/Google result analytics

2is10:
@leogrim

### DIFF
--- a/packrat/scripts/google_inject.js
+++ b/packrat/scripts/google_inject.js
@@ -39,7 +39,7 @@ if (searchUrlRe.test(document.URL)) !function () {
   var refinements = -1;   // how many times the user has refined the search on the same page. No searches at all yet.
   var showMoreOnArrival;
   var clicks = {kifi: [], google: []};  // clicked result link hrefs
-  var tQuery, tGoogleResultsShown, tKifiResultsReceived, tKifiResultsShown;  // for timing stats
+  var time = {google: {shown: 0}, kifi: {queried: 0, received: 0, shown: 0}};  // for timing stats
 
   var $q = $(), $qf = $q, $qp = $q, keyTimer;
   $(function() {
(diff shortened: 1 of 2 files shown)


packrat/scripts/google_inject.js
```diff
@@ -86,9 +86,9 @@ if (searchUrlRe.test(document.URL)) !function () {
         "maxResults": response.prefs.maxResults,
         "kifiResults": response.hits.length,
         "kifiExpanded": response.expanded || false,
-        "kifiTime": tKifiResultsReceived - tQuery,
-        "kifiShownTime": tKifiResultsShown - tQuery,
-        "thirdPartyShownTime": tGoogleResultsShown - tQuery,
+        "kifiTime": Math.max(-1, time.kifi.received - time.kifi.queried),
+        "kifiShownTime": Math.max(-1, time.kifi.shown - time.kifi.queried),
+        "thirdPartyShownTime": Math.max(-1, time.google.shown - time.kifi.queried),
```
2is10, line 91: `-1` will signify that the event did not occur.

packrat/scripts/google_inject.js
```diff
@@ -258,9 +257,12 @@ if (searchUrlRe.test(document.URL)) !function () {
     if (isVertical) return;
     if (attachKifiRes() > 0) {
       log('[withMutations] Google results inserted');
-      tGoogleResultsShown = Date.now();
-      if (!(tKifiResultsShown >= tKifiResultsReceived)) {
-        tKifiResultsShown = tGoogleResultsShown;
+      var now = Date.now();
+      if (time.google.shown < time.kifi.queried) {  // updating if we haven't searched would make us look artificially good
+        time.google.shown = now;
+      }
```
2is10, line 263: The new `if` guard here should help prevent a lot of the negative times we were seeing before any refinements.
(Our notion of refinements is different from Google’s, so the timing comparison is inherently unreliable when `.refinement != 0`.)

PR #12210, opened 2014-11-06
add xsl to the sitemap

eishay:
@andrewconner @raymondng


raymondng:
Cool. Do we plan to expose it to end users?

eishay:
For hacker/excited users who want a list of all libraries and for us to do a quick view on the xml. Its not going to be linked from the site. 
We should have a proper designed pages for regular users. 
The RSS, will actually be the end user endpoints.

raymondng:
Good idea :+1:

PR #12211, opened 2014-11-06
Fix sizeImage problem due to invisibility of keeps on loading.

yipingliao:
@andrewconner @ahsu1230


ahsu1230:
thanks for the fix! :us:

PR #12215, opened 2014-11-06
add response status to LibraryFail -> allow LibraryCommander to persist ...

ahsu1230:
...specific response status to Controllers

I've decided to add a `status` field to LibraryFail to easily differentiate 4xx status codes, so they may (1) persist to the Controllers and (2) be easily changed in the future to be consistent among all clients.
I know right now, the site & mobile don't check for 4xx status codes when it comes to library, but in case they would like them in the future, we can easily change responses accordingly

@andrewconner @2is10

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
@@ -169,7 +169,7 @@ class LibraryCommander @Inject() (
     (libInfo, accessStr)
   }
 
-  def getLibraryWithOwnerAndCounts(libraryId: Id[Library], viewerUserId: Id[User]): Either[(Int, String), (Library, BasicUser, Int, Int)] = {
+  def getLibraryWithOwnerAndCounts(libraryId: Id[Library], viewerUserId: Id[User]): Either[LibraryFail, (Library, BasicUser, Int, Int)] = {
     db.readOnlyReplica { implicit s =>
       val library = libraryRepo.get(libraryId)
       if (library.visibility == LibraryVisibility.PUBLISHED ||
@@ -180,7 +180,7 @@ class LibraryCommander @Inject() (
         val followerCount = libraryMembershipRepo.countWithLibraryIdByAccess(library.id.get).apply(LibraryAccess.READ_ONLY)
         Right(library, owner, keepCount, followerCount)
       } else {
-        Left(403, "library_access_denied")
+        Left(LibraryFail(FORBIDDEN, "library_access_denied"))
       }
     }
   }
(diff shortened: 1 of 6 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -180,7 +180,7 @@ class LibraryCommander @Inject() (
         val followerCount = libraryMembershipRepo.countWithLibraryIdByAccess(library.id.get).apply(LibraryAccess.READ_ONLY)
         Right(library, owner, keepCount, followerCount)
       } else {
-        Left(403, "library_access_denied")
+        Left(LibraryFail(FORBIDDEN, "library_access_denied"))
```
2is10, line 183: To me the `LibraryFail` type seems redundant with `Left`, and just makes the code longer. It’s already clear that `Left` means failure—the scaladoc for `Either` says “Convention dictates that Left is used for failure and Right is used for success.”

kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtLibraryController.scala
```diff
@@ -92,7 +92,7 @@ class ExtLibraryController @Inject() (
     decode(libraryPubId) { libraryId =>
       implicit val context = heimdalContextBuilder.withRequestInfoAndSource(request, KeepSource.keeper).build
       libraryCommander.removeLibrary(libraryId, request.userId) match {
-        case Some((status, message)) => Status(status)(Json.obj("error" -> message))
+        case Some(fail) => Status(fail.status)(Json.obj("error" -> fail.message))
```
2is10, line 95: I guess it’s a matter of taste as to whether you like having a wrapper object on the receiving end, like here.

2is10:
:+1:

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
ahsu1230, line 183: I think LibraryFail is used consistently as the Left field of LibraryCommander API return signatures. So maybe it is redundant and might as well just get rid of it altogether? 
But on the other hand, if we want to add fields to LibraryFail, having the wrapper class instead of a tuple would be useful right?

andrewconner, line 183: Yep

2is10, line 183: True. I generally choose not writing extra code today to optimize for what ifs. I try to optimize code for what it does today. But this is a very small issue and doesn’t really matter either way.

PR #12216, opened 2014-11-06
mitigation of deadlock in sequence number assigners

ymatsuda:
@stkem @eishay @yingjieMiao , Please review carefully.

Deadlocks we are seeing in Curator seem to be caused by conflicts between the raw item ingestion and the sequence number assigner. If I am right, both tasks update the same table in batch manner. Since they work as batch, a deadlock is possible unless we constrain them completely mutually exclusive. It is a possible solution. But, I am not doing it here. This change is only making the sequence number assigner more tolerable to errors.

The change:
- keep track of the number of transaction errors
- upon an error
  -  if the error count is already 1, throw exception
  - otherwise
  - increment the error count
  - reduce the batch size to the half of the current batch size
- upon a successful transaction
  - reset the error count to 0
  - reset the batch size to the default

Possible improvements:
- the tolerable error count can be bigger or configurable
- the logic of the batch size adjustment can be smarter

A similar error handling can be done by the ingestion side, too.
Or, simply breaking ingestion in to smaller transactions may be enough

### DIFF
--- a/kifi-backend/modules/sqldb/app/com/keepit/common/db/DbSequenceAssigner.scala
+++ b/kifi-backend/modules/sqldb/app/com/keepit/common/db/DbSequenceAssigner.scala
@@ -4,6 +4,8 @@ import com.keepit.common.db.slick.{ Database, Repo, SeqNumberFunction }
 import com.keepit.common.healthcheck.AirbrakeNotifier
 import com.keepit.common.logging.Logging
 import com.keepit.common.plugin.{ SequenceNumberAssignmentStalling, SequenceAssigner }
+import java.util.concurrent.atomic.AtomicInteger
+import scala.math._
 
 abstract class DbSequenceAssigner[M <: ModelWithSeqNumber[M]](
     db: Database,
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/sqldb/app/com/keepit/common/db/DbSequenceAssigner.scala
```diff
@@ -11,14 +13,28 @@ abstract class DbSequenceAssigner[M <: ModelWithSeqNumber[M]](
...
+        currentBatchSize = batchSize
+      } catch {
+        case e: UnsupportedOperationException =>
+          reportUnsupported(e)
+          throw e
+        case e: Throwable =>
+          if (errorCount.getAndIncrement > 0) {
+            throw e
+          } else {
+            currentBatchSize = max(currentBatchSize / 2, 1)
```
yingjieMiao, line 34: minor: maybe slightly higher lower bound?

yingjieMiao:
thanks. lgtm

stkem:
nice, lgtm!

kifi-backend/modules/sqldb/app/com/keepit/common/db/DbSequenceAssigner.scala
ymatsuda, line 34: Currently only one error is tolerated, so the lowest batch size is 10 (the default is 20). This is just a safe guard just in case we change the tolerable error count or change the default by overriding the default batch size.

PR #12217, opened 2014-11-06
Get Facebook Invites to behave (hopefully)

leogrim:
@andrewconner @stkem @joshm1 
This code gets rid of redirects for invites and replaces OpenGraph tags with the invite text. 
It does not insert the tag if it's not already there, and we need to fix one of the two marketing pages in that regard.


stkem:
:+1:

PR #12218, opened 2014-11-07
Style create library modal to display server-side errors.

yipingliao:
@ahsu1230 

Thanks for your help on this!


ahsu1230:
Thanks! I think we talked about it before, but the response comes back with `library_name_or_slug_exists`
Do you want to add that into that error switch statement?

yipingliao:
Are you going to change that to branch into two different errors? Either way is fine. If it's a single error, I will just say so in the error message.

2is10:
+1 for splitting that error into two for a better user experience.

PR #12223, opened 2014-11-07
more library click and view event tracking

joshm1:
@stkem


kifi-backend/modules/shoebox/angular/src/signup/signupService.js
```diff
@@ -83,6 +83,10 @@ angular.module('kifi')
     };
 
     var registerModal = function () {
+      if ($scope.userData.libraryId) {
+        $rootScope.$broadcast('trackLibraryEvent', 'view', { type: 'signupLibrary' });
```
stkem, line 87: Would it perhaps encapsulate this better to use `emit` and then listen on the root scope? Might be faster too since there is no propagation.

PR #12226, opened 2014-11-07
feed: top libraries

raymondng:
Baby steps. For today just one simple endpoint based on `mostMembersSince`. 

Need a lot more work but it's functional and more importantly RSS-compliant (AFAIK, new to the format myself)

Note I have added a new top level `/feeds` for the RSS endpoints. 

Plan is to publish this and then see if it works properly in feed reader, etc. And go from there. Feedback appreciated.

@andrewconner @eishay


kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
```diff
@@ -55,6 +63,51 @@ class LibraryController @Inject() (
...
+
+    val libIds = tops.map(_._1)
+    val libMap = db.readOnlyReplica { implicit request =>
+      libraryRepo.getLibraries(libIds.toSet)
+    }
+    val libraries = libIds.map(libMap(_))
+    val ownerIds = CollectionHelpers.dedupBy(libraries.map(_.ownerId))(id => id)
+    val ownerMap = db.readOnlyMaster { implicit request => userRepo.getUsers(ownerIds) }
+    val formatter = DateTimeFormat.forPattern("E, d MMM y HH:mm:ss Z") // rfc822
+    val prefix = if (mode == Mode.Dev) "http://dev.ezkeep.com:9000" else "https://www.kifi.com"
```
eishay: maybe use `FortyTwoConfig#applicationBaseUrl` ?

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
```diff
@@ -55,6 +63,51 @@ class LibraryController @Inject() (
...
+    val items = tops map {
+      case (libId, count) =>
+        val lib = libMap(libId)
+        val owner = ownerMap(lib.ownerId)
+        <item>
+          <title>{ lib.slug.value }</title>
```
eishay: why not `lib.name`?

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
raymondng: Cool. Didn't know -- I was going to add something like this.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
```diff
@@ -55,6 +63,51 @@ class LibraryController @Inject() (
...
+          <description>{ lib.description.getOrElse("") }</description>
```
eishay: See `LibraryCommander#libraryMetaTags`
How about using instead

``` scala
library.description.getOrElse(s"${owner.fullName}'s ${library.name} Kifi Library")
```

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
raymondng: that's a good idea. thx.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
```diff
@@ -55,6 +63,51 @@ class LibraryController @Inject() (
...
+        </channel>
+      </rss>
+
+    val elems = Enumerator.enumerate(rss)
+    val toBytes = Enumeratee.map[Node] { n => n.toString.getBytes }
+    val header = Enumerator("<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n".getBytes)
+    val combined = header.andThen(elems &> toBytes)
+    log.info(s"[getTopLibraries] published #${libIds.size} libraries. ${libIds.take(20)} ...")
+    Result(
+      header = ResponseHeader(200, Map(CONTENT_TYPE -> "application/xml")),
```
eishay: http://tools.ietf.org/id/draft-nottingham-rss-media-type-00.txt

``` application/rss+xml```
```

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
raymondng: Sounds good.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
raymondng: Good catch. Feed readers processed it without issues, thought I am compliant, but guess not.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
```diff
@@ -55,6 +63,51 @@ class LibraryController @Inject() (
...
+        val owner = ownerMap(lib.ownerId)
+        <item>
+          <title>{ lib.slug.value }</title>
+          <description>{ lib.description.getOrElse("") }</description>
+          <link>{ s"$prefix${Library.formatLibraryPath(owner.username, owner.externalId, lib.slug)}" }</link>
+          <pubDate>{ lib.createdAt.toString(formatter) }</pubDate>
+        </item>
+    }
+    val rss =
+      <rss>
```
eishay: not 

``` xml
<rss version="2.0">
```

http://en.wikipedia.org/wiki/RSS#Example
?

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
```diff
@@ -55,6 +63,51 @@ class LibraryController @Inject() (
...
+    val prefix = if (mode == Mode.Dev) "http://dev.ezkeep.com:9000" else "https://www.kifi.com"
+    val items = tops map {
+      case (libId, count) =>
+        val lib = libMap(libId)
+        val owner = ownerMap(lib.ownerId)
+        <item>
+          <title>{ lib.slug.value }</title>
+          <description>{ lib.description.getOrElse("") }</description>
+          <link>{ s"$prefix${Library.formatLibraryPath(owner.username, owner.externalId, lib.slug)}" }</link>
+          <pubDate>{ lib.createdAt.toString(formatter) }</pubDate>
```
eishay: GUID helps scrapers a lot. We can use the library external id. E.e. 

``` scala
<guid>{ lib.externalid.toString }</guid>
```

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
raymondng: I was following Google doc (which does not include the version) but I can add it.

eishay:
lets use the `author` tag http://cyber.law.harvard.edu/rss/rss.html#hrelementsOfLtitemgt with the `owner.fullName`

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
raymondng: Interesting. I actually looked into GUID a bit and thought it's not necessary -- I checked a few top sites ... they seem to just dup it with link or drop it entirely. I can add it though.

eishay: I used the GUID at linkedin when i built the scraper to dedup articles.
Think about renaming the library, updating the slug or moving the owner.
The GUID will tell the scraper its still the same web resource.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
```diff
@@ -55,6 +64,49 @@ class LibraryController @Inject() (
   implicit val config: PublicIdConfiguration)
     extends UserActions with LibraryAccessActions with ShoeboxServiceController {
 
+  def getTopLibraries() = MaybeUserAction { request =>
```
andrewconner, line 66: Why `MaybeUserAction`?

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
```diff
@@ -55,6 +64,49 @@ class LibraryController @Inject() (
...
+    val tops = db.readOnlyReplica { implicit request =>
```
andrewconner: Shadowing `request`

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
```diff
@@ -55,6 +64,49 @@ class LibraryController @Inject() (
...
+    }
+
+    val libIds = tops.map(_._1)
+    val libMap = db.readOnlyReplica { implicit request =>
+      libraryRepo.getLibraries(libIds.toSet)
+    }
+    val libraries = libIds.map(libMap(_))
+    val ownerIds = CollectionHelpers.dedupBy(libraries.map(_.ownerId))(id => id)
+    val ownerMap = db.readOnlyMaster { implicit request => userRepo.getUsers(ownerIds) }
+    val formatter = DateTimeFormat.forPattern("E, d MMM y HH:mm:ss Z") val items = tops map {
```
andrewconner: Two lines combined?

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
raymondng: Ah will make it clearer

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
raymondng: Weird. I am sure I didn't do this ... will fix.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
raymondng, line 66: Knowing that a user is visiting might be helpful, but it's not necessary. Are you concerned about overhead?

andrewconner, line 66: It's not being used.

raymondng, line 66: ah that will be fixed shortly :)

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
raymondng: Ah we have an in-house expert! You should have told me earlier so I can bug you instead :)
No problem -- will add GUID. BTW we have no external Id, only public Id -- is that what you had in mind?

eishay: That would work as well.

andrewconner:
Wahoo. Let's get this added to the marketing homepage.

eishay:
Sure, get it out. 
I would like to get some stats on it quickly e.g.:
- How frequently its pulled (maybe we should cache) 
- Adding tracking to the urls @stkem can help with how to do it

raymondng:
Alright. On it!

PR #12228, opened 2014-11-07
frontend library tracking  fixes

joshm1:
@stkem

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/signup/signupService.js
+++ b/kifi-backend/modules/shoebox/angular/src/signup/signupService.js
@@ -100,7 +100,7 @@ angular.module('kifi')
       }
 
       if ($scope.userData.libraryId) {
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/angular/src/signup/signupService.js
```diff
@@ -100,7 +100,7 @@ angular.module('kifi')
       }
 
       if ($scope.userData.libraryId) {
-        $rootScope.$broadcast('trackLibraryEvent', 'click', { type: 'signupLibrary', action: 'clickSignUpButton' });
+        $scope.$emit('trackLibraryEvent', 'click', { type: 'signupLibrary', action: 'clickSignUpButton' });
```
stkem, line 103: If this is in response to my previous comment, I actually meant to use `$rootScope.emit` and `$rootScope.on`, that way there is no event propagation at all, it just stays in the root scope.

PR #12232, opened 2014-11-07
fix admin js parsing

yingjieMiao:
@stkem  FYI

since when json became so type-safe?


yingjieMiao:
I guess Json was looking for quotes....

PR #12241, opened 2014-11-07
rss: wire-in meta tags

raymondng:
Wire-in `libraryMetaTags` 
Add image for channel

@eishay @andrewconner


raymondng:
Went over changes with @eishay in person. Sticking with rss for now -- I have a different branch with atom, if needed. Will push to see if it works as expected.

PR #12242, opened 2014-11-07
modify library needs to check if modified name/slug already exists

ahsu1230:
Surprising how we didn't catch this before =(

And some additional sanity checkings to addLibrary
@andrewconner 

Thanks for the heads up @yipingliao


andrewconner:
Good catch.

PR #12243, opened 2014-11-07
Getting rid of HTML tag

andrewconner:
@eishay


kifi-backend/modules/shoebox/app/com/keepit/commanders/PublicPageMetaTags.scala
```diff
@@ -73,7 +73,6 @@ case class PublicPageMetaFullTags(unsafeTitle: String, url: String, urlPathOnly:
     } getOrElse ""
 
     s"""
-      |<html itemscope itemtype="http://schema.org/Product">
```
2is10, line 77: no matching `</html>` to remove?

andrewconner, line 77: That's why this was removed. There was none.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -450,8 +450,8 @@ class LibraryCommander @Inject() (
               db.readOnlyMaster { implicit s =>
                 libraryRepo.getBySlugAndUserId(userId, slug)
               } match {
-                case Some(_) => Left(LibraryFail(BAD_REQUEST, "library_slug_exists"))
-                case None => Right(slug)
+                case Some(other) if other.id.get != libraryId => Left(LibraryFail(BAD_REQUEST, "library_slug_exists"))
+                case _ => Right(slug)
```
ahsu1230, line 454: I see. I assumed wrongly that `name` would be different from the existing name or `None` otherwise. Sorry about this -bad mistake on my part. Thanks for the change

kifi-backend/modules/shoebox/angular/src/common/platformService.js
```diff
@@ -16,6 +16,7 @@ angular.module('kifi')
     };
 
     var goToAppOrStore = function (url) {
+      url = url || '//kifi.com';
```
2is10, line 19: It looks like this is going to produce `'intent:////kifi.com#Intent...'` on Android. Intended?

PR #12245, opened 2014-11-07
adding last modification tag

eishay:
@raymondng @andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/common/seo/SiteMapGenerator.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/common/seo/SiteMapGenerator.scala
@@ -3,13 +3,16 @@ package com.keepit.common.seo
 import com.google.inject.{ Singleton, Inject }
 import com.keepit.commanders.LibraryCommander
 import com.keepit.common.CollectionHelpers
+import com.keepit.common.time._
 import com.keepit.common.actor.ActorInstance
 import com.keepit.common.akka.FortyTwoActor
 import com.keepit.common.db.slick.Database
 import com.keepit.common.healthcheck.AirbrakeNotifier
 import com.keepit.common.logging.Logging
 import com.keepit.common.plugin.{ SchedulerPlugin, SchedulingProperties }
+import com.keepit.inject.FortyTwoConfig
 import com.keepit.model.{ KeepRepo, UserRepo, Library, LibraryRepo }
+import org.joda.time.DateTime
 import play.api.{ Play, Plugin }
 import Play.current
 import scala.concurrent.Future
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/app/com/keepit/model/KeepRepo.scala
```diff
@@ -436,4 +437,9 @@ class KeepRepoImpl @Inject() (
     import StaticQuery.interpolation
     sql"""select b.library_id, count(*) as cnt from bookmark b, library l where l.id = b.library_id and l.state='active' and l.visibility='published' and b.created_at > $since group by b.library_id order by count(*) desc limit $count""".as[(Id[Library], Int)].list
   }
+
+  def latestKeepInLibrary(libraryId: Id[Library])(implicit session: RSession): Option[DateTime] = {
```
raymondng, line 441: :+1: Was thinking the same. We should probably cache something like this as well.

kifi-backend/modules/shoebox/app/com/keepit/common/seo/SiteMapGenerator.scala
```diff
@@ -73,27 +77,39 @@ class SiteMapGenerator @Inject() (
...
-    val paths = libraries.filter { lib =>
+    val libs = libraries.filter { lib =>
       owners.get(lib.ownerId).isDefined && db.readOnlyMaster { implicit ro => keepRepo.getCountByLibrary(lib.id.get) > 3 } // proxy for quality; need bulk version
-    } map { lib =>
+    }
+
+    def path(lib: Library): String = {
       val owner = owners(lib.ownerId)
-      s"https://www.kifi.com${Library.formatLibraryPath(owner.username, owner.externalId, lib.slug)}"
+      s"${fortyTwoConfig.applicationBaseUrl}{Library.formatLibraryPath(owner.username, owner.externalId, lib.slug)}"
```
raymondng, line 86: Thanks!

kifi-backend/modules/shoebox/app/com/keepit/model/KeepRepo.scala
raymondng, line 441: Probably should name it something like `lastUpdatedAt` for libraries.

kifi-backend/modules/shoebox/app/com/keepit/common/seo/SiteMapGenerator.scala
```diff
@@ -73,27 +77,39 @@ class SiteMapGenerator @Inject() (
...
+    }
+
+    def lastMod(lib: Library): DateTime = {
+      db.readOnlyReplica { implicit s =>
+        keepRepo.latestKeepInLibrary(lib.id.get) match {
+          case Some(date) if date.isAfter(lib.updatedAt) => date
+          case _ => lib.updatedAt
+        }
+      }
```
2is10, line 95: This won’t update the last-modified date if someone removes a keep from a library, which could sometimes be important. Probably worth explicitly storing a `last_modified_at` field in each library—if not today, then eventually. We’ve been bitten by basing logic on `createdAt` and `updatedAt` in the past.

eishay, line 95: agree

PR #12247, opened 2014-11-07
Always have inviters first in followers list

stkem:
Order is arbitrary if there are multiple inviters
@andrewconner @ahsu1230

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
@@ -193,7 +193,16 @@ class LibraryCommander @Inject() (
 
     val followerInfosByLibraryId = libraries.map { library =>
       val (collaborators, followers, _, counts) = getLibraryMembers(library.id.get, 0, maxMembersShown, fillInWithInvites = false)
-      library.id.get -> ((collaborators ++ followers).map(_.userId), counts)
+      val inviters: Seq[LibraryMembership] = viewerUserIdOpt.map { userId =>
+        db.readOnlyReplica { implicit session =>
+          libraryInviteRepo.getWithLibraryIdAndUserId(library.id.get, userId).filter { invite =>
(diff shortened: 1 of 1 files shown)


andrewconner:
Looks good, thanks!

ahsu1230:
awesome! tyty

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -193,7 +193,16 @@ class LibraryCommander @Inject() (
 
     val followerInfosByLibraryId = libraries.map { library =>
       val (collaborators, followers, _, counts) = getLibraryMembers(library.id.get, 0, maxMembersShown, fillInWithInvites = false)
-      library.id.get -> ((collaborators ++ followers).map(_.userId), counts)
+      val inviters: Seq[LibraryMembership] = viewerUserIdOpt.map { userId =>
+        db.readOnlyReplica { implicit session =>
+          libraryInviteRepo.getWithLibraryIdAndUserId(library.id.get, userId).filter { invite =>
```
ahsu1230, line 198: if you prefer, you could use `filterNot` in this case

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -193,7 +193,16 @@ class LibraryCommander @Inject() (
...
+            invite.inviterId != library.ownerId
+          }.map { invite =>
+            libraryMembershipRepo.getWithLibraryIdAndUserId(library.id.get, invite.inviterId)
+          }
+        }.flatten
+      }.getOrElse(Seq.empty)
+      library.id.get -> ((inviters ++ collaborators.filter(!inviters.contains(_)) ++ followers.filter(!inviters.contains(_))).map(_.userId), counts)
```
ahsu1230, line 205: same here if you think it's more readable =)

PR #12250, opened 2014-11-07
hamburger thingy shows left bar by default. Press hamburger to see right...

ahsu1230:
... column.

Basically just flipped from the original function

@andrewconner @yipingliao @stkem @2is10

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/layout/header/HeaderCtrl.js
+++ b/kifi-backend/modules/shoebox/angular/src/layout/header/HeaderCtrl.js
@@ -16,7 +16,7 @@ angular.module('kifi')
     $window.addEventListener('message', function (event) {
       if (event.data === 'show_left_column') {  // for guide
         $scope.$apply(function () {
-          $rootElement.find('html').addClass('kf-sidebar-inactive');
+          $rootElement.find('html').removeClass('kf-sidebar-inactive');
         });
       }
     });
--- a/kifi-backend/modules/shoebox/angular/src/layout/header/HeaderCtrl.js
+++ b/kifi-backend/modules/shoebox/angular/src/layout/header/HeaderCtrl.js
@@ -12,6 +12,7 @@ angular.module('kifi')
       $rootElement.find('html').toggleClass('kf-sidebar-inactive');
     };
 
+    // reminder: this triggers left sidebar to show in case the default for small windows is to hide left sidebar
     $window.addEventListener('message', function (event) {
       if (event.data === 'show_left_column') {  // for guide
         $scope.$apply(function () {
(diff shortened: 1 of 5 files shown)


andrewconner:
Product knows about it? May want to double check. I like the idea.

ahsu1230:
Yup, will double check.

yipingliao:
Awesome!

kifi-backend/modules/shoebox/angular/src/layout/sidebar/sidebar.styl
```diff
@@ -17,14 +17,14 @@
 
 @media only screen and (max-width: 1024px)
   .kf-sidebar
-    left: -135px
-    z-index: -1
+    left: 0
+    z-index: 1
```
2is10, line 21: Why does the z-index need to be lowered from `10` to `1` on this screen size?

kifi-backend/modules/shoebox/angular/src/layout/sidebar/sidebar.styl
```diff
@@ -17,14 +17,14 @@
 
 @media only screen and (max-width: 1024px)
   .kf-sidebar
-    left: -135px
-    z-index: -1
+    left: 0
```
2is10, line 20: What does this `left: 0` accomplish? If nothing, let’s omit it.
Oh, is it required to make the transition work?

ahsu1230, line 20: it makes the transition much smoother. Without it, it gets pretty choppy

kifi-backend/modules/shoebox/angular/src/layout/header/HeaderCtrl.js
```diff
@@ -9,13 +9,14 @@ angular.module('kifi')
...
-      $rootElement.find('html').toggleClass('kf-sidebar-active');
+      $rootElement.find('html').toggleClass('kf-sidebar-inactive');
     };
 
+    // reminder: this triggers left sidebar to show in case the default for small windows is to hide left sidebar
     $window.addEventListener('message', function (event) {
       if (event.data === 'show_left_column') {  // for guide
         $scope.$apply(function () {
-          $rootElement.find('html').addClass('kf-sidebar-active');
+          $rootElement.find('html').addClass('kf-sidebar-inactive');
```
2is10: Scratch my earlier comment. Shouldn’t this be `.removeClass` now?

2is10:
nice

kifi-backend/modules/shoebox/angular/src/layout/header/HeaderCtrl.js
ahsu1230: if we removed the class, wouldn't we have to `.addClass` later? 
Oh wait. nvm.

~ edit 1 ~ if we `.removeClass` then would we still be able to `.toggleClass` afterwards?

ahsu1230:
@weeewake says it's a go!

PR #12251, opened 2014-11-07
getWithLibraryIdsAndUserId: bulk version of getWithLibraryIdAndUserId

ymatsuda:
@eishay @andrewconner @ahsu1230 

Getting a bulk of LibraryMembership instances in one db call.


andrewconner:
Awesome!

PR #12253, opened 2014-11-07
Add generic error modal pop-ups for library and keep actions.

yipingliao:
@andrewconner @ahsu1230 
https://app.asana.com/0/17575799252940/19685910579384

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/keep/keep.js
+++ b/kifi-backend/modules/shoebox/angular/src/keep/keep.js
@@ -441,11 +441,7 @@ angular.module('kifi')
 
               libraryService.addToLibraryCount(clickedLibrary.id, -1);
               scope.$emit('keepRemoved', { url: scope.keep.url }, clickedLibrary);
-            })['catch'](function () {
-              modalService.open({
-                template: 'common/modal/genericErrorModal.tpl.html'
-              });
-            });
+            })['catch'](modalService.openGenericErrorModal);
 
           // Keep.
           } else {
@@ -466,12 +462,7 @@ angular.module('kifi')
             if (scope.keep && scope.keep.id) {
               keepToLibrary = keepActionService.copyToLibrary([scope.keep.id], clickedLibrary.id).then(function (result) {
                 if (result.successes.length > 0) {
-                  return keepActionService.fetchFullKeepInfo(scope.keep)
-                    .then(fetchKeepInfoCallback)['catch'](function () {
-                      modalService.open({
-                        template: 'common/modal/genericErrorModal.tpl.html'
-                      });
-                    });
+                  return keepActionService.fetchFullKeepInfo(scope.keep).then(fetchKeepInfoCallback);
                 }
               });
             } else {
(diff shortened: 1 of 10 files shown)


kifi-backend/modules/shoebox/angular/src/keep/keep.js
```diff
@@ -129,22 +129,6 @@ angular.module('kifi')
           keep.descHtml = formatDesc(keep.siteName || keep.url);
         }
 
-        function keepOne (keep) {
```
yipingliao, line 132: I cleaned up some unused code in this file.

ahsu1230, line 132: :+1:

kifi-backend/modules/shoebox/angular/src/keeps/keepActionService.js
```diff
@@ -214,17 +214,6 @@ angular.module('kifi')
       });
     }
 
-    function togglePrivateMany(keeps) {
```
yipingliao, line 217: More clean-up: no more toggling public and private keeps.

kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.js
```diff
@@ -81,21 +81,18 @@ angular.module('kifi')
 
             var error = err.data && err.data.error;
             switch (error) {
-              case 'library name already exists for user':  // deprecated
-              case 'library_name_exists':
-                scope.$error.general = 'You already have a library with this name. Pick another.';
-                break;
-              case 'invalid library name':  // deprecated
               case 'invalid_name':
+                scope.$error.general = 'The name you picked isn\'t valid. Try using only letters and numbers.';
```
yipingliao, line 85: @ahsu1230 I updated these so that they are current.

yipingliao:
@andrewconner Aaron and I are debating whether this pattern of catching errors in the controllers and directives is good or whether it pollutes that code too much. One idea is to catch errors in the action services, which will DRY up things and have a better guarantee that the error path is considered but it has the drawback of less flexibility in UI treatment, and I don't like opening modals in a service. Thoughts?

kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.js
ahsu1230, line 85: nice! :ice_cream:

yipingliao:
Updated per your very helpful code review comments, @andrewconner and @ahsu1230

kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.js
```diff
@@ -81,21 +81,18 @@ angular.module('kifi')
...
+                break;
+              case 'invalid_slug':
+                scope.$error.general = 'The URL you picked isn\'t valid. Try using only letters and numbers.';
```
2is10, line 88: One benefit of smart quotes is that they never have to be escaped!  :smiley: 
`“` ← `Opt` `[`
`”` ← `Shift` `Opt` `[`
`‘` ← `Opt` `]`
`’` ← `Shift` `Opt` `]`

yipingliao, line 88: Oh! Thanks!

PR #12254, opened 2014-11-07
caching sitemap and serving it dynamically with last modification time

eishay:


eishay:
@raymondng @andrewconner

raymondng:
Cool. Thanks.

Concerned about the memory impact on memcached and shoebox as the file size grows. We should probably persist the file and just serve from file system direct going forward. WDYT?

PR #12256, opened 2014-11-08
humble beginnings of a proper keptAt field for keeps

stkem:
@andrewconner @leogrim 
the do not merge is mostly because it's Friday afternoon and this touches the keep object and table


stkem:
The keep object as well as the raw keep object gained a separate optional timestamp and the bookmark imports sets it when available. Next steps:
1) Backfill DB and make it not optional for keeps (it will remain optional for raw_keeps) 
2) Start using it everywhere appropriate instead of created at (that _might_ need extra indices on the DB)

leogrim:
:+1:

ymatsuda:
Keep me in the loop. This affects the search backend, right?

leogrim:
We need to switch the field we use in the `KeepIndexer` from `createdAt` to `keptAt`, I can't think of anything else?

andrewconner:
Awesome, thanks for this.

kifi-backend/modules/common/app/com/keepit/model/Keep.scala
```diff
@@ -84,11 +85,11 @@ object Keep {
...
-    Keep(id, createdAt, updatedAt, externalId, title, uriId, isPrimary.exists(b => b), inDisjointLib.exists(b => b), urlId, url, bookmarkPath, visibility.getOrElse(isPrivateToVisibility(isPrivate)), userId, state, source, kifiInstallation, seq, libraryId)
+  def applyFromDbRow(id: Option[Id[Keep]], createdAt: DateTime, updatedAt: DateTime, externalId: ExternalId[Keep], title: Option[String], uriId: Id[NormalizedURI], isPrimary: Option[Boolean], inDisjointLib: Option[Boolean], urlId: Id[URL], url: String, bookmarkPath: Option[String], isPrivate: Boolean, userId: Id[User], state: State[Keep], source: KeepSource, kifiInstallation: Option[ExternalId[KifiInstallation]], seq: SequenceNumber[Keep], libraryId: Option[Id[Library]], visibility: Option[LibraryVisibility], keptAt: Option[DateTime]) = {
+    Keep(id, createdAt, updatedAt, externalId, title, uriId, isPrimary.exists(b => b), inDisjointLib.exists(b => b), urlId, url, bookmarkPath, visibility.getOrElse(isPrivateToVisibility(isPrivate)), userId, state, source, kifiInstallation, seq, libraryId, keptAt)
```
andrewconner, line 89: how about `Some(keptAt.getOrElse(createdAt))`?

PR #12259, opened 2014-11-08
short-circuiting

ymatsuda:
@eishay @andrewconner @ahsu1230 
Short-circuiting getWithLibraryIdsAndUserId when there is no library ids.

Just found that Slick still issue SQL like this when there is no item for IN predicate.

select ...
from `library_membership` x2 where (false and (x2.`user_id` = 9827)) and (not (x2.`state` = 'inactive'))

`false` in the WHERE clause is the remnant of the IN predicate with no items.


andrewconner:
Nice catch

eishay:
nice

PR #12260, opened 2014-11-08
auto submitting sitemap to google and bing

eishay:
@andrewconner @raymondng

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/common/seo/SiteMapGenerator.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/common/seo/SiteMapGenerator.scala
@@ -5,6 +5,7 @@ import com.google.inject.{ Singleton, Inject }
 import com.keepit.commanders.LibraryCommander
 import com.keepit.common.CollectionHelpers
 import com.keepit.common.cache._
+import com.keepit.common.net.{ DirectUrl, HttpClient }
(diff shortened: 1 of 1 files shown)


andrewconner:
Awesome

kifi-backend/modules/shoebox/app/com/keepit/common/seo/SiteMapGenerator.scala
```diff
@@ -33,30 +34,30 @@ class SiteMapGeneratorPluginImpl @Inject() (
   override def enabled: Boolean = true
   override def onStart() {
     for (app <- Play.maybeApplication) {
-      val (initDelay, freq) = if (Play.isDev) (30 seconds, 10 minutes) else (15 minutes, 60 minutes)
+      val (initDelay, freq) = (15 minutes, 60 minutes)
```
andrewconner: Mind setting the `15 minutes` to higher, like `60 minutes`? As-is, server restarts within an hour will call it many times.

raymondng:
Nice :+1:

kifi-backend/modules/shoebox/app/com/keepit/common/seo/SiteMapGenerator.scala
eishay: sure

PR #12263, opened 2014-11-08
adding the robots!

eishay:


eishay:
@andrewconner @raymondng

andrewconner:
Yeee ta ta ta

PR #12265, opened 2014-11-08
Revert "hamburger thingy shows left bar by default. Press hamburger to s...

andrewconner:
This is breaking public library pages when the browser isn't large.

Feel free to fix and re-merge.


andrewconner:
@ahsu1230 @yipingliao

kifi-backend/modules/shoebox/angular/src/layout/scrollablePane/scrollablePane.styl
```diff
@@ -7,9 +7,9 @@
 @media only screen and (max-width: 1024px)
   .kf-scrollable-pane.kf-scrollable-pane
     width: 100%
-    padding-left: 270px
```
andrewconner, line 10: I believe this was the issue, but we need to test this on different devices and resolutions. @ahsu1230

ahsu1230:
Sorry my bad. Thanks for the revert

PR #12267, opened 2014-11-08
fix url generation, circular dependency, and added test for sitemap gene...

eishay:
...ration
@raymondng @andrewconner


andrewconner:
Nice

PR #12270, opened 2014-11-09
feed: misc

raymondng:
Misc tweaks.

Note that I have left the image size alone after viewing the feed via feedly -- the big image actually looks good (on mobile). Will continue to investigate.

@eishay @andrewconner


eishay:
great, on email the image is too large, I think it would be a good idea to keep it at some max size. Maybe add a width element which is `min(image.width, 400px)` or something like that. This way the browser will do the resizing. Not perfect but a good first step. What do you think?

raymondng:
Ah. Good idea.

raymondng:
BTW are you using some RSS + email tool? Which one?

eishay:
IFTTT recipe
https://ifttt.com/recipes/219315-email-on-trending-kifi-libraries

raymondng:
Ah. Of course. Let me push this first. Will continue to investigate images -- right now the size info is not part of the meta tag. Will need a separate call.

raymondng:
Correction: will push Monday AM.

PR #12272, opened 2014-11-10
easing down load on the scraper

eishay:


eishay:
@raymondng @andrewconner @ymatsuda @leogrim @stkem

kifi-backend/modules/common/conf/prod/prod.conf
```diff
@@ -161,15 +161,15 @@ ws {
...
+    min = 168                       # hours -> 7 days     1 week
+    max = 1024                      # hours -> 42.6 days, 6 weeks
+    increment = 12                  # hours
     decrement = 2                   # hours
   }
 
   initialBackoff = 3                 # hours
   maxBackoff = 1024                  # hours
-  maxRandomDelay = 600               # seconds
+  maxRandomDelay = 21600             # seconds -> 6 hours (spreading out large import batches)
```
ymatsuda, line 172: So, in the worst case, a content of newly kept page won't be scraped for 6 hours, and the content remain not searchable. Please let the product team know this so that they understand this is not a search bug.

eishay, line 172: Oh, I didn't know the delay applies to the initial scrape, I thought its only applying to the next scrapes and the initial `nextScrape` is the `START_OF_TIME` (i.e. asap). 
I'll revert this change until its clarified.

ymatsuda, line 172: Oh. Sorry. You may be right.

ymatsuda, line 172: I think you are right. We seem to apply random delay only to next scrape.

PR #12273, opened 2014-11-10
making the random spread be only 10 minutes

eishay:
@ymatsuda @raymondng


eishay:
its basically a revert to the prev config

PR #12277, opened 2014-11-10
Trying again - switch hamburger functionality (show libraries by default)

ahsu1230:
Last merge broke public pages (too much padding on the left side when the window was small)
Added a line that would have no padding for public pages.


kifi-backend/modules/shoebox/angular/src/layout/scrollablePane/scrollablePane.styl
```diff
@@ -7,9 +7,12 @@
...
     -webkit-transition: padding-left sidebarTransitionDuration;
     transition: padding-left sidebarTransitionDuration;
 
-  .kf-sidebar-active .kf-scrollable-pane
-      padding-left: 270px
+  .kf-sidebar-inactive .kf-scrollable-pane
+      padding-left: 0px
+
+.kf-logged-out .kf-scrollable-pane
+  padding-left: 0px
```
ahsu1230, line 18: fix on public pages

PR #12282, opened 2014-11-10
noop when unfriending no one or adding no one

ymatsuda:
@eishay @andrewconner 

These two very expensive methods, unfriendConnections and addConnections, should be noop when `users` is empty. It looks obvious too me, but please doublecheck that there is no hidden side effect. I will wait for your approval.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/model/UserConnectionRepo.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/model/UserConnectionRepo.scala
@@ -122,46 +122,52 @@ class UserConnectionRepoImpl @Inject() (
   }
 
   def unfriendConnections(userId: Id[User], users: Set[Id[User]])(implicit session: RWSession): Int = {
-    val ids = (for {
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/app/com/keepit/model/UserConnectionRepo.scala
```diff
@@ -122,46 +122,52 @@ class UserConnectionRepoImpl @Inject() (
   }
 
   def unfriendConnections(userId: Id[User], users: Set[Id[User]])(implicit session: RWSession): Int = {
-    val ids = (for {
-      c <- rows if c.user2 === userId && c.user1.inSet(users) || c.user1 === userId && c.user2.inSet(users)
-    } yield c.id).list
+    if (users.nonEmpty) {
```
eishay, line 125: sure :-) thought I think that if users is empty it may indicate a bug, right?

eishay:
lgtm, though I would feel better if @leogrim will take a look as well

leogrim:
lgtm

PR #12283, opened 2014-11-10
fix scrolling keeps over nav bar

ahsu1230:
Currently you can do this, if the window is small (hamburger is up) and you scroll to the right, the nav bar seems like it just hovers over the keeps

left-col-scroll-left

Have a change to keep content behind the nav bar hidden while still keeping things aligned.

@andrewconner @joshm1 @yipingliao


yipingliao:
Thanks for fixing this!

PR #12284, opened 2014-11-10
making a couple of db calls lazy

ymatsuda:
@leogrim @eishay 

When `userIds` is empty, `invitesByUserId` will not be used. Similarly when `emailAddresses` is empty, `invitesByEmailAddress` will not be used. By making them lazy, we can avoid unnecessary db calls.


eishay:
:+1:

PR #12295, opened 2014-11-10
propagating chunk measurements to mixpanel

2is10:
@leogrim

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/heimdal/SearchAnalytics.scala
+++ b/kifi-backend/modules/common/app/com/keepit/heimdal/SearchAnalytics.scala
@@ -45,7 +45,9 @@ case class BasicSearchContext(
   kifiShownTime: Option[Int],
   thirdPartyShownTime: Option[Int],
   kifiResultsClicked: Option[Int],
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/common/app/com/keepit/heimdal/SearchAnalytics.scala
```diff
@@ -45,7 +45,9 @@ case class BasicSearchContext(
   kifiShownTime: Option[Int],
   thirdPartyShownTime: Option[Int],
   kifiResultsClicked: Option[Int],
-  thirdPartyResultsClicked: Option[Int])
+  thirdPartyResultsClicked: Option[Int],
+  chunkDelta: Option[Int],
+  chunksSplit: Option[Boolean])
```
2is10, line 50: `chunkDelta` is the number of milliseconds between the arrival of the first and second chunk of our results
`chunksSplit` is whether we think the rendering of the Google results happened between the arrival times of our two chunks — `true` essentially meaning that it could have mattered if we would have tried to render our results incrementally

PR #12296, opened 2014-11-10
[DO NOT MERGE] Library search refactor in preparation for public library search

yipingliao:
@ahsu1230 @andrewconner 
https://app.asana.com/0/18079737520750/18807532114522

This pull refactors the current library search to use a different url. Instead of `https://www.kifi.com/find?q=car&f=a&l=l9Rre2MSpAPb`, we now have `https://www.kifi.com/lydialaurenson/halloween/find?q=car&f=a`.

Please do not merge yet; I still need to work on the permissions part.

@ahsu1230 Your review is much appreciated, since I changed a lot of code in `HeaderCtrl`.

Thanks!

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/search/search.js
+++ b/kifi-backend/modules/shoebox/angular/src/search/search.js
@@ -63,7 +63,6 @@ angular.module('kifi')
       libraryIdPromise.then(function (libraryId) {
         library = libraryId;
 
-
         if (!query) { // No query or blank query.
           $location.path('/');
         }
@@ -255,6 +254,11 @@ angular.module('kifi')
       });
     });
     $scope.$on('$destroy', deregisterKeepAddedListener);
+
+    var deregisterUpdateLibrarySearch = $rootScope.$on('librarySearched', function () {
+      init();
+    });
+    $scope.$on('$destroy', deregisterUpdateLibrarySearch);
 
 
     init();
--- a/kifi-backend/modules/shoebox/angular/src/search/search.js
+++ b/kifi-backend/modules/shoebox/angular/src/search/search.js
@@ -10,6 +10,8 @@ angular.module('kifi')
     //
     var query;
     var filter;
+    var userName;
+    var librarySlug;
     var library;
     var lastResult = null;
     var selectedCount = 0;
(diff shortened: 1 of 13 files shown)


kifi-backend/modules/shoebox/angular/src/search/search.js
```diff
@@ -67,6 +41,46 @@ angular.module('kifi')
...
+        $scope.hasMore = true;
+        $scope.scrollDistance = '100%';
+        $scope.loading = false;
+
+        $window.document.title = 'Kifi • ' + query;
+
+        searchActionService.reset();
+        $scope.getNextKeeps(true);
+      })['catch'](function (resp) {
+        if (resp.status && resp.status === 403) {
```
yipingliao, line 81: What should we do when a user tries an url that searches a library that the user has no access to? We could redirect to login, show the not-found page, or show the need-to-login page (that prompts for the passphrase).

kifi-backend/modules/shoebox/angular/src/search/search.js
```diff
@@ -3,18 +3,22 @@
...
     //
     var query;
-    var filter = $routeParams.f || 'm';
-    var library = $routeParams.l || '';
-
+    var filter;
+    var library;
     var lastResult = null;
     var selectedCount = 0;
+    var authToken = $location.search().authToken || $location.search().authCode || $location.search().accessToken || '';
```
ahsu1230: We should only be using use `authToken` now. The other `authCode` or `accessToken` were when we were still confused / deciding the name lol

kifi-backend/modules/shoebox/angular/src/search/search.js
andrewconner, line 81: Not found would be closest in line to what we do currently.

kifi-backend/modules/shoebox/angular/src/search/search.js
yipingliao: Will do. Thanks!

kifi-backend/modules/shoebox/angular/src/search/search.js
yipingliao, line 81: Okay, will do.

PR #12301, opened 2014-11-11
new libraries feed

raymondng:
Factored out common logic to `FeedCommander`
Moved endpoint to `FeedController`
DRY-er implementation in general
Simple test added for new feed
More work to do (kcid, auto-ping google pubsubhub, etc.)

@eishay @andrewconner


andrewconner:
Great thanks

eishay:
Could not find in the code, are we having some quality measures like not showing libraries with less then three keeps?

raymondng:
No not yet. For the new feed, I thought this is not necessary -- but it'd be trivial to add. Let me know.

raymondng:
As discussed, we'll do filtering on the `top` feed.

raymondng:
Merging to build on it. Thanks for the feedback!

PR #12308, opened 2014-11-11
getting images of multiple keeps in one db call

ymatsuda:
@eishay @andrewconner @leogrim


eishay:
thanks!

andrewconner:
Awesome, thanks Yasuhiro.

PR #12311, opened 2014-11-11
using new `keptAt` field on keeps in recommendations

stkem:
@yingjieMiao

### DIFF
--- a/kifi-backend/modules/curator/app/com/keepit/curator/commanders/AllKeepsSeedIngestionHelper.scala
+++ b/kifi-backend/modules/curator/app/com/keepit/curator/commanders/AllKeepsSeedIngestionHelper.scala
@@ -103,9 +103,9 @@ class AllKeepSeedIngestionHelper @Inject() (
         uriId = keep.uriId,
         url = keep.url,
         userId = None,
-        firstKept = keep.createdAt,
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/curator/app/com/keepit/curator/commanders/AllKeepsSeedIngestionHelper.scala
```diff
@@ -103,9 +103,9 @@ class AllKeepSeedIngestionHelper @Inject() (
         uriId = keep.uriId,
         url = keep.url,
         userId = None,
-        firstKept = keep.createdAt,
-        lastKept = keep.createdAt,
-        lastSeen = keep.createdAt,
+        firstKept = keep.keptAt.getOrElse(keep.createdAt),
+        lastKept = keep.keptAt.getOrElse(keep.createdAt),
+        lastSeen = keep.keptAt.getOrElse(keep.createdAt),
         priorScore = None,
```
yingjieMiao, line 109: Mind adding this as a common API in `Keep`, or `CuratorKeep`? Looks like this can be used in many places, and we could avoid this boilerplate code.

yingjieMiao:
lgtm

PR #12312, opened 2014-11-11
adding .sbtopts for our project

2is10:
As described in [this post](https://medium.com/@jan______/sbtconfig-is-deprecated-650d6ff10236).

Also relevant is this message from [the homebrew sbt formula](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/sbt.rb):

> This formula is now using the standard typesafe sbt launcher script.
> Project specific options should be placed in .sbtopts in the root of your project.
> Global settings should be placed in /etc/sbtopts

Feel free to tweak the values. These are just what I’ve been using recently. The `PermGen` options are omitted because they’re [obsolete in Java 8](http://www.infoq.com/articles/Java-PERMGEN-Removed). I haven’t needed any Metaspace options.

This file will not get deployed to production, correct?

@andrewconner @stkem @ymatsuda @eishay

### DIFF
--- a/kifi-backend/.sbtopts
+++ b/kifi-backend/.sbtopts
@@ -1,0 +1,9 @@
+-J-Xms1g
+-J-Xmx4g
+-J-Xss4M
+-J-XX:+CMSClassUnloadingEnabled
+-J-XX:ReservedCodeCacheSize=2g
+-J-XX:+HeapDumpOnOutOfMemoryError
+-J-XX:+DoEscapeAnalysis
+-J-XX:+UseCodeCacheFlushing
+-J-XX:+UseConcMarkSweepGC


stkem:
Interesting, they quadruple the default stack size.

andrewconner:
Awesome. I actually have used one of these for a while. I'll go through my system backup and find it to compare.

2is10:
Note: The options and values in this .sbtopts file are somewhat arbitrary—they’ve been passed around the company, made it to me, and I may have changed some of them. Let’s tweak this file over time and use whatever options and values we feel make sense most of the time for our local development.

andrewconner:
As a note, we should _really_ unify `sbt` starter scripts. There's 3 versions out in the wild, and it causes some incompatibilities.

"SBT" refers to 3 things:
- `sbt` library, which builds projects, manages dependencies, etc. This is set by the project and we're all on the same version.
- `sbt-launcher.jar` which provides a command line for the library. This does not need to be upgraded on every release. As such, people are likely running different versions.
- `sbt`, the bash script. This reads config files (like the one committed) and runs `sbt-launcher.jar`. There are 3 versions out there.
  - "the standard script" which is only a few lines. This is what brew uses.
  - [Paul Phillips's `sbt-extras`](https://github.com/sbt/sbt-launcher-package), which is the most updated and feature-rich.
  - Typesafe's fork of `sbt-extras`, which has been maintained, but separately. Its documentation is not as up-to-date.

I use / prefer Paul Phillips's version, but am open to discussion. We may want to just include one of them in our project, so everyone can standardize.

2is10:
I’ve been using "the standard script" because `brew install sbt` seems to be the simplest way to get rolling with sbt and it’s what our [New Engineer guide](https://team42.atlassian.net/wiki/display/ENG/New+Engineer%3A+Get+Started) says to use. I haven’t needed any “extras,” but maybe I just don’t know what I’m missing.

PR #12316, opened 2014-11-11
3.3.16: hiding keeper faster after keep box/compose

2is10:


packrat/styles/google_inject.css
```diff
@@ -84,7 +84,7 @@
 }
 .kifi-res-bar:not(:hover)>.kifi-res-bar-status[href][data-n='0'][data-of]::after {
   opacity: 0;
-  transition-delay: 0;
+  transition-delay: 0s;
```
2is10, line 87: Firefox requires units for a zero duration, Chrome doesn’t.

PR #12320, opened 2014-11-12
Fixing no title on new bookmark

andrewconner:
@raymondng @stkem


stkem:
lgtm

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepInterner.scala
```diff
@@ -209,7 +211,7 @@ class KeepInterner @Inject() (
         }
 
         val savedKeep = bookmark.copy(
-          title = title orElse bookmark.title orElse uri.title,
+          title = trimmedTitle orElse bookmark.title orElse uri.title,
```
raymondng, line 214: Not sure how pervasive the issue is -- do we need to do the same check for `bookmark.title` and `uri.title`?

andrewconner, line 214: Not an issue I've noticed, just did a little cleanup.

raymondng, line 214: Oh cool

PR #12326, opened 2014-11-12
if the library is of a low quality (will iterate on that definition), te...

eishay:
...ll the search engine crawler to skip it
@andrewconner


eishay:
@andrewconner

andrewconner:
Awesome.

PR #12327, opened 2014-11-12
Cleanup in preparation for image changes.

andrewconner:
`uriId` isn't used yet, but we need to swap out the serialization on all services that use this.

@stkem @leogrim


leogrim:
:+1:

stkem:
lgtm

PR #12330, opened 2014-11-12
Removing url, option of uriId from UriSummaryRequest

andrewconner:
@leogrim @stkem 

I love statically typed languages.


stkem:
lgtm

PR #12334, opened 2014-11-12
optimizing getConnectionsPage

ymatsuda:
@eishay @andrewconner 
Just listing friends on Kifi should be simple and straight forward. But the original code wouldn't scale when a user had many connections.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/UserConnectionsCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/UserConnectionsCommander.scala
@@ -145,17 +145,20 @@ class UserConnectionsCommander @Inject() (
   }
 
   def getConnectionsPage(userId: Id[User], page: Int, pageSize: Int): (Seq[ConnectionInfo], Int) = {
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/UserConnectionsCommander.scala
```diff
@@ -145,17 +145,20 @@ class UserConnectionsCommander @Inject() (
...
+      val list = connections.drop(page * pageSize).take(pageSize)
+      val basicUsers = basicUserRepo.loadAll(list.map(_._1).toSet)
+      val infos = list.map {
         case (friendId, unfriended) =>
-          ConnectionInfo(basicUserRepo.load(friendId), friendId, unfriended, searchFriends.contains(friendId))
+          ConnectionInfo(basicUsers(friendId), friendId, unfriended, searchFriends.contains(friendId))
       }
+      (infos, connectionIds.size)
     }
-    (infos.drop(page * pageSize).take(pageSize), infos.filterNot(_.unfriended).size)
```
eishay, line 158: yes that part was bad

PR #12335, opened 2014-11-12
Update library search urls and enable non-logged-in user public library search.

yipingliao:
@andrewconner @ahsu1230 
https://app.asana.com/0/18079737520750/18807532114522

UI is still pending discussion with Danny. Current UI for dev purposes is attached as screenshot. Please review for functionality.

Besides allowing for public library search for non-logged-in users, this pull also:
- Uses a `/:userName/:librarySlug/find?q=` url for library searches for both logged-in and logged-out users; this url can be pasted and the correct search results will render along with the correct library chip in the header search input.
- Moves library search into `LibraryController` and `library.tpl.html` (so as to enable non-logged-in user library search to render search results under the library header card)
- Fixes a small bug where the "All Keeps" link on a search results page is disabled if no other users besides the user and the users' friend have keeps that match the search query

screen shot 2014-11-12 at 1 12 47 pm

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/layout/header/HeaderCtrl.js
+++ b/kifi-backend/modules/shoebox/angular/src/layout/header/HeaderCtrl.js
@@ -24,7 +24,6 @@ angular.module('kifi')
     $scope.isFocused = false;
     $scope.library = {};
     $scope.search = { text: '', showName: false };
-    $scope.stayInLibraryPath = '';
 
     // Temp callout method. Remove after most users know about libraries. (Oct 26 2014)
     var calloutName = 'tag_callout_shown';
@@ -38,18 +37,24 @@ angular.module('kifi')
       profileService.savePrefs(save);
     };
 
+
     //
     // Watchers & Listeners
     //
     var deregisterLibraryChip = $rootScope.$on('libraryUrl', function (e, library) {
       $scope.library = library;
-      $scope.search.text = '';
+
+      if ($scope.library.owner && !$scope.library.owner.picUrl) {
+        $scope.library.owner.picUrl = friendService.getPictureUrlForUser($scope.library.owner);
+      }
+
       if ($scope.library.id) {
         $scope.search.showName = true;
-        $scope.stayInLibraryPath = $scope.library.url;
       } else {
         $scope.clearLibraryName();
       }
+
+      $scope.search.text = $routeParams.q || '';
     });
     $scope.$on('$destroy', deregisterLibraryChip);
 
@@ -74,45 +79,44 @@ angular.module('kifi')
 
     $scope.clearLibraryName = function () {
       $scope.search.showName = false;
-      $scope.stayInLibraryPath = '';
       $scope.library = {};
-      if ($location.path() === '/find') {
-        $location.search('l', '');
-      }
+      $scope.changeSearchInput();
     };
 
     $scope.search.text = $routeParams.q || '';
     $scope.changeSearchInput = _.debounce(function () {
-      if ($scope.search.text === '' && $scope.stayInLibraryPath !== '') {
-        $timeout(function() {
-          $location.url($scope.stayInLibraryPath);
-        }, 0);
-        $scope.clearInput();
-      } else {
-        $timeout(function() {
-          var libId = $scope.library.id;
-          if ($location.path() !== '/find') {
-            if (libId) {
-              $location.url('/find?q=' + $scope.search.text + '&f=a' + '&l=' + libId);
-            } else {
-              $location.url('/find?q=' + $scope.search.text);
+      $timeout(function() {
+        // We are not already on the search page.
+        if ($location.path() !== '/find') {
+          if ($scope.library && $scope.library.url) {
+            if ($scope.search.text) {
+              /* jshint ignore:start */
+              if ($routeParams.q) {
+                $location.search('q', $scope.search.text).replace();
+              } else {
+                $location.url($scope.library.url + '/find?q=' + $scope.search.text + '&f=a');
+              }
+              /* jshint ignore:end */
+            } else {  // jshint ignore:line
+              $location.url($scope.library.url);
             }
-          } else {
-            $location.search('q', $scope.search.text).replace(); // this keeps any existing URL params
+          } else if ($scope.search.text) {
+            $location.url('/find?q=' + $scope.search.text);
           }
-        });
-      }
+        }
+
+        // We are already on the search page.
+        else {
+          $location.search('q', $scope.search.text).replace(); // this keeps any existing URL params
+        }
+      });
     }, 100, {
       'leading': true
     });
 
     $scope.clearInput = function () {
-      if ($scope.stayInLibraryPath !== '') {
-        $timeout(function () {
-          $location.url($scope.stayInLibraryPath);
-        }, 0);
-      }
       $scope.search.text = '';
+      $scope.changeSearchInput();
     };
 
     var KEY_ESC = 27, KEY_DEL = 8;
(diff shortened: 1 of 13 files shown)


kifi-backend/modules/shoebox/angular/src/layout/header/HeaderCtrl.js
```diff
@@ -74,45 +79,44 @@ angular.module('kifi')
...
-            if (libId) {
-              $location.url('/find?q=' + $scope.search.text + '&f=a' + '&l=' + libId);
-            } else {
-              $location.url('/find?q=' + $scope.search.text);
+      $timeout(function() {
+        // We are not already on the search page.
+        if ($location.path() !== '/find') {
+          if ($scope.library && $scope.library.url) {
+            if ($scope.search.text) {
+              /* jshint ignore:start */
```
yipingliao, line 93: jshint is unhappy that there are so many nested if's. :(

kifi-backend/modules/shoebox/angular/src/layout/header/header.tpl.html
```diff
@@ -8,28 +8,28 @@
     </div>
 
     <div class="kf-header-search-bar">
-        <div class="kf-query-lib-container" ng-class="{empty: isEmpty()}">
```
yipingliao, line 11: Just indentation fixes in this file.

kifi-backend/modules/shoebox/angular/src/libraries/library.js
```diff
@@ -4,21 +4,20 @@ angular.module('kifi')
 
 .controller('LibraryCtrl', [
   '$scope', '$rootScope', '$location', '$routeParams', 'keepDecoratorService', 'libraryService',
-  'modalService', 'profileService', 'util', '$window', '$analytics',
+  'modalService', 'profileService', 'util', '$window', '$analytics', 'librarySearch',
```
yipingliao, line 7: `librarySearch` is a boolean that indicates whether we should display just the library keeps or display search results per the query parameters.

kifi-backend/modules/shoebox/angular/src/libraries/library.js
```diff
@@ -32,10 +31,15 @@ angular.module('kifi')
       $analytics.pageTrack(url, trackLibraryAttributes);
     }
 
+    function setTitle(lib) {
```
yipingliao, line 34: There is some movement of code into the appropriate sections that is not related to library search.

kifi-backend/modules/shoebox/angular/src/libraries/library.tpl.html
```diff
@@ -12,33 +12,39 @@
...
-      keeps-has-more="hasMore"
-      scroll-disabled="!hasMore"
-      scroll-distance="scrollDistance"
-      scroll-next="getNextKeeps"
-      edit-mode="editMode"
-      toggle-edit="toggleEdit"
-      update-selected-count="updateSelectedCount(numSelected)"
-      keep-click="libaryKeepClicked"
-    ></div>
+    <div ng-if="librarySearch" ng-include="'search/search.tpl.html'" ng-controller="SearchCtrl"></div>
```
yipingliao: This is where search gets included inside libraries.

kifi-backend/modules/shoebox/angular/src/search/search.js
```diff
@@ -24,47 +30,58 @@ angular.module('kifi')
...
+    // Internal methods.
     //
+    function getFilterCount(type) {
+      switch (type) {
+      case 'm':
+        return $scope.resultTotals.myTotal;
+      case 'f':
+        return $scope.resultTotals.friendsTotal;
+      case 'a':
+        return $scope.resultTotals.myTotal + $scope.resultTotals.friendsTotal + $scope.resultTotals.othersTotal;
```
yipingliao, line 42: Originally, this was just `$scope.resultTotals.othersTotal`, which disables the "All Keeps" link once you click on another link like "My Keeps" or "Friends Keeps" if `othersTotal` is 0. :(

kifi-backend/modules/shoebox/angular/src/search/search.tpl.html
```diff
@@ -1,18 +1,19 @@
 <div class="kf-main-search">
   <div class="kf-main-head">
-    <div class="kf-search-filters">
+    <div ng-if="$root.userLoggedIn" class="kf-search-filters">
```
yipingliao, line 3: Show edit options only if the user is logged in.

kifi-backend/modules/shoebox/angular/src/search/searchActionService.js
```diff
@@ -30,50 +30,56 @@ angular.module('kifi')
       return JSON.parse(JSON.stringify(obj));
     }
 
-    function decompressHit(hit, users, libraries) {
+    function decompressHit(hit, users, libraries, userLoggedIn) {
```
yipingliao, line 33: For post-processing of search results, we now take into account whether we have a logged-in user.

kifi-backend/modules/shoebox/angular/src/search/searchActionService.js
```diff
@@ -30,50 +30,56 @@ angular.module('kifi')
...
       var decompressedKeepers = [];
       var decompressedLibraries = [];
-      var myLibraries = [];
+      var myLibraries = [];  // myLibraries doesn't seem to be used anywhere.
```
yipingliao, line 36: @stkem Can I remove `myLibraries`? It doesn't seem to be used anywhere.

kifi-backend/modules/shoebox/angular/src/layout/header/HeaderCtrl.js
ahsu1230, line 93: yikes this is a tough one to shorten =/

kifi-backend/modules/shoebox/angular/src/route.js
```diff
@@ -49,9 +49,17 @@ angular.module('kifi')
       controller: 'ManageTagCtrl'
     })
     // ↓↓↓↓↓ Important: This needs to be last! ↓↓↓↓↓
+    .when('/:username/:librarySlug/find', {
+      templateUrl: 'libraries/library.tpl.html',
+      controller: 'LibraryCtrl',
+      resolve: { librarySearch: function () { return true; } },
```
ahsu1230, line 55: is this how the `LibraryCtrl` receives `librarySearch` as a boolean value?

yipingliao, line 55: Yes. In addition, in `libraryCard.js`, I also emit an event that will update `librarySearch` when the user moves from non-search to search and vice versa. In that case, I am preventing a route reload so this route logic will not run and I have to force update `librarySearch`.

kifi-backend/modules/shoebox/angular/src/layout/header/HeaderCtrl.js
ahsu1230, line 93: to take out some nested-if's to make this function more modular
you could try having the smaller if's `return` before coming to the main logic part.
For example:

```
if ($location.path() === '/find') { // on search page
   $location.search('q', $scope.search.text).replace();
   return;
}
var atLibrary = Boolean(library && library.url)
if (!atLibrary) {
   if ($scope.search.text) {
      $location.url('/find?q=' + $scope.search.text);
   }
   return;
}
// at this point, you know library && library.url is truthy
// from here, you can do the rest of the logic
```

It might make the code a little more readable, and with the more trivial if's taken out, you might not need jshint.
Whatever makes more sense =)

yipingliao, line 93: Thanks! I'll look into this some more.

kifi-backend/modules/shoebox/angular/src/search/searchActionService.js
stkem, line 36: If you are sure it's not used, go ahead and remove it (I can't think of a usage)

PR #12337, opened 2014-11-12
digest email updates

joshm1:
@stkem 

other than my unit tests I don't have a easy method of testing this email in dev.  I'll need to send myself a digest email in prod


stkem:
:+1:

PR #12338, opened 2014-11-12
When login link is clicked on a mobile device, go to the app or the app store.

yipingliao:
@andrewconner 
How can I test this in dev?

https://app.asana.com/0/19898402862947/18964106149131

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/layout/header/PublicHeaderCtrl.js
+++ b/kifi-backend/modules/shoebox/angular/src/layout/header/PublicHeaderCtrl.js
@@ -8,13 +8,18 @@ angular.module('kifi')
 
     $scope.join = function ($event) {
       $event.preventDefault();
-      $scope.$emit('getCurrentLibrary', { callback: function (lib) {
-        var userData;
-        if (lib && lib.id) {
-          userData = { libraryId: lib.id };
-        }
-        signupService.register(userData);
-      }});
+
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/angular/src/layout/header/PublicHeaderCtrl.js
```diff
@@ -8,13 +8,18 @@ angular.module('kifi')
...
-      $scope.$emit('getCurrentLibrary', { callback: function (lib) {
-        var userData;
-        if (lib && lib.id) {
-          userData = { libraryId: lib.id };
-        }
-        signupService.register(userData);
-      }});
+
+      if (platformService.isSupportedMobilePlatform()) {
+        platformService.goToAppOrStore();
```
andrewconner, line 13: Do we need a `$event.preventDefault`?

andrewconner:
Chrome DevTools can pretend to be a phone. Happy to show you.

kifi-backend/modules/shoebox/angular/src/layout/header/PublicHeaderCtrl.js
yipingliao, line 13: Yes, otherwise the browser will go to "/signup".

yipingliao:
Works in dev now. Had to pull in your latest changes. Gonna merge.

PR #12341, opened 2014-11-12
adding people microdata tags

eishay:
see https://support.google.com/webmasters/answer/146646


eishay:
@andrewconner @yipingliao

eishay:
see https://github.com/kifi/keepit/pull/12342

PR #12344, opened 2014-11-12
Youtube Video Spotlight (play youtube in a keep card)

ahsu1230:
Figure out if the link is in youtube format
then display it with a 3-max line description below it (uses ellipsis directive)

@andrewconner @yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/keep/keep.js
+++ b/kifi-backend/modules/shoebox/angular/src/keep/keep.js
@@ -271,11 +271,12 @@ angular.module('kifi')
         };
 
         scope.showSmallImage = function (keep) {
-          return keep.hasSmallImage && !useBigLayout;
+          return keep.hasSmallImage && !useBigLayout && !scope.isYoutubeCard;
         };
 
         scope.showBigImage = function (keep) {
-          return keep.hasBigImage || (keep.summary && useBigLayout);
+          var bigImageReady = keep.hasBigImage || (keep.summary && useBigLayout);
+          return bigImageReady && !scope.isYoutubeCard;
         };
 
         scope.hasTag = function (keep) {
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/shoebox/angular/src/keep/keep.js
```diff
@@ -82,6 +82,7 @@ angular.module('kifi')
         var useBigLayout = false;
         var strippedSchemeRe = /^https?:\/\//;
         var domainTrailingSlashRe = /^([^\/]*)\/$/;
+        var youtubeLinkRe = /.*(?:youtu.be\/|v\/|u\/\w\/|embed\/|watch\?v=)([^#\&\?]*).*/;
```
ahsu1230, line 85: taken from http://stackoverflow.com/questions/3452546/javascript-regex-how-to-get-youtube-video-id-from-url

kifi-backend/modules/shoebox/angular/src/keep/keep.js
```diff
@@ -114,6 +116,17 @@ angular.module('kifi')
...
+        function isYoutubeCard(url) {
+          var strippedSchemeLen = (url.match(strippedSchemeRe) || [''])[0].length;
+          var match = url.substr(strippedSchemeLen).match(youtubeLinkRe);
+          if (match && match[1]) {
+            scope.isYoutubeCard = true;
+            scope.youtubeId = match[1];
+          }
+          return match;
+        }
+        isYoutubeCard(scope.keep.url);
```
yipingliao: This function call should be moved to the bottom of the controller code, in the "on link" section. Alternatively (and I like this better), @andrewconner has started to make an `init()` function where all the initializing code goes, and invokes `init()` at the bottom of a controller or directive's link code.  For an example, see `search.js`.

yipingliao:
:+1:

kifi-backend/modules/shoebox/angular/src/keep/keep.js
ahsu1230: I see, will do that

PR #12347, opened 2014-11-12
use different `top` css value for sticky button on mobile/desktop

joshm1:
also fixes z-index issue in iOS when scrolling over the header

@stkem


stkem:
Great!

PR #12357, opened 2014-11-13
Implement basic clickable url in library description.

yipingliao:
@andrewconner @2is10 cc @ahsu1230 

Probably too basic. There is a bunch of code in https://github.com/kifi/keepit/blob/master/packrat/scripts/formatting.js that I don't understand (and didn't include). Will discuss in detail with @2is10 .

https://app.asana.com/0/19898402862947/19898402862972

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
@@ -5,10 +5,10 @@ angular.module('kifi')
 .directive('kfLibraryCard', [
   '$FB', '$location', '$q', '$rootScope', '$window', 'env', 'friendService', 'libraryService', 'modalService',
   'profileService', 'platformService', 'signupService', '$twitter', '$timeout', '$routeParams', '$route',
-  'locationNoReload',
+  'locationNoReload', 'util',
   function ($FB, $location, $q, $rootScope, $window, env, friendService, libraryService, modalService,
       profileService, platformService, signupService, $twitter, $timeout, $routeParams, $route,
-      locationNoReload) {
+      locationNoReload, util) {
     return {
       restrict: 'A',
       replace: true,
@@ -25,6 +25,7 @@ angular.module('kifi')
         //
         // Internal data.
         //
+        var uriRe = /(?:\b|^)((?:(?:(https?|ftp):\/\/|www\d{0,3}[.])?(?:[a-z0-9](?:[-a-z0-9]*[a-z0-9])?\.)+(?:com|edu|biz|gov|in(?:t|fo)|mil|net|org|name|coop|aero|museum|[a-z][a-z]\b))(?::[0-9]{1,5})?(?:\/(?:[^\s()<>]*[^\s`!\[\]{};:.'",<>?«»()“”‘’]|\((?:[^\s()<>]+|(?:\([^\s()<>]+\)))*\))*|\b))(?=[\s`!()\[\]{};:.'",<>?«»“”‘’]|$)/;  // jshint ignore:line
         var authToken = $location.search().authToken || '';
         var prevQuery = '';
 
@@ -104,18 +105,20 @@ angular.module('kifi')
           });
         }
 
-        function processUrls(text) {
-          var uriRe = /(?:\b|^)((?:(?:(https?|ftp):\/\/|www\d{0,3}[.])?(?:[a-z0-9](?:[-a-z0-9]*[a-z0-9])?\.)+(?:com|edu|biz|gov|in(?:t|fo)|mil|net|org|name|coop|aero|museum|[a-z][a-z]\b))(?::[0-9]{1,5})?(?:\/(?:[^\s()<>]*[^\s`!\[\]{};:.'",<>?«»()“”‘’]|\((?:[^\s()<>]+|(?:\([^\s()<>]+\)))*\))*|\b))(?=[\s`!()\[\]{};:.'",<>?«»“”‘’]|$)/;  // jshint ignore:line
 
+        function processUrls(text) {
           var parts = text.split(uriRe);
+
           for (var i = 1; i < parts.length; i += 3) {
             var uri = parts[i];
             var scheme = parts[i+1];
-            var url = (scheme ? '' : 'http://') + uri;
+            var url = (scheme ? '' : 'http://') + util.htmlEscape(uri);
 
             parts[i] = '<a target="_blank" href="' + url + '">' + url;
             parts[i+1] = '</a>';
+            parts[i-1] = util.htmlEscape(parts[i-1]);
           }
+          parts[parts.length-1] = util.htmlEscape(parts[parts.length-1]);
 
           return parts.join('');
         }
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.tpl.html
```diff
@@ -39,7 +39,7 @@
 
       <div ng-switch="clippedDescription" class="kf-keep-lib-description-container">
         <p ng-switch-when="true" class="kf-keep-lib-description">
-          {{library.shortDescription}}&hellip;
+          <span ng-bind-html="library.shortDescription"></span>&hellip;
```
yipingliao, line 42: Sanitize the short description. Currently the long description is sanitized but the short one is not. :(

andrewconner, line 42: Yeah, mind adding this? Long desc gets sanitized with `angular.element('<div>').text(scope.library.description).text()`

kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
```diff
@@ -104,6 +104,22 @@ angular.module('kifi')
           });
         }
 
+        function processUrls(text) {
+          var uriRe = /(?:\b|^)((?:(?:(https?|ftp):\/\/|www\d{0,3}[.])?(?:[a-z0-9](?:[-a-z0-9]*[a-z0-9])?\.)+(?:com|edu|biz|gov|in(?:t|fo)|mil|net|org|name|coop|aero|museum|[a-z][a-z]\b))(?::[0-9]{1,5})?(?:\/(?:[^\s()<>]*[^\s`!\[\]{};:.'",<>?«»()“”‘’]|\((?:[^\s()<>]+|(?:\([^\s()<>]+\)))*\))*|\b))(?=[\s`!()\[\]{};:.'",<>?«»“”‘’]|$)/;  // jshint ignore:line
```
2is10: It would be better to only instantiate this regular expression once, instead of once per `processUrls` call.

kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
```diff
@@ -104,6 +104,22 @@ angular.module('kifi')
...
+
+          var parts = text.split(uriRe);
+          for (var i = 1; i < parts.length; i += 3) {
+            var uri = parts[i];
+            var scheme = parts[i+1];
+            var url = (scheme ? '' : 'http://') + uri;
+
+            parts[i] = '<a target="_blank" href="' + url + '">' + url;
```
2is10, line 117: `url` should be HTML-escaped before being inserted into an HTML attribute value or element text. Angular may protect against exploits, but it can’t repair malformed HTML (e.g. if the URL contains `"`).

Here’s an efficient `htmlEscape` function. I can’t find one exposed by Angular—it tries to abstract this away from you, and doesn’t seem to like raw HTML string construction, like we’re doing here.

``` js
  var HTML_ESCAPE_CHARS = /[&<>"'\/]/g;
  var HTML_ESCAPES = {
    '&': '&amp;',
    '<': '&lt;',
    '>': '&gt;',
    '"': '&quot;',
    "'": '&#x27;',
    '/': '&#x2F;'
  };
  function htmlEscape(text) {
    return text == null ? '' : String(text).replace(HTML_ESCAPE_CHARS, htmlEscapeReplace);
  }
  function htmlEscapeReplace(ch) {
    return HTML_ESCAPES[ch];
  }
```

2is10, line 117: (It would be much more efficient to run auto-linkification only once, on our servers, when text is saved/sent, rather than in every viewer’s browser on every view. But when we were prototyping messaging, the team didn’t want to invest the time.)

kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.tpl.html
2is10, line 42: I believe `ng-bind-html` is for rendering HTML, not plain text. If a user types `(<o>)` in the description, for whatever reason, we want to display it that way. I’m afraid `ng-bind-html` would end up displaying `()` instead.

2is10, line 42: I’m not sure why we need to do the sanitization Andrew describes. In my opinion, we should be treating the description as text, not HTML. The user didn’t type it as HTML. The user typed it as text. I believe Angular automatically HTML-escapes values included in templates like this: `{{library.shortDescription}}`. Mustache and Handlebars do.

In the long description case, where we’re building a trusted HTML string from untrusted text, we can’t render it that way (between `{{` and `}}`) if that HTML-escapes it. We’ll need to use `ng-bind-html` or something like that. The string we plug in there should be comprised solely of HTML characters we inserted (such as `<a href="`) and HTML-escaped untrusted text from the user. This means we manually HTML-escape any detected URLs _and_ every other text fragment before/after/between URLs. Otherwise, even if `ng-bind-html` eliminates threats, we’ll still end up rendering the user’s text as HTML, which can ruin it (e.g. transforming `&amp;` into `&` when the user really wanted `&amp;` and making things that look like HTML tags render invisibly, or with visible effects on some of the text, such as bolding). We don’t yet support rich text in descriptions. When we do, I don’t think it should be accomplished by allowing the user to type certain HTML tags. I hope this is clear.

kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
```diff
@@ -104,6 +104,22 @@ angular.module('kifi')
...
+            parts[i+1] = '</a>';
+          }
+
+          return parts.join('');
```
2is10, line 123: As I described in my long comment below, the non-URL text fragments need to be HTML-escaped too, before joining here. I mean `parts[0]`, `parts[3]`, `parts[6]`, etc.

kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
andrewconner, line 117: Sounds like a good backend project — we'd need a migration path where the client would accept HTML, but seems pretty straight forward.

kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.tpl.html
andrewconner, line 42: Good points. My comment was about using `ng-bind-html` (I didn't investigate why we'd need to switch, we may not need to). `{{library.shortDescription}}` is safe (HTML is escaped).

2is10, line 42: Here’s a good test library for you guys. Looks like the long description is implemented incorrectly today.
https://www.kifi.com/jaredfortytwo/html-examples

2is10, line 42: I think we want to continue rendering the short desc using `{{library.shortDescription}}`. It doesn’t need sanitization because Angular automatically HTML-escapes it.

I think we need to stop doing this to the full description, if, in fact, we currently are:
`angular.element('<div>').text(scope.library.description).text()`

`{{library.description}}` would be safe, but to support our auto-detected links, I think we’ll instead need to use `ng-bind-html` and manually HTML-escape every user-typed text fragment.

kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
yipingliao: Will do. Thanks!

yipingliao:
@2is10 Thank you so much for your help! I've made a new commit that should correctly escape HTML so that HTML and links show up correctly for both the short and long description.
- Since a link can be in the short or long description, and that link should not be escaped, I am now using `ng-bind-html` for both of them (instead of `{{ }}`).
- The description text is now processed such that urls are wrapped in links and all user-entered text is HTML escaped.

2is10:
​LGTM

PR #12361, opened 2014-11-13
extra check for displaying small images in keep card

ahsu1230:
Try keeping this:
https://en.wikipedia.org/wiki/Chromesthesia

You will see in your keep card that the entire description is left as is (no truncation). 
Turns out `maybeSizeImage()` in keep.js looks for images above 110px height & width. 
But `showSmallImage()` doesn't check for that so it accidentally displays small images.

Hopefully I'm not missing an edge case.

Also, keep descriptions now use kf-ellipsis directive!!
@andrewconner @yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/common/directives/ellipsis/ellipsis.js
+++ b/kifi-backend/modules/shoebox/angular/src/common/directives/ellipsis/ellipsis.js
@@ -25,7 +25,7 @@ angular.module('kifi')
     },
     compile: function(/*elem, attr, linker*/) {
 
-      return function(scope, element, attributes) {
+      return function(scope, element) {
         var copyElement = element.clone()
           .css({
             'display': 'block',
@@ -51,7 +51,7 @@ angular.module('kifi')
             var currentHeight = copyElement.height();
             var currentNumLines = currentHeight / heightPerLine;
             var maxIndex = scope.ngBind.length;
-            var maxNumLines = attributes.maxNumLines || 1;
+            var maxNumLines = scope.maxNumLines || currentNumLines;
 
             if (currentHeight === 0) {
               element.html(scope.ngBind);
--- a/kifi-backend/modules/shoebox/angular/src/common/directives/youtube.js
+++ b/kifi-backend/modules/shoebox/angular/src/common/directives/youtube.js
@@ -7,13 +7,16 @@ angular.module('kifi')
   function () {
 
     function videoIdToSrc(videoId) {
-      return '//www.youtube.com/v/' + videoId +
-        '&rel=0&theme=light&showinfo=0&disablekb=1&modestbranding=1&controls=1&hd=1&autohide=1&color=white&iv_load_policy=3';
+      // return '//www.youtube.com/v/' + videoId +
+        // '&rel=0&theme=light&showinfo=0&disablekb=1&modestbranding=1&controls=1&hd=1&autohide=1&color=white&iv_load_policy=3';
+      return '//www.youtube.com/embed/' + videoId +
+        '?rel=0&theme=light&showinfo=0&disablekb=1&modestbranding=1&controls=1&hd=1&autohide=1&color=white&iv_load_policy=3';
     }
 
     function videoEmbed(src) {
-      return '<embed src="' + src +
-        '" type="application/x-shockwave-flash" allowfullscreen="true" style="width:100%; height: 100%;" allowscriptaccess="always"></embed>';
+      // return '<embed src="' + src +'"
+      //    type="application/x-shockwave-flash" allowfullscreen="true" style="width:100%; height: 100%;" allowscriptaccess="always"></embed>';
+      return '<iframe width="100%" height="100%" src="' + src + '" frameborder="none" allowfullscreen="true" allowscriptaccess="always"/>';
     }
 
     return {
--- a/kifi-backend/modules/shoebox/angular/src/common/directives/youtube.js
+++ b/kifi-backend/modules/shoebox/angular/src/common/directives/youtube.js
@@ -7,16 +7,22 @@ angular.module('kifi')
   function () {
 
     function videoIdToSrc(videoId) {
-      // return '//www.youtube.com/v/' + videoId +
-        // '&rel=0&theme=light&showinfo=0&disablekb=1&modestbranding=1&controls=1&hd=1&autohide=1&color=white&iv_load_policy=3';
       return '//www.youtube.com/embed/' + videoId +
-        '?rel=0&theme=light&showinfo=0&disablekb=1&modestbranding=1&controls=1&hd=1&autohide=1&color=white&iv_load_policy=3';
+        '?rel=0&theme=light&showinfo=0&disablekb=1&modestbranding=1&controls=1&hd=1&autoplay=1&autohide=1&color=white&iv_load_policy=3';
     }
 
     function videoEmbed(src) {
-      // return '<embed src="' + src +'"
-      //    type="application/x-shockwave-flash" allowfullscreen="true" style="width:100%; height: 100%;" allowscriptaccess="always"></embed>';
       return '<iframe width="100%" height="100%" src="' + src + '" frameborder="none" allowfullscreen="true" allowscriptaccess="always"/>';
+    }
+
+    function getVideoImage(videoId) {
+      return '//img.youtube.com/vi/' + videoId + '/hqdefault.jpg';
+    }
+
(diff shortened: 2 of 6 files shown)


kifi-backend/modules/shoebox/angular/src/common/directives/ellipsis/ellipsis.js
```diff
@@ -51,7 +51,7 @@ angular.module('kifi')
             var currentHeight = copyElement.height();
             var currentNumLines = currentHeight / heightPerLine;
             var maxIndex = scope.ngBind.length;
-            var maxNumLines = attributes.maxNumLines || 1;
```
ahsu1230, line 54: surprised how this worked before
attributes.maxNumLines turned out to be a string and apparently...
"2" <= 3 is true and "4" >= 3 is false
wut. :jack_o_lantern:

kifi-backend/modules/shoebox/angular/src/keep/keep.js
```diff
@@ -275,7 +270,8 @@ angular.module('kifi')
         };
 
         scope.showSmallImage = function (keep) {
-          return keep.hasSmallImage && !useBigLayout && !scope.isYoutubeCard;
+          var hasImage = keep.summary.imageWidth > 110 && keep.summary.imageHeight > 110;
```
ahsu1230, line 283: don't display small images

kifi-backend/modules/shoebox/angular/src/keep/keepCard.tpl.html
```diff
@@ -31,11 +31,7 @@
         </div>
 
         <div class="kf-keep-info-wrap" ng-if="!isYoutubeCard">
```
andrewconner: May be better to go the other way. We may have several variants, and `ng-if="!isYoutubCard && !isVimeoCard && !isSoundcloudCard && ..."` will become unmaintainable quickly.

kifi-backend/modules/shoebox/angular/src/keep/keepCard.tpl.html
```diff
@@ -31,11 +31,7 @@
...
-          <div class="kf-keep-info">
-            <div class="kf-keep-description">
-              {{keep.summary.description}}
-            </div>
-          </div>
+          <div kf-ellipsis data-ng-bind="keep.summary.description" data-max-num-lines="linesToShow" class="kf-keep-description"></div>
```
andrewconner: Wahoo

andrewconner:
Thanks! Let me pull this and double check some edge cases before merging. There's several ways it can get messed up. It looks right, but swapping elements may change semantics in unexpected ways.

ahsu1230:
Ok thanks! I've been using different keeps to check logged in & logged out keep cards. I think it looks right. But some double checking would probably be helpful

yipingliao:
Nice!

kifi-backend/modules/shoebox/angular/src/keep/keep.js
```diff
@@ -229,23 +229,18 @@ angular.module('kifi')
...
 
           keep.sizeCard = function () {
             var $content = element.find('.kf-keep-content-line');
             //$content.height(Math.floor(bestRes.hi) + 4); // 4px padding on image
-            $content.find('.kf-keep-small-image').width(asideWidthPercent + '%');
-            if (calcTextHeight > 20) {
-              element.find('.kf-keep-info').css({
-                'height': calcTextHeight + 'px'
-              }).addClass('kf-dyn-positioned');
+            if ($content && $content.length > 0) {
```
andrewconner: Was `$content` ever undefined?

kifi-backend/modules/shoebox/angular/src/keep/keep.js
```diff
@@ -229,23 +229,18 @@ angular.module('kifi')
...
+          scope.linesToShow = linesToShow;
 
           keep.sizeCard = function () {
             var $content = element.find('.kf-keep-content-line');
             //$content.height(Math.floor(bestRes.hi) + 4); // 4px padding on image
-            $content.find('.kf-keep-small-image').width(asideWidthPercent + '%');
-            if (calcTextHeight > 20) {
-              element.find('.kf-keep-info').css({
-                'height': calcTextHeight + 'px'
-              }).addClass('kf-dyn-positioned');
```
andrewconner, line 243: is `kf-dyn-positioned` not needed anymore?

andrewconner, line 243: Oh, probably not, because of your ellipses directive. You can remove that css class.

kifi-backend/modules/shoebox/angular/src/keep/keepCard.tpl.html
andrewconner: `data-` is not needed (usually, maybe your directive needs it?)

kifi-backend/modules/shoebox/angular/src/common/directives/ellipsis/ellipsis.js
andrewconner, line 54: Hm?

```
> "4" >= 3
  true
> "2" <= 3
  true
```

kifi-backend/modules/shoebox/angular/src/keep/keep.js
ahsu1230, line 243: Ok. kf-dyn-positioned is not used elsewhere

kifi-backend/modules/shoebox/angular/src/keep/keep.js
ahsu1230: woops. It was a change I had before. will change it back

kifi-backend/modules/shoebox/angular/src/keep/keepCard.tpl.html
ahsu1230: how about a scope variable `isSpecialCard` that checks if any of `isYoutubeCard`, `isVimeoCard`, etc. is true?

andrewconner: How about `cardType`, and we have several types, such as `small`, `big`, `youtube`, etc.

kifi-backend/modules/shoebox/angular/src/common/directives/ellipsis/ellipsis.js
ahsu1230, line 54: woops I meant to type "4" <= 3 is false.
But, basically `attributes.maxNumLines` is the string parameter you pass in T_T

kifi-backend/modules/shoebox/angular/src/keep/keepCard.tpl.html
ahsu1230: that works better

ahsu1230:
now uses `<iframe>`

in the mean time, double checking some stuff...

kifi-backend/modules/shoebox/angular/src/common/directives/youtube.js
```diff
@@ -7,13 +7,16 @@ angular.module('kifi')
   function () {
 
     function videoIdToSrc(videoId) {
-      return '//www.youtube.com/v/' + videoId +
-        '&rel=0&theme=light&showinfo=0&disablekb=1&modestbranding=1&controls=1&hd=1&autohide=1&color=white&iv_load_policy=3';
+      // return '//www.youtube.com/v/' + videoId +
+        // '&rel=0&theme=light&showinfo=0&disablekb=1&modestbranding=1&controls=1&hd=1&autohide=1&color=white&iv_load_policy=3';
+      return '//www.youtube.com/embed/' + videoId +
+        '?rel=0&theme=light&showinfo=0&disablekb=1&modestbranding=1&controls=1&hd=1&autohide=1&color=white&iv_load_policy=3';
```
ahsu1230: any reason why we would want to leave the old way commented out?

andrewconner: Nope. Git will have it

ahsu1230:
now shows an image for the youtube keep card, then replaces with embedded youtube video

andrewconner:
It looks great. Two minor things:
- Let's not hotlink their image for the play button. Add it to our images.
- The image's aspect ratio seems to be variable, so sometimes it is stretched. Perhaps we should crop it?

ahsu1230:
gotcha. I'll add the play button image.
For the aspect ratio are you talking about the video screen shot? Or does the play button look stretched?

PR #12363, opened 2014-11-13
support some type of user agent

yingjieMiao:
e.g.

Kifi/18168 CFNetwork/672.1.13 Darwin/14.0.0

device and os version is not available from there.


yingjieMiao:
@eishay  FYI

PR #12365, opened 2014-11-13
Differentiate between missing keep image and no keep image

leogrim:
@andrewconner 
The `Option[Option[KeepImage]]` API looks a bit weird. An alternative would be `Either[MissingImage, Option[KeepImage]]` with `MissingImage` being a singleton object, so that's semantically equivalent to options.


andrewconner:
Interesting. Seems like externally you always want to flatten. Could this be handled internally, so it doesn't leak the fact that it differentiates?

leogrim:
Well it depends what you mean by "externally". In order to build the keep summary (which seems "external" to KeepImageCommander), we need to know what the keep image state is. What would you have in mind?

PR #12368, opened 2014-11-13
Fix bug where a short description may chop off in the middle of a link.

yipingliao:
@andrewconner cc @ahsu1230 
Related to:
https://app.asana.com/0/19898402862947/19898402862972

The previous truncates at 150 characters to generate the library's short description after link insertion is performed. This may result in the truncation happening in the middle of a link. The change is to perform the truncation before the link insertion.


yipingliao:
Merging fast since it's a bug fix. Will address any comments later.

PR #12372, opened 2014-11-13
angular tests lol

ahsu1230:
thanks @andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/angular/test/unit/common/directives/youtube.spec.js
+++ b/kifi-backend/modules/shoebox/angular/test/unit/common/directives/youtube.spec.js
@@ -38,17 +38,16 @@ describe('kifi.youtube', function () {
       element.remove();
     });
 
-    it('should append an embed element as a child', function () {
-      var child = element.find('embed')[0];
-      expect(child.nodeName).toBe('EMBED');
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/angular/test/unit/common/directives/youtube.spec.js
```diff
@@ -38,17 +38,16 @@ describe('kifi.youtube', function () {
       element.remove();
     });
 
-    it('should append an embed element as a child', function () {
-      var child = element.find('embed')[0];
-      expect(child.nodeName).toBe('EMBED');
-      expect(child.type).toBe('application/x-shockwave-flash');
+    it('should append an iframe element as a child', function () {
```
2is10, line 41: by "iframe" you mean "img", of course, yes?

ahsu1230, line 41: oh yea, sorry, good catch.

ahsu1230, line 41: actually woops, I later changed it to a div. Sothe test description is correct now =P

PR #12373, opened 2014-11-13
Revert "Adding 'active' field to BasicUser"

eishay:
Reverts kifi/keepit#12366
We'll deal with inactive users in a different way, keeping the BasicUser object small.
I'll update the cache
@andrewconner @ymatsuda


eishay:
bumped the cache version up, ready to merge

andrewconner:
Looks good.

PR #12374, opened 2014-11-14
library names in left sidebar use ellipsis directive!

ahsu1230:
no more lame javascript space splitting!

@andrewconner @yipingliao


kifi-backend/modules/shoebox/angular/src/layout/nav/nav.tpl.html
```diff
@@ -98,8 +98,7 @@
                 <img class="library-icon-owner-img" ng-if="!library.isMine" ng-src="{{library.owner.image || 'https://www.kifi.com/assets/img/ghost.200.png'}}" title="{{library.owner.firstName + ' ' + library.owner.lastName}}"></img>
               </div>
               <div class="kf-nav-lib-info">
-                <div class="kf-nav-lib-name">{{library.firstLine}}</div>
-                <div class="kf-nav-lib-name">{{library.secondLine}}</div>
+                <div kf-ellipsis ng-bind="library.name" max-num-lines=2 class="kf-nav-lib-name"></div>
```
yipingliao, line 101: Are you using `ng-bind` instead of `{{ }}` to prevent a display of raw content?

kifi-backend/modules/shoebox/angular/src/libraries/libraryService.js
```diff
@@ -87,43 +87,11 @@ angular.module('kifi')
     //
     // Internal helper methods.
     //
-    function shortenLibName(fullName) {
```
yipingliao, line 90: Yay!

yipingliao:
Nice!

kifi-backend/modules/shoebox/angular/src/layout/nav/nav.tpl.html
ahsu1230, line 101: yea, is that unconventional?

kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.tpl.html
```diff
@@ -20,7 +20,7 @@
 
       <div ng-switch="recommendation" class="kf-keep-lib-name-container kf-keep-title">
         <a ng-switch-when="true" class="kf-keep-lib-name" ng-href="{{library.url}}" ng-attr-title="{{library.name}}" ng-bind-html="library.name"></a>
-        <span ng-switch-when="false" class="kf-keep-lib-name" title="{{library.name}}" kf-ellipsis data-ng-bind="library.name" data-max-num-lines=2></span>
+        <span ng-switch-when="false" class="kf-keep-lib-name" title="{{library.name}}" kf-ellipsis ng-bind="library.name" max-num-lines=2 num-reloads=3></span>
```
ahsu1230, line 23: ellipsis directive can also take in a parameter `num-reloads` which will redo the sizing after intervals. This was originally for the font-loading problem we had before for the library cards. But, I don't think it's critical enough to have to  reload multiple times for library names & keep descriptions =P

kifi-backend/modules/shoebox/angular/src/layout/nav/nav.tpl.html
2is10, line 101: `ng-bind` is for displaying HTML. `{{...}}` is for displaying plain text.

A library name is user-entered plain text, not HTML, so we should be using `{{...}}`.

Here’s [a library I created](https://www.kifi.com/jaredfortytwo/html-examples) to illustrate the difference. The title is currently displayed incorrectly. If you hover over the title, the browser will show you the correct title.

PR #12377, opened 2014-11-14
Fixing undefined regex parsing when there's no description

andrewconner:
@eishay @yipingliao This may not be the best holistic solution, but fixes the JS error.


eishay:
thanks

yipingliao:
Thanks for the fix!

PR #12378, opened 2014-11-14
Augment description

yipingliao:
@andrewconner 

Tonight was the second time that the website code has tripped up because of the lack of the description field in the library object, due to libraries created with the extension having no such field. This pull retains your change to fix the JS error (always good to have null check) but also augments all libraries with the description field.


andrewconner:
Looks good. It _is_ an optional field, so we'll always need to deal with it not being included. There may be others, I'll check.

PR #12379, opened 2014-11-14
Google finds many of our meta tag descriptions too short, this should he...

eishay:
...lp fixing it.


eishay:
@andrewconner @raymondng

PR #12385, opened 2014-11-14
youtube image cropping

ahsu1230:
1) Adding youtube play button to our images
2) Turning youtube image to div so it can be set as background image & cropped (variable width) accordingly.
3) ellipsis directive had an ng-bind which was actually a scope variable. The naming was very confusing.

Thanks for the comments & help!
@andrewconner @yipingliao


andrewconner:
Awesome

yipingliao:
Hmm, I need to start keeping Youtube videos. :)

PR #12388, opened 2014-11-14
stop gap throtteling of the repo reaper

stkem:
to stop alerts. I’ll come up with a better solution soon.
@andrewconner @yingjieMiao


kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationCleanupCommander.scala
```diff
@@ -18,6 +18,7 @@ class RecommendationCleanupCommander @Inject() (
     val userToClean = db.readOnlyReplica { implicit session => uriRecoRepo.getUsersWithRecommendations() }.toSeq
     db.readWriteBatch(userToClean) { (session, userId) =>
       uriRecoRepo.cleanupLowMasterScoreRecos(userId, overrideLimit.getOrElse(defaultLimitNumRecosForUser), overrideTimeCutoff.getOrElse(currentDateTime.minusDays(4)))(session)
+      Thread.sleep(500)
```
yingjieMiao, line 21: So this makes each session 500 milli seconds longer?

andrewconner, line 21: Each session will be 500ms \* number of users longer. We've seen sessions this long time out (random dropped connections), so we'll see if this works as expected.

stkem, line 21: yeah, this is just a bandaid. I'll have a better solution Monday (randomly clean up just a couple of users, but do it frequently)

PR #12391, opened 2014-11-14
eat elephant

raymondng:
Wrapping up some progress on refactoring / improving the token-based login path:
- factored out (`SecureSocial`-free) core layer
- added adaptor layer to ensure `SecureSocial` interop
- injected new providers back into legacy `SecureSocial` provider so current `SecureSocial` code path continues to work as before
- wired-in new layer with token-based login code path
- added relevant `Fake` modules and supportive classes
- (finally) added test coverage for token-based login (+ve and -ve tests)

Some benefits of the new layer includes async API, modularity, ease-of-testing, improved robustness (type-safety) and in general fewer random/unexpected assumptions (wrt how classes are loaded, etc.)

Still a lot more work to do but I think this is a step in the right direction for us.

@andrewconner 
cc: @eishay @stkem


andrewconner:
Awesome, will review this weekend. Thanks!

andrewconner:
Thanks for the new tests. Looks good. If you want help testing locally to be sure it doesn't cause issues, happy to help.

raymondng:
Thanks Andrew. I think we are ok -- I tested locally and also, new tests :)

raymondng:
Will merge after fixit. I'll deal with the merge conflicts, if any.

PR #12392, opened 2014-11-14
updating open graph meta tags, some where broken and some missing. thanks to Leo

eishay:


eishay:
@leogrim @andrewconner

PR #12394, opened 2014-11-15
My Libraries First Switch to LHR

ahsu1230:
@andrewconner @yipingliao


kifi-backend/modules/shoebox/marketing
```diff
@@ -1 +1 @@
-Subproject commit f1f3709e7ab24b2d7bf43761f65027103646e445
+Subproject commit c68c815783dc93cd0a9af85ec7391957eae34054
```
andrewconner, line 1: This may have been an accident. `f1f3` is the latest. ie, don't merge this.

ahsu1230, line 1: yeah I realized something wrong with my branching =(

PR #12404, opened 2014-11-15
small changes to LHR

ahsu1230:
suggestions from Mark
@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/layout/nav/nav.tpl.html
+++ b/kifi-backend/modules/shoebox/angular/src/layout/nav/nav.tpl.html
@@ -84,7 +84,7 @@
     <div class="kf-scrollable-libs" antiscroll="{ autoHide: false }">
       <ul class="kf-nav-lib-users">
         <li class="kf-nav-lib-separator" ng-show="sortingMenu.myLibsFirst && myLibsToShow.length > 0">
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/angular/src/layout/nav/nav.tpl.html
```diff
@@ -120,10 +120,10 @@
 
 
         <li class="kf-nav-lib-separator" ng-show="!sortingMenu.myLibsFirst && userLibsToShow.length > 0">
-          <div class="kf-nav-lib-separator-title">All Libraries</div>
+          <div class="kf-nav-lib-separator-title">ALL LIBRARIES</div>
         </li>
         <li class="kf-nav-lib-separator" ng-show="sortingMenu.myLibsFirst && userLibsToShow.length > 0">
-          <div class="kf-nav-lib-separator-title">Others' Libraries</div>
+          <div class="kf-nav-lib-separator-title">LIBRARIES YOU FOLLOW</div>
```
2is10, line 126: FYI, all-caps can also be achieved in CSS using `text-transform: uppercase`. Not very important, just nice to know.

ahsu1230, line 126: oh yea, just recently found that out too lol

PR #12409, opened 2014-11-16
[Tutorial] Switch SocialConnectionRepo to deferred sequence number

leogrim:
Here are the required steps to to switch a repo deferred / batch sequence number assignment:
1. Switch from `sequence.incrementAndGet` to `deferredSeqNum` in the repo `save` method
2. Implement a `DbSequenceAssigner`, a `SequencingActor` and a `SequencingPlugin`
3. Bind the repo's sequencing plugin to its implementation in the service's DBSequencingModule
4. Make sure the sequencing plugin is not lazy loaded by adding an explicit injection to the service's Global
5. After pushing to production, make sure everything is ok by querying the database table for rows with negative sequence numbers: you should find some but they should _not_ accumulate: 
`select min(seq) from yourTableName where seq < 0`

See comments for actual exemples.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/model/SocialConnectionRepo.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/model/SocialConnectionRepo.scala
@@ -3,7 +3,7 @@ package com.keepit.model
 import com.google.inject.{ Provider, Inject, Singleton, ImplementedBy }
 import com.keepit.common.cache.{ JsonCacheImpl, FortyTwoCachePlugin, Key, CacheStatistics }
 import com.keepit.common.logging.AccessLog
-import com.keepit.common.db.{ State, SequenceNumber, Id }
+import com.keepit.common.db.{ DbSequenceAssigner, State, SequenceNumber, Id }
 import com.keepit.common.db.slick.DBSession.{ RWSession, RSession }
 import com.keepit.common.db.slick._
 import com.keepit.common.service.RequestConsolidator
@@ -11,6 +11,9 @@ import com.keepit.common.time.Clock
 import com.keepit.social.{ SocialNetworks, SocialNetworkType }
 import scala.concurrent.{ Await, Future }
 import scala.concurrent.duration._
+import com.keepit.common.plugin.{ SchedulingProperties, SequencingActor, SequencingPlugin }
+import com.keepit.common.actor.ActorInstance
+import com.keepit.common.healthcheck.AirbrakeNotifier
 
 @ImplementedBy(classOf[SocialConnectionRepoImpl])
 trait SocialConnectionRepo extends Repo[SocialConnection] with SeqNumberFunction[SocialConnection] {
@@ -56,7 +59,7 @@ class SocialConnectionRepoImpl @Inject() (
   def table(tag: Tag) = new SocialConnectionTable(tag)
 
   override def save(socialConnection: SocialConnection)(implicit session: RWSession): SocialConnection = {
-    val toSave = socialConnection.copy(seq = sequence.incrementAndGet())
+    val toSave = socialConnection.copy(seq = deferredSeqNum())
     super.save(toSave)
   }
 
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/app/com/keepit/model/SocialConnectionRepo.scala
```diff
@@ -56,7 +59,7 @@ class SocialConnectionRepoImpl @Inject() (
   def table(tag: Tag) = new SocialConnectionTable(tag)
 
   override def save(socialConnection: SocialConnection)(implicit session: RWSession): SocialConnection = {
-    val toSave = socialConnection.copy(seq = sequence.incrementAndGet())
+    val toSave = socialConnection.copy(seq = deferredSeqNum())
     super.save(toSave)
   }
```
leogrim, line 64: 1- Switch from `sequence.incrementAndGet` to `deferredSeqNum` in `save` method

kifi-backend/modules/shoebox/app/com/keepit/model/SocialConnectionRepo.scala
```diff
@@ -177,3 +180,20 @@ class SocialConnectionRepoImpl @Inject() (
...
+  override val interval: FiniteDuration = 20 seconds
+}
+
+@Singleton
+class SocialConnectionSequenceNumberAssigner @Inject() (db: Database, repo: SocialConnectionRepo, airbrake: AirbrakeNotifier)
+  extends DbSequenceAssigner[SocialConnection](db, repo, airbrake)
+
+class SocialConnectionSequencingActor @Inject() (
+  assigner: SocialConnectionSequenceNumberAssigner,
+  airbrake: AirbrakeNotifier) extends SequencingActor(assigner, airbrake)
```
leogrim, line 199: 2- Introduce `SocialConnectionSequenceNumberAssigner`, `SocialConnectionSequencingActor` and `SocialConnectionSequencingPlugin` / `SocialConnectionSequencingPluginImpl`

kifi-backend/modules/shoebox/app/com/keepit/shoebox/ShoeboxDbSequencingModule.scala
```diff
@@ -11,5 +11,6 @@ case class ShoeboxDbSequencingModule() extends ScalaModule {
     bind[UserConnectionSequencingPlugin].to[UserConnectionSequencingPluginImpl].in[AppScoped]
     bind[LibrarySequencingPlugin].to[LibrarySequencingPluginImpl].in[AppScoped]
     bind[LibraryMembershipSequencingPlugin].to[LibraryMembershipSequencingPluginImpl].in[AppScoped]
+    bind[SocialConnectionSequencingPlugin].to[SocialConnectionSequencingPluginImpl].in[AppScoped]
   }
 }
```
leogrim, line 16: 3- Bind `SocialConnectionSequencingPlugin` to its implementation `SocialConnectionSequencingPluginImpl` in `ShoeboxDbSequencingModule`

kifi-backend/modules/shoebox/app/com/keepit/shoebox/ShoeboxGlobal.scala
```diff
@@ -53,5 +53,6 @@ trait ShoeboxServices { self: FortyTwoGlobal =>
     require(injector.instance[UserConnectionSequencingPlugin] != null) //make sure its not lazy loaded
     require(injector.instance[LibrarySequencingPlugin] != null) //make sure its not lazy loaded
     require(injector.instance[LibraryMembershipSequencingPlugin] != null) //make sure its not lazy loaded
+    require(injector.instance[SocialConnectionSequencingPlugin] != null) //make sure its not lazy loaded
   }
 }
```
leogrim, line 58: 4- Make sure `SocialConnectionSequencingPlugin` is not lazy loaded in `ShoeboxGlobal`

yingjieMiao:
:+1:

leogrim:
5- Push to production and check the database for rows with negative sequence numbers, these rows should quickly get processed by the plugin that assigns them a with a positive sequence number:
`select min(seq) from social_connection where seq < 0`

kifi-backend/modules/shoebox/app/com/keepit/model/SocialConnectionRepo.scala
raymondng, line 199: Is `20 seconds` the recommended interval? I see that `ImageInfo` sets it to `5 seconds`

raymondng, line 199: Actually based on `SequencingPlugin` `5 seconds` is the default. I assume we'll simply use that and override only in the exception cases. Please confirm.

leogrim, line 199: @ymatsuda please correct if I'm wrong:
yes let's stick to the default `5 seconds` for minimal latency, if that's too much load we can always slow it down in the future.

PR #12410, opened 2014-11-17
ignore sensitivity for keeper

yingjieMiao:
@2is10 

I chose to overwrite `sensitive` field in the controller. In the future, if we ignore `sensitive` in the extension, we can remove the `sensitive` field from the `KeeperPageInfo` object.


2is10:
:+1:

PR #12412, opened 2014-11-17
better (randomized) recommendation reaper

stkem:
This should avoid the load spikes without a loooong transaction, any Thread.sleeps or the need to keep state.
Statistically speaking every user should be cleaned at leas once every few days (and generally much more frequently than that), which is going to be plenty.
@yingjieMiao @joshm1 @andrewconner


yingjieMiao:
:+1: 

we need to adjust these as user base grows. (Or, you can make the update frequency or batchsize as a function of # of users) So that in the future you can just change the userSize number, instead of computing some statistics again.

PR #12413, opened 2014-11-17
changedURI use defer seq

yingjieMiao:
@leogrim @raymondng

### DIFF
--- a/kifi-backend/modules/shoebox/test/com/keepit/integrity/OrphanCleanerTest.scala
+++ b/kifi-backend/modules/shoebox/test/com/keepit/integrity/OrphanCleanerTest.scala
@@ -106,6 +106,8 @@ class OrphanCleanerTest extends Specification with ShoeboxTestInjector {
         db.readWrite { implicit session =>
           uris.drop(1).map { uri => changedURIRepo.save(ChangedURI(oldUriId = uri.id.get, newUriId = uris(0).id.get, state = ChangedURIStates.APPLIED)) }
         }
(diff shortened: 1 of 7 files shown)


leogrim:
:+1:

kifi-backend/modules/shoebox/test/com/keepit/integrity/OrphanCleanerTest.scala
```diff
@@ -106,6 +106,8 @@ class OrphanCleanerTest extends Specification with ShoeboxTestInjector {
         db.readWrite { implicit session =>
           uris.drop(1).map { uri => changedURIRepo.save(ChangedURI(oldUriId = uri.id.get, newUriId = uris(0).id.get, state = ChangedURIStates.APPLIED)) }
         }
+        val seqAssigner = inject[ChangedURISeqAssigner]
+        seqAssigner.assignSequenceNumbers()
```
ahsu1230, line 110: Sorry, a bit unclear to me. What does assigning sequence numbers in tests do?

leogrim, line 110: Previously, the sequence number would be assigned immediately on `save`. Now if your test does rely on sequence numbers, you want to make sure they are indeed assigned by the assigner (vs hoping for the asynchronous plugin to do it).

ahsu1230, line 110: I see thanks!

PR #12415, opened 2014-11-17
AdminLibraryCommander.index

leogrim:


andrewconner:
Looks like this one is breaking the admin page. `AdminLibraryController.scala:117`, perhaps not every library has a `statsByLibraryId` entry? @leogrim

andrewconner:
See: https://admin.kifi.com/admin/libraries

PR #12418, opened 2014-11-17
making kfEllipsis render plain text instead of HTML

2is10:
This fixes library names in the left column.

The only other field this directive seems to be used on currently, other than library titles, is keep card descriptions (`keep.summary.description`). I think we should be treating those as plain text, too, and not as HTML. They primarily come from embedly descriptions, which generally come from the plain-text description meta tag or plain-text open graph description tag (see the embedly [extract API documentation](http://embed.ly/docs/api/extract/endpoints/1/extract)).

@ahsu1230 @yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/common/directives/ellipsis/ellipsis.js
+++ b/kifi-backend/modules/shoebox/angular/src/common/directives/ellipsis/ellipsis.js
@@ -3,11 +3,9 @@
  *  inputs:
  *  (required) ng-bind
  *  (optional) ellipsisAppend - string to append at end of truncated text
- *  (optional) ellipsisSymbol - string to use as ellipsis - default is '...'
  *
  *  how to use:
  *  <p data-ellipsis data-ng-bind="longName"></p>
- *  <p data-ellipsis data-ng-bind="longName" data-ellipsis-symbol="---"></p>
  *  <p data-ellipsis data-ng-bind="longName" data-ellipsis-append="read more"></p>
  *
  */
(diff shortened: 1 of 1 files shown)


andrewconner:
Awesome, thank Jared.

yipingliao:
Thanks!

ahsu1230:
thanks Jared!

kifi-backend/modules/shoebox/angular/src/common/directives/ellipsis/ellipsis.js
```diff
@@ -41,32 +39,29 @@ angular.module('kifi')
         var lastWindowWidth = 0;
 
         function buildEllipsis() {
-          if (typeof(scope.fullText) !== 'undefined') {
+          if (typeof scope.fullText === 'string') {
```
2is10, line 42: Note: [`typeof`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof) is a built-in unary operator like `void` or `+` or `-` or `!`. Parentheses aren’t normally used with unary operators.

kifi-backend/modules/shoebox/angular/src/common/directives/ellipsis/ellipsis.js
```diff
@@ -41,32 +39,29 @@ angular.module('kifi')
...
-              element.html(scope.fullText);
+              element.text(scope.fullText);
               return;
             }
 
-            var appendString =
-              (typeof(scope.ellipsisAppend) !== 'undefined' && scope.ellipsisAppend !== '') ?
-                scope.ellipsisAppend :
-                '<span>&hellip;</span>';
+            var appendString = typeof scope.ellipsisAppend === 'string' && scope.ellipsisAppend || '\u2026';
```
2is10, line 64: The new code here uses `scope.ellipsisAppend` if it’s a non-empty string, as before, or else an ellipsis. `null` and other non-string, non-`undefined` values will now be ignored instead of being coerced to a string. We’re now using the ellipsis Unicode character directly instead of the HTML character entity and not inserting a `<span>` because we’re interpreting the string as plain text now, so no tags are allowed. I checked all uses of `kf-ellipsis` and didn’t see any purpose for the span. Please let me know if I missed something.

ahsu1230, line 64: Yup, you're right. Now that we're using plain text, the span should not be used anymore.
Thanks for this change, it's much cleaner!

PR #12420, opened 2014-11-17
deference Seq num for InvitationRepo

ahsu1230:
@leogrim @andrewconner @yingjieMiao


leogrim:
:+1: make sure to check the database once it's in deployed

PR #12421, opened 2014-11-17
fixing integration test

ahsu1230:
@raymondng @stkem @2is10 @andrewconner 

misleading error - wasn't really a JSON parsing error
seems like the export test just had an extra line that was recently added :fried_shrimp:


kifi-backend/modules/shoebox/app/com/keepit/model/InvitationRepo.scala
```diff
@@ -177,8 +177,6 @@ trait InvitationSequencingPlugin extends SequencingPlugin
 class InvitationSequencingPluginImpl @Inject() (
     override val actor: ActorInstance[InvitationSequencingActor],
     override val scheduling: SchedulingProperties) extends InvitationSequencingPlugin {
-
-  override val interval: FiniteDuration = 20 seconds
```
ahsu1230, line 181: leaving out interval override until needed

2is10:
thanks for fixing

raymondng:
thanks!

PR #12429, opened 2014-11-17
uri candidate cache

joshm1:
@stkem


kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
```diff
@@ -252,4 +243,34 @@ class RecommendationGenerationCommander @Inject() (
...
+
+    candidatesF map { candidates =>
+      val notFromCacheResults = (notFromCache zip candidates).map {
+        case (id, bool) =>
+          candidateUriCache.put(id, bool)
+          (id, bool)
+      }.toMap
+
+      // puts the candidate booleans back in the order they were passed in
+      val initial = IndexedSeq.fill(uriIds.size)(false)
```
stkem: I think an imperative approach might be better here, i.e. just make an array and fill that in a loop. Should be quite a bit faster too.

stkem:

PR #12430, opened 2014-11-17
ignoring page sensitivity and removing the sensitivity setting

2is10:
cc @yingjieMiao @stkem


yingjieMiao:
:+1: 

I assume we can remove the `sensitive` field from the object now?

PR #12431, opened 2014-11-17
no longer checking whether a page is sensitive on every page view

2is10:
Also removing unused `isSensitiveURI` internal shoebox endpoint and `NOT_SENSITIVE` experiment.
@yingjieMiao


yingjieMiao:
Cool. Will look into that.

PR #12432, opened 2014-11-17
SQL for count

raymondng:
Rather direct translation into SQL -- explain shows a clear difference in terms of execution plan.

If needed we can also add an efficient bulk version -- similar to what we did for helprank queries using jdbc, we can likely re-use it here.

@ymatsuda


raymondng:
As the bookmark count cache is no longer frequently accessed, adjusted ttl per discussion with @ymatsuda

PR #12435, opened 2014-11-17
user repo use defer seq

yingjieMiao:
@raymondng @ahsu1230


raymondng:
:+1:

kifi-backend/modules/common/conf/evolutions/shoebox/114.sql
```diff
@@ -9,7 +9,7 @@
 CREATE SEQUENCE user_sequence;
 ---
 
-ALTER TABLE user add seq Int NOT NULL DEFAULT 0;
+ALTER TABLE user add seq bigint(20) NOT NULL DEFAULT 0;
```
ahsu1230, line 12: nice catch :+1:

kifi-backend/modules/common/conf/evolutions/shoebox/114.sql
```diff
@@ -9,7 +9,7 @@
...
 CREATE INDEX user_i_seq ON user(seq);
 
 INSERT INTO evolutions (name, description) VALUES ('114.sql', 'add sequence number to user');
```
ahsu1230, line 15: maybe a newline after this line?

yingjieMiao, line 15: oh, good catch. It was an old 'bug'.

PR #12436, opened 2014-11-18
UserCommander.getFriendRecommendations

ymatsuda:
@eishay @ahsu1230 @andrewconner @leogrim 
bulk version of getConnectedUsers


kifi-backend/modules/shoebox/app/com/keepit/model/UserConnectionRepo.scala
```diff
@@ -99,6 +100,21 @@ class UserConnectionRepoImpl @Inject() (
...
+      val query =
+        ((for (c <- rows if c.user1.inSet(missing) && c.state === UserConnectionStates.ACTIVE) yield (c.user1, c.user2)) union
+          (for (c <- rows if c.user2.inSet(missing) && c.state === UserConnectionStates.ACTIVE) yield (c.user2, c.user1)))
+
+      query.list().groupBy(_._1).map {
+        case (id, connections) => UserConnectionIdKey(id) -> connections.map(_._2.id).toArray
+      }.toMap
+    }
+    ret.map { case (key, value) => key.userId -> value.map(Id[User]).toSet }.toMap
+  }
```
ahsu1230, line 116: :+1: thanks, this is a great example to use for bulk GETs

PR #12437, opened 2014-11-18
KeepRepo: deferred seq

raymondng:
standard stuff except for the interactive keep we stick with the old style to avoid latency

@ymatsuda @leogrim please take a look


raymondng:
Reviewed with @ymatsuda in person. Will monitor.

PR #12439, opened 2014-11-18
optimizing UserController.friends

ahsu1230:
https://app.asana.com/0/20314588032618/20153720309488
- calling getConnectionCount in map in a transaction is slow
- need getConnectionCounts(ids: Set[Id[User])
  - use bulkGetOrElse to access cache
  - use IN predicates and SELECT GROUP BY to compute counts for muiltple users in a single DB call.

@ymatsuda @andrewconner @stkem

### DIFF
--- a/integration/site-keeps-export_spec.js
+++ b/integration/site-keeps-export_spec.js
@@ -9,16 +9,15 @@ Do Not Edit! -->\n\
 <Title>Kifi Bookmarks Export</Title>\n\
 <H1>Bookmarks</H1>\n\
 <DL>\n\
-<DT><A HREF="https://www.kifi.com/integrator-aa/test-2" ADD_DATE="1416253967">Test 2 by Integrator AA • Kifi</A>\n\
-<DT><A HREF="https://www.kifi.com/friends/invite" ADD_DATE="1403807124" TAGS="kifi Support">kifi • Find friends your friends on kifi</A>\n\
-<DT><A HREF="http://support.kifi.com/customer/portal/emails/new" ADD_DATE="1403807125" TAGS="kifi Support">kifi • Contact Us</A>\n\
-<DT><A HREF="http://support.kifi.com/customer/portal/articles/1397866-introduction-to-kifi-" ADD_DATE="1403807126" TAGS="kifi Support">kifi • How to Use kifi</A>\n\
-<DT><A HREF="https://www.kifi.com/install" ADD_DATE="1403807127" TAGS="kifi Support">kifi • Install kifi on Firefox and Chrome</A>\n\
-<DT><A HREF="http://www.youtube.com/watch?v=_OBlgSz8sSM" ADD_DATE="1403807128" TAGS="Example Keep,Funny">Charlie bit my finger - again ! - YouTube</A>\n\
-<DT><A HREF="http://twistedsifter.com/2013/01/50-life-hacks-to-simplify-your-world/" ADD_DATE="1403807129" TAGS="Example Keep,Read Later">50 Life Hacks to Simplify your World «TwistedSifter</A>\n\
-<DT><A HREF="https://www.airbnb.com/locations/san-francisco/mission-district" ADD_DATE="1403807130" TAGS="Example Keep,Travel">Mission District, San Francisco Guide - Airbnb Neighborhoods</A>\n\
-<DT><A HREF="http://www.amazon.com/Hitchhikers-Guide-Galaxy-25th-Anniversary/dp/1400052920/" ADD_DATE="1403807131" TAGS="Example Keep,Shopping Wishlist">The Hitchhiker\'s Guide to the Galaxy, 25th Anniversary Edition: Douglas Adams: 9781400052929: Amazon.com: Books</A>\n\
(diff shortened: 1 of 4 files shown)


integration/site-keeps-export_spec.js
```diff
@@ -9,16 +9,15 @@ Do Not Edit! -->\n\
 <Title>Kifi Bookmarks Export</Title>\n\
 <H1>Bookmarks</H1>\n\
 <DL>\n\
-<DT><A HREF="https://www.kifi.com/integrator-aa/test-2" ADD_DATE="1416253967">Test 2 by Integrator AA • Kifi</A>\n\
```
ahsu1230: fix for integration test. Uses userB instead of userA

kifi-backend/modules/shoebox/test/com/keepit/controllers/website/UserControllerTest.scala
```diff
@@ -260,6 +260,37 @@ class UserControllerTest extends Specification with ShoeboxTestInjector {
...
+        }
+        val userController = inject[UserController]
+
+        inject[FakeUserActionsHelper].setUser(userGW)
+        val request1 = FakeRequest("GET", routes.UserController.friends().url)
+        val result1: Future[Result] = userController.friends(0, 5)(request1)
+        status(result1) must equalTo(OK)
+        contentType(result1) must beSome("application/json")
+
+        val resultBody = contentAsString(result1)
```
andrewconner: You can do `contentAsJson`

kifi-backend/modules/shoebox/test/com/keepit/controllers/website/UserControllerTest.scala
```diff
@@ -260,6 +260,37 @@ class UserControllerTest extends Specification with ShoeboxTestInjector {
       }
     }
 
+    "get friends" in {
```
andrewconner, line 262: :+1:

kifi-backend/modules/shoebox/app/com/keepit/model/UserConnectionRepo.scala
```diff
@@ -74,6 +75,18 @@ class UserConnectionRepoImpl @Inject() (
       } yield c).length).first
     }
   }
+  def getConnectionCounts(userIds: Set[Id[User]])(implicit session: RSession): Map[Id[User], Int] = {
+    val ret = connCountCache.bulkGetOrElse(userIds map UserConnectionCountKey) { keys =>
+      val missingIds = keys.map(_.userId)
+      val query =
+        ((for (c <- rows if c.user1.inSet(missingIds) && c.state === UserConnectionStates.ACTIVE) yield (c.user1, 1)) union
+          (for (c <- rows if c.user2.inSet(missingIds) && c.state === UserConnectionStates.ACTIVE) yield (c.user2, 1)))
+      query.list().groupBy(_._1).map {
```
ymatsuda: This transfers all connections from MySQL to Shoebox as tuples. Counting happens in Scala.
Why not using SQL directly?

kifi-backend/modules/shoebox/app/com/keepit/model/UserConnectionRepo.scala
```diff
@@ -75,6 +76,18 @@ class UserConnectionRepoImpl @Inject() (
       } yield c).length).first
     }
   }
+  def getConnectionCounts(userIds: Set[Id[User]])(implicit session: RSession): Map[Id[User], Int] = {
+    val ret = connCountCache.bulkGetOrElse(userIds map UserConnectionCountKey) { keys =>
+      val missingIds = keys.map(_.userId)
+      val query =
+        ((for (c <- rows if c.user1.inSet(missingIds) && c.state === UserConnectionStates.ACTIVE) yield (c.user1, 1)) union
```
ahsu1230: Thanks for the help @leogrim 
@ymatsuda @andrewconner 
I was deciding between using this slick query vs. a direct SQL query. But a SQL query would have us use `Set.mkString(",")` for the IN predicates. What do you think?

ahsu1230: I'm also using a tuple `(c.user, 1)` so `groupBy` can aggregate. Would there be a better way to do this aggregation (maybe without having to create a tuple pair per yielded entry)?

kifi-backend/modules/shoebox/test/com/keepit/controllers/website/UserControllerTest.scala
ahsu1230: Sorry, I'm not sure I follow. The result1 would return a `{ friends : [ {...}, {...} ], total : 3 }`. Each of those objects has like 8 fields and there isn't really a Read formatter. Is there something that could easily parse a field out of a `JsArray`?

andrewconner: Yep. You can use `\\` for recursive paths. https://www.playframework.com/documentation/2.3.x/ScalaJson

ahsu1230: Ah right, forgot that you could just bulk cast as a specific object

ahsu1230: thanks!

PR #12443, opened 2014-11-18
Don't barf at non-fully initialized social users (simply reject) in websocket connections

stkem:
A live mobile client is currently doing that which has been regularly triggering pagerduties.

@andrewconner please review carefully


stkem:
Works in dev mode

PR #12445, opened 2014-11-18
Easiest way to understand it is to remove code.

andrewconner:
@stkem @leogrim


stkem:
lgtm. last nail in the coffin of page peeker (or our usage thereof), I guess :-)

PR #12447, opened 2014-11-18
use deferred for updates

raymondng:
Per discussion with @ymatsuda 

cc: @leogrim @yingjieMiao


yingjieMiao:
:+1:

kifi-backend/modules/shoebox/test/com/keepit/integrity/OrphanCleanerTest.scala
```diff
@@ -19,6 +21,14 @@ class OrphanCleanerTest extends Specification with ShoeboxTestInjector {
     FakeKeepImportsModule()
   )
 
+  @inline def assignSeqNums(db: Database)(repos: SeqNumberFunction[_]*): Unit = {
```
raymondng, line 24: @yingjieMiao I've added a helper method here to deal with sequence numbers. Please take a look.

PR #12448, opened 2014-11-18
introduce a new cache helper method

yingjieMiao:
@leogrim @stkem 

I only applied this to one of the graph caches. Another change comes next, but that one requires many more API changes.

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/common/cache/ObjectCache.scala
+++ b/kifi-backend/modules/common/app/com/keepit/common/cache/ObjectCache.scala
@@ -1,5 +1,7 @@
 package com.keepit.common.cache
 
+import org.joda.time.DateTime
+import com.keepit.common.time._
 import scala.concurrent._
 import scala.concurrent.duration._
 import play.api.libs.concurrent.Execution.Implicits._
@@ -88,6 +90,24 @@ trait ObjectCache[K <: Key[T], T] {
         val valueOptFuture = orElse
         valueOptFuture.onSuccess { case valueOpt => set(key, valueOpt) }
         valueOptFuture
+    }
+  }
+
+  def getOldAndAsyncRefresh(key: K, bePatient: Boolean, freshInterval: Long)(getTimeStamp: T => DateTime)(refresh: => Future[Option[T]]): Future[Option[T]] = {
+    get(key) match {
+      case Some(value) =>
+        if (getTimeStamp(value).plus(freshInterval) < currentDateTime) {
+          refresh.onSuccess { case valueOpt => set(key, valueOpt) }
+        }
+        Future.successful(Some(value))
+      case None =>
+        if (bePatient) {
+          val valueOptF = refresh
+          valueOptF.onSuccess { case valueOpt => set(key, valueOpt) }
+          valueOptF
+        } else {
+          Future.successful(None)
+        }
     }
   }
 
--- a/kifi-backend/modules/common/app/com/keepit/common/cache/ObjectCache.scala
+++ b/kifi-backend/modules/common/app/com/keepit/common/cache/ObjectCache.scala
@@ -73,9 +73,13 @@ trait ObjectCache[K <: Key[T], T] {
     }
   }
 
-  def getOrElseFuture(key: K)(orElse: => Future[T]): Future[T] = {
+  def getOrElseFuture(key: K, predicate: T => Boolean = Function.const(true))(orElse: => Future[T]): Future[T] = {
     get(key) match {
-      case Some(value) => Promise.successful(value).future
+      case Some(value) =>
+        if (!predicate(value)) {
+          orElse.onSuccess { case value => set(key, value) }
+        }
+        Promise.successful(value).future
       case None =>
         val valueFuture = orElse
         valueFuture.onSuccess { case value => set(key, value) }
@@ -83,9 +87,13 @@ trait ObjectCache[K <: Key[T], T] {
     }
   }
 
-  def getOrElseFutureOpt(key: K)(orElse: => Future[Option[T]]): Future[Option[T]] = {
+  def getOrElseFutureOpt(key: K, predicate: Option[T] => Boolean = Function.const(true))(orElse: => Future[Option[T]]): Future[Option[T]] = {
     internalGet(key) match {
-      case Found(valueOpt) => Promise.successful(valueOpt).future
+      case Found(valueOpt) =>
+        if (!predicate(valueOpt)) {
+          orElse.onSuccess { case valOpt => set(key, valOpt) }
+        }
+        Future.successful(valueOpt)
       case _ =>
         val valueOptFuture = orElse
         valueOptFuture.onSuccess { case valueOpt => set(key, valueOpt) }
--- a/kifi-backend/modules/common/app/com/keepit/common/cache/ObjectCache.scala
+++ b/kifi-backend/modules/common/app/com/keepit/common/cache/ObjectCache.scala
@@ -101,24 +101,6 @@ trait ObjectCache[K <: Key[T], T] {
     }
   }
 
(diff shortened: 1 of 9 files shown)


kifi-backend/modules/common/app/com/keepit/common/cache/ObjectCache.scala
```diff
@@ -91,6 +93,24 @@ trait ObjectCache[K <: Key[T], T] {
...
+        if (bePatient) {
+          val valueOptF = refresh
+          valueOptF.onSuccess { case valueOpt => set(key, valueOpt) }
+          valueOptF
+        } else {
+          Future.successful(None)
+        }
+    }
+  }
+
```
leogrim: As discussed with @stkem a while back, `bePatient` isn't a good pattern since the behavior of the code depends on the presence / absence of the cache.
As for async updates, I think the best generic approach would be to modify `getOrElse` and `getOrElseFuture` to take an additional `andThenInvalidate: T => Boolean` parameter that would default to false. 
@andrewconner @eishay feel free to chip in :)

yingjieMiao: For method names/arguments, there is a trade off between 'being generic' and 'being clear and specific'.

For example, we could let the `cache` has only 1 method, and takes various flags, optional parameters. This would necessarily complicate the code inside the method. Another reason to create a new endpoint is that I see this as a frequent pattern (e.g. check time stamp), so I think it deserves a separate endpoint. 

Also, we don't have to over-design or over-generic when there is no clear usage yet. As of now in the code base, this pattern is only used with timestamp.

yingjieMiao:
@raymondng  you might be interested

kifi-backend/modules/common/app/com/keepit/common/cache/ObjectCache.scala
leogrim: More problematic in my opinion is the risk to see the `orElse` blocks diverge if we have several APIs. This is a concern I encountered when I refactored `Typeahead` with this async update pattern. Having a single API will ensure that the block of code used to populate the cache is uniquely defined. Since Scala makes it easy to supply default parameters, this would be an invisible addition in existing use cases.

yingjieMiao: BTW, the `bePatient` thing is to support the current method signature in graphClient. Looks like AbookClient uses this flag. I am not familiar with the history of these endpoints. If we only set the flag = true or only set the flag = false, then we need to remove the flag from related methods.

kifi-backend/modules/common/app/com/keepit/common/cache/ObjectCache.scala
```diff
@@ -91,6 +93,24 @@ trait ObjectCache[K <: Key[T], T] {
     }
   }
 
+  def getOldAndAsyncRefresh(key: K, bePatient: Boolean, freshInterval: Long)(getTimeStamp: T => DateTime)(refresh: => Future[Option[T]]): Future[Option[T]] = {
+    get(key) match {
+      case Some(value) =>
+        if (getTimeStamp(value).plus(freshInterval) < currentDateTime) {
```
stkem: I wonder if the cache itself actually lets us get a timestamp (It obviously has it, otherwise it couldn't expire things). That way we could avoid having the actual model be augmented for this functionality.

kifi-backend/modules/common/app/com/keepit/common/cache/ObjectCache.scala
stkem: If that is needed, then we should do that. Having a `bePatient` basically turns the cache into a communication channel, instead of a storage mechanism, which is a bad pattern.

kifi-backend/modules/common/app/com/keepit/graph/GraphServiceClient.scala
```diff
@@ -124,20 +124,11 @@ class GraphServiceClientImpl @Inject() (
     futureUserScores.map(userScores => userScores.map { case ConnectedUserScore(friendId, score) => friendId -> score })
   }
 
-  private val consolidateSociallyRelatedEntities = new RequestConsolidator[Id[User], Unit](10 seconds)
-  def refreshSociallyRelatedEntities(userId: Id[User]): Future[Unit] = consolidateSociallyRelatedEntities(userId) { id =>
-    call(Graph.internal.refreshSociallyRelatedEntities(id), callTimeouts = longTimeout).map(_ => ())
-  }
-
   def getSociallyRelatedEntities(userId: Id[User], bePatient: Boolean): Future[Option[SociallyRelatedEntities]] = {
```
stkem, line 128: The `bePatient` here slipped through the cracks earlier, I don't think it is a good pattern. (We need this data, so we should always wait)

kifi-backend/modules/common/app/com/keepit/common/cache/ObjectCache.scala
raymondng: Agreed. That's the approach we took with `ObjMetadata` -- it's best to get it from the source (cache or store).

kifi-backend/modules/common/app/com/keepit/common/cache/ObjectCache.scala
stkem: Actually, never mind on the last point. This does return the result from the actual call, so the communication channel is still the call. However, it should ideally not fail if there is a problem with cache, i.e. the call should go through and return data even when the cache fails (for whatever reason).

yingjieMiao:
@leogrim @stkem  updated.

kifi-backend/modules/common/app/com/keepit/common/cache/ObjectCache.scala
```diff
@@ -71,9 +73,13 @@ trait ObjectCache[K <: Key[T], T] {
     }
   }
 
-  def getOrElseFuture(key: K)(orElse: => Future[T]): Future[T] = {
+  def getOrElseFuture(key: K, predicate: T => Boolean = Function.const(true))(orElse: => Future[T]): Future[T] = {
```
stkem: how about calling it `refresh` and reversing the logic? Makes it easier to understand imho.

kifi-backend/modules/common/app/com/keepit/common/cache/ObjectCache.scala
```diff
@@ -81,9 +87,13 @@ trait ObjectCache[K <: Key[T], T] {
     }
   }
 
-  def getOrElseFutureOpt(key: K)(orElse: => Future[Option[T]]): Future[Option[T]] = {
+  def getOrElseFutureOpt(key: K, predicate: Option[T] => Boolean = Function.const(true))(orElse: => Future[Option[T]]): Future[Option[T]] = {
```
stkem: I think we should also add it to the non-future methods (`getOrElse`, `getOrElseOpt`) to keep the api complete (and keep others from hacking something when they need the functionality there)

leogrim: Agree.

kifi-backend/modules/common/app/com/keepit/common/cache/ObjectCache.scala
leogrim: +1

yingjieMiao: OK with that.

kifi-backend/modules/common/app/com/keepit/common/cache/ObjectCache.scala
yingjieMiao: In some sense, I avoided early-design (or early-generalization). This sometimes just introduce unused code. 

I am fine with adding support to the other APIs, this would necessarily introduce `SafeFuture` inside the block.

yingjieMiao:
@leogrim @stkem  updated.

stkem:
lgtm

PR #12449, opened 2014-11-18
apparently naming our Firefox extension icons magically gets them used more

2is10:
(e.g. in the warning dialog at installation time)


packrat/adapters/firefox/package.json
```diff
@@ -7,8 +7,6 @@
   "homepage": "https://www.kifi.com/",
   "version": "see build.properties",
   "id": "kifi@42go.com",
-  "icon": "data/icons/kifi.48.png",
-  "icon64": "data/icons/kifi.64.png",
```
2is10, line 11: These were getting the icons shown in about:addons but not in the installation prompt/warning.

PR #12450, opened 2014-11-18
Add integration tests for all Angular routes

yipingliao:
Also:
- Add checks for JS errors
- Use multiple instances of Chrome to run Protractor tests

@andrewconner @ahsu1230


kifi-backend/modules/shoebox/angular/test/protractor.conf.js
```diff
@@ -1,7 +1,9 @@
 module.exports.config = {
   // Use phantomjs for headless testing.
   capabilities: {
-    browserName: 'chrome'
+    browserName: 'chrome',
+    shardTestFiles: true,
+    maxInstances: 3
```
yipingliao, line 6: When I used a single instance, the time it took to run the tests was 12 minutes. Two instances got me down to 7 minutes. Three instances reduced that further to 6.67 minutes, so it seems like any more instances past 3 will be only diminishing returns. After merging, I'll check what the running is on Jenkins. FYI, currently it takes 1 minutes to run the tests on Jenkins, but now the number of tests has increased from 4 to 47.

kifi-backend/modules/shoebox/angular/src/tagManage/tagManage.js
```diff
@@ -6,8 +6,9 @@ angular.module('kifi')
             'routeService', '$http', '$location', 'modalService', '$timeout', '$rootScope',
   function (tagService, $scope, $window, manageTagService, libraryService,
               routeService, $http, $location, modalService, $timeout, $rootScope) {
-    $scope.selected = {};
+    $window.document.title = 'Kifi • Manage Your Tags';
```
yipingliao, line 9: Previously this page had no set title.

kifi-backend/modules/shoebox/angular/test/e2e/find.spec.js
```diff
@@ -0,0 +1,63 @@
...
+  // Helper functions.
+  //
+  function loginUserWithCookies() {
+    // Set cookies to authenticate user.
+    // See: https://github.com/kifi/keepit/blob/master/integration/auth_headers.js
+    browser.manage().addCookie('KIFI_SECURESOCIAL', '567b766f-cae2-4bb9-b8f2-02e3ecb5cedd', '/', '.kifi.com');
+    browser.manage().addCookie('KIFI_SESSION', 'bd4cee3dd5d4170c702ff599302e223a900a8c4d-fortytwo_user_id=8397', '/', '.kifi.com');
+  }
+
+  function checkJSErrors() {
```
yipingliao: This function is now run after every test. It checks for any console messages that contains our Kifi minified JS file. I have tested this with a few example failures, including that bug from last week when the code tried to call `split` on `undefined`.

kifi-backend/modules/shoebox/angular/src/tagManage/tagManage.js
ahsu1230, line 9: nice catch ty!

andrewconner:
Nice! How about pulling `checkJSErrors` to a separate file? Any way to require it?

andrewconner:
Basically all of "Helper functions". Keeping tests small and easy to read will really help us.

kifi-backend/modules/shoebox/angular/test/protractor.conf.js
andrewconner, line 6: Hmm. We need to make this as fast as possible. Curious what takes so long. Did you find a way to recycle browser instances?

andrewconner, line 6: I also noticed previously that it waits until every `timeout` and promise is resolved. Perhaps we have some timeouts that are artificially delaying things for every load for the tests?

yipingliao, line 6: I'll look into this some more, but it seems like navigating to a page before each test is the standard way of doing things and I have not seen any examples otherwise. Will also look into the `timeout` part.

yipingliao:
Yup, will also look into DRYing up all helper functions.

kifi-backend/modules/shoebox/angular/test/protractor.conf.js
andrewconner, line 6: I was trying to help, but protractor is crashing locally for me. Sigh.

Two things to check out: 
1) https://docs.angularjs.org/api/ngMock/service/$timeout has a `$timeout.flush`, perhaps we should use this?
2) We may be able to trim some of the timeouts, or not run them during tests. I know of one that checks if there's a social import going on after a few seconds.

kifi-backend/modules/shoebox/angular/test/e2e/logged-out-library.spec.js
```diff
@@ -0,0 +1,62 @@
...
+        console.log('log: ' + require('util').inspect(kifiJsErrors));
+      }
+    });
+  }
+
+
+  //
+  // Test cases.
+  //
+  describe('Kifi library page for logged-out users', function () {
```
ahsu1230, line 6: how about also testing users can't see private libraries while logged out?
I guess you could also do the same for main/secret system libraries

yipingliao, line 6: I thought about it, but didn't want to make too many tests since they are sooo slow.
Looks like they are too slow anyway! :) Will add these cases.

andrewconner, line 6: > how about also testing users can't see private libraries while logged out?

Not sure if the frontend should test this. Perhaps it could be a test to see if the right error page shows, but this is really determined on the backend. We don't trust the frontend to do the right thing, from a data security point of view.

andrewconner, line 6: Also, better suited for a karma test.

ahsu1230, line 6: yea I was thinking testing the error shows up or maybe making sure no keeps are displayed.

yipingliao:
@andrewconner I DRY'ed up the helper functions but have not gotten anywhere with speeding things up. :( Setting `browser.ignoreSynchronization = true` results in tests breaking sometimes even when we shouldn't be waiting for any timeouts or http for the test (like looking for title). I also tried ignoring synchronization and then using `browser.wait()` to wait for the specific element that I am looking for. I had hoped that this means that Protractor will not wait for any irrelevant timeouts and http requests. The result is that the test seem to always pass with an explicit `browser.wait()`, but it takes just as long (in fact, longer) as when we are waiting for synchronization. I am a little bit out of ideas.

yipingliao:
Just added a commit where I:
- "Commented out" tests for "/friends", "/invite", "/keep/:keepid", "/tags/manage", and "/profile". By "commenting out" I meant renaming the test files so that they will not grabbed by the protractor gulp task.
- Manually parallelized "library.spec.js" so that the different test cases inside can run on up to 3 browser instances at the same time ("library.spec.js" was split into 6 separate files).

With these improvements, we're down to 3.6 minutes. :)

andrewconner:
Wow. Hm, that's crazy — based on timers I did, it seems like most of the time is spent doing nothing.

We may need to trim it back some more, and perhaps selectively run them. But this shouldn't be an issue — I'll spend some time on it.

yipingliao:
Down to 2 minutes now after truncating more tests. Gonna merge!

ahsu1230:
yahoo!

PR #12453, opened 2014-11-18
adding users with recent extension into A/B experiment

2is10:
@stkem


kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtAuthController.scala
```diff
@@ -83,6 +84,12 @@ class ExtAuthController @Inject() (
 
     if (isUpdate || isInstall) {
       SafeFuture {
+        if (version >= KifiExtVersion(3, 3, 18) && !request.experiments.contains(ExperimentType.VISITED)) {
+          db.readWrite { implicit s =>
```
stkem: Idiomatic would be to use `LocalUserExperimentCommander.addExperimentForUser` instead of going directly to the db (although that call does exactly that).

stkem:

2is10:
thanks

PR #12454, opened 2014-11-18
prevent users who already have a membership to a library to be invited (...

ahsu1230:
...backend check)

https://app.asana.com/0/19898402862947/20521229605002/f
Not sure how it happened since our site has a guard to see if a user already following a library. 
But this backend guard prevents an invite from being created if the invitee user already has an active membership to the library.

@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
@@ -735,19 +735,27 @@ class LibraryCommander @Inject() (
       }
 
       futureInvitedContactsByEmailAddress.map { invitedContactsByEmailAddress =>
-        val invitesAndInvitees = for (i <- inviteList) yield {
-          val (recipient, inviteAccess, msgOpt) = i
-          // TODO (aaron): if non-owners invite that's not READ_ONLY, we need to change API to present "partial" failures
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -735,19 +735,27 @@ class LibraryCommander @Inject() (
...
-              (LibraryInvite(libraryId = libraryId, inviterId = inviterId, userId = Some(id), access = access, message = msgOpt), Left(invitedBasicUsersById(id)))
-            case Right(email) =>
-              (LibraryInvite(libraryId = libraryId, inviterId = inviterId, emailAddress = Some(email), access = access, message = msgOpt), Right(invitedContactsByEmailAddress(email)))
+        val invitesAndInvitees = db.readOnlyMaster { implicit s =>
+          for ((recipient, inviteAccess, msgOpt) <- inviteList) yield {
+            val access = if (targetLib.ownerId != inviterId) LibraryAccess.READ_ONLY else inviteAccess
+            recipient match {
+              case Left(userId) =>
+                libraryMembershipRepo.getWithLibraryIdAndUserId(libraryId, userId) match {
+                  case Some(mem) if mem.access == access =>
```
ahsu1230, line 744: let's say an inviter invites me for first READ_ONLY then READ_WRITE. Then we should be creating a new invite in the Database. It's only if an inviter invites me for READ_ONLY access that it's not okay. But since all invites are all READ_ONLY anyway, this check should be suffice.

PR #12455, opened 2014-11-18
Remove unused Angular routes.

yipingliao:
@andrewconner @ahsu1230 

When writing the protractor tests, I noticed that these old routes were still hanging around.


ahsu1230:
thanks! confirmed they're unused lol

PR #12457, opened 2014-11-18
marketing site

joshm1:
@stkem


stkem:
why are we doing this?

joshm1:
because the old landing page is blank and goes nowhere after the redirect, this at least give the user something to see if they navigate back to the page or click not to open the app (as it does in facebook's embedded browser)

we can discuss with jen

stkem:
ok, fine with me

stkem:
code looks good

PR #12464, opened 2014-11-19
update the library membership lastEmailSent for digest emails

joshm1:
@stkem

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/common/routes/Router.scala
+++ b/kifi-backend/modules/common/app/com/keepit/common/routes/Router.scala
@@ -167,7 +167,7 @@ object Shoebox extends Service {
     def getLibrariesChanged(seqNum: SequenceNumber[Library], fetchSize: Int) = ServiceRoute(GET, "/internal/shoebox/database/getLibrariesChanged", Param("seqNum", seqNum), Param("fetchSize", fetchSize))
     def getLibraryMembershipsChanged(seqNum: SequenceNumber[LibraryMembership], fetchSize: Int) = ServiceRoute(GET, "/internal/shoebox/database/getLibraryMembershipsChanged", Param("seqNum", seqNum), Param("fetchSize", fetchSize))
     def canViewLibrary() = ServiceRoute(POST, "/internal/shoebox/libraries/canView")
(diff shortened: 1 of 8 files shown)


kifi-backend/modules/shoebox/test/com/keepit/commanders/NewKeepsInLibraryCommanderTest.scala
```diff
@@ -33,134 +33,31 @@ class NewKeepsInLibraryCommanderTest extends TestKitSupport with SpecificationLi
     FakeElizaServiceClientModule()
   )
 
-  def setupUsers()(implicit injector: Injector) = {
```
joshm1, line 36: move all of this setup code out so it can be used in other tests

stkem, line 36: :+1:

kifi-backend/modules/common/app/com/keepit/common/routes/Router.scala
```diff
@@ -167,7 +167,7 @@ object Shoebox extends Service {
     def getLibrariesChanged(seqNum: SequenceNumber[Library], fetchSize: Int) = ServiceRoute(GET, "/internal/shoebox/database/getLibrariesChanged", Param("seqNum", seqNum), Param("fetchSize", fetchSize))
     def getLibraryMembershipsChanged(seqNum: SequenceNumber[LibraryMembership], fetchSize: Int) = ServiceRoute(GET, "/internal/shoebox/database/getLibraryMembershipsChanged", Param("seqNum", seqNum), Param("fetchSize", fetchSize))
     def canViewLibrary() = ServiceRoute(POST, "/internal/shoebox/libraries/canView")
-    def newKeepsInLibrary(userId: Id[User], max: Int) = ServiceRoute(GET, "/internal/shoebox/database/newKeepsInLibrary", Param("userId", userId), Param("max", max))
+    def newKeepsInLibraryForEmail(userId: Id[User], max: Int) = ServiceRoute(GET, "/internal/shoebox/database/newKeepsInLibraryForEmail", Param("userId", userId), Param("max", max))
```
stkem, line 170: Unless I'm missing something here, you'll also have to update `shoebox.routes`

PR #12470, opened 2014-11-19
Add name microdata to follower pics on logged-out library pages.

yipingliao:
@eishay cc @andrewconner


kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.tpl.html
```diff
@@ -70,7 +70,7 @@
                 ng-click="hideTooltip()"
                 itemprop="photo"
                 alt="{{follower.firstName}} {{follower.lastName}}"
-              >
+              ><span itemprop="name" style="display:none">{{follower.firstName}} {{follower.lastName}}</span>
```
yipingliao, line 73: Adding the span right after the previous element's closing tag (without a newline separator) fixes our image dropping off problem.

PR #12472, opened 2014-11-20
[DO NOT MERGE] Add library attribute for users on library page and add system libraries too.

yipingliao:
@andrewconner @ahsu1230 cc @leogrim 
https://app.asana.com/0/19898402862947/19898402863094

Add the front-end part of adding library attributions. Do not merge right now, because of [this issue](https://app.asana.com/0/15523651812135/20646352218293/), - the library attributions are not always consistent with the most recent keep-to-library activities. @leogrim is working on that; if it proves too difficult on the backend, front-end can also reconcile the inconsistencies. This is an existing bug but will become much more obvious once user library attribution is displayed.

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/common/directives/keepWho/keepWho.styl
+++ b/kifi-backend/modules/shoebox/angular/src/common/directives/keepWho/keepWho.styl
@@ -23,6 +23,9 @@
   &.secret
     background-color: libraryRed
 
+  &.system
+    background-color: #000
+
 
 .kf-keep-who-lib
   font-size: 10px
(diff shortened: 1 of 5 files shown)


andrewconner:
@leogrim and I talked: we'll do this on the frontend. The history of this is that the name of the two special libraries has changed, and will change in the future (because we create them, our product needs will change over time).

Instead of updating every single system library's name whenever product needs to, we do the naming superficially when the row is pulled from the db. @raymondng and I talked about this at the time, and felt it was the best way.

What I missed is that search has a copy of this name, and so doesn't get this "superficial naming" update. Thus, incorrectly named old libraries.

It's a hard issue, because either search needs to request to shoebox about the name, OR we do it on the frontend. I think it's best to do it on the frontend since we already augment data there. @yipingliao, let me know if this makes sense.

andrewconner:
@leogrim @yipingliao What did you figure out about the data issue?

yipingliao:
@leogrim is going to look into data reconciliation on the server-side tomorrow. :)

yipingliao:
@leogrim by the way I am going to do the system library name thing on the front-end ("My Private Library" vs. "Secret Library"). I also haven't had time to do it today.

eishay:
This is a 15 days old pull request. What's going on with it?

yipingliao:
I'll sync up with Leo about this this morning.

Yiping

PR #12473, opened 2014-11-20
Sticky LH sidebar titles

ahsu1230:
For this sticky sidebar, most of the trickiness comes from figuring out what the current stick title should be based on scrollbar offset. And, if you switch between `My Libraries First` in the dropdown menu, the current scrollbar offset will dictate which title is sticky.

I wanted to use the sticky directive `kf-sticky` but I think it may not be exactly what I'm looking for. I don't use a callback function, and would have to add in more parameters like which element is the scrolly-window, z-index value, a doThisWhileScrolling function of some sort? and maybe more. The sticky part here (which is all I need), can be easily achieved just using `position: 'fixed'` and a `top` value.

Open to thoughts though =)

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
+++ b/kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
@@ -40,7 +40,6 @@ angular.module('kifi')
         //
         // Internal methods.
         //
-
         function setLibListHeight() {
           if (scrollableLibList.offset()) {
             scrollableLibList.height(w.height() - (scrollableLibList.offset().top - w[0].pageYOffset));
@@ -66,7 +65,6 @@ angular.module('kifi')
         //
         // Scope methods.
         //
-
         // Temp callout method. Remove after most users know about libraries. (Oct 26 2014)
         var calloutName = 'library_callout_shown';
         scope.showCallout = function () {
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
```diff
@@ -172,6 +183,47 @@ angular.module('kifi')
           }
         }, 100);
 
+        antiscrollLibList.bind('scroll', _.debounce(function() {
+          var lastStickTitle = scope.currentStickTitle;
+          setCurrentStickTitle();
+          if (lastStickTitle !== scope.currentStickTitle) {
+            scope.$apply();
+          }
+        }, 10));
```
ahsu1230, line 192: trying to debounce so it calculates less often. Anymore of a delay, and the switch-eroo is too lagged =/

andrewconner:
Interesting. I believe `kf-sticky` can still do this by using the same trick antiscroll uses (find the first element that _can_ scroll, stick to that one). (the callback is optional, not really needed in this case or even all that often).

I'll come over in a few and play with this one

kifi-backend/modules/shoebox/angular/src/search/search.js
```diff
@@ -58,6 +58,7 @@ angular.module('kifi')
         });
       } else {
         libraryIdPromise = $q.when('');
+        $rootScope.$emit('libraryUrl', {});
```
ahsu1230, line 61: thanks @yipingliao for the help!

PR #12477, opened 2014-11-20
Shorten regex expression for detecting 2-letter TLDs.

yipingliao:
@2is10 Thanks for your advice. This is much shorter.


2is10:
:+1:

PR #12478, opened 2014-11-20
type-safing the use of access token

eishay:
and testing


eishay:
@raymondng  @andrewconner

PR #12485, opened 2014-11-21
AutogenReap virtual lib exp handling

raymondng:
We recently added virtual library experiment type. Reaper needs to be updated to handle it.
Added a simple filter.

(Please ignore the merge history. I reverted some refactoring changes)

@andrewconner @eishay


raymondng:
Will push to reduce noise in prod. Feedback appreciated.

PR #12491, opened 2014-11-21
sticky scrollbar menu panes push off each other

ahsu1230:
(should work with filtering & toggling dropdown menu options)

@andrewconner @yipingliao


kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
```diff
@@ -62,6 +61,85 @@ angular.module('kifi')
...
+            firstLimit = scope.userLibsToShow.length * libItemHeight;
+            firstLimitOverlay = firstLimit + separatorHeight;
+            secondLimit = firstLimit + scope.invitedLibsToShow.length * libItemHeight;
+            secondLimitOverlay = secondLimit + separatorHeight;
+          } else {
+            firstLimit = scope.invitedLibsToShow.length * libItemHeight;
+            firstLimitOverlay = firstLimit + separatorHeight;
+            // no more limits needed
+          }
+          fixSeparators(offset, firstLimit, firstLimitOverlay, secondLimit, secondLimitOverlay, separatorHeight);
```
ahsu1230, line 103: had to separate to a different function to prevent cyclomatic complexity being too high

kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
```diff
@@ -88,12 +166,11 @@ angular.module('kifi')
           return loc === path || util.startsWith(loc, path + '/');
         };
 
-        scope.padList = function () {
-          if (scope.sortingMenu.myLibsFirst) {
-            return scope.myLibsToShow.length > 0;
-          } else {
-            return scope.userLibsToShow.length > 0;
-          }
+        scope.toggleMyLibsFirstOn = function() {
```
yipingliao: These two toggle methods can be combined into one.
`scope.sortingMenu.myLibsFirst = !scope.sortingMenu.myLibsFirst;`

kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
```diff
@@ -62,6 +61,85 @@ angular.module('kifi')
           scope.$broadcast('refreshScroll');
         }
 
+        function setCurrentStick(fromScroll) {
+          var offset = antiscrollLibList.scrollTop();
+          var libItemHeight = 0, separatorHeight = 0;
```
yipingliao: nit: separate into two lines.

kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
yipingliao, line 103: Nice!

kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
```diff
@@ -62,6 +61,85 @@ angular.module('kifi')
           scope.$broadcast('refreshScroll');
         }
 
+        function setCurrentStick(fromScroll) {
```
yipingliao: This is just a suggestion, but it looks like this function is for sticking separators, and it may be more informative if the name reflects that (setSeparatorStickValues, perhaps?).

Second suggestion is the argument `fromScroll`. My understanding is that when `fromScroll` is false, then that means that we need to re-query the DOM to get the updated set of separator DOM elements. This is a little indirect. How about if we pass in an argument of refetchSeparators, also a boolean, that is true if separators need to be reloaded? That would make the code more self-documenting:

``` javascript
if (refetchSeparators) {
  separators = antiscrollLibList.find('.kf-nav-lib-separator');
}
```

Then, when you call it, you won't pass in an argument in the on scroll handler, but pass in an argument of true in `scope.changeList`, which would be understandable since when we change the list of libraries, we should refetch the separators.

yipingliao:
This looks awesome!

kifi-backend/modules/shoebox/angular/src/layout/nav/nav.js
ahsu1230: I see, yes better name reflecting! Thanks!

kifi-backend/modules/shoebox/app/views/marketing/helper/page.scala.html
```diff
@@ -16,7 +16,7 @@
       mixpanel.init("cff752ff16ee39eda30ae01bb6fa3bd6");
     </script>
     @headScripts
-    <link rel="shortcut icon" type="image/png" href="/assets/images/kifi-favi-50.png">
+    <link rel="shortcut icon" type="image/png" href="//d1dwdv9wd966qu.cloudfront.net/img/favicon64x64.7cc6dd4.png">
```
ahsu1230, line 19: Thanks for the help @joshm1

ahsu1230:
thanks for looking it over @yipingliao

PR #12494, opened 2014-11-21
remove unnecessary endpoint calls on public libraries

ahsu1230:
on public libraries, 
`/site/libraries` was being called for library summaries which is unneeded for public libraries.

@andrewconner @yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
@@ -170,7 +170,7 @@ angular.module('kifi')
             '&kcid=na-vf_twitter-library_invite-lid_' + scope.library.id);
           scope.library.shareText = 'Discover this amazing @Kifi library about ' + scope.library.name + '!';
 
-          if (scope.isUserLoggedOut) {
+          if (scope.$root.userLoggedIn === false) {
             scope.$evalAsync(function () {
               angular.element('.white-background').height(element.height() + 20);
             });
--- a/kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
@@ -170,28 +170,29 @@ angular.module('kifi')
             '&kcid=na-vf_twitter-library_invite-lid_' + scope.library.id);
           scope.library.shareText = 'Discover this amazing @Kifi library about ' + scope.library.name + '!';
 
-          // Figure out whether this library is a library that the user has been invited to.
-          // If so, display an invite header.
-          var promise = null;
-          if (libraryService.invitedSummaries.length) {
-            promise = $q.when(libraryService.invitedSummaries);
-          } else {
-            promise = libraryService.fetchLibrarySummaries(true).then(function () {
-              return libraryService.invitedSummaries;
-            });
-          }
 
-          promise.then(function (invitedSummaries) {
-            var maybeLib = _.find(invitedSummaries, { 'id' : scope.library.id });
-            if (maybeLib) {
-              scope.library.invite = {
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
```diff
@@ -170,32 +170,12 @@ angular.module('kifi')
             '&kcid=na-vf_twitter-library_invite-lid_' + scope.library.id);
           scope.library.shareText = 'Discover this amazing @Kifi library about ' + scope.library.name + '!';
 
-          // Figure out whether this library is a library that the user has been invited to.
```
ahsu1230, line 173: replaced with $watch instead

kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
```diff
@@ -430,6 +410,20 @@ angular.module('kifi')
           }
         });
 
+        // Figure out whether this library is a library that the user has been invited to.
```
andrewconner, line 413: We should add a backend field for this. Mobile needs it also, and this could be expensive.

kifi-backend/modules/shoebox/angular/src/libraries/libraryService.js
```diff
@@ -144,6 +144,9 @@ angular.module('kifi')
       },
 
       fetchLibrarySummaries: function (invalidateCache) {
+        if (!$rootScope.userLoggedIn) {
```
andrewconner: Why is `fetchLibrarySummaries` being called? We should stop _that_. Otherwise, we end up with tons of guards in our services to check for login state.

yipingliao: `fetchLibrarySummaries` is called once, upon initial page load.

andrewconner: Ah. What if we moved that call to app.js, around line 134?

yipingliao: After page load, `fetchLibrarySummaries` is called when the user has taken some action that changes library, like adding keeps and creating libraries. Not sure how we can stop calling this?

kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
```diff
@@ -170,32 +170,12 @@ angular.module('kifi')
...
-            if (maybeLib) {
-              scope.library.invite = {
-                inviterName: maybeLib.inviter.firstName + ' ' + maybeLib.inviter.lastName,
-                actedOn: false
-              };
-            }
-          });
-
-          if (scope.$root.userLoggedIn === false) {
+          if (scope.isUserLoggedOut) {
```
yipingliao: FYI, I'm removing this (it's on one of my dev branches). :) Can be done in CSS only.

kifi-backend/modules/shoebox/angular/src/libraries/libraryService.js
yipingliao: I think that works.

kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
ahsu1230: ooh nice!

ahsu1230: moved it back to the original line (same logic anyway) so you don't get a merge conflict or anything =)

yipingliao:
Thanks for cleaning this up!

kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
ahsu1230, line 413: Sorry, I think I'm missing something. What would the field be?

PR #12497, opened 2014-11-22
Curator diverse recommendations by topic

joshm1:
@stkem @yingjieMiao @leogrim 

With topics stored for URI recommendations, the recommendation retrieval algorithm (enabled by experiment) groups recommendations by primary topic (topic1) and applies an exponential decay function the scores of URIs (sorted by score descending).

I tested a few coefficients for the function, and ended with going with 1/15, which would mean the 2nd recommendation in a topic has a score multiplier of 0.93, the 5th has .71.  Of course we'll adjust this coefficient (or improve the algorithm), but my goal was to not apply too big of a decay that we end up rejecting great recommendations in favor of mediocre ones.

### DIFF
--- a/kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationRetrievalCommander.scala
+++ b/kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationRetrievalCommander.scala
@@ -1,16 +1,13 @@
 package com.keepit.curator.commanders
 
-import com.keepit.common.logging.Logging
-import com.keepit.cortex.models.lda.LDATopic
-import com.keepit.curator.model._
 import com.keepit.common.db.Id
+import com.keepit.curator.model.{ RecoInfo, RecommendationClientType, UriScores, PublicFeedRepo, UriRecommendationRepo, UriRecommendation }
 import com.keepit.model.User
 import com.keepit.common.db.slick.Database
 import com.keepit.common.akka.SafeFuture
 
 import play.api.libs.concurrent.Execution.Implicits.defaultContext
 
-import scala.collection.immutable.SortedSet
 import scala.util.Random
 
 import com.google.inject.{ Inject, Singleton }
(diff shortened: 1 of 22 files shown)


yingjieMiao:
lgtm overall. :+1:

joshm1:
thanks for the feedback @yingjieMiao!  I updated per your suggestions.

yingjieMiao:
one small comment. lgtm otherwise.

yingjieMiao:
I assume  we are going to propagate topic ids first, then start the experiment with diverse strategy. otherwise the groupBy will group quite a bit `None`.

joshm1:
yep. certainly enabling the experiment will be useless until the topicIds propagate to curator.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationRetrievalCommander.scala
```diff
@@ -12,6 +12,36 @@ import scala.util.Random
 
 import com.google.inject.{ Inject, Singleton }
 
+case class UriRecoScore(score: Float, reco: UriRecommendation)
+
+trait RecoSortStrategy {
+  def sort(recosByTopScore: Seq[UriRecoScore]): Seq[UriRecoScore]
+}
+
+class TopScoreRecoSortStrategy(val minScore: Float = 5f, val limit: Int = 10) extends RecoSortStrategy {
```
stkem, line 21: I  think having the limit in here conflates the sorting with the batch size (and also the name is not entirely correct, since it really does selection, not sorting). Maybe move the `take` out of this (could be just in the `topRecos` method in one place, after the sorting)

stkem:
:+1:

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationRetrievalCommander.scala
joshm1, line 21: Agree the name is not the best.  `take` is here to be consistent with `DiverseRecoSortStrategy` where the `limit` is used to both remove lower scoring recos from _each_ topic before the decay is applied to the scores. A previous implementation also used it to remove topics that didn't have a top 10 scoring reco (which reduced the # of decay ops from 1000 to 100 or less).  That was removed for clarity, and all URIs from removing lower-scoring topics required additional pre-sorting which may be slower than just applying the decay to all 1000 URIs.

stkem, line 21: In that case I suggest renaming it to `*RecoSelectionStrategy`?

PR #12500, opened 2014-11-24
Do not show tag list on keep cards for logged-out users.

yipingliao:
@andrewconner cc @ahsu1230 
https://app.asana.com/0/19898402862947/20950859174981


ahsu1230:
:+1:

2is10:
We don’t want search engines indexing tags?

yipingliao:
The plan is to put them back in when we deploy public library search. How does that sound?

2is10:
Fine with me. I was just curious about the reasoning. Thanks for explaining
in person.

PR #12504, opened 2014-11-24
update keep endpoint for website (title only)

ahsu1230:
remove old updateKeep website endpoint
and replace with new endpoint

Decided to keep route naming consistent with extension (since they use the same commander function)

@andrewconner @2is10

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/KeepsController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/KeepsController.scala
@@ -128,25 +128,6 @@ class KeepsController @Inject() (
     }
   }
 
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/shoebox/app/com/keepit/controllers/website/KeepsController.scala
```diff
@@ -128,25 +128,6 @@ class KeepsController @Inject() (
     }
   }
 
-  def updateKeepInfo(id: ExternalId[Keep]) = UserAction { request =>
```
ahsu1230, line 131: remove unused endpoint

2is10:
:+1:

PR #12513, opened 2014-11-25
3.3.22: no longer suggesting recent or popular tags for a page

2is10:
This undoes or disables some of the work done in 16b6afa (#12071) and in c7e7017.

### DIFF
--- a/packrat/build.properties
+++ b/packrat/build.properties
@@ -1,1 +1,1 @@
-version=3.3.21
+version=3.3.22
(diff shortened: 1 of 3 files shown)


2is10:
cc @leogrim 

Danny wants to disable tag suggestions until he has some time to design a UI for it. The current UI is a bit clunky, as it’s not clear where the suggestions came from and why they’re being shown before you type anything.

PR #12514, opened 2014-11-25
Curator library ingest

joshm1:
@stkem

### DIFF
--- a/kifi-backend/modules/curator/app/com/keepit/curator/commanders/AllLibrarySeedIngestionHelper.scala
+++ b/kifi-backend/modules/curator/app/com/keepit/curator/commanders/AllLibrarySeedIngestionHelper.scala
@@ -1,0 +1,80 @@
+package com.keepit.curator.commanders
+
+import com.google.inject.{ Inject, Singleton }
+import com.keepit.common.db.slick.DBSession.RWSession
+import com.keepit.common.db.slick.Database
+import com.keepit.common.db.{ SequenceNumber, State }
+import com.keepit.common.healthcheck.AirbrakeNotifier
+import com.keepit.common.logging.Logging
+import com.keepit.curator.model.{ CuratorLibraryInfo, CuratorLibraryInfoRepo }
+import com.keepit.model.{ DetailedLibraryView, Library, Name, SystemValueRepo }
+import com.keepit.shoebox.ShoeboxServiceClient
+
+import play.api.libs.concurrent.Execution.Implicits.defaultContext
+
+import scala.concurrent.Future
+
+@Singleton
+class AllLibrarySeedIngestionHelper @Inject() (
+    systemValueRepo: SystemValueRepo,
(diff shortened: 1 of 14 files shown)


kifi-backend/modules/curator/app/com/keepit/curator/commanders/AllLibrarySeedIngestionHelper.scala
```diff
@@ -0,0 +1,81 @@
...
+      lastFollowed = library.lastFollowed,
+      kind = library.kind,
+      libraryLastUpdated = library.updatedAt,
+      state = State[CuratorLibraryInfo](library.state.value)
+    ))
+  }
+
+  private def processUpdatedLibrary(library: DetailedLibraryView, libraryInfo: CuratorLibraryInfo)(implicit session: RWSession): Unit = {
+    libraryInfoRepo.save(libraryInfo.copy(
+      libraryId = library.id.get,
```
stkem: You are getting into this code path by looking up the library by id. This line thus does nothing.

kifi-backend/modules/common/conf/evolutions/shoebox/264.sql
```diff
@@ -0,0 +1,27 @@
...
+    keep_count int NOT NULL,
+    visibility varchar(32) NOT NULL,
+    last_kept DATETIME,
+    last_followed DATETIME,
+    state varchar(20) NOT NULL,
+    kind varchar(64) NOT NULL,
+    library_last_updated DATETIME NOT NULL,
+
+    PRIMARY KEY (id),
+    UNIQUE INDEX curator_library_info_u_keep_id (library_id),
```
stkem: `keep_id`? Copy/paste issue?

kifi-backend/modules/curator/app/com/keepit/curator/commanders/SeedIngestionCommander.scala
```diff
@@ -44,22 +45,37 @@ class SeedIngestionCommander @Inject() (
...
+
+      ingestKeepsF.onComplete {
+        case Failure(ex) =>
+          log.error("Failure occurred during keeps ingestion.")
+          airbrake.notify("Failure occurred during keeps ingestion.", ex)
         case _ =>
       }
-      fut.map(_ => true)
+
+      Future.sequence(ingestLibMembershipsF :: ingestKeepsF :: ingestLibrariesF :: Nil) map (_ => true)
```
stkem: This works, but I think it would be better to have a separate lock for libraries/memberships, so they can proceed completely independently from keeps etc. The way it currently is, a really long running keep ingestions would block the next (not the current) scheduled library ingestion.

stkem:
made some small comments, lgtm overall. Nice work! Next step is the actual library recommendation generation pipeline. This should structurally be very similar to the uri recommendation generation. We can talk about details when I get to the office.

joshm1:
@stkem thanks for the feedback. Updated per your suggestions.

PR #12516, opened 2014-11-25
add a few fields to library topic repo

yingjieMiao:
@stkem  @joshm1 

Similar to those fields for uri topic. One more field on 'entropy'. Might be useful later on.


stkem:
lgtm

PR #12517, opened 2014-11-25
non-youtube sometimes match as youtube video

ahsu1230:
was getting some false-positives for matching youtube videos like: 

https://foursquare.com/v/ramen-dojo/4bbe2b79061fb713aed7edce

hopefully should fix edge cases. 

@andrewconner @yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/keep/keep.js
+++ b/kifi-backend/modules/shoebox/angular/src/keep/keep.js
@@ -82,7 +82,7 @@ angular.module('kifi')
         var useBigLayout = false;
         var strippedSchemeRe = /^https?:\/\//;
         var domainTrailingSlashRe = /^([^\/]*)\/$/;
-        var youtubeLinkRe = /.*(?:youtu.be\/|v\/|u\/\w\/|embed\/|watch\?v=)([^#\&\?]*).*/;
+        var youtubeLinkRe = /^.*(youtu.be\/|v\/|e\/|u\/\w+\/|embed\/|v=)([^#\&\?]*).*/;
 
         var tagDragMask = element.find('.kf-tag-drag-mask');
         var mouseX, mouseY;
@@ -124,8 +124,8 @@ angular.module('kifi')
         function isYoutubeCard(url) {
           var strippedSchemeLen = (url.match(strippedSchemeRe) || [''])[0].length;
           var match = url.substr(strippedSchemeLen).match(youtubeLinkRe);
-          if (match && match[1] && match[1].length === 11) { // youtube Ids are always 11 characters
-            scope.youtubeId = match[1];
+          if (match && match[2] && match[2].length === 11) { // youtube Ids are always 11 characters
+            scope.youtubeId = match[2];
           }
           return scope.youtubeId;
         }
--- a/kifi-backend/modules/shoebox/angular/src/keep/keep.js
+++ b/kifi-backend/modules/shoebox/angular/src/keep/keep.js
@@ -124,7 +124,7 @@ angular.module('kifi')
         function isYoutubeCard(url) {
           var strippedSchemeLen = (url.match(strippedSchemeRe) || [''])[0].length;
           var match = url.substr(strippedSchemeLen).match(youtubeLinkRe);
-          if (match && match[1]) {
+          if (match && match[1] && match[1].length === 11) { // youtube Ids are always 11 characters
             scope.youtubeId = match[1];
           }
           return scope.youtubeId;


kifi-backend/modules/shoebox/angular/src/keep/keep.js
```diff
@@ -82,7 +82,7 @@ angular.module('kifi')
         var useBigLayout = false;
         var strippedSchemeRe = /^https?:\/\//;
         var domainTrailingSlashRe = /^([^\/]*)\/$/;
-        var youtubeLinkRe = /.*(?:youtu.be\/|v\/|u\/\w\/|embed\/|watch\?v=)([^#\&\?]*).*/;
+        var youtubeLinkRe = /^.*(youtu.be\/|v\/|e\/|u\/\w+\/|embed\/|v=)([^#\&\?]*).*/;
```
andrewconner, line 85: The `.*` is pretty aggressive. Perhaps optionally matching `https?:\/\/`?

andrewconner, line 85: and optional `www`

ahsu1230, line 85: The `strippedSchemeRe` variable (line 83) above will try to strip out `http://` or `https://` before trying to match.

ahsu1230, line 85: I think that was originally in there to help relieve the regex parsing?

kifi-backend/modules/shoebox/angular/src/keep/keep.js
```diff
@@ -124,8 +124,8 @@ angular.module('kifi')
         function isYoutubeCard(url) {
           var strippedSchemeLen = (url.match(strippedSchemeRe) || [''])[0].length;
           var match = url.substr(strippedSchemeLen).match(youtubeLinkRe);
-          if (match && match[1]) {
-            scope.youtubeId = match[1];
+          if (match && match[2] && match[2].length === 11) { // youtube Ids are always 11 characters
```
2is10, line 127: The 11-character thing can go in the regular expression. Use `{11}` instead of `*` or `+`.

kifi-backend/modules/shoebox/angular/src/keep/keep.js
2is10, line 85: In my opinion, complex regular expressions are pretty worthless without a list of positive and negative test cases.

This looks like it will have many false positives as well. Why can’t this be limited to a fixed set of domains?

ahsu1230, line 85: so this `/^.*(youtu.be\/|v\/|e\/|u\/\w+\/|embed\/|v=)([^#\&\?]*).*/`(new regex pattern) should match the following: 

```
_http://www.youtube.com/v/0zM3nApSvMg?fs=1&hl=en_US&rel=0
_http://www.youtube.com/embed/0zM3nApSvMg?rel=0
_http://www.youtube.com/watch?v=0zM3nApSvMg&feature=feedrec_grec_index
_http://www.youtube.com/watch?v=0zM3nApSvMg#t=0m10s
_http://www.youtube.com/user/IngridMichaelsonVEVO#p/a/u/1/KdwsulMb8EQ
_http://youtu.be/dQw4w9WgXcQ
_http://www.youtube.com/embed/dQw4w9WgXcQ
_http://www.youtube.com/v/dQw4w9WgXcQ
_http://www.youtube.com/e/dQw4w9WgXcQ
_http://www.youtube.com/watch?v=dQw4w9WgXcQ
_http://www.youtube.com/?v=dQw4w9WgXcQ
_http://www.youtube.com/watch?feature=player_embedded&v=dQw4w9WgXcQ
_http://www.youtube.com/?feature=player_embedded&v=dQw4w9WgXcQ
_http://www.youtube.com/user/IngridMichaelsonVEVO#p/u/11/KdwsulMb8EQ
_http://www.youtube-nocookie.com/v/6L3ZvIMwZFM?version=3&hl=en_US&rel=0
```

taken from http://stackoverflow.com/questions/3452546/javascript-regex-how-to-get-youtube-video-id-from-url

The tricky part is for getting the video ID out of each of these urls, since they are in a lot of different cases. So, should we only apply this regex after checking the domain name is either `www.youtube.com` or `youtu.be`?

2is10, line 85: I like that idea. Here’s a regex that does it. I’ve tested it on all of the positive examples above. We should be sure to include negative test cases too.

``` js
var youtubeVideoUrlRe = /^(?:https?:\/\/)?(?:youtu\.be|(?:www\.)?youtube(?:-nocookie)?\.com)\/(?:|user\/[^\/?#]+)?(?:|.*?[\/=])([a-zA-Z0-9]{11})\b/;
function getVideoIdIfYouTubeVideo(url) {
  var match = url.match(youtubeVideoUrlRe);
  return match && match[1];
}
```

This regular expression only accepts the following domains:
- youtu.be
- youtube.com
- www.youtube.com
- youtube-nocookie.com
- www.youtube-nocookie.com

Either the `http:` or `https:` scheme is acceptable, but not required.
The path may optionally begin with `/user/` and a user identifier path segment.
The 11-character video ID must be preceded by either `/` or `=` and must either terminate the URL or be followed by a non-word character.

ahsu1230, line 85: trying to use in #12605

PR #12520, opened 2014-11-25
trying to add stacktraces to the test log in jenkins

eishay:
See
http://www.scalatest.org/user_guide/using_scalatest_with_sbt
http://www.scala-sbt.org/0.13.2/docs/Howto/logging.html
and fixing the build (maybe)


eishay:
@andrewconner looks like it fixed the stacktrace issue on jenkins

andrewconner:
I wonder what broke this. Thanks!

PR #12532, opened 2014-11-25
update play to serve marketing html files from gulp-generated files

joshm1:
@stkem @andrewconner


joshm1:
The first step to test locally is to run `mkt-npm install` then `mkt` in sbt.  Then `run` and make sure the marketing pages load in localhost:9000

stkem:
code lgtm, we should monitor carefully when deployed

PR #12534, opened 2014-11-25
penalizing a few newsy domains heavily

stkem:
list form sterling; will likely get longer
@joshm1 @yingjieMiao


yingjieMiao:
cool. we can start thinking using the feedback loop. (Maybe something like multi-armed bandit learning, which isn't too hard to implement. Or even cheaper alternatives. )

PR #12535, opened 2014-11-25
consolidator ttl tweak for getSignature

raymondng:
It's not the real fix for the issues observed today, but given that the "loop" extends beyond 30s -- this may help.

@ymatsuda @leogrim @yingjieMiao @eishay


yingjieMiao:
cool. it's fair to assume url's signature won't change within 5 minutes.

eishay:
nice

leogrim:
Thanks!

PR #12540, opened 2014-11-26
consolidate (external) fetch requests

raymondng:
Along the same lines, add request consolidation to (external) fetch requests. 

Or -- the world's simplest (and almost useful) rate-limiter for scraper.

Under normal circumstances this is likely no-op. But if somehow we end up sending multiple fetch request for the same URL, this serves as a safety net and may (help) avoid sites blocking us due to overly aggressive scraping.

@yingjieMiao @leogrim @ymatsuda @eishay @andrewconner


yingjieMiao:
lgtm

PR #12548, opened 2014-11-26
don't like object Singleton around, can't inject stuff to them

eishay:


eishay:
@andrewconner i would like to replace some functionality with image magic fu

PR #12559, opened 2014-12-01
parsing the keep image file extension from the URL path

2is10:
@eishay 

This fixes the exact case I encountered, at least, which was:
http://images.vice.com/motherboard/content-images/article/17117/1416839855864368.png?crop=1xw:0.932295719844358xh;*,*&resize=2300:*&output-format=jpeg&output-quality=90

https://app.asana.com/0/18005085042422/21147414406232


2is10:
merging because another is coming

PR #12561, opened 2014-12-01
sniffing the content type from a downloaded file if necessary

2is10:
@andrewconner @eishay 

We were only using `URLConnection.guessContentTypeFromStream` on uploaded images, so this is a quick fix to use it on downloaded files too.

Potential optimizations:
- [`URLConnection.guessContentTypeFromStream`](http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/7-b147/java/net/URLConnection.java#URLConnection.guessContentTypeFromStream%28java.io.InputStream%29) only uses the first 16 bytes of a stream. We could run it much sooner than when the download completes.
- We could also reject large files much earlier based on the `Content-Length` header.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/ProcessedImageHelper.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/ProcessedImageHelper.scala
@@ -140,10 +140,8 @@ trait ProcessedImageHelper {
 
           if (headers.status != 200) {
             Future.failed(new RuntimeException(s"Image returned non-200 code, ${headers.status}, $imageUrl"))
-          } else if (formatOpt.isEmpty) {
-            Future.failed(new RuntimeException(s"Unknown image type, ${headers.headers.get("Content-Type")}, $imageUrl"))
           } else {
-            val tempFile = TemporaryFile(prefix = "remote-file", suffix = "." + formatOpt.get.value)
+            val tempFile = TemporaryFile(prefix = "remote-file")
             tempFile.file.deleteOnExit()
             val outputStream = new FileOutputStream(tempFile.file)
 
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/ProcessedImageHelper.scala
```diff
@@ -140,10 +140,8 @@ trait ProcessedImageHelper {
 
           if (headers.status != 200) {
             Future.failed(new RuntimeException(s"Image returned non-200 code, ${headers.status}, $imageUrl"))
-          } else if (formatOpt.isEmpty) {
-            Future.failed(new RuntimeException(s"Unknown image type, ${headers.headers.get("Content-Type")}, $imageUrl"))
           } else {
-            val tempFile = TemporaryFile(prefix = "remote-file", suffix = "." + formatOpt.get.value)
+            val tempFile = TemporaryFile(prefix = "remote-file")
```
2is10, line 144: I’m assuming here that the temp file name suffix isn’t really important, since this function returns the format explicitly along with the file. Valid assumption?

andrewconner, line 144: I vaguely remember a null pointer exception when prefix and suffix weren't set.

andrewconner:
Thanks, good improvement

kifi-backend/modules/shoebox/app/com/keepit/commanders/ProcessedImageHelper.scala
```diff
@@ -160,11 +158,17 @@ trait ProcessedImageHelper {
...
-                result.get
-            }.map(_ => (formatOpt.get, tempFile))
+            } flatMap { _ =>
+              formatOpt.orElse(detectImageType(tempFile)) map { format =>
+                Future.successful((format, tempFile))
+              } getOrElse {
+                tempFile.file.delete
+                Future.failed(new Exception(s"Unknown image type, ${headers.headers.get("Content-Type")}, $imageUrl"))
+              }
+            }
```
2is10, line 170: Since this detection is pretty cheap (a seek and 16-byte read), we might consider always performing it and preferring the detected format over the declared format.

kifi-backend/modules/shoebox/app/com/keepit/commanders/ProcessedImageHelper.scala
2is10, line 144: Hmm, the tests cover this code path. I’ll verify and then merge.

PR #12563, opened 2014-12-02
[DO NOT MERGE] Twitter token (oauth1a) support

raymondng:
Twitter token (oauth1a) support
Refactored token-based sign-up/login code for sharing

OAuth2 token based code paths are already covered
OAuth1a token based code paths are still based on `SecureSocial` -- tested locally
If we have time, will try to factor out OAuth1a from `SecureSocial` similar to how OAuth2 providers were factored out earlier

Actually this can be merged -- the `DO NOT MERGE` is really `hold off` until tomorrow when everyone is in office and can test/react in a timely manner.

@andrewconner 

cc: @jaredpetker @thassss -- in theory, you can start testing the endpoints


raymondng:
Per discussion with @andrewconner -- will push and test

PR #12572, opened 2014-12-02
removing concurrent check for test user email, reading emails from db us...

eishay:
expecting to have better error code from IM on failure


eishay:
@andrewconner @ymatsuda @stkem @raymondng

eishay:
don't merge yet, test failed locally

PR #12577, opened 2014-12-02
retrying urban airship push notifications twice before quitting the aler...

eishay:
...ting.

see https://support.urbanairship.com/entries/68913408-Best-practices-for-using-the-API-or-why-you-should-log-the-Push-ID


eishay:
@stkem @raymondng

PR #12581, opened 2014-12-03
library recommendations in curator

joshm1:
@stkem @yingjieMiao 

This is very much a WIP (parts of the code are obviously incomplete and features/tests are missing). Sharing anyway if you don't mind taking a quick look since the PR is already quite big.

### DIFF
--- a/kifi-backend/modules/common/conf/evolutions/shoebox/267.sql
+++ b/kifi-backend/modules/common/conf/evolutions/shoebox/267.sql
@@ -1,0 +1,25 @@
+# CURATOR
+
+# --- !Ups
+
+CREATE TABLE library_recommendation (
+    id bigint(20) NOT NULL AUTO_INCREMENT,
+    created_at datetime NOT NULL,
+    updated_at datetime NOT NULL,
+    state varchar(20) NOT NULL,
+
+    user_id bigint(20) NOT NULL,
+    library_id bigint(20) NOT NULL,
+    master_score float(10) NOT NULL,
+    all_scores text NOT NULL,
+    followed boolean NOT NULL,
+
+    PRIMARY KEY (id),
+
+    UNIQUE INDEX library_recommendation_u_library_id_user_id (library_id, user_id),
+    INDEX library_recommendation_i_user_id_master_score (user_id, master_score)
+);
+
+insert into evolutions (name, description) values('267.sql', 'create library_recommendation table');
+
+# --- !Downs
--- a/kifi-backend/modules/curator/app/com/keepit/curator/LibraryScoringHelper.scala
+++ b/kifi-backend/modules/curator/app/com/keepit/curator/LibraryScoringHelper.scala
@@ -1,0 +1,104 @@
+package com.keepit.curator
+
+import com.google.inject.Inject
+import com.keepit.common.db.slick.Database
+import com.keepit.common.time._
+import com.keepit.curator.commanders.LibraryRecoCandidate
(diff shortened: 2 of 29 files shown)


stkem:
Nice! Will review thoroughly today.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/LibraryRecommendationGenerationCommander.scala
```diff
@@ -0,0 +1,136 @@
...
+      map(users => users.map(_.id.get).toSeq)
+
+  def getTopRecommendations(userId: Id[User], howManyMax: Int): Future[Seq[LibraryRecommendation]] = {
+    db.readOnlyReplicaAsync { implicit session =>
+      libraryRecRepo.getByTopMasterScore(userId, howManyMax)
+    }
+  }
+
+  private def shouldInclude(scoredLibrary: ScoredLibraryInfo): Boolean = {
+    scoredLibrary.masterScore > 0 // TODO(josh)
```
joshm1, line 43: going to keep this at 0 for initial internal testing to get an idea of what kind of scores the current algorithm derives with real data

joshm1:
@stkem @yingjieMiao  I think my first draft is done. I'm moving to the admin page next, but this can be deployed after review(s)

yingjieMiao:
sorry, I was late on this. Will go through this tonight.

kifi-backend/modules/curator/app/com/keepit/curator/LibraryScoringHelper.scala
```diff
@@ -0,0 +1,120 @@
...
+    }
+  }
+
+  private def getPopularityScore(candidate: CuratorLibraryInfo): Float = {
+    1 / (1 + Math.exp(-candidate.memberCount / 5)).toFloat
+  }
+
+  private def getSizeScore(candidate: CuratorLibraryInfo): Float = {
+    1 / (1 + Math.exp(-candidate.keepCount / 15)).toFloat
+  }
```
yingjieMiao, line 118: cortex also has 'entropy' info for a library. Big Library tends to have 'maxed entropy', in a 256 dim model, this is log_2(256) = 8.0

We could have some penalty for libraries with big entropy. (I can include this in the user/library score in the future)

yingjieMiao:
lgtm overall. (didn't scrutinize though). It's probably better to push smaller pull requests. (i.e. push out self-contained modules early). 

:+1:

kifi-backend/modules/common/conf/evolutions/shoebox/267.sql
```diff
@@ -0,0 +1,25 @@
+# CURATOR
+
+# --- !Ups
+
+CREATE TABLE library_recommendation (
```
stkem, line 5: Does not need to be in this PR, but this table should probably gain a few more fields similar to uri_recommendation, i.e. the topics of the library and possibly feedback fields ('clicked', 'followed', 'trashed' and such, these will depend on the final UI)

stkem:
Nice work! Let's get it into prod :-)

kifi-backend/modules/curator/app/com/keepit/curator/commanders/LibraryRecommendationGenerationCommander.scala
stkem, line 43: :+1:

PR #12582, opened 2014-12-03
taking only the first frame of a GIF and converting it to a PNG before r...

eishay:
...esizing/optimizing it


eishay:
@andrewconner @2is10 
https://app.asana.com/0/17575799253041/21147414406237

2is10:
sweet, thanks

PR #12586, opened 2014-12-03
mobile endpoints: add, modify, delete libraries

ahsu1230:
Mobile endpoints for:
- Create Library (POST /m/1/libraries/add)
  request: `{name, description, visibility}`, response: LibraryInfo + membership
- Modify Library (POST /m/1/libraries/:id/modify)
  request: `{newName, newDescription, newVisibility, newSlug}`, response: LibraryInfo + membership
- Delete Library (POST /m/1/libraries/:id/delete)
  (no body), response: `No Content`

with extra tests

keeps for libraries coming in a bit
@andrewconner @thassss @jaredpetker

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileLibraryController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileLibraryController.scala
@@ -35,11 +35,11 @@ class MobileLibraryController @Inject() (
 
   def createLibrary() = UserAction(parse.tolerantJson) { request =>
     val jsonBody = request.body
-    val title = (jsonBody \ "title").as[String]
+    val name = (jsonBody \ "name").as[String]
     val description = (jsonBody \ "description").asOpt[String]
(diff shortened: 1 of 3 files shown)


andrewconner:
Thanks!

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileLibraryController.scala
```diff
@@ -33,6 +33,49 @@ class MobileLibraryController @Inject() (
   implicit val config: PublicIdConfiguration)
     extends UserActions with LibraryAccessActions with ShoeboxServiceController {
 
+  def createLibrary() = UserAction(parse.tolerantJson) { request =>
+    val jsonBody = request.body
+    val name = (jsonBody \ "name").as[String]
```
ahsu1230, line 38: will use `name` instead of `title` for library names just to be consistent with extension & website library requests, hope it's not an issue =) @thassss

PR #12588, opened 2014-12-04
[DO NOT MERGE] OAuth1

raymondng:
The goal is to bring OAuth1 (token-based) to more-or-less same level as OAuth2 -- more modular/pluggable and testable, remove `SecureSocial` dependencies where possible, etc.

To that end, the main value-add is `OAuth1TokenTest` -- much of the work is done to make this possible. This test is currently quite rudimentary, and certainly not on par with the OAuth2-based test, but I think the foundations are there and adding new tests should be simple -- I plan to add more before the merge.

The `DO NOT MERGE` applies as before -- review feedback appreciated but we'll wait until tomorrow to merge.

Now we have coverage, and code is locally tested but will again need help ensure nothing unexpected breaks after merge. Thanks in advance!

@andrewconner 
cc: @jaredpetker @thassss


kifi-backend/modules/common/app/com/keepit/common/oauth/TwitterOAuthProvider.scala
```diff
@@ -0,0 +1,64 @@
...
+      .get()
+
+    call map { response =>
+      log.info(s"[fillProfile] response.body=${response.body}")
+      val me = response.json
+      // should get screen name and follower count at a minimum
+      val userId = (me \ Id).as[Long]
+      val name = (me \ Name).as[String]
+      val splitted = name.split(' ')
+      val (firstName, lastName) = if (splitted.length < 2) (name, "") else (splitted.head, splitted.takeRight(1).head)
```
eishay, line 50: There's a pattern matching trick you could use instead of the `takeRight(1).head`

``` scala
val (firstName, lastName) = splitted match {
  case Seq(firstName, lastName) => (firstName, lastName)
  case _ => (name, "")
}
```

Writing the code above I wonder how would you split the name "Jean Luc Picard" to first and last names. I wouldn't mess with him...

eishay, line 50: not sure i saw a test for it

2is10, line 50: I think first names are more often multiple words than last names, unless the second-to-last word begins with a lowercase letter.

Examples:
`Jean Luc` `Picard`
`Jayne Anne` `Shaw`
`Ruth Ann` `Daines`
`Kevin` `de León`
`Helen` `Richardson-Walsh`

Of course there are exceptions. We may want to have a heuristic (machine-learned?) for classifying middle names as part of the first or the last. And then there are suffixes like `III` and `Jr.` which we should include in our last name field, since we really use it logically as “everything but first name”.

eishay, line 50: Another example:
`Jean-Claude Van Damme`

andrewconner, line 50: We should also consider that Twitter does not have a real name policy — the name can be anything, including nonsensical names.

raymondng, line 50: Exactly.

Twitter's name is not nearly as useful/relevant as facebook's or linkedin's -- I have highlighted this as issue#2 in the document [Twiter Auth Issues](https://docs.google.com/a/42go.com/document/d/1q6GDS8fQCZFMZlxtLcOi88n4RsEezS1LlzGnggQPLpQ/edit#heading=h.pu6mizgwjc9). 
In fact, the Twitter documentation lists `twitterati` as an example for “name” field, which is not a name at all (this is also mentioned in the doc but repeated here for completeness)
As for whether we should move away from the `first name` and `last name` requirement, it's a bit of a larger discussion -- we should probably discuss it in the context of the [Twitter Auth Issues](https://docs.google.com/a/42go.com/document/d/1q6GDS8fQCZFMZlxtLcOi88n4RsEezS1LlzGnggQPLpQ/edit#heading=h.pu6mizgwjc9) document.

raymondng, line 50: Also, thanks @eishay for the reminder -- as mentioned in the general comments, a lot more tests on the way!

raymondng:
Additional tests added to bring `OAuth1TokenTest` on par with `OAuth2TokenTest`

raymondng:
Per discussion with @andrewconner -- will merge and test. Will continue to look into naming issue as well.

PR #12589, opened 2014-12-04
add keep to library, remove keep from library

ahsu1230:
@andrewconner @jaredpetker @thassss 

add keep to library (POST    /m/1/libraries/:id/keeps)
takes in `{title, url, imageUrl, tagNames}`. title & imageUrl are optional. For tagNames, if no tags, use `[]`
remove keep from library (POST    /m/1/libraries/:id/keeps/delete)

### DIFF
--- a/kifi-backend/modules/shoebox/conf/shoebox.routes
+++ b/kifi-backend/modules/shoebox/conf/shoebox.routes
@@ -238,6 +238,8 @@ POST    /m/1/libraries/:id/join                 @com.keepit.controllers.mobile.M
 POST    /m/1/libraries/:id/decline              @com.keepit.controllers.mobile.MobileLibraryController.declineLibrary(id: PublicId[Library])
 POST    /m/1/libraries/:id/leave                @com.keepit.controllers.mobile.MobileLibraryController.leaveLibrary(id: PublicId[Library])
 GET     /m/1/libraries/:id/keeps                @com.keepit.controllers.mobile.MobileLibraryController.getKeeps(id: PublicId[Library], offset: Int ?= 0, limit: Int ?= 10, imageWidth: Option[Int] ?= None, imageHeight: Option[Int] ?= None)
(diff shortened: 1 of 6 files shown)


kifi-backend/modules/shoebox/conf/shoebox.routes
```diff
@@ -238,6 +238,8 @@ POST    /m/1/libraries/:id/join                 @com.keepit.controllers.mobile.M
 POST    /m/1/libraries/:id/decline              @com.keepit.controllers.mobile.MobileLibraryController.declineLibrary(id: PublicId[Library])
 POST    /m/1/libraries/:id/leave                @com.keepit.controllers.mobile.MobileLibraryController.leaveLibrary(id: PublicId[Library])
 GET     /m/1/libraries/:id/keeps                @com.keepit.controllers.mobile.MobileLibraryController.getKeeps(id: PublicId[Library], offset: Int ?= 0, limit: Int ?= 10, imageWidth: Option[Int] ?= None, imageHeight: Option[Int] ?= None)
+POST    /m/1/libraries/:id/keeps                @com.keepit.controllers.mobile.MobileLibraryController.keepToLibrary(id: PublicId[Library])
+POST    /m/1/libraries/:id/keeps/delete         @com.keepit.controllers.mobile.MobileLibraryController.unkeepFromLibrary(id: PublicId[Library])
```
2is10, line 242: Why not put the keep ID in the URL and use DELETE method, like the website and extension do? Can mobile apps not make DELETE requests? Even if so, it would be nice (RESTful) to put the keep ID in the URL.

`POST    /m/1/libraries/:id/keeps/:k/delete`

andrewconner, line 242: Agreed. If it's one keep, `DELETE`. And if you want to allow a set of keep IDs (perhaps not useful for mobile?), we've historically used `POST`.

ahsu1230, line 242: Ok will let Tommy know about this. Thanks @2is10  for the catch

PR #12590, opened 2014-12-04
some refactoring and breaking db transaction to two, one that needs mast...

eishay:
...er db and one that can use replica (long calls that don't use cache). getting better visibility into what the search client augmentation can give us


eishay:
@ahsu1230

ahsu1230:
:+1: pretty neat!
Thanks =)

PR #12592, opened 2014-12-04
batching the "add to collection" so we won't try to add thousands of tag...

eishay:
...s in one huge session.

if the batch fail then retry and move on.
if the "insertAll" fails, throw an exception and not ignore it

With @ymatsuda

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
@@ -605,35 +605,43 @@ class KeepsCommander @Inject() (
     } getOrElse 0
   }
 
-  def addToCollection(collectionId: Id[Collection], keeps: Seq[Keep], updateIndex: Boolean = true)(implicit context: HeimdalContext): Set[KeepToCollection] = timing(s"addToCollection($collectionId,${keeps.length})") {
-    val result = db.readWrite { implicit s =>
-      val keepsById = keeps.map(keep => keep.id.get -> keep).toMap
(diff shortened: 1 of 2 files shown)


eishay:
@raymondng @andrewconner @ahsu1230

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
```diff
@@ -605,35 +605,43 @@ class KeepsCommander @Inject() (
...
-      val newK2C = newKeepIds map { kId => KeepToCollection(keepId = kId, collectionId = collectionId) }
-      timing(s"addToCollection($collectionId,${keeps.length}) -- keepToCollection.insertAll", 50) {
-        keepToCollectionRepo.insertAll(newK2C.toSeq)
-      }
-      val activated = existing collect {
-        case ktc if ktc.state == KeepToCollectionStates.INACTIVE && keepsById.contains(ktc.keepId) =>
-          keepToCollectionRepo.save(ktc.copy(state = KeepToCollectionStates.ACTIVE, createdAt = clock.now()))
-      }
+  def addToCollection(collectionId: Id[Collection], allKeeps: Seq[Keep], updateIndex: Boolean = true)(implicit context: HeimdalContext): Set[KeepToCollection] = timing(s"addToCollection($collectionId,${allKeeps.length})") {
+    val result: Iterator[KeepToCollection] = allKeeps.grouped(50) flatMap { keeps =>
```
ahsu1230, line 609: cool didn't know you could do that (grouping, i mean)

raymondng:
Thanks @eishay! As discussed we may want to consider `readWriteBatch` as well.

PR #12593, opened 2014-12-04
get writeable library summaries & given a url, give which libraries url ...

ahsu1230:
...is kept in already!
The endpoint is: (POST    /m/1/libraries)

also changed unkeep from library mobile endpoint to: 
DELETE  /m/1/libraries/:id/keeps/:kId

@andrewconner @thassss


kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtPageController.scala
```diff
@@ -15,19 +15,6 @@ class ExtPageController @Inject() (
   pageCommander: PageCommander)
     extends UserActions with ShoeboxServiceController {
 
-  def getPageDetails() = UserAction(parse.tolerantJson) { request =>
```
ahsu1230, line 18: This endpoint is not used anymore @2is10 . I checked with logentries, it hasn't been used over the past 30 days. OK to remove?

kifi-backend/modules/shoebox/conf/shoebox.routes
```diff
@@ -233,13 +231,14 @@ POST    /m/1/libraries/add                      @com.keepit.controllers.mobile.M
...
 POST    /m/1/libraries/:id/join                 @com.keepit.controllers.mobile.MobileLibraryController.joinLibrary(id: PublicId[Library])
 POST    /m/1/libraries/:id/decline              @com.keepit.controllers.mobile.MobileLibraryController.declineLibrary(id: PublicId[Library])
 POST    /m/1/libraries/:id/leave                @com.keepit.controllers.mobile.MobileLibraryController.leaveLibrary(id: PublicId[Library])
 GET     /m/1/libraries/:id/keeps                @com.keepit.controllers.mobile.MobileLibraryController.getKeeps(id: PublicId[Library], offset: Int ?= 0, limit: Int ?= 10, imageWidth: Option[Int] ?= None, imageHeight: Option[Int] ?= None)
 POST    /m/1/libraries/:id/keeps                @com.keepit.controllers.mobile.MobileLibraryController.keepToLibrary(id: PublicId[Library])
-POST    /m/1/libraries/:id/keeps/delete         @com.keepit.controllers.mobile.MobileLibraryController.unkeepFromLibrary(id: PublicId[Library])
+DELETE  /m/1/libraries/:id/keeps/:kId           @com.keepit.controllers.mobile.MobileLibraryController.unkeepFromLibrary(id: PublicId[Library], kId: ExternalId[Keep])
```
ahsu1230, line 241: Thanks for the suggestions @2is10 @andrewconner

kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtPageController.scala
2is10, line 18: Sure, thanks. Please also delete `PageCommander.getPageDetails` and any relevant helper code, including the `KeeperInfo` type.

2is10, line 18: Probably would have been best to do it in a different pull request.

ahsu1230, line 18: Turns out mobile is still using `PageCommander.getPageDetails` in two of its endpoints

2is10, line 18: Ok, good call. The data in the response was chosen to meet the extension’s
needs. It’s likely that the mobile apps aren’t using all of the data, in
which case we could optimize away a few calls from the commander method and
a few fields from the response. Not urgent, of course.​

ahsu1230, line 18: kk, sounds good!

PR #12594, opened 2014-12-04
including total number of keepers in /ext/page response

2is10:
@leogrim 
cc @eishay


2is10:
merging because Léo came and spoke to me about this pull request in person. happy to address any comments.

PR #12595, opened 2014-12-04
Add cutPoint field to old search API (mobile)

leogrim:
@ymatsuda per @mrbrown14's request, I'm not confident this is necessary (the old merger tail cuts on the backend doesn't it?) but the number of results returned always seems to be maxHits.


leogrim:
Solved and closing, the request was made with a filter `f=a`

PR #12599, opened 2014-12-04
starting to return some library info in /ext/page response

2is10:
@eishay 
You’ve been in meetings, so I made a quick start on this. If the approach is not what you had in mind, then I’ll be happy to scrap this.

cc @leogrim


2is10:
We may need to ask `searchClient.augment` for more libraries than we want in case we need to filter some out. It would be cool if that endpoint supported the kind of filtering we need.

eishay:
@2is10 you're too fast for me. Thanks!

leogrim:
Which kind of filtering would be needed?

2is10:
There are still some TODOs for you!

PR #12600, opened 2014-12-04
excluding viewer from count of total keepers in /ext/page response

2is10:
@leogrim 

This should minimize the impact of any discrepancy between the search service and shoebox. Also, this way the extension will not have to update this value in its cache each time the viewer keeps or unkeeps.


leogrim:
:+1:

PR #12603, opened 2014-12-04
simple admin page to see raw library reco scores + logging for library reco generation

joshm1:
@stkem 

lib recos aren't generating. I couldn't figure out why in code so hopefully these temporary logs will help.

admin page is not user friendly yet. I'll improve it.


kifi-backend/modules/shoebox/app/views/admin/curator/librecos/listUserRecos.scala.html
```diff
@@ -0,0 +1,3 @@
+@(recos: Seq[com.keepit.curator.model.LibraryRecoInfo])
+
+@recos
```
joshm1, line 3: this file isn't used yet

stkem:
lgtm. I'd be happy to pair debug for a bit with you if this doesn't reveal anything.

PR #12605, opened 2014-12-04
remove users who are already following in share dropdown

ahsu1230:
and a fix for detecting youtube keep @2is10 
@andrewconner @stkem


kifi-backend/modules/shoebox/angular/src/keep/keep.js
```diff
@@ -82,7 +82,7 @@ angular.module('kifi')
         var useBigLayout = false;
         var strippedSchemeRe = /^https?:\/\//;
         var domainTrailingSlashRe = /^([^\/]*)\/$/;
-        var youtubeLinkRe = /^.*(youtu.be\/|v\/|e\/|u\/\w+\/|embed\/|v=)([^#\&\?]*).*/;
+        var youtubeVideoUrlRe = /^(?:https?:\/\/)?(?:youtu\.be|(?:www\.)?youtube(?:-nocookie)?\.com)\/(?:|user\/[^\/?#]+)?(?:|.*?[\/=])([a-zA-Z0-9_-]{11})\b/;
```
ahsu1230: Thanks for the tip @2is10 
from https://github.com/kifi/keepit/pull/12517

2is10: Nice. No tests?

ahsu1230: oh. yeaaaa that'd be a good idea

kifi-backend/modules/shoebox/angular/test/unit/common/util.spec.js
```diff
@@ -37,4 +37,24 @@ describe('util', function () {
     });
   });
 
+  describe('util.isYoutube', function () {
```
ahsu1230, line 40: Thanks for the tips @yipingliao

andrewconner, line 40: Awesome!

andrewconner:
Replied to asana issue. This works, but also removes users from when the inviter types in their name. Check the issue for more info.

PR #12612, opened 2014-12-05
fix column order of curator library info required by GetResult

joshm1:
@stkem I think this will fix it... issue was get GetResult depends on the order of the columns in the database - wasn't an error in H2 I think because the order of the fields in the case class was the same


stkem:
ohhh, classic. This very problem has cost several people (myself included) quite a bit of time in the past. I should have spotted it, sorry.

PR #12614, opened 2014-12-05
minor YouTube URL code refactoring

2is10:
@ahsu1230

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/common/util.js
+++ b/kifi-backend/modules/shoebox/angular/src/common/util.js
@@ -15,10 +15,6 @@ angular.module('util', [])
       '/': '&#x2F;'
     };
     var youtubeVideoUrlRe = /^(?:https?:\/\/)?(?:youtu\.be|(?:www\.)?youtube(?:-nocookie)?\.com)\/(?:|user\/[^\/?#]+)?(?:|.*?[\/=])([a-zA-Z0-9_-]{11})\b/;
-    function getYoutubeIdFromUrl(url) {
-      var match = url.match(youtubeVideoUrlRe);
-      return match && match[1];
-    }
 
     return {
       startsWith: function (str, prefix) {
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/angular/src/common/util.js
```diff
@@ -70,8 +66,9 @@ angular.module('util', [])
            (/Trident/.exec(navigator.userAgent) != null))
         );
       },
-      isYoutubeUrl: function (url) {
-        return getYoutubeIdFromUrl(url);
+      getYoutubeIdFromUrl: function (url) {
+        var match = url.match(youtubeVideoUrlRe);
+        return match && match[1];
       },
```
2is10, line 72: The name `isYoutubeUrl` sets the expectation that the function will return `true`/`false`. I think it’s important for function names to convey or hint at their return types. We could go with `getYoutubeIdFromUrlIfTheUrlIsIndeedAYouTubeVideoUrlOrElseJustReturnNull` but I’ve compromised.

kifi-backend/modules/shoebox/angular/src/keep/keep.js
```diff
@@ -120,16 +120,9 @@ angular.module('kifi')
           return text;
         }
 
-        function isYoutubeCard(url) {
-          var youtubeId = util.isYoutubeUrl(url);
-          if (youtubeId) {
-            scope.youtubeId = youtubeId;
-          }
-          return scope.youtubeId;
-        }
```
2is10, line 129: One rule of thumb that I’ve really come to value over the years is that a function should _either_ return a value _or_ have a side effect—not both. It makes for much clearer code.

Notice that `isYoutubeCard` was both returning a value and modifying `scope.youtubeId`. The side effect was surprising given the predicate-style name.

joshm1, line 129: :+1:

kifi-backend/modules/shoebox/angular/src/common/util.js
ahsu1230, line 72: lol =)

kifi-backend/modules/shoebox/angular/src/keep/keep.js
ahsu1230, line 129: :+1:  I see

ahsu1230:
Thanks for the cleanup! looks a lot better =)

kifi-backend/modules/shoebox/angular/src/keep/keep.js
andrewconner, line 129: Very nice.

PR #12616, opened 2014-12-05
deal with canonical url escaping

eishay:


eishay:
@2is10 @raymondng @andrewconner @stkem

andrewconner:
Looks good.

kifi-backend/modules/scraper/app/com/keepit/scraper/extractor/Extractor.scala
```diff
@@ -29,6 +31,12 @@ trait Extractor {
       case _ => None
     }
   }
+  private def escapedDestination(destinationUrl: String, url: String): Boolean = {
+    val escaped = escapeHtml4(destinationUrl)
+    if (escaped == url) return true
+    val doubleEscaped = escapeHtml4(escaped)
+    doubleEscaped == url
+  }
```
2is10, line 39: This works, but only one escaping and comparison should be necessary. (We already unescape once when we read the attribute value.)

It occurs to me that there are many different equivalent strings that are partially or fully HTML-escaped, whereas _unescaping_ is a unifying operation, so it would be better to check this in the opposite direction. I’ll send a pull request.

PR #12620, opened 2014-12-05
refactor curator lib gen & add statsd for generation

joshm1:
@stkem


joshm1:
also fixes/improves how curator_lib_reco seq's are saved. the previous method was flawed in multiple ways

PR #12621, opened 2014-12-05
using HTML unescaping instead of escaping, and only doing it once

2is10:
@eishay 
cc @andrewconner @stkem @raymondng


andrewconner:
Looks good.

2is10:
I just added a second small commit to:
- avoid the unescape operation in the common case
- ignore the bad canonical URL entirely instead of using it

eishay:
Thanks!

PR #12627, opened 2014-12-05
fixing broken pane on some pages due to initial show transition never running

2is10:
cc @ahsu1230


2is10:
This happens because some pages manage to block all rendering while loading slow scripts—even after document is ready.

ahsu1230:
thanks!

andrewconner:
Nice

PR #12630, opened 2014-12-05
fix debug option for library search

ymatsuda:
@leogrim 
This should fix the debug option for Library Search.

A new field `debug' is added to LibrarySearchRequest without any compatibility support. So, library search may fail during deployment.


leogrim:
Thanks!

PR #12635, opened 2014-12-05
collections -> hashTags

ahsu1230:
@jaredpetker


ahsu1230:
only using 'hashTags' now for both requests & responses @thassss

PR #12637, opened 2014-12-05
3.3.27: fixing JS error when formatting certain hits on google.com

2is10:


2is10:
By the way, @mrbrown14, this fixes the error we saw on my computer the other day when we were talking about search results (`cutPoint` and such).

PR #12639, opened 2014-12-05
#cAmElCaSe4LyFe

ahsu1230:
hashTags -> hashtags @jaredpetker


2is10:
what’s wrong with `tags`?

ahsu1230:
I think maybe to differentiate away from the old `collections`/`tags`? Since, technically, hashtags now only work with string names

2is10:
I see. We made that migration in the extension back in Sept or Oct and still use `tags` in the API calls—both for tagging/untagging and in search results (e.g. `"tags": ["Design", "Sketch Tutorial"]`) and in routes (e.g. `.../tags/...`).

I know Léo introduced a `Hashtag` type server-side for clarity, possibly b/c `Tag` could conflict with other types (e.g. `com.keepit.common.mail.template.Tag`) but in my opinion we haven’t fundamentally changed what these things are. In our UI, we still call them “tags”. We may begin prefixing them with `#` someday, but we haven’t yet.

PR #12644, opened 2014-12-06
refactored lib reco generation, add refreshLibRecos endpoint, and improved lib reco admin page

joshm1:
- allow custom score weights
- remove the redudantly passing around params by moving the generation logic for a user in it's own class
- admin page will in the future allow us to input custom selection parameters
- ability to reset already processed library recommendations
- adjusted default score weights

@stkem


joshm1:
the diff is not very helpful because a lot of code was moved around.  most of the core logic is the same and the tests didn't change (and still pass)

stkem:
:+1:

PR #12651, opened 2014-12-07
User Factory

eishay:


kifi-backend/modules/curator/test/com/keepit/curator/FeedDigestEmailSenderTest.scala
```diff
@@ -89,18 +89,7 @@ class FeedDigestEmailSenderTest extends Specification with CuratorTestInjector w
...
-        pictureName = Some("mrt"))
-      val friend3 = User(id = Some(Id[User](46)), firstName = "Dolly", lastName = "Parton", username = Username("test"), normalizedUsername = "test",
-        pictureName = Some("mrt"))
-      val friend4 = User(id = Some(Id[User](47)), firstName = "Benedict", lastName = "Arnold", username = Username("test"), normalizedUsername = "test",
-        pictureName = Some("mrt"))
-      val friend5 = User(id = Some(Id[User](48)), firstName = "Winston", lastName = "Churchill", username = Username("test"), normalizedUsername = "test",
-        pictureName = Some("mrt"))
-      val friend6 = User(id = Some(Id[User](49)), firstName = "Bob", lastName = "Marley", username = Username("test"), normalizedUsername = "test",
-        pictureName = Some("0"))
+      val (friend1, friend2, friend3, friend4, friend5, friend6) = (UserFactory().get, UserFactory().get, UserFactory().get, UserFactory().get, UserFactory().get, UserFactory().get)
```
eishay, line 92: optionally can do a for loop

2is10, line 92:
``` scala
val Seq(f1, f2, f3, f4, f5, f6) = Seq.fill(6)(UserFactory().get)
```

eishay, line 92: Thanks
https://github.com/kifi/keepit/pull/12652/files#diff-346cddc1bc0083558c3082928e874008R15

PR #12664, opened 2014-12-08
fixing neverblue trackers

stkem:


kifi-backend/modules/shoebox/public/js/install.js
```diff
@@ -20,8 +20,24 @@
...
+        xhrFields: {
+          withCredentials: true
+        }
+      });
+      $.ajax({
+        url: "https://byvue.com/p.ashx?o=35028&f=js&t=INSTALL_PAGE_CHROME",
+        xhrFields: {
+          withCredentials: true
+        }
+      });
```
2is10, line 40: FYI, for less repetition, default options can be set just once for the page.

``` js
$.ajaxSetup({xhrFields: {withCredentials: true}});
...
$.get("...");
$.get("...");
$.get("...");
```

or there’s:

``` js
var opts = {xhrFields: {withCredentials: true}};
$.ajax($.extend({url: "..."}, opts));
$.ajax($.extend({url: "..."}, opts));
$.ajax($.extend({url: "..."}, opts));
```

PR #12665, opened 2014-12-08
rudimentary library quality filter for social tooltip

2is10:
@eishay 
cc @leogrim


2is10:
This is temporary, until search filters out crummy public libraries.

Merging so I can use this. Comments appreciated.

leogrim:
:+1:

PR #12666, opened 2014-12-08
Public library search no mobile

yipingliao:
@andrewconner Same as the last pull, with one more commit that takes out mobile search code. "Follow" button still jumps slightly as you observed. This is caused by adding some timeouts to `sticky.js`, which were needed because otherwise the library header will not show up in mobile at all, as we saw on Friday. :(


andrewconner:
Any way to make mobile just not have whatever causes the header not to be there?

andrewconner:
i.e., how it is how in production?

yipingliao:
Okay, public library search sans mobile is finally out and not obviously breaking anything. _phew_ I am going to look into the Follow Button/kfSticky issue this upcoming weekend, out of band. It's bugging me but it's probably not a good idea to spend more time on it right now.

PR #12667, opened 2014-12-08
preparation for removing some callbacks

ymatsuda:
@stkem @leogrim @andrewconner @raymondng @eishay 

This is a preparation for removing some callbacks from scraper (uri summary related ones).

In the original code, http accesses (Embedly and image files) and db accesses (shoebox callbacks) were intermingled. PageInfo was saved before fetching images. and the execution was paused until the call comes back. I believe it is unnecessary to wait for shoebox.

In the new code, db calls are made at the top level, `fetchFromEmbedly`. Embedly call and image fetches happen first. There is no shoebox call until everything is fetched. Then, shoebox is called without waiting for completion. `URISummary` is constructed and returned immediately.  This means that even when `URISummary` is returned, it is not guaranteed that `PageInfo` and `ImageInfo` are persisted successfully. I think this is safe.

We should improve the response time as the result of async db write.
The next step is to use `fetchPageInfoAndImageInfo` instead of `fetchFromEmbedly`, and let shoebox construct URISummary and save `PageInfo` and `ImageInfo` to db.


andrewconner:
Nice, looks good

PR #12675, opened 2014-12-09
user_used_kifi events for digest/message emails

joshm1:
@stkem 

Tests included for the sendgrid commander to create the new events but did not test the actual links in the emails. There are 0 tests for the code I modified for messages; could take some time to create these.  Thoughts?


stkem:
It makes me sad to say it, but we don't have time to add any kind of descent test coverage there right now. Your change is fairly minor, so I think making one run through in prod will have to do here.

stkem:
Code looks good to me

joshm1:
Thanks, I was thinking the same thing.  Changes are minor, but regardless I always get uneasy making changes, especially with code I'm less familiar with, w/o some tests.

PR #12676, opened 2014-12-10
similar libraries

yingjieMiao:
@stkem @joshm1 

just an endpoint. client endpoints later.


stkem:
Nice! Could you add some statsd timing to this, since it is going to be fairly performance sensitive once in front of users?

yingjieMiao:
Sure. I can do that in the controller.

joshm1:
Looks good!

PR #12678, opened 2014-12-10
easier testing setup blocks

eishay:


eishay:
@andrewconner @ahsu1230

PR #12681, opened 2014-12-10
Commander + Repo level: Library Images!

ahsu1230:
Main Changes:
- introduce `BaseImage` which `KeepImage` & `LibraryImage` extend from. With `BaseImage`, a lot of code used in `KeepImageCommander` that isn't `KeepImage` specific can just be applied to any `BaseImage` (less duplication of code). Those functions are moved in `ProcessedImageHelper` which basically works with only classes of `BaseImage`. Anything in `LibraryImageCommander` or `KeepImageCommander` should only have logic regarding `LibraryImage` and `KeepImage` respectively
- create `LibraryImage, LibraryImageCommander, LibraryImageRepo, LibraryImageRequest, LibraryImageRequestRepo, LibraryImageStore`
- `LibraryImage` also has `(centerX, centerY, selectedWidth, selectedHeight)`
- `LibraryImageCommanderTest`
- evolution for library_image & library_image_request

New things: LibraryImage, LibraryImageCommander, 

@andrewconner @2is10


kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepImageCommander.scala
```diff
@@ -325,86 +322,6 @@ class KeepImageCommanderImpl @Inject() (
     }
   }
 
-  private def buildPersistSet(sourceImage: ImageProcessState.ImageLoadedAndHashed): Either[KeepImageStoreFailure, Set[ImageProcessState.ReadyToPersist]] = {
```
ahsu1230, line 328: moved to ProcessedImageHelper. Is used for both KeepImage & LibraryImage

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepImageCommander.scala
```diff
@@ -439,31 +356,3 @@ class KeepImageCommanderImpl @Inject() (
   }
 
 }
-
```
ahsu1230, line 442: moved to BaseImage since both KeepImages & LibraryImages could use these states

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryImageCommander.scala
```diff
@@ -0,0 +1,263 @@
...
+
+  def getUrl(libraryImage: LibraryImage): String = {
+    s3ImageConfig.cdnBase + "/" + libraryImage.imagePath
+  }
+
+  def getBestImageForLibrary(libraryId: Id[Library], idealSize: ImageSize): Option[LibraryImage] = {
+    val validLibraryImages = db.readOnlyMaster { implicit s =>
+      libraryImageRepo.getForLibraryId(libraryId)
+    }
+    ProcessedImageSize.pickBestImage(idealSize, validLibraryImages).asInstanceOf[Option[LibraryImage]]
```
ahsu1230: pickBestImage() returns an `Option[BaseImage]`, but we want a `Option[LibraryImage]` so
I believe this is the way to turn cast a parent class to its extended class?

kifi-backend/modules/shoebox/app/com/keepit/model/KeepImage.scala
```diff
@@ -15,17 +15,18 @@ case class KeepImage(
...
-  val imageSize = ImageSize(width, height)
+    isOriginal: Boolean) extends BaseImage with Model[KeepImage] {
+  //val imageSize = ImageSize(width, height)
   def withId(id: Id[KeepImage]) = copy(id = Some(id))
   def withUpdateTime(now: DateTime) = copy(updatedAt = now)
 }
 
 object KeepImageStates extends States[KeepImage]
 
+/*
```
ahsu1230: whoops removing this

yingjieMiao:
wow, this is big. My personal taste is to ship self-contained and deployable 'small' modules. 

For me, it's always a bit challenging to do code review for PR over 600 lines of code. It's harder to capture bugs for pull request this size. 

Just my 2 cents.

ahsu1230:
Yea, you're right. Sorry about that. =( 
I had trouble figuring out the overall picture as I was going back & forth in between the commander & repo levels.

yingjieMiao:
No worries. Sometime this happens inevitably.

kifi-backend/modules/common/conf/evolutions/shoebox/271.sql
```diff
@@ -0,0 +1,46 @@
...
+  image_format varchar(16) NOT NULL,
+  width smallint(6) unsigned NOT NULL,
+  height smallint(6) unsigned NOT NULL,
+  center_x smallint(6) unsigned NOT NULL,
+  center_y smallint(6)  unsigned NOT NULL,
+  selected_width smallint(6) unsigned NOT NULL,
+  selected_height smallint(6) unsigned NOT NULL,
+  source varchar(32) NOT NULL,
+  source_file_hash varchar(32) DEFAULT NULL,
+  source_image_url varchar(3072) DEFAULT NULL,
```
andrewconner: This currently has no meaning, since users can't give us a URL. Are you using it?

kifi-backend/modules/common/conf/evolutions/shoebox/271.sql
```diff
@@ -0,0 +1,46 @@
...
+  library_id bigint(20) NOT NULL,
+  image_path varchar(64) NOT NULL,
+  image_format varchar(16) NOT NULL,
+  width smallint(6) unsigned NOT NULL,
+  height smallint(6) unsigned NOT NULL,
+  center_x smallint(6) unsigned NOT NULL,
+  center_y smallint(6)  unsigned NOT NULL,
+  selected_width smallint(6) unsigned NOT NULL,
+  selected_height smallint(6) unsigned NOT NULL,
+  source varchar(32) NOT NULL,
```
andrewconner, line 17: Is this used / needed?

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepImageCommander.scala
```diff
@@ -69,7 +68,7 @@ class KeepImageCommanderImpl @Inject() (
       None
     } else Some {
       val validKeepImages = keepImages.filter(_.state == KeepImageStates.ACTIVE)
-      ProcessedImageSize.pickBest(idealSize, validKeepImages)
+      ProcessedImageSize.pickBestImage(idealSize, validKeepImages).asInstanceOf[Option[KeepImage]]
```
andrewconner: `asInstanceOf` should be avoided in all\* cases. Let's figure out a better API. Perhaps generic typing?
- Very rarely, especially in serialization, it may be permitted. Otherwise, it's not needed.

kifi-backend/modules/shoebox/app/com/keepit/commanders/ProcessedImageHelper.scala
```diff
@@ -28,6 +29,88 @@ trait ProcessedImageHelper {
...
+          case Failure(ex) =>
+            Left(BaseImageProcessState.HashFailed(ex))
+        }
+    }.recover {
+      case ex: Throwable =>
+        Left(BaseImageProcessState.SourceFetchFailed(ex))
+    }
+  }
+
+  def buildPersistSet(sourceImage: BaseImageProcessState.ImageLoadedAndHashed, baseLabel: String)(implicit photoshop: Photoshop): Either[BaseImageStoreFailure, Set[BaseImageProcessState.ReadyToPersist]] = {
```
andrewconner: +1, thanks for keeping it generic

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryImageCommander.scala
```diff
@@ -0,0 +1,263 @@
...
+import com.keepit.model._
+import play.api.libs.Files.TemporaryFile
+import play.api.libs.concurrent.Execution.Implicits.defaultContext
+
+import scala.concurrent.Future
+import scala.concurrent.duration.DurationInt
+import scala.util.{ Failure, Success }
+
+@ImplementedBy(classOf[LibraryImageCommanderImpl])
+trait LibraryImageCommander {
```
andrewconner, line 20: Very good API. Super easy to understand. +1

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryImageCommander.scala
andrewconner: As commented above, this can be avoided.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryImageCommander.scala
```diff
@@ -0,0 +1,263 @@
...
+
+  private def fetchAndSet(imageFile: TemporaryFile, libraryId: Id[Library], selection: LibraryImageSelection, source: BaseImageSource, overwriteExistingImage: Boolean)(implicit requestId: Option[Id[LibraryImageRequest]]): Future[BaseImageProcessDone] = {
+    runFetcherAndPersist(libraryId, selection, source, overwriteExistingImage)(fetchAndHashLocalImage(imageFile))
+  }
+
+  private def runFetcherAndPersist(libraryId: Id[Library], selection: LibraryImageSelection, source: BaseImageSource, overwriteExistingImage: Boolean)(fetcher: => Future[Either[BaseImageStoreFailure, BaseImageProcessState.ImageLoadedAndHashed]])(implicit requestId: Option[Id[LibraryImageRequest]]): Future[BaseImageProcessDone] = {
+    val existingImagesForLibrary = db.readOnlyMaster { implicit session =>
+      libraryImageRepo.getForLibraryId(libraryId)
+    }
+    if (existingImagesForLibrary.nonEmpty && !overwriteExistingImage) {
```
andrewconner: `overwriteExistingImage` appears to be always `true` (and it _feels_ like it'll always be that way). `KeepImage`s have some specialness because of auto-fetching and race conditions. I believe you can get rid of `overwriteExistingImage` and not make the `existingImagesForLibrary` call.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryImageCommander.scala
```diff
@@ -0,0 +1,263 @@
...
+          if (existingSameHash.isEmpty || overwriteExistingImage) {
+            // never seen this image, or we're reprocessing an image
+            buildPersistSet(loadedImage, "library")(photoshop) match {
+              case Right(toPersist) =>
+                updateRequestState(LibraryImageRequestStates.PERSISTING)
+                uploadAndPersistImages(loadedImage, toPersist, libraryId, selection, source, overwriteExistingImage)
+              case Left(failure) =>
+                Future.successful(failure)
+            }
+          } else {
```
andrewconner: Never runs?

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryImageCommander.scala
```diff
@@ -0,0 +1,263 @@
...
+    } else {
+      updateRequestState(LibraryImageRequestStates.FETCHING)
+
+      fetcher.flatMap {
+        case Right(loadedImage) =>
+          updateRequestState(LibraryImageRequestStates.PROCESSING)
+          val existingSameHash = db.readOnlyReplica { implicit session =>
+            libraryImageRepo.getBySourceHash(loadedImage.hash)
+          }
+          if (existingSameHash.isEmpty || overwriteExistingImage) {
```
andrewconner: Unnecessary / always runs?

kifi-backend/modules/common/conf/evolutions/shoebox/271.sql
```diff
@@ -0,0 +1,46 @@
...
+  state varchar(20) NOT NULL,
+  library_id bigint(20) NOT NULL,
+  image_path varchar(64) NOT NULL,
+  image_format varchar(16) NOT NULL,
+  width smallint(6) unsigned NOT NULL,
+  height smallint(6) unsigned NOT NULL,
+  center_x smallint(6) unsigned NOT NULL,
+  center_y smallint(6)  unsigned NOT NULL,
+  selected_width smallint(6) unsigned NOT NULL,
+  selected_height smallint(6) unsigned NOT NULL,
```
2is10: I think we decided on rounded percentages for the four cropping rectangle values, so they can be `tinyint(3)`. We can always add precision later if we want more.

Also, you might consider these names to clarify that all four values are related:
 `crop_center_x`
 `crop_center_y`
 `crop_width`
 `crop_height`

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryImageCommander.scala
```diff
@@ -0,0 +1,263 @@
...
+    }
+    if (existingImagesForLibrary.nonEmpty && !overwriteExistingImage) {
+      Future.successful(BaseImageProcessState.ExistingStoredImagesFound(existingImagesForLibrary))
+    } else {
+      updateRequestState(LibraryImageRequestStates.FETCHING)
+
+      fetcher.flatMap {
+        case Right(loadedImage) =>
+          updateRequestState(LibraryImageRequestStates.PROCESSING)
+          val existingSameHash = db.readOnlyReplica { implicit session =>
```
andrewconner: Not used / never needed because of `overwriteExistingImage=true`?

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryImageCommander.scala
```diff
@@ -0,0 +1,263 @@
...
+            Future.successful(copyExistingImagesAndReplace(libraryId, selection, source, existingSameHash))
+          }
+        case Left(failure) => Future.successful(failure)
+      }
+    }
+  }
+
+  private def uploadAndPersistImages(originalImage: BaseImageProcessState.ImageLoadedAndHashed, toPersist: Set[BaseImageProcessState.ReadyToPersist], libraryId: Id[Library], selection: LibraryImageSelection, source: BaseImageSource, overwriteExistingImage: Boolean): Future[BaseImageProcessDone] = {
+    val uploads = toPersist.map { image =>
+      log.info(s"[kic] Persisting ${image.key} (${image.bytes} B)")
```
andrewconner: All `[kic]`s can be replaced since this isn't a keep image :)

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryImageCommander.scala
```diff
@@ -0,0 +1,263 @@
...
+        }.isEmpty
+      }
+
+      log.info("[kic] Activating:" + toActivate.map(_.imagePath) + "\nDeactivating:" + toDeactivate.map(_.imagePath) + "\nCreating:" + toCreate.map(_.imagePath))
+      toDeactivate.foreach(libraryImageRepo.save)
+      (toActivate ++ toCreate).map(libraryImageRepo.save)
+    }
+  }
+
+  private def copyExistingImagesAndReplace(libraryId: Id[Library], selection: LibraryImageSelection, source: BaseImageSource, existingSameHash: Seq[LibraryImage]) = {
```
andrewconner: Not needed?

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryImage.scala
```diff
@@ -0,0 +1,36 @@
...
+    isOriginal: Boolean) extends BaseImage with Model[LibraryImage] {
+
+  val imageSelection = LibraryImageSelection(centerX, centerY, selectedWidth, selectedHeight)
+  def withId(id: Id[LibraryImage]) = copy(id = Some(id))
+  def withUpdateTime(now: DateTime) = copy(updatedAt = now)
+}
+
+case class LibraryImageSelection(
+  centerX: Int, centerY: Int,
+  selectedWidth: Int, selectedHeight: Int)
```
2is10: `selectedWidth` → `width`
`selectedHeight` → `height`
(“selected” is implied by the type name)

2is10: Also, we use the term “position” in the route `/site/library/:id/image/position` and in the JSON field name `"imagePos"`. It would probably be clearest to be consistent with our terminology and name this `LibraryImagePosition`.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryImageCommander.scala
```diff
@@ -0,0 +1,263 @@
...
+import scala.concurrent.duration.DurationInt
+import scala.util.{ Failure, Success }
+
+@ImplementedBy(classOf[LibraryImageCommanderImpl])
+trait LibraryImageCommander {
+
+  def getUrl(libraryImage: LibraryImage): String
+  def getBestImageForLibrary(libraryId: Id[Library], idealSize: ImageSize): Option[LibraryImage]
+  def setLibraryImageFromFile(image: TemporaryFile, libraryId: Id[Library], imageSizing: LibraryImageSelection, source: BaseImageSource, requestId: Option[Id[LibraryImageRequest]] = None): Future[BaseImageProcessDone]
+  def setLibraryImageSizing(libraryId: Id[Library], imageSizing: LibraryImageSelection): Seq[LibraryImage]
```
2is10: Okay if we use the term “position” instead of “sizing” and/or “selection” here and throughout?

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryImageCommander.scala
2is10: You can have `pickBestImage` accept any subclass of `BaseImage` and return the same type. Then no casting is necessary.

andrewconner: Yep. `T <: BaseImage`

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryImageCommander.scala
```diff
@@ -0,0 +1,263 @@
...
+      libraryImageRepo.getForLibraryId(libraryId).map { libImage =>
+        libraryImageRepo.save(libImage.copy(state = LibraryImageStates.INACTIVE))
+      }
+    }
+    images.nonEmpty
+  }
+
+  private def fetchAndSet(imageFile: TemporaryFile, libraryId: Id[Library], selection: LibraryImageSelection, source: BaseImageSource, overwriteExistingImage: Boolean)(implicit requestId: Option[Id[LibraryImageRequest]]): Future[BaseImageProcessDone] = {
+    runFetcherAndPersist(libraryId, selection, source, overwriteExistingImage)(fetchAndHashLocalImage(imageFile))
+  }
```
2is10: I don’t think we’ll have a position at the time of the upload. Probably better to make the position optional, and be absent initially (i.e. removed as an argument to this `fetchAndSet` method, `null` in the database). We could even have a non-`ACTIVE` state for an uploaded library image that has not yet been positioned, if we want, like `UPLOADED` or `UNPOSITIONED`.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryImageCommander.scala
```diff
@@ -0,0 +1,263 @@
...
+      }
+      val toActivate = shouldBeActive.filter(_.state != LibraryImageStates.ACTIVE).map(_.copy(state = LibraryImageStates.ACTIVE))
+      val toDeactivate = shouldBeInactive.filter(_.state != LibraryImageStates.INACTIVE).map(_.copy(state = LibraryImageStates.INACTIVE))
+      val toCreate = newLibraryImages.filter { newImg =>
+        oldLibraryImages.find { existingImg =>
+          existingImg.sourceFileHash == newImg.sourceFileHash && existingImg.width == newImg.width && existingImg.height == newImg.height
+        }.isEmpty
+      }
+
+      log.info("[kic] Activating:" + toActivate.map(_.imagePath) + "\nDeactivating:" + toDeactivate.map(_.imagePath) + "\nCreating:" + toCreate.map(_.imagePath))
```
2is10: `[kic]` → `[lic]` throughout this file (for `LibraryImageCommander`)

kifi-backend/modules/shoebox/app/com/keepit/commanders/ProcessedImageHelper.scala
```diff
@@ -28,6 +29,88 @@ trait ProcessedImageHelper {
   val log: Logger
   val webService: WebService
 
+  val originalLabel: String = "_o"
+
+  def fetchAndHashLocalImage(file: TemporaryFile): Future[Either[BaseImageStoreFailure, BaseImageProcessState.ImageLoadedAndHashed]] = {
+    log.info(s"[kic] Fetching ${file.file.getAbsolutePath}")
```
2is10: `[kic]` → `[pih]` throughout this file?

kifi-backend/modules/common/conf/evolutions/shoebox/271.sql
ahsu1230, line 17: well for LibraryImage it is always `user_upload`. But I figured we could leave it in for future use (users can pick a keep image to be a library image?)

kifi-backend/modules/common/conf/evolutions/shoebox/271.sql
ahsu1230: Nope, not used. It's always None. I'll go ahead and remove it.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryImageCommander.scala
ahsu1230: I see, thanks for that tip.

kifi-backend/modules/common/conf/evolutions/shoebox/271.sql
ahsu1230: that makes more sense, thanks!

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryImageCommander.scala
ahsu1230: :+1:

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepImageCommander.scala
```diff
@@ -31,13 +31,12 @@ trait KeepImageCommander {
   def getBestImagesForKeeps(keepIds: Set[Id[Keep]], idealSize: ImageSize): Map[Id[Keep], Option[KeepImage]]
   def getExistingImageUrlForKeepUri(nUriId: Id[NormalizedURI])(implicit session: RSession): Option[String]
 
-  def autoSetKeepImage(keepId: Id[Keep], localOnly: Boolean = true, overwriteExistingChoice: Boolean = false): Future[ImageProcessDone]
-  def setKeepImageFromUrl(imageUrl: String, keepId: Id[Keep], source: KeepImageSource, requestId: Option[Id[KeepImageRequest]] = None): Future[ImageProcessDone]
-  def setKeepImageFromFile(image: TemporaryFile, keepId: Id[Keep], source: KeepImageSource, requestId: Option[Id[KeepImageRequest]] = None): Future[ImageProcessDone]
+  def autoSetKeepImage(keepId: Id[Keep], localOnly: Boolean = true, overwriteExistingChoice: Boolean = false): Future[BaseImageProcessDone]
+  def setKeepImageFromUrl(imageUrl: String, keepId: Id[Keep], source: BaseImageSource, requestId: Option[Id[KeepImageRequest]] = None): Future[BaseImageProcessDone]
+  def setKeepImageFromFile(image: TemporaryFile, keepId: Id[Keep], source: BaseImageSource, requestId: Option[Id[KeepImageRequest]] = None): Future[BaseImageProcessDone]
```
2is10: I would recommend only adding the word `Base` to the names of types that are actually subclassed. This code would read better if `BaseImageSource` were instead `KeepImageSource` or just `ImageSource` (depending on whether the type is keep-specific) and `BaseImageProcessDone` remained `ImageProcessDone`.

Abstract types should be as invisible as possible, ideally only referenced in the definitions of a subclasses and generic methods.

kifi-backend/modules/shoebox/app/com/keepit/commanders/ProcessedImageHelper.scala
```diff
@@ -210,7 +293,7 @@ object ProcessedImageSize {
     allSizes.map(s => s -> maxDivergence(s.idealSize, size)).sortBy(_._2).head._1
   }
 
-  def pickBest(idealSize: ImageSize, images: Seq[KeepImage]): Option[KeepImage] = {
+  def pickBestImage[T <: BaseImage](idealSize: ImageSize, images: Seq[T]): Option[T] = {
```
ahsu1230, line 296: now uses generic typing to accept subclasses of `BaseImage` and return that type

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryImageCommander.scala
```diff
@@ -0,0 +1,214 @@
...
+
+  def getUrl(libraryImage: LibraryImage): String = {
+    s3ImageConfig.cdnBase + "/" + libraryImage.imagePath
+  }
+
+  def getBestImageForLibrary(libraryId: Id[Library], idealSize: ImageSize): Option[LibraryImage] = {
+    val validLibraryImages = db.readOnlyMaster { implicit s =>
+      libraryImageRepo.getForLibraryId(libraryId)
+    }
+    ProcessedImageSize.pickBestImage(idealSize, validLibraryImages)
```
ahsu1230, line 50: no more `InstanceOf` =D

kifi-backend/modules/shoebox/app/com/keepit/commanders/ProcessedImageHelper.scala
andrewconner, line 296: Nice, great job.

andrewconner:
I'll go through this again in a little bit as a sanity check. Thanks for all the updates.

ahsu1230:
Thanks @2is10 @andrewconner for all the checks & tips!

2is10:
Of course. Excellent work!​

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryImageCommander.scala
```diff
@@ -0,0 +1,214 @@
...
+import play.api.libs.concurrent.Execution.Implicits.defaultContext
+
+import scala.concurrent.Future
+
+@ImplementedBy(classOf[LibraryImageCommanderImpl])
+trait LibraryImageCommander {
+
+  def getUrl(libraryImage: LibraryImage): String
+  def getBestImageForLibrary(libraryId: Id[Library], idealSize: ImageSize): Option[LibraryImage]
+  def setLibraryImageFromFile(image: TemporaryFile, libraryId: Id[Library], imagePos: LibraryImagePosition, source: ImageSource, requestId: Option[Id[LibraryImageRequest]] = None): Future[ImageProcessDone]
```
ahsu1230: whoops, may still need to add Option[LibraryImagePosition] to signature.

ahsu1230:
I just made positions in `LibraryImageRepo` to be nullable. So `LibraryImage` will have optional positions.
This is because when somebody uploads an image, a position set won't be ready at that point.

Users may not want to apply the library image even after uploading. So the position could be null & we can later set the positions using a "setPosition" endpoint. Thanks @2is10 for the discussion

ahsu1230:
I'll merge soon & apply evolutions so this PR doesn't get any bigger. Would be happy to address comments after.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryImageCommander.scala
```diff
@@ -0,0 +1,214 @@
...
+  }
+
+  def getBestImageForLibrary(libraryId: Id[Library], idealSize: ImageSize): Option[LibraryImage] = {
+    val validLibraryImages = db.readOnlyMaster { implicit s =>
+      libraryImageRepo.getForLibraryId(libraryId)
+    }
+    ProcessedImageSize.pickBestImage(idealSize, validLibraryImages)
+  }
+
+  def setLibraryImageFromFile(image: TemporaryFile, libraryId: Id[Library], position: Option[LibraryImagePosition], source: ImageSource, requestId: Option[Id[LibraryImageRequest]]): Future[ImageProcessDone] = {
```
2is10: I think this method and the private helpers it calls should not take a `position` argument at all, since there’s no actual use case for it.

ahsu1230: yup just removed it!

ahsu1230:
positionX and positionY are now optional for LibraryImage

ahsu1230:
evolutions applied onto mysql

2is10:
Yay! Great job.

ahsu1230:
yahooooo build passss lol

PR #12683, opened 2014-12-10
make sure inactive library has inactive lda feature

yingjieMiao:
@stkem @joshm1

### DIFF
--- a/kifi-backend/modules/cortex/app/com/keepit/cortex/models/lda/LDALibraryUpdater.scala
+++ b/kifi-backend/modules/cortex/app/com/keepit/cortex/models/lda/LDALibraryUpdater.scala
@@ -48,6 +48,10 @@ class LDALibraryUpdaterImpl @Inject() (
    * main update() logic is driven by keep's seqNum. This cleanup process is driven by Library's seqNum. This is mostly designed for:
    * library changes state from active to inactive or vice versa. Such event is not caputured from keep events, we want to
    * make sure inactive library has inactive library LDA feature.
+   * NOTE: In update(), we never bring an inactive library's feature from inactive to active, and for an active library, we never bring
+   * its feature state from active to inactive. In cleanup(), the only mutation is switching active and inactive state of library features.
+   * Thus, update() and cleanup() perform "disjoint" mutations. This is critical, because we are listening on two seqNums, and the two seqNum
+   * may proceed at different speed. We don't want mutation from one method is overwritten by the other method, in an unexpected way.
    */
   private def cleanup(version: ModelVersion[DenseLDA]): Unit = {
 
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/cortex/app/com/keepit/cortex/models/lda/LDALibraryUpdater.scala
```diff
@@ -39,6 +40,51 @@ class LDALibraryUpdaterImpl @Inject() (
...
+    val commitOpt = db.readOnlyReplica { implicit s => commitRepo.getByModelAndVersion(cleanupSeq, version.version) }
+    if (commitOpt.isEmpty) db.readWrite { implicit s => commitRepo.save(FeatureCommitInfo(modelName = cleanupSeq, modelVersion = version.version, seq = 0L)) }
+    val fromSeq = SequenceNumber[CortexLibrary](commitOpt.map { _.seq }.getOrElse(0L))
+    val libs = db.readOnlyReplica { implicit s => libRepo.getSince(fromSeq, fetchSize) }
+
+    // make sure non-system library has consitent library state and library feature state
+    libs.filter { x => x.kind != LibraryKind.SYSTEM_MAIN && x.kind != LibraryKind.SYSTEM_SECRET }
+      .foreach { lib =>
+        db.readWrite { implicit s =>
+          val libFeatOpt = libraryLDARepo.getByLibraryId(lib.libraryId, version)
```
joshm1, line 68: possible to do a batch query for all library IDs that pass the filter on line 65 to get these libFeat objects so we're not doing as many as `fetchSize` queries to get this `libFeatOpt` object?

joshm1:
Other than my minor note, LGTM! :+1:

kifi-backend/modules/cortex/app/com/keepit/cortex/models/lda/LDALibraryUpdater.scala
yingjieMiao, line 68: It's possible. But I would like to reuse existing endpoint unless performance become a problem. currently we have 27K libs and only 3K of them are user_created. In some sense, the density is pretty low. And in practice, library seqNum is not increased very frequently (keeping to library doesn't bump up seqNum). So this cleanup() method should not be very busy.

I will definitely keep an eye on it when I deploy (it will create some db write spikes)

yingjieMiao, line 68: I will use a different fetch size for the cleanup process. The update() process was listening on keeps, it used to be 2000, I bumped up to 10000 for catching up quickly when I deployed new model.

PR #12685, opened 2014-12-10
making library card template a little smaller by not passing scope.library to helper methods

2is10:
@yipingliao (trying to avoid overloading Andrew)

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
@@ -254,14 +254,14 @@ angular.module('kifi')
         //
         // Scope methods.
         //
-        scope.acceptInvitation = function (library) {
-          scope.followLibrary(library);
+        scope.acceptInvitation = function () {
+          scope.followLibrary();
         };
 
-        scope.ignoreInvitation = function (library) {
-          if (library.invite) {
-            libraryService.declineToJoinLibrary(library.id).then(function () {
-              library.invite.actedOn = true;
+        scope.ignoreInvitation = function () {
+          if (scope.library.invite) {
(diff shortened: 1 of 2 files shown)


yipingliao:
I've seen this done both ways, and have been opting for passing in the arguments into scope functions, even if the arguments are already on the scope itself. I like this way better because it's easier to see what the function depends on.

Here is a stack overflow discussion on the issue:
http://stackoverflow.com/questions/18967127/angularjs-when-to-pass-scope-variable-to-function

What do you think? If you feel strongly about not passing in scope arguments into scope functions, then we should make that a uniform practice across the code base.

2is10:
My feeling is summarized well in the accepted answer to that stack overflow discussion:

> People from OOP background would think that: we only pass in parameters that don't belong to object. Like a class Person already has hands, we don't need to pass in hands for every object method.

This is the library card directive. Passing a library object into nearly every predicate felt a little weird. It made me wonder whether this card might for some reason be dealing with other libraries too, in addition to the one being rendered.

That being said, I don’t feel strongly about it, and if passing scope properties into scope predicates is the convention in this codebase, then I’ll abide by it. I suppose I didn’t guess that it’s a convention because `scope.canBeShared` uses `scope.isUserLoggedOut` and the handler methods use many scope properties—both reading and writing to them. (Predicates and handlers are currently hard to distinguish at a glance.)

yipingliao:
:+1: 
Okay, let's stop passing in scope variables to scope methods. I'll also change them as I see them from now on.

2is10:
Sounds good.

I’m all for pure functions, but in this case the framework seems neutral or opinionated against them. It could have encouraged them by providing a separate place (scope) for defining template predicates (for `ng-if` and friends).

It’s certainly not urgent to change our codebase. Doing it here just helped me make sense of this template I’m working in a little better.

PR #12686, opened 2014-12-10
renaming class in library card to avoid potential conflict with keep card

2is10:
@yipingliao 

To be clear, because all of the library card rules are scoped with ancestor selectors, there isn’t an actual conflict. This just may help avoid conflicts going forward (e.g. keep card changes unintentionally affecting the library card).

By the way, I’d like to gradually simplify our selectors. Most rules shouldn’t need ancestor scoping.


yipingliao:
Nice!

PR #12687, opened 2014-12-10
using routeService.formatPicUrl for user picture URLs

2is10:
@yipingliao 
diff looks cleaner if you add `?w=0` to the URL


yipingliao:
We had this in so many places... thanks for fixing.
Should we be calling `routeService.formatPicUrl` from now on (instead of the methods in `profileService` and `friendService`)?

2is10:
Thanks for reviewing.

Yes, it may be possible to consolidate further. Right now the wrapper methods all appear to be a bit more specific than `routeService.formatPicUrl`.

PR #12690, opened 2014-12-10
OAuth2Test++ 

raymondng:
Refactor/clean-up of (old) OAuth2 code along the lines of what's already done with token-based code-paths.

Many code-level changes, but the overall approach is not new (already established and working with token-based code paths)

A few goals:
- reduce unnecessary dependency on `SecureSocial` as much as possible
- reduce unnecessary dependency on `[dev|prod]-securesocial.conf`
- cleanly separate parts that are dependent on `SecureSocial` and the core parts (async/non-blocking, more robust and less surprising)
- remove unused/redundant code (e.g. `PimpedOAuthSettings`)
- reduce `UserIdentityProvider` to bare minimum
- improve modularity and test-friendliness
- added (very basic) test for OAuth2-flow (`OAuth2Test`)
- enable local (dev-mode) testing for all social logins (FB/LNKD/TWTR)

Good news is now we can test all social logins locally; and based on that I can ascertain that all social signups/logins are working properly 

Additional help still required after push to prod (this is required until we have reasonable coverage; right now it's very basic)

More work required to bring additional coverage. Plan is to merge this and build on it.

@andrewconner 
cc: @jaredpetker @thassss


raymondng:
Discussed changes in person with @andrewconner. Will push and test.

PR #12703, opened 2014-12-11
nonlinear scoring experiment

ymatsuda:
@yingjieMiao @joshm1 @leogrim @stkem @eishay 

An experimental code
- Recomputes master scores at retrieval time
- A score is split into two parts, the interest part and the social part.
- The interest part and the social part are added together after the social part is weighted by the interest part, i.e., the social part is discounted when the interest scores are low.

### DIFF
--- a/kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationRetrievalCommander.scala
+++ b/kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationRetrievalCommander.scala
@@ -72,6 +72,9 @@ class NonLinearRecoScoringStrategy extends RecoScoringStrategy {
       7 * scores.recentInterestScore +
       4 * scores.libraryInducedScore.getOrElse(0.0f)
     )
+    val normalizer = (5f + 7f + scores.libraryInducedScore.fold(0f)(_ => 4f))
+    val factor = interestPart / normalizer
+
     val socialPart = (
       4 * scores.socialScore +
       2 * scores.priorScore +
@@ -81,7 +84,6 @@ class NonLinearRecoScoringStrategy extends RecoScoringStrategy {
       3 * scores.discoveryScore +
       4 * scores.curationScore.getOrElse(0.0f)
     )
-    val factor = interestPart / (5f + 7f + scores.libraryInducedScore.fold(0f)(_ => 4f)) // normalized to [0, 1]
 
     (interestPart + factor * socialPart) * scores.multiplier.getOrElse(1.0f)
   }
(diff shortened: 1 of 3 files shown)


joshm1:
lgtm! :+1:

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationRetrievalCommander.scala
```diff
@@ -49,22 +49,52 @@ class DiverseRecoSelectionStrategy(val minScore: Float = 5f) extends RecoSelecti
...
+    val socialPart = (
+      4 * scores.socialScore +
+      2 * scores.priorScore +
+      1 * scores.recencyScore +
+      1 * scores.popularityScore +
+      6 * scores.rekeepScore +
+      3 * scores.discoveryScore +
+      4 * scores.curationScore.getOrElse(0.0f)
+    )
+    val factor = interestPart / (5f + 7f + 4f) // normalized to [0, 1]
```
eishay: instead of `(5f + 7f + 4f)`
shouldn't it be like this?

``` scala
(5f + 7f + scores.libraryInducedScore.map(4f).getOrElse(0f)) 
```

eishay:
great!

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationRetrievalCommander.scala
```diff
@@ -49,22 +49,52 @@ class DiverseRecoSelectionStrategy(val minScore: Float = 5f) extends RecoSelecti
...
+      4 * scores.socialScore +
+      2 * scores.priorScore +
+      1 * scores.recencyScore +
+      1 * scores.popularityScore +
+      6 * scores.rekeepScore +
+      3 * scores.discoveryScore +
+      4 * scores.curationScore.getOrElse(0.0f)
+    )
+    val factor = interestPart / (5f + 7f + scores.libraryInducedScore.fold(0f)(_ => 4f)) // normalized to [0, 1]
+
```
yingjieMiao, line 87: cool. Minor comment: How about have

```
val normalizer = (5f + 7f + scores.libraryInducedScore.fold(0f)(_ => 4f)
val factor = interestPart/normalizer
``
after the definition of `interestPart`? IMHO, this removes the need of comment, and make a clear connection of the 'magic consts' in the normalizer to the consts in the interestPart. 
```

yingjieMiao:
:+1: 

Looks like this will affect the score persist in the DB, right?

ymatsuda:
This doesn't affect stored master scores. The master scores in DB are still computed as they are now. So, it is safe to turn on/off the non-linear scoring experiment casually.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationRetrievalCommander.scala
ymatsuda, line 87: got it.

yingjieMiao:
Nice. would be interesting to see results

PR #12705, opened 2014-12-11
library image controller (3 endpoints -> upload, position, remove)

ahsu1230:
LibraryImageController with 3 endpoints
- upload `POST /site/libraries/:id/image/upload?is=wxh&x=40&y=60`
  - request: { file }
  - response: `{ "imagePath" : "library/f293764f93g_640x480.jpg"}`
- position `POST /site/libraries/:id/image/position`
  - request: `{ "imagePath" : "library/f293764f93g_640x480.jpg", "x" : 30, "y": 25 }`
  - response: `NoContent`
- remove `DELETE /site/libraries/:id/image`
  - response: `NoContent`

Biggest Changes to other files:
- `LibraryImageCommander` to fit Controller signatures
- Tests

@andrewconner @2is10

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryImageCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryImageCommander.scala
@@ -21,9 +21,9 @@ import scala.concurrent.Future
 trait LibraryImageCommander {
 
   def getUrl(libraryImage: LibraryImage): String
-  def getBestImageForLibraryAndHash(libraryId: Id[Library], hash: ImageHash, idealSize: ImageSize): Option[LibraryImage]
-  def uploadLibraryImageFromFile(image: TemporaryFile, libraryId: Id[Library], source: ImageSource, requestId: Option[Id[LibraryImageRequest]] = None): Future[ImageProcessDone]
-  def positionLibraryImage(libraryId: Id[Library], imageHash: ImageHash, imagePosition: LibraryImagePosition): Either[(Int, String), Seq[LibraryImage]]
+  def getBestImageForLibrary(libraryId: Id[Library], idealSize: ImageSize): Option[LibraryImage]
+  def uploadLibraryImageFromFile(image: TemporaryFile, libraryId: Id[Library], position: LibraryImagePosition, source: ImageSource, requestId: Option[Id[LibraryImageRequest]] = None): Future[ImageProcessDone]
+  def positionLibraryImage(libraryId: Id[Library], imagePosition: LibraryImagePosition): Seq[LibraryImage]
   def removeImageForLibrary(libraryId: Id[Library]): Boolean // Returns true if images were removed, false otherwise
 }
 
@@ -44,43 +44,33 @@ class LibraryImageCommanderImpl @Inject() (
     s3ImageConfig.cdnBase + "/" + libraryImage.imagePath
   }
 
-  def getBestImageForLibraryAndHash(libraryId: Id[Library], hash: ImageHash, idealSize: ImageSize): Option[LibraryImage] = {
+  def getBestImageForLibrary(libraryId: Id[Library], idealSize: ImageSize): Option[LibraryImage] = {
     val targetLibraryImages = db.readOnlyMaster { implicit s =>
-      libraryImageRepo.getByLibraryIdAndHash(libraryId, hash)
+      libraryImageRepo.getForLibraryId(libraryId)
     }
     ProcessedImageSize.pickBestImage(idealSize, targetLibraryImages)
   }
 
-  def uploadLibraryImageFromFile(image: TemporaryFile, libraryId: Id[Library], source: ImageSource, requestId: Option[Id[LibraryImageRequest]]): Future[ImageProcessDone] = {
-    fetchAndSet(image, libraryId, source)(requestId).map { done =>
+  def uploadLibraryImageFromFile(image: TemporaryFile, libraryId: Id[Library], position: LibraryImagePosition, source: ImageSource, requestId: Option[Id[LibraryImageRequest]]): Future[ImageProcessDone] = {
+    fetchAndSet(image, libraryId, position, source)(requestId).map { done =>
       finalizeImageRequestState(libraryId, requestId, done)
       done
     }
   }
 
-  def positionLibraryImage(libraryId: Id[Library], imageHash: ImageHash, position: LibraryImagePosition): Either[(Int, String), Seq[LibraryImage]] = {
+  def positionLibraryImage(libraryId: Id[Library], position: LibraryImagePosition): Seq[LibraryImage] = {
     db.readWrite { implicit s =>
-      val (toPosition, toDeactivate) = libraryImageRepo.getForLibraryIdOrWithHash(libraryId, imageHash).partition { img =>
-        img.sourceFileHash == imageHash
+      val toPosition = libraryImageRepo.getForLibraryId(libraryId)
+      val positionedImages = toPosition.map { libImage =>
+        val setX = position.x.getOrElse(libImage.positionX.getOrElse(50))
+        val setY = position.y.getOrElse(libImage.positionY.getOrElse(50))
+        libraryImageRepo.save(libImage.copy(
+          positionX = Some(setX),
+          positionY = Some(setY),
+          state = LibraryImageStates.ACTIVE))
       }
-
-      if (toPosition.isEmpty) {
-        Left((NOT_FOUND, "hash_not_found"))
-      } else {
-        toDeactivate.map { libImage =>
-          libraryImageRepo.save(libImage.copy(state = LibraryImageStates.INACTIVE))
-        }
-        val positionedImages = toPosition.map { libImage =>
-          val setX = position.x.getOrElse(libImage.positionX.getOrElse(50))
-          val setY = position.y.getOrElse(libImage.positionY.getOrElse(50))
-          libraryImageRepo.save(libImage.copy(
-            positionX = Some(setX),
-            positionY = Some(setY),
-            state = LibraryImageStates.ACTIVE))
-        }
-        log.info("[lic] Deactivating: " + toDeactivate.map(_.imagePath) + "\nPositioning: " + toPosition.map(_.imagePath))
-        Right(positionedImages)
-      }
+      log.info("[lic] Positioning: " + toPosition.map(_.imagePath))
+      positionedImages
     }
   }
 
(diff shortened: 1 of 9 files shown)


kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryImageController.scala
```diff
@@ -0,0 +1,86 @@
...
+
+  def uploadLibraryImage(pubId: PublicId[Library], imageSize: Option[String] = None) = (UserAction andThen LibraryWriteAction(pubId)).async(parse.temporaryFile) { request =>
+    val libraryId = Library.decodePublicId(pubId).get
+    val imageRequest = db.readWrite { implicit session =>
+      libraryImageRequestRepo.save(LibraryImageRequest(libraryId = libraryId, source = ImageSource.UserUpload))
+    }
+    val uploadImageF = libraryImageCommander.uploadLibraryImageFromFile(request.body, libraryId, ImageSource.UserUpload, Some(imageRequest.id.get))
+    uploadImageF.map {
+      case fail: ImageStoreFailure =>
+        println("**************" + fail.reason)
```
ahsu1230: woops removing

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryImageController.scala
```diff
@@ -0,0 +1,86 @@
...
+  def removeLibraryImage(pubId: PublicId[Library]) = (UserAction andThen LibraryWriteAction(pubId)) { request =>
+    val libraryId = Library.decodePublicId(pubId).get
+    libraryImageCommander.removeImageForLibrary(libraryId)
+    NoContent
+  }
+}
+
+object LibraryImageController {
+  val defaultImageSize = ImageSize(600, 480)
+}
```
ahsu1230: and newline

kifi-backend/modules/shoebox/app/com/keepit/model/BaseImage.scala
```diff
@@ -61,8 +64,8 @@ object ImageProcessState {
   case class CDNUploadFailed(ex: Throwable) extends ImageStoreFailureWithException(ex, "cdn_upload_failed")
 
   // Success
-  case object StoreSuccess extends ImageProcessState with ImageProcessSuccess
-  case class ExistingStoredImagesFound(images: Seq[BaseImage]) extends ImageProcessState with ImageProcessSuccess
+  case class StoreSuccess(hashSet: Set[ImageHash]) extends ImageProcessSuccess(hashSet)
```
ahsu1230: `StoreSuccess` should have a set of successful imageHashes which we can later use to pick out the best image for a library (given libraryId, imageHash, idealSize) - look at LibraryImageController (line 44)

kifi-backend/modules/shoebox/conf/shoebox.routes
```diff
@@ -365,6 +365,10 @@ POST    /site/libraries/:id/importTag   @com.keepit.controllers.website.LibraryC
 POST    /site/libraries/:id/moveTag     @com.keepit.controllers.website.LibraryController.moveKeepsFromCollectionToLibrary(id: PublicId[Library], tag: String)
 POST    /site/libraries/:id/import-file  @com.keepit.controllers.website.BookmarkImporter.importFileToLibrary(id: PublicId[Library])
 
+POST    /site/libraries/:id/image/upload    @com.keepit.controllers.website.LibraryImageController.uploadLibraryImage(id: PublicId[Library], is: Option[String] ?= None)
```
andrewconner: Does the site know the dimensions? @2is10

kifi-backend/modules/shoebox/app/com/keepit/model/BaseImage.scala
andrewconner: Why would there be multiple hashes?

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryImageCommander.scala
```diff
@@ -20,9 +21,9 @@ import scala.concurrent.Future
 trait LibraryImageCommander {
 
   def getUrl(libraryImage: LibraryImage): String
-  def getBestImageForLibrary(libraryId: Id[Library], idealSize: ImageSize): Option[LibraryImage]
```
2is10, line 23: Will we need this method to include the image in JSON responses when libraries are requested?

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryImageCommander.scala
```diff
@@ -116,15 +134,21 @@ class LibraryImageCommanderImpl @Inject() (
           val isOriginal = uploadedImage.key.takeRight(7).indexOf(originalLabel) != -1
           val libImg = LibraryImage(libraryId = libraryId, imagePath = uploadedImage.key, format = uploadedImage.format,
             width = uploadedImage.image.getWidth, height = uploadedImage.image.getHeight, positionX = None, positionY = None,
-            source = source, sourceFileHash = originalImage.hash, isOriginal = isOriginal)
+            source = source, sourceFileHash = originalImage.hash, isOriginal = isOriginal, state = LibraryImageStates.INACTIVE)
```
ahsu1230: Uploading an image no longer auto sets that as the library image. So it must be initiated as inactive

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryImageCommander.scala
```diff
@@ -20,9 +21,9 @@ import scala.concurrent.Future
 trait LibraryImageCommander {
 
   def getUrl(libraryImage: LibraryImage): String
-  def getBestImageForLibrary(libraryId: Id[Library], idealSize: ImageSize): Option[LibraryImage]
-  def setLibraryImageFromFile(image: TemporaryFile, libraryId: Id[Library], source: ImageSource, requestId: Option[Id[LibraryImageRequest]] = None): Future[ImageProcessDone]
-  def positionLibraryImage(libraryId: Id[Library], imagePosition: LibraryImagePosition): Seq[LibraryImage]
+  def getBestImageForLibraryAndHash(libraryId: Id[Library], hash: ImageHash, idealSize: ImageSize): Option[LibraryImage]
```
andrewconner: Why signature change? A library can only have one image, right?

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryImageCommander.scala
```diff
@@ -20,9 +21,9 @@ import scala.concurrent.Future
...
+  def uploadLibraryImageFromFile(image: TemporaryFile, libraryId: Id[Library], source: ImageSource, requestId: Option[Id[LibraryImageRequest]] = None): Future[ImageProcessDone]
+  def positionLibraryImage(libraryId: Id[Library], imageHash: ImageHash, imagePosition: LibraryImagePosition): Either[(Int, String), Seq[LibraryImage]]
```
andrewconner: Same. Given a library id, we want to just position each of the active images. Why is hash needed?

kifi-backend/modules/shoebox/conf/shoebox.routes
2is10: Yes, the client does know the dimensions, but I believe that’s irrelevant here.
The `is` parameter is for “ideal size” (what size image it would like returned in the response, as a path).

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryImageCommander.scala
andrewconner: `INACTIVE` for us usually means deleted / finished / dead. Perhaps make another state if you need it?

kifi-backend/modules/shoebox/conf/shoebox.routes
andrewconner: Ah, I see. Makes sense.

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryImageRepo.scala
```diff
@@ -9,7 +9,8 @@ import com.keepit.common.time.Clock
 @ImplementedBy(classOf[LibraryImageRepoImpl])
 trait LibraryImageRepo extends Repo[LibraryImage] {
   def getForLibraryId(libraryId: Id[Library], excludeState: Option[State[LibraryImage]] = Some(LibraryImageStates.INACTIVE))(implicit session: RSession): Seq[LibraryImage]
-  def getBySourceHash(hash: ImageHash)(implicit session: RSession): Seq[LibraryImage]
+  def getForLibraryIdOrWithHash(libraryId: Id[Library], hash: ImageHash)(implicit session: RSession): Seq[LibraryImage]
```
2is10: Given the in-person discussion we just had, I think we won’t need this method. If we were to need it, I’d recommend calling it `getByLibraryIdAndEitherActiveOrWithHash` to make it really clear that the library ID is always honored.

kifi-backend/modules/shoebox/test/com/keepit/commanders/KeepImageCommanderTest.scala
```diff
@@ -62,12 +62,12 @@ class KeepImageCommanderTest extends Specification with ShoeboxTestInjector with
         {
           val savedF = commander.setKeepImageFromFile(fakeFile1, keep1.id.get, ImageSource.UserUpload)
           val saved = Await.result(savedF, Duration("10 seconds"))
-          saved === ImageProcessState.StoreSuccess
+          saved.isInstanceOf[ImageProcessState.StoreSuccess] === true // if this test fails, make sure imagemagick is installed. Use `brew install imagemagick`
```
2is10: Prefer [assertions](http://etorreborre.github.io/specs2/guide/org.specs2.guide.Matchers.html) that provide the most helpful messages when they fail.

``` scala
saved must beInstanceOf[ImageProcessState.StoreSuccess]
```

(It’d be nice to see in the message what `saved` is, rather than just seeing “expected `true`, got `false`”.)

andrewconner: And if `StoreSuccess` drops back to not contain hashes, this isn't necessary (but useful for other situations).

kifi-backend/modules/shoebox/test/com/keepit/controllers/website/LibraryImageControllerTest.scala
```diff
@@ -0,0 +1,175 @@
...
+             | { "imagePath" : "library/26dbdc56d54dbc94830f7cfc85031481_66x38_o.png" }
+           """.stripMargin)
+
+        // Upload again to different library - same image path
+        val uploadResp3 = uploadFile(user1, libPubId2, fakeFile)
+        status(uploadResp3) === OK
+        Json.parse(contentAsString(uploadResp3)) === Json.parse(
+          s"""
+             | { "imagePath" : "library/26dbdc56d54dbc94830f7cfc85031481_66x38_o.png" }
+           """.stripMargin)
```
2is10: Let’s please stop using `.stripMargin` before a `Json.parse` call. `Json.parse` correctly ignores extra whitespace.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryImageController.scala
```diff
@@ -0,0 +1,86 @@
...
+    }
+  }
+
+  def positionLibraryImage(pubId: PublicId[Library]) = (UserAction andThen LibraryWriteAction(pubId))(parse.tolerantJson) { request =>
+    val libraryId = Library.decodePublicId(pubId).get
+    val imagePath = (request.body \ "imagePath").as[String]
+    val posX = (request.body \ "x").asOpt[Int]
+    val posY = (request.body \ "y").asOpt[Int]
+
+    val pathRegex = """library/(.*)_(\d+)x(\d+).+""".r
```
2is10: Using `.*` in the first group will result in the engine always scanning to the end of the input and then backtracking to find a match. You can prevent that by making the quantifier reluctant `.*?` or by clarifying that it can’t include an underscore `[^_]*`.

Also, probably not worth grouping the width and height, since they’re never used. The regexp could just end with `_.*` or `_.+`.

And it would be more efficient to only compile the regexp once, in a `LibraryImageController` companion object.

2is10:
great tests!

ahsu1230:
Sorry about the complications earlier. I think it should be correct now.
- Upload endpoint automatically sets the uploaded image (with position) as the library's image
  - it also still responds with an `imagePath` 
- Reposition endpoint still takes an `imagePath` for sanity checking with the current library image
- BaseImage no longer needs to be changed

ahsu1230:
merging... let me know if there are any comments =)

2is10:
Yay! Excited to try the endpoints. Still need the images to be sent in responses when libraries are fetched.

ahsu1230:
Yup! So I just had a thought. `LibraryImageCommander.getBestImageForLibrary()` requires an idealSize.
I'm guessing when we view a library, we will need to pass in an idealSize as part of the request, so we can retrieve the right image and respond with a library image path

2is10:
Yes, each endpoint that returns library JSON should understand `?is=WxH` and, if absent, fall back to a reasonable default. It may make sense to use a different default (fallback) size per endpoint (e.g. smaller for mobile).

PR #12708, opened 2014-12-11
Absolute URLs in attribution bar

andrewconner:
@eishay @yipingliao


yipingliao:
:+1:

eishay:
great!
how about tag searches?
`<a ng-href="/find?q=tag:OpenWrt" ng-click="onTagClick()" class="ng-binding" href="/find?q=tag:OpenWrt">OpenWrt</a>`

eishay:
that actually has not `href`, only `ng-href`

eishay:
can we add the `href` there?
in any rate, lets push this change asap to prod and reindex

andrewconner:
Added tags links, switched to using `href` instead of `ng-href`. This _shouldn't_ make a difference, but you never know. A slight risk is if Google tries to crawl/index the raw unparsed string before it gets replaced. Unlikely, but something to watch for. If we see tons of broken links after this, that's why.

PR #12710, opened 2014-12-12
Fix build

leogrim:
@andrewconner


leogrim:
Thanks, please see other PR, some Play! engineer probably deserves it more than I do :)

eishay:
yea, saw it :-)

2is10:
Methvin!!​

PR #12714, opened 2014-12-12
clean footers

ahsu1230:
looks like this!
https://cloud.githubusercontent.com/assets/377205/5174059/bf83e6da-73e2-11e4-9a79-1dda43ea04aa.jpg
@andrewconner @yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/library.tpl.html
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/library.tpl.html
@@ -80,4 +80,18 @@
     </div>
   </div>
 
+  <div class="kf-lib-info-footer" ng-if="!$root.userLoggedIn">
+    <div class="first-line">
+      <a class="kf-footer-item-link" href="https://www.kifi.com/how_it_works">How it works</a>&middot;
+      <a class="kf-footer-item-link" target="_top" href="https://www.kifi.com/about">About</a> &middot;
+      <a class="kf-footer-item-link" href="http://blog.kifi.com">Blog</a>
+    </div>
+    <span ng-if="!isMobile">
+      &middot;
+    </span>
+    <div class="second-line">
+      <a class="kf-footer-item-link" target="_top" href="https://www.kifi.com/privacy">Privacy</a> &middot;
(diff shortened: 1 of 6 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/library.tpl.html
```diff
@@ -80,4 +80,18 @@
     </div>
   </div>
 
+  <div class="kf-lib-info-footer" ng-if="!$root.userLoggedIn">
+    <div class="first-line">
+      <a class="kf-footer-item-link" href="https://www.kifi.com/how_it_works">How it works</a>&middot;
+      <a class="kf-footer-item-link" target="_top" href="https://www.kifi.com/about">About</a> &middot;
+      <a class="kf-footer-item-link" href="http://blog.kifi.com">Blog</a>
+    </div>
+    <span ng-if="!isMobile">
```
ahsu1230, line 89: add dot if desktop. Otherwise, no dot for mobile

kifi-backend/modules/shoebox/angular/src/libraries/library.tpl.html
```diff
@@ -80,4 +80,18 @@
     </div>
   </div>
 
+  <div class="kf-lib-info-footer" ng-if="!$root.userLoggedIn">
+    <div class="first-line">
+      <a class="kf-footer-item-link" href="https://www.kifi.com/how_it_works">How it works</a>&middot;
+      <a class="kf-footer-item-link" target="_top" href="https://www.kifi.com/about">About</a> &middot;
+      <a class="kf-footer-item-link" href="http://blog.kifi.com">Blog</a>
+    </div>
```
2is10, line 88: It seems as though we should either use `target="_top"` on all of these links or none. Personally, I don’t see a reason for them, but they don’t do any harm.

Also, for equal spacing, it seems we should either have whitespace before each `&middot;` or not.

andrewconner, line 88: The target is so Angular doesn't route it. Weird semantic, but that's how you do it.

PR #12715, opened 2014-12-12
related libs controller

yingjieMiao:
@joshm1 @stkem 

@joshm1 feel free to make suggestions on field names, returning format, etc.


kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
```diff
@@ -510,6 +511,18 @@ class LibraryController @Inject() (
     }
 
   }
+
+  def relatedLibraries(pubId: PublicId[Library]) = MaybeUserAction.async { request =>
```
joshm1, line 515: do you need to add this to the routes file?

joshm1:
lgtm other than missing route

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
yingjieMiao, line 515: yes. will.

PR #12718, opened 2014-12-12
Using real links for bots.

andrewconner:
@eishay @yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/common/directives/keepWho/keepWhoLib.tpl.html
+++ b/kifi-backend/modules/shoebox/angular/src/common/directives/keepWho/keepWhoLib.tpl.html
@@ -1,9 +1,14 @@
 <span ng-if="!library.hidden">
-  <a class="kf-keep-who-pic-box" href="{{library.absPath}}" ng-class="{ 'secret': library.visibility === 'secret' }" ng-click="onLibraryAttributionClicked()">
-    <span class="kf-keep-who-pic" ng-attr-style="background-image: url({{library.keeperPic}})"></span>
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/angular/src/common/directives/keepWho/keepWhoLib.tpl.html
```diff
@@ -1,9 +1,14 @@
...
+      <span class="kf-keep-who-pic" ng-attr-style="background-image: url({{library.keeperPic}})"></span>
+      <span class="kf-keep-who-lib">{{library.name}}</span>
+      <span kf-tooltip position="top" padding="10">
+        <span kf-library-mini-card library="library"></span>
+      </span>
+    </a>
+  </span>
+  <span ng-if="isBot" class="kf-keep-who-pic-box">
+    <a href="{{library.absPath}}" target="_self">{{library.name}}</a>
+  </span>
```
2is10, line 13: Is the `target="_self"` the only thing bots need? I think it’d be best to minimize both the perceptible UI differences and the amount of code in each branch.

It looks like [someone has recommended using a directive to conditionally add the target attribute to links](http://stackoverflow.com/questions/23584201/conditionally-add-target-blank-to-links-with-angular-js).

andrewconner, line 13: In this case, I was trying a couple different things at once. Namely, `target="_self"` (causes Angular not to internally route the link) and making he link markup more simple (just an `a` with text body).

If Google seems happy now, I'll isolate which of these may have contributed, and make it clean (and we'll use it everywhere).

PR #12724, opened 2014-12-12
adding library membership visibility (db patch as well). counting lib nu...

eishay:
...mbers for anonymous users


eishay:
@andrewconner @yipingliao

eishay:
sql patch applyed

PR #12728, opened 2014-12-12
related library cards frontend

joshm1:
@andrewconner @2is10 @yipingliao @stkem 

these should be hidden unless you put grl=1 in the URL (query param)
Using a query parameter to test this so it's easy to test on mobile - any objections?


kifi-backend/modules/shoebox/angular/src/libraries/relatedLibraryCard.tpl.html
```diff
@@ -0,0 +1,31 @@
+<div class="kf-related-lib-card" ng-attr-style="background-color: {{library.cardColor}}">
```
yipingliao: This card will be re-used in other places (e.g., profile! :) ).  Can we get it out into its own directive (and template file) for ease of reuse? Maybe "smallLibraryCard"? relatedLibraryCard can be a wrapper around smallLibraryCard. Also, smallLibraryCard styling should have 0 margin and plain positioning. Margins and positions can be done with the wrapper div (relatedLibraryCard) so that we can have different margins and positions for the different use cases without affecting smallLibraryCard. Thanks!

kifi-backend/modules/shoebox/angular/src/libraries/library.js
```diff
@@ -168,6 +178,7 @@ angular.module('kifi')
       if (eventType === 'click') {
         if (!$rootScope.userLoggedIn) {
           attributes.type = attributes.type || 'libraryLanding';
+          console.log(attributes);
```
yipingliao: Will this pass jshint?

kifi-backend/modules/shoebox/angular/src/libraries/relatedLibraries.styl
```diff
@@ -0,0 +1,324 @@
+// TEMPORARY
+keepsColumnWidth = 800px
+relatedLibrariesColumnWidth = 226px
+desktopBorderRadius = 19px
+mobileBorderRadius = 54px
+
+.kf-logged-out {
+  .kf-library-results-wrapper {
+    width: keepsColumnWidth;
```
yipingliao, line 8: CSS syntax?

kifi-backend/modules/shoebox/angular/src/libraries/libraryService.js
```diff
@@ -360,6 +375,26 @@ angular.module('kifi')
         return defaultAttributes;
       },
 
+      getRelatedLibraries: function (libraryId) {
+        var deferred = $q.defer();
+
+        // temporary - put ?grl=1 in URL to see related libraries
+        // using a query param so it's easy to test on mobile devices
+        if ($location.url().indexOf('grl=1') < 0) {
```
yipingliao, line 383: `$routeParams.grl`also does the trick, if you like how that looks.
https://docs.angularjs.org/api/ngRoute/service/$routeParams

kifi-backend/modules/shoebox/angular/src/libraries/relatedLibraryCard.tpl.html
joshm1: Thanks for the feedback. I extracted out all that's not directly related to a "related" library into a smallLibraryCard. I didn't change the position though.  It has to be static for the "stack of cards" effect with the absolute positioned child elements.  I suppose adding an additional container element with plain positioning would suffice.

kifi-backend/modules/shoebox/angular/src/libraries/relatedLibraries.styl
joshm1, line 8: Sorry, I was probably 200+ lines in before I realized I wasn't using the no-braces no-semi-colon syntax of stylus.  Habit from years using SASS (SCSS).

kifi-backend/modules/shoebox/angular/src/libraries/smallLibraryCard.tpl.html
```diff
@@ -0,0 +1,31 @@
+<div class="kf-small-library-card" ng-attr-style="background-color: {{library.cardColor}}">
```
yipingliao, line 1: Thank you very much!

kifi-backend/modules/shoebox/angular/src/libraries/libraryService.js
```diff
@@ -102,6 +102,21 @@ angular.module('kifi')
...
+      this.numKeeps = library.numKeeps;
+      this.ownerPicUrl = routeService.formatPicUrl(library.owner.id, library.owner.pictureName, 200);
+      this.name = library.name;
+      this.imageUrl = null; // TODO(josh) add when we get this image from the backend
+      this.libraryUrl = library.url;
+      this.followers = library.followers.map(function (user) {
+        return _.merge(user, { picUrl: routeService.formatPicUrl(user.id, user.pictureName, 200) });
+      });
+      // default color until we're passed colors from the backend
+      this.cardColor = '#73c785';
```
andrewconner, line 117: Nice, future proofing :)

PR #12732, opened 2014-12-15
twitter connect / disconnect in user settings

ahsu1230:
@andrewconner @yipingliao @raymondng


kifi-backend/modules/shoebox/angular/src/profile/profile.js
```diff
@@ -212,6 +212,41 @@ angular.module('kifi')
...
 ])
 
+.directive('kfTwitterConnectButton', [
+  'socialService',
+  function (socialService) {
+    return {
+      restrict: 'A',
+      link: function (scope) {
+        // todo (aaron): endpoint doesn't return profileUrl yet. Leave this out for now
+        //scope.isTwitterConnected = socialService.twitter && !!socialService.twitter.profileUrl;
```
ahsu1230, line 222: currently the twitter endpoint doesn't provide `profileUrl` so that part is going to be ignored for now.
When you click on the Twitter social button, it won't go to your twitter profile page (which is the case for facebook & linkedin)

kifi-backend/modules/shoebox/angular/src/profile/profile.styl
```diff
@@ -161,7 +161,6 @@ input.profile-input-input
 .profile-input-placeholder
 input.profile-input-input
   border-bottom: 1px solid #d2dae3
-  padding-right: 58px
```
ahsu1230: fixes a problem in user settings which covers the first name

kifi-backend/modules/shoebox/angular/src/profile/profile.tpl.html
```diff
@@ -46,7 +46,7 @@
           </li>
 
           <li kf-linkedin-connect-button data-network="linkedin" ng-class="{ 'connected': isLinkedInConnected }">
-            <a class="profile-nw nw-linkedin" href="https://www.kifi.com/link/linkedin" target="_self" ng-click="connectFacebook()" ng-show="!isLinkedInConnected">LinkedIn</a>
```
ahsu1230, line 49: called facebook for some reason instead of linkedin =0 (I think it only affects analytics, maybe...)

kifi-backend/modules/shoebox/angular/src/profile/profile.js
andrewconner, line 222: Could we get it added? @raymondng

raymondng, line 222: Sure thing.

andrewconner:
Nice, looks good (with the `profileUrl` Ray is adding)

ahsu1230:
Thanks for checking over! @raymondng let me know when it's in =)

raymondng:
:+1: Thanks!

PR #12738, opened 2014-12-15
libraries have hex color (mobile support only for now)

ahsu1230:
@andrewconner 

@thassss @jaredpetker simply add a `color` field to the Json body when creating or modifying library
`{ ..., 'color': '#aa00aa', ... }`


kifi-backend/modules/common/app/com/keepit/model/Library.scala
```diff
@@ -209,3 +212,14 @@ case class BasicLibraryStatistics(memberCount: Int, keepCount: Int)
 object BasicLibraryStatistics {
   implicit val format = Json.format[BasicLibraryStatistics]
 }
+
+class HexColor private (val hex: String) {
```
ahsu1230, line 216: requires `hex` to be 7 characters long, all lower case, and in format `#aaa000`

kifi-backend/modules/common/app/com/keepit/model/Library.scala
```diff
@@ -209,3 +212,14 @@ case class BasicLibraryStatistics(memberCount: Int, keepCount: Int)
...
+  require(hex.matches(HexColor.regexStr), "hex color must be in format '#aaaaaa'")
+}
+object HexColor {
+  implicit def format[T]: Format[HexColor] =
+    Format(__.read[String].map(HexColor(_)), new Writes[HexColor] { def writes(o: HexColor) = JsString(o.hex) })
+
+  val regexStr = "^#[0-9a-f]{6}"
```
andrewconner, line 223: More efficient to build the regex once and then use that:

``` scala
val regex = "^#[0-9a-f]{6}".r
```

ahsu1230, line 223: so I think the `matches()` function for `String` takes in a `String` parameters only.
Finding something that takes in a Regex type as a parameter and returns maybe a boolean...

I've found this: `regex.pattern.matcher(hex).matches` but there might be something a bit more concise lol

2is10, line 223:
``` scala
val hexRe = "^#[0-9a-f]{6}$".r  // notice the $ at the end
...
require(hexRe.findFirstIn(hex).isDefined, ...)
```

PR #12739, opened 2014-12-15
fixing an issue persisting an uploaded library image with a position

2is10:
also defaulting absent image positions to 50 at read time rather than before write
@ahsu1230

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryImageCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryImageCommander.scala
@@ -22,7 +22,7 @@ trait LibraryImageCommander {
   def getUrl(libraryImage: LibraryImage): String
   def getBestImageForLibrary(libraryId: Id[Library], idealSize: ImageSize): Option[LibraryImage]
   def uploadLibraryImageFromFile(image: TemporaryFile, libraryId: Id[Library], position: LibraryImagePosition, source: ImageSource, requestId: Option[Id[LibraryImageRequest]] = None): Future[ImageProcessDone]
-  def positionLibraryImage(libraryId: Id[Library], imagePosition: LibraryImagePosition): Seq[LibraryImage]
+  def positionLibraryImage(libraryId: Id[Library], position: LibraryImagePosition): Seq[LibraryImage]
   def removeImageForLibrary(libraryId: Id[Library]): Boolean // Returns true if images were removed, false otherwise
 }
 
@@ -61,11 +61,9 @@ class LibraryImageCommanderImpl @Inject() (
     db.readWrite { implicit s =>
       val toPosition = libraryImageRepo.getForLibraryId(libraryId)
       val positionedImages = toPosition.map { libImage =>
-        val setX = position.x.getOrElse(libImage.positionX.getOrElse(50))
-        val setY = position.y.getOrElse(libImage.positionY.getOrElse(50))
         libraryImageRepo.save(libImage.copy(
-          positionX = Some(setX),
-          positionY = Some(setY),
+          positionX = position.x,
+          positionY = position.y,
           state = LibraryImageStates.ACTIVE))
       }
       log.info("[lic] Positioning: " + toPosition.map(_.imagePath))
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryImageCommander.scala
```diff
@@ -22,7 +22,7 @@ trait LibraryImageCommander {
   def getUrl(libraryImage: LibraryImage): String
   def getBestImageForLibrary(libraryId: Id[Library], idealSize: ImageSize): Option[LibraryImage]
   def uploadLibraryImageFromFile(image: TemporaryFile, libraryId: Id[Library], position: LibraryImagePosition, source: ImageSource, requestId: Option[Id[LibraryImageRequest]] = None): Future[ImageProcessDone]
-  def positionLibraryImage(libraryId: Id[Library], imagePosition: LibraryImagePosition): Seq[LibraryImage]
+  def positionLibraryImage(libraryId: Id[Library], position: LibraryImagePosition): Seq[LibraryImage]
```
2is10, line 25: just making the parameter name consistent with the method definition below

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryImageCommander.scala
```diff
@@ -166,6 +156,10 @@ class LibraryImageCommanderImpl @Inject() (
     }
   }
 
+  private def sameHashAndWidthAndHeightAsAnyOf(images: Set[LibraryImage])(image: LibraryImage): Boolean = {
+    images.exists(i => i.sourceFileHash == image.sourceFileHash && i.width == image.width && i.height == image.height)
+  }
```
2is10, line 161: This predicate saves 8 lines and (I think) makes the method above easier to understand.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryImageCommander.scala
```diff
@@ -143,29 +139,27 @@ class LibraryImageCommanderImpl @Inject() (
...
-      val (shouldBeActive, shouldBeInactive) = oldLibraryImages.partition { existingImg =>
-        newLibraryImages.find { newImg =>
-          existingImg.sourceFileHash == newImg.sourceFileHash && existingImg.width == newImg.width && existingImg.height == newImg.height
-        }.nonEmpty
-      }
-      val toActivate = shouldBeActive.filter(_.state != LibraryImageStates.ACTIVE).map(_.copy(state = LibraryImageStates.ACTIVE))
+      val (shouldBeActive, shouldBeInactive) = oldLibraryImages.partition(sameHashAndWidthAndHeightAsAnyOf(newLibraryImages))
+      val toUpdate = shouldBeActive
+        .filter(i => i.state != LibraryImageStates.ACTIVE || i.positionX != position.x || i.positionY != position.y)
+        .map(_.copy(state = LibraryImageStates.ACTIVE, positionX = position.x, positionY = position.y))
```
2is10, line 149: This is the fix. We need to update the row if the position is different.

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryImage.scala
```diff
@@ -21,14 +21,14 @@ case class LibraryImage(
...
+  def position = LibraryImagePosition(positionX, positionY)
   def withId(id: Id[LibraryImage]) = copy(id = Some(id))
   def withUpdateTime(now: DateTime) = copy(updatedAt = now)
 }
 
-@json case class LibraryImageInfo(path: String, x: Option[Int], y: Option[Int])
+@json case class LibraryImageInfo(path: String, x: Int, y: Int)
 object LibraryImageInfo {
-  def createInfo(img: LibraryImage) = LibraryImageInfo(img.imagePath, img.positionX, img.positionY)
+  def createInfo(img: LibraryImage) = LibraryImageInfo(img.imagePath, img.positionX.getOrElse(50), img.positionY.getOrElse(50))
```
ahsu1230, line 31: I see, so don't save the default positions to the database. The default positions are only given when serializing responses.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryImageController.scala
```diff
@@ -45,8 +45,8 @@ class LibraryImageController @Inject() (
       case success: ImageProcessSuccess =>
         val idealSize = imageSize.flatMap { s => Try(ImageSize(s)).toOption }.getOrElse(LibraryImageController.defaultImageSize)
         libraryImageCommander.getBestImageForLibrary(libraryId, idealSize) match {
-          case Some(img) =>
-            Ok(Json.obj("imagePath" -> img.imagePath))
+          case Some(img: LibraryImage) =>
+            Ok(Json.toJson(LibraryImageInfo.createInfo(img)))
```
2is10, line 49: changing the endpoint’s response format to be consistent with the format we use in `FullLibraryInfo` and to preserve the default positioning behavior

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryImageCommander.scala
ahsu1230, line 149: :+1:

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryImageController.scala
```diff
@@ -83,6 +83,6 @@ class LibraryImageController @Inject() (
 }
 
 object LibraryImageController {
-  val defaultImageSize = ImageSize(600, 480)
-  val pathRegex = """^library/(.*?)_\d+x\d+.+""".r
+  val defaultImageSize = ImageSize(800, 600)
+  val pathHashRe = "library/([^_]+)_.*".r
```
2is10, line 87: being a little more lenient with this regular expression, since its purpose is just to extract the hash, not to validate the path

ahsu1230:
:+1: thanks for the cleanup

PR #12740, opened 2014-12-16
Few front end changes

ahsu1230:
- Long library breaks hover card
  https://app.asana.com/0/13310647066703/19068363415722
- hex color regex is created once

@andrewconner @2is10

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/model/Library.scala
+++ b/kifi-backend/modules/common/app/com/keepit/model/Library.scala
@@ -214,12 +214,12 @@ object BasicLibraryStatistics {
 }
 
 class HexColor private (val hex: String) {
-  require(hex.matches(HexColor.regexStr), "hex color must be in format '#aaaaaa'")
+  require(HexColor.regex.findFirstIn(hex).isDefined, "hex color must be in format '#aaaaaa'")
 }
 object HexColor {
   implicit def format[T]: Format[HexColor] =
     Format(__.read[String].map(HexColor(_)), new Writes[HexColor] { def writes(o: HexColor) = JsString(o.hex) })
 
-  val regexStr = "^#[0-9a-f]{6}"
+  val regex = "^#[0-9a-f]{6}$".r
   def apply(hex: String): HexColor = new HexColor(hex.toLowerCase)
 }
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/shoebox/angular/src/profile/profile.styl
```diff
@@ -161,7 +161,6 @@ input.profile-input-input
 .profile-input-placeholder
 input.profile-input-input
   border-bottom: 1px solid #d2dae3
-  padding-right: 58px
```
ahsu1230, line 164: fixes covering over user's first name in user settings page

kifi-backend/modules/common/app/com/keepit/model/Library.scala
```diff
@@ -214,12 +214,12 @@ object BasicLibraryStatistics {
 }
 
 class HexColor private (val hex: String) {
-  require(hex.matches(HexColor.regexStr), "hex color must be in format '#aaaaaa'")
+  require(HexColor.regex.findFirstIn(hex).isDefined, "hex color must be in format '#aaaaaa'")
```
ahsu1230, line 217: thanks for the suggestions on this one! @andrewconner @2is10

andrewconner:
Thanks

kifi-backend/modules/shoebox/angular/src/libraries/libraryMiniCard.tpl.html
```diff
@@ -17,7 +17,7 @@
   </div>
 
   <div class="kf-library-mini-card-body">
-    <div class="kf-library-mini-card-name" ng-click="view()">{{library.name}}</div>
+    <div kf-ellipsis full-text="library.name" max-num-lines=1 class="kf-library-mini-card-name" ng-click="view()"></div>
```
2is10, line 20: One-line ellipsis is simpler and faster with just CSS.

``` css
white-space: nowrap;
overflow: hidden;
text-overflow: ellipsis;
```

2is10:
thanks!

kifi-backend/modules/shoebox/angular/src/libraries/libraryMiniCard.tpl.html
ahsu1230, line 20: arggh you're right #facepalm
thanks lol

PR #12742, opened 2014-12-16
endpoint for marketing site libraries

joshm1:
@andrewconner @stkem

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
@@ -68,6 +68,7 @@ class LibraryCommander @Inject() (
     uriSummaryCommander: URISummaryCommander,
     socialUserInfoRepo: SocialUserInfoRepo,
     experimentCommander: LocalUserExperimentCommander,
+    systemValueRepo: SystemValueRepo,
     implicit val publicIdConfig: PublicIdConfiguration,
     clock: Clock) extends Logging {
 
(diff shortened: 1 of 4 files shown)


andrewconner:
Awesome! What was the problem with the test?

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -273,12 +280,13 @@ class LibraryCommander @Inject() (
...
     val relevantInviteStates = Set(LibraryInviteStates.ACTIVE)
 
+    val memberCount = db.readOnlyMaster { implicit s => libraryMembershipRepo.countWithLibraryIdByAccess(libraryId) }
+
     if (limit > 0) db.readOnlyMaster { implicit session =>
       // Get Collaborators
       val collaborators = libraryMembershipRepo.pageWithLibraryIdAndAccess(libraryId, offset, limit, collaboratorsAccess)
       val collaboratorsShown = collaborators.length
 
-      val memberCount = libraryMembershipRepo.countWithLibraryIdByAccess(libraryId)
```
joshm1, line 281: In my new test (which created 2 libraries), this worked on the first library but failed on the second because the db session closed.  I never figured out why it failed on the second.  Replaced this call with the bulk call above (which is better anyway). But we still need to figure out why this failed.  @andrewconner @stkem

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -273,12 +280,13 @@ class LibraryCommander @Inject() (
     val followersAccess: Set[LibraryAccess] = Set(LibraryAccess.READ_ONLY)
     val relevantInviteStates = Set(LibraryInviteStates.ACTIVE)
 
+    val memberCount = db.readOnlyMaster { implicit s => libraryMembershipRepo.countWithLibraryIdByAccess(libraryId) }
```
joshm1, line 283: This method had a bug when you called it with `limit == 0` because it would return Map.empty, and the caller would then call `map(key)` which throws a a key not found exception.  Updated to return the memberCounts even if limit = 0 (which isn't used for looking up these values anyway).

PR #12743, opened 2014-12-16
more relaxation in retrieving similar libraries

yingjieMiao:
@stkem @ymatsuda @joshm1 

hope we have better recall


stkem:
lgtm

PR #12744, opened 2014-12-16
DRYing search.explain

leogrim:
This should enable us to share the explainer and scoring code paths, in order to DRY things up and avoid future divergence that may lead us to look at an explainer that does not represent the real thing. There might be room for some more DRYing, I'll see it as I move forward.

@ymatsuda hope I didn't miss anything obvious :)

@stkem next: split and dry Explanation to reuse the HTML generation code before getting to the actual LibrarySearch explainer - which may end up being very little work after this, reframe the problem and you're half-way to the solution :)


ymatsuda:
lgtm!

PR #12746, opened 2014-12-16
add color to related and marketing-suggested libraries

joshm1:
@stkem

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/libraryService.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/libraryService.js
@@ -108,13 +108,13 @@ angular.module('kifi')
       this.numKeeps = library.numKeeps;
       this.ownerPicUrl = routeService.formatPicUrl(library.owner.id, library.owner.pictureName, 200);
       this.name = library.name;
-      this.imageUrl = null; // TODO(josh) add when we get this image from the backend
+      this.image = library.image;
       this.libraryUrl = library.url;
       this.followers = library.followers.map(function (user) {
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/libraryService.js
```diff
@@ -387,6 +387,7 @@ angular.module('kifi')
 
         $http.get(routeService.getRelatedLibraries(libraryId)).then(function (resp) {
           var decoratedLibs = resp.data.libs.map(function (lib) {
+            lib.image = lib.image || {};
```
2is10, line 390: I’d vote for leaving `.image` null or missing when a library doesn’t have one. That way checking whether a library has an image is easier (just `.image`) and we don’t introduce objects that appear to be images but really aren’t.

kifi-backend/modules/shoebox/angular/src/libraries/smallLibraryCard.tpl.html
```diff
@@ -24,7 +24,7 @@
         </span>
       </div>
     </div>
-    <div ng-if="library.imageUrl"  class="kf-small-library-image" ng-attr-style="background-image: url({{library.imageUrl}})"></div>
+    <div ng-if="library.image.url"  class="kf-small-library-image" ng-attr-style="background-image: url({{library.image.url}})"></div>
```
2is10, line 27: I’m not seeing where `library.image.url` gets set. The server endpoint serializes `.image` as a `LibraryImageInfo`, which has a `.path` that needs `env.picBase + '/'` prepended.

PR #12749, opened 2014-12-16
TwitterSocialGraph: WIP

raymondng:
WIP, but the basic graph functionality is there, and graph (based on mutual-follows) is persisted in the same manner as FB and LNKD.

Some things are available/functional:
- web-based authentication/login/logout
- twitter user info handling, including profile image
- twitter mutual-follow-as-friend model
- persisted in `SocialUserInfo` in a similar manner as FB friends and LNKD connections
- if two twitter users (mutual follow) sign-up, they should be auto-friended
- twitter typeahead (somewhat arbitrarily slotted TWTR between FB and LNKD; product spec TBD)
  *\* while the results will show up, you cannot do anything with it (see limitations)

A simple `TwitterSocialGraphTest` has been added to provide basic coverage. More work required.

A bunch of known limitations, the main ones: 
- Invitation (spec pending) -- it's not implemented. So when you try to send an invitation to a Twitter friend, it will error out
- Typeahead (spec pending)
- Large data set handling (if you have more than 75K followers or friends, it does not fetch all of them)
- Rate limit handling

See [asana task](https://app.asana.com/0/7517704133688/21793251446662) for a  laundry list of TODOs.

@andrewconner 
cc: @ahsu1230 
cc: @weeewake @JRuff42


kifi-backend/modules/shoebox/test/com/keepit/social/TwitterSocialGraphTest.scala
```diff
@@ -0,0 +1,295 @@
...
+import com.keepit.test.ShoeboxTestInjector
+import org.specs2.mutable.Specification
+import play.api.libs.json.{ JsArray, Json, JsNull, JsValue }
+import securesocial.core.{ IdentityId, AuthenticationMethod, SocialUser, OAuth1Info }
+
+import scala.concurrent.Future
+
+class TwitterSocialGraphTest extends Specification with ShoeboxTestInjector {
+
+  val tweetfortytwoRaw =
```
andrewconner, line 16: Since you're parsing this as JSON, the `|`'s are not needed. Leaving them out has the great side effect of the string being valid JSON.

raymondng, line 16: Cool. Will take care of it.

raymondng:
Also added experiment per discussion with @andrewconner. Will merge and build on it.

PR #12758, opened 2014-12-16
add description field to related libraries and a caption to marketing site libraries

joshm1:
marketing site library gets it's caption from the db system value, which allows us to set the value manually specific to the marketing site

@stkem


stkem:
lgtm

PR #12762, opened 2014-12-17
fix keep tracking for mobile

yingjieMiao:
@stkem 

A quick fix. 

@leogrim suggested that we do tracking in keepInterner, this is where all keep endpoints converge. I think it's a great idea. Considering huge batch imports, I am not quite sure about the performance impact if we do all tracking there.


leogrim:
But even better / decoupled / more reliable would be to listen on sequence
numbers.

stkem:
Too much lag for this use case. We already listen on sequence number of the
bookmark table.

PR #12763, opened 2014-12-17
User Settings for profile (show libraries I follow optional)

ahsu1230:
In the `UserValueRepo`, `User_Profile_Settings` maps to a Json Array of settings
i.e.

```
{ 'show_followed_library' : false }
```

there are two endpoints that can be used to get/set settings
GET     /site/user/settings   (returns all settings with default values)
POST    /site/user/settings  (takes in a json body of settings you want to set)

@andrewconner @eishay


kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -1207,7 +1208,7 @@ class LibraryCommander @Inject() (
 
   private def countLibrariesForOtherUser(userId: Id[User], friendId: Id[User]): Int = {
     db.readOnlyReplica { implicit s =>
-      val showFollowLibraries = true
+      val showFollowLibraries = getUserValueSetting(userId, UserValueName.SHOW_FOLLOWED_LIBRARIES)
```
ahsu1230, line 1213: @eishay you can use this as an example of how to retrieve a certain setting preference in the `UserValueRepo`

PR #12756, opened 2014-12-16
library cover images UI MVP

2is10:
I’ve set expectations that this feature will go live today, so I’d appreciate a prompt once-over looking for any site-breaking problems.

I’d also highly value any other comments you might have, as I’m an Angular n00b. I intend to move the cover image editing stuff into its own directive, but that refactoring may take ½ day or more. Thanks!

@andrewconner @yipingliao

PR #12764, opened 2014-12-17
Using `FullLibraryInfo` in recommendations directly

stkem:
@andrewconner


andrewconner:
Thanks! Needed for #12756

PR #12767, opened 2014-12-17
paginating over user's owned libraries from three views (owner, anonymous, connection)

eishay:
the feature is not complete but i rather split the commits to avoid merge conflicts.


kifi-backend/modules/shoebox/app/com/keepit/commanders/UserCommander.scala
```diff
@@ -103,17 +103,23 @@ class UserCommander @Inject() (
...
-    //    val friends = userConnectionRepo.getConnectionCount(user.id.get) //cached
-    //    val numFollowers = libraryMembershipRepo.countFollowersWithOwnerId(user.id.get) //cached
-    val numKeeps = keepRepo.getCountByUser(user.id.get)
-    val isConnected: Option[Boolean] = viewer flatMap {
-      case me if me.id.get == user.id.get => None //if i see my own profile, is connected is not relevant
-      case other => Some(userConnectionRepo.getConnectionOpt(user.id.get, other.id.get).isDefined)
-    }
-    UserProfile(user = user, viewer = viewer.map(_.id.get), numKeeps = numKeeps, isConnected = isConnected)
+  def userFromUsername(username: Username): User = db.readOnlyReplica { implicit session =>
+    userRepo.getByUsername(username).getOrElse(throw UserNotFoundException(username)) //cached
```
andrewconner, line 107: Why throw an exception? Can't we just return a `None`, and then return a 404 to the client? This will happen since it's user-provided data — typos, someone poking around, etc airbrake.

eishay, line 107: Good point, will do

PR #12770, opened 2014-12-17
twtr: fix reference

raymondng:
Fix param reference. 
@andrewconner


raymondng:
This is a simple bug fix. Large data set not currently supported, so Twitter will still complain. I may tone down airbrake as well.

PR #12774, opened 2014-12-17
preventing cover image transitions on page load or in-page nav

2is10:
@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.tpl.html
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.tpl.html
@@ -27,22 +27,22 @@
     </div>
 
     <div ng-if="recommendation && coverImageUrl" class="kf-keep-lib-cover-image" style="background-image:url({{coverImageUrl}});background-position:{{coverImagePos}}"></div>
-    <div ng-if="!recommendation && (coverImageUrl || coverImagePreview)" class="kf-keep-lib-cover-image" ng-class="{'animate-if':coverImagePreview}" ng-style="{'background-image':'{{coverImageUrl ? 'url(' + coverImageUrl + ')' : ''}}','background-position':'{{coverImagePos}}'}">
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.tpl.html
```diff
@@ -27,22 +27,22 @@
     </div>
 
     <div ng-if="recommendation && coverImageUrl" class="kf-keep-lib-cover-image" style="background-image:url({{coverImageUrl}});background-position:{{coverImagePos}}"></div>
-    <div ng-if="!recommendation && (coverImageUrl || coverImagePreview)" class="kf-keep-lib-cover-image" ng-class="{'animate-if':coverImagePreview}" ng-style="{'background-image':'{{coverImageUrl ? 'url(' + coverImageUrl + ')' : ''}}','background-position':'{{coverImagePos}}'}">
-      <a ng-if="isMyLibrary() && !coverImagePreview" class="kf-keep-lib-cover-image-adjust animate-if" href="javascript:" ng-mousedown="onCoverImageAdjustEngage($event)" ng-click="onCoverImageAdjustEngage($event)"></a>
-      <div ng-if="coverImageMenuShowing" class="kf-keep-lib-cover-image-menu animate-if">
```
2is10, line 32: The animation example on the `ng-if` documentation page used an `animate-if` class, and I thought it was indicator to Angular to please bother with `ng-enter`, `ng-enter-active`, `ng-leave`, `ng-leave-active`, but I should have trusted Yiping, who told me that Angular does all of that work for every `ng-if` and `ng-show` toggle.

My eyes must have deceived me, or there was a rendering stall when I tested my first fix for this issue.

2is10:
I feel this solution is a little hacky. I’d really prefer delaying an in-page navigation until the primary data needed for it is available. Here’s [a stack overflow question](http://stackoverflow.com/questions/20094273/stop-angular-ui-router-navigation-until-promise-is-resolved) about how to achieve that.

Besides less janky rendering (multiple layout passes), a side benefit would be that if loading the data fails, the user stays on the current page rather than navigating to an empty shell of a page.

PR #12777, opened 2014-12-17
use library image for social share

eishay:
@andrewconner


andrewconner:
Thanks!

eishay:
does not work yet, i need to wipe out the cache. coming next

PR #12778, opened 2014-12-17
exclude followed libraries from library recommendations

joshm1:
@stkem 

to update existing recos I can just set `library_membership_seq_num_sequence` to 0?


stkem:
lgtm. And yes, resetting the seq num should do the trick.

PR #12781, opened 2014-12-17
camelCase for user profile settings endpoints

ahsu1230:
for better usage on front-end side =P

@andrewconner


ahsu1230:
merging so I can apply changes on my front-end push. Happy to address comments later!

PR #12790, opened 2014-12-18
Switch back to v1 styles for user profiles.

yipingliao:
@2is10 @andrewconner 

In user profiles v1, we only show number of libraries and number of keeps and there is no need to navigate to friends, followers, etc. I had previously added those links to test nested views with angular ui router, but am not changing them back so we can continue with v1 UI development.


kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
```diff
@@ -99,6 +99,24 @@ angular.module('kifi')
...
+      });
+
+      switch ($state.current.data.libraryType) {
+        case 'My':
+          $scope.libraryNavLinks[0].selected = true;
+          break;
+        case 'Following':
+          $scope.libraryNavLinks[1].selected = true;
+          break;
+        case 'Invited':
```
2is10: I don’t know where these strings originate, but between this code and `libraryNavLinksConfig` on line 26, we have three different cases for each library type (e.g. `my`, `My`, and `MY`). Could we standardize on a lowercase for all of them? In most of our code that I’ve seen, we’ve gone with all-lower (e.g. routes), lower-camel (e.g. data model) or lower-underscore (e.g. error codes). One problem with using strings that have just the first letter capitalized is that they look like UI labels. If these are in fact UI labels, can we move them out of the controller and into a template? Thanks.

kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
```diff
@@ -99,6 +99,24 @@ angular.module('kifi')
       });
     }
 
+    function updateSelectedLibraryNavLink() {
+      _.forEach($scope.libraryNavLinks, function (navLink) {
+        navLink.selected = false;
+      });
```
2is10, line 127: We could scrap the `switch` statement below if line 104 above could instead be:

``` js
  navLink.selected = navLink.type === $state.current.data.libraryType;
```

kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
yipingliao: Will do. Thanks!

kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
yipingliao, line 127: Thanks! Will do.

yipingliao:
Made updates per comments; move header template into its own file; add back in numbers for number of libraries and number of keeps.

kifi-backend/modules/shoebox/angular/src/userProfile/userProfileLibraries.tpl.html
```diff
@@ -2,7 +2,7 @@
   <nav class="kf-user-profile-libraries-nav">
     <a class="kf-user-profile-libraries-nav-link" ng-repeat="libraryNavLink in libraryNavLinks"
        ui-sref="{{libraryNavLink.routeState}}" ng-class="{'selected': libraryNavLink.selected}">
-      <span class="kf-user-profile-libraries-nav-link-text">{{libraryNavLink.name}}</span>
+      <span class="kf-user-profile-libraries-nav-link-text">{{libraryNavLink.name | uppercase}}</span>
```
2is10: Oh, I didn’t realize `.name` was a UI label. I guessed otherwise because I saw it being used in controller logic (comparisons). Have about giving each link a `type` (or `libraryType`) and a `label` (with desired capitalization, only used for display purposes).

yipingliao: Will do.

yipingliao:
@2is10 I cleaned up the code a little more per our discussion; maybe that'll make things clearer? Or maybe not. :)  I am going to merge this now so we have something to start with.

2is10:
:+1:

PR #12791, opened 2014-12-18
twtr: copy oauth1 info

raymondng:
- Added missing oauth1 info for the oauth1 signup code path used by mobile. 
- Added test coverage as well.

@andrewconner @thassss


raymondng:
Simple fix for mobile. Will merge. Feedback appreciated.

PR #12792, opened 2014-12-18
Manage profile settings (behind profiles_beta experiment)

ahsu1230:
uses the endpoints for toggling profile settings (and only behind profile_beta experiment)

@andrewconner @yipingliao


kifi-backend/modules/shoebox/angular/src/profile/profileManageSettings.tpl.html
```diff
@@ -0,0 +1,30 @@
...
+  <a class="profile-manage-settings-title-wrapper" href="javascript:" ng-click="toggle()">
+    <div class="profile-setting-arrow" ng-class="{'active':isOpen}"></div>
+    <div class="profile-manage-settings-title" ng-class="{'active':isOpen}">Manage your Profile</div>
+  </a>
+
+  <div class="profile-manage-settings-change" ng-show="isOpen">
+
+    <div class="profile-manage-settings-group">
+
+      <div class="profile-manage-settings-option" ng-click="settings.showFollowedLibraries=true">
```
ahsu1230: Sorry @yipingliao I couldn't figure it out how to customize with `<input>` radio buttons. I'll try again tomorrow :fish_cake:

yipingliao: We can look at it together if you like.

kifi-backend/modules/shoebox/angular/src/profile/profileManageSettings.tpl.html
```diff
@@ -0,0 +1,30 @@
...
+        <span class="profile-manage-settings-option-text">Show Libraries you are following on your Profile</span>
+      </div>
+
+      <div class="profile-manage-settings-option" ng-click="settings.showFollowedLibraries=false">
+        <div class="profile-manage-settings-option-button">
+          <div class="checked" ng-show="settings.showFollowedLibraries==false"></div>
+        </div>
+        <span class="profile-manage-settings-option-text">Hide Libraries you are following on your Profile</span>
+      </div>
+
```
yipingliao, line 24: nit: remove some extraneous newlines.

kifi-backend/modules/shoebox/angular/src/profile/profileManageSettings.js
```diff
@@ -0,0 +1,40 @@
...
+          profileService.getSettings().then(function (val) {
+            scope.settings = val;
+            return scope.settings;
+          });
+        }
+
+        scope.$watch(function() {
+          return scope.settings.showFollowedLibraries;
+        }, function(newVal, oldVal) {
+          if (_.isBoolean(newVal) && _.isBoolean(oldVal)) {
```
yipingliao, line 28: Why does `oldVal` have to be boolean in order for new settings to be saved?

kifi-backend/modules/shoebox/angular/src/profile/profileManageSettings.js
```diff
@@ -0,0 +1,40 @@
...
+.directive('kfProfileManageSettings', [
+  'profileService',
+  function (profileService) {
+    return {
+      restrict: 'A',
+      scope: {},
+      templateUrl: 'profile/profileManageSettings.tpl.html',
+      link: function (scope) {
+        scope.isOpen = false;
+        scope.settings = {};
```
yipingliao: Naming suggestion: it looks like `settings` is referring to whether the user's followed libraries should be shown, but there is no indication of that anywhere outside the template file. The word "settings" is generic but in this case it is referring to something very specific.

kifi-backend/modules/shoebox/angular/src/profile/profileManageSettings.js
ahsu1230, line 28: `scope.settings.showFollowedLibraries` is first `undefined` and then when the endpoint comes back with data, it becomes a boolean value. Without the `_.isBoolean(oldVal)` the watch triggers and calls the setSettings endpoint (which is an extraneous call). I only wanted the endpoint to be called when you have a boolean value and it gets set to a different boolean value. Thoughts of a better way?

kifi-backend/modules/shoebox/angular/src/profile/profileManageSettings.js
ahsu1230: I was in anticipation of more setting options coming into play soon. so like `settings = { showFollowedLibraries, showFollowingUsers, hideMyProfile, ... }`

kifi-backend/modules/shoebox/angular/src/profile/profile.styl
```diff
@@ -335,6 +372,7 @@ a.profile-email-pending-cancel
     color: darkBlueColor
     transition: all 0.3s
 
+a.profile-manage-settings-title-wrapper
```
2is10: Sorry, I created a merge conflict for you here without realizing it.
After you pull and resolve, you won’t need the `a` at the front of this selector.

kifi-backend/modules/shoebox/angular/src/profile/profile.styl
```diff
@@ -285,9 +285,45 @@ a.profile-email-pending-cancel
...
   margin-top: 25px
 
+  .profile-manage-settings-option:first-child
+    margin-top: 10px
+
+  .profile-manage-settings-option
+    margin-left: 20px
+    margin-bottom: 5px
+
+    .profile-manage-settings-option-button
```
2is10: Please remove all indentation here. When using unambiguous selectors (e.g. unique class names), we shouldn’t nest them, since that results in more expensive (and larger) compound selectors.

kifi-backend/modules/shoebox/angular/src/profile/profile.styl
```diff
@@ -285,9 +285,45 @@ a.profile-email-pending-cancel
 .profile-email-address-pending-email
   color: #777c87
 
-.profile-email-addresses
+.profile-manage-settings
   margin-top: 25px
 
+  .profile-manage-settings-option:first-child
+    margin-top: 10px
```
2is10, line 279: Let’s embrace the brevity and organization Stylus allows and put this beneath the rule below, indented as `&:first-child`.

kifi-backend/modules/shoebox/angular/src/profile/profile.styl
```diff
@@ -285,9 +285,45 @@ a.profile-email-pending-cancel
...
+        width: 6px
+        height: 6px
+        border-radius: 50%
+        background-color: darkBlueColor
+
+
+    .profile-manage-settings-option-text
+      cursor: pointer
+    .profile-manage-settings-option-text:hover
+      color: darkBlueColor
```
2is10: Let’s indent and abbreviate the selector to `&:hover`.

kifi-backend/modules/shoebox/angular/src/profile/profileManageSettings.js
2is10: How about `profileSettings`?

kifi-backend/modules/shoebox/angular/src/profile/profileManageSettings.js
2is10, line 28: How about initiating the watch only after the initial load (in the `profileService.getSettings().then` callback above)?

kifi-backend/modules/shoebox/angular/src/profile/profileManageSettings.js
yipingliao: Or `userProfileSettings`? I've been using `userProfile` to refer to the user profile pages we're building, to disambiguate from the "profile" page where users can currently update their emails, etc.

kifi-backend/modules/shoebox/angular/src/profile/profileManageSettings.js
yipingliao, line 28: Ah, I see. Maybe this problem will go away when we use `<input>`...

kifi-backend/modules/shoebox/angular/src/profile/profileManageSettings.js
```diff
@@ -0,0 +1,40 @@
...
+        function init() {
+          profileService.getSettings().then(function (val) {
+            scope.settings = val;
+            return scope.settings;
+          });
+        }
+
+        scope.$watch(function() {
+          return scope.settings.showFollowedLibraries;
+        }, function(newVal, oldVal) {
```
2is10: `$watch` has a convenient shorthand for simple scope expressions:

``` js
scope.$watch('settings.showFollowedLibraries', function (...) {
  ...
});
```

kifi-backend/modules/shoebox/angular/src/profile/profileManageSettings.js
```diff
@@ -0,0 +1,40 @@
...
+        scope.settings = {};
+
+        scope.toggle = function () {
+          scope.isOpen = !scope.isOpen;
+        };
+
+        function init() {
+          profileService.getSettings().then(function (val) {
+            scope.settings = val;
+            return scope.settings;
```
2is10: Does this `return` statement have a purpose? I don’t see it.

kifi-backend/modules/shoebox/angular/src/profile/profileManageSettings.js
```diff
@@ -0,0 +1,40 @@
...
+  function (profileService) {
+    return {
+      restrict: 'A',
+      scope: {},
+      templateUrl: 'profile/profileManageSettings.tpl.html',
+      link: function (scope) {
+        scope.isOpen = false;
+        scope.settings = {};
+
+        scope.toggle = function () {
```
2is10: `toggleOpen` might be clearer

kifi-backend/modules/shoebox/angular/src/profile/profileService.js
```diff
@@ -286,9 +286,23 @@ angular.module('kifi')
...
+      return $http.get(routeService.profileSettings).then(function (res) {
+        return res.data;
+      });
+    }
+
+    function setSettings(changedSettings) {
+      return $http.post(routeService.profileSettings, changedSettings).then(function (res) {
+        return res.data;
+      });
+    }
```
2is10, line 295: These helpers may make it difficult for callers to distinguish success from errors. If we just eliminate the `.then` calls here, then callers could use `.success(function (data) { ... })` and `.error(function (data) { ... })`.

2is10:
:+1: Nice work. I hope my comments don’t feel overwhelming. They’re all quite minor. Hope they’re helpful.

kifi-backend/modules/shoebox/angular/src/profile/profileManageSettings.js
2is10: Ok by me. I think we need to eventually rename our existing “profile” page to “settings” or “account”. The term “profile” connotes something visible to others, and so it’s a misnomer for the settings page.

kifi-backend/modules/shoebox/angular/src/profile/profileManageSettings.js
ahsu1230: nope there is none

ahsu1230:
Thanks @2is10 and @yipingliao for the comments! Resolving the merge conflict now

2is10:
LGTM (once `/marketing` is updated)

PR #12794, opened 2014-12-18
new unread and sent messages endpoints for mobile

stkem:
@thassss @mrbrown14


kifi-backend/modules/eliza/conf/elizaService.routes
```diff
@@ -20,6 +20,8 @@ GET     /eliza/emailPreview/:msgExtId     @com.keepit.eliza.controllers.ext.ExtM
 
 GET     /m/1/eliza/ws                     @com.keepit.eliza.controllers.shared.SharedWsMessagingController.websocket(version: Option[String], eip: Option[String])
 GET     /m/1/eliza/notifications          @com.keepit.eliza.controllers.mobile.MobileMessagingController.getNotifications(n: Int, before: Option[String] ?= None)
+GET     /m/1/eliza/notifications/unread          @com.keepit.eliza.controllers.mobile.MobileMessagingController.getUnreadNotifications(n: Int, before: Option[String] ?= None)
```
stkem, line 23: @thassss @mrbrown14 these are your new routes. semantics are exactly the same as the existing one, except that they return only unread/sent messages, respectively.

stkem:
Live now

PR #12795, opened 2014-12-18
lowering the specificity of our default styles for links

2is10:
cc @yipingliao 

For fun, I went through one stylesheet (profile.styl) and removed the `a` specificity boosts.

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/profile/profile.styl
+++ b/kifi-backend/modules/shoebox/angular/src/profile/profile.styl
@@ -18,7 +18,7 @@
   padding: 18px 32px 20px 32px
   position: relative
 
-a.profile-logout
+.profile-logout
   position: absolute
   top: 32px
   right: 50px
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/angular/src/profile/profile.styl
```diff
@@ -91,23 +91,6 @@ a.profile-logout
...
-  letter-spacing: .5px
-  transition: background 0.2s ease
-
-  &:hover
-  &:focus
-    background-color: #C5D2DE
-
-  &:active
-    background-color: #9FACB8
-
```
2is10, line 110: These styles were obsolete. Today `profile-image-change` is only used on a `<div>`.

kifi-backend/modules/shoebox/angular/src/profile/profile.styl
```diff
@@ -499,32 +479,7 @@ input.profile-email-address-add-input
   padding-right: 58px
   border-radius: 2px
 
-a.profile-email-address-add-save
```
2is10, line 502: No elements use a `profile-email-address-add-save` class.

yipingliao:
Nice!

PR #12797, opened 2014-12-18
also include library from same owner

yingjieMiao:
@stkem @joshm1


stkem:
lgtm, although @joshm1 is probably the more qualified reviewer here. Would love to get some statsd timing on the top level method, so we can see how fast it is overall.

kifi-backend/modules/shoebox/app/com/keepit/commanders/RelatedLibraryCommander.scala
```diff
@@ -30,52 +30,78 @@ class RelatedLibraryCommanderImpl @Inject() (
...
-      .map { libs => libs.filter { x => !userLibs.contains(x.id.get) }.take(5) }
-      .flatMap { libs => libCommander.createFullLibraryInfos(userIdOpt, true, 10, 0, ProcessedImageSize.Large.idealSize, libs, ProcessedImageSize.Large.idealSize) }
+    suggestedLibsFut
+      .map { libs => libs.filter { case RelatedLibrary(lib, kind) => !userLibs.contains(lib.id.get) }.take(RETURN_SIZE) }
+      .flatMap { relatedLibs =>
+        val libs = relatedLibs.map { _.library }
+        val kinds = relatedLibs.map { _.kind }
+        val fullInfosFut = libCommander.createFullLibraryInfos(userIdOpt, true, 10, 0, ProcessedImageSize.Large.idealSize, libs, ProcessedImageSize.Large.idealSize)
+        fullInfosFut.map { info =>
+          assert(info.size == kinds.size)
```
joshm1: can we Future.recover from this, airbrake.notify and return an empty seq so the caller doesn't have to deal with the future fail?  Or make sure caller(s) (the controller) gracefully handles this and returns an empty seq to the client.

kifi-backend/modules/shoebox/app/com/keepit/commanders/RelatedLibraryCommander.scala
```diff
@@ -30,52 +30,78 @@ class RelatedLibraryCommanderImpl @Inject() (
...
-      suggestedLibs <- suggestedLibsFut
-    } yield (info, suggestedLibs._2)
   }
 
-  def suggestedLibraries(libId: Id[Library]): Future[(Seq[Library], Boolean)] = consolidater(libId) { libId =>
+  def suggestedLibraries(libId: Id[Library]): Future[Seq[RelatedLibrary]] = consolidater(libId) { libId =>
     for {
-      related <- relatedLibraries(libId)
+      topicRelated <- topicRelatedLibraries(libId)
+      ownerLibs <- librariesFromSameOwner(libId)
```
joshm1: Looks like `librariesFromSameOwner(libId)` can be a wasted call if `topicRelated` has enough qualified libraries?
Same for `topFollowedLibraries()` below.  If we want to call them anyway (incase some get filtered out later) you can assign the futures to vals before the for comprehension so they run in parallel.

yingjieMiao: Good point.

yingjieMiao:
@joshm1  updated.

PR #12798, opened 2014-12-18
Merge Search and Explain code + librarySearch explainer logic complete #Portland

leogrim:
@ymatsuda @stkem 
I still need to write some controller code for the library search explainer to be demoed tomorrow, but the logic is complete. 

I'll also make some adjustments to the uri search explainer e.g. move the additional boosting logic from `UriSearch` into `UriResultCollector` and add it to the `UriSearchExplanation`. 

Will do before I reach Europe :)


ymatsuda:
Awesome!

PR #12801, opened 2014-12-18
style radio buttons on manage profile settings page =D

ahsu1230:
Thanks @yipingliao for the advice before!
@andrewconner @2is10 

and removing some things from twitter connect controller now that `profileUrl` is in

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/profile/profile.styl
+++ b/kifi-backend/modules/shoebox/angular/src/profile/profile.styl
@@ -313,7 +313,7 @@ input.profile-last-name
     border-radius: 50%
     content: ' '
 
-input.profile-manage-input:checked ~ .profile-manage-radio-check:after
+.profile-manage-input.profile-manage-input:checked ~ .profile-manage-radio-check:after
   display: block
   left: 2px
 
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/angular/src/profile/profile.styl
```diff
@@ -268,44 +268,59 @@ input.profile-last-name
...
-    border-radius: 50%
+    width: 8px
+    height: 8px
     background-color: darkBlueColor
+    border-radius: 50%
+    content: ' '
 
-.profile-manage-settings-option-text
-  cursor: pointer
+.profile-manage-input.profile-manage-input:checked ~ .profile-manage-radio-check:after
```
2is10, line 316: This selector can be shortened to `:checked + .profile-manage-radio-check::after`. `+` is for adjacent siblings. Also, a single colon works to create pseudoelements, but we prefer the standard two colons (`::before` and `::after`).

2is10:
:+1:

yipingliao:
I may be remembering the wrong thing, but did this change help with those `_.isBoolean()` checks from before?

ahsu1230:
unfortunately not. the watch still picks up the changed values from `undefined` to whatever initial value they should be in

PR #12802, opened 2014-12-18
remove trailing slashes from URL paths automatically

2is10:
@yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/route.js
+++ b/kifi-backend/modules/shoebox/angular/src/route.js
@@ -11,14 +11,19 @@ angular.module('kifi')
     $httpProvider.defaults.withCredentials = true;
 
     // URL redirects.
+    var trailingSlashRe = /^([^?#]+)\/([?#].*|$)/;
     $urlRouterProvider
+      .rule(function ($injector, $location) {
+        var match = $location.url().match(trailingSlashRe);
+        if (match) {
+          return match[1] + match[2]; // remove trailing slash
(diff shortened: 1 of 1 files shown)


2is10:
This is broadly useful, but the use case that prompted it for me was deleting a library slug in the location bar to get to the user profile.

kifi-backend/modules/shoebox/angular/src/route.js
```diff
@@ -11,7 +11,14 @@ angular.module('kifi')
     $httpProvider.defaults.withCredentials = true;
 
     // URL redirects.
+    var trailingSlashRe = /^([^?#]+)\/([?#].*|$)/;
     $urlRouterProvider
+      .rule(function ($injector, $location) {
+        var match = $location.url().match(trailingSlashRe);
+        if (match) {
+          return match[1] + match[2]; // remove trailing slash
```
yipingliao, line 19: I am curious about what `match[2]` is. Testing it against a string with a trailing slash gives me an empty string for `match[2]`.

yipingliao:
Nice!

kifi-backend/modules/shoebox/angular/src/route.js
2is10, line 19: It can be a query string and/or hash fragment.

yipingliao, line 19: Thanks for the explanation.

PR #12803, opened 2014-12-18
make the response type be html in popup so the relevant script will run.

eishay:
i think this was broken for couple of months now since the UserAction refactoring
Thanks to @2is10 for pointing the problem


eishay:
@raymondng @andrewconner 
need to review UserController#576

2is10:
Andrew & Ray, it seems the refactoring changed our default content type from `text/html` to `text/plain`.

andrewconner:
There is no default type in Play. It's just based on the type you pass to the Status class — `Html` is `text/html`, `String` is `text/plain`, etc.

andrewconner:
Previously, we basically overrode whatever Play decided it would be with `text/html`. Not a good pattern in general.

2is10:
Agreed.

PR #12806, opened 2014-12-18
twtr: DM-invite (WIP)

raymondng:
- Added DM-invite per spec from @weeewake 
- Added coverage for sending invite via Twitter.

Validated Twitter DM API locally as well.

WIP (error-handling needs work in general) but the functionality is there. 

@andrewconner


raymondng:
Went over main changes with @andrewconner in person. Will merge and build upon it. Feedback appreciated.

PR #12809, opened 2014-12-18
using ng-change instead of a $watch

2is10:
@ahsu1230 
cc @yipingliao


ahsu1230:
:+1: thanks for the fix. `ng-change` definitely makes it cleaner.

yipingliao:
Yay!

PR #12811, opened 2014-12-18
library membership visibility in FullLibraryInfo, createLibrary & modify...

ahsu1230:
...Library (for website only)

@andrewconner @eishay

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/model/LibraryMembership.scala
+++ b/kifi-backend/modules/common/app/com/keepit/model/LibraryMembership.scala
@@ -35,11 +35,7 @@ case class LibraryMembership(
     LibraryMembershipView(id = id.get, libraryId = libraryId, userId = userId, access = access, createdAt = createdAt, state = state, seq = seq, showInSearch = showInSearch)
 }
 
-case class LibraryMembershipVisibility(value: String)
-
-object LibraryMembershipVisibility {
-  implicit val format = Json.format[LibraryMembershipVisibility]
-}
+@json case class LibraryMembershipVisibility(value: String)
 
 object LibraryMembershipVisibilityStates {
   val HIDDEN = LibraryMembershipVisibility("hidden")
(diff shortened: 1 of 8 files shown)


kifi-backend/modules/common/app/com/keepit/model/LibraryMembership.scala
```diff
@@ -35,11 +35,7 @@ case class LibraryMembership(
     LibraryMembershipView(id = id.get, libraryId = libraryId, userId = userId, access = access, createdAt = createdAt, state = state, seq = seq, showInSearch = showInSearch)
 }
 
-case class LibraryMembershipVisibility(value: String)
-
-object LibraryMembershipVisibility {
-  implicit val format = Json.format[LibraryMembershipVisibility]
-}
+@json case class LibraryMembershipVisibility(value: String)
```
ahsu1230, line 38: using format, the json for FullLibraryInfo comes out to be `{... membershipVisibility : {"value" : "visible"}...}`
since `LibraryMembershipVisibility` only has one value, `@json` macro automatically turns the json to just "visible" which is much easier to handle =)

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileLibraryController.scala
```diff
@@ -64,8 +64,9 @@ class MobileLibraryController @Inject() (
     val newDescription = (json \ "newDescription").asOpt[String]
     val newVisibility = (json \ "newVisibility").asOpt[LibraryVisibility]
     val newSlug = (json \ "newSlug").asOpt[String]
-    val newColor = (json \ "color").asOpt[HexColor]
-    val res = libraryCommander.modifyLibrary(libId, request.userId, newName, newDescription, newSlug, newVisibility, newColor)
+    val newColor = (json \ "newColor").asOpt[HexColor]
```
ahsu1230, line 67: @thassss @jaredpetker in case, you're using colors... I'm changing request json for modifying libraries to `newColor` instead of `color` to keep consistency with the other parameters.

2is10, line 67: Interesting. This is the first endpoint I’ve seen that uses “new” prefixes.

ahsu1230:
Thanks for bringing up the discussion @2is10 
closing merge. Will go ahead and add `listed: Option[Boolean]` to `LibraryMembership` table
will leave `LibraryMembershipVisibility` for now =)

PR #12814, opened 2014-12-19
more error tracking, restricting emails to use fully qualified domain na...

eishay:
...mes
@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/public/js/auth.js
+++ b/kifi-backend/modules/shoebox/public/js/auth.js
@@ -28,7 +28,7 @@
 var kifi = {};
 kifi.form = (function () {
   'use strict';
-  var emailAddrRe = /^[a-zA-Z0-9.!#$%&'*+\/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$/;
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/public/js/auth.js
```diff
@@ -231,6 +231,12 @@ kifi.form = (function () {
         if (o && o.error === 'user_exists_failed_auth') {
           Tracker.track('visitor_viewed_page', { error: 'incorrectPassword' });
           kifi.form.showError($password, 'Account exists, incorrect password');
+        } else if (o && o.error === 'error.email') {
+          Tracker.track('visitor_viewed_page', { error: 'errorEmail' });
+          kifi.form.showError($password, 'Error With Email Format');
```
2is10, line 236: The error message should be attached to the email address input field, not the password one. Also, we capitalize only the first word in the other error messages. When the JS doesn’t like the email address format it says “Invalid email address”. How about showing the same message when the server doesn’t like it?

``` js
kifi.form.showError($email, 'Invalid email address');
```

They’re tracked with different codes, so we’ll still be able to tell the difference in mixpanel.

kifi-backend/modules/shoebox/public/js/auth.js
```diff
@@ -231,6 +231,12 @@ kifi.form = (function () {
...
+        } else {
+          Tracker.track('visitor_viewed_page', { error: 'unknownSignupError' });
+          kifi.form.showError($password, 'Signup Error');
```
2is10, line 239: We typically apologize and give a suggestion. Maybe: `'Whoops, server error. Try again?'`.

PR #12817, opened 2014-12-19
hardening reactivelock

ymatsuda:
@stkem 

The existing `withLockFuture` has a problem when `f` throws an exception. The exception prevents `innerPromise.completeWith` from being executed.


andrewconner:
:+1:

stkem:
Oh, nice find! Thanks!

stkem:
lgtm

PR #12818, opened 2014-12-19
[DO NOT MERGE] User profiles header and rework some layout

yipingliao:
@andrewconner cc @2is10 
Hi Andrew,

As we discussed, this is a somewhat messy pull since I haven't had a chance to clean it yet (there are a couple of nested divs and ng-classes that may not be necessary), but I wanted to make this PR so you can take a look and see if anything jumps out at you.

I tested this on logged-in, logged-out, desktop, and mobile, and with the library search and infinite scroll, with related libraries and library pictures, and it's looking good, but I'd like to test it some more before merging and deploying.

Thanks!

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/app.styl
+++ b/kifi-backend/modules/shoebox/angular/src/app.styl
@@ -84,7 +84,7 @@ menu, ol, ul
   margin: 0 auto
 
 .kf-logged-out .kf-col-island
-  max-width: nonUserAppMaxWidth
+  // max-width: nonUserAppMaxWidth
   margin: 0 auto
 
 .kf-col
--- a/kifi-backend/modules/shoebox/angular/src/app.styl
+++ b/kifi-backend/modules/shoebox/angular/src/app.styl
@@ -83,10 +83,6 @@ menu, ol, ul
   max-width: userAppMaxWidth
   margin: 0 auto
 
-.kf-logged-out .kf-col-island
-  // max-width: nonUserAppMaxWidth
-  margin: 0 auto
-
 .kf-col
   display: inline-block
   vertical-align: top
--- a/kifi-backend/modules/shoebox/angular/src/common/mobile.styl
+++ b/kifi-backend/modules/shoebox/angular/src/common/mobile.styl
@@ -130,6 +130,8 @@
           padding: 18px 50px
           border-radius: 7px
 
+  .kf-main-library-keeps
+
     .kf-main-keeps
 
       .kf-keep-description
--- a/kifi-backend/modules/shoebox/angular/src/common/mobile.styl
+++ b/kifi-backend/modules/shoebox/angular/src/common/mobile.styl
@@ -50,91 +50,85 @@
   .kf-col-island
     margin-top: 120px
 
-  .kf-main-library-keeps
+  .kf-library-card-header
+    margin-bottom: -10px
 
-    .kf-library-card-header
-      margin-bottom: -10px
+    .library-card
+      overflow: hidden
 
-      .library-card
-        overflow: hidden
+      .kf-keep-lib-name
+        font-size: 55px
 
-        .kf-keep-lib-name-container
-          padding-top: 70px
+      .kf-keep-lib-info
 
-          .kf-keep-lib-name
-            font-size 55px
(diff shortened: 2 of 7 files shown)


kifi-backend/modules/shoebox/angular/src/app.styl
```diff
@@ -83,10 +83,6 @@ menu, ol, ul
   max-width: userAppMaxWidth
   margin: 0 auto
 
-.kf-logged-out .kf-col-island
```
yipingliao, line 86: Removing this gives us the screen-wide background for the library card header as well as the user profile header.

kifi-backend/modules/shoebox/angular/src/common/mobile.styl
```diff
@@ -48,93 +48,89 @@
     margin: 24px 0 36px
 
   .kf-col-island
-    margin-top: 90px
+    margin-top: 120px
 
-  .kf-main-library-keeps
```
yipingliao, line 53: I moved `.kf-library-card-header`, previously nested under `.kf-main-library-keeps`, to be sibling to it, so a lot of these indentations are for that.

kifi-backend/modules/shoebox/angular/src/libraries/library.styl
```diff
@@ -14,6 +14,11 @@ invite-header-orange = #ff7c3c
 .delighted-visible ~ div .kf-library-card-header
   margin-top: 10px
 
+.kf-logged-out .kf-main-library-keeps
+  width: nonUserAppMaxWidth
```
yipingliao, line 23: Constraining the width here instead.

kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
```diff
@@ -57,24 +56,6 @@ angular.module('kifi')
         //
         // Internal methods.
         //
-        function init() {
-          if (scope.isUserLoggedOut && !platformService.isSupportedMobilePlatform()) {
```
yipingliao, line 61: We can now remove some hacks added when infinite scroll was broken. :)

kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.styl
```diff
@@ -77,3 +84,33 @@ a.kf-user-profile-libraries-nav-link
 .kf-user-profile-libraries-nav-link-text
   position: relative
   top: 10px
+
+
+//
```
yipingliao, line 94: In general, as we expose more stuff on mobile, I think that mobile styles should be defined in the file where the styles for that page/component is defined. Currently there are a lot of stuff in `mobile.styl` that are for the library because that was the only thing on mobile, but I'd like to move them out too so that `mobile.styl` should only contain the global mobile styles.

andrewconner, line 94: Actually, I think it's better for there to be a separate mobiles style file, and we only serve one or the other. Will take a little bit of gulp work, but will make things a LOT cleaner.

kifi-backend/modules/shoebox/angular/src/app.styl
2is10, line 86: sweet!

kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
2is10, line 61: Yay!

kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.styl
2is10, line 94: I tend to agree with Andrew, that it’s better to serve less CSS to mobile clients. To restate what I think Andrew meant, we can probably achieve that while also authoring the mobile styles and desktop styles for each widget close together in our source code.

yipingliao, line 94: So, for now I am not going to move the mobile styles out into the separate file?

yipingliao:
Closing this and then reopening a rebased/conflicts-free pull request.

PR #12819, opened 2014-12-19
add subsource to reco action events and rename reco clientType to source

joshm1:
"source" is consistently used for analytics
renamed RecommendationClientType to RecommendationSource so it's consistent with the event property and it's not confusing to other devs in the future, and i thought it would've been confusing to have a RecommendationSubSource without a RecommendationSource

@stkem


stkem:
lgtm

PR #12820, opened 2014-12-19
Add library explainer endpoint #Amsterdam

leogrim:
@stkem @ymatsuda


kifi-backend/modules/shoebox/app/com/keepit/admin/AdminSearchController.scala
```diff
@@ -16,16 +17,17 @@ case class ArticleSearchResultHitMeta(uri: NormalizedURI, hit: ArticleHit)
...
+    implicit val publicIdConfig: PublicIdConfiguration) extends AdminUserActions with Logging {
 
-  def explain(query: String, uriId: Id[NormalizedURI], lang: String, debug: Option[String]) = AdminUserPage.async { request =>
-    searchClient.explainResult(query, request.userId, uriId, lang, debug).map(Ok(_))
+  def explainUriResult(query: String, uriId: Id[NormalizedURI], lang: String, debug: Option[String]) = AdminUserPage.async { request =>
+    searchClient.explainUriResult(query, request.userId, uriId, lang, debug).map(Ok(_))
+  }
+
+  def explainLibraryResult(query: String, libraryId: PublicId[Library], lang: String, debug: Option[String]) = AdminUserPage.async { request =>
+    searchClient.explainLibraryResult(query, request.userId, Library.decodePublicId(libraryId).get, lang, debug).map(Ok(_))
```
leogrim, line 30: @ymatsuda given that the explainer now runs against the real search, it should no longer be necessary to explicitly supply a language is it?

ymatsuda, line 30: I see. Good point.

PR #12822, opened 2014-12-19
simplifying user profile library nav links code and layout

2is10:
@yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.styl
+++ b/kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.styl
@@ -33,47 +33,28 @@
       margin-left: 5px
       content: '·'
 
-.kf-user-profile-libraries-nav-container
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.styl
```diff
@@ -33,47 +33,28 @@
...
-  display: inline-block
-  box-sizing: border-box
-  width: 25%
-  max-width: 180px
-  min-width: 100px
+.kf-upl-nav-link
+  display: table-cell
+  width: 33%
   height: 38px
+  line-height: 38px
```
2is10, line 53: FYI: `line-height` does vertical centering of text

2is10:
merging b/c I just remembered Yiping is out today

yipingliao:
Thanks!

PR #12824, opened 2014-12-19
Bumping prob. experiment generator cache version

stkem:
it contains stale data with a 30 day TTL because I modified the db
directly. I’ll have another pull request coming to avoid this problem
in the future.

@yingjieMiao


yingjieMiao:
thanks

PR #12826, opened 2014-12-19
adding library image to library JSON for user profile

2is10:
@eishay


kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -1243,35 +1244,42 @@ class LibraryCommander @Inject() (
...
+          libraryMembershipRepo.getOwnedLibrariesForOtherUser(user.id.get, other.id.get, page) map libraryRepo.get //cached
+      }
+      libs map profileLibraryView(idealSize)
     }
-    libs map profilelibraryView
   }
 
-  private def profilelibraryView(library: Library)(implicit session: RSession) = {
+  private def profileLibraryView(idealSize: ImageSize)(library: Library)(implicit session: RSession): ProfileLibraryView = {
+    val image = ProcessedImageSize.pickBestImage(idealSize, libraryImageRepo.getForLibraryId(library.id.get))
```
2is10, line 1262: We probably ought to optimize these image lookups into a single query, or add caching.

2is10:
merging so we can use the images in development—happy to receive any comments

PR #12828, opened 2014-12-19
specifying height and width for user profile header picture

2is10:
Specifying the size of an `` in the HTML or CSS allows the browser to reserve the proper amount of space for it until it loads, so the user won’t see content below jump down when the image loads.

@yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
+++ b/kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
@@ -67,7 +67,7 @@ angular.module('kifi')
       $scope.profile.picUrl = keepWhoService.getPicUrl({
         id: $scope.profile.id,
         pictureName: $scope.profile.pictureName
-      }, 100);
+      }, 200);
     }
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
```diff
@@ -67,7 +67,7 @@ angular.module('kifi')
       $scope.profile.picUrl = keepWhoService.getPicUrl({
         id: $scope.profile.id,
         pictureName: $scope.profile.pictureName
-      }, 100);
+      }, 200);
```
2is10, line 70: we use a minimum of 2x resolution:
- 100 when displaying in a space 50x50 or smaller
- 200 when displaying in a space 51x51 to 100x100
- we don’t display the user picture in a space larger than 100x100

yipingliao, line 70: Ah, I see. Thanks!

PR #12830, opened 2014-12-19
renaming "my" to "own" in user profile library route state

2is10:
Changes here are for consistency with minor changes we made today in the library server endpoints.

cc @yipingliao


kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.styl
```diff
@@ -60,3 +60,6 @@
   &.kf-selected
     background: #fff
     color: #000
+
+  &.kf-my::before
+    content: 'My '
```
yipingliao, line 65: :)

yipingliao:
:+1:

PR #12833, opened 2014-12-19
library recommendation feedback plumbing in curator

joshm1:
still need clicked/trashed/followed to be wired up from shoebox and angular
still need to increment delivered for lib recos

@stkem


stkem:
Appears to be missing the evolution to add `delivered`, `clicked` and `trashed` columns, otherwise lgtm

PR #12834, opened 2014-12-19
twtr: improve error handling

raymondng:
Various error handling related improvements:
- more consistent & DRY-er implementation
- update `UserValue` with current cursor (record only for now)
- update `SocialUserInfo` with `APP_NOT_AUTHORIZED` for invalid/expired token

Note that now `TwitterSocialGraph` generates more noise than other graphs -- `LinkedInSocialGraph` for instance does not airbrake at all. Added a toggle to reduce noise should that become annoying.

@andrewconner

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/model/UserValue.scala
+++ b/kifi-backend/modules/common/app/com/keepit/model/UserValue.scala
@@ -65,6 +65,10 @@ object UserValueName {
   val LATEST_EMAIL_TIPS_SENT = UserValueName("latest_email_tips")
   val LIBRARY_SORTING_PREF = UserValueName("library_sorting_pref")
 
+  val TWITTER_FOLLOWERS_CURSOR = UserValueName("twitter_followers_cursor")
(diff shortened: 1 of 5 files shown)


raymondng:
Will merge and validate. Feedback appreciated.

kifi-backend/modules/common/app/com/keepit/model/UserValue.scala
```diff
@@ -65,6 +65,10 @@ object UserValueName {
   val LATEST_EMAIL_TIPS_SENT = UserValueName("latest_email_tips")
   val LIBRARY_SORTING_PREF = UserValueName("library_sorting_pref")
 
+  val TWITTER_FOLLOWERS_CURSOR = UserValueName("twitter_followers_cursor")
```
andrewconner, line 68: What are these used for?

raymondng, line 68: Just for tracking purposes for now. If we hit rate limit these will be recorded. Going forward we can pick it up from where it's left off, etc.

PR #12836, opened 2014-12-19
Adding listed to Library Membership

ahsu1230:
As discussed with several people,
`LibraryMembershipVisibility` is a bit unclear to demonstrate that a library does/does not show up on a user's public profile. Instead, we'll have a `listed: Boolean` which is defaulted to true.
I also set `showInSearch`, `visibility` to have default values so there is no requirement for explicitly setting for `LibraryMembership` each usage.

In next PR, we'll remove LibraryMembershipVisibility and the visibility column

@andrewconner @2is10 @eishay


kifi-backend/modules/shoebox/app/com/keepit/model/LibraryMembership.scala
```diff
@@ -1,58 +0,0 @@
-package com.keepit.model
-
-import com.keepit.common.cache.{ PrimitiveCacheImpl, JsonCacheImpl, FortyTwoCachePlugin, CacheStatistics, Key }
```
ahsu1230, line 3: This file is also named `LibraryMembership` (but in shoebox)
and the formatter & caches can be moved to LibraryMembership in common. It's also easier to find (it took me a while to find this formatter T_T)

2is10:
I think this pull request would be clearer if `visibility` were removed at the same time—a clean swap. We already made the `visibility` column optional in the database, so there’s no hard requirement that this be done in stages. Just something to consider.

ahsu1230:
@2is10 Oh right, I just removed LibraryMembershipVisibility. So, let's remove `visibility` column from the table now?

2is10:
Only remove `visibility` column from the database after all shoebox servers have new code that no longer expects it to be there, and are stable (no chance we’ll need to roll back).

kifi-backend/modules/common/test/com/keepit/model/LibraryMembershipFactory.scala
```diff
@@ -9,7 +9,7 @@ object LibraryMembershipFactory {
 
   def membership(): PartialLibraryMembership = {
     new PartialLibraryMembership(LibraryMembership(id = Some(Id[LibraryMembership](idx.incrementAndGet())),
-      libraryId = Id[Library](idx.incrementAndGet()), userId = Id[User](idx.incrementAndGet()), access = LibraryAccess.OWNER, showInSearch = true, visibility = LibraryMembershipVisibilityStates.VISIBLE))
+      libraryId = Id[Library](idx.incrementAndGet()), userId = Id[User](idx.incrementAndGet()), access = LibraryAccess.OWNER, showInSearch = true, listed = true))
```
2is10: remove optional args?

kifi-backend/modules/common/test/com/keepit/model/LibraryMembershipFactory.scala
```diff
@@ -24,7 +24,7 @@ object LibraryMembershipFactory {
     def withUser(id: Int) = new PartialLibraryMembership(membership.copy(userId = Id[User](id)))
     def withUser(id: Id[User]) = new PartialLibraryMembership(membership.copy(userId = id))
     def withUser(user: User) = new PartialLibraryMembership(membership.copy(userId = user.id.get))
-    def invisible() = new PartialLibraryMembership(membership.copy(visibility = LibraryMembershipVisibilityStates.HIDDEN))
+    def invisible() = new PartialLibraryMembership(membership.copy(listed = false))
```
2is10: `unlisted`?

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminLibraryController.scala
```diff
@@ -62,7 +62,7 @@ class AdminLibraryController @Inject() (
       libraryMembershipRepo.save(currentOwnership.copy(access = LibraryAccess.READ_ONLY))
       libraryMembershipRepo.getWithLibraryIdAndUserId(libraryId, toUserId) match {
         case None =>
-          libraryMembershipRepo.save(LibraryMembership(userId = toUserId, libraryId = libraryId, access = LibraryAccess.OWNER, showInSearch = currentOwnership.showInSearch, visibility = currentOwnership.visibility))
+          libraryMembershipRepo.save(LibraryMembership(userId = toUserId, libraryId = libraryId, access = LibraryAccess.OWNER, showInSearch = currentOwnership.showInSearch, listed = currentOwnership.listed))
```
2is10: Would this be equivalent, clearer and safer?

``` scala
libraryMembershipRepo.save(currentOwnership.copy(id = None, userId = toUserId))
```

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryMembershipRepo.scala
```diff
@@ -67,8 +67,8 @@ class LibraryMembershipRepoImpl @Inject() (
     def showInSearch = column[Boolean]("show_in_search", O.NotNull)
     def lastViewed = column[Option[DateTime]]("last_viewed", O.Nullable)
     def lastEmailSent = column[Option[DateTime]]("last_email_sent", O.Nullable)
-    def visibility = column[LibraryMembershipVisibility]("visibility", O.NotNull)
-    def * = (id.?, libraryId, userId, access, createdAt, updatedAt, state, seq, showInSearch, visibility, lastViewed, lastEmailSent) <> ((LibraryMembership.apply _).tupled, LibraryMembership.unapply)
+    def listed = column[Boolean]("listed", O.NotNull)
+    def * = (id.?, libraryId, userId, access, createdAt, updatedAt, state, seq, showInSearch, listed, lastViewed, lastEmailSent) <> ((LibraryMembership.apply _).tupled, LibraryMembership.unapply)
```
2is10, line 71: I think `visibility` is used in some raw sql queries below (in this repo).

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminLibraryController.scala
ahsu1230: I see, so the .save() will assign a new `LibraryMembership id`

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryMembership.scala
2is10, line 3: Is this extra code needed outside of shoebox? Another option might be to put a comment at the bottom of the common LibraryMembership.scala indicating that helper classes can be found in ____.scala in the shoebox module. I don’t feel strongly—Andrew or Eishay might have opinions.

2is10:
nice work. thanks.

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryMembershipRepo.scala
ahsu1230, line 71: they're removed now! Thanks.

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryMembership.scala
ahsu1230, line 3: Sure, would be glad to move them into shoebox again. Maybe a new file name would be better though =)

ahsu1230:
Thanks @2is10 I'll merge and once the build is good, I'll make the mysql change

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryMembershipRepo.scala
```diff
@@ -272,9 +272,9 @@ class LibraryMembershipRepoImpl @Inject() (
   def countLibrariesOfUserFromAnonymos(userId: Id[User], countFollowLibraries: Boolean)(implicit session: RSession): Int = {
     import StaticQuery.interpolation
     val query = if (countFollowLibraries) {
-      sql"select count(*) from library_membership lm, library lib where lm.library_id = lib.id and lm.user_id = $userId and lib.state = 'active' and lm.state = 'active' and lm.visibility = 'visible' and lib.visibility = 'published'"
+      sql"select count(*) from library_membership lm, library lib where lm.library_id = lib.id and lm.user_id = $userId and lib.state = 'active' and lm.state = 'active' and lib.visibility = 'published'"
     } else {
-      sql"select count(*) from library_membership lm, library lib where lm.library_id = lib.id and lm.user_id = $userId and lib.state = 'active' and lm.state = 'active' and lm.visibility = 'visible' and lib.visibility = 'published' and lm.access='owner'"
+      sql"select count(*) from library_membership lm, library lib where lm.library_id = lib.id and lm.user_id = $userId and lib.state = 'active' and lm.state = 'active' and lib.visibility = 'published' and lm.access='owner'"
```
2is10, line 277: I think we should have replaced `lm.visibility = 'visible'` → `lm.listed` instead of just deleting the clauses.

PR #12847, opened 2014-12-22
Randomness in nonlinear scoring

ymatsuda:
@yingjieMiao @stkem 

Making the non-linear scoring strategy to use linear scoring for 10% of URIs.


yingjieMiao:
lgtm

PR #12849, opened 2014-12-22
stubbing out "filter" request parameter and updating response format

2is10:
@ahsu1230


kifi-backend/modules/sqldb/app/com/keepit/common/db/slick/Database.scala
```diff
@@ -1,6 +1,5 @@
 package com.keepit.common.db.slick
 
-
```
2is10, line 3: auto-formatter removed this line

ahsu1230:
lgtm :+1:

PR #12850, opened 2014-12-22
ignoring objects already in the db, updating state if needed and trying ...

eishay:
...to insert only new image objects

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepImageCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepImageCommander.scala
@@ -252,8 +252,20 @@ class KeepImageCommanderImpl @Inject() (
       }
       ImageProcessState.ExistingStoredImagesFound(saved)
     } else {
-      val copiedImages = existingSameHash.map { prev =>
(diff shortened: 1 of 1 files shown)


eishay:
@ahsu1230 @andrewconner

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepImageCommander.scala
```diff
@@ -252,8 +252,20 @@ class KeepImageCommanderImpl @Inject() (
...
+            None
+          } else {
+            log.info(s"re-activating image: $prev")
+            Some(prev.copy(state = KeepImageStates.ACTIVE, source = source))
+          }
+        } else {
+          val image = KeepImage(state = KeepImageStates.ACTIVE, keepId = keepId, imagePath = prev.imagePath, format = prev.format, width = prev.width, height = prev.height, source = source, sourceFileHash = prev.sourceFileHash, sourceImageUrl = prev.sourceImageUrl, isOriginal = prev.isOriginal)
+          log.info(s"saving new image [$image] derived from [$prev]")
+          Some(image)
+        }
```
2is10, line 268: Only the last `else` branch should ever execute, since we already know `allForThisKeep.isEmpty`.

Would it possible to implement this using `update` queries, falling back to `insert` when the update achieves nothing?

PR #12853, opened 2014-12-22
making sure we're not trying to persist clones of the same keep image ob...

eishay:
...ject


eishay:
@ahsu1230 @andrewconner

PR #12854, opened 2014-12-22
trying to unique the images even more using a smaller dimension set base...

eishay:
...d on the db index


eishay:
@andrewconner @ahsu1230

PR #12857, opened 2014-12-22
filtering followers for the profile page the same way we do for recommended libraries

2is10:
@ahsu1230


2is10:
merging, comments welcome

PR #12858, opened 2014-12-22
misc lda version clean up

yingjieMiao:
@ymatsuda @stkem @joshm1 

minor clean ups
- admin gets the default version from `cortexClient`, instead of hard-coded number
- renamed `ModelVersions.denseLDAVersion` to `ModelVersions.defaultLDAVersion`
- removed `PublishingVersions`. Semantically, this is equivalent to `ModelVersions.defaultLDAVersion`


ymatsuda:
:+1:

PR #12863, opened 2014-12-22
evolutions remove visibility column from library membership

ahsu1230:
@2is10

### DIFF
--- a/kifi-backend/modules/common/conf/evolutions/shoebox/278.sql
+++ b/kifi-backend/modules/common/conf/evolutions/shoebox/278.sql
@@ -2,11 +2,14 @@
 
 # --- !Ups
 
-DROP INDEX library_membership_i_visible ON library_membership;
-ALTER TABLE library_membership DROP COLUMN visibility;
+drop index library_membership_i_visible;
+alter table library_membership drop column visibility;
 
-CREATE INDEX library_membership_i_listed ON library_membership (listed);
+create index library_membership_i_listed on library_membership(listed);
 
-insert into evolutions (name, description) values('278.sql', 'drop visibility column from library membership');
+INSERT INTO evolutions (name, description) VALUES('278.sql', 'delete visibility column from library_membership');
+
 
 # --- !Downs
+
+
--- a/kifi-backend/modules/common/conf/evolutions/shoebox/278.sql
+++ b/kifi-backend/modules/common/conf/evolutions/shoebox/278.sql
@@ -2,8 +2,9 @@
 
 # --- !Ups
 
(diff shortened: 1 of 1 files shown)


2is10:
Looks like you’re also going to need to delete an index and create a new one.

``` sql
mysql> show create table library_membership \G
*************************** 1. row ***************************
       Table: library_membership
Create Table: CREATE TABLE `library_membership` (
  `id` bigint(20) NOT NULL AUTO_INCREMENT,
  `library_id` bigint(20) NOT NULL,
  `user_id` bigint(20) NOT NULL,
  `access` varchar(20) NOT NULL,
  `state` varchar(20) NOT NULL,
  `created_at` datetime NOT NULL,
  `updated_at` datetime NOT NULL,
  `seq` bigint(20) NOT NULL,
  `show_in_search` tinyint(1) NOT NULL DEFAULT '1',
  `last_viewed` datetime DEFAULT NULL,
  `last_email_sent` datetime DEFAULT NULL,
  `visibility` varchar(20) DEFAULT 'visible',
  `listed` tinyint(1) NOT NULL DEFAULT '1',
  PRIMARY KEY (`id`),
  UNIQUE KEY `library_membership_u_library_user` (`library_id`,`user_id`),
  KEY `library_member_f_user` (`user_id`),
  KEY `library_membership_i_seq` (`seq`),
  KEY `library_membership_i_last_email_sent` (`last_email_sent`),
  KEY `library_membership_i_last_viewed` (`last_viewed`),
  KEY `library_membership_i_visible` (`visibility`),
  CONSTRAINT `library_member_f_library` FOREIGN KEY (`library_id`) REFERENCES `library` (`id`),
  CONSTRAINT `library_member_f_user` FOREIGN KEY (`user_id`) REFERENCES `user` (`id`)
) ENGINE=InnoDB AUTO_INCREMENT=33377 DEFAULT CHARSET=utf8mb4
1 row in set (0.00 sec)
```

See `library_membership_i_visible` above.

ahsu1230:
oh good catch @2is10, replaced with a new index!

kifi-backend/modules/common/conf/evolutions/shoebox/278.sql
```diff
@@ -0,0 +1,11 @@
+# SHOEBOX
+
+# --- !Ups
+
+ALTER TABLE library_membership DROP COLUMN visibility;
+DROP INDEX library_membership_i_visible ON library_membership;
```
2is10: The index needs to be dropped before the column can be removed, but otherwise looks good.

ahsu1230:
evolutions applied

PR #12864, opened 2014-12-22
Add float property to related libraries bottom footer element.

yipingliao:
@2is10 cc @joshm1 

Fixes a problem where sometimes the related library card "run into" each other because the last element in that section was not floated. Could be caused by a race condition where this problem is visible when the data for the related libraries is returned with different delays.

Tested with logged-in, logged-out, and mobile views.

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/relatedLibraries.styl
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/relatedLibraries.styl
@@ -3,111 +3,96 @@ relatedLibrariesColumnWidth = 226px
 desktopBorderRadius = 19px
 mobileBorderRadius = 54px
 
-.kf-logged-out {
-  .kf-library-results-wrapper {
-    width: keepsColumnWidth;
-    float: left;
-  }
-  &.kf-mobile .kf-library-results-wrapper {
-    max-width: initial;
-    float: none;
-  }
+.kf-logged-out
+  .kf-library-results-wrapper
+    width: keepsColumnWidth
+    float: left
 
-  .kf-library-body {
-    min-width: keepsColumnWidth + relatedLibrariesColumnWidth;
-    .kf-main-keeps {
-      max-width: keepsColumnWidth;
+  &.kf-mobile .kf-library-results-wrapper
+    max-width: initial
+    float: none
+
+  .kf-library-body
+    min-width: keepsColumnWidth + relatedLibrariesColumnWidth
+    .kf-main-keeps
+      max-width: keepsColumnWidth
+
+  &.kf-mobile .kf-library-body
+    min-width: initial
+    .kf-main-keeps
+      max-width: initial
+
+.kf-related-libraries
+  margin-top: 42px
+  float: left
+  width: relatedLibrariesColumnWidth
+  box-sizing: border-box
+  padding: 0
+  .kf-mobile &
+    margin-top: 0
+    float: none
+    width: initial
+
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/relatedLibraries.styl
```diff
@@ -3,111 +3,96 @@ relatedLibrariesColumnWidth = 226px
 desktopBorderRadius = 19px
 mobileBorderRadius = 54px
 
-.kf-logged-out {
```
yipingliao, line 6: Most changes are in converting CSS syntax to Stylus syntax.

kifi-backend/modules/shoebox/angular/src/libraries/relatedLibraries.styl
```diff
@@ -3,111 +3,96 @@ relatedLibrariesColumnWidth = 226px
...
-  width: 100%;
-  margin-bottom: 1em;
-  .kf-mobile & {
-    margin: 40px 25px 10px 30px;
-    float: none;
-    width: auto;
-  }
-}
+.kf-related-library-footer
+  float: left
```
yipingliao, line 98: This is the CSS rule being added.

kifi-backend/modules/shoebox/angular/src/libraries/relatedLibraries.styl
```diff
@@ -3,111 +3,96 @@ relatedLibrariesColumnWidth = 226px
...
-    width: initial;
-  }
-}
+  &.kf-mobile .kf-library-body
+    min-width: initial
+    .kf-main-keeps
+      max-width: initial
 
-.kf-related-libraries-see-more {
-  display: block;
```
yipingliao, line 44: @2is10 Removed display block.

kifi-backend/modules/shoebox/angular/src/libraries/relatedLibraries.tpl.html
```diff
@@ -1,14 +1,14 @@
-<div class="kf-related-libraries">
-  <div class="kf-related-library-explore-more-header" ng-if="!$root.userLoggedIn">Explore More Kifi Libraries</div>
+<section class="kf-related-libraries">
+  <header class="kf-related-library-explore-more-header" ng-if="!$root.userLoggedIn">Explore More Kifi Libraries</header>
   <div class="kf-related-library-card" ng-repeat="library in relatedLibraries">
     <div kf-small-library-card library="library" origin="rl" action="clickedLibraryRec"></div>
   </div>
-  <div ng-if="!$root.userLoggedIn">
+  <footer ng-if="!$root.userLoggedIn" class="kf-related-library-footer">
```
yipingliao, line 6: Added `kf-related-library-footer` class to float this.

kifi-backend/modules/shoebox/angular/src/libraries/relatedLibraries.tpl.html
```diff
@@ -1,14 +1,14 @@
-<div class="kf-related-libraries">
-  <div class="kf-related-library-explore-more-header" ng-if="!$root.userLoggedIn">Explore More Kifi Libraries</div>
+<section class="kf-related-libraries">
```
yipingliao, line 1: Using `section`, `header`, and `footer` tags since I started calling the bottom element "footer".

2is10:
:+1: thanks!

PR #12868, opened 2014-12-23
adding calls to get which libraries does a user follow 

eishay:
and refactoring out some classes from library commander

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
@@ -1280,21 +1280,55 @@ class LibraryCommander @Inject() (
         } else {
           (0, Seq.empty)
         }
-        LibraryCardInfo(
-          id = Library.publicId(lib.id.get),
-          name = lib.name,
-          description = lib.description,
-          color = lib.color,
-          image = image.map(LibraryImageInfo.createInfo),
-          slug = lib.slug,
-          owner = ownerBasicUser,
-          numKeeps = numKeeps,
-          numFollowers = numFollowers,
-          followers = LibraryCardInfo.showable(followersSample, viewer.isDefined),
-          caption = None)
+        createLibraryCardInfo(lib, image, ownerBasicUser, numKeeps, numFollowers, followersSample, viewer.isDefined)
       }
     }
   }
+
+  def getFollowingLibraries(owner: User, viewer: Option[User], page: Paginator, idealSize: ImageSize): Seq[LibraryCardInfo] = {
+    val ownerBasicUser = BasicUser.fromUser(owner)
+    db.readOnlyMaster { implicit session =>
+      val libs = viewer match {
+        case None =>
+          libraryRepo.getLibrariesUserFollowFromAnonymos(owner.id.get, page)
+        case Some(other) if other.id == owner.id =>
+          libraryRepo.getLibrariesOfSelf(owner.id.get, page)
+        case Some(other) =>
+          libraryMembershipRepo.getOwnedLibrariesForOtherUser(owner.id.get, other.id.get, page) map libraryRepo.get //cached
+      }
+      libs map { lib => // may want to optimize queries below into bulk queries
+        val image = ProcessedImageSize.pickBestImage(idealSize, libraryImageRepo.getForLibraryId(lib.id.get))
+        val numKeeps = keepRepo.getCountByLibrary(lib.id.get)
+        val (numFollowers, followersSample) = if (lib.memberCount > 1) {
+          val count = libraryMembershipRepo.countWithLibraryIdAndAccess(lib.id.get, LibraryAccess.READ_ONLY)
+          val sample = libraryMembershipRepo.pageWithLibraryIdAndAccess(lib.id.get, 0, 2, Set(LibraryAccess.READ_ONLY)) map { lm =>
+            basicUserRepo.load(lm.userId)
+          }
+          (count, sample)
+        } else {
+          (0, Seq.empty)
+        }
+        createLibraryCardInfo(lib, image, ownerBasicUser, numKeeps, numFollowers, followersSample, viewer.isDefined)
+      }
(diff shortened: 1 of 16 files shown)


eishay:
@ahsu1230

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryView.scala
```diff
@@ -0,0 +1,155 @@
+package com.keepit.model
```
ahsu1230, line 1: thanks for separating these classes out of `LibraryCommander`!

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -1268,171 +1268,62 @@ class LibraryCommander @Inject() (
...
+      val libs = viewer match {
+        case None =>
+          val showFollowLibraries = getUserValueSetting(owner.id.get, UserValueName.SHOW_FOLLOWED_LIBRARIES)
+          if (!showFollowLibraries) Seq.empty
+          else libraryMembershipRepo.getFollowingLibrariesFromAnonymos(owner.id.get, page) map libraryRepo.get //cached
+        case Some(other) if other.id == owner.id =>
+          libraryMembershipRepo.getFollowingLibrariesOfSelf(owner.id.get, page) map libraryRepo.get //cached
+        case Some(other) =>
+          val showFollowLibraries = getUserValueSetting(owner.id.get, UserValueName.SHOW_FOLLOWED_LIBRARIES)
+          if (!showFollowLibraries) Seq.empty
```
ahsu1230, line 1287: just a minor preference: it would be a bit easier to read if we had the `true` case first in the if-block

```
if (showFollowLibraries) ...
else ...
```

because `showFollowLibraries` naming already implies the boolean value.
If we wanted to falsey case first, then naming the variable `hideFollowLibraries` would be more appropriate. But, just preferences =)

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -1268,171 +1268,62 @@ class LibraryCommander @Inject() (
...
+          else libraryMembershipRepo.getFollowingLibrariesForOtherUser(owner.id.get, other.id.get, page) map libraryRepo.get //cached
+      }
+      createLibraryCardInfos(libs, ownerBasicUser, idealSize, viewer.isDefined)
```
ahsu1230, line 1290: @eishay correct me if I'm wrong about this. But if we want to show libraries I follow on my profile page, `createLibraryCardInfos` shouldn't take in a single `BasicUser`parameter since we want to library owner information of every library in `libs` (which could be different owners). So `createLibraryCardInfos` will probably need to map over every library and retreive a BasicUser from the library owner Id. Hope I'm understanding it correctly.

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryMembershipRepo.scala
```diff
@@ -312,7 +315,31 @@ class LibraryMembershipRepoImpl @Inject() (
...
+
+  def getFollowingLibrariesOfSelf(userId: Id[User], page: Paginator)(implicit session: RSession): Seq[Id[Library]] = {
+    import StaticQuery.interpolation
+    val query = sql"select lib.id from library_membership lm, library lib where lm.library_id = lib.id and lm.user_id = $userId and lib.state = 'active' and lm.state = 'active' and lm.access != 'owner' order by lm.id desc limit ${page.itemsToDrop}, ${page.size}"
+    query.as[Id[Library]].list
+  }
+
+  def getFollowingLibrariesFromAnonymos(userId: Id[User], page: Paginator)(implicit session: RSession): Seq[Id[Library]] = {
+    import StaticQuery.interpolation
+    val query = sql"select lib.id from library_membership lm, library lib where lm.library_id = lib.id and lm.user_id = $userId and lib.state = 'active' and lm.state = 'active' and lm.listed and lib.visibility = 'published' and lm.access != 'owner' order by lm.id desc limit ${page.itemsToDrop}, ${page.size}"
```
2is10, line 342: would prepared statements make these much faster on average?

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryMembershipRepo.scala
```diff
@@ -43,6 +43,9 @@ trait LibraryMembershipRepo extends Repo[LibraryMembership] with RepoWithDelete[
   def countLibrariesToSelf(userId: Id[User])(implicit session: RSession): Int
   def countLibrariesForOtherUser(userId: Id[User], friendId: Id[User], countFollowLibraries: Boolean)(implicit session: RSession): Int
   def getOwnedLibrariesForOtherUser(userId: Id[User], friendId: Id[User], page: Paginator)(implicit session: RSession): Seq[Id[Library]]
+  def getFollowingLibrariesForOtherUser(userId: Id[User], friendId: Id[User], page: Paginator)(implicit session: RSession): Seq[Id[Library]]
+  def getFollowingLibrariesOfSelf(userId: Id[User], page: Paginator)(implicit session: RSession): Seq[Id[Library]]
+  def getFollowingLibrariesFromAnonymos(userId: Id[User], page: Paginator)(implicit session: RSession): Seq[Id[Library]]
```
2is10, line 48: `Anonymous` (sp.)

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
eishay, line 1287: good point

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
eishay, line 1290: right! my bad, i'll add a test for that

PR #12870, opened 2014-12-23
first cut at user profile library cards

2is10:
@yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/userProfile/userProfileActionService.js
+++ b/kifi-backend/modules/shoebox/angular/src/userProfile/userProfileActionService.js
@@ -5,21 +5,27 @@ angular.module('kifi')
 .factory('userProfileActionService', [
   '$http', '$q', 'routeService', 'Clutch',
   function ($http, $q, routeService, Clutch) {
-    var clutchParamsRecos = {
+    var clutchParams = {
       cacheDuration: 2000
     };
 
+    function getData(res) {
+      return res.data;
+    }
+
     var userProfileService = new Clutch(function (username) {
-      return $http.get(routeService.getUserProfile(username)).then(function (res) {
-        if (res && res.data) {
-          return res.data;
-        }
-      });
-    }, clutchParamsRecos);
+      return $http.get(routeService.getUserProfile(username)).then(getData);
+    }, clutchParams);
+    var userLibrariesService = new Clutch(function (username) {
+      return $http.get(routeService.getUserLibraries(username)).then(getData);
+    }, clutchParams);
 
     var api = {
       getProfile: function (username) {
         return userProfileService.get(username);
+      },
+      getLibraries: function (username) {
+        return userLibrariesService.get(username);
       }
     };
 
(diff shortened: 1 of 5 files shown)


yipingliao:
Yay! Will definitely look through this in detail.

kifi-backend/modules/shoebox/angular/src/userProfile/userProfileLibrariesList.tpl.html
```diff
@@ -1 +1,21 @@
...
+        <div class="kf-upl-card-name-clip">
+          {{lib.name}}
+        </div>
+      </div>
+    </div>
+    <div class="kf-upl-counts">
+      <ng-pluralize class="kf-upl-count" count="lib.numKeeps" when="{'1':'{} Keep','other':'{} Keeps'}" ng-if="lib.numKeeps > 0"></ng-pluralize>
+      <ng-pluralize class="kf-upl-count" count="lib.numFollowers" when="{'1':'{} Follower','other':'{} Followers'}" ng-if="lib.numFollowers > 0"></ng-pluralize>
+    </div>
+    <div class="kf-upl-card-byline">{{lib.owner.firstName}} {{lib.owner.lastName}}</div>
```
yipingliao, line 17: I am surprised to see that the owner's name is nested within the library's information and when clicked, will go to the library. I was expecting that the owner's name will be with the owner's picture and when clicked will go to the owner's user profile, like the picture. This kind of grouping will also be easier for tagging microdata.

kifi-backend/modules/shoebox/angular/src/userProfile/userProfileActionService.js
```diff
@@ -5,21 +5,27 @@ angular.module('kifi')
 .factory('userProfileActionService', [
   '$http', '$q', 'routeService', 'Clutch',
   function ($http, $q, routeService, Clutch) {
-    var clutchParamsRecos = {
+    var clutchParams = {
       cacheDuration: 2000
     };
 
+    function getData(res) {
+      return res.data;
```
yipingliao, line 13: Do you think it's okay to have removed the null check for `res`?

yipingliao:
Wow. I'm looking forward to seeing these library cards!

kifi-backend/modules/shoebox/angular/src/userProfile/userProfileLibrariesList.tpl.html
2is10, line 17: Yeah, I wasn’t sure how it should work. Maybe the whole author area should link to the author’s profile, but that’s also the area that shows the library’s cover image. Or maybe just the picture and name. For now I just made it the picture until I can have a conversation with Mark. Easy to change.

kifi-backend/modules/shoebox/angular/src/userProfile/userProfileActionService.js
2is10, line 13: Yes. From `$http` [doc](https://docs.angularjs.org/api/ng/service/$http#usage):

> Returns a `promise` object with the standard `then` method and two http specific methods: `success` and `error`. The `then` method takes two arguments a success and an error callback which will be called with a response object. ...

2is10:
Thanks for reviewing! Still a lot to do.

kifi-backend/modules/shoebox/angular/src/userProfile/userProfileLibrariesList.tpl.html
yipingliao, line 17: Ah, I see.

PR #12871, opened 2014-12-23
website & mobile endpoints for profile libraries

ahsu1230:
Thanks @eishay for the commander implementations
@2is10 @andrewconner 

should match the spec found in this googly doc
https://docs.google.com/a/kifi.com/document/d/1eR3smK3cOZibRy4IzbAwe6Um9acSn0dTdnT3A6f7KXo

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileLibraryController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileLibraryController.scala
@@ -354,30 +354,33 @@ class MobileLibraryController @Inject() (
         NotFound(username.value)
       case Some(user) =>
         val viewer = request.userOpt
+        val paginator = Paginator(page, pageSize)
+        val imageSize = ProcessedImageSize.Medium.idealSize
         filter match {
           case "own" =>
-            val libs = libraryCommander.getOwnProfileLibraries(user, viewer, Paginator(page, pageSize), ProcessedImageSize.Medium.idealSize)
+            val libs = libraryCommander.getOwnProfileLibraries(user, viewer, paginator, imageSize)
             Ok(Json.obj("own" -> libs.map(LibraryCardInfo.writesWithoutOwner.writes)))
 
           case "following" =>
-            val libs = libraryCommander.getFollowingLibraries(user, viewer, Paginator(page, pageSize), ProcessedImageSize.Medium.idealSize)
+            val libs = libraryCommander.getFollowingLibraries(user, viewer, paginator, imageSize)
             Ok(Json.obj("following" -> Json.toJson(libs)))
 
           case "invited" =>
-            val libs = libraryCommander.getInvitedLibraries(user, viewer, Paginator(page, pageSize), ProcessedImageSize.Medium.idealSize)
+            val libs = libraryCommander.getInvitedLibraries(user, viewer, paginator, imageSize)
             Ok(Json.obj("invited" -> Json.toJson(libs)))
 
-          case "all" =>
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileLibraryController.scala
```diff
@@ -344,6 +346,43 @@ class MobileLibraryController @Inject() (
...
+        val viewer = request.userOpt
+        filter match {
+          case "own" =>
+            val libs = libraryCommander.getOwnProfileLibraries(user, viewer, Paginator(page, pageSize), ProcessedImageSize.Medium.idealSize)
+            Ok(Json.obj("own" -> libs.map(LibraryCardInfo.writesWithoutOwner.writes)))
+
+          case "following" =>
+            //val libs = libraryCommander.getFollowingProfileLibraries(user, viewer, Paginator(page, pageSize), ProcessedImageSize.Medium.idealSize) // todo (aaron, eishay): implement retrieving following libraries
+            val libs = Seq.empty[LibraryCardInfo]
+            Ok(Json.obj("following" -> libs.map(LibraryCardInfo.writesWithoutOwner.writes)))
```
ahsu1230: woops changing to include owner

kifi-backend/modules/shoebox/test/com/keepit/controllers/website/LibraryControllerTest.scala
```diff
@@ -484,6 +484,55 @@ class LibraryControllerTest extends Specification with ShoeboxTestInjector {
...
+        contentType(result2) must beSome("application/json")
+        Json.parse(contentAsString(result2)) must equalTo(Json.parse(
+          s"""
+            {
+              "following": [
+                {
+                  "id":"${pubId3.id}",
+                  "name":"lib3",
+                  "slug":"lib3",
+                  "owner":{
```
ahsu1230, line 505: @eishay this might be a mistake
In this case, library3 is owned by user2 and is followed by user1. This endpoint is called by user1, so in the list of "following" libraries, the owner of this library should be user2 (not user1).

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileLibraryController.scala
```diff
@@ -344,6 +346,41 @@ class MobileLibraryController @Inject() (
...
+            Ok(Json.obj("following" -> Json.toJson(libs)))
+
+          case "invited" =>
+            val libs = libraryCommander.getInvitedLibraries(user, viewer, Paginator(page, pageSize), ProcessedImageSize.Medium.idealSize)
+            Ok(Json.obj("invited" -> Json.toJson(libs)))
+
+          case "all" =>
+            val ownLibs = libraryCommander.getOwnProfileLibraries(user, viewer, Paginator(page, pageSize), ProcessedImageSize.Medium.idealSize)
+            val followLibs = libraryCommander.getFollowingLibraries(user, viewer, Paginator(page, pageSize), ProcessedImageSize.Medium.idealSize)
+            val invitedLibs = libraryCommander.getInvitedLibraries(user, viewer, Paginator(page, pageSize), ProcessedImageSize.Medium.idealSize)
```
2is10: The six `libraryCommander` calls could be a bit shorter (more readable) if the last two arguments (common to all calls) were factored out to local `paginator` and `idealSize` vals before the `filter match`.

Also, we might want to ignore `page` in the `"all"` case and always use `0` (or return 400 status if it’s nonzero) because it never makes sense to paginate through all three filters at once. `filter=all` is only intended as a way to preload the first page for all three filters. And it could be expensive to answer those queries.

2is10:
:+1:

PR #12872, opened 2014-12-23
User profile gear and connect

yipingliao:
Add gear and connect icons to the user profile page, which are displayed when the viewing user is viewing her own page and when the viewing user can connect to the displayed user as a Kifi friend, respectively.

Also, add item property of "url" to the user's picture on the user profile page.

@2is10 cc @eishay


kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
```diff
@@ -83,6 +88,24 @@ angular.module('kifi')
...
+    $scope.connect = function () {
+      if ($scope.ignoreClick.connect) {
+        return;
+      }
+
+      var $connect = angular.element('.kf-user-profile-connect');
+      $connect.text('SENDING REQUEST...');
+      $scope.ignoreClick.connect = true;
+
+      inviteService.friendRequest($scope.profile.id).then(function () {
```
yipingliao, line 124: Clicking will result in a friend request being sent. Detailed UI still needs to be reviewed by Mark. I think we should have a dialog box that tells the user what action he is about to take before taking it.

kifi-backend/modules/shoebox/angular/src/userProfile/userProfileHeader.tpl.html
```diff
@@ -1,16 +1,16 @@
...
+         ng-click="connect()" ng-class="{'clickable': ignoreClick.connect === false}">
+      CONNECT
     </div>
+    <a ng-switch-when="settings" class="kf-user-profile-user-settings svg-settings-gear" ui-sref="settings"></a>
   </div>
 
   <div class="kf-user-profile-basic-info" itemscope itemtype="http://data-vocabulary.org/Person">
-    <img class="kf-user-profile-picture" ng-src="{{profile.picUrl || 'https://www.kifi.com/assets/img/ghost.200.png'}}"
-         alt="{{profile.firstName}} {{profile.lastName}}" itemprop="photo">
+    <a href="{{userProfileRootUrl}}" itemprop="url">
```
yipingliao, line 10: @eishay Adding url itemprop here. Will check with @andrewconner that this is also done with links on the site that link to this profile page.

kifi-backend/modules/shoebox/angular/src/userProfile/userProfileHeader.tpl.html
```diff
@@ -1,16 +1,16 @@
...
   <div ng-switch="optionalAction" class="kf-user-profile-optional-action">
-    <div ng-switch-when="connect" class="kf-user-profile-connect">
-      <!-- need connect graphic -->CONNECT
-    </div>
-    <div ng-switch-when="settings" class="kf-user-profile-user-settings">
-      <!-- need gear graphic -->SETTINGS
+    <div ng-switch-when="connect" class="kf-user-profile-connect svg-connect-to-user"
+         ng-click="connect()" ng-class="{'clickable': ignoreClick.connect === false}">
+      CONNECT
     </div>
```
2is10: We should use an `<a href="javascript:">`, `button`, or other form control for all click targets, for usability and accessibility. When a link is only sometimes clickable, I typically remove the `href` attribute when it’s not. Adding/removing the `disabled` attribute (no value) or equivalently toggling the `.disabled` property (`true`/`false`) is the standard way for form controls. These techniques are described in the HTML5 spec, and the browser properly handles some usability concerns for free when we use them.

2is10:
LGTM. Nice work.

kifi-backend/modules/shoebox/angular/src/userProfile/userProfileHeader.tpl.html
yipingliao: Will do. Thanks for the help!

PR #12873, opened 2014-12-23
add getInvitedLibraries

eishay:
sill push tests next, need to do some refactoring for it but at least have the controller be able to call this method

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
@@ -6,7 +6,7 @@ import com.keepit.common.akka.SafeFuture
 import com.keepit.common.cache._
 import com.keepit.common.crypto.{ PublicIdConfiguration, PublicId }
 import com.keepit.common.db.slick.DBSession.{ RSession, RWSession }
-import com.keepit.common.db.{ Id, ExternalId }
+import com.keepit.common.db.{ State, Id, ExternalId }
 import com.keepit.common.db.slick.Database
 import com.keepit.common.healthcheck.AirbrakeNotifier
 import com.keepit.common.logging.{ AccessLog, Logging }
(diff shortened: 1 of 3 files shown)


eishay:
@ahsu1230

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -1291,6 +1291,21 @@ class LibraryCommander @Inject() (
     }
   }
 
+  def getInvitedLibraries(owner: User, viewer: Option[User], page: Paginator, idealSize: ImageSize): Seq[LibraryCardInfo] = {
+    val ownerBasicUser = BasicUser.fromUser(owner)
```
2is10, line 1295: `owner` isn’t really the right name here. `invitee`, perhaps?

2is10, line 1295: In fact, only the invitee should be able to view this tab, correct? So `viewer` doesn’t need to be passed. Just one `User` argument, and it may as well be called `user`, I guess.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -1291,6 +1291,21 @@ class LibraryCommander @Inject() (
...
+        db.readOnlyMaster { implicit session =>
+          val libs = libraryInviteRepo.getActiveWithUserId(owner.id.get, page) map libraryRepo.get //cached
+          createLibraryCardInfos(libs, ownerBasicUser, idealSize, viewer.isDefined)
+        }
+      case Some(other) =>
+        Seq.empty
+    }
+  }
+
   private def createLibraryCardInfos(libs: Seq[Library], ownerBasicUser: BasicUser, idealSize: ImageSize, isAuthenticatedRequest: Boolean)(implicit session: RSession): Seq[LibraryCardInfo] = {
```
2is10, line 1309: In the general case, each library may have a different owner. Maybe better to pass a `Seq[(Library, User)]` or `Seq[(Library, BasicUser)]`?

eishay, line 1309: correct, @ahsu1230 pointed it out as well, there's a pull request that fixes and tests it

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
eishay, line 1295: True, there's a name change in the next pull request.
The reason I kept it this way is to enforce ACL on this layer so we'll be consisted with other getLibraries methods in this class used for profile.
Also, in the future the inviter may be able to see invitations they made.

PR #12879, opened 2014-12-23
randomize cache ttl by default

ymatsuda:
@eishay @leogrim @yingjieMiao @stkem @andrewconner 

The default `minTTL` is set to `maxTTL * 0.9`


kifi-backend/modules/common/app/com/keepit/common/cache/FortyTwoCache.scala
```diff
@@ -187,6 +174,10 @@ class FortyTwoCacheImpl[K <: Key[T], T](
     val serializer: Serializer[T],
     override val outerCache: Option[ObjectCache[K, T]]) extends FortyTwoCache[K, T] {
 
+  assert(minTTL.toMillis <= (30 days).toMillis, "minTTL too long")
```
eishay, line 177: great!

eishay:
nice!

PR #12882, opened 2014-12-23
testing invitation getters and add test factory

eishay:


eishay:
@ahsu1230

ahsu1230:
oooo thanks for this! :dog:

PR #12885, opened 2014-12-23
speed optimizations and add parallel support to profile mobile endpoint

eishay:


eishay:
@ahsu1230 @2is10

2is10:
very nice

PR #12891, opened 2014-12-23
landing page v3 is alive!

eishay:


eishay:
@joshm1 @stkem @andrewconner

PR #12892, opened 2014-12-23
show library on profile (backend support)

ahsu1230:
for createLibrary & modifyLibrary (website)
the response looks a tad different, but the response body is not being used on the FE website. 
The response looks like this:

```
{
  "library" : {...},
  "listed" : true
}
```

When we need to include more information about memberships, we can include `listed` as part of the `membership` object. But, for now, since we only have listed. I was thinking we'll just keep it like that =)

@eishay @2is10

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/model/LibraryMembership.scala
+++ b/kifi-backend/modules/common/app/com/keepit/model/LibraryMembership.scala
@@ -22,7 +22,7 @@ case class LibraryMembership(
     state: State[LibraryMembership] = LibraryMembershipStates.ACTIVE,
     seq: SequenceNumber[LibraryMembership] = SequenceNumber.ZERO,
     showInSearch: Boolean = true,
-    listed: Boolean = true, // does library appear on user's profile?
+    listed: Boolean = true, // whether library appears on user's profile
     lastViewed: Option[DateTime] = None,
     lastEmailSent: Option[DateTime] = None) extends ModelWithState[LibraryMembership] with ModelWithSeqNumber[LibraryMembership] {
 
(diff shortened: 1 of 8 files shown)


kifi-backend/modules/common/app/com/keepit/model/LibraryMembership.scala
```diff
@@ -22,7 +22,7 @@ case class LibraryMembership(
     state: State[LibraryMembership] = LibraryMembershipStates.ACTIVE,
     seq: SequenceNumber[LibraryMembership] = SequenceNumber.ZERO,
     showInSearch: Boolean = true,
-    listed: Boolean = true, // does library appear on user's profile?
+    listed: Boolean = true, // whether library appears on user's profile
```
2is10, line 25: “may appear” would be technically more accurate than “appears” since there are other factors at play (e.g. who the viewer is, what the library’s visibility is)

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -534,6 +537,7 @@ class LibraryCommander @Inject() (
             libraryAliasRepo.reclaim(ownerId, newSlug)
             libraryAliasRepo.alias(ownerId, targetLib.slug, targetLib.id.get)
           }
+          libraryMembershipRepo.save(targetMembership.copy(listed = newListed))
           libraryRepo.save(targetLib.copy(name = newName, slug = newSlug, visibility = newVisibility, description = newDescription, color = newColor, state = LibraryStates.ACTIVE))
```
2is10, line 541: We generally try to avoid writing rows unnecessarily. If only `listed` is changing, we shouldn’t update the `library` row. If only something else is changing, we shouldn’t update the `library_membership` row.

Figuring out what’s changing is sometimes cumbersome. This is one advantage of having separate endpoints for different kinds of changes.

PR #12893, opened 2014-12-23
making the user profile library tabs work (library filtering)

2is10:
...and adding a very simple empty state.

@yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/libraryService.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/libraryService.js
@@ -128,10 +128,6 @@ angular.module('kifi')
       invitedSummaries: invitedSummaries,
       recentLibraries: recentLibraries,
 
-      isAllowed: function () {
-        return profileService.me.experiments && profileService.me.experiments.indexOf('libraries') !== -1;
-      },
-
       isSystemLibrary: function (libraryId) {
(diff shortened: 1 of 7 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/libraryService.js
```diff
@@ -128,10 +128,6 @@ angular.module('kifi')
       invitedSummaries: invitedSummaries,
       recentLibraries: recentLibraries,
 
-      isAllowed: function () {
-        return profileService.me.experiments && profileService.me.experiments.indexOf('libraries') !== -1;
-      },
-
```
2is10, line 134: dead code

kifi-backend/modules/shoebox/angular/src/route.js
```diff
@@ -98,7 +96,6 @@ angular.module('kifi')
           .state('userProfile.libraries.invited', {
             url: '/libraries/invited',
             templateUrl: 'userProfile/userProfileLibrariesList.tpl.html',
-            controller: 'UserProfileLibrariesListCtrl',
```
2is10, line 101: since `UserProfileLibrariesCtrl` has so little UI (the library filter tabs), I thought I’d give it the responsibility of managing the library list too

yipingliao:
Nice! :+1:

PR #12895, opened 2014-12-23
splitting library and user sitemaps, giving them a numerical (starts wit...

eishay:
...h 0) for future compatibility;

making sure sitemap is cached before calling the search engine


eishay:
@raymondng

PR #12896, opened 2014-12-23
showing followers or when last updated at bottom of library cards

2is10:
@yipingliao


kifi-backend/modules/shoebox/angular/src/userProfile/userProfileLibrariesList.tpl.html
```diff
@@ -14,10 +14,20 @@
...
-  <div class="kf-upl-card-foot"></div>
+  <div class="kf-upl-card-foot">
+    <div class="kf-upl-followers" ng-if="lib.followers.length > 0">
+      <span class="kf-upl-follower" ng-repeat="user in lib.followers" ng-attr-style="background-image:url({{user.picUrl}})"></span>
+      <span class="kf-upl-followers-text"><!-- ng-if="lib.numFollowers > lib.followers.length" -->
+        + <ng-pluralize count="lib.numFollowers - lib.followers.length" when="{'1':'{} other follows ','other':'{} others follow '}"></ng-pluralize> this library
+      </span>
+    </div>
+    <div class="kf-upl-updated" ng-if="lib.followers.length === 0">
+      Updated N days ago
```
2is10, line 28: shoebox doesn’t yet sent when library was last updated

yipingliao:
Thanks again!

PR #12897, opened 2014-12-24
increased ConnectedUserScoreCache to 6-12 hours

eishay:
the problems with full gc on graph are still causing the graph to do a full pause and get into unavailable state. the increase in cache time is tmp solution until we'll figure out this problem.
Yasu reports that he sees an increase of heap size from 8gb to 12 gb over one minute just before the gc stops the vm.


eishay:
@leogrim @ymatsuda @stkem @yingjieMiao

PR #12906, opened 2014-12-24
add listed to get library (which uses FullLibraryInfo) website endpoint

ahsu1230:
The get response endpoints look like this:

```
{
 "library" : {...},
 "membership" : "none",           // "read_only", "owner", etc.
 "listed" : false
}
```

I was originally going to place `listed` into a membership object to look something like this

```
"membership" : {
   "access" : "read_only",
   "listed" : false
}
```

and if there's no membership, it doesn't show up in the JSON. However, changing it to an object could break some website stuff. So I figured I should make a change like that later when I'm sure it won't break the FE website

@2is10 @eishay

### DIFF
--- a/kifi-backend/modules/shoebox/conf/shoeboxService.routes
+++ b/kifi-backend/modules/shoebox/conf/shoeboxService.routes
@@ -207,7 +207,7 @@ GET     /m/1/contacts/search                    @com.keepit.controllers.mobile.M
 POST    /m/1/invite                             @com.keepit.controllers.mobile.MobileInviteController.inviteConnection
 
 # Profile Page
-GET     /m/1//user/:username/libraries          @com.keepit.controllers.mobile.MobileLibraryController.getProfileLibraries(username: Username, page: Int ?= 0, size: Int ?= 12, filter: String ?= "own")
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/shoebox/conf/shoeboxService.routes
```diff
@@ -207,7 +207,7 @@ GET     /m/1/contacts/search                    @com.keepit.controllers.mobile.M
 POST    /m/1/invite                             @com.keepit.controllers.mobile.MobileInviteController.inviteConnection
 
 # Profile Page
-GET     /m/1//user/:username/libraries          @com.keepit.controllers.mobile.MobileLibraryController.getProfileLibraries(username: Username, page: Int ?= 0, size: Int ?= 12, filter: String ?= "own")
+GET     /m/1/user/:username/libraries          @com.keepit.controllers.mobile.MobileLibraryController.getProfileLibraries(username: Username, page: Int ?= 0, size: Int ?= 12, filter: String ?= "own")
```
2is10: please keep `@` aligned with others above and below. thanks.

2is10:
:+1:

PR #12909, opened 2014-12-29
fixing long escaping in redirects

eishay:


eishay:
@leogrim @raymondng

eishay:
for context:

``` sql

mysql> select count(*) from normalized_uri where url like "%25252525%"\G
*************************** 1. row ***************************
count(*): 2748
1 row in set (5.55 sec)
```

eishay:
``` sql

mysql> select count(*) from normalized_uri where url like "%2525%"\G
*************************** 1. row ***************************
count(*): 4393
1 row in set (5.54 sec)
```

eishay:
accidentally added a fix for another issue for the same pull request

PR #12910, opened 2014-12-29
not trying to get images from domains we can't deal with

eishay:
and avoiding a bad "return" in the middle of a function

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/common/images/ImageFetcher.scala
+++ b/kifi-backend/modules/common/app/com/keepit/common/images/ImageFetcher.scala
@@ -43,54 +43,62 @@ class ImageFetcherImpl @Inject() (
   private def getBufferedImage(is: InputStream) = Try { imageUtils.forceRGB(ImageIO.read(is)) }
 
   override def fetchRawImage(url: URI): Future[Option[BufferedImage]] = {
-    val trace = new StackTrace()
-    val timer = accessLog.timer(Access.HTTP_OUT)
-    val uriObj = URI.parse(url.toString()) match {
-      case Success(goodUri) => goodUri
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/common/app/com/keepit/common/images/ImageFetcher.scala
```diff
@@ -43,54 +43,62 @@ class ImageFetcherImpl @Inject() (
...
 
   override def fetchRawImage(url: URI): Future[Option[BufferedImage]] = {
-    val trace = new StackTrace()
-    val timer = accessLog.timer(Access.HTTP_OUT)
-    val uriObj = URI.parse(url.toString()) match {
-      case Success(goodUri) => goodUri
+    val validUri = URI.parse(url.toString()) match {
+      case Success(goodUri) =>
+        if (goodUri.host.exists(_.domain.exists(_.contains("_")))) {
+          log.error(s"Url [$url] has a hostname which contains an underscore, our current http client would not like that!")
```
andrewconner, line 49: @eishay Is this true? Seeing many images on hosts such as http://bwcom_blog.s3.amazonaws.com/wp-content/uploads/2014/12/caravan_logo.png.

eishay, line 49: for future note:

``` scala
scala> new java.net.URI("http://bwcom_blog.s3.amazonaws.com/wp-content/uploads/2014/12/caravan_logo.png").getHost
res10: String = null
```

andrewconner, line 49: Thanks for investigating!

PR #12913, opened 2014-12-29
avoid autoboxing of Long in ProbabilityDensity

ymatsuda:
@eishay @leogrim @yingjieMiao @stkem 

It seems that the autoboxing of Long in ProbabilityDensity creates a huge number of objects and consuming Graph's heap rapidly (analyzed using yourkit). This could be a cause of full GCs.


ymatsuda:
I meant Double, not Long. Sorry.

stkem:
Nice find! Also, check out http://scala-miniboxing.org/index.html which might help with stuff like that if we find more occurrences.

raymondng:
Nice :+1:

leogrim:
Cool! @stkem even had a task to try Miniboxing out at some point, among other ideas: https://app.asana.com/0/11645953228207/14934497049443

PR #12914, opened 2014-12-29
Fix broken infinite scroll on search.

yipingliao:
@leogrim @andrewconner 

Still investigating why I stuck this line in to begin with.


leogrim:
thanks!

PR #12915, opened 2014-12-29
some extra checks on valid uris and ignoring invalid uri from the db

eishay:


eishay:
updated old db rows to reflect the code

PR #12918, opened 2014-12-29
when a user re-authenticates, update the state of the social user info to "created" instead of "inactive" of "app not auth"

eishay:


eishay:
@raymondng @andrewconner @leogrim

PR #12928, opened 2014-12-30
scraper cleanups/logging

eishay:
trying to figure out why the scraper would not scrape //www.autotrader.com/research/article/car-reviews/217339/2014-honda-odyssey-vs-2014-toyota-sienna-which-is-better.jsp (curl from that machine works). the error message from the server is "status:HTTP/1.1 403 Forbidden"


eishay:
@leogrim @andrewconner @stkem @raymondng 
i'm very interested in using embeddly, they could scrape it with out a problem

PR #12933, opened 2014-12-30
Add empty state to own libraries on user profiles; fix flash of empty state.

yipingliao:
@2is10 

A previous change I made causes a "flash" of empty state on "following" and "invited" libraries; this fixes that.

Also, add empty state language to "own" libraries, when the user is viewing the profile page of a user who has not created any publicly viewable libraries.

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/userProfile/userProfileLibrariesList.tpl.html
+++ b/kifi-backend/modules/shoebox/angular/src/userProfile/userProfileLibrariesList.tpl.html
@@ -31,6 +31,7 @@
     </div>
   </div>
   <div class="kf-upl-empty" ng-if="libraries && libraries.length === 0">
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/angular/src/userProfile/userProfileLibrariesList.tpl.html
```diff
@@ -31,6 +31,7 @@
     </div>
   </div>
   <div class="kf-upl-empty" ng-if="libraries && libraries.length === 0">
+    <span ng-if="libraryType == 'own'">{{viewingOwnProfile ? 'You have' : profile.firstName + ' has'}} not created any publicly viewable libraries.</span>
```
2is10, line 34: I think I originally left this out because (based on a mock) a user will always see his/her own Main and Private libraries on his/her own profile. This is definitely needed for the viewing another’s profile case. Thanks.

2is10:
:+1:

PR #12934, opened 2014-12-30
Do not submit feedback for recommended libraries.

yipingliao:
@andrewconner @stkem @joshm1 cc @eishay 

All the feedback we submit for recommendations are based on the assumption that the recommendation is a keep. AFAIK there is no backend support for feedback for recommended libraries. This pull fixes a bug where we get a JavaScript error when we try to submit feedback for a recommended library by first checking that the recommendation is a keep.

Feedback for recommended libraries perhaps coming in the future? :)

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/recos/recos.js
+++ b/kifi-backend/modules/shoebox/angular/src/recos/recos.js
@@ -42,7 +42,9 @@ angular.module('kifi')
     };
 
     $scope.trash = function (reco) {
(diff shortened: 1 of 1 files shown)


joshm1:
@yipingliao @andrewconner @stkem @eishay Most of the backend is already in place for recommended library feedback.  :)

Unfortunately I didn't get to the frontend. :(

yipingliao:
Oh nice!

PR #12936, opened 2014-12-31
Implement front-end for per-library listing visibility.

yipingliao:
@ahsu1230 

Aaron, this is the front-end work that I've done to allow the user to adjust a library's listing visibility.

A couple of issues:
- Adjusting the `listed` field does not seem to affect the libraries being displayed. That is, after making a library `listed` be `false`, I still see that library on my public user profile page. :(
- Also, when we create/modify a library with a visibility of `secret`, then `listed` should be `false` regardless of the what `listed` value is being sent in the post payload. That is, `secret` should override `listed`. However, I still see `listed` being `true` in the response when I send over a `secret` visibility for the library.

Screenshot of the UI pasted.


ahsu1230:
Thanks! I'll take a look at the profiles endpoint and take a look at where the issue is.
As for the create/modify endpoints - I'll make that quick change!

PR #12937, opened 2014-12-31
adding og:see_also to public meta tags using the library recommendations

eishay:


eishay:
@joshm1

joshm1:
lgtm!

kifi-backend/modules/shoebox/app/com/keepit/commanders/PageMetaTagsCommander.scala
```diff
@@ -0,0 +1,112 @@
+package com.keepit.commanders
```
andrewconner, line 1: Nice, thanks for splitting it out.

andrewconner:
Looks good.

PR #12938, opened 2014-12-31
Style the gear and connect icons in the user profile page header so that...

yipingliao:
... they don't move around on width resizing.

@2is10


2is10:
:+1:

PR #12939, opened 2014-12-31
Upl links update

yipingliao:
@2is10 cc @eishay 
- Add microdata to owner pictures in user profile library cards
- Add links to follower pictures in user profile library cards (no microdata for these yet since followers are not shown for logged-out users; question noted for product - we can add microdata when we decide to show followers)

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
+++ b/kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
@@ -135,8 +135,8 @@ angular.module('kifi')
 
 
 .controller('UserProfileLibrariesCtrl', [
-  '$scope', '$rootScope', '$state', '$stateParams', 'routeService', 'keepWhoService', 'profileService', 'userProfileActionService',
-  function ($scope, $rootScope, $state, $stateParams, routeService, keepWhoService, profileService, userProfileActionService) {
+  '$scope', '$rootScope', '$state', '$stateParams', 'routeService', 'keepWhoService', 'profileService', 'userProfileActionService', 'userService',
+  function ($scope, $rootScope, $state, $stateParams, routeService, keepWhoService, profileService, userProfileActionService, userService) {
     var colors = ['#C764A2', '#E35957', '#FF9430', '#2EC89A', '#3975BF', '#955CB4', '#FAB200'];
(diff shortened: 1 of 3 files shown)


eishay:
Great!
Do we have microdata link refs on user faces on library card?

kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
```diff
@@ -179,6 +179,10 @@ angular.module('kifi')
       return hasMoreLibraries;
     };
 
+    $scope.getUserProfileUrl = function (user) {
+      return userService.getProfileUrl(user.username);
```
2is10, line 183: What would you think of having `routeService` be responsible for all formatting of our URLs?

kifi-backend/modules/shoebox/angular/src/userProfile/userProfileLibrariesList.tpl.html
```diff
@@ -18,10 +18,15 @@
       </div>
       <div class="kf-upl-card-byline"><span class="kf-upl-card-by">by</span> {{lib.owner.firstName}} {{lib.owner.lastName}}</div>
     </a>
-    <a class="kf-upl-card-owner-pic" href="/{{lib.owner.username}}" ng-style="{'background-image': 'url(' + lib.ownerPicUrl + ')'}"></a>
+    <div itemscope itemtype="http://data-vocabulary.org/Person">
+      <a href="/{{lib.owner.username}}" itemprop="url">
```
2is10, line 22: should this be `getUserProfileUrl(lib.owner)`?

2is10:
:+1:

kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
yipingliao, line 183: I like it. I will refactor.

kifi-backend/modules/shoebox/angular/src/userProfile/userProfileLibrariesList.tpl.html
yipingliao, line 22: Yes! Will fix.

PR #12940, opened 2014-12-31
non-linear scoring for library recommendations

ymatsuda:
@eishay @joshm1 @stkem @yingjieMiao 

Things done
- added non-linear scoring for library recommendations.
- added incrementDeliveredCount to LibraryRecommendationRepo and actually increment the counts when recommendations are delivered.

Things missing
- diversification is not in place because topic information is not available for each library reco, but I added the default selection strategy so that we can added the diversified selection strategy easily.
- analytics is missing
- feedback for library recos has not endpoint in shoebox.


ymatsuda:
The library non-linear scoring is under the same experiment (curator_nonlinear_scoring).

yingjieMiao:
lgtm

joshm1:
nice. :+1:

eishay:
Cool!

PR #12941, opened 2014-12-31
Preliminarily style mobile for user profile pages.

yipingliao:
@2is10 This will need review and discussion with Mark, but here is a preliminary styling of the user profile library cards on mobile.
screen shot 2014-12-31 at 2 41 10 pm


2is10:
:+1: thanks

PR #12942, opened 2015-01-02
secret library visibility overrides listed setting

ahsu1230:
If a library's visibility is `SECRET`, it should override the `listed` setting (become false) when stored in the database.

@andrewconner
cc: @yipingliao


yipingliao:
Awesome. Thanks!

2is10:
@ahsu1230 @yipingliao  I’m not sure this pull request is a good idea. We want `listed` to be `false` only if the user specifically de-lists a published library. If a library starts out as private and is later published, it should be listed (show up on the user’s profile) automatically. Won’t this pull request prevent that?

2is10:
Better, I think, would be to store private libraries with `listed` being `true` (by default) and just not serve them on profile views when the viewer doesn’t have access.

yipingliao:
Agreed. I think that the intention was that "if a library starts out as private and is later published", that it _should_ show up listed automatically. I didn't read the description about what is being stored in the db carefully - sorry.

PR #12943, opened 2015-01-02
not updating updated_at when incrementing delivered counts

ymatsuda:
@stkem @yingjieMiao @joshm1 

I think we shouldn't update the `updated_at` columns when incrementing the `delivered` column, otherwise the clean up process are not able to pick up delivered recommendations.


ymatsuda:
not good enough. closing for now.

PR #12944, opened 2015-01-02
profiles endpoint (for self) return whether libraries are listed

ahsu1230:
or hidden on public profile page, so UI can differentiate which libraries are listed and which aren't.

@andrewconner @2is10


andrewconner:
Thanks

yipingliao:
Nice. Thanks!

2is10:
:+1:

PR #12945, opened 2015-01-02
hiding recos shown 10 times or more

ymatsuda:
@stkem @yingjieMiao @joshm1 

Could make the threshold a parameter, but hardcoding right now.


joshm1:
:+1:

PR #12946, opened 2015-01-02
trying to fix unstable LibraryRecommendationGenerationCommanderTest

ymatsuda:
@joshm1 

This test intermittently fails. I think it is caused by the deferred sequence number assignment. Updates are ordered by millisecond resolution when the sequence numbers are assigned. Millisecond is not fine enough in the test situation, so the order or sequence number assignment is non-deterministic.


joshm1:
thanks!

PR #12947, opened 2015-01-03
settings for profile library cards

ahsu1230:
- setting icon on top-left corner of all library cards (that you own)
- setting icon does not show up for Main or Secret Libraries
- "Hidden on Public Profile" message shows if `listed` is false
- Clicking on setting icon opens up the ModifyLibrary Modal so you can change the name, listed, and even slug.

@andrewconner @yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
+++ b/kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
@@ -4,9 +4,9 @@ angular.module('kifi')
 
 .controller('UserProfileCtrl', [
   '$scope', '$rootScope', '$state', '$stateParams', '$timeout', '$window',
-  'env', 'inviteService', 'keepWhoService', 'profileService', 'userProfileActionService',
+  'env', 'inviteService', 'keepWhoService', 'profileService', 'userProfileActionService', 'modalService', 'libraryService',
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
```diff
@@ -127,6 +127,30 @@ angular.module('kifi')
       });
     };
 
+    $scope.isMyLibrary = function(libraryOwnerId) {
+      return libraryOwnerId === $scope.profile.id;
+    };
+
+    $scope.isHidden = function(library) {
+      return library.visibility === 'published' && library.listed === false;
```
ahsu1230, line 135: now that I think about it. It's a bit weird when "Hidden on Public Profile" doesn't show up when a library's visibility is `secret`. Maybe consider having a message also show up when a library is "secret" so users know that secret libraries aren't showing up on a public profile?

yipingliao:
Thanks!

kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
yipingliao, line 135: Agreed. More questions for product.

PR #12951, opened 2015-01-05
Add title to profile pages (and fix a couple of things per code review)

yipingliao:
@2is10 
- Add title to user profile pages
- Fix hardcoded route to profile page on owner image in profile library card
- Refactor `userService.getProfileUrl` to be `routeService.getProfileUrl`

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
+++ b/kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
@@ -164,8 +164,8 @@ angular.module('kifi')
 
 
 .controller('UserProfileLibrariesCtrl', [
-  '$scope', '$rootScope', '$state', '$stateParams', 'routeService', 'keepWhoService', 'profileService', 'userProfileActionService', 'userService',
-  function ($scope, $rootScope, $state, $stateParams, routeService, keepWhoService, profileService, userProfileActionService, userService) {
+  '$scope', '$rootScope', '$state', '$stateParams', 'routeService', 'keepWhoService', 'profileService', 'userProfileActionService',
+  function ($scope, $rootScope, $state, $stateParams, routeService, keepWhoService, profileService, userProfileActionService) {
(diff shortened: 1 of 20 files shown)


kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
```diff
@@ -64,12 +64,17 @@ angular.module('kifi')
...
       userProfileActionService.getProfile(username).then(function (profile) {
+        setTitle(profile);
         initProfile(profile);
         initViewingUserStatus();
         initUserNavLinks();
       });
     }
 
+    function setTitle(profile) {
+      $window.document.title = profile.firstName + ' ' + profile.lastName + ' • Kifi' ;
```
yipingliao, line 75: This is where the title is set. It's kinda deeply buried among all the refactoring name changes.

2is10, line 75: I’d prefer that we configure page titles where we configure our routes, so it’s easy to see review them all in one place, to help ensure consistency, and to help remind people to set a title when adding a new route.

Here are some ways to do it with ui-router that a quick search turned up:
http://stackoverflow.com/questions/23813599/set-page-title-using-ui-router
https://coderwall.com/p/zk7qww/dynamic-site-page-titles-with-angular-and-ui-router

2is10:
:+1: Thanks!

kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
yipingliao, line 75: Hmm... I spent some time on this and haven't been able to find a more elegant way - reason being the title we want to set depends on the user's first and last names, which needs a backend call and is not yet available at the time that the routing is done.

2is10, line 75: Thanks for trying. I’ll think about this along with delayed (promised-based) navigation over the next few days.

PR #12953, opened 2015-01-05
profile cards: replaced 'updated N days ago' with real times

ahsu1230:
if there are no followers for a library, the display message can be:
updated just now
updated minute(s) ago
updated hour(s) ago
updated day(s) ago

and document title change when you visit a public profile

@andrewconner @yipingliao


yipingliao:
Looks good, but you may want to check out `angularMoment`. We use this for time-ago strings, like in the keep cards. See: https://github.com/kifi/keepit/blob/master/kifi-backend/modules/shoebox/angular/src/keep/keepCard.tpl.html#L20

ahsu1230:
ah woops, thanks @yipingliao for the heads up!

ahsu1230:
thanks @yipingliao `angularMoment` works better lol

PR #12955, opened 2015-01-06
Mobile pic binary upload endpoint

andrewconner:
200 response means you're good. It included a URL if you need it.

500 means resize failed (usually not much you can do to rectify)

15MB max filesize.


eishay:
that was fast :-)

mrbrown14:
confirmed it works!

PR #12956, opened 2015-01-06
profile libraries: fix followers wording & settings gear only show up on hover

ahsu1230:
@yipingliao @andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/userProfile/userProfileLibrariesList.tpl.html
+++ b/kifi-backend/modules/shoebox/angular/src/userProfile/userProfileLibrariesList.tpl.html
@@ -31,7 +31,7 @@
         <a href="{{getUserProfileUrl(user)}}" class="kf-upl-follower" ng-repeat="user in lib.followers" ng-attr-style="background-image:url({{user.picUrl}})"></a>
         <span ng-if="lib.followers.length==1"> follows this library</span>
         <span ng-if="lib.followers.length > 1">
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/angular/src/userProfile/userProfileLibrariesList.tpl.html
```diff
@@ -29,7 +29,10 @@
     <div class="kf-upl-card-foot">
       <div class="kf-upl-followers" ng-if="lib.followers.length > 0">
         <a href="{{getUserProfileUrl(user)}}" class="kf-upl-follower" ng-repeat="user in lib.followers" ng-attr-style="background-image:url({{user.picUrl}})"></a>
-        <ng-pluralize count="lib.numFollowers - lib.followers.length" when="{'1':'+{} other ','other':'+{} others '}"></ng-pluralize> follow this library
+        <span ng-if="lib.followers.length==1"> follows this library</span>
```
yipingliao, line 32: nit: use `===`.

yipingliao:
:+1:

kifi-backend/modules/shoebox/angular/src/userProfile/userProfileLibrariesList.tpl.html
2is10, line 32: @ahsu1230 @yipingliao  It’s possible for `lib.followers.length === 1` and `lib.numFollowers > 1` because the server filters out followers that have  no custom picture. I’ll take another stab at this.

PR #12957, opened 2015-01-06
3.3.31: fixing issue where backspace can delete wrong recipient

2is10:
cc @andrewconner


andrewconner:
Whoa, bug fixing by removing code.

PR #12959, opened 2015-01-06
front/backend support for library feedback

joshm1:
@stkem


kifi-backend/modules/curator/app/com/keepit/curator/model/LibraryRecommendationRepo.scala
```diff
@@ -120,14 +121,17 @@ class LibraryRecommendationRepoImpl @Inject() (
       if (feedback.clicked.exists(true ==))
         sql"UPDATE library_recommendation SET clicked=clicked+1, updated_at=$currentDateTime WHERE user_id=$userId AND library_id=$libraryId".asUpdate.first > 0
       else true
-    val trashedResult =
-      if (feedback.trashed.exists(true ==)) rowz.map(row => (row.trashed, row.updatedAt)).update((feedback.trashed.get, currentDateTime)) > 0
-      else true
-    val followedResult =
-      if (feedback.followed.exists(true ==)) rowz.map(row => (row.followed, row.updatedAt)).update((feedback.followed.get, currentDateTime)) > 0
-      else true
```
joshm1, line 128: even though we don't allow the client to set the trashed/followed field from true to false, I don't see a reason to add this constraint at this low of a level

PR #12960, opened 2015-01-06
fix linking on lefthand nav bar

ahsu1230:
The bug was introduced when we had nested `<a></a>`. The inner hyperlink <a> was for clicking on profile pictures to get to profiles.

@yipingliao @andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/layout/nav/nav.tpl.html
+++ b/kifi-backend/modules/shoebox/angular/src/layout/nav/nav.tpl.html
@@ -40,6 +40,7 @@
                              'one': '1 keep',
                              'other': '{} keeps'}">
               </ng-pluralize>
+            </div>
           </div>
         </a>
       </li>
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/shoebox/angular/src/layout/nav/nav.tpl.html
```diff
@@ -89,34 +90,32 @@
         </li>
 
         <li ng-show="sortingMenu.myLibsFirst" ng-repeat="library in myLibsToShow" class="kf-nav-lib-item" ng-class="{ active: isActive(library.url) }">
-          <div class="kf-nav-link-box">
-            <a class="kf-nav-link" ng-href="{{library.url}}">
-              <div class="library-icon-holder">
-                <div class="library-icon svg-private-library-icon" ng-if="library.visibility=='secret'"></div>
-                <div class="library-icon svg-public-library-icon" ng-if="library.visibility=='published'"></div>
+          <div class="kf-nav-link-box" ng-click="redirectTo(library.url)">
```
ahsu1230, line 93: to fix, I placed an ng-click on the enveloping element to link to the library url instead of the outer `<a>`

yipingliao:
Oops, I was the one who broke it! I think that we need both of the `<a>` tags. We can make the profile image be a sibling element that is positioned absolutely. See: https://github.com/kifi/keepit/blob/master/kifi-backend/modules/shoebox/angular/src/userProfile/userProfileLibrariesList.tpl.html#L24

ahsu1230:
is that better than using a ng-click on the parent element? I guess then you don't need to have a scope function to redirect

2is10:
> is that better than using a ng-click on the parent element?

Yes, we should always use `<a href="url">...</a>` for navigation links for usability and accessibility. For example, users should be able to hover over a link to see where it will go (in the browser status bar) and right click to copy the URL or open it in a different tab.

PR #12964, opened 2015-01-06
library reco action tracking

joshm1:
@stkem @yingjieMiao

### DIFF
--- a/kifi-backend/modules/curator/app/com/keepit/curator/commanders/LibraryRecommendationGenerationCommander.scala
+++ b/kifi-backend/modules/curator/app/com/keepit/curator/commanders/LibraryRecommendationGenerationCommander.scala
@@ -47,13 +47,10 @@ class LibraryRecommendationGenerationCommander @Inject() (
       recoSortStrategy.sort(recosByTopScore) take howManyMax
     }
 
-    SafeFuture {
+    if (source != RecommendationSource.Admin) SafeFuture {
       db.readWrite { implicit session =>
-        recos.map { recoScore =>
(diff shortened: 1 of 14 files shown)


kifi-backend/modules/curator/app/com/keepit/curator/commanders/LibraryRecommendationGenerationCommander.scala
```diff
@@ -36,8 +37,8 @@ class LibraryRecommendationGenerationCommander @Inject() (
     experimentCommander.getUsersByExperiment(ExperimentType.CURATOR_LIBRARY_RECOS).
       map(users => users.map(_.id.get).toSeq)
 
-  def getTopRecommendations(userId: Id[User], howManyMax: Int, recoSortStrategy: LibraryRecoSelectionStrategy, scoringStrategy: LibraryRecoScoringStrategy): Seq[LibraryRecoInfo] = {
-
+  def getTopRecommendations(userId: Id[User], howManyMax: Int, source: RecommendationSource, subSource: RecommendationSubSource,
+    recoSortStrategy: LibraryRecoSelectionStrategy, scoringStrategy: LibraryRecoScoringStrategy): Seq[LibraryRecoInfo] = {
```
joshm1, line 41: I'm not particularly happy with this implementation of passing source/subSource all the way from the shoebox controller to a curator commander, which clutters an reasonably concise and clean API.  I followed the pattern used for URI recommendations, but I think there's room for improvement with a clean abstraction for creating the reco action events.

PR #12965, opened 2015-01-06
creating component iterator lazily

ymatsuda:
@leogrim @yingjieMiao 

A component iterator of IncomingEdgeReader is not used in wandering. We can save memory by opening the iterator lazily.


yingjieMiao:
:+1:  great observation!

leogrim:
All edges in the graph are currently undirected, or more exactly, double-directed: each outgoing edge from source to destination has a corresponding outgoing edge from destination to source. This is why we can correctly traverse the graph relying only on outgoing edges. 

Now we could save even more memory (~1/2 the current number of edges) by having single directed edges and leveraging 'IncomingEdgeReader' in traversals. I have added the 'IncomingEdgeReader' feature later on, together with the ability to delete vertices, but I haven't remodeled the graph and updated the wandering logic. If memory really is an issue these days, that's probably a nice easy win.

ymatsuda:
Oh, that is why it is not used. The lazy creation of iterator seems orthogonal to what you said, Leo. I will merge this anyway.

PR #12969, opened 2015-01-06
making "isFriend" correctly handle unfriended connections

2is10:
@ahsu1230


kifi-backend/modules/shoebox/app/com/keepit/commanders/UserCommander.scala
```diff
@@ -109,15 +109,16 @@ class UserCommander @Inject() (
 
   def profile(username: Username, viewer: Option[User]): Option[UserProfile] = {
     userFromUsername(username) map { user =>
+      val isConnected: Option[Boolean] = viewer.filter(_.id != user.id).map { viewer =>
+        db.readOnlyMaster { implicit session =>
+          userConnectionRepo.areConnected(viewer.id.get, user.id.get)
+        }
+      }
```
2is10, line 116: using `readOnlyMaster` because the friendship status is actually quite important to have up-to-date

ahsu1230:
:+1:

PR #12971, opened 2015-01-06
Make search input box more responsive.

yipingliao:
@andrewconner @2is10 
@eishay pointed out that our search input text box is not very responsive - when you type into it at a moderately fast pace, the new letters take some delay to show up. This pull is an attempt to fix that. Please let me know what you think. Thanks!

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/search/search.js
+++ b/kifi-backend/modules/shoebox/angular/src/search/search.js
@@ -122,7 +122,6 @@ angular.module('kifi')
         return;
       }
 
-
       $scope.loading = true;
       searchActionService.find(query, filter, library, lastResult && lastResult.context, $rootScope.userLoggedIn).then(function (q, result) {
         if (q !== query) {  // query was updated
@@ -136,40 +135,55 @@ angular.module('kifi')
           $scope.resultTotals.othersTotal = 0;
         }
 
+        function processHit() {
+          if (q !== query) {
+            return;
+          }
+
+          if (hitIndex >= hits.length) {
+            processHitDeferred.resolve();
+            return;
+          }
+
+          var hit = hits[hitIndex];
+          var searchKeep = new keepDecoratorService.Keep(hit);
+          if (!!searchKeep.id) {
+            searchKeep.buildKeep(searchKeep);
+          }
+
+          // TODO remove after we get rid of the deprecated code and update new code to use 'tags' instead of 'hashtags'
+          searchKeep.hashtags = searchKeep.tags;
+          $scope.resultKeeps.push(searchKeep);
+
+          hitIndex++;
+          $timeout(processHit);
+        }
+
         var hits = result.hits;
-        hits.forEach(function (hit) {
-          // Break up this update-query-results loop so that the event loop has
-          // a chance to pick up keystrokes from the search input.
+        var hitIndex = 0;
+
+        // Process one hit per event loop turn to allow other events to come through.
+        $timeout(processHit);
+
+        var processHitDeferred = $q.defer();
+        var processHitPromise = processHitDeferred.promise;
+
+        // Wait for results to be rendered before displaying totals.
+        processHitPromise.then(function () {
+          // Another segment to be executed on the next turn of the event loop for keystrokes to come through.
           $timeout(function () {
             if (q !== query) { // query was updated
               return;
             }
 
-            var searchKeep = new keepDecoratorService.Keep(hit);
-            if (!!searchKeep.id) {
-              searchKeep.buildKeep(searchKeep);
-            }
+            $scope.resultTotals.myTotal = $scope.resultTotals.myTotal || result.myTotal;
+            $scope.resultTotals.friendsTotal = $scope.resultTotals.friendsTotal || result.friendsTotal;
+            $scope.resultTotals.othersTotal = $scope.resultTotals.othersTotal || result.othersTotal;
 
-            // TODO remove after we get rid of the deprecated code and update new code to use 'tags' instead of 'hashtags'
-            searchKeep.hashtags = searchKeep.tags;
-
-            $scope.resultKeeps.push(searchKeep);
+            $scope.hasMore = !!result.mayHaveMore;
+            lastResult = result;
+            $scope.loading = false;
           });
-        });
-
-        // Another segment to be executed on the next turn of the event loop for keystrokes to come through.
-        $timeout(function () {
-          if (q !== query) { // query was updated
-            return;
-          }
-
-          $scope.resultTotals.myTotal = $scope.resultTotals.myTotal || result.myTotal;
-          $scope.resultTotals.friendsTotal = $scope.resultTotals.friendsTotal || result.friendsTotal;
-          $scope.resultTotals.othersTotal = $scope.resultTotals.othersTotal || result.othersTotal;
-
-          $scope.hasMore = !!result.mayHaveMore;
-          lastResult = result;
-          $scope.loading = false;
         });
       }.bind(null, query));
     };
--- a/kifi-backend/modules/shoebox/angular/src/search/search.js
+++ b/kifi-backend/modules/shoebox/angular/src/search/search.js
@@ -245,6 +245,9 @@ angular.module('kifi')
       }
 
       newSearchPromise = $timeout(function () {
+        // Use $state.params instead of $stateParams because changes to $stateParams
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/angular/src/search/search.js
```diff
@@ -123,12 +123,11 @@ angular.module('kifi')
       }
 
       $scope.loading = true;
-      var searchedQuery = query;
-
       searchActionService.find(query, filter, library, lastResult && lastResult.context, $rootScope.userLoggedIn).then(function (result) {
-        if (searchedQuery !== query) { // query was updated
+        if (result.query !== query) { // query was updated
```
yipingliao: I don't think the previous test worked because `searchedQuery` will always equal to `query` since `query` is not saved in a closure. I am now comparing `result.query`, which is returned as part of the search result and has the query value when that search call was made.

kifi-backend/modules/shoebox/angular/src/search/search.js
```diff
@@ -138,23 +137,38 @@ angular.module('kifi')
 
         var hits = result.hits;
 
-        hits.forEach(function (hit) {
-          var searchKeep = new keepDecoratorService.Keep(hit);
-          if (!!searchKeep.id) {
-            searchKeep.buildKeep(searchKeep);
+        // Break up this update-query-results function into segments so that the event loop has
```
yipingliao: One part of the fix is to break up this long-ish function so that the key strokes in the search input box has a chance to come through.

kifi-backend/modules/shoebox/angular/src/search/search.js
```diff
@@ -221,16 +235,20 @@ angular.module('kifi')
     //
     // Watches and event listeners.
     //
-    var newSearch = _.debounce(function () {
-        // Use $state.params instead of $stateParams because changes to $stateParams
-        // does not propagate to HeaderCtrl when it is injected there.
-        // See: http://stackoverflow.com/questions/23081397/ui-router-stateparams-vs-state-params
+
+    // Wait for a quiescent period of 250ms before firing off a new search.
```
yipingliao: The second part of the fix is to wait for a quiescent period on the text input before firing off a new search. Fewer new searches are now fired compared to the previous debounced function.

kifi-backend/modules/shoebox/angular/src/search/searchActionService.js
```diff
@@ -187,6 +187,7 @@ angular.module('kifi')
           pageSession: pageSession
         };
 
+        resData.query = query;
```
yipingliao: Add the query that invoked the search to the search result for comparison.

2is10: This is fine. Another option might be to bind the query to the callback function where it’s defined (closer to where it’s needed):

``` js
searchActionService.find(query, filter, ...).then(function (q, result) {
  if (query !== q) { // query has changed
    return;
  }
  ...
}.bind(null, query));
```

kifi-backend/modules/shoebox/angular/src/search/search.js
```diff
@@ -221,16 +235,23 @@ angular.module('kifi')
...
         // does not propagate to HeaderCtrl when it is injected there.
         // See: http://stackoverflow.com/questions/23081397/ui-router-stateparams-vs-state-params
         _.assign($state.params, $location.search());
         init();
-      },
-      250,
-      { 'leading': true }
-    );
+      }, 250);
+    };
```
2is10: How is what you’ve done different from simply removing `{'leading': true}`? What you’ve got now looks to me to be equivalent to `_.debounce(..., 250)`.

kifi-backend/modules/shoebox/angular/src/search/searchActionService.js
yipingliao: Will do.

kifi-backend/modules/shoebox/angular/src/search/search.js
2is10: Good find!

kifi-backend/modules/shoebox/angular/src/search/search.js
2is10: As we just discussed, a timeline recording seems to suggest that the rendering of the hits is what needs to be broken up. We may want append only one hit to `.resultKeeps` per turn of the event loop. It might also pay to request fewer than 10 hits in the first batch, since typically only 1 to 3 can fit on a screen at a time (maybe 5 in the pathological case that none have descriptions or images).

yipingliao:
Thanks for all your help, @2is10 ! I am going to merge this first thing tomorrow morning just to be safe since it affects search - looking forward to trying it out in prod.

kifi-backend/modules/shoebox/angular/src/search/search.js
```diff
@@ -136,26 +135,57 @@ angular.module('kifi')
...
+        processHitPromise.then(function () {
+          // Another segment to be executed on the next turn of the event loop for keystrokes to come through.
+          $timeout(function () {
+            if (q !== query) { // query was updated
+              return;
+            }
+
+            $scope.resultTotals.myTotal = $scope.resultTotals.myTotal || result.myTotal;
+            $scope.resultTotals.friendsTotal = $scope.resultTotals.friendsTotal || result.friendsTotal;
+            $scope.resultTotals.othersTotal = $scope.resultTotals.othersTotal || result.othersTotal;
```
2is10: I’d suggest updating these earlier, either when the first hit is rendered or before, since these counts are visible at the top of the page (and aren’t expensive to render).

kifi-backend/modules/shoebox/angular/src/search/search.js
```diff
@@ -136,26 +135,57 @@ angular.module('kifi')
...
+        // Process one hit per event loop turn to allow other events to come through.
+        $timeout(processHit);
+
+        var processHitDeferred = $q.defer();
+        var processHitPromise = processHitDeferred.promise;
+
+        // Wait for results to be rendered before displaying totals.
+        processHitPromise.then(function () {
+          // Another segment to be executed on the next turn of the event loop for keystrokes to come through.
+          $timeout(function () {
```
2is10: I don’t think this `$timeout` is helpful. The way `processHit` is currently structured, `processHitPromise` is already being resolved on a fast turn of the event loop (no hits being rendered, since they have all been rendered already).

kifi-backend/modules/shoebox/angular/src/search/search.js
```diff
@@ -136,26 +135,57 @@ angular.module('kifi')
...
-        $scope.loading = false;
-      });
+        var hits = result.hits;
+        var hitIndex = 0;
+
+        // Process one hit per event loop turn to allow other events to come through.
+        $timeout(processHit);
+
+        var processHitDeferred = $q.defer();
+        var processHitPromise = processHitDeferred.promise;
```
2is10: Naming suggestions:
`processHit` → `processNextHit`
`processHitDeferred` → `processAllHitsDeferred` or just `deferred` (no ambiguity)
`processHitPromise` → doesn’t need to exist (only used once, can be inlined)

In fact, a deferred is most useful when its promise can be returned/exposed as part of an API. Here the deferred has the same visibility as the promise, so it might be clearer not to use deferred/promise at all.

Don’t feel any obligation to act on this comment. I’ll be happy to attempt a minor refactoring after this is merged and ask your opinion of it.

2is10:
Very nicely done! I also can’t wait to try it out.

kifi-backend/modules/shoebox/angular/src/search/search.js
yipingliao: Will do. Thanks!

kifi-backend/modules/shoebox/angular/src/search/search.js
yipingliao: Removing this whole thing from this logic. Original motivation was so that the number of results being displayed, which was dependent on the length of the array of keeps, won't change every time we push into the array. This was a pretty bad way of resolving that. I am making changes to resolve that better.

kifi-backend/modules/shoebox/angular/src/search/search.js
yipingliao: Yes, it was a little convoluted. :) Thanks!

PR #12972, opened 2015-01-06
refactored how library deliveries are tracked and updated

joshm1:
@ymatsuda @yingjieMiao
cc @stkem 

I separated the 2 concerns of fetching library recommendations and analytics for delivering library recommendations.

This is to solve 2 problems:
1) curator doesn't actually know that the library recommendations it returns are delivered to the client
2) the logic required to fetch a library recommendations for a user doesn't care what the source/subSource are. Only Heimdal (mixpanel) care about these values (at this point).


yingjieMiao:
lgtm

PR #12973, opened 2015-01-06
adding "friendRequestSentAt" or "friendRequestReceivedAt" to user profile response JSON

2is10:
@eishay 
cc @ahsu1230 @yipingliao @andrewconner @thassss @mrbrown14 @jaredpetker @DerekBurch 

What do you all think of this format? Using dates instead of booleans so that we have the option of exposing to users how long ago a friend request was sent.


andrewconner:
Works for me

eishay:
like

2is10:
Thanks. It looks like we’ll need this [ `BasicUser` + friend status ] unit elsewhere, so I’ll define a case class for it in a subsequent pull request.

PR #12974, opened 2015-01-06
Add dropdown menu to user profile icon in header.

yipingliao:
@andrewconner @2is10 
Paths to user profile pages are behind experiment.


2is10:
:+1: nice!

PR #12975, opened 2015-01-06
no longer requiring authentication to serve the followers list

2is10:
I’m not sure why this was ever a restriction, but Mark said we could remove it.

@joshm1 
cc @yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/model/LibraryView.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/model/LibraryView.scala
@@ -146,13 +146,8 @@ object LibraryCardInfo {
     JsObject((Json.toJson(o).asInstanceOf[JsObject].value - "owner").toSeq)
   }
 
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/app/com/keepit/model/LibraryView.scala
```diff
@@ -146,13 +146,8 @@ object LibraryCardInfo {
...
 
-  def showable(followers: Seq[BasicUser], isAuthenticatedRequest: Boolean): Seq[BasicUser] = {
-    if (isAuthenticatedRequest) {
-      val goodLooking = followers.filter(_.pictureName != "0.jpg")
-      if (goodLooking.size <= 7) goodLooking else goodLooking.take(3) // only 7 can fit
-    } else {
-      Seq.empty
-    }
+  def showable(followers: Seq[BasicUser]): Seq[BasicUser] = {
+    followers.filter(_.pictureName != "0.jpg").take(3)
```
2is10, line 150: never taking more than 3 because we’re always going to have text explaining that the faces are followers

joshm1:
lgtm

PR #12977, opened 2015-01-07
new mobile endpoints to include library recos

joshm1:
@stkem @andrewconner

### DIFF
--- a/kifi-backend/modules/common/test/com/keepit/model/UserFactory.scala
+++ b/kifi-backend/modules/common/test/com/keepit/model/UserFactory.scala
@@ -14,7 +14,7 @@ object UserFactory {
 
   def users(count: Int): Seq[PartialUser] = List.fill(count)(user())
 
-  class PartialUser private[UserFactory] (user: User) {
+  class PartialUser private[UserFactory] (user: User, _experiments: Seq[ExperimentType] = Seq.empty) {
     def withId(id: Id[User]) = new PartialUser(user.copy(id = Some(id)))
     def withId(id: Int) = new PartialUser(user.copy(id = Some(Id[User](id))))
     def withId(id: ExternalId[User]) = new PartialUser(user.copy(externalId = id))
@@ -24,7 +24,9 @@ object UserFactory {
     def withUsername(name: Username) = new PartialUser(user.copy(username = name))
     def withPictureName(name: String) = new PartialUser(user.copy(pictureName = Some(name)))
     def withState(state: State[User]) = new PartialUser(user.copy(state = state))
+    def withExperiments(experiments: ExperimentType*) = new PartialUser(user, experiments ++ _experiments)
     def get: User = user
+    def experiments: Seq[ExperimentType] = _experiments
   }
 
   implicit class PartialUserSeq(users: Seq[PartialUser]) {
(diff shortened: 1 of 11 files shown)


kifi-backend/modules/shoebox/test/com/keepit/controllers/mobile/MobileRecommendationsControllerTest.scala
```diff
@@ -55,55 +58,42 @@ class MobileRecommendationsControllerTest extends TestKitSupport with Specificat
 
   "MobileRecommendationsController" should {
```
joshm1, line 59: A lot of the changed code in this file is moving code around for reuse between tests for /m/1 and /m/2

kifi-backend/modules/shoebox/test/com/keepit/controllers/mobile/MobileRecommendationsControllerTest.scala
```diff
@@ -55,55 +58,42 @@ class MobileRecommendationsControllerTest extends TestKitSupport with Specificat
...
-        inject[UserActionsHelper].asInstanceOf[FakeUserActionsHelper].setUser(user, Set())
-
-        val route = com.keepit.controllers.mobile.routes.MobileRecommendationsController.topRecos(true, 0.75f).url
-        route === "/m/1/recos/top?more=true&recency=0.75"
-
-        val request = FakeRequest("GET", route)
+    // Most of this block is setup code that is shared between 2 different tests
+    "topRecos" should {
 
-        val controller = inject[RecommendationsController]
```
joshm1, line 71: the existing test wasn't testing the right controller. :)

kifi-backend/modules/common/test/com/keepit/model/UserFactory.scala
```diff
@@ -24,7 +24,9 @@ object UserFactory {
     def withUsername(name: Username) = new PartialUser(user.copy(username = name))
     def withPictureName(name: String) = new PartialUser(user.copy(pictureName = Some(name)))
     def withState(state: State[User]) = new PartialUser(user.copy(state = state))
+    def withExperiments(experiments: ExperimentType*) = new PartialUser(user, experiments ++ _experiments)
     def get: User = user
+    def experiments: Seq[ExperimentType] = _experiments
```
joshm1, line 29: @eishay I can create a UserExperimentFactory instead if you'd prefer single-model isolation with these test factories.  I tried this out because `user().withExperiments(...)` is nice and short

kifi-backend/modules/shoebox/conf/shoeboxService.routes
```diff
@@ -228,7 +228,7 @@ POST    /m/1/attribution/appsflyer              @com.keepit.controllers.mobile.M
 
 POST    /m/1/disconnect/:provider               @com.keepit.controllers.mobile.MobileUserController.disconnect(provider)
 
-GET     /m/1/recos/top                          @com.keepit.controllers.mobile.MobileRecommendationsController.topRecos(more: Boolean, recency: Float)
+GET     /m/$version<[12]>/recos/top             @com.keepit.controllers.mobile.MobileRecommendationsController.topRecos(more: Boolean, recency: Float, version: Int)
```
joshm1: on the other hand I can just as easily 2 hardcoded paths with 2 different controller actions.  I started with this because the uri recommendation part was identical between the 2, but now I'm second guessing that

stkem: I think just having two routes here is the better way to go. Minimally more code and easier to understand. Plus no risk of masking later routes with wildcards and the like (albeit that is kind of minimal here).

andrewconner: Agreed. The idea of versions is to be able to chop off old ones when we can.

joshm1: Updated, thanks!

PR #12978, opened 2015-01-07
turn links on nav bar back to <a> tags

ahsu1230:
Thanks @yipingliao @2is10 for the suggestions

The links (for profile & library) on the left hand side nav bar are both <a> where the profile link (user image) is `position: absolute`

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/layout/nav/nav.tpl.html
+++ b/kifi-backend/modules/shoebox/angular/src/layout/nav/nav.tpl.html
@@ -90,19 +90,10 @@
         </li>
 
         <li ng-show="sortingMenu.myLibsFirst" ng-repeat="library in myLibsToShow" class="kf-nav-lib-item" ng-class="{ active: isActive(library.url) }">
-          <div class="kf-nav-link-box" ng-click="redirectTo(library.url)">
-            <div class="library-icon-holder">
-              <div class="library-icon svg-private-library-icon" ng-if="library.visibility=='secret'"></div>
-              <div class="library-icon svg-public-library-icon" ng-if="library.visibility=='published'"></div>
-
-              <!-- TODO: remove this when discoverable libraries have been converted to public libraries in the db. -->
-              <div class="library-icon svg-yellow-card-with-globe" ng-if="library.visibility=='discoverable'"></div>
-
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/angular/src/layout/nav/nav.tpl.html
```diff
@@ -90,19 +90,10 @@
...
         <li ng-show="sortingMenu.myLibsFirst" ng-repeat="library in myLibsToShow" class="kf-nav-lib-item" ng-class="{ active: isActive(library.url) }">
-          <div class="kf-nav-link-box" ng-click="redirectTo(library.url)">
-            <div class="library-icon-holder">
-              <div class="library-icon svg-private-library-icon" ng-if="library.visibility=='secret'"></div>
-              <div class="library-icon svg-public-library-icon" ng-if="library.visibility=='published'"></div>
-
-              <!-- TODO: remove this when discoverable libraries have been converted to public libraries in the db. -->
-              <div class="library-icon svg-yellow-card-with-globe" ng-if="library.visibility=='discoverable'"></div>
-
-              <a ng-if="inUserProfileBeta" href="{{getLibraryOwnerProfileUrl(library)}}">
```
ahsu1230, line 101: removed image link because when it's your own library, no owner picture will show up anyway

kifi-backend/modules/shoebox/angular/src/layout/nav/nav.tpl.html
```diff
@@ -90,19 +90,10 @@
         </li>
 
         <li ng-show="sortingMenu.myLibsFirst" ng-repeat="library in myLibsToShow" class="kf-nav-lib-item" ng-class="{ active: isActive(library.url) }">
-          <div class="kf-nav-link-box" ng-click="redirectTo(library.url)">
-            <div class="library-icon-holder">
```
ahsu1230, line 94: no use for icon holder because all it holds is the green, red libary-icon

kifi-backend/modules/shoebox/angular/src/layout/nav/nav.tpl.html
yipingliao, line 101: :)

yipingliao:
:+1:

2is10:
:+1:

PR #12979, opened 2015-01-07
Add `extra` field to Global notifications

andrewconner:
This is for information useful to mobile clients who are not going to `linkUrl`. Immediately, this will enable tapping on new followers to go to their profile (and not the user's own library).

@stkem: Obviously the current API is driven 100% by the extension's needs. Perhaps we should re-evaluate a good intra-service API for notifications soon?

@mrbrown14


andrewconner:
Of note, this does not fix old notifications.

PR #12980, opened 2015-01-07
refactor profile libraries database queries out of libraryMembershipRepo...

ahsu1230:
... and into libraryRepo

Just in my opinion, `LibraryMembershipRepo` should only give back `LibraryMembership`. When we want to retrieve Libraries, we should retrieve it out of the `LibraryRepo` instead. So here, when we want to retrieve libraries for profiles, they should be coming from the `LibraryRepo` because (1) we can use a `GetResult` in `LibraryRepo` and (2) it will be faster than going back into the table to fetch from libraryIds

@andrewconner @2is10 @eishay


andrewconner:
Works for me

2is10:
:+1:  seems more consistent

eishay:
thanks

PR #12981, opened 2015-01-07
turn back visibility to not override listed

ahsu1230:
@yipingliao @2is10 

reverting a few changes as discussed from https://github.com/kifi/keepit/pull/12942


2is10:
:+1: thanks

PR #12984, opened 2015-01-07
better error message for urban airship

eishay:
the errors we had yesterday night where when we tried to get the status of a device so we'll know not to try to send messages to a "dead" device. they did not effect the actual push notifications. 
with better error messages we should get a better understanding of what's wrong.


eishay:
@andrewconner @stkem

andrewconner:
Looks good

stkem:
:+1:

PR #12988, opened 2015-01-07
Lib recos updates

joshm1:
@yingjieMiao @stkem


joshm1:
I'll reset sequence numbers to make sure curator libraries are updated with the 2 new fields and to recompute all lib recos

PR #12989, opened 2015-01-07
introducing BasicUserWithFriendStatus JSON model for user profiles and library cards

2is10:
@ahsu1230 
cc @eishay @andrewconner


kifi-backend/modules/common/app/com/keepit/search/user/UserSearchResult.scala
```diff
@@ -21,16 +21,16 @@ object UserHit {
...
     (__ \ 'hits).format[Array[UserHit]] and
     (__ \ 'context).format[String]
   )(UserSearchResult.apply, unlift(UserSearchResult.unapply))
 }
 
 object UserSearchRequest {
-  implicit val userIdFormat = Id.format[User]
-  implicit val userSearchRequestFormat = (
+  private implicit val userIdFormat = Id.format[User]
+  implicit val format = (
```
2is10, line 33: I’ve done this in a few places. I think it’s a much better pattern to expose `UserSearchRequest.format` instead of `UserSearchRequest.userSearchRequestFormat`. And to avoid exposing helper implicit formats.

kifi-backend/modules/common/app/com/keepit/social/BasicUser.scala
```diff
@@ -16,47 +16,57 @@ trait BasicUserLikeEntity {
...
+  def fullName: String = s"$firstName $lastName"
+}
+
+object BasicUserFields {
+  val format =
+    (__ \ 'id).format[ExternalId[User]] and
+      (__ \ 'firstName).format[String] and
+      (__ \ 'lastName).format[String] and
+      (__ \ 'pictureName).format[String] and
+      (__ \ 'username).format[Username]
```
2is10, line 52: `#scalariformFAIL`

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -1306,7 +1306,7 @@ class LibraryCommander @Inject() (
       color = lib.color,
       image = image.map(LibraryImageInfo.createInfo),
       slug = lib.slug,
-      owner = owner,
+      owner = BasicUserWithFriendStatus.from(owner),
```
2is10: This has no effect today. If we decide we want to include friend status here eventually, we can look it up and then include it here in the future. Yiping says we may need it for tracking purposes on the profile page, and it would be handy for library owner hovercards.

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryView.scala
```diff
@@ -49,6 +51,47 @@ case class LibraryFail(status: Int, message: String)
...
+    (__ \ 'friendRequestSentAt).formatNullable[DateTime] and
+    (__ \ 'friendRequestReceivedAt).formatNullable[DateTime]
+  )(BasicUserWithFriendStatus.apply, unlift(BasicUserWithFriendStatus.unapply))
+
+  def from(u: User): BasicUserWithFriendStatus = from(BasicUser.fromUser(u))
+  def from(u: User, isFriend: Boolean): BasicUserWithFriendStatus = from(BasicUser.fromUser(u), isFriend)
+  def from(u: User, friendRequestSentAt: Option[DateTime] = None, friendRequestReceivedAt: Option[DateTime] = None): BasicUserWithFriendStatus = {
+    from(BasicUser.fromUser(u), friendRequestSentAt, friendRequestReceivedAt)
+  }
+  def from(u: BasicUser): BasicUserWithFriendStatus = from(u, None, None, None)
```
andrewconner: Hm, worried that this may cause confusion. People see `BasicUserWithFriendStatus`, but there's no friend status, and they'd have to track down why. What do you think?

kifi-backend/modules/shoebox/test/com/keepit/model/LibraryViewTest.scala
```diff
@@ -0,0 +1,49 @@
...
+        lastName = "Christie",
+        pictureName = "1.jpg",
+        username = Username("cc"),
+        isFriend = Some(false),
+        friendRequestSentAt = Some(new DateTime(23948192381L)),
+        friendRequestReceivedAt = None)) ===
+        Json.parse("""{
+          "id": "58328718-0222-47bf-9b12-d2d781cb8b0c",
+          "firstName": "Chris",
+          "lastName": "Christie",
```
andrewconner, line 24: How edgy, http://www.ibtimes.com/chris-christie-pushed-port-authority-give-contract-jerry-jones-firm-1774154

kifi-backend/modules/common/app/com/keepit/social/BasicUser.scala
andrewconner, line 52: IntelliJ actually does that for me, but not scaliform. I always have to manually remove it, and then it's fine.

kifi-backend/modules/common/app/com/keepit/search/user/UserSearchResult.scala
andrewconner, line 33: Good call

kifi-backend/modules/shoebox/test/com/keepit/model/LibraryViewTest.scala
2is10, line 24: I can’t take credit, unfortunately. Just copied and modified this fake data from elsewhere.

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryView.scala
2is10: That’s a valid concern. The most common case where we won’t include friend status is when the viewer is not logged in. It’d be a shame to have to use all different class hierarchies to handle that.

Are you just objecting to this convenience method, specifically? I’d be happy to remove it if you think the resulting (larger) code at call sites would be clearer.

kifi-backend/modules/common/app/com/keepit/social/BasicUser.scala
2is10, line 52: Hmm... I’m not using IntelliJ. Just tried it again—compiling reformats this.

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryView.scala
andrewconner: We've been bitten several times by classes that become overly versatile where they primarily have optional members and factories that make it easy to construct "incomplete" objects. See `KeepInfo` for an extreme example: few places actually populate the `KeepInfo` fully.

This seems like it's very easy to go from `BasicUser` to `BasicUserWithFriendStatus`, leading a consumer to believe that because they have a `BasicUserWithFriendStatus`, the fields are correct.

Not sure of a generic solution. Moving the `None, None, None` to call sites doesn't really fix this, because it just widens the surface area of sites that do not fully populate it. Thoughts?

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
andrewconner: For example, here. A consumer may believe that because the type is `BasicUserWithFriendStatus`, it would contain correct friend status information.

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryView.scala
2is10: I’ve made the factory methods clearer. Thanks for your comments and in-person discussion.

PR #12990, opened 2015-01-07
count of libraries on profile is only created libraries (no followed libraries count)

ahsu1230:
@andrewconner @eishay


kifi-backend/modules/shoebox/app/com/keepit/model/LibraryRepo.scala
```diff
@@ -196,13 +195,6 @@ class LibraryRepoImpl @Inject() (
     query.as[Int].firstOption.getOrElse(0)
```
2is10, line 191: Should the `countFollowLibraries` option be removed here (and other methods)?

ahsu1230, line 191: Yea, I guess it should. We're making a final decision to only include created library counts?

2is10, line 191: Yes, we may also want to be able to count followed libraries eventually, but we’ll never want to add owned + followed.

PR #12991, opened 2015-01-07
optimizing ProbabilityDensity further

ymatsuda:
@leogrim @yingjieMiao 
- not holding onto the density parameter (less memory used by caching in Wanderer)
- less autoboxing

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/common/math/ProbabilityDensity.scala
+++ b/kifi-backend/modules/common/app/com/keepit/common/math/ProbabilityDensity.scala
@@ -7,19 +7,18 @@ import scala.collection.mutable
 case class Probability[A](outcome: A, probability: Double)
 
 //Note: The order of the argument Seq here is very important. It, along with the salt, uniquely determines which users go in which bin.
-case class ProbabilityDensity[A](density: Seq[Probability[A]]) {
-  require(density.forall(_.probability >= 0), "Probabilities must be non-negative")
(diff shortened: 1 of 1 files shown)


yingjieMiao:
I see 'density' and 'cumulative' is swapped. How does this affect json serialization and cache?

yingjieMiao:
oh, never mind. The apply method changed.

kifi-backend/modules/common/app/com/keepit/common/math/ProbabilityDensity.scala
```diff
@@ -52,19 +59,35 @@ object ProbabilityDensity {
...
+    if (weight > 0) {
+      buf += Probability(outcome, weight)
+      sum += weight
+    }
+  }
 
-  def build(): ProbabilityDensity[A] = ProbabilityDensity.normalized(buf)
+  def build(): ProbabilityDensity[A] = {
+    if (sum > 0.0) ProbabilityDensity(buf, sum) else ProbabilityDensity(buf)
+  }
```
yingjieMiao, line 92: I am not sure if this changes the behavior. 

Originally, we have a un-normalized distribution, and call `normalize` manually in the code.

consider density: (0.1, 0.2). With old code, there is a 0.7 probability we sample a None. 
With new code, this is normalized automatically? So 1/3 probability for 1st item, 2/3 prob for 2nd item?

yingjieMiao, line 92: never mind. discussed in person.

PR #12992, opened 2015-01-07
Fix library search clear bug.

yipingliao:
@andrewconner cc @ahsu1230 

Ugh, I found this ugly bug about an hour ago. When you clear the search input in a library search or delete all the text in a library search, the page redirects to home! This is the fix for it.


andrewconner:
Nice catch

PR #12999, opened 2015-01-08
Add a couple more tweaks to make the search text input experience better

yipingliao:
@2is10 
These are a couple more tweaks to the search input boxes UI experience; Danny and Jonathan said that the inputs were still not responsive enough, so I am trying these out.

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/layout/header/HeaderCtrl.js
+++ b/kifi-backend/modules/shoebox/angular/src/layout/header/HeaderCtrl.js
@@ -110,7 +110,7 @@ angular.module('kifi')
           $location.search('q', $scope.search.text).replace(); // this keeps any existing URL params
         }
       });
-    }, 100);
+    }, 250);
 
     $scope.onSearchBarClicked = function () {
       if ($scope.library && $scope.library.id) {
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/angular/src/search/search.js
```diff
@@ -239,12 +245,12 @@ angular.module('kifi')
     // Watches and event listeners.
     //
     var newSearch = _.debounce(function () {
-        // Use $state.params instead of $stateParams because changes to $stateParams
-        // does not propagate to HeaderCtrl when it is injected there.
-        // See: http://stackoverflow.com/questions/23081397/ui-router-stateparams-vs-state-params
-        _.assign($state.params, $location.search());
-        init();
-      }, 250);
+      // Use $state.params instead of $stateParams because changes to $stateParams
```
yipingliao, line 248: This is just an indentation change.

kifi-backend/modules/shoebox/angular/src/layout/header/HeaderCtrl.js
```diff
@@ -110,7 +110,7 @@ angular.module('kifi')
           $location.search('q', $scope.search.text).replace(); // this keeps any existing URL params
         }
       });
-    }, 100);
+    }, 250);
```
yipingliao, line 113: Increasing the debounce time for the search input text change seems to help with seeing the initial burst of search input text faster.

2is10, line 113: sounds good to me

2is10:
:+1:

PR #13001, opened 2015-01-08
remove chip from search bar when going to profile

ahsu1230:
and quick change for viewing your own followed libraries (forgot to order correctly)

@andrewconner @yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.styl
+++ b/kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.styl
@@ -73,6 +73,9 @@
   background: #e8e8e8
   text-align: center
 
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.styl
```diff
@@ -73,6 +73,9 @@
   background: #e8e8e8
   text-align: center
 
+.kf-upl-card-list
+  margin-bottom: 40px
+
```
ahsu1230, line 78: this is added so there's no hard cut-off at the bottom of the list and to allow a bit of space. It just bothered me - couldn't help myself =P

2is10:
:+1:

PR #13002, opened 2015-01-08
adding friend status to library owners in profile library responses

2is10:
@eishay 
cc @ahsu1230 @yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
@@ -65,6 +65,8 @@ class LibraryCommander @Inject() (
     countByLibraryCache: CountByLibraryCache,
     typeaheadCommander: TypeaheadCommander,
     collectionRepo: CollectionRepo,
+    connectionRepo: UserConnectionRepo,
+    friendRequestRepo: FriendRequestRepo,
     s3ImageStore: S3ImageStore,
     emailOptOutCommander: EmailOptOutCommander,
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -1306,4 +1309,34 @@ class LibraryCommander @Inject() (
...
 
+  private def augmentWithFriendStatus(owners: Map[Id[User], BasicUser], viewerId: Option[Id[User]])(implicit session: RSession): Map[Id[User], BasicUserWithFriendStatus] = {
+    val otherOwnerIds = owners.keySet -- viewerId.toSet
+    if (viewerId.isDefined && otherOwnerIds.nonEmpty) {
+      val friendIds = connectionRepo.getConnectedUsers(viewerId.get)
+      val nonFriendOwnerIds = otherOwnerIds -- friendIds
+      if (nonFriendOwnerIds.nonEmpty) {
+        val reqsSent = friendRequestRepo.getBySender(viewerId.get, Set(FriendRequestStates.ACTIVE, FriendRequestStates.IGNORED))
+        val reqsReceived = friendRequestRepo.getByRecipient(viewerId.get, Set(FriendRequestStates.ACTIVE))
+        owners.map {
```
2is10, line 1320: This has a little room for optimization. We don’t need to load all connected user IDs and friend requests if there is only one user ID we’re checking for. I have a subsequent pull request in the works that does additional optimization and refactors this to its own commander for reuse.

PR #13004, opened 2015-01-08
Hide user profile invited tab

yipingliao:
@andrewconner @2is10 @ahsu1230 

A small pull that somehow seems like should be cc'ed to three people. :) Please see commit messages and inline comments for details. Thanks!

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
+++ b/kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
@@ -227,8 +227,8 @@ angular.module('kifi')
 
 
 .controller('UserProfileLibrariesCtrl', [
-  '$scope', '$rootScope', '$state', '$stateParams', 'routeService', 'keepWhoService', 'profileService', 'userProfileActionService',
-  function ($scope, $rootScope, $state, $stateParams, routeService, keepWhoService, profileService, userProfileActionService) {
+  '$scope', '$rootScope', '$state', '$stateParams', 'routeService', 'keepWhoService', 'profileService', 'userProfileActionService', 'libraryService',
+  function ($scope, $rootScope, $state, $stateParams, routeService, keepWhoService, profileService, userProfileActionService, libraryService) {
     var colors = ['#C764A2', '#E35957', '#FF9430', '#2EC89A', '#3975BF', '#955CB4', '#FAB200'];
     var username = $stateParams.username;
     var fetchPageSize = 12;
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
```diff
@@ -166,7 +166,7 @@ angular.module('kifi')
     };
 
     $scope.isMyLibrary = function(libraryOwnerId) {
-      return libraryOwnerId === $scope.profile.id;
+      return $scope.profile && (libraryOwnerId === $scope.profile.id);
```
yipingliao, line 169: @ahsu1230 We can't be sure that `$scope.profile` is non-null because that is populated by an XHR call - adding a null check here since it's being used as a scope method that will be invoked every digest cycle.

ahsu1230, line 169: I see, I guess that's where the bug came in. Thanks!

kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
```diff
@@ -271,6 +271,12 @@ angular.module('kifi')
       return hasMoreLibraries;
     };
 
+    $scope.showInvitedLibraries = function () {
+      // TODO: modify the profile endpoint to return number of invited libraries instead of relying
+      // on libraryService.invitedSummaries.length.
+      return $scope.viewingOwnProfile && libraryService.invitedSummaries.length;
```
ahsu1230, line 277: so I guess we need to keep `numLibraries` and add in `numInvited`?

kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
```diff
@@ -271,6 +271,12 @@ angular.module('kifi')
       return hasMoreLibraries;
     };
 
+    $scope.showInvitedLibraries = function () {
+      // TODO: modify the profile endpoint to return number of invited libraries instead of relying
```
yipingliao, line 275: @2is10 What do you think about adding a field to the user profile endpoint that indicates how many libraries they are invited to (or just a boolean value that indicates whether any invited libraries should be shown). Is that a proper field for the profile endpoint? I am asking because the `libraryService.invitedSummaries` that I am using here may go away (@andrewconner has previously indicated that that may go away when we remove the left nav panel).

kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
yipingliao, line 277: Yeah, not sure about this one... have asked @2is10 for advice.

2is10, line 277: Seems fine to me to add a field. Since it will be in a JSON object representing the user profile, I think we should be clear that we’re talking about _library_ invitations. Maybe `numPendingLibraryInvitations`? Or just `numInvitedLibraries`.

PR #13006, opened 2015-01-08
[BE] assign a random library color to all user_created libraries

ahsu1230:
(1) when a library is created, assign a random library color (based on the current color selection on the front-end)
(2) there's also an endpoint to be called that will paginate between all `user_created` libraries and each saving a randomly chosen color. This endpoint will be removed later once all non-system libraries have a color

@andrewconner @2is10

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/model/Library.scala
+++ b/kifi-backend/modules/common/app/com/keepit/model/Library.scala
@@ -18,6 +18,8 @@ import scala.concurrent.duration.Duration
 import com.keepit.social.BasicUser
 import com.keepit.common.json
 
+import scala.util.Random
+
 case class Library(
     id: Option[Id[Library]] = None,
     createdAt: DateTime = currentDateTime,
@@ -254,4 +256,18 @@ object HexColor {
 
   val regex = "^#[0-9a-f]{6}$".r
   def apply(hex: String): HexColor = new HexColor(hex.toLowerCase)
+
+  val allColors = Seq(
+    HexColor("#C764A2"), // magenta (pink)
+    HexColor("#E35957"), // red
+    HexColor("#FF9430"), // orange (a bit darker)
+    HexColor("#2EC89A"), // indigo (light green)
+    HexColor("#3975BF"), // blue
+    HexColor("#955CB4"), // purple
+    HexColor("#FAB200")) // light orange (more yellow-ish)
+
+  def pickRandomLibraryColor(): HexColor = {
+    val rnd = new Random
+    allColors(rnd.nextInt(allColors.size))
+  }
 }
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
@@ -48,6 +48,28 @@ class LibraryController @Inject() (
   implicit val config: PublicIdConfiguration)
     extends UserActions with LibraryAccessActions with ShoeboxServiceController with Logging {
 
+  def colorAllUserLibraries() = UserAction { request =>
+    val pageSize = 500
+    var c = 0
+    var paginator = Paginator(c, pageSize)
+    var libs = db.readOnlyMaster { implicit s =>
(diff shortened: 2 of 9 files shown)


kifi-backend/modules/common/app/com/keepit/model/Library.scala
```diff
@@ -254,4 +256,18 @@ object HexColor {
...
   def apply(hex: String): HexColor = new HexColor(hex.toLowerCase)
+
+  val allColors = Seq(
+    HexColor("#C764A2"), // magenta (pink)
+    HexColor("#E35957"), // red
+    HexColor("#FF9430"), // orange (a bit darker)
+    HexColor("#2EC89A"), // indigo (light green)
+    HexColor("#3975BF"), // blue
+    HexColor("#955CB4"), // purple
+    HexColor("#FAB200")) // light orange (more yellow-ish)
```
2is10, line 267: How confident are we of these colors? I saw a mobile color picker design from Teehan (looking for it now) and I’m not sure it had exactly these colors.

kifi-backend/modules/shoebox/conf/shoeboxService.routes
```diff
@@ -354,6 +354,8 @@ POST    /site/libraries/add         @com.keepit.controllers.website.LibraryContr
 POST    /site/libraries/copy        @com.keepit.controllers.website.LibraryController.copyKeeps()
 POST    /site/libraries/move        @com.keepit.controllers.website.LibraryController.moveKeeps()
 
+POST    /site/libraries/colorAll    @com.keepit.controllers.website.LibraryController.colorAllUserLibraries()
```
2is10: This should be an admin endpoint in an admin controller, no?

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
```diff
@@ -48,6 +48,28 @@ class LibraryController @Inject() (
...
+        }
+      }
+      c += 1
+      paginator = Paginator(c, pageSize)
+      libs = db.readOnlyMaster { implicit s =>
+        libraryRepo.pageByKind(LibraryKind.USER_CREATED, paginator)
+      }
+    }
+    NoContent
+  }
```
2is10: If we end up needing to run this more than once for whatever reason, it’d be more efficient to filter out libraries that already have colors at the database level.

2is10:
:+1:

kifi-backend/modules/shoebox/test/com/keepit/controllers/website/LibraryControllerTest.scala
```diff
@@ -65,6 +67,44 @@ class LibraryControllerTest extends Specification with ShoeboxTestInjector {
...
+          library().withUser(user5.id.get).withKind(LibraryKind.SYSTEM_SECRET).secret().saved
+
+          libraryRepo.count === 9
+          val allLibs = libraryRepo.all
+          allLibs.count(_.color.isEmpty) === 8
+          allLibs.count(_.color.nonEmpty) === 1
+          user1
+        }
+
+        val libraryController = inject[AdminLibraryController]
```
ahsu1230, line 93: It's in `AdminLibraryController`, but there's no `AdminLibraryControllerTest` so I'll put the test in here for now and remove it later!

2is10, line 93: OK by me

kifi-backend/modules/common/app/com/keepit/model/Library.scala
ahsu1230, line 267: Just confirmed with Mark that these library colors are finalized!

kifi-backend/modules/common/app/com/keepit/model/Library.scala
```diff
@@ -254,4 +256,18 @@ object HexColor {
 
   val regex = "^#[0-9a-f]{6}$".r
   def apply(hex: String): HexColor = new HexColor(hex.toLowerCase)
```
andrewconner, line 258: You may want to match just on the allowed colors, so a client can't pick a non-"approved" color.

PR #13009, opened 2015-01-08
hide related libs during library search

joshm1:
@yipingliao


2is10:
using `ng-hide` so they’ll come back if the search is canceled?

joshm1:
@2is10 yes, that's why i didn't put it in `ng-if`

yipingliao:
Thanks! :+1:

PR #13011, opened 2015-01-08
many edit library actions are very slow, trying to fix some of it.

eishay:
the slowness is especially for super users (lots of libs).
I noticed that few times we call searchClient reindex and analytics inside a db transaction or in a blocking way while the user is waiting. Moved some to a future.


eishay:
@andrewconner @ahsu1230

andrewconner:
Works for me

ahsu1230:
thanks for this! :+1:

PR #13013, opened 2015-01-08
no functional change, just fixing two JSHint warnings

2is10:
@yipingliao


kifi-backend/modules/shoebox/angular/src/layout/header/HeaderCtrl.js
```diff
@@ -85,29 +85,21 @@ angular.module('kifi')
...
+            $location.url($scope.library.url);
           }
-        }
-
-        // We are already on the search page.
-        else {
-          $location.search('q', $scope.search.text).replace(); // this keeps any existing URL params
+        } else if ($scope.search.text) {
+          $location.url('/find?q=' + $scope.search.text);
         }
```
2is10, line 103: just restructured the branches to eliminate one level of nesting

yipingliao:
Thanks!

PR #13014, opened 2015-01-08
using real link for entire click target

2is10:
This is important for usability. Also makes the assumption in `clickCard` that the clicked element is a link correct again.
@joshm1 re: #12993


joshm1:
Thanks!

PR #13017, opened 2015-01-08
linkifying owner name on profile library card

2is10:
@yipingliao


yipingliao:
Not directly related to this pull request: I wonder if we should do anything to notify the user that they are about to click on something that takes them to a different place than clicking on anywhere else on the card. E.g., underlining the user's name on hover and/or change opacity on the user/follower's pictures on hover, because those go to profile pages, not library pages.

2is10:
Agreed, valid concern.

PR #13018, opened 2015-01-08
Do not show library nav or settings/connect UI components for Kifi curators

yipingliao:
@2is10 

Per @andrewconner , there is no indication in the db that a particular user is a "special" Kifi curator. We're keeping a hardcoded list on the frontend to detect this condition.


2is10:
:+1:

PR #13020, opened 2015-01-08
only removing braces and semicolons

2is10:
for consistency with our other .styl files
@joshm1


andrewconner:
Don't ya love the minimalism?

joshm1:
:)

PR #13021, opened 2015-01-08
profile endpoint includes number of distinct libraries you're invited to

ahsu1230:
The Json field `numInvitedLibraries` only returns when you're viewing your own profile
The number is distinct libraries you're invited to

and set library colors are in, so removing admin endpoint that sets all the library colors

@2is10 cc: @yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/UserController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/UserController.scala
@@ -594,19 +594,17 @@ class UserController @Inject() (
     userCommander.profile(Username(username), viewer) match {
       case None => NotFound(s"can't find username $username")
       case Some(profile) =>
-        val numLibraries = libraryCommander.countLibraries(profile.userId, viewer.map(_.id.get))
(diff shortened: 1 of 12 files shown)


yipingliao:
Awesome. Thanks!

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/UserController.scala
```diff
@@ -594,10 +595,19 @@ class UserController @Inject() (
...
-        ))
+        )
+        viewer match {
+          case Some(user) if user.id.get == profile.userId =>
+            val numInvitedLibs = db.readOnlyMaster { implicit s =>
+              libraryInviteRepo.countDistinctWithUserId(profile.userId)
+            }
+            Ok(json ++ Json.obj("numInvitedLibraries" -> numInvitedLibs))
+          case _ => Ok(json)
+        }
```
2is10, line 608: Nice!
One possible optimization would be to reuse the DB session we’re already using in `libraryCommander.countLibraries`.

2is10:
:+1:

PR #13023, opened 2015-01-08
trying some hover effects for library curator picture and name

2is10:
@yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.styl
+++ b/kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.styl
@@ -212,6 +212,9 @@ name-line-height = 25px
   letter-spacing: normal
   margin-right: 2px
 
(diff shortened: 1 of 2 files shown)


yipingliao:
Nice!

kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.styl
```diff
@@ -249,6 +252,17 @@ name-line-height = 25px
...
+  &:hover
+    transform: scale(1.03)
+
+.kf-upl-card-owner-pic
+  position: absolute
+  top: 0
+  left: 0
+  width: 100%
+  height: 100%
+  border-radius: 50%
```
2is10, line 265: This might be unnecessary, but I’ve been bitten many times in the past by unstyled links ending up bigger than their content, so now positioning and sizing the link explicitly and the image inside too.

PR #13027, opened 2015-01-09
sampling edges without creating ProbabilityDensity

ymatsuda:
@leogrim @yingjieMiao @stkem @eishay 

Half of graph's Full GC we are seeing are caused by "promotion failure", which implies that JVM failed to find a continuous free space to store large objects during GC. My guess is that ProbabilityDensity is causing this. It can become quite big when hitting a large node like a topic node and cause "promotion failure".

`SimpleSampler` samples outcomes by probabilities. It creates sample points (`Array[Double]`) before scanning, and choose outcomes as the cumulative probability passes the sample points. The drawback of this logic is that it creates a large `Array[Double]` when the sample size is big. And our sample size can be big actually.

To cope with this, `SegmentedSampler` is defined. It segments the range of cumulative probability into equal size segments, and works on a single segment at a time, so it can reuse the fixed size array for sample points.

Note that the sampling is not strictly random because each segments always has the same number of samples. It maybe good for our purpose.

The sample size allowed by `SegmentedSampler` is a multiple of `segmentSize`. It should always used though `CombinedSampler` to allow any sample size. `CombinedSampler` combines `SegmentSampler` and `SimpleSampler` to do that.


eishay:
nice!

kifi-backend/modules/common/app/com/keepit/common/math/Sampler.scala
```diff
@@ -0,0 +1,113 @@
...
+
+  require(numSegments > 0, "the number of segments must be greater than 0")
+
+  val sampleSize = segmentSize * numSegments
+
+  private[this] var sampledCount: Int = 0
+
+  private[this] val rnd: Array[Double] = new Array[Double](segmentSize)
+  private[this] val segWidth = 1.0 / numSegments
+  private[this] var segSampledCount: Int = 0
```
yingjieMiao, line 92: is `segSampledCount` ever mutated anywhere?

kifi-backend/modules/common/app/com/keepit/common/math/Sampler.scala
```diff
@@ -0,0 +1,113 @@
...
+  randomize(rnd, 0.0, segWidth)
+
+  private[this] var cumulative: Double = 0.0
+
+  def put(outcome: A, probability: Double): Unit = {
+    cumulative += probability
+    val lastCount = sampledCount
+
+    while (segNo < numSegments && cumulative > rnd(segSampledCount)) {
+      sampledCount += 1
```
yingjieMiao, line 104: also incre `segSampledCount` here?

kifi-backend/modules/common/app/com/keepit/common/math/Sampler.scala
```diff
@@ -0,0 +1,113 @@
...
+    while (i < rnd.length) {
+      rnd(i) = Random.nextDouble() * width + base
+      i += 1
+    }
+    Arrays.sort(rnd)
+  }
+}
+
+private[math] class SimpleSampler[A](val sampleSize: Int, out: (A, Int) => Unit) extends Sampler[A] {
+  import Sampler._
```
yingjieMiao: Send a 'mutating' method to the class constructor looks a bit confusing for me. Looks like the `Int` argument in this `out` method is generated during sampling, i.e. num of sampled points. 

I am wondering if we can make `put` method return this number, instead of Unit. And the caller is responsible for dealing with the number. 

Based on my understanding, every time we run the sampler, some external variables are mutated. It's kind of unusual for a sampler. 

I am sure there are some performance considerations here, or maybe I missed something.

yingjieMiao:
Cool. It's like a 'moving window' Monte Carlo sampling.

:+1:

yingjieMiao:
One thing regarding 'gigantic topic' vertexes. This could be solved if we consider 'hierarchical' connection. every topic connection this way will be super lightweight.

kifi-backend/modules/common/app/com/keepit/common/math/Sampler.scala
```diff
@@ -0,0 +1,113 @@
...
+
+  def apply[A](sampleSize: Int, out: (A, Int) => Unit) = {
+    require(sampleSize > 0, "the sample size must be greater than 0")
+
+    if (sampleSize <= segmentSize) new SimpleSampler[A](sampleSize, out)
+    else new CombinedSampler[A](sampleSize, out)
+  }
+
+  private[math] def randomize(rnd: Array[Double], base: Double, width: Double): Unit = {
+    var i = 0
```
yingjieMiao: probably more conventional/user friendly to use `low`, `high`, instead of `base` and `width`, where `width = high - low`

kifi-backend/modules/common/app/com/keepit/common/math/Sampler.scala
ymatsuda: I will your suggestion. It is much cleaner. Thanks!

kifi-backend/modules/common/app/com/keepit/common/math/Sampler.scala
ymatsuda, line 104: Oops. Thank you for catching this.

kifi-backend/modules/common/app/com/keepit/common/math/Sampler.scala
ymatsuda: got it.

yingjieMiao:
lgtm!

PR #13028, opened 2015-01-09
Limit width upl logged out

yipingliao:
@2is10 cc @ahsu1230 
- Limit width of user profile library cards to 3 cards per row, per Mark
- Switch to using `numInvitedLibraries` on profile endpoint that @ahsu1230 just pushed

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
+++ b/kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
@@ -285,9 +285,7 @@ angular.module('kifi')
     };
 
     $scope.showInvitedLibraries = function () {
-      // TODO: modify the profile endpoint to return number of invited libraries instead of relying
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
```diff
@@ -285,9 +285,7 @@ angular.module('kifi')
     };
 
     $scope.showInvitedLibraries = function () {
-      // TODO: modify the profile endpoint to return number of invited libraries instead of relying
-      // on libraryService.invitedSummaries.length.
-      return $scope.viewingOwnProfile && libraryService.invitedSummaries.length;
+      return $scope.profile && $scope.profile.numInvitedLibraries && $scope.viewingOwnProfile;
```
ahsu1230, line 288: thanks!

2is10:
:+1:

PR #13030, opened 2015-01-09
making related library cards look just like user profile library cards

2is10:
@yipingliao @joshm1 

Also removing hard-coding of colors, since Aaron has populated colors for every library and one is randomly chosen for every newly created library now.

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/relatedLibraries.styl
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/relatedLibraries.styl
@@ -32,6 +32,7 @@ mobileBorderRadius = 54px
     margin-top: 0
     float: none
     width: initial
+    text-align: center
 
 .kf-related-libraries-see-more
   margin: 0 4px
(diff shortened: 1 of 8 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/relatedLibraries.styl
```diff
@@ -83,16 +84,21 @@ mobileBorderRadius = 54px
...
-  margin-bottom: 1em
+  margin-bottom: 2em
   .kf-mobile &
-    margin: 40px 25px 10px 30px
+    margin: 0 auto 500px
     float: none
-    width: auto
+    width: 210px
+    transform: scale(2.5)
+    transform-origin: center top
```
2is10, line 98: using big margin-bottom + scaling to get accurate larger rendering on mobile

2is10:
I started by making small changes to the small library card widget to make its appearance more similar, then decided it would be faster to just replace the whole template and all styles with those from the user profile. Just using a different class prefix.

2is10:
This may still need tweaking. Merging now to get it out for evaluation and QA. Happy to address any comments in a follow-up.

kifi-backend/modules/shoebox/angular/src/libraries/smallLibraryCard.styl
```diff
@@ -1,198 +1,174 @@
...
+  border-radius: 16px
+  vertical-align: top
   position: relative
-  border-radius: desktopBorderRadius desktopBorderRadius
-  z-index: 1
+  overflow: hidden
   text-align: center
-  color: #fff
-  font-family: Lato
+  box-shadow: 0 3px #f5f5f5, 0 4px #c9ccc0, 0 6px #e5e5e5, 0 7px #c4c6cb, 0 8px #d4d4d4, 0 9px #bec1c5
```
joshm1, line 11: nice! didn't know you could do this

joshm1:
lgtm, thanks!

PR #13033, opened 2015-01-09
formatting profile URLs ahead of time

2is10:
@yipingliao 

This also fixes some broken links in the new related library card.


yipingliao:
Nice! Thanks!

PR #13034, opened 2015-01-09
Greatly simplify number of keeps information in keep cards.

yipingliao:
@andrewconner 

cc @eishay Note that you can drop the `createdAt` timestamp field on keeps returned from the backend; the front end currently handles that gracefully.


eishay:
thanks!

andrewconner:
Nice, removing code!

PR #13036, opened 2015-01-09
styling the settings gear on profile library cards

2is10:
@ahsu1230 cc @yipingliao


kifi-backend/modules/shoebox/angular/img/svg/settings-gear.svg
```diff
@@ -5,9 +5,9 @@
     <desc>Created with Sketch.</desc>
     <defs></defs>
     <g id="Welcome" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd" sketch:type="MSPage">
-        <g id="Desktop-HD-Profile-w/-Main--Logged-In" sketch:type="MSArtboardGroup" transform="translate(-878.000000, -130.000000)" fill="#FFFFFF">
-            <g id="User-Info-2" sketch:type="MSLayerGroup" transform="translate(0.000000, 63.000000)">
-                <g id="Settings-Icon-5" transform="translate(877.055556, 67.000000)" sketch:type="MSShapeGroup">
+        <g id="Desktop-HD-Profile-w/-Main--Logged-In" sketch:type="MSArtboardGroup" fill="#FFFFFF">
+            <g id="User-Info-2" sketch:type="MSLayerGroup">
+                <g id="Settings-Icon-5" sketch:type="MSShapeGroup">
```
2is10, line 10: Removing the SVG translation here fixed the rendering glitch. It seems our SVG minifier doesn’t handle it properly, but I didn’t fully pinpoint the cause.

kifi-backend/modules/shoebox/angular/src/userProfile/userProfileLibrariesList.tpl.html
```diff
@@ -1,6 +1,6 @@
 <div class="kf-upl-card-list" smart-scroll scroll-distance="'100%'" scroll-disabled="!hasMoreLibraries()" scroll-next="fetchLibraries()">
   <div class="kf-upl-card" ng-repeat="lib in libraries" ng-style="{'background-color': lib.color}">
-    <div class="kf-upl-card-setting svg-settings-gear" ng-if="isMyLibrary(lib.owner.id) && lib.kind==='user_created'" ng-click="openModifyLibrary(lib)"></div>
+    <a class="kf-upl-card-setting svg-settings-gear" ng-if="isMyLibrary(lib.owner.id) && lib.kind==='user_created'" href="javascript:" ng-click="openModifyLibrary(lib)"></a>
```
ahsu1230, line 3: just curious, why turn this to an `<a>` when it's not technically a link (user does not go to a different page), but a popup button?

ahsu1230:
nice thanks!

kifi-backend/modules/shoebox/angular/src/userProfile/userProfileLibrariesList.tpl.html
2is10, line 3: Anything that can be interacted with should be a link or form control, for accessibility reasons. Blind people can only discover actions available to them (using screen readers) if they are a link or form control. Parts of a page that can be interacted with should also be focusable (reachable via Tab key and clickable via Enter key). This is nice for anyone, but especially important for people who have difficulty using pointers and/or who use text-only browsers.

The [links section](http://www.w3.org/TR/html401/struct/links.html) of the HTML 4 spec is pretty explicit about one purpose of links being linking to a specific element in the current document (page). It intends for people to use named anchors for that, but people have long used scripts to achieve that functionality too.

In this particular case, the gear icon effectively links to the settings form for the library, which happens to be shown in a dialog and to be added to the page only when needed. We could assign an `id` to the dialog and use it in the `href` attribute of this link, to better conform with the HTML spec, but that can be kind of error prone, so our convention is to use `javascript:` as the `href` URL for links handled by scripts.

PR #13037, opened 2015-01-09
removing translation here too

2is10:
@ahsu1230 cc @yipingliao


yipingliao:
Argh, so that's what caused it! I had gone in there to play with the SVG a bit, but didn't succeed in finding what was responsible for the shift. Thanks!

PR #13041, opened 2015-01-09
showing lock icon in place of owner picture on non-system private libs

2is10:
@yipingliao 

This is where Mark wants it for now. Top left corner of card didn’t look very good.


kifi-backend/modules/shoebox/angular/img/svg/card-star.svg
```diff
@@ -5,7 +5,7 @@
     <desc>Created with Sketch.</desc>
     <defs></defs>
     <g id="Page-1" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd" sketch:type="MSPage">
-        <g id="Keep-2-Choose-Library-2" sketch:type="MSArtboardGroup" transform="translate(-70.000000, -202.000000)" stroke="#FFFFFF" stroke-width="3" opacity="0.5">
+        <g id="Keep-2-Choose-Library-2" sketch:type="MSArtboardGroup" transform="translate(-70.000000, -202.000000)" stroke="#FFFFFF" stroke-width="3">
```
2is10, line 8: removing opacity from these SVG files so we can control it from CSS
(didn’t actually end up changing it in any cases for this pull request)

PR #13046, opened 2015-01-09
related lib frontend updates, tiny update to non-linear scoring

joshm1:
prevents non-linear from returning a score of 0 if topicScore=0

@stkem @2is10 
cc @ymatsuda

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/smallLibraryCard.styl
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/smallLibraryCard.styl
@@ -141,6 +141,9 @@ name-line-height = 25px
   height: 35px
   border-top: 1px solid rgba(255,255,255,.3)
   -webkit-font-smoothing: antialiased
+  text-align: center
+  color: #fff
+  line-height: 35px
 
 follower-pic-diameter = 23px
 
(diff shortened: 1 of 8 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/smallLibraryCard.styl
```diff
@@ -141,6 +141,9 @@ name-line-height = 25px
   height: 35px
   border-top: 1px solid rgba(255,255,255,.3)
   -webkit-font-smoothing: antialiased
+  text-align: center
+  color: #fff
```
2is10, line 145: We’re using partially transparent white for the other footers types, for a nice effect. Did you consider `rgba(255,255,255.7)`?

Also, each kind of footer has its own container, and those are styled differently (one level deeper). This new styling at the level will break the other kinds of footers, so it would be better to either remove them or move this new styling one level deeper.

2is10, line 145: Oh, I see now that you did remove the portion of the template for the alternate footers. Would be good to remove the corresponding styles, too (everything else below this point in this stylesheet).

2is10:
:+1:

kifi-backend/modules/shoebox/angular/src/libraries/smallLibraryCard.styl
joshm1, line 145: I agree, partially transparent will look nicer. I just used what Mark said "white". I'll update per your suggestions. Thanks!

PR #13057, opened 2015-01-10
do not show libraries from fake user

yingjieMiao:
@joshm1 @stkem

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/RelatedLibraryCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/RelatedLibraryCommander.scala
@@ -30,6 +30,7 @@ class RelatedLibraryCommanderImpl @Inject() (
     libMemRepo: LibraryMembershipRepo,
     libCommander: LibraryCommander,
     cortex: CortexServiceClient,
+    experimentCommander: LocalUserExperimentCommander,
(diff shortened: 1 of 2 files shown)


joshm1:
lgtm

kifi-backend/modules/shoebox/app/com/keepit/commanders/RelatedLibraryCommander.scala
```diff
@@ -40,11 +41,25 @@ class RelatedLibraryCommanderImpl @Inject() (
   // main method
   def suggestedLibrariesInfo(libId: Id[Library], userIdOpt: Option[Id[User]]): Future[(Seq[FullLibraryInfo], Seq[RelatedLibraryKind])] = {
     val suggestedLibsFut = suggestedLibraries(libId)
+
+    val fakeOwnersFut = suggestedLibsFut.map { libs =>
+      val owners = libs.map { _.library.ownerId }.distinct
+      owners.filter { id => experimentCommander.getExperimentsByUser(id).contains(ExperimentType.FAKE) }.toSet
```
eishay: instead of fetching all users experiments from cache one by one I recommend getting `getAllFakeUsers` from the shoebox client and then `filterNot` if the owner's id in this set.

yingjieMiao:
@eishay  updated per your comment. It's much cleaner now.

PR #13058, opened 2015-01-10
Move out code that was unnecessarily dependent on fetch-library-summary promise resolutions

yipingliao:
@andrewconner cc @eishay 

Thanks to @eishay's pointers, found two instances where we are waiting for `fetchLibrarySummaries` to return before moving ahead with UI changes that were actually unnecessary. Note that we have to wait for `fetchLibrarySummaries` to return before we can follow a library because we need the most current information about whether the user is already following the library.


yipingliao:
@eishay History of fetching the libraries summaries in this pull: https://github.com/kifi/keepit/pull/12922 Actually this was less than 2 weeks ago. Per the pull description, if the backend can return an indication of whether the user has joined the library already, we won't have to refetch library summaries at all. If this is not urgent, I'll do the backend bit; am going to get Aaron to help me become backend-literate.

andrewconner:
Awesome. I'm happy to help as well.

PR #13059, opened 2015-01-10
update marketing site

joshm1:
see https://github.com/kifi/FrogSpark/pull/78

@andrewconner


andrewconner:
Why not do the redirect in Play? I thought we had code for this.

andrewconner:
@joshm1 This needs to be server side. The fix now leaves the browser on that page (the user will see it when they open Safari again). We can't show it.

cc @eishay

joshm1:
@andrewconner we don't redirect in Play b/c we want the page to be appealing on mobile and not a dead-end if they go back to Safari (like you said).  There is also some tracking code in place here so we know how many mobile users view the landing page (which I understand could still be done if we rendered completely different HTML like we used to).  I spoke with Sterling, we know that the site is still vulnerable to letting users sign in from this landing page on mobile. This is going to be removed first thing Monday and replaced with some "Get the app" button.

cc @eishay

PR #13061, opened 2015-01-11
for library header, prefer to show first users with a profile picture

eishay:


eishay:
@andrewconner @ahsu1230

andrewconner:
That should work, looks good.

ahsu1230:
:+1:

PR #13062, opened 2015-01-12
checking for a specific query parameter more specifically

2is10:
I think this is the idiomatic way to do it. The `o=lr` is supposed to be in the query string, yes?
@joshm1


joshm1:
yep! thanks

joshm1:
@2is10 I'm working on section of code now, I'll commit this change. thanks!

PR #13063, opened 2015-01-12
css updates for lib recos and related libs

joshm1:
@stkem @2is10

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/recos/recos.styl
+++ b/kifi-backend/modules/shoebox/angular/src/recos/recos.styl
@@ -95,6 +95,11 @@
   .kf-keep-lib-header
     padding-top: 4px
 
+// library recos with invite need additional spacing between the control header and the invite banner
+.kf-reco-with-invite
+  .kf-reco-top
+    padding-bottom: 10px
+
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/angular/src/recos/recos.styl
```diff
@@ -95,6 +95,11 @@
   .kf-keep-lib-header
     padding-top: 4px
 
+// library recos with invite need additional spacing between the control header and the invite banner
+.kf-reco-with-invite
+  .kf-reco-top
+    padding-bottom: 10px
```
2is10, line 101: Looks like there’s also padding set for `.kf-reco-top` around lines 160-170. Best to keep all of those close together, so it’s clearer what the adjustment is actually doing, and so they can be maintained together.

This same rule can be achieved this way down there:

``` css
.kf-reco-top
  padding: ...

  .kf-reco-with-invite &
    padding-bottom: 10px
```

joshm1, line 101: good idea! :+1:

2is10:
:+1:

PR #13064, opened 2015-01-12
keep keptAt is not optional

eishay:


eishay:
@andrewconner @stkem

kifi-backend/modules/shoebox/marketing
```diff
@@ -1 +1 @@
-Subproject commit 56906cc7c0733af9e06e415f3b7d88a3c1b894c6
+Subproject commit 17af4006a9f9347561fa4b118748a12c65c63afa
```
2is10: unintentional?

2is10:
Extension still isn’t sending bookmark times, but I could add it today.
https://app.asana.com/0/2456298849186/20632537806440

kifi-backend/modules/shoebox/marketing
eishay: right

eishay:
@2is10 that will be nice

stkem:
:+1:

PR #13066, opened 2015-01-12
lib recos for everybody!

joshm1:
@stkem


stkem:
lgtm! I assume you checked with the mobile guys to make sure none of the versions currently in the wild will choke in this?

joshm1:
@stkem yeah, they're aware the /m/2 version was added last week and they've tested it; /m/1 is unchanged

PR #13068, opened 2015-01-12
Sampling

ymatsuda:
@yingjieMiao @leogrim @stkem @eishay 
- I think I fixed a bug which caused high CPU
- added a switch to enable the sampler (the default is using ProbabilityDensity)
- renamed SegmentedSampler to StratifiedSampler

If it is ok to use a stratified sampling with a single sample per stratum, we don't need to allocate `Array[Double]` at all (and simplifies the code a lot).
What do you think? Shall we go extreme?


yingjieMiao:
lgtm. 

'single sample per stratum': I am not sure I get this part.

PR #13076, opened 2015-01-12
adaptive font size for library card titles

2is10:
@yipingliao cc @joshm1

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.styl
+++ b/kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.styl
@@ -134,6 +134,7 @@
   left: 0
   right: 0
   bottom: 0
+  user-select: none
 
 .kf-upl-card-icon
   position: absolute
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.styl
```diff
@@ -341,48 +334,3 @@ follower-pic-diameter = 23px
...
-    height: 200px
-    margin-left: -100px
-
-  .kf-upl-card-foot
-    height: 100px
-
-  .kf-upl-updated
-    line-height: 82px
-    font-size: 28px
-
```
2is10, line 388: Removing these mobile styles for now, rather than updating them. We’ll need to use scaling.

2is10:
Merging to get this live, since recommended library cards have already launched. Comments still appreciated.

andrewconner:
Awesome, nice

andrewconner:
Really nice utility. You missed the opportunity for a ki_fit_i pun, however.

joshm1:
Very nice!

PR #13078, opened 2015-01-13
removing library card background image blur on hover

2is10:
(Mark’s idea)

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/smallLibraryCard.styl
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/smallLibraryCard.styl
@@ -93,10 +93,14 @@
   left: -3px
   right: -3px
   bottom: -3px
+  background-color: #eee
   background-size: cover
   -webkit-filter: blur(3px)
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/smallLibraryCard.styl
```diff
@@ -93,11 +93,15 @@
   left: -3px
   right: -3px
   bottom: -3px
+  background-color: #eee
```
2is10, line 96: for library images that have transparency

kifi-backend/modules/shoebox/angular/src/libraries/smallLibraryCard.styl
```diff
@@ -93,11 +93,15 @@
...
   background-size: cover
   -webkit-filter: blur(3px)
-  -moz-filter: blur(3px)
```
2is10, line 98: There is no `-moz-filter` property. Firefox is adding support for `filter` in Firefox 35.

PR #13081, opened 2015-01-13
3.3.32: sending bookmark add times with browser bookmark uploads

2is10:
@stkem @andrewconner


packrat/adapters/firefox/lib/bookmarks.js
```diff
@@ -18,7 +18,7 @@ function runQuery(query) {
   result.containerOpen = true;
   for (let i = 0, n = result.childCount; i < n; ++i) {
     let node = result.getChild(i);
-    resultArray.push({id: node.itemId, title: node.title, url: node.uri});
+    resultArray.push({id: node.itemId, title: node.title, url: node.uri, addedAt: Math.round(node.dateAdded / 1000)});
```
2is10, line 21: Note: This value is sometimes `0` (e.g. for browser default bookmarks).

PR #13083, opened 2015-01-13
fixing z-index issue with library hovercard over profile library cards

2is10:
@yipingliao cc @ahsu1230

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/layout/sidebar/sidebar.styl
+++ b/kifi-backend/modules/shoebox/angular/src/layout/sidebar/sidebar.styl
@@ -2,10 +2,7 @@
   position: fixed
   width: 270px
   padding: 0 0 0 20px
-  z-index: 1
-
-  .kf-tag-count
-    float: right
+  z-index: 2
 
 .kf-sidebar-inner
   margin-top: 15px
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/angular/src/layout/sidebar/sidebar.styl
```diff
@@ -2,10 +2,7 @@
   position: fixed
   width: 270px
   padding: 0 0 0 20px
-  z-index: 1
-
-  .kf-tag-count
-    float: right
+  z-index: 2
```
2is10, line 5: The user picture in the user profile library cards also has `z-index: 1` so bumping this up to 2.

kifi-backend/modules/shoebox/angular/src/layout/sidebar/sidebar.styl
```diff
@@ -2,10 +2,7 @@
   position: fixed
   width: 270px
   padding: 0 0 0 20px
-  z-index: 1
-
-  .kf-tag-count
-    float: right
```
2is10, line 8: As far as I can tell, the left nav no longer ever contains a `kf-tag-count`.

kifi-backend/modules/shoebox/angular/src/layout/sidebar/sidebar.styl
```diff
@@ -20,14 +17,11 @@
...
+    transition-duration: sidebarTransitionDuration
 
-  .kf-sidebar-inactive .kf-sidebar
-    left: -135px
-    z-index: -1
+    .kf-sidebar-inactive &
+      transform: translate(-50%)
+      opacity: 0
+      visibility: hidden
+      transition-delay: 0, 0, sidebarTransitionDuration
```
2is10, line 27: `opacity` and `visibility` make for a better hiding transition than `z-index`.

2is10:
There’s still some funkiness where the first library heading jumps up/down when this sidebar transition is triggered. I don’t know why it has `fixed` position, but I have a feeling it’s a rabbit hole, so leaving alone for now.

yipingliao:
:+1:

kifi-backend/modules/shoebox/angular/src/layout/sidebar/sidebar.styl
ahsu1230, line 27: :+1:

PR #13084, opened 2015-01-13
Library and User Index modifications

leogrim:
@ymatsuda @yingjieMiao 
- renaming fields in UserIndex to be consistent with other indices and queries
- adding LibraryKind to LibraryIndex to be able to make explicit exclusions / boosts both in library search and augmentation (e.g. excluding Main and Private from Library search results, per Mark's request) 

@andrewconner @stkem if we can have special library kinds for imports, search can now explicitly filter them out.


yingjieMiao:
:+1:

PR #13085, opened 2015-01-13
inserting line breaks in titles more conservatively

2is10:
Our line balancing algorithm doesn’t consider all possible line break points (e.g. hyphens). So now if we can’t find a good place to insert line breaks, we don’t. We assume the browser already found an optimal solution.

Also measuring hypothetical line lengths more accurately—no longer assuming uniform width of spaces.

@ahsu1230


ahsu1230:
this is really cool, thanks!

PR #13092, opened 2015-01-13
Dropping NormalizedURI from ScraperURISummaryCommanderImpl

andrewconner:
`ScraperURISummaryCommander` uses NormalizedURI a ton, only needing `id`, `externalId`, and `url`. Since these parts are a bit hairy, doing them separately.

The risk here is if a client is ever currently sending a NormalizedURI without an id. We'll get serialization issues if that happens.

@stkem

### DIFF
--- a/kifi-backend/modules/scraper/app/com/keepit/commanders/ScraperURISummaryCommander.scala
+++ b/kifi-backend/modules/scraper/app/com/keepit/commanders/ScraperURISummaryCommander.scala
@@ -1,10 +1,12 @@
 package com.keepit.commanders
 
+import com.keepit.common.db.{ ExternalId, Id }
 import com.keepit.common.performance._
 
 import java.awt.image.BufferedImage
 
 import com.keepit.common.logging.Logging
+import com.kifi.macros.json
 
 import scala.concurrent.Future
 import scala.util.{ Failure, Success }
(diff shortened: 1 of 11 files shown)


andrewconner:
Mind going over this? I'd like to merge/deploy to watch it, though.

kifi-backend/modules/shoebox/test/com/keepit/commanders/URISummaryCommanderTest.scala
```diff
@@ -58,16 +56,13 @@ class URISummaryCommanderTestImageFetcher extends ImageFetcher {
 }
 
 case class URISummaryCommanderTestS3URIImageStore() extends S3URIImageStore {
-  def storeImage(info: ImageInfo, rawImage: BufferedImage, nUri: NormalizedURI): Try[(String, Int)] = Success((FakeS3URIImageStore.placeholderImageURL, FakeS3URIImageStore.placeholderSize))
-  def getDefaultScreenshotURL(nUri: NormalizedURI): Option[String] = Some(FakeS3URIImageStore.placeholderScreenshotURL)
```
andrewconner, line 62: So much of this wasn't even being used.

kifi-backend/modules/scraper/app/com/keepit/commanders/ScraperURISummaryCommander.scala
```diff
@@ -1,10 +1,12 @@
 package com.keepit.commanders
```
andrewconner, line 1: todo: rename to `com.keepit.scraper.commanders`

yingjieMiao:
lgtm

stkem:
lgtm

PR #13093, opened 2015-01-13
remove old uri recos

ymatsuda:
It is annoying to see old recos keep coming back. This change adds a code that remove old uri recos whose update_at is older than 30 days.

@stkem @yingjieMiao @joshm1


stkem:
:+1:

kifi-backend/modules/curator/app/com/keepit/curator/model/UriRecommendationRepo.scala
```diff
@@ -179,5 +181,11 @@ class UriRecommendationRepoImpl @Inject() (
   def deleteCache(model: UriRecommendation)(implicit session: RSession): Unit = {}
 
   def invalidateCache(model: UriRecommendation)(implicit session: RSession): Unit = {}
+
+  // for test
```
yingjieMiao, line 185: Looks like @leogrim wrote a FakeClock which allow us to manually manipulate timestamps in test.

Might worth consider using that, instead of have a public method in the repo just for that.

PR #13095, opened 2015-01-13
User profile misc issues

yipingliao:
@ahsu1230 

A couple of misc. issues. Please see commits for detailed description. Thanks!

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.tpl.html
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.tpl.html
@@ -70,7 +70,7 @@
                 <span class="kf-keep-lib-stats-number">{{library.numKeeps}}</span>
               </div>
 
-              <div ng-if="isUserCreatedLibrary()" class="kf-keep-lib-keeps-followers-stats">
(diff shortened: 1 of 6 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.tpl.html
```diff
@@ -70,7 +70,7 @@
                 <span class="kf-keep-lib-stats-number">{{library.numKeeps}}</span>
               </div>
 
-              <div ng-if="isUserCreatedLibrary()" class="kf-keep-lib-keeps-followers-stats">
+              <div ng-if="isUserCreatedLibrary()" class="kf-keep-lib-keeps-followers-stats kf-keep-lib-followers-stats" ng-click="showFollowers()">
```
2is10, line 73: For accessibility, everything that’s clickable should be a link, a button, or a form control.
[This article](http://www.karlgroves.com/2013/05/14/links-are-not-buttons-neither-are-divs-and-spans/) does a great job explaining why. (I wouldn’t have come down quite so hard against in-page links as he does, though.)

PR #13096, opened 2015-01-13
before sending email, check for alt attribute in HTML on image tags

joshm1:
@eishay cc @stkem 
- updates existing img elements to have alt attributes, there weren't many that didn't have them from what I found (mostly spacer/rounded-corner images)
- auto-adds alt="" attribute if it's missing
- airbrake to notify us so we can correct

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/SendgridCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/SendgridCommander.scala
@@ -208,7 +208,7 @@ class SendgridCommander @Inject() (
   private def addTrackingFromUrlParam(url: String, contextBuilder: HeimdalContextBuilder): Unit = {
     try {
       val uri = new java.net.URI(url)
-      val queryOpt = Option(uri.getQuery)
+      val queryOpt = Option(uri.getQuery) map { s => s.replace("&amp;", "&") }
       val keyValuesOpt = queryOpt map (_.split('&').map(str => str.splitAt(str.indexOf('='))).toSeq)
       val datParamOpt = keyValuesOpt flatMap { keyValues =>
         keyValues.find(_._1 == EmailTrackingParam.paramName) map (_._2.drop(1))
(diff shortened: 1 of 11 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/SendgridCommander.scala
```diff
@@ -208,7 +208,7 @@ class SendgridCommander @Inject() (
   private def addTrackingFromUrlParam(url: String, contextBuilder: HeimdalContextBuilder): Unit = {
     try {
       val uri = new java.net.URI(url)
-      val queryOpt = Option(uri.getQuery)
+      val queryOpt = Option(uri.getQuery) map { s => s.replace("&amp;", "&") }
```
2is10, line 211: A comment for this would be really helpful, I think, since many people find escaping confusing. Maybe:

``` scala
// Our observations indicate that ampersands are mistakenly double HTML-escaped
// in URLs in HTML attribute values much more often than the sequence &amp;
// occurs intentionally in URLs, so we always assume it's there by accident.
```

Caveat: I don’t really know why you added this. If this code only runs over our own URLs from our HTML emails, then I think we should fix whatever problem this new code is intended to address in a different way. For example, HTML-unescaping should be done at the time the URL is read from the HTML attribute rather than being done here.

joshm1, line 211: the problem it's solving is that old URLs in emails were using the unescaped `&` and new use `&amp;`.  When I added the HTML decorator to add the alt attribute, it escapes existing & characters in URLs.  I'm not sure if the URL sendgrid sends us escapes these ampersands or not (I'll find out soon), so that's why I added support for

2is10, line 211: I see. In that case, I think `SendgridController.parseEvent` would be the place to undo SendGrid’s erroneous HTML escaping. That way the source of the problem that’s being solved would be more apparent and `event.url` would be correct everywhere it’s used instead of just here.

PR #13098, opened 2015-01-14
blocking render to fit the first few library card titles

2is10:
This way the first few cards already have their title showing at the correct size when they are first drawn. The rest are either off-screen or not the user’s focal point, so it’s okay if their titles appear a bit later—we don’t block on their size calculation.
@ahsu1230

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/common/directives/fit.js
+++ b/kifi-backend/modules/shoebox/angular/src/common/directives/fit.js
@@ -158,20 +158,28 @@ angular.module('kifi')
         }
       }
 
+      var block = +attr.block || 0;
       var sizes = {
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/angular/src/common/directives/fit.js
```diff
@@ -158,20 +158,28 @@ angular.module('kifi')
...
           if (element.attr('kf-fit') !== 'false') {
             fit(element);
           }
           element.css('visibility', '');
-        });
+        }
+
+        if (block > 0) {
+          block--;
+          scope.$evalAsync(fitAndShow);  // blocking on binding/interpolation for first few only
```
2is10, line 179: Thanks for exposing me to `$evalAsync` Aaron!

PR #13100, opened 2015-01-14
For mobile web: scale up user profile library cards; make user profile header taller.

yipingliao:
@2is10 Thanks for the help! Screenshot attached.
screen shot 2015-01-13 at 5 58 36 pm


2is10:
:+1:

PR #13101, opened 2015-01-14
better respond to friend request

ahsu1230:
appears above the header now with a click Accept or Decline


kifi-backend/modules/shoebox/angular/src/userProfile/userProfileHeader.tpl.html
```diff
@@ -1,3 +1,13 @@
+<div class="kf-user-profile-friend-request-header" ng-if="connectionWithUser==='request_received'">
+  <span class="kf-user-profile-friend-respond accept" ng-click="acceptFriendRequest()">
+    Accept
+  </span>
+  <span class="kf-user-profile-friend-respond decline" ng-click="ignoreFriendRequest()">
+    Decline
+  </span>
```
2is10, line 7: `<button>` would be better ([why](http://www.karlgroves.com/2013/05/14/links-are-not-buttons-neither-are-divs-and-spans/))

2is10:
That was quick. Thanks. :+1:

PR #13102, opened 2015-01-14
prevent followers list from breaking after modifying profile library

ahsu1230:
After modifying a library, the followers list stops including the pictureUrl of each user, so no follower pictures show up


kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
```diff
@@ -112,7 +112,9 @@ angular.module('kifi')
           library: library,
           returnAction: function () {
             libraryService.getLibraryById(library.id, true).then(function (data) {
-              _.assign(library, data.library);
+              var followersList = library.followers; // contains userPicture
+              _.assign(library, data.library); // replaces all new data (data.library.followers does not have userPicture)
+              library.followers = followersList; // reset userPicture
```
ahsu1230: there might be a more clever way to do this. We want to reset all the old library data using `_.assign` except the `followers` field

2is10:
``` js
_.assign(library, data.library, {followers: library.followers});
```

ahsu1230: ohh I see thanks.

PR #13103, opened 2015-01-14
fix sticky follow button on desktop

joshm1:
@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/common/directives/sticky.js
+++ b/kifi-backend/modules/shoebox/angular/src/common/directives/sticky.js
@@ -3,8 +3,8 @@
 angular.module('kifi')
 
 .directive('kfSticky', [
-  '$window', '$timeout',
-  function ($window, $timeout) {
+  '$window', '$timeout', 'platformService',
+  function ($window, $timeout, platformService) {
     return {
       restrict: 'A',
       scope: {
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/angular/src/common/directives/sticky.js
```diff
@@ -58,7 +58,16 @@ angular.module('kifi')
...
             height = element.height() + borderTopWidth + borderBottomWidth;
-          });
+          }
+
+          // This timeout is needed for mobile Safari.
+          if (isMobile) {
+            $timeout(update);
+          } else {
+            // on desktop, the offsetTop calculation gradually increases in value every time updateProperties is called
+            // and becomes very different from what the actual offset is
```
joshm1, line 68: @2is10 I know you'll need to see it to understand what I'm asking, but any idea why this timeout is causing the `element.offset().top` value to "grow" and get far from the initial value every time `updateProperties` is called, while the same value not inside the timeout is the same correct value every time.

2is10, line 68: Not sure, but I’d recommend using `element[0].getBoundingClientRect()` instead of relying on jQuery’s `.offset()`. Then you also shouldn’t have to measure borders and margins. I’m surprised how complex this directive is. I think it could be much simpler. Maybe we could pair on it a bit tomorrow.

joshm1, line 68: that'd be great, I'm going to go ahead and merge this to fix the desktop issue and maybe we can pair later. thanks!

PR #13106, opened 2015-01-14
not allow double impersonation

eishay:
don't want to get into possible security holes, so just to be safe, you need to un-impersonate and then impersonate again to someone else


eishay:
@andrewconner

andrewconner:
Works for me.

PR #13107, opened 2015-01-14
detailed time logging

yingjieMiao:
@stkem @joshm1 @eishay @ymatsuda 

I noticed some slowness in relatedLibCommander.
I had tracking on the cortex side, it took about 50 ms (jump up a bit after my deployment on Tuesday)
The tracking on shoebox's controller shows about 200ms to 300ms. 

I am breaking down timing a bit to see what is the slow part.


joshm1:
thanks!

PR #13110, opened 2015-01-14
Fix broken search icons.

yipingliao:
@ahsu1230

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/keep/keepToLibraryWidget.styl
+++ b/kifi-backend/modules/shoebox/angular/src/keep/keepToLibraryWidget.styl
@@ -84,7 +84,7 @@ keep-to-library-lib-icon =
     width: 13px
     height: 14px
     background-image: url('/img/sprites.png')
-    background-position: -68px -88px
+    background-position: -110px -29px
     background-size: 132px auto
 
   .keep-to-library-search-input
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/angular/src/keep/keepToLibraryWidget.styl
```diff
@@ -84,7 +84,7 @@ keep-to-library-lib-icon =
     width: 13px
     height: 14px
     background-image: url('/img/sprites.png')
-    background-position: -68px -88px
+    background-position: -110px -29px
```
yipingliao, line 87: I cannot use `sprite2x` here because of scaling. Hand-coding this but we need to extend `sprite2x` to do scaling.

kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.tpl.html
```diff
@@ -49,7 +49,7 @@
               Add a cover photo
             </a>
             <div ng-if="!recommendation && isUserLoggedOut" class="kf-keep-lib-search">
-              <span class="kf-keep-lib-search-icon"></span>
+              <span class="kf-keep-lib-search-icon sprite-search-icon"></span>
```
yipingliao, line 52: So much cleaner now, yay!

ahsu1230:
:+1: :smile_cat:

kifi-backend/modules/shoebox/angular/src/keep/keepToLibraryWidget.styl
yipingliao, line 87: @ahsu1230 Per our discussion, a better solution is to get an SVG version of this asset.

PR #13112, opened 2015-01-14
User profiles misc bugs

yipingliao:
@ahsu1230 @2is10 

Fixes two issues:
- Curator images in library search chips are now clickable
- When a user deletes a library from the manage-library modal she opened on the user profiles page, the page will stay on the user profile page (it previously redirected to home) and the deleted library is removed from the view.

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/layout/header/header.tpl.html
+++ b/kifi-backend/modules/shoebox/angular/src/layout/header/header.tpl.html
@@ -11,7 +11,10 @@
       <div class="kf-query-lib-container" ng-class="{empty: isEmpty()}">
         <div class="kf-query-lib-icon" ng-class="{'focused': isFocused}"></div>
 
-        <img class="kf-search-bar-lib-owner-img" ng-show="search.showName" ng-src="{{library.owner.picUrl || 'https://www.kifi.com/assets/img/ghost.200.png'}}" title="{{library.owner.firstName + ' ' + library.owner.lastName}}"></img>
+        <a ng-if="inUserProfileBeta" href="{{userProfileUrl}}">
+          <img class="kf-search-bar-lib-owner-img" ng-show="search.showName" ng-src="{{library.owner.picUrl || 'https://www.kifi.com/assets/img/ghost.200.png'}}" title="{{library.owner.firstName + ' ' + library.owner.lastName}}"></img>
+        </a>
+        <img ng-if="!inUserProfileBeta" class="kf-search-bar-lib-owner-img" ng-show="search.showName" ng-src="{{library.owner.picUrl || 'https://www.kifi.com/assets/img/ghost.200.png'}}" title="{{library.owner.firstName + ' ' + library.owner.lastName}}"></img>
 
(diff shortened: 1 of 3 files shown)


ahsu1230:
lgtm! thanks!

kifi-backend/modules/shoebox/angular/src/layout/header/header.tpl.html
```diff
@@ -11,7 +11,10 @@
       <div class="kf-query-lib-container" ng-class="{empty: isEmpty()}">
         <div class="kf-query-lib-icon" ng-class="{'focused': isFocused}"></div>
 
-        <img class="kf-search-bar-lib-owner-img" ng-show="search.showName" ng-src="{{library.owner.picUrl || 'https://www.kifi.com/assets/img/ghost.200.png'}}" title="{{library.owner.firstName + ' ' + library.owner.lastName}}"></img>
+        <a ng-if="inUserProfileBeta" href="{{userProfileUrl}}">
+          <img class="kf-search-bar-lib-owner-img" ng-show="search.showName" ng-src="{{library.owner.picUrl || 'https://www.kifi.com/assets/img/ghost.200.png'}}" title="{{library.owner.firstName + ' ' + library.owner.lastName}}"></img>
+        </a>
+        <img ng-if="!inUserProfileBeta" class="kf-search-bar-lib-owner-img" ng-show="search.showName" ng-src="{{library.owner.picUrl || 'https://www.kifi.com/assets/img/ghost.200.png'}}" title="{{library.owner.firstName + ' ' + library.owner.lastName}}"></img>
```
2is10, line 17: The `` tag has no corresponding close tag `</img>` in HTML. It’s actually forbidden.

kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.js
```diff
@@ -124,8 +124,13 @@ angular.module('kifi')
           submitting = true;
           libraryService.deleteLibrary(scope.library.id).then(function () {
             $rootScope.$emit('librarySummariesChanged');
+            $rootScope.$emit('libraryDeleted', scope.library.id);
             scope.close();
-            $location.path('/');
+
+            // If we were on the deleted library's page, return to the homepage.
+            if ($state.is('library.keeps')) {
```
2is10, line 131: Would it be easy to also verify that `$state` corresponds to the specific library deleted? I can imagine clicking Delete, navigating to a related library, then being taken away from it when deletion of the first library completes.

2is10:
Nice! :+1:

kifi-backend/modules/shoebox/angular/src/layout/header/header.tpl.html
yipingliao, line 17: Autocomplete? Will fix. Thanks!

kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.js
yipingliao, line 131: Will do. Thanks!

PR #13113, opened 2015-01-14
removing recos affected by uri renormalization

ymatsuda:
@stkem @yingjieMiao @eishay 
- This adds a logic to remove recommendations when a uri renormalization is propagated to curator
- Those garbage recos already in curator db is not removed. They will be removed by regular clean up process eventually.


yingjieMiao:
lgtm. sorry, just saw this.

stkem:
:+1:

PR #13114, opened 2015-01-14
adding following field to LibraryCardInfo

ahsu1230:
A `following` field appears only when you are signed in & viewing someone else's library. It is true or false based on whether you have a membership to a library.

So anonymous users, and when viewing your own profile, this field is irrelevant and won't appear.

@andrewconner @2is10


kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -1165,6 +1165,7 @@ class LibraryCommander @Inject() (
               numFollowers = info.numFollowers,
               followers = Seq.empty,
               lastKept = info.lastKept,
+              following = None,
```
2is10, line 1168: If the viewer is logged in, we may want to populate this field and show a Follow button in the future. This is fine for now.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -1303,12 +1309,18 @@ class LibraryCommander @Inject() (
       } else {
         (0, Seq.empty)
       }
-      createLibraryCardInfo(lib, image, ownersWFS(lib.ownerId), numKeeps, numFollowers, followersSample)
+
+      val isFollowing = if (isUserViewingOtherProfile) {
+        Some(libraryMembershipRepo.getWithLibraryIdAndUserId(lib.id.get, viewer.get.id.get).isDefined)
```
2is10, line 1314: How about using `getWithLibraryIdsAndUserId` instead of making these 12 database queries?

2is10:
:+1:

PR #13115, opened 2015-01-14
indicating whether the user tried to join a library that it is already f...

eishay:
...ollowing


eishay:
@yipingliao @ahsu1230 @andrewconner

ahsu1230:
lgtm! thanks!

yipingliao:
Thanks!

eishay:
@yipingliao since it will add a field and not modify others, it seems to be backward compatible.

PR #13116, opened 2015-01-14
Add ability to see all followers from links in user profile library cards.

yipingliao:
@2is10 @ahsu1230 

Both number of keeps and number of followers are now clickable. Clicking the former takes you to the library; clicking the latter opens up the modal. Also, when there are more than 3 followers, clicking all the "+N" text opens up the modal.

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
+++ b/kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
@@ -158,8 +158,10 @@ angular.module('kifi')
 
 
 .controller('UserProfileLibrariesCtrl', [
-  '$scope', '$rootScope', '$state', '$stateParams', 'routeService', 'keepWhoService', 'profileService', 'userProfileActionService',
-  function ($scope, $rootScope, $state, $stateParams, routeService, keepWhoService, profileService, userProfileActionService) {
+  '$scope', '$rootScope', '$state', '$stateParams',
+  'routeService', 'keepWhoService', 'profileService', 'userProfileActionService', 'modalService',
+  function ($scope, $rootScope, $state, $stateParams,
+    routeService, keepWhoService, profileService, userProfileActionService, modalService) {
     var username = $stateParams.username;
     var fetchPageSize = 12;
     var fetchPageNumber = 0;
@@ -237,6 +239,15 @@ angular.module('kifi')
       });
       return lib;
     }
+
+    $scope.openFollowersList = function (lib) {
+      modalService.open({
+        template: 'libraries/libraryFollowersModal.tpl.html',
+        modalData: {
+          library: lib
+        }
+      });
+    };
   }
 ])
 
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
```diff
@@ -237,6 +239,15 @@ angular.module('kifi')
       });
       return lib;
     }
+
+    $scope.openFollowersList = function (lib) {
```
yipingliao, line 242: @ahsu1230 What do you think about the placement of this method (`$scope.openFollowersList`)? I put it in `UserProfileLibrariesCtrl` because the UI is in the library card, but I saw that you had previously put `$scope.openModifyLibrary` on the controller one level higher.

ahsu1230, line 242: now that I think about it, I think it is better in `UserProfileLibrariesCtrl` since you only need that function per library. `$scope.openModifyLibrary` should probably reside in the same place as `$scope.openFollowersList`

yipingliao, line 242: I'll move it, thanks!

ahsu1230, line 242: thanks!

kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.styl
```diff
@@ -406,6 +409,16 @@ follower-pic-diameter = 23px
   vertical-align: middle
   color: rgba(255,255,255,.7)
 
+.kf-upl-followers-number
+  &:link
+  &:visited
+  &:hover
+  &:active
```
2is10, line 416: why all the pseudoclasses?

kifi-backend/modules/shoebox/angular/src/userProfile/userProfileLibrariesList.tpl.html
```diff
@@ -28,7 +32,7 @@
           <ng-pluralize class="kf-upl-all-followers" count="lib.numFollowers" when="{'1':'follows this library','other':'are following'}"></ng-pluralize>
         </span>
         <span ng-if="lib.numFollowers > lib.followers.length">
-          + {{lib.numFollowers - lib.followers.length}} are following
+          <a href="javascript:" ng-click="openFollowersList(lib)" class="kf-upl-followers-number">+ {{lib.numFollowers - lib.followers.length}}</a> are following
```
2is10, line 35: When making changes to this template, it’s a good idea to think about which ones make sense for the related library card too. These probably do.

2is10:
:+1:

kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.styl
yipingliao, line 416: My bad. Fixing.

kifi-backend/modules/shoebox/angular/src/userProfile/userProfileLibrariesList.tpl.html
yipingliao, line 35: Will do. Thanks!

PR #13123, opened 2015-01-14
smoothing of library induced interest score

yingjieMiao:
@ymatsuda @stkem @joshm1 

we used to use max of scores, this creates a "single point failure" if you followed some weird library. That single library could ruin the whole uri recommendations (happened to Jeremy).

The approach here is more 'ensemble'.


ymatsuda:
:+1:

PR #13124, opened 2015-01-15
changing library colors and accepting color names

2is10:
@ahsu1230


kifi-backend/modules/common/app/com/keepit/model/Library.scala
```diff
@@ -247,36 +247,55 @@ object BasicLibraryStatistics {
   implicit val format = Json.format[BasicLibraryStatistics]
 }
 
-sealed abstract class HexColor(val hex: String)
-object HexColor {
+sealed abstract class LibraryColor(val hex: String)
+object LibraryColor {
```
2is10, line 251: renaming to `LibraryColor` to better reflect that it cannot represent an arbitrary hex color

kifi-backend/modules/common/app/com/keepit/model/Library.scala
```diff
@@ -247,36 +247,55 @@ object BasicLibraryStatistics {
...
       case PURPLE.hex => PURPLE
+
+      // TODO: Remove mappings below when migration from old colors is complete.
+      case "#3975bf" => BLUE
+      case "#fab200" => SKY_BLUE // was ORANGE_YELLOW
+      case "#2ec89a" => GREEN
+      case "#ff9430" => ORANGE
+      case "#e35957" => RED
+      case "#c764a2" => MAGENTA
+      case "#955cb4" => PURPLE
```
2is10, line 289: I’ll prepare a database migration endpoint in a separate commit.

ahsu1230, line 289: nice!

ahsu1230:
:+1:
 thanks

PR #13125, opened 2015-01-15
Update code per pull request comments

yipingliao:
@2is10


2is10:
:+1:

yipingliao:
@2is10 I just remembered why I couldn't combine the two elements (and also cannot combine the two classes) - I want to style the text to underline on hover but I don't want to style the bullet between the number of keeps and the number of followers to underline on hover, so having two elements/classes allows me to style them separately.

2is10:
Thanks.

PR #13126, opened 2015-01-15
an endpoint to migrate library colors

2is10:
@ahsu1230

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/model/LibraryRepo.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/model/LibraryRepo.scala
@@ -38,6 +38,7 @@ trait LibraryRepo extends Repo[Library] with SeqNumberFunction[Library] {
   def pagePublished(page: Paginator)(implicit session: RSession): Seq[Library]
   def countPublished(implicit session: RSession): Int
   def filterPublishedByMemberCount(minCount: Int, limit: Int = 100)(implicit session: RSession): Seq[Library]
+  def getByColor(hexColor: String)(implicit session: RSession): Seq[Library]
 }
 
 @Singleton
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/shoebox/app/com/keepit/model/LibraryRepo.scala
```diff
@@ -295,6 +296,11 @@ class LibraryRepoImpl @Inject() (
       t <- rows if t.visibility === (LibraryVisibility.PUBLISHED: LibraryVisibility) && t.state === LibraryStates.ACTIVE && t.memberCount >= minCount
     } yield t).sortBy(_.updatedAt.desc).take(limit).list
   }
+
+  def getByColor(hexColor: String)(implicit session: RSession): Seq[Library] = {
+    import StaticQuery.interpolation
+    sql"select * from library where color=$hexColor".as[Library].list
+  }
```
2is10, line 303: There are only roughly 600 rows per color, so no need to batch in groups of 500.

2is10, line 303: Using a `String` instead of `LibraryColor` because old colors cannot be represented with a `LibraryColor`.

kifi-backend/modules/shoebox/test/com/keepit/controllers/website/LibraryControllerTest.scala
```diff
@@ -174,27 +174,28 @@ class LibraryControllerTest extends Specification with ShoeboxTestInjector {
...
+                  "pictureName":"${basicUser1.pictureName}",
+                  "username":"${basicUser1.username.value}"
+                  },
+                "numKeeps":0,
+                "numFollowers":0,
+                "kind":"user_created"
+              },
+              "listed": true
+            }
+           """)
```
ahsu1230, line 198: thanks for removing the |

ahsu1230:
:+1:

PR #13127, opened 2015-01-15
Fix Kifi curator profile pages.

yipingliao:
@ahsu1230 Please check if I added the `ng-if` at the correct place.

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/userProfile/userProfileHeader.tpl.html
+++ b/kifi-backend/modules/shoebox/angular/src/userProfile/userProfileHeader.tpl.html
@@ -15,7 +15,7 @@
         <img class="kf-user-profile-picture" ng-src="{{profile.picUrl || 'https://www.kifi.com/assets/img/ghost.200.png'}}" alt="{{profile.firstName}} {{profile.lastName}}" itemprop="photo">
       </a>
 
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/angular/src/userProfile/userProfileLibraries.tpl.html
```diff
@@ -1,4 +1,4 @@
-<div ng-if="!isKifiCurator" class="kf-upl-nav-container">
```
yipingliao, line 1: We no longer hide the libraries nav, per yesterday's discussion.

kifi-backend/modules/shoebox/angular/src/userProfile/userProfileHeader.tpl.html
```diff
@@ -15,7 +15,7 @@
         <img class="kf-user-profile-picture" ng-src="{{profile.picUrl || 'https://www.kifi.com/assets/img/ghost.200.png'}}" alt="{{profile.firstName}} {{profile.lastName}}" itemprop="photo">
       </a>
 
-      <div ng-switch="connectionWithUser" class="kf-user-profile-available-action">
+      <div ng-if="!isKifiCurator" ng-switch="connectionWithUser" class="kf-user-profile-available-action">
```
ahsu1230, line 18: :+1:

PR #13131, opened 2015-01-15
twitter signup2

ahsu1230:
@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/core/AuthHelper.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/core/AuthHelper.scala
@@ -225,7 +225,7 @@ class AuthHelper @Inject() (
       }),
       "firstName" -> text, // todo(ray/andrew): revisit non-empty requirement for twitter
       "lastName" -> text,
-      "password" -> text.verifying("password_too_short", pw => AuthHelper.validatePwd(pw.toCharArray)),
+      "password" -> text.verifying("password_too_short", pw => AuthHelper.validatePwd(pw.toCharArray)), // todo (aaron): for twitter, there's no password. Should client generate random password? or do we do logic here? Cuz then form gets messed up
       "picToken" -> optional(text),
       "picHeight" -> optional(number),
       "picWidth" -> optional(number),
--- a/kifi-backend/modules/shoebox/app/views/auth/authGrey.scala.html
+++ b/kifi-backend/modules/shoebox/app/views/auth/authGrey.scala.html
@@ -101,8 +101,8 @@
       <div class="social-text-info">
         <div class="social-name">@firstName @lastName</div>
 
-        @if(network == Some(SocialNetworkType("twitter"))) {
-          <input class="social-email twitter-form" type="email" name="signup-2-email" value="@emailAddress" placeholder="Your email address">
+        @if(emailAddress == "") {
+          <input class="social-email" type="email" name="signup-2-email" value="@emailAddress" placeholder="Your email address">
         } else {
           <input class="social-email" type="email" name="signup-2-email" value="@emailAddress" placeholder="Your email address" disabled>
           <a class="social-change-email" href="javascript:" data-track-action="clickchangeEmail">Change email</a>
@@ -110,15 +110,13 @@
 
       </div>
     </div>
-    @if(network != Some(SocialNetworkType("twitter"))) {
-      <div class="form-creds">
-        <label class="password-label signup-2-label">
-          Create a password
-          <input class="form-password" type="password" name="signup-2-password" autofocus>
-          <span class="form-password-tip">at least 7 characters</span>
-        </label>
-      </div>
-    }
+    <div class="form-creds">
+      <label class="password-label signup-2-label">
+        Create a password
+        <input class="form-password" type="password" name="signup-2-password" autofocus>
+        <span class="form-password-tip">at least 7 characters</span>
+      </label>
+    </div>
     <button class="form-submit form-submit-sexy alt-button alt-blue" data-track-action="clicksignUpButton">
       <span class="progress"></span>
       <span class="text">Sign me up</span>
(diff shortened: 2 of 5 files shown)


kifi-backend/modules/shoebox/app/com/keepit/controllers/core/AuthHelper.scala
```diff
@@ -225,7 +225,7 @@ class AuthHelper @Inject() (
       }),
       "firstName" -> text, // todo(ray/andrew): revisit non-empty requirement for twitter
       "lastName" -> text,
-      "password" -> text.verifying("password_too_short", pw => AuthHelper.validatePwd(pw.toCharArray)),
+      "password" -> text.verifying("password_too_short", pw => AuthHelper.validatePwd(pw.toCharArray)), // todo (aaron): for twitter, there's no password. Should client generate random password? or do we do logic here? Cuz then form gets messed up
```
ahsu1230, line 228: should the client generate a fake password? or should we make the password here optional?
If we make it optional, it makes things a bit weird because then we're trusting the client to check for valid password when signing in from LinkedIn or Facebook.

I think generating a password for twitter is good for now because eventually we'll move to FB/LI also having a random password (in which case then we can just make `password` for `socialFinalizeAccountForm` optional)

andrewconner, line 228: Either way, hopefully password is completely removed in a day or so.

kifi-backend/modules/shoebox/app/views/auth/authGrey.scala.html
```diff
@@ -38,6 +38,10 @@ <h3 class="intro-header">Watch a quick introduction to Kifi</h3>
           <span class="form-network-icon"></span>
           <span class="form-network-text">Sign up using Facebook</span>
         </a>
+        <a class="form-network twitter" href="@routes.signup("twitter")" data-track-action="clickauthTwitter">
```
andrewconner: Hide this button for now. We know where it's going, and don't want to open it for everyone yet.

andrewconner: Commenting this out is fine.

kifi-backend/modules/shoebox/app/views/auth/authGrey.scala.html
```diff
@@ -100,17 +104,25 @@ <h3 class="intro-header">Watch a quick introduction to Kifi</h3>
       <div class="social-photo form-photo"@if(picturePath != ""){ style="background-image:url(@picturePath)"}></div>
       <div class="social-text-info">
         <div class="social-name">@firstName @lastName</div>
-        <input class="social-email" type="email" name="signup-2-email" value="@emailAddress" placeholder="Your email address" disabled>
-        <a class="social-change-email" href="javascript:" data-track-action="clickchangeEmail">Change email</a>
+
+        @if(network == Some(SocialNetworkType("twitter"))) {
```
andrewconner: Is this check necessary? If `emailAddress` is empty, then this case needs to show. Otherwise, the other case is fine.

kifi-backend/modules/shoebox/app/views/auth/authGrey.scala.html
```diff
@@ -100,17 +104,25 @@ <h3 class="intro-header">Watch a quick introduction to Kifi</h3>
...
       </div>
     </div>
-    <div class="form-creds">
-      <label class="password-label signup-2-label">
-        Create a password
-        <input class="form-password" type="password" name="signup-2-password" autofocus>
-        <span class="form-password-tip">at least 7 characters</span>
-      </label>
-    </div>
+    @if(network != Some(SocialNetworkType("twitter"))) {
```
andrewconner: Your solution is fine, but you could also show password box for now regardless of the network. Removing the password requirement is the same for all social networks, so a bit odd to special case twitter here, and then generate a password just for twitter.

andrewconner:
Thanks!

kifi-backend/modules/shoebox/app/views/auth/authGrey.scala.html
```diff
@@ -100,8 +100,14 @@ <h3 class="intro-header">Watch a quick introduction to Kifi</h3>
...
-        <input class="social-email" type="email" name="signup-2-email" value="@emailAddress" placeholder="Your email address" disabled>
-        <a class="social-change-email" href="javascript:" data-track-action="clickchangeEmail">Change email</a>
+
+        @if(emailAddress == "") {
+          <input class="social-email" type="email" name="signup-2-email" value="@emailAddress" placeholder="Your email address">
+        } else {
+          <input class="social-email" type="email" name="signup-2-email" value="@emailAddress" placeholder="Your email address" disabled>
+          <a class="social-change-email" href="javascript:" data-track-action="clickchangeEmail">Change email</a>
+        }
+
```
ahsu1230, line 110: if `emailAddress` is empty (twitter), we show email input. If it's not empty (fb & li), we show the whole change email input

kifi-backend/modules/shoebox/public/js/auth.js
```diff
@@ -314,8 +314,12 @@ kifi.form = (function () {
...
     var $email = $form.find('.social-email');
     var email = kifi.form.validateEmailAddress($email);
-    var password = kifi.form.validateNewPassword($form.find('.form-password'));
-    if (password) {
+    var $password = $form.find('.form-password');
+    var password = kifi.form.validateNewPassword($password);
+
+    // for FbLi we need valid password (already approved email)
+    // for Twitter we need valid password && email
+    if (password && email) {
```
ahsu1230, line 322: For now all social networks, password && email must be valid. email should be approved because Li & Fb ensure the email is not empty.

ahsu1230:
Sorry this is taking so long, I think I know what's going on now (and sbt being slow doesn't help either).
Next step is to remove password for all social network signups - generating random passwords for them.

PR #13132, opened 2015-01-15
add a request cosolidator to the graph client

ymatsuda:
@stkem @leogrim @yingjieMiao @joshm1 @eishay 

I saw bursts of graph calls, getConnectedUserScores, that are duplicates. These bursts sometimes cause full-GCs. I am adding a request consolidator for this call as a safeguard. I will investigate what is causing these bursts.


yingjieMiao:
good catch. should we also add one to `uriWandering` related code?

PR #13136, opened 2015-01-15
cleaner string manipulation and adding the html element in the error mes...

eishay:
...sage for quicker resolution and fix


eishay:
@joshm1

joshm1:
:+1:

PR #13137, opened 2015-01-15
stop burst by removing Future.sequence from Library reco generation

ymatsuda:
@joshm1 @stkem

It turned out that Library recommendation generation was causing the burst.


stkem:
Oh, nice find!

joshm1:
Nice find. thanks!

yingjieMiao:
:+1:

eishay:
nice!

PR #13138, opened 2015-01-15
Add create-library card to user profiles

yipingliao:
@2is10 Thanks for the help!

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
+++ b/kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
@@ -139,15 +139,16 @@ angular.module('kifi')
 
 
 .controller('UserProfileLibrariesCtrl', [
-  '$scope', '$rootScope', '$state', '$stateParams',
+  '$scope', '$rootScope', '$state', '$stateParams', '$timeout',
   'routeService', 'keepWhoService', 'profileService', 'userProfileActionService', 'libraryService', 'modalService',
-  function ($scope, $rootScope, $state, $stateParams,
+  function ($scope, $rootScope, $state, $stateParams, $timeout,
     routeService, keepWhoService, profileService, userProfileActionService, libraryService, modalService) {
     var username = $stateParams.username;
     var fetchPageSize = 12;
     var fetchPageNumber = 0;
     var hasMoreLibraries = true;
     var loading = false;
+    var newlyAddedLibraryIds = [];
 
     $scope.libraryType = $state.current.data.libraryType;
     $scope.libraries = null;
@@ -176,6 +177,7 @@ angular.module('kifi')
       $scope.libraries = null;
       fetchPageNumber = 0;
       hasMoreLibraries = true;
+      newlyAddedLibraryIds.length = 0;
       loading = false;
     }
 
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
```diff
@@ -206,7 +208,17 @@ angular.module('kifi')
           hasMoreLibraries = data[filter].length === fetchPageSize;
 
           var owner = filter === 'own' ? _.extend({username: username}, $scope.profile) : null;
-          $scope.libraries = ($scope.libraries || []).concat(data[filter].map(augmentLibrary.bind(null, owner)));
+          var filteredLibs = data[filter];
+
+          if (filter === 'own' && newlyAddedLibraryIds.length) {
```
yipingliao, line 213: This is the code that checks for whether we just fetched a card that's already on the page.

kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
```diff
@@ -222,6 +234,42 @@ angular.module('kifi')
...
+        $timeout(function () {
+          newLibrary.justAddedBelow = false;
+          newLibrary.justAddedBelowAcross = false;
+        });
+      }
+
+      modalService.open({
+        template: 'libraries/manageLibraryModal.tpl.html',
+        modalData: {
+          returnAction: function (newLibrary) {
```
yipingliao, line 261: This is a callback that gets executed when a new library has been created on the server.

yipingliao:
I am going to merge this so I can ask Jonathan to test it and because it's still behind an experiment; will address code review comments in a subsequent pull.

kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
```diff
@@ -222,6 +234,42 @@ angular.module('kifi')
...
 
+    $scope.openCreateLibrary = function () {
+      function addNewLibAnimationClass(newLibrary) {
+        // If the second system library card is under the create-library card,
+        // then there are two cards across and the new library will be
+        // below and across from the create-library card.
+        if ((Math.abs(angular.element('.kf-upl-create-card').offset().left -
+                      angular.element('.kf-upl-lib-card').eq(1).offset().left)) < 10) {
+          newLibrary.justAddedBelowAcross = true;
+        }
```
2is10, line 245: As long as we’re measuring where a specific card is in relation to the create card, we could easily compute the translation amount and set it using an inline style. That way this code and the CSS could avoid making assumptions about card spacing and the maximum number of cards per row. Not critical now, just something to consider going forward.

kifi-backend/modules/shoebox/angular/src/userProfile/userProfileLibrariesList.tpl.html
```diff
@@ -1,5 +1,10 @@
 <div class="kf-upl-card-list" smart-scroll scroll-distance="'100%'" scroll-disabled="!hasMoreLibraries()" scroll-next="fetchLibraries()">
-  <div class="kf-upl-card" ng-repeat="lib in libraries" ng-style="{'background-color': lib.color}">
+  <div ng-if="viewingOwnProfile && libraryType === 'own'" class="kf-upl-card kf-upl-create-card" ng-click="openCreateLibrary()">
+    <span class="kf-upl-create-card-add-icon svg-create-library"></span>
+    <div class="kf-upl-create-card-text">CREATE A LIBRARY</div>
+  </div>
+  <div class="kf-upl-card kf-upl-lib-card" ng-repeat="lib in libraries" ng-style="{'background-color': lib.color}"
+       ng-class="{'justAddedBelow': lib.justAddedBelow, 'justAddedBelowAcross': lib.justAddedBelowAcross}">
```
2is10, line 7: We typically use lower-hyphen case for class names.

2is10:
nice!

kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
yipingliao, line 245: Okay, will do!

kifi-backend/modules/shoebox/angular/src/userProfile/userProfileLibrariesList.tpl.html
yipingliao, line 7: Oops, will fix.

PR #13141, opened 2015-01-16
porting new library card design to related library cards

2is10:
@yipingliao @joshm1


kifi-backend/modules/shoebox/angular/src/common/directives/fit.js
```diff
@@ -1,186 +0,0 @@
-/**
```
joshm1, line 1: RIP `kf-fit`, we hardly knew you :sob:

PR #13143, opened 2015-01-16
fixing slug input wrapping and optimally sizing it

2is10:
@yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.tpl.html
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.tpl.html
@@ -27,9 +27,12 @@
             </div>
             <input class="manage-lib-input manage-lib-name-input ng-pristine ng-valid" ng-model="library.name" type="text" name="name" placeholder="Type a name for your library..." autocomplete="off" novalidate>
             <span ng-show="$error.name" class="manage-lib-name-error">{{$error.name}}</span>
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.tpl.html
```diff
@@ -27,9 +27,12 @@ <h1 class="dialog-header">
...
             <span ng-show="$error.name" class="manage-lib-name-error">{{$error.name}}</span>
-            <span class="manage-lib-slug">kifi.com/{{username}}/
-              <input ng-change="editSlug()" class="manage-lib-slug-input ng-pristine ng-valid" ng-model="library.slug" ng-class="{ 'non-empty-slug': !emptySlug }" type="text" name="slug" placeholder="new-library" novalidate>
-            </span>
+            <label class="manage-lib-slug">
+              <span class="manage-lib-slug-1">kifi.com/{{username}}/</span>
+              <span class="manage-lib-slug-2">
+                <input ng-change="editSlug()" class="manage-lib-slug-input ng-pristine ng-valid" ng-model="library.slug" ng-class="{ 'non-empty-slug': !emptySlug }" type="text" name="slug" placeholder="new-library" novalidate>
+              </span>
+            </label>
```
2is10, line 35: Using `<label>` here makes the input get focus when someone clicks on the path part in front of the input.

PR #13146, opened 2015-01-16
Fix broken url links from search page.

yipingliao:
@andrewconner cc @stkem

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/search/search.js
+++ b/kifi-backend/modules/shoebox/angular/src/search/search.js
@@ -249,7 +249,14 @@ angular.module('kifi')
       _.assign($state.params, $location.search());
       init();
     }, 250);
-    $scope.$on('$locationChangeSuccess', newSearch);
+    $scope.$on('$locationChangeSuccess', function (event, newState, oldState) {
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/angular/src/search/search.js
```diff
@@ -249,7 +249,14 @@ angular.module('kifi')
       _.assign($state.params, $location.search());
       init();
     }, 250);
-    $scope.$on('$locationChangeSuccess', newSearch);
```
yipingliao, line 252: Previously, every time we see a location change from the search page, we try to conduct a new search!

kifi-backend/modules/shoebox/angular/src/search/search.js
```diff
@@ -265,7 +272,7 @@ angular.module('kifi')
     $window.addEventListener('beforeunload', onUnload);
 
     $scope.$on('$destroy', function () {
-    onUnload();
+      onUnload();
```
yipingliao, line 275: Just an indentation.

stkem:
Thanks!

PR #13147, opened 2015-01-16
[DO NOT MERGE] Track origin/suborigin of profile page views preliminary

yipingliao:
@2is10 cc @joshm1 

This is what I've come up with for tracking a page view that includes information about where it came from (origin and suborigin). I saw that @joshm1 had previously added a query parameter but I couldn't figure out how to take that off once we land on the new page (Josh doesn't take the query parameter off). Here I am using a service to hold the origin/suborigin data until it's used by tracking. It's a global state, basically, but it's what services are for. I made this into a directive so it can be easily popped into other places.

Please let me know what you think (and if you hate it).

Thanks!

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/common/originTracking.js
+++ b/kifi-backend/modules/shoebox/angular/src/common/originTracking.js
@@ -1,0 +1,42 @@
+'use strict';
+
+angular.module('kifi')
+
+.factory('originTrackingService', [
+  function () {
+    var pageOrigin = null;
+
+    return {
+      set: function (newOrigin) {
+        pageOrigin = newOrigin;
+      },
+
+      get: function () {
+        var temp = pageOrigin;
+        pageOrigin = null;
+        return temp;
+      }
+    };
+  }
+])
+
+.directive('kfTrackOrigin', [
+  'originTrackingService',
+  function (originTrackingService) {
+    return {
+      restrict: 'A',
+      scope: {
+        origin: '@',
+        subOrigin: '@'
+      },
+      link: function (scope, element /*, attrs*/) {
+        element.on('click', function () {
+          originTrackingService.set({
+            origin: scope.origin,
+            subOrigin: scope.subOrigin
+          });
+        });
+      }
+    };
+  }
+]);
--- a/kifi-backend/modules/shoebox/angular/src/common/originTracking.js
+++ b/kifi-backend/modules/shoebox/angular/src/common/originTracking.js
@@ -8,7 +8,7 @@ angular.module('kifi')
 
     return {
       set: function (newOrigin) {
-        pageOrigin = newOrigin;
+        pageOrigin = _.cloneDeep(newOrigin);
       },
 
       get: function () {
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
```diff
@@ -39,6 +39,8 @@ angular.module('kifi')
       }
       $rootScope.$emit('libraryUrl', {});
 
+      console.dir(originTrackingService.get());
```
yipingliao, line 42: Whoever wants to use the data just calls the getter to retrieve the data.

kifi-backend/modules/shoebox/angular/src/userProfile/userProfileLibrariesList.tpl.html
```diff
@@ -18,7 +18,7 @@
       </div>
     </a>
     <div itemscope itemtype="http://data-vocabulary.org/Person">
-      <a class="kf-upl-card-owner" href="{{lib.ownerProfileUrl}}" ng-click="trackProfileClick()" itemprop="url">
+      <a class="kf-upl-card-owner" href="{{lib.ownerProfileUrl}}" kf-track-origin origin="profile" sub-origin="curator" itemprop="url">
```
yipingliao, line 21: Using this directive in just one place right now.

joshm1:
Nice. I like this and prefer it to use query params. :+1:

kifi-backend/modules/shoebox/angular/src/userProfile/userProfileLibrariesList.tpl.html
2is10, line 21: Nice. Might consider a more concise single-attribute syntax:

``` html
kf-origin="profile"
kf-origin="profile/curator"
```

kifi-backend/modules/shoebox/angular/src/common/originTracking.js
```diff
@@ -0,0 +1,42 @@
...
+    return {
+      set: function (newOrigin) {
+        pageOrigin = _.cloneDeep(newOrigin);
+      },
+
+      get: function () {
+        var temp = pageOrigin;
+        pageOrigin = null;
+        return temp;
+      }
```
2is10, line 18: It’s surprising for a method named `get` to be destructive. Might consider `getAndClear()` or `use()` or `take()`.

2is10:
I like this for in-app navigations. We’ll still need to use query or fragment parameters for entry into the app (e.g. from the browser extension). Might be simpler to just have the one method—the least common denominator—but it’s up to you. I guess the extension could also use message passing, but I’m afraid that would create a race condition.

kifi-backend/modules/shoebox/angular/src/userProfile/userProfileLibrariesList.tpl.html
yipingliao, line 21: Good idea. Thanks!

kifi-backend/modules/shoebox/angular/src/common/originTracking.js
yipingliao, line 18: Will do.

yipingliao:
We can have a wrapper method that gets the origin from the query parameters if they are available; and if they are not, retrieves them from the in-app service. I want to try to avoid having urls that have query parameters for tracking as much as possible. :)

PR #13148, opened 2015-01-16
remove password from social network logins

ahsu1230:
@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/core/AuthController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/core/AuthController.scala
@@ -34,6 +34,7 @@ import com.keepit.social.providers.ProviderController
 import securesocial.core.providers.utils.RoutesHelper
 
 import scala.concurrent.Future
+import scala.util.Random
 
 object AuthController {
   val LinkWithKey = "linkWith"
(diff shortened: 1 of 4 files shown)


ahsu1230:
do not merge yet!

kifi-backend/modules/shoebox/app/com/keepit/controllers/core/AuthController.scala
```diff
@@ -437,23 +438,37 @@ class AuthController @Inject() (
...
+                }
+                val sfi = SocialFinalizeInfo(
+                  firstName = User.sanitizeName(identity.firstName),
+                  lastName = User.sanitizeName(identity.lastName),
+                  email = EmailAddress(identity.email.getOrElse("")),
+                  password = password.toCharArray,
+                  picToken = None, picHeight = None, picWidth = None, cropX = None, cropY = None, cropSize = None)
+
+                val result = authHelper.handleSocialFinalizeInfo(sfi, None)(request)
+                result.header.headers.get("uri") match {
```
andrewconner, line 463: Is this necessary? Seems `finishSignup` (called from `handleSocialFinalizeInfo`) does this (and handles the cookie correctly):

``` scala
    val uri = request.session.get(SecureSocial.OriginalUrlKey) getOrElse {
      request.headers.get(USER_AGENT).flatMap { agentString =>
        val agent = UserAgent(agentString)
        if (agent.canRunExtensionIfUpToDate) Some("/install") else None
      } getOrElse "/" // In case the user signs up on a browser that doesn't support the extension
    }
```

ahsu1230, line 463: hm, from what I've tried out before, without the redirecting, I get a page that just has this:

```
{ "uri", "/install" }
```

this comes from the end of `AuthHelper.finishSignup()`, 

```
      authenticator => Ok(Json.obj("uri" -> uri))
        .withCookies(authenticator.toCookie).discardingCookies(DiscardingCookie("inv"))
        .withSession(request.session.setUserId(user.id.get))
    )
```

Is there supposed to be something to handle redirecting otherwise?

andrewconner, line 463: You're right, whoops! We do want to handle the case of redirecting to the cookie location. Instead of `case _ => Redirect("/")`, just go to the URL that's set.

PR #13149, opened 2015-01-16
cleaning up manage library dialog form and adding color picker

2is10:
@yipingliao 
image


kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.styl
```diff
@@ -63,256 +63,268 @@
...
-
-  .manage-lib-followers-button
-    border: 0
-    border-radius: 6px
-    padding: 15px 20px
-    background-color: #7eccf9
-    font-size: 12px
-    font-weight: 700
-    color: white
-
```
2is10, line 85: Unindented this block and a bunch of other rules that were unnecessarily using ancestor selectors. Add `?w=1` to URL in location bar to ignore whitespace changes in this diff.

kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.tpl.html
```diff
@@ -36,42 +34,44 @@ <h1 class="dialog-header">
...
+                  <label for="listed-visibility-checkbox">
+                    <span></span>
+                    List this library on my public profile
+                  </label>
                 </div>
-
-                <div ng-if="$error.general" class="manage-lib-error">{{$error.general}}</div>
+              </div>
+              <div ng-switch-when="secret" class="manage-lib-visibility-description-content">
+                Only people you invite can see this library
```
2is10, line 58: text here is from our mobile UI for the privacy setting

PR #13151, opened 2015-01-16
making the privacy toggle be left by default, right when privacy is on

2is10:
...and using gray background when public (privacy off).

Also making the new library widget wider so that the text can fit without wrapping.

@yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/keep/keepToLibraryWidget.tpl.html
+++ b/kifi-backend/modules/shoebox/angular/src/keep/keepToLibraryWidget.tpl.html
@@ -59,10 +59,6 @@
     </header>
     <div class="keep-to-library-create-heading">Name Your Library</div>
     <div class="keep-to-library-create-name-input-container">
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/shoebox/angular/src/keep/keepToLibraryWidget.tpl.html
```diff
@@ -59,10 +59,6 @@
     </header>
     <div class="keep-to-library-create-heading">Name Your Library</div>
     <div class="keep-to-library-create-name-input-container">
-      <span ng-switch="newLibrary.visibility" class="keep-to-library-lib-icon-container">
-        <span ng-switch-default class="keep-to-library-lib-icon svg-public-library-icon"></span>
-        <span ng-switch-when="secret" class="keep-to-library-lib-icon svg-private-library-icon"></span>
-      </span>
```
2is10, line 65: Removing library icon next to name. (We removed it from the extension a long time ago.)

PR #13154, opened 2015-01-16
falling back to Library.createdAt if .lastKept is missing (for cards)

2is10:
Also fixing an issue where `"following": true` was being returned for viewer’s own libraries when viewed on someone else’s profile.

@ahsu1230

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
@@ -1136,38 +1136,33 @@ class LibraryCommander @Inject() (
       ).zipWithIndex.map { case (value, idx) => value.id -> (value, idx) }.toMap
 
       val libIds = systemValueLibraries.keySet
-      val libIdsMap = db.readOnlyReplica { implicit s => libraryRepo.getLibraries(libIds) } filter {
-        case (_, lib) => lib.visibility == LibraryVisibility.PUBLISHED
+      val libs = db.readOnlyReplica { implicit s =>
+        libraryRepo.getLibraries(libIds).values.toSeq.filter(_.visibility == LibraryVisibility.PUBLISHED)
+      }
+      val infos: ParSeq[LibraryCardInfo] = db.readOnlyMaster { implicit s =>
+        val owners = basicUserRepo.loadAll(libs.map(_.ownerId).toSet)
+        createLibraryCardInfos(libs, owners, None, false, ProcessedImageSize.Medium.idealSize)
       }
 
-      val fullLibInfosF = createFullLibraryInfos(viewerUserIdOpt = None,
-        showPublishedLibraries = true,
-        maxMembersShown = 0,
-        maxKeepsShown = 0,
-        idealKeepImageSize = ProcessedImageSize.Medium.idealSize,
-        libraries = libIdsMap.values.toSeq,
-        idealLibraryImageSize = ProcessedImageSize.Medium.idealSize, withKeepTime = true)
-
-      fullLibInfosF map { libInfos =>
-        libInfos map {
-          case (libId, info) =>
-            val (extraInfo, idx) = systemValueLibraries(libId)
-            val card = LibraryCardInfo(
+      SafeFuture {
+        (infos.zip(libs).map {
+          case (info, lib) =>
+            val (extraInfo, idx) = systemValueLibraries(lib.id.get)
+            idx -> LibraryCardInfo(
               id = info.id,
               name = info.name,
               description = None, // not currently used
               color = info.color,
               image = info.image,
               slug = info.slug,
-              owner = BasicUserWithFriendStatus.fromWithoutFriendStatus(info.owner),
+              owner = info.owner,
               numKeeps = info.numKeeps,
               numFollowers = info.numFollowers,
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -1136,38 +1136,33 @@ class LibraryCommander @Inject() (
...
-      fullLibInfosF map { libInfos =>
-        libInfos map {
-          case (libId, info) =>
-            val (extraInfo, idx) = systemValueLibraries(libId)
-            val card = LibraryCardInfo(
+      SafeFuture {
+        (infos.zip(libs).map {
+          case (info, lib) =>
+            val (extraInfo, idx) = systemValueLibraries(lib.id.get)
+            idx -> LibraryCardInfo(
```
2is10, line 1151: Here I switched `createFullLibraryInfos` to `createLibraryCardInfos` for the marketing page. It’s simpler to create them directly, and should be faster.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -1279,24 +1274,20 @@ class LibraryCommander @Inject() (
...
-        createLibraryCardInfos(libs, owners, user, viewer, idealSize)
+        createLibraryCardInfos(libs, owners, viewer, false, idealSize)
       }
     } else {
       ParSeq.empty
     }
   }
 
-  private def createLibraryCardInfos(libs: Seq[Library], owners: Map[Id[User], BasicUser], profileUser: User, viewer: Option[User], idealSize: ImageSize)(implicit session: RSession): ParSeq[LibraryCardInfo] = {
+  private def createLibraryCardInfos(libs: Seq[Library], owners: Map[Id[User], BasicUser], viewer: Option[User], withFollowing: Boolean, idealSize: ImageSize)(implicit session: RSession): ParSeq[LibraryCardInfo] = {
```
2is10, line 1284: This method isn’t necessarily specific to profile page views. It now takes a `withFollowing` boolean instead of taking `profileUser` and computing the boolean.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -1309,12 +1300,13 @@ class LibraryCommander @Inject() (
         (0, Seq.empty)
       }
 
-      val isFollowing = if (isUserViewingOtherProfile) {
+      val ownerWFS = ownersWFS(lib.ownerId)
+      val isFollowing = if (withFollowing && viewer.isDefined && viewer.get.externalId != ownerWFS.externalId) {
```
2is10, line 1304: The `viewer.get.externalId != ownerWFS.externalId` part is the fix. `following` is not applicable for the viewer’s own libraries, even if `following` is applicable for other libraries in this batch.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
```diff
@@ -557,7 +557,7 @@ class LibraryController @Inject() (
               numKeeps = info.numKeeps,
               numFollowers = info.numFollowers,
               followers = LibraryCardInfo.showable(info.followers),
-              lastKept = info.lastKept,
+              lastKept = info.lastKept.get,
```
2is10: The related libraries code path should probably also bypass creation of `FullLibraryInfo` objects, but since it doesn’t yet, we don’t currently have access to `library.createdAt` as a fallback. However, `lastKept.get` should be safe since we wouldn’t recommend a library that has no keeps.

...riiiiight? :wink:

Maybe I’ll change this to fallback to time `0`, for peace of mind. We don’t even use this value in the related library card.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
ahsu1230, line 1304: yup, nice!

ahsu1230:
lgtm! Thanks!

kifi-backend/modules/shoebox/test/com/keepit/controllers/website/LibraryControllerTest.scala
```diff
@@ -558,7 +554,7 @@ class LibraryControllerTest extends Specification with ShoeboxTestInjector {
         val result5Json = (contentAsJson(result5) \ "following").as[Seq[JsObject]]
         (result5Json.head \ "name").as[String] === "lib1"
         (result5Json.head \ "numFollowers").as[Int] === 1
-        (result5Json.head \ "following").as[Boolean] === true
+        (result5Json.head \ "following").asOpt[Boolean] === None
```
2is10, line 557: Here user 1 is looking at the libraries user 2 is following, and seeing his/her own library at the top of the list.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
ahsu1230: loooll I guess haha. The related library will probably always have the "from ____" on the footer anyway.

PR #13156, opened 2015-01-16
fix keeping from unfollowed libraries!

ahsu1230:
Nice catch @yipingliao @joshm1 
@andrewconner


yipingliao:
Thanks for fixing this! Profiles would suck without this feature. :)

PR #13157, opened 2015-01-16
stringify vertices lazily when persisting

ymatsuda:
@leogrim @eishay 

Reducing the peak memory usage during backup by lazily stringify vertices.
We have a backup task every two hours, and graph regularly has a full GC every two hours. This may stop it hopefully.


eishay:
few lines, huge difference. cool!

leogrim:
Oh good catch, thanks!

PR #13159, opened 2015-01-16
hiding Create a Library card until first batch of cards is inserted

2is10:
Also making it a button instead of a div.

@yipingliao


yipingliao:
Awesome. Thanks!

PR #13160, opened 2015-01-16
Track profile page view origin for the profile page and the library page

yipingliao:
Also, missing curator pic/url in follower modal bug.

@2is10 Based on the PR from yesterday re tracking origin. Currently tracking a subset of the links that link to user profile pages, but will add more after the release.

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/libraryFollowers.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/libraryFollowers.js
@@ -53,6 +53,14 @@ angular.module('kifi')
           }
         }
 
+        function augmentLibrary() {
+          // The passed-in library may have owner information on different properties;
+          // normalize the properties (scope.library properties override scope.library.owner properties).
+          scope.library.owner = scope.library.owner || {};
+          scope.library.owner.picUrl = scope.library.ownerPicUrl || scope.library.owner.picUrl;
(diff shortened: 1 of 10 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/libraryFollowers.js
```diff
@@ -53,6 +53,14 @@ angular.module('kifi')
           }
         }
 
+        function augmentLibrary() {
+          // The passed-in library may have owner information on different properties;
+          // normalize the properties (scope.library properties override scope.library.owner properties).
+          scope.library.owner = scope.library.owner || {};
```
yipingliao, line 59: This fixes a bug where when you open up a followers modal from a user profile library card, the curator information (pic and href) are missing.

kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
```diff
@@ -4,14 +4,22 @@ angular.module('kifi')
...
 
     //
     // Internal data.
     //
 
+    // Mapping of library type to origin contexts for tracking.
+    var originContexts = {
+      'own': 'MyLibraries',
+      'following': 'FollowedLibraries',
+      'invited': 'InvitedLibraries'
```
2is10, line 19: I think thus far we have used lower-camel case for origins in MixPanel, like `myLibraries`. Oh well?

PR #13163, opened 2015-01-16
Simplifying image size requirements, removing NormalizedURI param for fetching URISummaries

andrewconner:
@stkem @leogrim Should be pretty safe, scraper needs to be deployed first, then shoebox.

Léo: was finally able to remove NormalizedURI requirement to fetch a URI summary. A little can be cleaned, but it's better now.


kifi-backend/modules/common/app/com/keepit/scraper/ScraperServiceClient.scala
```diff
@@ -118,6 +119,8 @@ case class ScraperThreadInstanceInfo(info: AmazonInstanceInfo, jobInfo: Either[S
   }
 }
 
+@json case class NormalizedURIRef(id: Id[NormalizedURI], url: String, externalId: ExternalId[NormalizedURI])
```
stkem, line 122: :+1:

leogrim, line 122: :+1:

stkem:
lgtm

PR #13166, opened 2015-01-16
Add link to profiles from related library cards.

yipingliao:
@2is10 @joshm1


kifi-backend/modules/shoebox/angular/src/friends/rightColFriends.styl
```diff
@@ -1,6 +1,12 @@
 .kf-rightcol-friends
   text-align: center
 
+  a:hover
```
yipingliao, line 4: We had to remove a hover rule from `kf-footer` to get the link underlining in related lib cards to work; adding that rule back in for specific modules.

2is10:
:+1:

PR #13169, opened 2015-01-16
social signup sets up a Social_Signup user_value [BE]

ahsu1230:
@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/AuthCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/AuthCommander.scala
@@ -206,7 +206,7 @@ class AuthCommander @Inject() (
         email = sfi.email, passwordInfo = pInfo, firstName = sfi.firstName, lastName = sfi.lastName, isComplete = true)
 
       db.readWrite { implicit s =>
-        userValueRepo.setValue(userId, UserValueName.SOCIAL_SIGNUP, true)
+        userValueRepo.setValue(userId, UserValueName.HAS_NO_PASSWORD, true)
       }
 
       val userPreUsername = db.readOnlyMaster { implicit session =>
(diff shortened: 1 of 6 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/AuthCommander.scala
```diff
@@ -205,6 +205,10 @@ class AuthCommander @Inject() (
       val (emailPassIdentity, userId) = saveUserPasswordIdentity(None, Some(socialIdentity),
         email = sfi.email, passwordInfo = pInfo, firstName = sfi.firstName, lastName = sfi.lastName, isComplete = true)
 
+      db.readWrite { implicit s =>
+        userValueRepo.setValue(userId, UserValueName.SOCIAL_SIGNUP, true)
+      }
```
ahsu1230, line 210: in `finalizeSocialAccount()`, set the `SOCIAL_SIGNUP` to be true (default false) here right after the user is created

ahsu1230:
woops forgot something. Once you create your new password after, the socialSignup should be set to false again (and maybe even not appear anymore)

andrewconner:
How about naming it "has_no_password" then?

ahsu1230:
sure, it's more generic.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/UserController.scala
```diff
@@ -309,13 +309,18 @@ class UserController @Inject() (
   }
 
   private def getUserInfo[T](userId: Id[User]) = {
-    val user = db.readOnlyMaster { implicit session => userRepo.get(userId) }
+    val (user, socialSignup) = db.readOnlyMaster { implicit session =>
```
andrewconner: It may be better to do this in `UserController#SitePrefNames`, which will include it in the site prefs. One line, pipeline already exists. Changing user info for this is a bit messy imho.

ahsu1230: ah I see. that works, thanks

PR #13173, opened 2015-01-17
showing register dialog if logged out user clicks follow button

2is10:
@yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
+++ b/kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
@@ -155,10 +155,10 @@ angular.module('kifi')
 
 
 .controller('UserProfileLibrariesCtrl', [
-  '$scope', '$rootScope', '$state', '$stateParams', '$timeout',
-  'routeService', 'keepWhoService', 'profileService', 'userProfileActionService', 'libraryService', 'modalService',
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
```diff
@@ -318,6 +318,13 @@ angular.module('kifi')
     };
 
     $scope.onFollowButtonClick = function (lib, $event) {
+      if (platformService.isSupportedMobilePlatform()) {
+        var url = $location.absUrl();
+        platformService.goToAppOrStore(url + (url.indexOf('?') > 0 ? '&' : '?') + 'follow=true');
+        return;
+      } else if ($rootScope.userLoggedIn === false) {
+        return signupService.register({libraryId: lib.id});
```
2is10, line 326: I can get this to show the dialog locally, but it also redirects to the recommendations page. I’m not sure whether it will have the same behavior in production or not. Worth a try, I guess.

yipingliao, line 326: :)

2is10, line 326: Seems to work.

yipingliao, line 326: Yup! Just tried it too.

PR #13178, opened 2015-01-17
Fixes redirect/cookie issue for social accounts

andrewconner:
This is bad code. Sorry.


andrewconner:
@ahsu1230 Says it's great

PR #13179, opened 2015-01-17
Fix bug where newly created library upl cards are not linked to the library

yipingliao:
@2is10 The problem was that I was not fully augmenting the new library.


2is10:
:+1:

PR #13180, opened 2015-01-17
showing library color strip in large library cards

2is10:
@yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/library.styl
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/library.styl
@@ -21,18 +21,25 @@ invite-header-orange = #ff7c3c
   margin: 0 auto
 
 
-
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/library.styl
```diff
@@ -388,17 +395,17 @@ html:not(.kf-mobile)
       right: -100px
 
     .kf-keep-lib-search-input
-      width: 495px
+      width: 430px
```
2is10, line 398: The search input sometimes overlaps long names. Better would probably be to use `display: table-cell` to make the search input always fill up the available space next to the library owner name.

2is10:
tested:
- logged out library page
- logged out library search
- logged out library page on mobile
- logged in looking at someone else’s library
- logged in looking at my own library
- library recommendation card

PR #13183, opened 2015-01-17
right-aligning add a cover photo link

2is10:
@yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.tpl.html
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.tpl.html
@@ -34,7 +34,10 @@
           <div ng-if="coverImageProgress" class="kf-keep-lib-cover-image-progress"></div>
         </div>
         <div ng-if="coverImageError" class="kf-keep-lib-cover-image-error">{{coverImageError}}</div>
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.tpl.html
```diff
@@ -34,7 +34,10 @@
           <div ng-if="coverImageProgress" class="kf-keep-lib-cover-image-progress"></div>
         </div>
         <div ng-if="coverImageError" class="kf-keep-lib-cover-image-error">{{coverImageError}}</div>
-        <input ng-if="!recommendation && isMyLibrary() && isUserCreatedLibrary()" type="file" class="kf-keep-lib-pic-file-input" accept="image/jpeg,image/png,image/gif" onchange="angular.element(this).scope().onCoverImageFileChosen(this.files[0]);this.value=null">
+        <input ng-if="!recommendation && isMyLibrary() && isUserCreatedLibrary()" type="file" class="kf-keep-lib-cover-image-file" accept="image/jpeg,image/png,image/gif" onchange="angular.element(this).scope().onCoverImageFileChosen(this.files[0]);this.value=null">
+        <a ng-if="!recommendation && isMyLibrary() && isUserCreatedLibrary() && !coverImageUrl" class="kf-keep-lib-cover-image-add" href="javascript:" onclick="angular.element(this).scope().onClickAddCoverImage(event)">
+          Add a cover photo
+        </a>
```
2is10, line 40: renamed a couple of classes to have the common `kf-keep-lib-cover-image-` prefix

2is10:
not sure whether Danny will like this, but I think it makes sense now that there’s a line just above the link to anchor it

yipingliao:
:+1: I like it; it was kind of floating by itself previously.

PR #13185, opened 2015-01-17
join Kifi on public library pages broken

ahsu1230:
@andrewconner @yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/layout/header/PublicHeaderCtrl.js
+++ b/kifi-backend/modules/shoebox/angular/src/layout/header/PublicHeaderCtrl.js
@@ -18,7 +18,9 @@ angular.module('kifi')
     };
 
     $scope.join = function ($event) {
-      $event.preventDefault();
+      if (platformService.isSupportedMobilePlatform()) {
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/angular/src/layout/header/PublicHeaderCtrl.js
```diff
@@ -18,7 +18,9 @@ angular.module('kifi')
     };
 
     $scope.join = function ($event) {
-      $event.preventDefault();
+      if (platformService.isSupportedMobilePlatform()) {
```
ahsu1230, line 21: just copying how `$scope.login()` works. There wasn't this `if` block before.

ahsu1230:
Sorry I didn't clarify. On a public library page, it worked as intended. But the link was broken when you went to a public profile page. Right now, this is a really quick fix so the links aren't broken. Going to the signup page isn't ideal, we want that modal to pop up right? 
I'm still looking why it's having trouble. I think it's because at a profile page, the `getCurrentLibrary` doesn't really work so we need to differentiate if we're in a profile page or library page?

PR #13186, opened 2015-01-17
better temp fix

ahsu1230:
@andrewconner @yipingliao 

Nvm, I think is a better temp fix than https://github.com/kifi/keepit/pull/13185

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
+++ b/kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
@@ -135,6 +135,11 @@ angular.module('kifi')
     });
     $scope.$on('$destroy', deregister$stateChangeSuccess);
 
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
```diff
@@ -135,6 +135,11 @@ angular.module('kifi')
     });
     $scope.$on('$destroy', deregister$stateChangeSuccess);
 
+    var deregisterCurrentLibrary = $rootScope.$on('getCurrentLibrary', function (e, args) {
+      args.callback({});
+    });
+    $scope.$on('$destroy', deregisterCurrentLibrary);
```
ahsu1230, line 141: `getCurrentLibrary` is emitted, but there's nobody to receive it when you go to a public profile page.
So for now, I think returning an empty object is okay. So the modal at least pops up

PR #13187, opened 2015-01-18
on graph serialization, persisting smaller chunks of the graph straight ...

eishay:
...to disk without needing to allocate large string representation of the set of edges.

note that the serialization format had changed and the reader is backward compatible.

### DIFF
--- a/kifi-backend/modules/graph/app/com/keepit/graph/simple/MutableVertex.scala
+++ b/kifi-backend/modules/graph/app/com/keepit/graph/simple/MutableVertex.scala
@@ -41,7 +41,7 @@ class MutableVertex(var data: VertexDataReader, val outgoingEdges: MutableOutgoi
    */
   def write(out: ObjectOutputStream): Unit = {
     out.writeUTF(VertexDataReader.writes.writes(data).toString)
-    out.writeInt(outgoingEdges.edges.foldLeft(0)((_, edges) => edges._2.size))
+    out.writeInt(outgoingEdges.edges.foldLeft(0)((count, edges) => count + edges._2.size))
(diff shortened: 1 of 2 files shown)


eishay:
@ymatsuda @leogrim

kifi-backend/modules/graph/app/com/keepit/graph/simple/MutableVertex.scala
```diff
@@ -33,11 +35,43 @@ class MutableVertex(var data: VertexDataReader, val outgoingEdges: MutableOutgoi
...
     if (incomingEdges.edges(component).isEmpty) { incomingEdges.edges -= component }
   }
+
+  /**
+   * UTF, Int, (Long, UTF)
+   */
+  def write(out: ObjectOutputStream): Unit = {
+    out.writeUTF(VertexDataReader.writes.writes(data).toString)
+    out.writeInt(outgoingEdges.edges.foldLeft(0)((count, edges) => count + edges._2.size))
+    outgoingEdges.edges foreach {
```
ymatsuda, line 45: `outgoingEdges.edges` is mutable. There is a slight chance that the count becomes inconsistent.

eishay, line 45: @leogrim are we not locking writes when persisting the graph?
If not, I can clone it to an immutable list.

leogrim, line 45: Yes we do, see `GraphManagerActor`, calls `graph.update` and `graph.backup` are synchronous there.

leogrim:
:+1: Thanks!

PR #13192, opened 2015-01-20
Properly close manage-lib modal when clicking on user image in share dropdown

yipingliao:
Also fix missing search icon in share dropdown in same modal.

@2is10


2is10:
:+1:

PR #13193, opened 2015-01-20
Properly validate email addresses in AuthController + use SafeFuture for service calls

leogrim:
@andrewconner @eishay @stkem @ymatsuda 
Following Yesterday's issues: 
- Adding proper typing (hence proper validation) for emails in AuthController
- Somehow the call to `getUserIndexable` ends up wrapped in a SafeFuture in `Search`, but was silently failing in `Graph`. Our HTTP client explicitly reports failed service calls, but returns a regular Future so that exceptions further up (e.g. deserialization) may be swallowed => now wrapping all service calls ServiceClient with a `SafeFuture`.

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/common/service/ServiceClient.scala
+++ b/kifi-backend/modules/common/app/com/keepit/common/service/ServiceClient.scala
@@ -119,13 +119,13 @@ trait ServiceClient extends CommonServiceUtilities with Logging {
       }
       if (ignoreFailure) {
         call match {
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/common/app/com/keepit/common/service/ServiceClient.scala
```diff
@@ -125,7 +127,7 @@ trait ServiceClient extends CommonServiceUtilities with Logging {
         case c @ ServiceRoute(POST, _, _*) => httpClient.withTimeout(callTimeouts).postFuture(httpUri, body)
       }
     }
-  }
+  })
```
stkem: You may want to give the safe future a name. Makes for better errors.

PR #13195, opened 2015-01-20
require library recos to have >= 2 followers and >= 5 keeps

joshm1:
@stkem


stkem:
> =10 keeps might be a bit high. Maybe 5?

stkem:
lgtm otherwise

PR #13196, opened 2015-01-20
avoiding an error dialog on follow/unfollow in some cases

2is10:
This pull request breaks up the `libraryUpdated` event into three more specific ones: `libraryJoined`, `libraryLeft`, and `libraryKeepCountChanged`. Each now contains less information. This way we don’t need to load a fresh copy of all information about a library from the server when some of these events occur (which sometimes isn’t possible).

Also showing a more specific error message when the user tries to follow on a library card but the owner has made the library private.

@yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
@@ -161,10 +161,7 @@ angular.module('kifi')
           }
 
           scope.library.followers = scope.library.followers || [];
-          scope.library.followers.forEach(function (follower) {
-            follower.picUrl = friendService.getPictureUrlForUser(follower);
-            follower.profileUrl = routeService.getProfileUrl(follower.username);
-          });
+          scope.library.followers.forEach(augmentFollower);
 
           var maxLength = 150;
           if (scope.library.description.length > maxLength && !scope.isUserLoggedOut) {
@@ -199,6 +196,11 @@ angular.module('kifi')
             scope.coverImageUrl = env.picBase + '/' + image.path;
             scope.coverImagePos = formatCoverImagePos(image);
           }
+        }
+
+        function augmentFollower(follower) {
+          follower.picUrl = friendService.getPictureUrlForUser(follower);
+          follower.profileUrl = routeService.getProfileUrl(follower.username);
         }
 
         function preloadSocial() {
@@ -789,28 +791,7 @@ angular.module('kifi')
             return signupService.register({libraryId: scope.library.id});
           }
 
-          libraryService.joinLibrary(scope.library.id).then(function (result) {
-            (scope.library.invite || {}).actedOn = true;
-
-            if (result === 'already_joined') {
-              modalService.openGenericErrorModal({
-                modalData: {
-                  genericErrorMessage: 'You are already following this library!'
-                }
-              });
-              return;
-            }
-
-            scope.library.followers.push({
-              id: profileService.me.id,
-              firstName: profileService.me.firstName,
-              lastName: profileService.me.lastName,
-              pictureName: profileService.me.pictureName
-            });
-
-            augmentData();
-            adjustFollowerPicsSize();
-          })['catch'](modalService.openGenericErrorModal);
+          libraryService.joinLibrary(scope.library.id)['catch'](modalService.openGenericErrorModal);
         };
 
         scope.unfollowLibrary = function () {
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
```diff
@@ -789,28 +791,7 @@ angular.module('kifi')
...
-            (scope.library.invite || {}).actedOn = true;
-
-            if (result === 'already_joined') {
-              modalService.openGenericErrorModal({
-                modalData: {
-                  genericErrorMessage: 'You are already following this library!'
-                }
-              });
-              return;
-            }
```
2is10, line 802: I didn’t think this error message was very user friendly. If this situation arises somehow, it’s not the user’s fault. It’s ours for having shown the “Follow” button. Fortunately, the following action can be idempotent. The ideal behavior, I think, is for the UI to update to indicate that the user is now following the library.

Notice that the `.alreadyJoined` and `already_joined` indicators are no longer used at all. A web app’s cached state getting out of sync with the authoritative state on the server is _expected_. It happens. Some data can be changed or read by multiple people. And sometimes the same user will alternate between two devices or browser tabs. So some requests to the server will look to the server like nonsense (redundant). This is okay. It’s not a problem.

What’s more important is that the web app detect whether new data it receives is different from what it already had cached. It doesn’t need extra help from the server to accomplish this. The new code below does exactly this. (See `'libraryJoined'` handler.)

kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
```diff
@@ -789,28 +791,7 @@ angular.module('kifi')
...
-
-            scope.library.followers.push({
-              id: profileService.me.id,
-              firstName: profileService.me.firstName,
-              lastName: profileService.me.lastName,
-              pictureName: profileService.me.pictureName
-            });
```
2is10, line 809: `username` was missing here. I added it in the new code below.

kifi-backend/modules/shoebox/angular/src/libraries/libraryService.js
```diff
@@ -280,48 +266,46 @@ angular.module('kifi')
...
+          $rootScope.$emit('libraryJoined', libraryId);
         });
       },
 
       leaveLibrary: function (libraryId) {
         return $http.post(routeService.leaveLibrary(libraryId)).then(function () {
-          _.remove(librarySummaries, { id: libraryId });
-          $rootScope.$emit('librarySummariesChanged');
-
-          return api.getLibraryById(libraryId, true).then(function (data) {
```
2is10, line 304: This was the problematic request that got me started on this pull request. As we discussed, after leaving a library, the user may not have permission to access it.

kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.js
```diff
@@ -129,9 +129,6 @@ angular.module('kifi')
 
           submitting = true;
           libraryService.deleteLibrary(scope.library.id).then(function () {
-            $rootScope.$emit('librarySummariesChanged');
-            $rootScope.$emit('libraryDeleted', scope.library.id);
-
```
2is10, line 134: moved these into `libraryService` for consistency with other library actions

kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
```diff
@@ -947,14 +928,38 @@ angular.module('kifi')
...
+            if (libraryId === scope.library.id) {
+              scope.library.numKeeps = keepCount;
+            }
+          }),
+          $rootScope.$on('libraryJoined', function (e, libraryId) {
+            var lib = scope.library;
+            if (lib && libraryId === lib.id && lib.access === 'none') {
+              (lib.invite || {}).actedOn = true;
+              lib.access = 'read_only';
+              lib.numFollowers++;
```
yipingliao, line 942: Should this line be moved to inside the conditional that starts on line 944? If the user has already joined the library and this is a second attempt to join, we don't want to increment the number of followers.

yipingliao:
:+1: Thank you so much for cleaning this up and the pointers you gave.

kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
```diff
@@ -947,14 +928,38 @@ angular.module('kifi')
...
+          $rootScope.$on('libraryKeepCountChanged', function (e, libraryId, keepCount) {
+            if (libraryId === scope.library.id) {
+              scope.library.numKeeps = keepCount;
+            }
+          }),
+          $rootScope.$on('libraryJoined', function (e, libraryId) {
+            var lib = scope.library;
+            if (lib && libraryId === lib.id && lib.access === 'none') {
+              (lib.invite || {}).actedOn = true;
+              lib.access = 'read_only';
```
2is10, line 941: In the future, if necessary for disambiguation, we can have the server return the access level and have it trickle down to here as an event argument.

kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
2is10, line 942: Good question. I don’t think so because I don’t think `.followers` is necessarily a complete list.

The `lib.access === 'none'` check above (line 939) should protect us from double-counting.

PR #13197, opened 2015-01-20
Framework for new images API

andrewconner:
@stkem `URIPreviewFetchResult` will be the data transmit type for scraper to return embedly-related data. This includes resized images.


andrewconner:
I have the class that does the image resizing, but am trying to do this in small chunks since the refactorings change (hopefully, not "break") old APIs. Most were broken to begin with.

I want to roll this out and test the new endpoint to get an idea how fast I can expect embedly responses to return (through scraper).

stkem:
lgtm

PR #13198, opened 2015-01-20
fixing twitter signup for mutual followers

ahsu1230:
@andrewconner


ahsu1230:
let me know if I should change any of comments for better clarity

andrewconner:
Great. As mentioned in person, add logging so we can follow it in prod.

PR #13199, opened 2015-01-21
Prefix search

leogrim:
@ymatsuda @yingjieMiao @stkem 
Attempt to blend a regular search (search full text for query keywords) with a prefix search (search entities by name prefix). 

This would be available for library search:
- on mobile for admins with the `PREFIX_SEARCH` experiments
- on the website experimental API `https://search.kifi.com/site/search` with the query parameter `doPrefixSearch`

Please criticize thoroughly :)


kifi-backend/modules/search/app/com/keepit/search/engine/query/KPrefixQuery.scala
```diff
@@ -0,0 +1,95 @@
...
+import java.util.{ Set => JSet }
+
+import scala.collection.mutable.ArrayBuffer
+
+object KPrefixQuery {
+  def get(nameValueField: String, prefixField: String, queryText: String): Option[KPrefixQuery] = {
+    val terms = PrefixFilter.tokenize(queryText)
+    if (terms.isEmpty) None else Some(new KPrefixQuery(nameValueField, prefixField, terms))
+  }
+}
```
leogrim, line 23: Here is the new query, the scoring may have to be adjusted to be comparable with other Lucene queries, and there's already a todo to adjust for field length.

kifi-backend/modules/search/app/com/keepit/search/engine/query/KPrefixQuery.scala
```diff
@@ -0,0 +1,95 @@
...
+  def explain(context: AtomicReaderContext, doc: Int) = booleanWeight.explain(context, doc)
+
+  override def scorer(context: AtomicReaderContext, acceptDocs: Bits): Scorer = {
+    val nameDocsValues = context.reader.getBinaryDocValues(query.nameValueField)
+    val booleanScorer = booleanWeight.scorer(context, acceptDocs)
+    if (nameDocsValues == null || booleanScorer == null) null else new KPrefixScorer(this, booleanScorer, query.terms, nameDocsValues)
+  }
+
+  override def getWeights(out: ArrayBuffer[(Weight, Float)]): Unit = {
+    booleanWeight.getWeights(out)
```
leogrim, line 69: @ymatsuda not sure if this makes sense.

kifi-backend/modules/search/app/com/keepit/search/engine/parser/KQueryParser.scala
```diff
@@ -26,6 +27,7 @@ class KQueryParser(
     override val altStemmingAnalyzer = qp.altStemmingAnalyzer
     override val siteBoost = config.asFloat("siteBoost")
     override val concatBoost = config.asFloat("concatBoost")
+    override val prefixBoost = if (doPrefixSearch) config.asFloat("prefixBoost") else 0.0f
     override val lang: Lang = qp.analyzer.lang
   }
```
leogrim, line 32: @ymatsuda not sure if this is the most elegant way to turn it on and off (vs in ScoreVectorSource query projections)

kifi-backend/modules/search/app/com/keepit/search/index/Indexable.scala
```diff
@@ -122,6 +124,17 @@ trait Indexable[T, S] extends Logging {
...
   }
 
+  protected def buildPrefixField(fieldName: String, fieldValue: String, maxPrefixLength: Int, minPrefixLength: Int = 1): Field = {
+    val tokens = PrefixFilter.tokenize(fieldValue).toSet
+    val prefixes = for {
+      token <- tokens
+      prefixLength <- minPrefixLength to maxPrefixLength
+    } yield {
+      token.take(prefixLength)
+    }
```
yingjieMiao, line 134: tokens is converted to Set. Is order still correct here?

yingjieMiao, line 134: never mind. you only care order within a token, I guess.

leogrim, line 134: It doesn't matter since they're indexed, does it?

yingjieMiao, line 134: yeah, you are right. My search mind is rusted...

yingjieMiao:
lgtm at high level. @ymatsuda is the best person to check details

kifi-backend/modules/search/app/com/keepit/search/engine/query/KPrefixQuery.scala
ymatsuda, line 69: KPrefixQuery should be opaque from ScoreVectorSource, thus I think getWeights should be:

out += ((this, 0.0f))

If you want the prefix query to influence the percent match value, you can push a positive value (> 0.0f).

leogrim, line 69: But then no document would make it through thanks to a prefix match, would one?

ymatsuda, line 69: hmmm... that is true.

ymatsuda, line 69: We definitely have to push KPrefixWeight to let ScoreVectorSource to create KPrefixScorer. I am not sure how to work with the percent match, though.
It may be cleaner if we create a KPrefixQuery and attach it to KTextQuery for each term.

leogrim, line 69: Since the threshold on percentMatch isn't 100%, that would allow a prefix query to match partially wouldn't it? E.g. if I type "Léo Gr", it may match people named Léo whose name does not start with Gr. In our current prefix search I understand that is not considered a match, so it would increase an already large search space (+ may feel surprising in ux, not sure about that).

leogrim:
I'll give this a try, see how it behaves. Please feel free to follow up with more comments, wip :)

PR #13200, opened 2015-01-21
More profiles tracking

yipingliao:
@2is10 

This completes the page view profile tracking for user profile pages on the website.
Asana task: https://app.asana.com/0/21601564789646/21771458106675/f


2is10:
Nice work.

Part of me wonders if it would be possible and perhaps cleaner and easier to derive the origin from `$location` in the `kfTrackOrigin` click handler. That would work for most tracked links, I think. The HTML attribute could be an optional override for cases where the simple location-based mapping isn’t what’s desired.

yipingliao:
Hmm, good idea. Or, we can use `$state.current.name`. I'll look into it.

PR #13203, opened 2015-01-21
Retries for this user value — got a deadlock

andrewconner:
@ahsu1230 

While registering a user, got a deadlock here. Not quite sure why, but this should help.

@ymatsuda Any ideas why this line would cause a deadlock?


ahsu1230:
herm =/

PR #13208, opened 2015-01-21
Migrate from data-vocabulary.org to schema.org

yipingliao:
Also, add microdata to follower pics on user profile library cards.

@2is10 cc @eishay

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/common/directives/keepWho/keepWhoLib.tpl.html
+++ b/kifi-backend/modules/shoebox/angular/src/common/directives/keepWho/keepWhoLib.tpl.html
@@ -1,10 +1,10 @@
 <span ng-if="!library.hidden">
   <span ng-if="!isBot">
     <span class="kf-keep-who-pic-box" ng-class="{ 'secret': library.visibility === 'secret' }">
-      <span ng-if="inUserProfileBeta" itemscope itemtype="http://data-vocabulary.org/Person">
+      <span ng-if="inUserProfileBeta" itemscope itemtype="http://schema.org/Person">
         <a href="{{library.keeperProfileUrl}}" itemprop="url"
            kf-track-origin origin="{{currentPageOrigin}}/alsoKeptInLibraryBy">
-          <span class="kf-keep-who-pic" ng-attr-style="background-image: url({{library.keeperPic}})" itemprop="photo"></span>
+          <span class="kf-keep-who-pic" ng-attr-style="background-image: url({{library.keeperPic}})" itemprop="image"></span>
         </a>
         <meta itemprop="name" content="{{library.keeperName}}" />
       </span>
--- a/kifi-backend/modules/shoebox/angular/src/userProfile/userProfileLibrariesList.tpl.html
+++ b/kifi-backend/modules/shoebox/angular/src/userProfile/userProfileLibrariesList.tpl.html
@@ -17,9 +17,9 @@
         </div>
       </div>
     </a>
-    <div itemscope itemtype="http://data-vocabulary.org/Person">
+    <div itemscope itemtype="http://schema.org/Person">
       <a class="kf-upl-card-owner" href="{{lib.ownerProfileUrl}}" kf-track-origin origin="{{currentPageOrigin}}/curator" itemprop="url">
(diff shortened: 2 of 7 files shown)


kifi-backend/modules/shoebox/angular/src/common/directives/keepWho/keepWhoLib.tpl.html
```diff
@@ -1,10 +1,10 @@
 <span ng-if="!library.hidden">
   <span ng-if="!isBot">
     <span class="kf-keep-who-pic-box" ng-class="{ 'secret': library.visibility === 'secret' }">
-      <span ng-if="inUserProfileBeta" itemscope itemtype="http://data-vocabulary.org/Person">
+      <span ng-if="inUserProfileBeta" itemscope itemtype="http://schema.org/Person">
         <a href="{{library.keeperProfileUrl}}" itemprop="url"
            kf-track-origin origin="{{currentPageOrigin}}/alsoKeptInLibraryBy">
-          <span class="kf-keep-who-pic" ng-attr-style="background-image: url({{library.keeperPic}})" itemprop="photo"></span>
+          <span class="kf-keep-who-pic" ng-attr-style="background-image: url({{library.keeperPic}})" itemprop="image"></span>
```
yipingliao, line 7: With schema.org, `photo` is not a property of `Person`, but `image` is, so all `"photo"`s are being updated to be `"image"`s.

eishay, line 7: thanks!

kifi-backend/modules/shoebox/angular/src/userProfile/userProfileLibrariesList.tpl.html
```diff
@@ -36,8 +36,13 @@
     <button ng-class="{'kf-upl-card-follow': true, 'kf-following': lib.following}" ng-if="lib.owner.id !== me.id" ng-click="onFollowButtonClick(lib, $event)"></button>
     <div class="kf-upl-card-foot" ng-switch="lib.visibility === 'published' && !lib.listed ? 'unlisted' : lib.followers.length ? 'followers' : lib.lastKept ? 'updated' : ''">
       <div class="kf-upl-followers" ng-switch-when="followers">
-        <a href="{{user.profileUrl}}" class="kf-upl-follower" ng-repeat="user in lib.followers" kf-track-origin origin="{{currentPageOrigin}}/follower" ng-click="trackProfileClick()"
-           ng-attr-style="background-image:url({{user.picUrl}})"></a>
+        <span ng-repeat="user in lib.followers" itemscope itemtype="http://schema.org/Person">
```
yipingliao, line 39: @2is10 Previously you had mentioned that maybe we shouldn't add microdata to something as small as a follower pic on the user profile library card, but schema.org only says that it's used for identifying the things on a web page to a search engine, so I think that these photos do properly identify "Person" objects.

@eishay Thoughts?

2is10, line 39: Yes, I still feel that way. The _why_ on schema.org is:

> Search engines ... rely on this markup to improve the display of search results.

I don’t think we want follower pictures affecting the display of search results. For one thing, we pick which follower pictures to show somewhat randomly, and not consistently over time on any given library page.

I think we ought to mark up a user only on his/her profile page and on his/her library pages.

kifi-backend/modules/shoebox/angular/src/common/directives/keepWho/keepWhoLib.tpl.html
2is10, line 7: This whole section is inside of `ng-if="!isBot"` so this markup is not shown to search engines. Also, as with follower pictures, I don’t think we should be marking up people here. The “who kept” pictures shouldn’t affect how a page shows up in search results.

2is10:
:+1:  Thanks for these updates, @yipingliao!

yipingliao:
Thanks for your help on this,@2is10 ! I am going to clean this up in another PR. In the meanwhile, I am going to close this PR and do another one that takes out the profiles_beta experiments hooks first; that'll make the markup a lot cleaner for this work.

PR #13212, opened 2015-01-21
add path to the image info

eishay:


eishay:
@andrewconner

kifi-backend/modules/common/conf/evolutions/shoebox/283.sql
```diff
@@ -0,0 +1,9 @@
+# CORTEX
```
yingjieMiao, line 1: typo here?

eishay, line 1: yes, thanks, will fix

PR #13213, opened 2015-01-21
Modify ServiceClient.broadcast signature and fix graph admin

leogrim:
@eishay

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/common/service/ServiceClient.scala
+++ b/kifi-backend/modules/common/app/com/keepit/common/service/ServiceClient.scala
@@ -1,8 +1,8 @@
 package com.keepit.common.service
 
 import com.keepit.common.akka.SafeFuture
+import com.keepit.common.amazon.AmazonInstanceInfo
 
-import scala.collection.mutable.{ SynchronizedSet, HashSet, Set }
 import scala.concurrent.Future
 import scala.util.Random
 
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/common/app/com/keepit/common/service/ServiceClient.scala
```diff
@@ -143,10 +143,24 @@ trait ServiceClient extends CommonServiceUtilities with Logging {
...
+      url.serviceInstance.instanceInfo -> callUrl(call, url, body)
+    }
+  }.toMap
+
+  protected def collectSuccessfulResponses(calls: Map[AmazonInstanceInfo, Future[ClientResponse]]): Future[Map[AmazonInstanceInfo, ClientResponse]] = {
+    val safeCalls = calls.mapValues { call =>
+      call.map(Some(_)).recover {
+        case error: Throwable =>
+          airbrakeNotifier.notify(s"Failed service call was ignored.", error)
+          None
```
eishay: nice, though you could return an `Either[ClientResponse, Throwable]` and not loose this information.
in fact, I'm not sure how the client would use it if you already airbraked so it may not be that important.

eishay:
:+1:

kifi-backend/modules/common/app/com/keepit/common/service/ServiceClient.scala
leogrim: Oh good idea, I'll do that

leogrim:
Here with `Try`

PR #13217, opened 2015-01-21
tracking user profile views from the extension

2is10:
@yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/app.js
+++ b/kifi-backend/modules/shoebox/angular/src/app.js
@@ -69,49 +69,35 @@ angular.module('kifi', [
   }
 ])
 
-.factory('injectedState', [
+.factory('initParams', [
   '$location',
   function ($location) {
-    var state = {};
-
-    if (_.size($location.search()) > 0) {
-      // There may be URL parameters that we're interested in extracting.
-      _.forOwn($location.search(), function (value, key) {
-        state[key] = value;
-      });
+    var params = ['m', 'o', 'friend', 'subtype'];
+    var search = $location.search();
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/shoebox/angular/src/app.js
```diff
@@ -69,49 +69,35 @@ angular.module('kifi', [
...
+      $location.search(_.omit(search, params)).replace();
     }
-
-    return {
-      state: state,
-      pushState: pushState
-    };
+    return state;
   }
 ])
```
2is10, line 83: @andrewconner I hope this change to `injectedState` is okay with you. It looks like `pushState` wasn’t being used.

2is10, line 83: Note that we now strip the specific params that we’re interested in retrieving this way, if present, from the location bar using `$location.replace()`. This happens early enough that it doesn’t seem to cause any visual glitches or extra network requests.

kifi-backend/modules/shoebox/angular/src/common/originTracking.js
```diff
@@ -3,8 +3,14 @@
...
 .factory('originTrackingService', [
-  function () {
+  'initParams',
+  function (initParams) {
     var pageOrigin = '';
+    var pageOrigins = {
+      xmp: 'extension/messagePane',
+      xsr: 'extension/searchResult',
+      xst: 'extension/socialTooltip'
+    };
```
2is10, line 14: This mechanism should be easy to reuse for any other entry points to our app by adding more mappings here.

yipingliao:
:+1:

PR #13218, opened 2015-01-21
twitter buttons on signup/login & public library page (only if 'twitter'...

ahsu1230:
... query param)

for:
- www.kifi.com/signup
- public library or profile page -> click on Sign Up, Log In or Follow

add `?twitter=1` to see what it would look like for twitter launch

@andrewconner


kifi-backend/modules/shoebox/public/js/auth.js
```diff
@@ -158,6 +158,26 @@ kifi.form = (function () {
...
+  }
+
+  // todo (aaron): remove below for twitter launch
+  if (getParameterByName('twitter')) { // do not show linkedIn (if twitter is a url query param)
+    var linkedInButtons = $('.form-network.linkedin');
+    for (var i=0; i < linkedInButtons.length; i++) {
+      linkedInButtons.eq(i).hide();
+    }
+  } else { // do not show twitter
+    var twitterButtons = $('.form-network.twitter');
```
andrewconner, line 177: How about make it hidden by default? That way it won't flash if JS doesn't run immediately.

ahsu1230, line 177: yea, I think that's a better idea. Thanks!

PR #13226, opened 2015-01-21
off by 1

yingjieMiao:
@stkem 

a hidden bug for quite a while. not major. Side effect is when user's applicable uris number is 5, we have its topic feature, but mark the state as NOT_APPLICABLE. In theory, NOT_APPLICABLE iff the feature vector is NULL.


stkem:
:-)

PR #13235, opened 2015-01-21
Clean up microdata

yipingliao:
@2is10 Per our earlier discussion.

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.tpl.html
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.tpl.html
@@ -60,7 +60,7 @@
 
           <div ng-switch="clippedDescription" class="kf-keep-lib-description-container">
             <p ng-switch-when="true" class="kf-keep-lib-description">
-              <span ng-bind-html="library.shortDescription" itemtype="headline"></span>&hellip;
+              <span ng-bind-html="library.shortDescription" itemtype="description"></span>&hellip;
               <a href="javascript:" class="kf-desc-more" ng-click="showLongDescription()">More</a>
             </p>
             <p ng-switch-when="false" class="kf-keep-lib-description" ng-bind-html="library.formattedDescription"></p>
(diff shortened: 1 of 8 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.tpl.html
```diff
@@ -58,7 +60,7 @@
 
           <div ng-switch="clippedDescription" class="kf-keep-lib-description-container">
             <p ng-switch-when="true" class="kf-keep-lib-description">
-              <span ng-bind-html="library.shortDescription"></span>&hellip;
+              <span ng-bind-html="library.shortDescription" itemtype="headline"></span>&hellip;
```
2is10: I’m not sure “headline of the article” captures what this really is. How about `description`?
Danny’s latest designs show the full description, btw.

yipingliao: Will do. Thanks!

kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.tpl.html
```diff
@@ -1,5 +1,6 @@
 <div class="kf-library-card-header-wrapper" ng-class="{ 'loggedOutUser': isUserLoggedOut, 'library-search-in-progress': librarySearchInProgress, 'library-search': librarySearch }">
-  <div class="kf-library-card-header" ng-class="{ 'loggedOutUser': isUserLoggedOut, 'library-search-in-progress': librarySearchInProgress, 'library-search': librarySearch, 'pageScrolled': pageScrolled }">
+  <div class="kf-library-card-header" ng-class="{ 'loggedOutUser': isUserLoggedOut, 'library-search-in-progress': librarySearchInProgress, 'library-search': librarySearch, 'pageScrolled': pageScrolled }" itemscope itemtype="http://schema.org/CollectionPage">
```
2is10, line 2: Nice. I think this `itemscope` and `itemtype` should eventually be on a container element that contains the creative work (all of the keep cards), but this is fine for now.

2is10:
:+1: Huge improvement. Thanks!

yipingliao:
Added a commit for the modal max-height work we just did to this pull request (it was accidentally done on this branch, but the commit message is separate and describes that work).

PR #13239, opened 2015-01-22
Increase max height of modal to show entire manage-lib modal.

yipingliao:
@2is10 Enlarging the maximum height of our modals a bit; this allows the entire manage-library modal to show up. After the addition of the color picker, that modal is not showing up entirely on smaller screens (e.g., my MacBook Air) even when the browser window fully occupies the screen.


2is10:
Interesting, thanks. While working on that, I reduced the height taken by the privacy section to compensate for the new color section, and I thought there was no net height increase, but I didn’t test on a small screen and could be wrong. :+1:

PR #13241, opened 2015-01-22
dealing with "no image provider" case

eishay:


eishay:
@andrewconner

andrewconner:
Looks good

PR #13242, opened 2015-01-22
using the image key, not the full url

eishay:


eishay:
@andrewconner

andrewconner:
Yep thanks

PR #13246, opened 2015-01-22
Make entire text re followers in upl card footer clickable.

yipingliao:
@2is10 Currently only the "+25" section is clickable; this pull makes the whole follower text clickable.


andrewconner:
Nice :)

2is10:
:+1:

PR #13253, opened 2015-01-22
remove password step from FB signin on public library/profile page

ahsu1230:
Go to a public library/profile page
click on Follow or Sign In
click on Continue with Facebook

After authenticating, you normally would be brought to a finalize modal (You're almost there! Enter in a password phase)
We want to skip this step and just go straight to "Thanks for Registering" install extension modal.

@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/signup/signupService.js
+++ b/kifi-backend/modules/shoebox/angular/src/signup/signupService.js
@@ -209,7 +209,8 @@ angular.module('kifi')
       $scope.registerFinalizeSubmitted = true;
       $scope.requestActive = true;
       var fields;
-      if (!form.$valid) {
+
+      if (form && !form.$valid) {
         $scope.requestActive = false;
         return false;
       } else if ($scope.userData.method === 'social') {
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/angular/src/signup/signupService.js
```diff
@@ -166,9 +166,13 @@ angular.module('kifi')
...
             } else if (resp.code && resp.code === 'continue_signup') {
               // todo: handle case when emails match, so backend auto-logs in user - andrew
-              modalService.close();
+
               $scope.userData.method = 'social';
-              registerFinalizeModal();
+
+              // skip password setting (set randomly)
+              $scope.userData.password = Math.random().toString(36).substring(8);
+              $scope.registerFinalizeSubmit();
```
andrewconner, line 174: Shouldn't the server handle this? ie, user auths with Facebook, call server to register, and when it responds, they're finished.

andrewconner:
Since you're leaving them on the page for this, are you doing the same for twitter?

ahsu1230:
did we agree that we would use `signup/twitter` for authentication? and use cookies to redirect them back?

andrewconner:
If you want to do it that way for twitter, why not just do facebook the same way? Will be much easier to maintain in the future.

ahsu1230:
Ok that sounds good. The facebook pop-up gets blocked by ad-blockers anyway lol.

PR #13254, opened 2015-01-22
Remove mini library hover cards.

yipingliao:
@2is10 This is quick to revert if we change our minds later, but I am removing the mini hover cards in this pull.


2is10:
:+1:

PR #13256, opened 2015-01-22
use recommendationGenerationLock for throttling

ymatsuda:
@yingjieMiao @stkem 

recommendationGenerationLock was not actually used. Now the futures are created using recommendationGenerationLock.
I found the `clear` method of `ReactiveLock` is dangerous. It does not completes `Promise`s (thus `Future`s of them) in the task queue. They will never be completed once the task queue is cleared. I removed the call from `RecommendationGenerationCommander` because I don't think it is essential.


stkem:
The lock is used in line 206 in `precomputeRecommendationsForUser`.  Am I missing something?
You are right on `clear`, it should probably fail all futures still waiting to execute or not be used at all. In this case probably the latter.

ymatsuda:
Ah.  You are right. It was out of my IDE window. The line is too long. :(

PR #13258, opened 2015-01-22
Revert "use recommendationGenerationLock for throttling"

ymatsuda:
Reverts kifi/keepit#13256


ymatsuda:
@stkem @yingjieMiao 
something went wrong

PR #13259, opened 2015-01-22
Fix bug in library search chip profile link.

yipingliao:
@2is10 Bug was that the link would go to the user's own profile page. :(


2is10:
:+1:

PR #13260, opened 2015-01-22
making image info path be not optional and dropping the page keeper clie...

eishay:
...nt


eishay:
@andrewconner

andrewconner:
Wowza, nice.

andrewconner:
Reminder: needs SQL change as well (remove nullable).

eishay:
yes

PR #13261, opened 2015-01-22
twitter signup redirect (with query parameter)

ahsu1230:
I'm still testing and wanted to know if I'm on the right track.
@andrewconner 

cc: @yipingliao I've added install modal to library & profile controllers to appear only when there's `?install=1` as query parameter in url.

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/library.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/library.js
@@ -4,9 +4,9 @@ angular.module('kifi')
 
 .controller('LibraryCtrl', [
   '$scope', '$rootScope', '$analytics', '$location', '$state', '$stateParams', '$timeout', '$window',
-  'keepDecoratorService', 'libraryService', 'modalService', 'platformService', 'profileService', 'util',
+  'keepDecoratorService', 'libraryService', 'modalService', 'platformService', 'profileService', 'util', 'initParams', 'installService',
   function ($scope, $rootScope, $analytics, $location, $state, $stateParams, $timeout, $window,
-    keepDecoratorService, libraryService, modalService, platformService, profileService, util) {
+    keepDecoratorService, libraryService, modalService, platformService, profileService, util, initParams, installService) {
     //
     // Internal data.
     //
@@ -14,10 +14,35 @@ angular.module('kifi')
     var prePopulated = false;
     var authToken = $location.search().authToken || '';
 
-
     //
     // Internal functions
     //
+    function maybeShowInstallModal() {
+      if (!installService.installedVersion) {
+        if (installService.canInstall) {
+          if (installService.isValidChrome) {
+            $scope.platformName = 'Chrome';
+          } else if (installService.isValidFirefox) {
+            $scope.platformName = 'Firefox';
+          }
+          $scope.installExtension = installService.triggerInstall;
+          $scope.thanksVersion = 'installExt';
+        } else {
+          $scope.thanksVersion = 'notSupported';
+        }
+
+        if ($scope.library.id) {
+          $rootScope.$emit('trackLibraryEvent', 'view', { type: 'installLibrary' });
+        }
+
+        $scope.close = modalService.close;
+        modalService.open({
+          template: 'signup/thanksForRegisteringModal.tpl.html',
+          scope: $scope
+        });
+      }
+    }
+
     function trackPageView(attributes) {
       var url = $analytics.settings.pageTracking.basePath + $location.url();
       var library = $scope.library;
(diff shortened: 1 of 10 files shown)


kifi-backend/modules/shoebox/angular/src/signup/signupService.js
```diff
@@ -94,6 +94,12 @@ angular.module('kifi')
       }));
     };
 
+    $scope.getTwitterSignupUrl = function() {
+      var url = '/signup/twitter';
+      url += '?redirect=' + $location.url().split('?')[0]; // remove query params when redirecting back to this page
+      return url;
```
ahsu1230, line 100: step 1: Go to `signup/twitter?redirect=a.aron/asdf`

kifi-backend/modules/shoebox/app/com/keepit/controllers/core/AuthController.scala
```diff
@@ -309,12 +309,16 @@ class AuthController @Inject() (
...
-  def signup(provider: String) = Action.async(parse.anyContent) { implicit request =>
+  def signup(provider: String, redirect: Option[String] = None) = Action.async(parse.anyContent) { implicit request =>
     val authRes = ProviderController.authenticate(provider)
     authRes(request).map { result =>
       authHelper.authHandler(request, result) { (_, sess: Session) =>
         // TODO: set FORTYTWO_USER_ID instead of clearing it and then setting it on the next request?
-        result.withSession((sess + (SecureSocial.OriginalUrlKey -> routes.AuthController.signupPage().url)).deleteUserId)
+        val res = result.withSession((sess + (SecureSocial.OriginalUrlKey -> routes.AuthController.signupPage().url)).deleteUserId)
+        if (redirect.isDefined)
+          res.withCookies(Cookie("redirect", redirect.get))
```
ahsu1230, line 319: step 2: at `signup/twitter` you should see a redirect parameter. Set cookie

kifi-backend/modules/shoebox/app/com/keepit/controllers/core/AuthHelper.scala
```diff
@@ -210,11 +210,15 @@ class AuthHelper @Inject() (
...
     request.session.get("kcid").map(saveKifiCampaignId(user.id.get, _))
 
+    val cookieRedirect = request.cookies.get("redirect")
+
     Authenticator.create(newIdentity).fold(
       error => Status(INTERNAL_SERVER_ERROR)("0"),
       authenticator => {
         val result = if (isFinalizedImmediately) {
           Redirect(uri)
+        } else if (cookieRedirect.isDefined) {
```
ahsu1230: at `finishSignup()` prioritize if a uri (like `/install`) returns and redirect there? Otherwise, redirect to the library/profile path from cookie.

kifi-backend/modules/shoebox/angular/src/libraries/library.js
```diff
@@ -14,10 +14,35 @@ angular.module('kifi')
     var prePopulated = false;
     var authToken = $location.search().authToken || '';
 
-
     //
     // Internal functions
     //
+    function maybeShowInstallModal() {
+      if (!installService.installedVersion) {
```
ahsu1230: putting some of this logic into `installService` so both library & profiles can use it

ahsu1230:
tested normal signup with social networks & signing up through public library/profile page
merging now... to test fully in prod!

kifi-backend/modules/shoebox/app/com/keepit/controllers/core/AuthHelper.scala
```diff
@@ -210,10 +210,15 @@ class AuthHelper @Inject() (
...
     request.session.get("kcid").map(saveKifiCampaignId(user.id.get, _))
 
+    val cookieRedirect = request.cookies.get("redirect")
+
     Authenticator.create(newIdentity).fold(
       error => Status(INTERNAL_SERVER_ERROR)("0"),
       authenticator => {
-        val result = if (isFinalizedImmediately) {
+        val result = if (cookieRedirect.isDefined) {
+          val redirectUrl = java.net.URLDecoder.decode(cookieRedirect.get.value, "UTF-8") + "?install=1" // tell browser to pop-up install modal
```
2is10, line 219: @ahsu1230 Do we know for sure that the redirect URL doesn’t already contain any query parameters?

ahsu1230, line 219: for now yes. It's just a path to either a library page or user profile page
`%2Faaron%2Fsomelib`
`%2Faaron`
which then get decoded

PR #13265, opened 2015-01-22
Fix problem where a user's pic is not showing up correctly when she joins a library.

yipingliao:
@2is10 I found a couple of issues with new follower pics. Details to follow in comments.
cc @joshm1

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
@@ -200,6 +200,8 @@ angular.module('kifi')
         function augmentFollower(follower) {
           follower.picUrl = friendService.getPictureUrlForUser(follower);
           follower.profileUrl = routeService.getProfileUrl(follower.username);
+
+          return follower;
         }
 
         function preloadSocial() {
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
```diff
@@ -200,6 +200,8 @@ angular.module('kifi')
         function augmentFollower(follower) {
           follower.picUrl = friendService.getPictureUrlForUser(follower);
           follower.profileUrl = routeService.getProfileUrl(follower.username);
+
+          return follower;
```
yipingliao, line 204: In the event listener for `libraryJoined`, `augmentFollower()` is pushed into the array of followers but it was returning an `undefined` so the follower picture ended up being a generic ghost image.

kifi-backend/modules/shoebox/angular/src/recos/recoDecoratorService.js
```diff
@@ -41,6 +41,7 @@ angular.module('kifi')
         // All recommended libraries are published user-created libraries.
         this.recoLib.kind = 'user_created';
         this.recoLib.visibility = 'published';
+        this.recoLib.access = 'none';
```
yipingliao, line 44: The `libraryJoined` event listener does a check that previously the user had no access to the library. However, the access field was not defined for recommended libraries. @joshm1 I am hard-coding this here but please let me know if this should come from the server and if you'd like to send it over from the server side. Thanks!

2is10, line 44: Recommended libraries are using a different data model from the library page header, and yet they are both rendered using the same widget (directive)? Yikes.

Good catch, @yipingliao, and this looks like a reasonable solution. I don’t think the server would ever recommend a library the user owns or is already following.

2is10:
:+1:

PR #13269, opened 2015-01-23
more recall of related libraries

yingjieMiao:
@eishay 
As long as the library contains 1 classifiable URI, we will consider it as a vertex in the graph. 

cc: @stkem


yingjieMiao:
don't know the impact on the graph size, and quality impact. let's see.

yingjieMiao:
better recall now. But graph size also grew considerably. Computing the full graph, trimming and saving takes about 14 minutes now.

I can tweak up the threshold to 2 keeps. A library with 1 keep is trivial

stkem:
Could you start a google doc similar to the one for recos to track algorithm changes? That way we can correlate changes with patterns in the analytics after the fact much easier.
Maybe make it sub sheet in the recos tracking doc?

yingjieMiao:
sure.

yingjieMiao:
added a few.

PR #13270, opened 2015-01-23
Increase bottom margin on first panel of manage-lib modal

yipingliao:
... to completely cover content from second panel.

Thanks for spotting this, @leogrim !
@2is10

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.styl
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.styl
@@ -96,6 +96,9 @@
   &.kf-color
     font-size: 0
 
+  &:last-child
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.styl
```diff
@@ -96,6 +96,9 @@
   &.kf-color
     font-size: 0
 
+  &:last-child
+    margin-bottom: 25px
+
```
2is10, line 101: This is a roundabout and somewhat fragile way of addressing the issue. It looks like the followers list has a `height: 100%` style on it. I think the same style on this initial form would be a better solution. Change coming...

PR #13271, opened 2015-01-23
eliza endpoint to get latest unread messages

joshm1:
@stkem

### DIFF
--- a/kifi-backend/modules/eliza/app/com/keepit/eliza/model/UserThreadRepo.scala
+++ b/kifi-backend/modules/eliza/app/com/keepit/eliza/model/UserThreadRepo.scala
@@ -25,6 +25,8 @@ trait UserThreadRepo extends Repo[UserThread] with RepoWithDelete[UserThread] {
 
   def getUserThreads(userId: Id[User], uriId: Id[NormalizedURI])(implicit session: RSession): Seq[UserThread]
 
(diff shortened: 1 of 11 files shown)


kifi-backend/modules/eliza/app/com/keepit/eliza/model/UserThreadRepo.scala
```diff
@@ -161,6 +163,11 @@ class UserThreadRepoImpl @Inject() (
     (for (r <- rows if r.user === userId && r.uriId === uriId) yield r).list()
   }
 
+  def getLatestUnreadUnmutedThreads(userId: Id[User], howMany: Int)(implicit session: RSession): Seq[UserThread] = {
+    (for (row <- rows if row.user === userId && row.unread && row.muted === false) yield row).
+      sortBy { row => row.lastActive.desc }.take(howMany).list
```
stkem, line 168: lastActive is the last time the user was active in the thread, not the last time the thread was active overall. Just want to make sure that that is what you intended?

stkem:
:+1:

kifi-backend/modules/eliza/app/com/keepit/eliza/model/UserThreadRepo.scala
joshm1, line 168: ah, nope. my intention is the last time the thread was active

PR #13273, opened 2015-01-23
properly setting height of two panels in manage library modal

2is10:
@yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.styl
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.styl
@@ -96,9 +96,6 @@
   &.kf-color
     font-size: 0
 
-  &:last-child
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.styl
```diff
@@ -398,8 +396,9 @@ button.manage-lib-delete-button
 
 .manage-lib-members-content
   position: relative
-  margin-top: 10px
-  height: 100%
+  margin-top = 10px
+  margin-top: margin-top
+  height: 'calc(100% - %s)' % margin-top
```
2is10, line 401: the last 10px of this panel were being occluded. making it 10px shorter.

yipingliao:
Thanks!

PR #13281, opened 2015-01-23
simpler and faster related lib query

yingjieMiao:
@stkem @ymatsuda 

takes out the logic of querying library by topic Ids. This should have minimal affect on production quality


stkem:
Why take it out?

yingjieMiao:
the old algorithm has very low recall. Almost all of the related libs 
you see now are from the semantic graph.

the old algorithm uses partial information, the graph is constructed 
from full information.

the old algorithm is slower.

One advantage of old algorithm is (just occurred to me), it is more 
reactive to newly created libraries. (the topic ids are
updated on the fly. Graph is only recomputed on machine restart, or 
every N hours, currently N = 12.

stkem:
Got it. I don't think the 12 hour delay is a problem. We can always reduce the time interval if it becomes one.

PR #13283, opened 2015-01-23
Remove hover mini from left

yipingliao:
@andrewconner This removes the mini hover card from the left nav only. In addition, per Mark's advice, reducing/removing the transition delays for showing the mini hover card. Also, to match the mini hover card, I've reduced the delay for the user picture tooltip as well.


andrewconner:
Great, thanks.

PR #13284, opened 2015-01-23
using new custom content type for library

eishay:


eishay:
@leogrim

kifi-backend/modules/shoebox/app/com/keepit/commanders/PublicPageMetaTags.scala
```diff
@@ -115,7 +115,7 @@ case class PublicPageMetaFullTags(unsafeTitle: String, url: String, urlPathOnly:
   } else ""
 
   def formatOpenGraphForLibrary: String = {
-    formatOpenGraph("article") +
+    formatOpenGraph("fortytwoinc:library") +
       s"""
       |<meta name="author" content="$firstName $lastName">
       |<meta property="article:published_time" content="${ISO_8601_DAY_FORMAT.print(createdAt)}">
```
2is10, line 121: Does `article:published_time` make sense as a property if it’s not an `article`?

leogrim, line 121: Good catch, I think we need to add all properties to the new custom type on Facebook (for some reason it can't inherit from article)

PR #13285, opened 2015-01-23
making env a constant

2is10:
This makes the `$FBProvider` config block simpler and cleaner, avoiding duplication of logic.
@yipingliao


yipingliao:
Nice!

PR #13288, opened 2015-01-23
combining two body templates into one for now

2is10:
@yipingliao cc @andrewconner 

I think this is clearer for now. It makes the differences and similarities in the HTML structure more apparent.

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/layout/body.tpl.html
+++ b/kifi-backend/modules/shoebox/angular/src/layout/body.tpl.html
@@ -1,0 +1,10 @@
+<div ng-class="{'public-body': !$root.userLoggedIn}">
+  <header ng-include="$root.userLoggedIn ? 'layout/header/header.tpl.html' : 'layout/header/publicHeader.tpl.html'"></header>
+  <div class="kf-col-island">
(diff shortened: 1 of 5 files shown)


yipingliao:
:+1:

kifi-backend/modules/shoebox/angular/src/layout/body.tpl.html
```diff
@@ -0,0 +1,10 @@
+<div ng-class="{'public-body': !$root.userLoggedIn}">
+  <header ng-include="$root.userLoggedIn ? 'layout/header/header.tpl.html' : 'layout/header/publicHeader.tpl.html'"></header>
+  <div class="kf-col-island">
+    <aside class="kf-col kf-sidebar" ng-include="'layout/sidebar/sidebar.tpl.html'" ng-if="$root.userLoggedIn"></aside>
+    <section class="kf-scrollable-pane"><!--
```
andrewconner, line 5: Are the comments necessary? I believe the angular templates compiler does this? (removes whitespace)

andrewconner:
Works for me

kifi-backend/modules/shoebox/angular/src/layout/body.tpl.html
2is10, line 5: I don’t see why they’d be necessary, but I assumed the funky formatting in the old loggedIn/body.tpl.html was there for a reason and so was playing it safe. Happy to remove them.

PR #13290, opened 2015-01-23
eliminating analyticsState

2is10:
It doesn’t seem to be helping with anything.
@yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/invite/invite.js
+++ b/kifi-backend/modules/shoebox/angular/src/invite/invite.js
@@ -274,20 +274,18 @@ angular.module('kifi')
 ])
 
 .directive('kfFriendRequestBanner', [
-  '$analytics', '$timeout', 'analyticsState', 'initParams', 'keepWhoService',
+  '$analytics', '$timeout', 'initParams', 'keepWhoService',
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/angular/src/invite/invite.js
```diff
@@ -331,16 +329,14 @@ angular.module('kifi')
...
 
-      scope.friendRequestBannerHeader = 'Send a friend request';
-      scope.showFriendRequestBanner = true;
-      setupShowFriendRequestBanner(scope, externalId);
+      var externalId = initParams.friend;
+      if (externalId && /^[a-f0-9-]{36}$/.test(externalId)) {
+        scope.friendRequestBannerHeader = 'Send a friend request';
+        scope.showFriendRequestBanner = true;
+        setupShowFriendRequestBanner(scope, externalId, initParams.subtype || 'contactJoined');
+      }
```
2is10, line 339: The only functional change here is ignoring `friend` query param values that don’t look like external IDs. Passing the subtype into `setupShowFriendRequestBanner` so that the `initParams` reads are closer together.

yipingliao:
:+1:

PR #13292, opened 2015-01-23
removing a couple of unnecessary things

2is10:
@andrewconner


2is10:
Just reread `ng-cloak` doc. Leaving it as a matter of principle.

PR #13293, opened 2015-01-24
minor formatting of routes

2is10:
@yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/route.js
+++ b/kifi-backend/modules/shoebox/angular/src/route.js
@@ -63,68 +63,66 @@ angular.module('kifi')
         url: '/find?q&f',
         templateUrl: 'search/search.tpl.html',
         controller: 'SearchCtrl',
-
-        // SearchCtrl manually reloads search results on search input change in a debounced function.
-        reloadOnSearch: false
+        reloadOnSearch: false  // controller handles search query changes itself
       })
       .state('userProfile', {
-        url: '/:username?upb',
+        url: '/:username',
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/angular/src/route.js
```diff
@@ -63,68 +63,66 @@ angular.module('kifi')
...
         templateUrl: 'search/search.tpl.html',
         controller: 'SearchCtrl',
-
-        // SearchCtrl manually reloads search results on search input change in a debounced function.
-        reloadOnSearch: false
+        reloadOnSearch: false  // controller handles search query changes itself
       })
       .state('userProfile', {
-        url: '/:username?upb',
+        url: '/:username',
```
2is10, line 69: `upb` does not seem to be referenced anywhere else

kifi-backend/modules/shoebox/angular/src/route.js
```diff
@@ -63,68 +63,66 @@ angular.module('kifi')
...
         templateUrl: 'userProfile/userProfile.tpl.html',
         controller: 'UserProfileCtrl',
         'abstract': true
       })
```
2is10, line 73: The edits below are just flattening the indentation. Didn’t want to mix subsequent edits in with this simple formatting change.

kifi-backend/modules/shoebox/angular/src/route.js
yipingliao, line 69: Oops, thanks. That should have been removed as part of removing the hooks on profiles_beta experiment.

yipingliao:
:+1:

PR #13294, opened 2015-01-24
[DO NOT MERGE] Split up library card into libraryCard and libraryHeader

yipingliao:
@2is10 cc @andrewconner 

Split up our current `libraryCard` into `libraryCard` (for placement into a list of keep-like cards, i.e., recommendations) and `libraryHeader` (the header for a library).

Some notes:
- No CSS renaming yet - I want to get this in before it becomes too big, but as I expand the library header for the new layout, I am going to split up the CSS class names so that the two cards have separate and independent styles, like the related library cards and user profile library cards right now.
- `libraryHeader` template and JS files are mostly what was the old `libraryCard`, because the header contains a lot more functionality.
- The new `libraryCard` files are a stripped down version of the old `libraryCard`.
- The row of follower pictures have been refactored out into its own directive. It was the most complex part of the card that was shared, so I took it out. I may refactor out some smaller common things next week.

Please do not merge. This is kind of big and I want to merge it when there is more time in the day to debug if needed. Thanks!

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/library.styl
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/library.styl
@@ -597,9 +597,12 @@ html:not(.kf-mobile)
   .kf-keep-lib-stats-and-followers
     border-top: 1px solid #dadae2
     padding-top: 15px
+    height: 42px
+    overflow: hidden
 
   .kf-keep-lib-metrics
     display: inline-block
+    height: 42px
 
   .kf-keep-lib-keeps-followers-stats
     display: inline-block
--- a/kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
@@ -3,23 +3,15 @@
 angular.module('kifi')
 
 .directive('kfLibraryCard', [
-  '$FB', '$http', '$location', '$q', '$rootScope', '$state', '$stateParams', '$timeout', '$twitter', '$window',
-  'env', 'friendService', 'libraryService', 'modalService','profileService', 'platformService', 'signupService',
-  'routeService', 'util',
-  function ($FB, $http, $location, $q, $rootScope, $state, $stateParams, $timeout, $twitter, $window,
-            env, friendService, libraryService, modalService, profileService, platformService, signupService,
-            routeService, util) {
+  '$FB', '$q', '$rootScope', '$timeout', '$twitter', 'env', 'friendService', 'libraryService',
+  'modalService','profileService', 'routeService', 'util',
+  function ($FB, $q, $rootScope, $timeout, $twitter, env, friendService, libraryService,
+    modalService, profileService, routeService, util) {
     return {
       restrict: 'A',
       replace: true,
       scope: {
         library: '=',
-        username: '=',
-        librarySlug: '=',
-        recommendation: '=',
-        loading: '=',
-        toggleEdit: '=',
-        librarySearch: '=',
         followCallback: '&',
         clickLibraryCallback: '&'
       },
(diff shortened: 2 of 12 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/library.styl
```diff
@@ -597,9 +597,12 @@ html:not(.kf-mobile)
   .kf-keep-lib-stats-and-followers
     border-top: 1px solid #dadae2
     padding-top: 15px
+    height: 42px
```
yipingliao, line 600: These CSS additions are to make sure that when the row of followers shrinks upon the window width decreasing, we don't get a dropped-down row of followers before they become the new smaller width.

kifi-backend/modules/shoebox/angular/src/recos/recoDecoratorService.js
```diff
@@ -41,7 +41,6 @@ angular.module('kifi')
         // All recommended libraries are published user-created libraries.
         this.recoLib.kind = 'user_created';
         this.recoLib.visibility = 'published';
-        this.recoLib.access = 'none';
```
yipingliao, line 44: This is no longer needed, since now the logic for joining/leaving libraries are in two different files and we can use different logic for each different situation.

kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
```diff
@@ -29,64 +21,24 @@ angular.module('kifi')
         // Internal data.
         //
         var uriRe = /(?:\b|^)((?:(?:(https?|ftp):\/\/|www\d{0,3}[.])?(?:[a-z0-9](?:[-a-z0-9]*[a-z0-9])?\.)+(?:com|edu|biz|gov|in(?:t|fo)|mil|net|org|name|coop|aero|museum|a[cdegilmoqrstuz]|b[abefghimnrtyz]|c[acdfghiklmnoruxyz]|d[ejko]|e[cegst]|f[ijkmor]|g[befghilmnpqrstu]|h[kmnru]|i[delmnorst]|j[eop]|k[eghrwyz]|l[bciktuvy]|m[cdghkmnoqrstuwxyz]|n[acfilouz]|om|p[aeghklmnrty]|qa|r[eouw]|s[abcdeghikmnotuvz]|t[cdfhjmnoprtvwz]|u[agkmsyz]|v[eginu]|wf|y[t|u]|z[amrw]\b))(?::[0-9]{1,5})?(?:\/(?:[^\s()<>]*[^\s`!\[\]{};:.'",<>?«»()“”‘’]|\((?:[^\s()<>]+|(?:\([^\s()<>]+\)))*\))*|\b))(?=[\s`!()\[\]{};:.'",<>?«»“”‘’]|$)/;  // jshint ignore:line
```
2is10: `processUrls` and this regular expression would be great to factor out to `util`.

kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
```diff
@@ -163,7 +115,7 @@ angular.module('kifi')
           scope.library.followers.forEach(augmentFollower);
```
2is10, line 87: This code above probably doesn’t belong in the recommended library card:

``` js
// TODO(yiping): get real owner data when owner is not user.
if (!scope.library.owner) {
  scope.library.owner = profileService.me;
}
```

kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
```diff
@@ -771,11 +201,6 @@ angular.module('kifi')
           trackShareEvent('clickedShareTwitter');
         };
```
2is10, line 174: As I’m sure you’re probably well aware, the library sharing stuff look like a good candidate for factoring out too.

kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.tpl.html
```diff
@@ -73,56 +41,32 @@
...
             <span class="kf-keep-lib-footer-button-icon svg-twitter"></span>
             <span class="kf-keep-lib-footer-button-icon svg-twitter-white"></span><!-- for non-logged-in users -->
             <span class="kf-keep-lib-footer-share-link-text">Tweet</span>
           </a>
         </button>
-        <button ng-if="isPublic()" ng-click="shareFB()" class="kf-keep-lib-footer-button library-button-with-icon kf-keep-lib-footer-facebook-button">
+        <button ng-click="shareFB()" class="kf-keep-lib-footer-button library-button-with-icon kf-keep-lib-footer-facebook-button">
           <span class="kf-keep-lib-footer-button-icon svg-facebook"></span>
           <span class="kf-keep-lib-footer-button-icon svg-facebook-white"></span><!-- for non-logged-in users -->
           <a class="kf-keep-lib-footer-share-link-text" href="javascript:">Post</a>
```
2is10, line 72: When you duplicate and trim down the unused styles, might consider `kf-rlc-` as the prefix for the recommend library card, or `kf-lc-`. I think long prefixes can be counterproductive.

kifi-backend/modules/shoebox/angular/src/libraries/libraryHeader.js
```diff
@@ -0,0 +1,965 @@
...
+          }
+        };
+
+        scope.onChangeCoverImageMouseUp = function (event) {
+          if (event.which === 1) {
+            $timeout(hideCoverImageMenu);
+            angular.element('.kf-keep-lib-cover-image-file').click();
+            libraryService.trackEvent('user_clicked_page', scope.library, { action: 'clickedChangeCoverImage' });
+          }
+        };
```
2is10, line 688: My apologies for not better encapsulating all of this cover image stuff. It was my first angular feature. I’ll try to refactor it soon.

2is10:
Nice work, looking forward to the merge.

kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
yipingliao, line 87: Thanks for the catch. Will fix.

kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.tpl.html
yipingliao, line 72: Yup, will do.

yipingliao:
Closing to fix merge conflicts.

PR #13295, opened 2015-01-24
Handling signup redirecting & auto-follow libraries

ahsu1230:
@andrewconner 

Only applies to the twitter case, for now...

Case 1: Public Profile Page Signup
- click on `Sign up` on header -> should send `?redirect=a.aron`
- click on `Follow` on a library -> should send `?redirect=l/pubId&follow=1`

Case 2: Public Library Page Signup
- click on `Sign up` on header -> should send `?redirect=l/pubId`
- click on `Follow Library` on library card -> `?redirect=l/pubId&follow=1`
- click on `Create your own` on related cards -> `?redirect=l/pubId`

On backend,
on `/signup` we look for two parameters `redirect` & `follow`
If redirect exists and/or follow exists, set cookies

after authentication with social network,
we should read cookies for `redirect`
if `redirect` begins with `l/`... we know it's a libraryId. We take that libraryId to grab the librarypath, and if `follow` cookie is set, we auto follow.
Otherwise (does not begin with `l/`, simply redirect to the path.

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
@@ -787,7 +787,7 @@ angular.module('kifi')
             platformService.goToAppOrStore(url + (url.indexOf('?') > 0 ? '&' : '?') + 'follow=true');
             return;
           } else if ($rootScope.userLoggedIn === false) {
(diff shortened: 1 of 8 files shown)


ahsu1230:
Thanks for the advice @2is10 
will go ahead with the changes we discussed in another PR

PR #13297, opened 2015-01-24
unread messages in activity email

joshm1:
@stkem

### DIFF
--- a/kifi-backend/modules/eliza/app/com/keepit/eliza/model/UserThreadRepo.scala
+++ b/kifi-backend/modules/eliza/app/com/keepit/eliza/model/UserThreadRepo.scala
@@ -164,8 +164,8 @@ class UserThreadRepoImpl @Inject() (
   }
 
   def getLatestUnreadUnmutedThreads(userId: Id[User], howMany: Int)(implicit session: RSession): Seq[UserThread] = {
-    (for (row <- rows if row.user === userId && row.unread && row.muted === false) yield row).
-      sortBy { row => row.lastActive.desc }.take(howMany).list
+    (for (row <- rows if row.user === userId && row.unread && !row.muted) yield row).
+      sortBy { row => row.updatedAt.desc }.take(howMany).list
   }
 
   def getThreadIds(userId: Id[User], uriIdOpt: Option[Id[NormalizedURI]] = None)(implicit session: RSession): Seq[Id[MessageThread]] = {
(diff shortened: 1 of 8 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/emails/ActivityFeedEmailSender.scala
```diff
@@ -169,26 +214,31 @@ class ActivityFeedEmailSenderImpl @Inject() (
...
+    val maxLibrariesFollowedByFriends = 10
 
-    val maxNewFollowersOfLibraries = 5
+    val maxNewFollowersOfLibraries = 10
 
-    val maxNewFollowersOfLibrariesUsers = 5
+    val maxNewFollowersOfLibrariesUsers = 10
+
+    val minAgeOfUnreadNotificationThreads = currentDateTime.minusWeeks(6)
+    val maxAgeOfUnreadNotificationThreads = currentDateTime.minusHours(12)
```
joshm1, line 239: this time range was decided b/w Vivek and I before we saw the design that it may just be a number in the email. I'll clean this up after design is closer to final

kifi-backend/modules/shoebox/app/com/keepit/commanders/emails/ActivityFeedEmailSender.scala
```diff
@@ -169,26 +214,31 @@ class ActivityFeedEmailSenderImpl @Inject() (
...
+
+    val maxUnreadNotificationThreads = 10
```
joshm1, line 241: likewise, may not be needed

kifi-backend/modules/eliza/app/com/keepit/eliza/model/UserThreadRepo.scala
```diff
@@ -164,8 +164,8 @@ class UserThreadRepoImpl @Inject() (
   }
 
   def getLatestUnreadUnmutedThreads(userId: Id[User], howMany: Int)(implicit session: RSession): Seq[UserThread] = {
-    (for (row <- rows if row.user === userId && row.unread && row.muted === false) yield row).
-      sortBy { row => row.lastActive.desc }.take(howMany).list
+    (for (row <- rows if row.user === userId && row.unread && !row.muted) yield row).
+      sortBy { row => row.updatedAt.desc }.take(howMany).list
```
joshm1: this isn't the best since `updatedAt` can be out of sync with the thread's last activity, but this is the best w/o doing a join on the message_thread and/or message table to see when the last message was

PR #13299, opened 2015-01-24
auto following from redirects

ahsu1230:
On a public profile page:
- signup with Header -> send `/signup/twitter?redirect=/a.aron` (not encoded) to BE
- signup with Follow a Library -> `/signup/twitter?redirect=/a.aron/ramen&intent=follow`
  On a public library page:
- signup with Header -> `/signup/twitter?redirect=/a.aron/ramen`
- signup with Follow Library -> `/signup/twitter?redirect=/a.aron/ramen&intent=follow`
- signup with Create your own -> `/signup/twitter?redirect=/a.aron/ramen`

From backend, we first create cookies for `redirect` and `intent`
Then, after social network authentication, we read cookies.
If `intent` was set, we tack on `&intent=follow` after the redirected path

Once redirected to a library page, we check `intent`. 
If it's follow, the browser will call join library

@andrewconner @2is10

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/library.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/library.js
@@ -154,6 +154,9 @@ angular.module('kifi')
           if (initParams.install === '1' && !installService.installedVersion) {
             showInstallModal();
           }
+          if (initParams.intent === 'follow' && !$scope.library.isMine && $scope.library.access==='none') {
+            libraryService.joinLibrary($scope.library.id);
+          }
         });
       }
     });
--- a/kifi-backend/modules/shoebox/angular/src/libraries/library.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/library.js
@@ -154,7 +154,7 @@ angular.module('kifi')
           if (initParams.install === '1' && !installService.installedVersion) {
             showInstallModal();
           }
-          if (initParams.intent === 'follow' && !$scope.library.isMine && $scope.library.access==='none') {
+          if (initParams.intent === 'follow' && $scope.library.access==='none') {
             libraryService.joinLibrary($scope.library.id);
           }
         });
--- a/kifi-backend/modules/shoebox/angular/src/signup/signupService.js
+++ b/kifi-backend/modules/shoebox/angular/src/signup/signupService.js
@@ -49,9 +49,6 @@ angular.module('kifi')
     $scope.userData = $scope.userData || {};
     $scope.showTwitter = _.has($location.search(), 'twitter'); // todo (aaron): remove for twitter launch
 
-    var encodedCurrentPage = $window.encodeURIComponent($location.url().split('?')[0]); //remove query params when redirecting back to this page
-    $scope.twitterSignupUrl = '/signup/twitter?redirect=' + encodedCurrentPage; // twitter signup url with redirect
-
     function setModalScope($modalScope, onClose) {
       $modalScope.close = modalService.close;
       $modalScope.$on('$destroy', function () {
@@ -89,6 +86,13 @@ angular.module('kifi')
     var registerModal = function () {
       if ($scope.userData.libraryId) {
         $rootScope.$emit('trackLibraryEvent', 'view', { type: 'signupLibrary' });
+      }
+
+       // twitter signup url with redirect
+      var currentPath = $location.url().split('?')[0]; //remove query params when redirecting back to this page
+      $scope.twitterSignupUrl = '/signup/twitter?redirect=' + $window.encodeURIComponent(currentPath);
+      if ($scope.userData.intent === 'follow') {
+        $scope.twitterSignupUrl += '&intent=follow';
       }
 
       setModalScope(modalService.open({
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/core/AuthController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/core/AuthController.scala
@@ -309,16 +309,18 @@ class AuthController @Inject() (
     }
   }
 
-  def signup(provider: String, redirect: Option[String] = None) = Action.async(parse.anyContent) { implicit request =>
+  def signup(provider: String, redirect: Option[String] = None, intent: Option[String] = None) = Action.async(parse.anyContent) { implicit request =>
     val authRes = ProviderController.authenticate(provider)
     authRes(request).map { result =>
       authHelper.authHandler(request, result) { (_, sess: Session) =>
         // TODO: set FORTYTWO_USER_ID instead of clearing it and then setting it on the next request?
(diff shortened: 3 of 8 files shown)


kifi-backend/modules/shoebox/angular/src/signup/signupService.js
```diff
@@ -91,6 +88,13 @@ angular.module('kifi')
         $rootScope.$emit('trackLibraryEvent', 'view', { type: 'signupLibrary' });
       }
 
+       // twitter signup url with redirect
+      var currentPath = $location.url().split('?')[0]; //remove query params when redirecting back to this page
+      $scope.twitterSignupUrl = '/signup/twitter?redirect=' + $window.encodeURIComponent(currentPath);
```
2is10, line 93: Could we change `redirect` to `redirect_path` or `return_path`? I think it would be helpful to engineers unfamiliar with this flow to indicate in the name that the value is a URL path.

kifi-backend/modules/shoebox/app/com/keepit/controllers/core/AuthController.scala
```diff
@@ -401,10 +403,16 @@ class AuthController @Inject() (
...
+            val cookieRedirect = request.cookies.get("redirect")
+            if (cookieRedirect.isDefined) {
+              val redirectPath = java.net.URLDecoder.decode(cookieRedirect.get.value, "UTF-8")
+              val initParams = new StringBuilder("?m=0")
+
+              val cookieIntent = request.cookies.get("intent")
+              if (cookieIntent.isDefined) {
+                initParams ++= s"&intent=${cookieIntent.get.value}"
+              }
+              Redirect(s"${redirectPath}${initParams.toString}")
```
2is10, line 415: I think we should also be clearing the `redirect` and `intent` cookies here.

kifi-backend/modules/shoebox/angular/src/signup/signupService.js
2is10, line 93: Nevermind. I now see that, below, you use variables with `path` in the name when you read this parameter value, and that you’re also using the name `redirect` as a cookie name, and it’s nice for it to have the same name as this parameter and may not be worth changing both.

kifi-backend/modules/shoebox/app/com/keepit/controllers/core/AuthHelper.scala
```diff
@@ -210,14 +210,19 @@ class AuthHelper @Inject() (
...
-
     Authenticator.create(newIdentity).fold(
       error => Status(INTERNAL_SERVER_ERROR)("0"),
       authenticator => {
+        val cookieRedirect = request.cookies.get("redirect")
+        val cookieIntent = request.cookies.get("intent")
         val result = if (cookieRedirect.isDefined) {
-          val redirectUrl = java.net.URLDecoder.decode(cookieRedirect.get.value, "UTF-8") + "?install=1" // tell browser to pop-up install modal
-          Ok(Json.obj("uri" -> redirectUrl))
+          val redirectUrl = java.net.URLDecoder.decode(cookieRedirect.get.value, "UTF-8")
```
2is10, line 219: Let’s be as specific as possible, for clarity, and rename `redirectUrl` → `redirectPath`.

kifi-backend/modules/shoebox/angular/src/libraries/library.js
```diff
@@ -154,6 +154,9 @@ angular.module('kifi')
           if (initParams.install === '1' && !installService.installedVersion) {
             showInstallModal();
           }
+          if (initParams.intent === 'follow' && $scope.library.access==='none') {
```
2is10, line 157: style nit: spaces around `===`

2is10:
:+1: Nicely done.

Do you know if the cookies are marked server-only? If not, the webapp may as well read the cookies directly, rather than have the server transfer them to init params. That would simplify the server code. The webapp code would be different (reading cookies instead of init params), but no more complex.

As I mentioned in person, I’d consider it cleaner to add the information as parameters to the return URL that we send the social network, rather than stashing it in kifi.com cookies, since cookies are so broadly scoped. Unless a cookie has a path set, it could just as easily be intended for any or all endpoints on a site. Alternatively, consider scoping the cookies by setting the intended path on them. We already have a huge cookie problem on our site today:

image

ahsu1230:
@2is10 No I don't think they cookies are server-only. I'll take a look at the return path we send to social network authenticating. 

Could you explain more about what scoping a cookie would look like? Like, how could you set an intended path on a cookie?

ahsu1230:
Also, thanks for the comments! Addressing the smaller ones in #13310

2is10:
@ahsu1230 Here’s a brief read on how to scope a cookie.
http://en.wikipedia.org/wiki/HTTP_cookie#Domain_and_Path
The Play framework `Cookie` class has a `path` field:
https://www.playframework.com/documentation/2.2.x/api/scala/index.html#play.api.mvc.Cookie

andrewconner:
Our session cookies are (intentionally) HTTP only.

2is10:
I see, `httpOnly` is true by default, but we’d just need to add `httpOnly = false` to these two `Cookie` constructor calls above to change that. @andrewconner it sounds like you may have a reason not to consider this?

andrewconner:
Yes, I would prefer leaving it httpOnly. If we need a cookie shared between server/client, we should make a separate one. Even though it's signed, HTTP only cookies are [widely](https://www.ibm.com/developerworks/community/blogs/PortalL2Thoughts/entry/guarding_your_cookies_best_practices2?lang=en) [considered](http://www.troyhunt.com/2013/03/c-is-for-cookie-h-is-for-hacker.html) [best](http://coffeeonthekeyboard.com/session-fixation-and-hijacking-basic-security-part-6-801/) [practice](https://www.owasp.org/index.php/XSS_%28Cross_Site_Scripting%29_Prevention_Cheat_Sheet#Bonus_Rule_.231:_Use_HTTPOnly_cookie_flag), because without, a XSS can steal a user's (admin's) active session.

On that topic, we need to switch our cookie to SSL only. I recall it had something to do with social logins, but don't remember.

2is10:
@andrewconner We’re not talking about our Play session cookie. We’re talking about two new session cookies (in the traditional short-lived, non-Play sense): `redirect` and `intent`. Neither would contain anything sensitive.

2is10:
Still, I’d prefer not using cookies for this at all, and if we do use cookies, I don’t have a strong preference about how. Limiting them to a specific path (the return address from the social network auth flow) would be nice.

andrewconner:
I see, my previous comment is a bit over reaching. However, allowing them to be written from the client does potentially pose issues (open redirect, causing a user to auto-follow a library, etc).

Regarding not using cookies entirely, works for me if it's simple with our current backend.

2is10:
Sounds good.

FYI, `httpOnly` doesn’t provide any guarantees that values received by the server weren’t set by the client. It only helps prevent values set by the server from being stolen. So it doesn’t help mitigate any open redirect or auto-follow issues we might have.

PR #13301, opened 2015-01-24
persisting article open graph type into page info

eishay:
*\* When deploying we should **
1. apply the patch first
2. update shoebox
3. update scraper last


eishay:
@andrewconner @leogrim

leogrim:
I figured it out this morning! No need to supply to og:type explicitely, we can always use "object" but we still need a story for each subtype. E.g. the query parameter itself is "polymorphic", if we don't know the exact type we can just use "object" and then the more specific recipe will still match on their end. This is cool :)

eishay:
Oh, great!

eishay:
Should we revert this code? Is it useful?

andrewconner:
As-is, I believe this code would cause UriSummaryCommander to _not_ fetch from embedly if scraper beat it (which is often). So needs a bit of tweaking if we do merge this. Would need to look closer, will do.

leogrim:
We can just call the API immediately with "object" and "library" parameters, that should do it, much simpler, we shouldn't need this.

eishay:
Ok, so no merge.

leogrim:
I have already added quite a few custom stories for most common subtypes. What's cool is that the custom story rules are actually just defining the rendering of the data existing in open graph. They are independent of the API call that mutates the graph. The data supplied by our API calls just have to type check against the object and actions we defined - and always do if we use "object". 

Our backend doesn't have to know anything about these stories, it simply updates Facebook's graph. Thus on the other end, as we define more custom stories in Facebook's Kifi app, even prior events will be displayed accordingly in people's timelines.

PR #13311, opened 2015-01-26
Split up lib cards 2

yipingliao:
Like previous PR, but with merge conflicts resolved.

@2is10

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/libraryHeader.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/libraryHeader.js
@@ -1,0 +1,965 @@
+'use strict';
+
+angular.module('kifi')
+
+.directive('kfLibraryHeader', [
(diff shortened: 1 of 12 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/libraryHeader.js
```diff
@@ -0,0 +1,965 @@
...
+        //
+        // Internal methods.
+        //
+        function trackShareEvent(action) {
+          $timeout(function () {
+            $rootScope.$emit('trackLibraryEvent', 'click', { action: action });
+          });
+        }
+
+        function processUrls(text) {
```
2is10, line 65: Also in `util`?

yipingliao, line 65: Missed this one. A sign that the description part needs to be refactored as well!

PR #13312, opened 2015-01-26
Twitter archive import

andrewconner:
Currently ignores tweet text, does not follow redirects. Work in progress.

@stkem @eishay

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/common/social/twitter/RawTweet.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/common/social/twitter/RawTweet.scala
@@ -1,0 +1,200 @@
+package com.keepit.common.social.twitter
+
+import java.util.Locale
+
+import org.joda.time.DateTime
+import play.api.libs.functional.syntax._
+import play.api.libs.json._
+
+case class RawTweet(
+  id: RawTweet.TwitterId,
+  createdAt: DateTime,
+  text: String,
+  source: String,
+  user: RawTweet.User,
+  entities: RawTweet.Entities,
+  retweetedStatus: Option[RawTweet.RawRetweet],
+  favoriteCount: Option[Int],
+  retweetCount: Option[Int],
+  inReplyToStatusId: Option[RawTweet.TwitterId],
+  inReplyToUserId: Option[RawTweet.TwitterId],
+  inReplyToScreenName: Option[String],
+  lang: Option[String], // BCP 47
+  possiblySensitive: Option[Boolean])
+object RawTweet {
+  implicit def format: Reads[RawTweet] = (
+    (__ \ 'id_str).read[RawTweet.TwitterId] and
+    (__ \ 'created_at).read[DateTime](twitterDateReads(java.util.Locale.ENGLISH)) and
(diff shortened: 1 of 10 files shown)


andrewconner:
As a bonus, this includes a pretty thorough typesafeing of Twitter's tweet API response, so using this data in the future will be easier.

Todo:
- Redirect following
- Do not ignore tweet text
- Better file upload heuristics to determine what was uploaded

kifi-backend/modules/shoebox/app/com/keepit/common/social/twitter/RawTweet.scala
```diff
@@ -0,0 +1,200 @@
...
+      }
+
+      implicit def format: Reads[MediaEntity] = (
+        (__ \ 'id_str).read[TwitterId] and
+        (__ \ 'indices).read[TweetIndices] and
+        (__ \ 'media_url_https).read[String] and
+        (__ \ 'url).read[String] and
+        (__ \ 'display_url).read[String] and
+        (__ \ 'expanded_url).read[String] and
+        (__ \ 'sizes).read[Map[String, Size]].orElse(new Reads[Map[String, Size]] {
```
andrewconner, line 152: This is because it appears Twitter sometimes returns an object, sometimes an array of sizes.

kifi-backend/modules/shoebox/app/com/keepit/common/social/twitter/RawTweet.scala
```diff
@@ -0,0 +1,200 @@
...
+      implicit def format: Reads[UrlEntity] = (
+        (__ \ 'indices).read[TweetIndices] and
+        (__ \ 'url).read[String] and
+        (__ \ 'expanded_url).read[String] and
+        (__ \ 'display_url).read[String]
+      )(UrlEntity.apply _)
+    }
+  }
+
+  case class User(
```
stkem, line 191: I know it's in a different package/namespace, but perhaps call it `TwitterUser` just to avoid confusion?

tools/keepbot.py
```diff
@@ -81,7 +81,7 @@
 bookmarks = f.read()
 f.close()
 payload = {'file': bookmarks}
-url = 'https://www.kifi.com/site/keeps/file-import?public=1'
+url = 'https://www.kifi.com/site/libraries/lB8BIuUOCTi3/import-file'
```
stkem, line 84: Remember to update this on marvin too

stkem:
Nice!

kifi-backend/modules/shoebox/app/com/keepit/common/social/twitter/RawTweet.scala
andrewconner, line 191: This is _inside_ the `RawTweet` object — it won't autocomplete and must be referenced as `RawTweet.User` (unless someone imports `RawTweet._`...)

andrewconner:
Merging soon to keep the size down. Of note, this does not handle redirects yet or fixes the keep title scraper bug.

andrewconner:
It now (hopefully) fixes the keep title problem, thanks to help from @leogrim

PR #13313, opened 2015-01-26
Facebook Publishing

leogrim:
@eishay @andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/FacebookPublishingCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/FacebookPublishingCommander.scala
@@ -1,0 +1,86 @@
+package com.keepit.commanders
+
+import java.net.URLEncoder
+
+import com.google.inject.Inject
+import com.keepit.common.db.Id
+import com.keepit.common.db.slick.Database
+import com.keepit.common.healthcheck.{ AirbrakeNotifier, StackTrace }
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/FacebookPublishingCommander.scala
```diff
@@ -0,0 +1,86 @@
+package com.keepit.commanders
```
eishay, line 1: great, thanks for refactoring it out!

kifi-backend/modules/shoebox/app/com/keepit/commanders/PublicPageMetaTags.scala
```diff
@@ -105,7 +105,7 @@ case class PublicPageMetaFullTags(unsafeTitle: String, url: String, urlPathOnly:
   } getOrElse ""
 
   private def facebookIdTag = facebookId.map { id =>
-    s"""<meta property="article:author" content="$id">"""
+    s"""<meta property="library_owner" content="$id">"""
```
eishay, line 108: nice

andrewconner, line 108: Is this name fixed? Seems a bit weird to have a `library_owner` field be a Facebook ID.

PR #13296, opened 2015-01-24
using ui-router resolve to load user profile data

2is10:
@yipingliao

PR #13318, opened 2015-01-26
Revert "using ui-router resolve to load user profile data"

yipingliao:
Reverts kifi/keepit#13296  This breaks user profile pages
@2is10


2is10:
Thanks, and sorry. Investigating why the difference in dev v. prod.

PR #13319, opened 2015-01-26
Revert "Temporarily remove failing protractor test for user profile page...

yipingliao:
Reverts kifi/keepit#13316 Apparently this test _was_ catching something. Reverting removal of test...

@2is10


2is10:
:+1:

PR #13320, opened 2015-01-26
Update FB metadata

leogrim:


eishay:
:+1:

kifi-backend/modules/shoebox/app/com/keepit/commanders/PublicPageMetaTags.scala
```diff
@@ -105,7 +105,7 @@ case class PublicPageMetaFullTags(unsafeTitle: String, url: String, urlPathOnly:
   } getOrElse ""
 
   private def facebookIdTag = facebookId.map { id =>
-    s"""<meta property="library_owner" content="$id">"""
+    s"""<meta property="library:owner" content="$id">"""
```
eishay, line 108: should we have another property with a link to the owner profile ?

eishay, line 108: i.e. the kifi profile

leogrim, line 108: No, it's the Facebook id, as we used to have, but you're right it may make more sense to have the owner's Kifi profile.

PR #13324, opened 2015-01-26
using ui-router resolve to load user profile data

2is10:
re-applying #13296 with a dependency injection fix
@yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/route.js
+++ b/kifi-backend/modules/shoebox/angular/src/route.js
@@ -69,6 +69,12 @@ angular.module('kifi')
         url: '/:username',
         templateUrl: 'userProfile/userProfile.tpl.html',
         controller: 'UserProfileCtrl',
+        resolve: {
+          userProfileActionService: 'userProfileActionService',
+          profile: ['userProfileActionService', '$stateParams', function (userProfileActionService, $stateParams) {
+            return userProfileActionService.getProfile($stateParams.username);
+          }]
+        },
         'abstract': true
       })
       .state('userProfile.libraries', {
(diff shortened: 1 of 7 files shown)


kifi-backend/modules/shoebox/angular/src/route.js
```diff
@@ -69,6 +69,12 @@ angular.module('kifi')
         url: '/:username',
         templateUrl: 'userProfile/userProfile.tpl.html',
         controller: 'UserProfileCtrl',
+        resolve: {
+          userProfileActionService: 'userProfileActionService',
+          profile: ['userProfileActionService', '$stateParams', function (userProfileActionService, $stateParams) {
+            return userProfileActionService.getProfile($stateParams.username);
+          }]
```
2is10, line 76: This fix is using quoted strings for the names of the injected arguments here, as we do elsewhere.

2is10:
The protractor test that failed before passes with this fix, and I’ve manually verified that user profile navigations work by pausing the protractor tests using `browser.pause()`.

kifi-backend/modules/shoebox/angular/src/route.js
yipingliao, line 76: Ah, that's why it breaks with prod assets!

yipingliao:
:+1:

kifi-backend/modules/shoebox/angular/src/route.js
2is10, line 76: @yipingliao Yeah, unfortunately, none of the ui-router examples I saw used this approach, so I thought it must somehow be exempt. Silly me.

PR #13326, opened 2015-01-26
Fix refresh-social-networks modal CSS class names

yipingliao:
@2is10 Leo has found a bug where the modal content was hidden by an ad-blocker extension due to clashing CSS class names.

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/social/connectNetworks.styl
+++ b/kifi-backend/modules/shoebox/angular/src/social/connectNetworks.styl
@@ -1,41 +1,30 @@
-.social-connect
+.kf-social-connect
   min-width: 430px
   margin: 0 auto
   text-align: center
   width: 0px
 
-.social-connect-title
+.kf-social-connect-title
   font-size: 24px
   font-weight: 300
   font-family: Lato, Arial, sans-serif
   font-style: italic
 
-.social-connect-desc
+.kf-social-connect-desc
   font-size: 16px
   margin: 15px auto 35px
   max-width: 350px
 
-.friends-wrapper, .invite-wrapper
-  .social-connect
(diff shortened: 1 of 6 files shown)


kifi-backend/modules/shoebox/angular/src/social/connectNetworks.styl
```diff
@@ -1,41 +1,30 @@
...
   font-family: Lato, Arial, sans-serif
   font-style: italic
 
-.social-connect-desc
+.kf-social-connect-desc
   font-size: 16px
   margin: 15px auto 35px
   max-width: 350px
 
-.friends-wrapper, .invite-wrapper
```
yipingliao, line 18: From here until line 27, these styles are not being used.

kifi-backend/modules/shoebox/angular/src/social/connectNetworks.tpl.html
```diff
@@ -1,114 +1,114 @@
-<div class="social-connect" ng-show="data.show">
```
yipingliao, line 1: Turned out that this `ng-show` is a no-op.

kifi-backend/modules/shoebox/angular/src/social/social.js
```diff
@@ -11,9 +11,6 @@ angular.module('kifi')
       restrict: 'A',
       templateUrl: 'social/connectNetworks.tpl.html',
       link: function (scope/*, element, attrs*/) {
-        scope.data = scope.data || {};
-        scope.data.show = true;
```
yipingliao, line 15: Related to previous comment re no-op `ng-show`.

2is10:
LGTM as long as everything still works. :wink:

PR #13329, opened 2015-01-26
removing old redundant "secret" key from libraries in website search results

2is10:
@leogrim 
cc @yipingliao 

We also need to add the library color. I would have, but it didn’t appear to be readily available. We should probably add it to `BasicLibrary` since the color will soon be displayed everywhere a library is.

I’ve verified that the website isn’t using the `"secret"` key. It’s only using `"visibility"`.


leogrim:
I can add library color to `BasicLibrary`, when do you need this?

2is10:
@leogrim Thanks. It’s not urgent. Some time this week?

PR #13330, opened 2015-01-26
caching samples instead of density

ymatsuda:
@leogrim @yingjieMiao @stkem 

another memory saving idea on graph
- caching a batch of samples instead of the probability density of a transition
- the batch are recomputed when cached sample is exhausted
- the batch size is doubled each time recomputation happens


yingjieMiao:
lgtm

PR #13331, opened 2015-01-27
fix the clear method of ReactiveLock

ymatsuda:
@stkem 
- make all tasks in the queue fail
- maintain waitingCount (we have a code that relies on this)


stkem:
lgtm. Thanks!

PR #13335, opened 2015-01-27
Library Score normalization

leogrim:
@ymatsuda @yingjieMiao @stkem 
Basic heuristics, comments welcome :)

### DIFF
--- a/kifi-backend/modules/search/app/com/keepit/search/engine/library/LibraryFromKeepsScoreVectorSource.scala
+++ b/kifi-backend/modules/search/app/com/keepit/search/engine/library/LibraryFromKeepsScoreVectorSource.scala
@@ -7,7 +7,7 @@ import com.keepit.search.{ SearchFilter, SearchConfig }
 import com.keepit.search.index.graph.keep.KeepFields
 import com.keepit.search.index.{ Searcher, WrappedSubReader }
 import com.keepit.search.util.join.{ DataBuffer, DataBufferWriter }
-import org.apache.lucene.index.AtomicReaderContext
+import org.apache.lucene.index.{Term, AtomicReaderContext}
 import org.apache.lucene.search.DocIdSetIterator.NO_MORE_DOCS
 import org.apache.lucene.search.{ Query, Scorer }
 import scala.concurrent.Future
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/search/app/com/keepit/search/engine/library/LibraryFromKeepsScoreVectorSource.scala
```diff
@@ -52,7 +52,9 @@ class LibraryFromKeepsScoreVectorSource(
         val visibility = keepVisibilityEvaluator(docId, libId)
 
         if (visibility != Visibility.RESTRICTED) {
-          val boost = getRecencyBoost(recencyScorer, docId)
+          val recencyBoost = getRecencyBoost(recencyScorer, docId)
+          val inverseLibraryFrequencyBoost = 1.0f / (1 + Math.log(1 + searcher.freq(new Term(KeepFields.libraryField, libId.toString))).toFloat) // number of keeps from this shard in this library (correlated with number of keeps in this library)
+          val boost = recencyBoost * inverseLibraryFrequencyBoost
```
yingjieMiao, line 58: I like the idea of idf. I assume the `IDF` is balanced with `TF` somehow (via searcher).

leogrim, line 58: More via the join performed by the engine: if many keeps match the query in a specific library, they all end up contributing their score for this specific library. An alternative is to perform this boost at the collector level (cheaper, maybe simpler), though those contributions from keeps in huge libraries (Twitter, bookmark imports etc...) are what we really want to tame.

kifi-backend/modules/search/app/com/keepit/search/engine/query/KPrefixQuery.scala
```diff
@@ -85,8 +87,9 @@ class KPrefixScorer(weight: KPrefixWeight, subScorer: Scorer, queryTerms: Array[
   override def score(): Float = {
     val name = getName()
     val distance = PrefixMatching.distance(name, queryTerms)
-    val boost = (PrefixMatching.maxDist - distance).toFloat / PrefixMatching.maxDist // todo(Léo): boost shorter names
-    val score = subScorer.score() * boost
+    val prefixSimilarity = (PrefixMatching.maxDist - distance).toFloat / PrefixMatching.maxDist
+    val lengthSimilarity = (1.0f + weight.query.length) / (1 + name.length)
```
yingjieMiao, line 91: Is this `lengthSim` always in [0,1] ?

yingjieMiao:
lgtm!

kifi-backend/modules/search/app/com/keepit/search/engine/query/KPrefixQuery.scala
leogrim, line 91: It should since it's a prefix search. We can't get a match if the length of the query is longer than the length of the matched field.

PR #13336, opened 2015-01-27
updating gulp-minify-html to fix stripped whitespace issue in prod

2is10:
@andrewconner @yipingliao 
I already had 0.1.8 locally, but marvin has 0.1.4. This should get everybody onto the latest release with consistent whitespace handling (collapsing to a single space, like the browser).

### DIFF
--- a/kifi-backend/modules/shoebox/angular/package.json
+++ b/kifi-backend/modules/shoebox/angular/package.json
@@ -35,7 +35,7 @@
     "gulp-ignore": "^1.2.0",
     "gulp-jshint": "~1.8.0",
     "gulp-livereload": "~2.1.0",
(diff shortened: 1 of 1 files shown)


andrewconner:
Nice, good catch. Not sure I would have realized that one.

2is10:
Note: This may cause some new layout issues in prod (making prod look like my dev has). We’ll need to fix them as we find them. Good thing we’re redesigning much of the site soon.

Also, you may start seeing me add HTML comments around line breaks to avoid introducing whitespace-only text nodes. Our minifier strips these comments out, and I prefer that approach to funky wrapping. Another approach that works sometimes is setting `font-size: 0` on a parent of the whitespace and then setting the appropriate `font-size` on siblings that need it.

PR #13338, opened 2015-01-27
no longer need to use ng-src or ng-href for safety here

2is10:
...since a recommended library card is only ever rendered when its data is already present.
@yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.tpl.html
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.tpl.html
@@ -9,16 +9,16 @@
       <div class="kf-keep-lib-body">
         <div class="kf-keep-lib-header">
           <div class="kf-keep-lib-name-container kf-keep-title">
-            <a class="kf-keep-lib-name" ng-href="{{library.url}}" ng-attr-title="{{library.name}}" ng-bind-html="library.name" ng-click="clickLibraryCallback()"></a>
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.tpl.html
```diff
@@ -9,16 +9,16 @@
       <div class="kf-keep-lib-body">
         <div class="kf-keep-lib-header">
           <div class="kf-keep-lib-name-container kf-keep-title">
-            <a class="kf-keep-lib-name" ng-href="{{library.url}}" ng-attr-title="{{library.name}}" ng-bind-html="library.name" ng-click="clickLibraryCallback()"></a>
+            <a class="kf-keep-lib-name" href="{{library.url}}" ng-click="clickLibraryCallback()">{{library.name}}</a>
```
2is10, line 12: Removed the `title` attribute. I find it annoying on most libraries. I think `kfEllipsis` should add it only if text is elided. I’ll take care of that eventually.

2is10, line 12: Also rendering `library.name` as plain text instead of HTML. Important distinction. My [HTML Examples (e.g. &lt;b&gt; and &amp;quot;)](https://www.kifi.com/jaredfortytwo/html-examples) library would have rendered incorrectly here.

yipingliao:
Thanks!

PR #13339, opened 2015-01-27
ensuring that the top margin does not animate on initial render

2is10:
@yipingliao 

May rename the `kf-loaded` class for more clarity in the future.


yipingliao:
:+1:

PR #13343, opened 2015-01-27
Fix typo

leogrim:
@eishay

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/PageMetaTagsCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/PageMetaTagsCommander.scala
@@ -149,7 +149,7 @@ class PageMetaTagsCommander @Inject() (
         unsafeTitle = s"${user.firstName} ${user.lastName}",
         url = url,
         urlPathOnly = url,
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/PageMetaTagsCommander.scala
```diff
@@ -149,7 +149,7 @@ class PageMetaTagsCommander @Inject() (
         unsafeTitle = s"${user.firstName} ${user.lastName}",
         url = url,
         urlPathOnly = url,
-        unsafeDescription = s"${user.firstName} ${user.lastName} is on Kifi. Join Kifi to connect with Eishay Smith and others you may know. Kifi connects people with knowledge.",
+        unsafeDescription = s"${user.firstName} ${user.lastName} is on Kifi. Join Kifi to connect with ${user.firstName} ${user.lastName} and others you may know. Kifi connects people with knowledge.",
```
eishay, line 152: wow, thanks!

2is10, line 152: Better not to repeat the last name in the second sentence?

eishay, line 152: works for FB

PR #13349, opened 2015-01-28
User Persona Commander basics

ahsu1230:
`UserPersonaCommander` has 3 main methods
- `getUserPersonas(userId)` returns `Seq[Persona]`
- `addUserPersonas(userId, Set[Persona])` returns `(Seq[Persona], Seq[Library])` list of personas and libraries that were actually added
- `removeUserPersonas(userId, Set[Persona])` returns `Seq[Persona]` list of personas actually removed

@andrewconner @stkem @yingjieMiao

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/UserPersonaCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/UserPersonaCommander.scala
@@ -1,0 +1,71 @@
+package com.keepit.commanders
+
+import com.google.inject.{ Provider, Inject }
+import com.keepit.common.db.Id
+import com.keepit.common.db.slick.Database
+import com.keepit.common.logging.Logging
+import com.keepit.common.time.Clock
+import com.keepit.heimdal.HeimdalContext
+import com.keepit.model._
+
+class UserPersonaCommander @Inject() (
+    db: Database,
+    userPersonaRepo: UserPersonaRepo,
+    personaRepo: PersonaRepo,
+    libraryCommander: LibraryCommander,
+    clock: Clock) extends Logging {
+
+  def getPersonasForUser(userId: Id[User]): Seq[Persona] = {
+    db.readOnlyMaster { implicit s =>
+      val personaIds = userPersonaRepo.getUserPersonas(userId).toSet
+      personaRepo.getPersonasByIds(personaIds)
+    }
+  }
+
+  def addPersonasForUser(userId: Id[User], personas: Set[String])(implicit context: HeimdalContext): (Seq[Persona], Seq[Library]) = {
+    val (newPersonas, currentPersonaIds) = db.readOnlyMaster { implicit s =>
+      val newPersonas = personaRepo.getByNames(personas)
+      val currentPersonaIds = userPersonaRepo.getUserPersonas(userId)
+      (newPersonas, currentPersonaIds)
+    }
+    val personasToPersist = newPersonas.filterNot(p => currentPersonaIds.contains(p.id.get))
+    db.readWrite { implicit s =>
+      personasToPersist.map { persona =>
+        userPersonaRepo.getByUserAndPersona(userId, persona.id.get) match {
+          case None =>
+            userPersonaRepo.save(UserPersona(userId = userId, personaId = persona.id.get, state = UserPersonaStates.ACTIVE))
+          case Some(up) =>
+            userPersonaRepo.save(up.copy(state = UserPersonaStates.ACTIVE))
+        }
+      }
+    }
+
+    // create libraries based on added personas
+    val librariesAdded = personasToPersist.map { persona =>
+      val libraryAddReq = LibraryAddRequest(persona.name, LibraryVisibility.PUBLISHED, None, persona.name)
+      libraryCommander.addLibrary(libraryAddReq, userId)
+    }.collect {
+      case Right(lib) => lib
+    }
+    (personasToPersist, librariesAdded)
+  }
+
+  def removePersonasForUser(userId: Id[User], personas: Set[String]): Seq[Persona] = {
+    val (newPersonas, currentPersonaIds) = db.readOnlyMaster { implicit s =>
+      val newPersonas = personaRepo.getByNames(personas)
+      val currentPersonaIds = userPersonaRepo.getUserPersonas(userId)
+      (newPersonas, currentPersonaIds)
+    }
+    val personasToRemove = newPersonas.filter(p => currentPersonaIds.contains(p.id.get))
+    db.readWrite { implicit s =>
+      personasToRemove.map { persona =>
+        userPersonaRepo.getByUserAndPersona(userId, persona.id.get) match {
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/shoebox/app/com/keepit/model/UserPersonaRepo.scala
```diff
@@ -37,15 +37,24 @@ class UserPersonaRepoImpl @Inject() (
   def deleteCache(model: UserPersona)(implicit session: RSession): Unit = {}
   def invalidateCache(model: UserPersona)(implicit session: RSession): Unit = {}
 
+  private val getByUserAndPersonaCompiled = Compiled { (userId: Column[Id[User]], personaId: Column[Id[Persona]]) =>
```
ahsu1230, line 40: using pre-compiled slick queries!

stkem:
lgtm! (although I expect this will see some tweaks as the client code materializes)

kifi-backend/modules/shoebox/app/com/keepit/commanders/UserPersonaCommander.scala
```diff
@@ -0,0 +1,71 @@
...
+import com.keepit.heimdal.HeimdalContext
+import com.keepit.model._
+
+class UserPersonaCommander @Inject() (
+    db: Database,
+    userPersonaRepo: UserPersonaRepo,
+    personaRepo: PersonaRepo,
+    libraryCommander: LibraryCommander,
+    clock: Clock) extends Logging {
+
```
yingjieMiao, line 17: it's probably helpful to have a trait for commander so its functions are clearly listed

kifi-backend/modules/shoebox/app/com/keepit/commanders/UserPersonaCommander.scala
```diff
@@ -0,0 +1,71 @@
...
+  def getPersonasForUser(userId: Id[User]): Seq[Persona] = {
+    db.readOnlyMaster { implicit s =>
+      val personaIds = userPersonaRepo.getUserPersonas(userId).toSet
+      personaRepo.getPersonasByIds(personaIds)
+    }
```
yingjieMiao, line 22: probably helpful to cache all personas by ids? not urgent

kifi-backend/modules/shoebox/app/com/keepit/commanders/UserPersonaCommander.scala
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  def addPersonasForUser(userId: Id[User], personas: Set[String])(implicit context: HeimdalContext): (Seq[Persona], Seq[Library]) = {
```
yingjieMiao, line 25: is it possible to combine addPersona and removePersona into one method? probably shorter/cleaner code?

stkem, line 25: Not sure I'm following. What would the signature look like? (Of not is that adding a persona creates a library, whereas removing the persona does _not_ remove the library, which makes unification difficult)

yingjieMiao, line 25: signature could be `modifyPersona(personas: Set[_])`, where `personas` defines currently selected personas on the UI. It should be easy to figure out what is being added, what is being removed. And Library creation is purely orthogonal to personaRepo editing. 

Moreover, 

```
val (newPersonas, currentPersonaIds) = db.readOnlyMaster { implicit s =>
```

this part is repeated exactly in two methods. 

Just a suggestion, if it can simplify the code.

ahsu1230, line 25: For website, the UI does not have a "save" button. So that means as users active/deactivate personas one a time, we should already be sending calls to the server (whether user activated or deactivated a certain persona)
If we had a `modifyPersona()` that takes in a Set, we would be sending a lot of redundant information for every persona the user checks off.

I personally like having endpoints be as separated and independent so it's clear to front-end people what kind of endpoint they're calling.
However, mobile may have something like a Save button or only persist Sets at a time because they don't want to be chatty with endpoints. So it's possible they'll need a `modifyPersona()` method.
We'll see when we get there! =)

kifi-backend/modules/shoebox/app/com/keepit/commanders/UserPersonaCommander.scala
andrewconner, line 17: +1, helps with compile time too!

kifi-backend/modules/shoebox/app/com/keepit/commanders/UserPersonaCommander.scala
andrewconner, line 22: Can this just be a join, so a single call? If so, is this method needed, since it'll just be a query left?

andrewconner:
Sorry I didn't see this sooner.

Would be nice to type safe a persona name, so we could enforce some data integrity rules — we don't want clients being able to add whatever they want to the database directly.

kifi-backend/modules/shoebox/app/com/keepit/commanders/UserPersonaCommander.scala
```diff
@@ -0,0 +1,71 @@
...
+      libraryCommander.addLibrary(libraryAddReq, userId)
+    }.collect {
+      case Right(lib) => lib
+    }
+    (personasToPersist, librariesAdded)
+  }
+
+  def removePersonasForUser(userId: Id[User], personas: Set[String]): Seq[Persona] = {
+    val (newPersonas, currentPersonaIds) = db.readOnlyMaster { implicit s =>
+      val newPersonas = personaRepo.getByNames(personas)
```
andrewconner, line 55: "new"?

kifi-backend/modules/shoebox/app/com/keepit/model/PersonaRepo.scala
```diff
@@ -1,13 +1,18 @@
...
 import com.keepit.common.db.slick.DBSession.RSession
 import com.keepit.common.db.slick.{ DataBaseComponent, DbRepo }
 import com.keepit.common.healthcheck.AirbrakeNotifier
 import com.keepit.common.time.Clock
 
 @ImplementedBy(classOf[PersonaRepoImpl])
-trait PersonaRepo extends DbRepo[Persona]
+trait PersonaRepo extends DbRepo[Persona] {
+  def getByName(name: String)(implicit session: RSession): Option[Persona]
+  def getByNames(names: Set[String])(implicit session: RSession): Seq[Persona]
```
andrewconner, line 13: Since this takes in a `Set`, should also return something unordered. For use cases like this, we've done `Map[String, Persona` (if it's important to easily go from name to object) or `Set[Persona]` (if only the collection matters).

kifi-backend/modules/shoebox/app/com/keepit/commanders/UserPersonaCommander.scala
ahsu1230, line 55: woops copy error

PR #13350, opened 2015-01-28
using ui-router resolve to load library

2is10:
@yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/common/services/mixpanelService.js
+++ b/kifi-backend/modules/shoebox/angular/src/common/services/mixpanelService.js
@@ -17,23 +17,19 @@
       $analyticsProvider.settings.pageTracking.autoTrackVirtualPages = false;
     }
   ])
-  .run(['$rootScope', '$window', '$log', 'profileService', '$http', 'env',
-    function (_$rootScope_, _$window_, _$log_, _profileService_, _$http_, _env_) {
+  .run(['$window', '$log', 'profileService', '$http', 'env',
+    function (_$window_, _$log_, _profileService_, _$http_, _env_) {
       $window = _$window_;
       $log = _$log_;
       profileService = _profileService_;
       $http = _$http_;
       env = _env_;
-
-      $rootScope = _$rootScope_;
-      $rootScope.$on('currentLibraryChanged', registerCurrentLibrary);
     }
   ]);
 
-  var $window, $rootScope, $log, profileService, $http, env;  // injected before any code below runs
+  var $window, $log, profileService, $http, env;  // injected before any code below runs
   var identifiedViewEventQueue = [];
   var userId;
-  var currentLibrary;
 
   var locations = {
     yourKeeps: /^\/$/,
@@ -45,10 +41,6 @@
     helpRankClicks: /^\/helprank\/clicks?$/,
     helpRankReKeeps: /^\/helprank\/rekeeps?$/
   };
-
-  function registerCurrentLibrary(event, library) {
-    currentLibrary = library;
-  }
 
   function trackEventThroughProxy(event, properties)  {
     return $http.post(env.xhrBase + '/events', [{
@@ -89,9 +81,6 @@
   }
 
   function getLocation(path) {
-    if (currentLibrary) {
-      return (profileService.me && profileService.me.id) ? 'library' : 'libraryLanding';
-    }
     for (var loc in locations) {
       if (locations[loc].test(path)) {
         return loc;
@@ -120,12 +109,12 @@
         $log.log('mixpanelService.pageTrackForUser(' + path + '):' + origin);
 
         attributes = _.extend({
-          type: getLocation(path),
+          type: attributes.type || getLocation(path),
           origin: origin,
           siteVersion: 2,
           userStatus: getUserStatus(),
           experiments: getExperiments()
-        }, attributes || {});
+        }, attributes);
 
         trackEventThroughProxy('user_viewed_page', attributes);
       } finally {
@@ -163,10 +152,10 @@
     $log.log('mixpanelService.pageTrackForVisitor(' + path + '):' + origin);
 
     attributes = _.extend({
-      type: getLocation(path),
+      type: attributes.type || getLocation(path),
       origin: origin,
       siteVersion: 2
-    }, attributes || {});
+    }, attributes);
 
     mixpanel.track('visitor_viewed_page', attributes);
   }
--- a/kifi-backend/modules/shoebox/angular/src/libraries/library.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/library.js
@@ -3,16 +3,14 @@
 angular.module('kifi')
 
 .controller('LibraryCtrl', [
-  '$scope', '$rootScope', '$analytics', '$location', '$state', '$stateParams', '$timeout', '$window',
+  '$scope', '$rootScope', '$analytics', '$location', '$state', '$stateParams', '$timeout', '$window', 'library',
   'keepDecoratorService', 'libraryService', 'modalService', 'platformService', 'profileService', 'util', 'initParams', 'installService',
-  function ($scope, $rootScope, $analytics, $location, $state, $stateParams, $timeout, $window,
+  function ($scope, $rootScope, $analytics, $location, $state, $stateParams, $timeout, $window, library,
     keepDecoratorService, libraryService, modalService, platformService, profileService, util, initParams, installService) {
     //
     // Internal data.
     //
     var selectedCount = 0;
-    var prePopulated = false;
-    var authToken = $location.search().authToken || '';
 
     //
     // Internal functions
@@ -26,7 +24,7 @@ angular.module('kifi')
         $scope.thanksVersion = 'notSupported';
       }
 
-      if ($scope.library.id) {
+      if (library.id) {
         $rootScope.$emit('trackLibraryEvent', 'view', { type: 'installLibrary' });
       }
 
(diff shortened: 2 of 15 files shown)


kifi-backend/modules/shoebox/angular/src/common/services/mixpanelService.js
```diff
@@ -17,23 +17,19 @@
...
+  .run(['$window', '$log', 'profileService', '$http', 'env',
+    function (_$window_, _$log_, _profileService_, _$http_, _env_) {
       $window = _$window_;
       $log = _$log_;
       profileService = _profileService_;
       $http = _$http_;
       env = _env_;
-
-      $rootScope = _$rootScope_;
-      $rootScope.$on('currentLibraryChanged', registerCurrentLibrary);
```
2is10, line 29: No longer broadcasting `currentLibraryChanged` since it was only used in this file to determine the `type` for page views and events, and it seemed better to pre-populate the `type` in the library event attributes instead.

kifi-backend/modules/shoebox/angular/src/common/services/mixpanelService.js
```diff
@@ -120,12 +109,12 @@
         $log.log('mixpanelService.pageTrackForUser(' + path + '):' + origin);
 
         attributes = _.extend({
-          type: getLocation(path),
+          type: attributes.type || getLocation(path),
```
2is10, line 112: a minor optimization for when `.type` is pre-populated

kifi-backend/modules/shoebox/angular/src/libraries/library.js
```diff
@@ -139,29 +133,26 @@ angular.module('kifi')
...
+    $scope.callImportBookmarks = function () {
+      $rootScope.$emit('showGlobalModal', 'importBookmarks');
+    };
+    $scope.callImportBookmarkFile = function () {
+      $rootScope.$emit('showGlobalModal', 'importBookmarkFile');
+    };
+    $scope.callTriggerInstall = function () {
+      $rootScope.$emit('triggerExtensionInstall');
+    };
+
```
2is10, line 148: These four methods were just moved up here (the methods section) from below. No changes.

kifi-backend/modules/shoebox/angular/src/libraries/library.js
```diff
@@ -139,29 +133,26 @@ angular.module('kifi')
...
 
     //
     // Watches and listeners.
     //
-    $scope.$watch('library.id', function (id) {
```
2is10, line 146: This watch struck me as odd. Seems like it should have only ever fired once, when the library associated with this controller instance first loaded. Now that the library is already loaded before we get here, I’ve just moved the code below that was inside this watch handler (the loading of related libraries) down to the initialization section.

yipingliao:
Wow, nice! I will need some time to comb through this in detail to make sure I understand everything. :) I like the clean up of all the "attempts to pre-populate" the library and the new passphrase directive. Please feel free to merge.

kifi-backend/modules/shoebox/angular/src/libraries/library.js
```diff
@@ -208,113 +197,62 @@ angular.module('kifi')
...
-    var deregisterUpdateLibrarySearch = $rootScope.$on('$stateChangeSuccess', function (event, toState) {
+    $rootScope.$on('$stateChangeSuccess', function (e, toState) {
       $scope.librarySearch = toState.data && toState.data.librarySearch;
+    }),
+
+    $rootScope.$on('userLoggedInStateChange', function (e, me) {
+      if (!me) {
+        reloadThisLibrary();
+      }
+    })
```
2is10, line 210: Rather than try to update this controller’s state to properly reflect a login or logout event, I thought it better to keep this controller simple and delegate the event to the router. In fact, we may want to do this globally in our application—not just on library pages.

kifi-backend/modules/shoebox/angular/src/libraries/library.js
```diff
@@ -208,113 +197,62 @@ angular.module('kifi')
...
-        return elem.url === path;
-      });
-
-      if (lib) {
-        $scope.page = 'library';
-        setTitle(lib);
-        util.replaceObjectInPlace($scope.library, lib);
-        prePopulated = true;
-      }
-    }
```
2is10, line 238: I eliminated this pre-population code. It caused choppy rendering of the header.
As a counter-measure, to make the library header appear noticeably faster, I delayed rendering of the first batch of keeps (see `$timeout` below).

kifi-backend/modules/shoebox/angular/src/libraries/library.js
```diff
@@ -208,113 +197,62 @@ angular.module('kifi')
...
+    setTitle(library);
 
+    $rootScope.$emit('libraryUrl', library);
+    $rootScope.$broadcast('relatedLibrariesChanged', []);
```
2is10, line 224: Clearing the related libraries immediately, rather than waiting until new ones are loaded. This helps them feel more like part of the page.

kifi-backend/modules/shoebox/angular/src/libraries/library.js
```diff
@@ -208,113 +197,62 @@ angular.module('kifi')
...
-    };
-    $scope.callTriggerInstall = function () {
-      $rootScope.$emit('triggerExtensionInstall');
-    };
-
-    var deregisterLogin = $rootScope.$on('userLoggedInStateChange', init.bind(this, true));
-    $scope.$on('$destroy', deregisterLogin);
-
-    init(true);
+      $scope.loading = false;
```
2is10, line 239: I changed `loading` to be initially `true`, so that we don’t see the “no keeps” message while waiting for the first batch of keeps to be drawn. Setting it to `false` here now that we’re actually drawing the keeps.

kifi-backend/modules/shoebox/angular/src/libraries/library.js
```diff
@@ -208,113 +197,62 @@ angular.module('kifi')
...
+      trackPageView({libraryRecCount: libraries.length});
+      $scope.relatedLibraries = libraries;
+      $rootScope.$broadcast('relatedLibrariesChanged', libraries);
+      if (initParams.install === '1' && !installService.installedVersion) {
+        showInstallModal();
+      }
+      if (initParams.intent === 'follow' && $scope.library.access === 'none') {
+        libraryService.joinLibrary($scope.library.id);
+      }
+    });
```
2is10, line 256: This block of code is unchanged from the old `library.id` watch above.

kifi-backend/modules/shoebox/angular/src/libraries/libraryCard.js
```diff
@@ -17,24 +17,10 @@ angular.module('kifi')
...
         //
-        function init() {
-          augmentData();
-
-          scope.$evalAsync(preloadSocial);
-
-          $timeout(function () {
-            element.addClass('kf-loaded');
-          });
-        }
```
2is10, line 37: I decided we don’t need `kf-loaded` in the library card because it’s just for enabling cover image editing and library search transitions.

kifi-backend/modules/shoebox/angular/src/libraries/libraryHeader.tpl.html
```diff
@@ -12,8 +12,7 @@
       <div class="kf-keep-lib-body">
         <div class="kf-keep-lib-header">
           <div class="kf-keep-lib-name-container kf-keep-title">
-            <span class="kf-keep-lib-name" title="{{library.name}}" kf-ellipsis full-text="library.name" max-num-lines=2 num-reloads=3
-              itemtype="name"></span>
+            <span class="kf-keep-lib-name" kf-ellipsis itemtype="name">{{library.name}}</span>
```
2is10: I’ll probably revert this line. I’m going to make some changes to `kf-ellipsis` so it won’t cause jumping, but I’ll do that in a later pull request.

kifi-backend/modules/shoebox/angular/src/libraries/libraryKeeps.tpl.html
```diff
@@ -1,17 +1,20 @@
...
+        <li><a href="javascript:" ng-click="callImportBookmarkFile()"><span class="sprite sprite-import-from-other-services-icon"></span>Import links from sites like Delicious, Pocket, Pinboard</a></li>
+        <li kf-max-version><a href="javascript:" ng-click="callTriggerInstall()"><span class="sprite sprite-install-extension-icon"></span>Install our browser extension on Chrome and Firefox</a></li>
+      </ul>
+    </div>
+    <div ng-if="!userIsOwner">
+      <p class="public-library-empty-explain">{{library.owner.firstName || 'The curator'}} hasn’t added anything to this library yet.</p>
+      <p class="public-library-empty-explain" ng-hide="$root.userLoggedIn">
+        Want to create and share your own libraries? <a href="/signup">Sign up</a> to join the Kifi family.
+      </p>
+    </div>
```
2is10, line 17: No real user-visible changes here. Just simpler logic.

kifi-backend/modules/shoebox/angular/src/libraries/libraryPassPhrase.tpl.html
```diff
@@ -0,0 +1,21 @@
...
+  <button class="kf-lib-passphrase-button" ng-click="submitPassPhrase()">
+    <span class="passphrase-key-icon sprite svg-white-key"></span>
+    <span class="passphrase-key-text">Let me in</span>
+  </button>
+  <div class="kf-lib-passphrase-example">Example: "hot yellow car"</div>
+  <div class="kf-lib-access-error clearfix" ng-if="showErrorMessage">
+    Oops, that didn’t work. Try again?<br>
+    Check the email you received for the correct pass phrase.
+  </div>
+</form>
```
2is10, line 21: Using `<form>` and `<button>` makes pressing Enter in the input field trigger `ng-click` on the button.

kifi-backend/modules/shoebox/angular/src/libraries/libraryHeader.tpl.html
```diff
@@ -12,8 +12,7 @@
       <div class="kf-keep-lib-body">
         <div class="kf-keep-lib-header">
           <div class="kf-keep-lib-name-container kf-keep-title">
-            <span class="kf-keep-lib-name" title="{{library.name}}" kf-ellipsis full-text="library.name" max-num-lines=2 num-reloads=3
-              itemtype="name"></span>
+            <span class="kf-keep-lib-name" itemtype="name">{{library.name}}</span>
```
2is10, line 15: If we’re not going to show the full library name at the top of library’s own page, then where? :smile:

PR #13352, opened 2015-01-28
using reactive lock for scraper tasks

eishay:


eishay:
@andrewconner

andrewconner:
Awesome

PR #13356, opened 2015-01-28
putting library search classes on only one element

2is10:
@yipingliao


kifi-backend/modules/shoebox/angular/src/libraries/library.styl
```diff
@@ -411,8 +410,7 @@ html:not(.kf-mobile)
     &.pageScrolled
       box-shadow: 0px 5px 5px #ddd
 
-  .library-search .library-card
-  .library-card.library-search-in-progress
+  .library-card
```
2is10, line 413: being careful not to change the specificity of any of the selectors

PR #13357, opened 2015-01-28
simplifying logout, removing old logout link on settings page

2is10:
@yipingliao


kifi-backend/modules/shoebox/angular/src/layout/header/header.tpl.html
```diff
@@ -48,7 +48,7 @@
         <menu class="kf-dropdown-menu">
           <a class="kf-dropdown-menu-item" href="{{userProfileUrl}}">Your Profile</a>
           <a class="kf-dropdown-menu-item" href="/profile">Settings</a>
-          <a class="kf-dropdown-menu-item" href="javascript:" ng-click="logout()">Log out</a>
+          <a class="kf-dropdown-menu-item" href="{{logoutUrl}}">Log out</a>
```
2is10, line 51: to avoid a vulnerability where rogue scripts or sites can log our users out, we should eventually make this an XSRF-proof POST request

PR #13358, opened 2015-01-28
not eagerly loading incoming friend requests anymore

2is10:
@yipingliao @andrewconner 

It looks like these no longer surface anywhere but on our friends page. May as well wait until user goes there, no?


kifi-backend/modules/shoebox/angular/src/layout/header/HeaderCtrl.js
```diff
@@ -159,9 +159,6 @@ angular.module('kifi')
       $document.off('keydown', addKeepsShortcut);
     });
 
-    friendService.getRequests();
-    $scope.friendRequests = friendService.requests;
```
2is10, line 163: (unused)

yipingliao:
:+1:

PR #13360, opened 2015-01-28
Returning to user early when no early exceptions.

andrewconner:
@stkem 

New short URLs came from the big import we did.


andrewconner:
Merging, comments appreciated.

PR #13361, opened 2015-01-28
re-enabling protractor tests, increasing their allowed timeout

2is10:
(see #13353)
@yipingliao 
cc @eishay @andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/angular/gulpfile.js
+++ b/kifi-backend/modules/shoebox/angular/gulpfile.js
@@ -487,7 +487,7 @@ gulp.task('server-dev', ['build-dev'], startDevServer);
 gulp.task('local-prod', ['build-prod'], startLocalProdServer);
 
 // This task is the one that should be run by the build script
-gulp.task('release', ['build-prod', 'karma']);
+gulp.task('release', ['build-prod', 'karma', 'assets:release', 'protractor:release']);
 
 // Use this task for normal development
 gulp.task('default', ['watch', 'server-dev']);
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/angular/gulpfile.js
```diff
@@ -487,7 +487,7 @@ gulp.task('server-dev', ['build-dev'], startDevServer);
 gulp.task('local-prod', ['build-prod'], startLocalProdServer);
 
 // This task is the one that should be run by the build script
-gulp.task('release', ['build-prod', 'karma']);
+gulp.task('release', ['build-prod', 'karma', 'assets:release', 'protractor:release']);
```
2is10, line 490: Listing `assets:release` explicitly here now so that if we need to remove `protractor:release` again in the future, we won’t stop publishing our assets by accident again.

kifi-backend/modules/shoebox/angular/test/protractor.conf.js
```diff
@@ -10,5 +10,5 @@ module.exports.config = {
 
   // Increase protractor timeout for page synchronization (default is 11s)
   // https://github.com/angular/protractor/blob/master/docs/timeouts.md
-  allScriptsTimeout: 20000
+  allScriptsTimeout: 60000
```
2is10, line 13: Increasing so that we can still make deployments that fix server performance regressions.

yipingliao:
Thanks!

PR #13362, opened 2015-01-28
shoebox client gives user active personas

yingjieMiao:
@stkem @ahsu1230 
- moved `UserPersona` to common package
- created `UserActivePersonas`, to store persona Ids and updated time stamp (could be useful)
- cache layer


kifi-backend/modules/common/app/com/keepit/model/UserPersona.scala
```diff
@@ -0,0 +1,34 @@
...
+    userId: Id[User],
+    personaId: Id[Persona],
+    state: State[UserPersona] = UserPersonaStates.ACTIVE) extends ModelWithState[UserPersona] {
+  def withId(id: Id[UserPersona]): UserPersona = copy(id = Some(id))
+  def withUpdateTime(now: DateTime): UserPersona = copy(updatedAt = now)
+}
+
+object UserPersonaStates extends States[UserPersona]
+
+@json case class UserActivePersonas(personas: Seq[Id[Persona]], updatedAt: Seq[DateTime])
```
ahsu1230, line 25: can this not be a `Map[personaId -> updatedAt]` or a `Map[personaId -> userPersona]`?
It would be easier to retrieve matching data

yingjieMiao, line 25: I would prefer `Map[personaId -> updatedAt]`. Storing `UserPersona` could be a waste.

yingjieMiao, line 25: I decided to keep it as is for now. IMHO, separate fields means easier to extend the object. Also, this way is slightly more efficient in serialization.

PR #13367, opened 2015-01-28
tracking page origin/subOrigin for library pages too

2is10:
The extension now sometimes sends people to library pages with `?o=xst`.
@yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
+++ b/kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
@@ -48,20 +48,13 @@ angular.module('kifi')
     }
 
     function trackPageView() {
-      var pageOrigin = originTrackingService.getAndClear();
       var url = $analytics.settings.pageTracking.basePath + $location.url();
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
```diff
@@ -48,20 +48,13 @@ angular.module('kifi')
...
         type: 'userProfile',
         profileOwnerUserId: $scope.profile.id,
         profileOwnedBy: $scope.viewingOwnProfile ? 'viewer' : ($scope.profile.isFriend ? 'viewersFriend' : 'other'),
-        origin: originsArray[0] || '',
-        subOrigin: originsArray[1] || '',
         libraryCount: $scope.profile.numLibraries
-      };
-
-      $analytics.pageTrack(url, profilePageTrackAttributes);
+      }));
```
2is10, line 57: moved the logic for parsing the origin and setting the tracking properties into `originTrackingService` to avoid duplicating it

PR #13368, opened 2015-01-28
accepting /:username/libraries and rewriting as /:username

2is10:
@andrewconner 
cc @yipingliao 

It’s good site design to let people remove the last path segment and get to a valid page.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/KifiSiteRouter.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/KifiSiteRouter.scala
@@ -166,6 +166,8 @@ class AngularRouter @Inject() (
             if (isUserAlias) Some(MovedPermanentlyRoute(redir)) else Some(SeeOtherRoute(redir))
           } else if (path.split.length == 1) { // user profile page
             Some(Angular(Some(userMetadata(user))))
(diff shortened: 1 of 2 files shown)


yipingliao:
:+1:

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/KifiSiteRouter.scala
```diff
@@ -166,6 +166,8 @@ class AngularRouter @Inject() (
             if (isUserAlias) Some(MovedPermanentlyRoute(redir)) else Some(SeeOtherRoute(redir))
           } else if (path.split.length == 1) { // user profile page
             Some(Angular(Some(userMetadata(user))))
+          } else if (path.split.length == 2 && path.split(1) == "libraries") { // user profile page (an alias that Angular will rectify)
```
andrewconner, line 169: Hm, couldn't we redirect at this stage?

andrewconner, line 169: Or would you prefer the angular app to be aware of it?

2is10, line 169: I considered that. It would mean an extra roundtrip. I guess I’m okay with it. If we go that route (server redirects), we should probably eliminate all of the Angular router rewrites at once.

PR #13372, opened 2015-01-28
make visibility not nullable in the db, use keptAt for sorting keeps, add a union methods for all private or public keeps to serve when asking for system libraries (with experiment)

eishay:


kifi-backend/modules/shoebox/app/com/keepit/model/KeepRepo.scala
```diff
@@ -434,16 +450,16 @@ class KeepRepoImpl @Inject() (
   }
 
   def getKeepsFromLibrarySince(since: DateTime, library: Id[Library], max: Int)(implicit session: RSession): Seq[Keep] = {
-    (for (b <- rows if b.libraryId === library && b.state === KeepStates.ACTIVE && b.createdAt > since) yield b).sortBy(_.createdAt asc).take(max).list
+    (for (b <- rows if b.libraryId === library && b.state === KeepStates.ACTIVE && b.keptAt > since) yield b).sortBy(_.createdAt asc).take(max).list
```
andrewconner: change to `sortBy(_.keptAt asc)`?

eishay: sure, thanks!

eishay: actually i missed many more, fixing

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -165,7 +165,22 @@ class LibraryCommander @Inject() (
...
+          val keeps = db.readOnlyMaster { implicit session =>
+            library.kind match {
+              case LibraryKind.USER_CREATED => keepRepo.getByLibrary(library.id.get, 0, maxKeepsShown)
+              case LibraryKind.SYSTEM_MAIN =>
+                assume(library.ownerId == viewerUserIdOpt.get, s"viewer ${library.ownerId} can't view a library he does not own: $library")
+                if (experimentCommander.userHasExperiment(library.ownerId, ExperimentType.ALL_KEEPS_VIEW)) {
+                  keepRepo.getNonPrivate(library.ownerId, 0, maxKeepsShown)
+                } else keepRepo.getByLibrary(library.id.get, 0, maxKeepsShown)
+              case LibraryKind.SYSTEM_SECRET =>
+                assume(library.ownerId == viewerUserIdOpt.get, s"viewer ${library.ownerId} can't view a library he does not own: $library")
```
2is10: These messages should have `${viewerUserIdOpt.get}` where `${library.ownerId}` currently is.

It would also be clearer if it said “system library” instead of just “library”.

2is10:
:+1:  what fun!

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
eishay: sure, fixed

kifi-backend/modules/shoebox/app/com/keepit/model/KeepRepo.scala
eishay: fixed

PR #13373, opened 2015-01-28
UserPersonaCommander changes

ahsu1230:
- changed `UserPersonaCommander` to be a trait
- change signatures (use Map or Set instead of Seq)
  - `def addPersonasForUser(userId: Id[User], personas: Set[String])`
  - `def removePersonasForUser(userId: Id[User], personas: Set[String]): Set[Persona]`
  - `def getByNames(names: Set[String])(implicit session: RSession): Map[String, Persona]`
- `UserPersonaRepo.getPersonasForUser()` joins with `PersonaRepo` to get a list of `Persona` objects, given a userId

@yingjieMiao @andrewconner @stkem


kifi-backend/modules/shoebox/app/com/keepit/commanders/UserPersonaCommander.scala
```diff
@@ -8,64 +8,80 @@ import com.keepit.common.time.Clock
 import com.keepit.heimdal.HeimdalContext
 import com.keepit.model._
 
-class UserPersonaCommander @Inject() (
+@ImplementedBy(classOf[UserPersonaCommanderImpl])
+trait UserPersonaCommander {
+  def addPersonaForUser(userId: Id[User], persona: String)(implicit context: HeimdalContext): (Option[Persona], Option[Library])
+  def addPersonasForUser(userId: Id[User], personas: Set[String])(implicit context: HeimdalContext): Map[Persona, Option[Library]]
+  def removePersonaForUser(userId: Id[User], persona: String): Option[Persona]
```
yingjieMiao, line 15: any reason why add methods has `HeimdalContext`, remove not? We only want to track add event? And so far, the context is not used.

ahsu1230, line 15: `LibraryCommander.addLibrary` requires a `HeimdalContext`, so naturally the `addPersonas` will need it as well.

I think it's good to add tracking on UserPersonas specifically once we get there and it's all specced out. Until then, I just like to leave it as is.

kifi-backend/modules/shoebox/app/com/keepit/commanders/UserPersonaCommander.scala
```diff
@@ -8,64 +8,80 @@ import com.keepit.common.time.Clock
...
 
-  def addPersonasForUser(userId: Id[User], personas: Set[String])(implicit context: HeimdalContext): (Seq[Persona], Seq[Library]) = {
-    val (newPersonas, currentPersonaIds) = db.readOnlyMaster { implicit s =>
-      val newPersonas = personaRepo.getByNames(personas)
-      val currentPersonaIds = userPersonaRepo.getUserPersonaIds(userId)
-      (newPersonas, currentPersonaIds)
+  def addPersonasForUser(userId: Id[User], personas: Set[String])(implicit context: HeimdalContext): Map[Persona, Option[Library]] = {
+    val personasToPersist = db.readOnlyMaster { implicit s =>
+      val currentPersonas = userPersonaRepo.getPersonasForUser(userId).toSet
+      val personaNamesToAdd = personas diff currentPersonas.map(_.name)
```
yingjieMiao: may need logic to deal with string normalization. e.g. Capitalization to lower case.

kifi-backend/modules/shoebox/app/com/keepit/commanders/UserPersonaCommander.scala
```diff
@@ -8,64 +8,80 @@ import com.keepit.common.time.Clock
...
-      val currentPersonaIds = userPersonaRepo.getUserPersonaIds(userId)
-      (newPersonas, currentPersonaIds)
+  def removePersonaForUser(userId: Id[User], persona: String): Option[Persona] = {
+    removePersonasForUser(userId, Set(persona)).headOption
+  }
+
+  def removePersonasForUser(userId: Id[User], personas: Set[String]): Set[Persona] = {
+    val personasToRemove = db.readOnlyMaster { implicit s =>
+      val currentPersonas = userPersonaRepo.getPersonasForUser(userId).toSet
+      val personaNamesToRemove = personas intersect currentPersonas.map(_.name)
```
yingjieMiao: same here, case sensitive?

yingjieMiao:
some minor comments. otherwise lgtm!

:+1:

kifi-backend/modules/shoebox/app/com/keepit/commanders/UserPersonaCommander.scala
ahsu1230: ah good call.

ahsu1230:
thanks for the comments! merging

andrewconner:
It makes sense to make it type safe. Arbitrary strings will quickly become unmanageable, for more reasons than case sensitivity.

As-is, I can call addPersonasForUser(userId, Set("hello aaron")), and it adds that to the database.

kifi-backend/modules/shoebox/app/com/keepit/commanders/UserPersonaCommander.scala
andrewconner, line 15: We'd pass a context to every CRUD function call then :). Better add that when it's needed, vs preventatively. Otherwise, we'll need to maintain it from here out.

Tracking may be done client side, or in a controller.

PR #13376, opened 2015-01-28
integrity check inside uriIntegrityPlugin

yingjieMiao:
@leogrim @ymatsuda 

@leogrim  suggested have this sanity check in integrityPlugin (to reduce 10+ minutes delay)
@ymatsuda : any performance concerns? Any side effects? race conditions?

I cannot directly refer to `orphanCleaner` in the plugin, circular dependency. So I made up a trait. 

tests passed


leogrim:
@andrewconner @stkem

andrewconner:
Ah this fixes the rescrape issue?

yingjieMiao:
hope so!
we used to explicitly handle scrapeInfo inside the plugin, but the logic was taken out long ago.
this brings it back.

One concern I have is that this may cause spikes in scraper activities. Maybe?

yingjieMiao:
shall we merge this one, except the merge confilcts?

andrewconner:
If it fixes the scraper issue for imported urls, then yes! :)

yingjieMiao:
OK. I will give it a try.

PR #13383, opened 2015-01-28
Fix stats formatting on mobile library header.

yipingliao:
@2is10 Looks like I broke some formatting on the public library header on mobile when I split up the library cards.


yipingliao:
Note: this is a quick fix for now, but I need to re-display the followers again. Will revisit right after hover A/B test is out.

2is10:
:+1: I don’t really know the mobile layout well, but seems fine to me.

PR #13384, opened 2015-01-28
User Persona Website Controller

ahsu1230:
- make `PersonaNames` type safe so `PersonaRepo` only persists correct personas
- website controller for retrieving a user's personas, adding a persona, and removing a persona

@andrewconner @yingjieMiao @stkem

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/model/Persona.scala
+++ b/kifi-backend/modules/common/app/com/keepit/model/Persona.scala
@@ -34,17 +34,17 @@ object PersonaName {
   case object ADVENTURER extends PersonaName("adventurer")
 
   def apply(str: String) = {
-    str match {
-      case FOODIE.`value` => FOODIE
-      case ARTIST.`value` => ARTIST
-      case PARENT.`value` => PARENT
-      case STUDENT.`value` => STUDENT
-      case DESIGNER.`value` => DESIGNER
-      case PHOTOGRAPHER.`value` => PHOTOGRAPHER
-      case ATHLETE.`value` => ATHLETE
-      case TECHIE.`value` => TECHIE
-      case GAMER.`value` => GAMER
-      case ADVENTURER.`value` => ADVENTURER
+    str.toLowerCase match {
+      case FOODIE.value => FOODIE
+      case ARTIST.value => ARTIST
+      case PARENT.value => PARENT
+      case STUDENT.value => STUDENT
+      case DESIGNER.value => DESIGNER
+      case PHOTOGRAPHER.value => PHOTOGRAPHER
+      case ATHLETE.value => ATHLETE
+      case TECHIE.value => TECHIE
+      case GAMER.value => GAMER
+      case ADVENTURER.value => ADVENTURER
     }
   }
 
--- a/kifi-backend/modules/common/app/com/keepit/model/Persona.scala
+++ b/kifi-backend/modules/common/app/com/keepit/model/Persona.scala
@@ -3,16 +3,52 @@ package com.keepit.model
 import com.keepit.common.db.{ Id, ModelWithState, State, States }
 import com.keepit.common.time._
 import org.joda.time.DateTime
+import play.api.libs.json._
 
 case class Persona(
     id: Option[Id[Persona]] = None,
     createdAt: DateTime = currentDateTime,
     updatedAt: DateTime = currentDateTime,
-    name: String,
+    name: PersonaName,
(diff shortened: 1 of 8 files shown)


kifi-backend/modules/common/app/com/keepit/model/Persona.scala
```diff
@@ -3,16 +3,52 @@ package com.keepit.model
...
   def withUpdateTime(now: DateTime): Persona = copy(updatedAt = now)
 }
 
+sealed abstract class PersonaName(val value: String)
+
+object PersonaName {
+  implicit def format[T]: Format[PersonaName] =
+    Format(__.read[String].map(PersonaName(_)), new Writes[PersonaName] { def writes(o: PersonaName) = JsString(o.value) })
+
+  case object FOODIE extends PersonaName("foodie")
```
ahsu1230, line 25: Just a few guesses on my part. These are not finalized yet

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/UserPersonaController.scala
```diff
@@ -0,0 +1,53 @@
...
+  def getAllPersonas() = UserAction { request =>
+    val (allPersonas, userPersonas) = db.readOnlyMaster { implicit s =>
+      val userPersonas = userPersonaRepo.getPersonasForUser(request.userId)
+      val allPersonas = personaRepo.getByState(PersonaStates.ACTIVE)
+      (allPersonas, userPersonas)
+    }
+    val personaMap = allPersonas.map { persona =>
+      (persona.name.value, userPersonas.contains(persona))
+    }.toMap
+    Ok(Json.obj("personas" -> Json.toJson(personaMap)))
```
ahsu1230, line 32: this method gets all active personas & returns `{personaName -> true/false }` depending on whether User has the persona active or not. This is so the UI can quickly display all the personas & highlight the active ones

yingjieMiao:
lgtm!

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/UserPersonaController.scala
```diff
@@ -0,0 +1,53 @@
...
+  db: Database,
+  userPersonaCommander: UserPersonaCommander,
+  personaRepo: PersonaRepo,
+  userPersonaRepo: UserPersonaRepo,
+  val userActionsHelper: UserActionsHelper,
+  heimdalContextFactory: HeimdalContextBuilderFactory,
+  clock: Clock)
+    extends UserActions with ShoeboxServiceController with Logging {
+
+  def getAllPersonas() = UserAction { request =>
```
stkem, line 23: We are going to need a mobile version for this that works for non users (obviously without info about the already selected personas) because mobile will show this pre signup.
Let's get together with @DerekBurch and @thassss  tomorrow to figure out exact mobile API.

kifi-backend/modules/common/app/com/keepit/model/Persona.scala
```diff
@@ -3,16 +3,52 @@ package com.keepit.model
 import com.keepit.common.db.{ Id, ModelWithState, State, States }
 import com.keepit.common.time._
 import org.joda.time.DateTime
+import play.api.libs.json._
 
 case class Persona(
```
stkem, line 8: This is also going to need to gain persona metadata, specifically `displayName` (or something like that) and `icon`.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/UserPersonaController.scala
stkem, line 23: Ah, just saw your other pull request :-)

ahsu1230, line 23: talked with @DerekBurch about this not too long ago. The mobile version supports non-users and gives you all the personas. =D

kifi-backend/modules/common/app/com/keepit/model/Persona.scala
ahsu1230, line 8: What will the `icon` look like? Is it a path, or maybe an icon name?

stkem, line 8: A cloudfront url, likely

stkem, line 8: Actually, I'm committing to that. It's a url.

PR #13385, opened 2015-01-29
distributing tasks by simple hashing

ymatsuda:
@eishay 
FYI @stkem @andrewconner 
- allowing distribution of tasks, which are currently executed by a leader, by hashing task names
- using this only in curator for now to see the behavior in prod.


andrewconner:
I like.

eishay:
wohooooo!!!

yingjieMiao:
nice!

eishay:
I think we should kill `scheduleTaskOnLeader` and use only the new method to reduce load on the leader machine.
@all What do you think?

stkem:
Sweet! +1 on what Eishay said, once this is validated.
I simple test for the hashing part in particular might be good, so people don't break it by accident in the future.

PR #13388, opened 2015-01-29
more careful filtering

yingjieMiao:
@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/RelatedLibraryCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/RelatedLibraryCommander.scala
@@ -50,17 +50,18 @@ class RelatedLibraryCommanderImpl @Inject() (
       case None => Set()
     }
 
-    def filterUnwantedRelatedLibs(relatedLibs: Seq[RelatedLibrary]) = {
-      relatedLibs.filter {
-        case RelatedLibrary(lib, kind) =>
-          !fakeUsers.contains(lib.ownerId) &&
-            !userLibs.contains(lib.id.get) &&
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/RelatedLibraryCommander.scala
```diff
@@ -50,17 +50,18 @@ class RelatedLibraryCommanderImpl @Inject() (
...
-        case RelatedLibrary(lib, kind) =>
-          !fakeUsers.contains(lib.ownerId) &&
-            !userLibs.contains(lib.id.get) &&
-            !libQualityHelper.isBadLibraryName(lib.name)
-      }
+    def isUnwantedLibrary(relatedLib: RelatedLibrary): Boolean = {
+      val lib = relatedLib.library
+
+      lib.state == LibraryStates.INACTIVE ||
+        fakeUsers.contains(lib.ownerId) ||
```
andrewconner, line 57: Ooo, this is quite expensive (1400 users).

yingjieMiao, line 57: it's a Set, get from cache. should be ok.

andrewconner:
Thanks. What's the number of keeps a library has to have to be recommended?

yingjieMiao:
for related libraries, min keep size is 2. @eishay wanted that, so that user can explore more.
popular libraries or same owner library should have 5+ keeps.

yingjieMiao:
Cortex side timely update library lda info. That library was turned to inactive almost immediately. However, the lib-lib related graph is NOT updated frequently, nor listening on any seqNum. It's a 12hour cron-job.

PR #13391, opened 2015-01-29
making library info generation more parallel and add image cache

eishay:


eishay:
@andrewconner

andrewconner:
I'm sure it's good :)

PR #13392, opened 2015-01-29
Implement A/B test for CTA hover on public library follow button.

yipingliao:
@2is10 Thanks for all your help!

cc @eishay


kifi-backend/modules/shoebox/angular/src/libraries/libraryHeaderFollowButtonCta.styl
```diff
@@ -0,0 +1,56 @@
...
+  font-family: Lato, Helvetica, Arial, sans-serif
+  font-size: 14px
+  text-align: center
+  background: #fff
+  z-index: 500
+
+.kf-lh-fbc-arrow-outer
+.kf-lh-fbc-arrow-inner
+  position: absolute
+  left: (kf-lh-fbc-width/2) + kf-lh-fbc-arrow-border-width
```
andrewconner, line 21: Hah, nice :)

kifi-backend/modules/shoebox/angular/src/libraries/libraryHeaderFollowButtonCta.js
```diff
@@ -0,0 +1,148 @@
...
+        // On link.
+        //
+        if (scope.isUserLoggedOut) {
+          // A/B test experiments.
+          var experiment = scope.library.abTestExperiment;
+          if (!experiment) {
+            return;
+          }
+
+          if (platformService.isSupportedMobilePlatform() ||
```
yipingliao: Not showing hover on mobile.

kifi-backend/modules/shoebox/angular/src/common/services/abTestService.js
```diff
@@ -0,0 +1,74 @@
...
+
+    function getExperiment(abTestId) {
+      var abTest = abTests[abTestId];
+
+      if (!abTest || !mixPanelId) {
+        return null;
+      }
+
+      return new Experiment(abTest);
+    }
```
2is10: It seems to me that what this function is really doing is choosing the appropriate treatment for the current user/visitor. Might be simpler to call it `.getTreatment` and avoid introducing a new “experiment” concept. We could clone the treatment object before returning it if you’re concerned about the caller modifying it.

2is10: Looks like the only extra information an experiment carries beyond the treatment data is the experiment name. Two possible ways of escaping that conundrum:
1. Key experiments by name instead of by numeric strings like `'1'`. Then the caller will already know the name.
2. Have the caller pass in the treatments array. That would greatly simplify this service. It would no longer need to know experiment IDs or names or know about all experiments in the app. Experiment configurations (treatments) would also then live closer to where they’re used.

yipingliao: Agreed re `.getTreatment`.
Agreed re keying experiments by name.
Not sure about the last one with the caller passing in the treatment array; I like having all the experiment configs in a central place instead of having to dig through the code for them, and being able to find out quickly if any A/B tests are active.

2is10: If you go with option 2 above, this service would then just be a simple function—might even fit in `util`.

``` js
chooseTreatment: function (treatments) {
  try {
    var i = parseInt($window.mixpanel.cookie.props.distinct_id.slice(-12), 16);
    return treatments[i % treatments.length];
  } catch (e) {
    // convention is that the first is always the control
    return treatments[0];  // or null?
  }
}
```

Notice that I changed `-14` to `-12` above. This is because 14 hex digits exceeds the amount of precision a double can faithfully represent.

``` js
> parseInt("fffffffffffffd", 16)
< 72057594037927940
> parseInt("fffffffffffffe", 16)
< 72057594037927940
> parseInt("ffffffffffffff", 16)
< 72057594037927940
```

13 hex digits is okay:

``` js
> parseInt("ffffffffffffd", 16)
< 4503599627370493
> parseInt("ffffffffffffe", 16)
< 4503599627370494
> parseInt("fffffffffffff", 16)
< 4503599627370495
```

I’m recommending 12 over 13 just because it’s a nice even number.

kifi-backend/modules/shoebox/angular/src/libraries/libraryHeaderFollowButtonCta.js
```diff
@@ -0,0 +1,148 @@
...
+            experiment.treatment_is_control || !experiment.treatment_data ||
+            !experiment.treatment_data.mainText || !experiment.treatment_data.quote || !experiment.treatment_data.quoteAttribution
+          ) {
+            return;
+          }
```
2is10, line 110: Defining the treatments in this file (see earlier suggestion) might allow us to trust them more and not have to check here that every single property is defined on them before using them.

Btw, let’s stick with `lowerCamelCase` for internally used property names.

kifi-backend/modules/shoebox/angular/src/libraries/libraryHeaderFollowButtonCta.js
```diff
@@ -0,0 +1,148 @@
...
+          // Populate the CTA content.
+          scope.ctaMainText = experiment.treatment_data.mainText;
+          scope.ctaQuote = experiment.treatment_data.quote;
+          scope.ctaQuoteAttribution = experiment.treatment_data.quoteAttribution;
+
+          // Show the CTA after a certain amount of time.
+          var autoShowCTAPromise = $timeout(function () {
+            showCTA(true);
+            trackHover('timer');
+          }, 5000);
```
2is10, line 135: Wow, aggressive. Interesting experiment.

2is10:
Nicely done! Excited to see the results.

kifi-backend/modules/shoebox/angular/src/common/services/abTestService.js
2is10: > I like having all the experiment configs in a central place instead of having to dig through the code for them,

Valid concern. Of course, a code search for `chooseTreatment` (or whatever name we choose) would locate them very quickly.

kifi-backend/modules/shoebox/angular/src/common/services/abTestService.js
```diff
@@ -0,0 +1,74 @@
...
+      $window.mixpanel.cookie.props &&
+      $window.mixpanel.cookie.props.distinct_id;
+
+    /**
+     * Return a new experiment based on the configuration of the passed-in abTest object.
+     */
+    function Experiment(abTest) {
+      this.name = abTest.name;
+
+      var treatmentIndex = parseInt(mixPanelId.slice(-14).replace('-', ''), 16) % abTest.treatments.length;
```
2is10: This treatment selection code has the unfortunate property that experiment group selection is correlated across experiments. For example, if we have two experiments that each have 3 treatments, then if a visitor is in the control group for one of them, he/she will also be in the control group for the other. To make treatment selection independent across all of our A/B tests, we need to define a unique “salt” for each experiment we run. Here’s one way it could work:

``` js
// To generate a salt for a new A/B test, run this code in your console:
// [0,0,0,0,0,0].map(function () { return Math.floor(Math.random() * 32).toString(32) }).join('')
// We use 32 because it’s the minimum number of hexademical digits in a distinct_id.
```

Then when we select a treatment, we use the `salt` associated with the experiment to select which hex characters from the `distinct_id` (six in this example) we’ll use, instead of always using the last 12 or 14:

``` js
// distinct_id is either a string (for visitors) or an object (for users)
var id = mixpanel.cookie.props.distinct_id;
var sourceDigits = (id.id || id).replace(/[^\da-f]/gi, '');
var chosenDigits = salt.split('').map(function (ch) {
  return sourceDigits[parseInt(ch, 32)];
}).join('');
var treatmentIndex = parseInt(chosenDigits, 16) % treatments.length;
```

yipingliao:
Thanks for all your comments, Jared! Addressing them right now before pushing out this test.

kifi-backend/modules/shoebox/angular/src/libraries/libraryHeader.js
```diff
@@ -141,6 +141,10 @@ angular.module('kifi')
             scope.coverImageUrl = env.picBase + '/' + image.path;
             scope.coverImagePos = formatCoverImagePos(image);
           }
+
+          scope.library.followButtonText = (scope.library.abTestTreatment && !scope.library.abTestTreatment.isControl) ?
+            scope.library.abTestTreatment.data.buttonText :
+            'Follow Library';
```
2is10, line 147: Do you think it would be clearer to put `.followButtonText`, `.abTest`, and `.abTestTreatment` on `scope`? I just realized they don’t really represent characteristics of the library.

yipingliao, line 147: Absolutely. In fact the ab test experiment/treatment object was originally on the scope. I moved them into the library object, however, because I needed to track them for library-related events and the code for tracking common library attributes lives inside `libraryService` and takes a library as its input.

PR #13395, opened 2015-01-29
mobile controller for user personas

ahsu1230:
@andrewconner @yingjieMiao @stkem 
cc: @jaredpetker @DerekBurch @thassss @mrbrown14 

two endpoints:
`GET     /m/1/user/personas` (can be called by a user or non-user)
response comes back with a name -> boolean which indicates whether persona is activated for that user or not. If non-user, all fields will be false

```
  personas: {
    "parent" : true
    "student" : false
    ...
  }
```

`POST    /m/1/user/personas` must be user & takes in:
`{ personas : [parent, adventurer, student]}`

which will persist all personas in that list. Any existing personas not in list will be removed.
responds with which personas have been added or removed

```
{
    added : [ "adventurer", "photographer" ]
    removed : ["student"]
}
```

Note: Backend has a safety guard as to which Persona you can add (kind of like library colors) - you can only add whatever the backend acknowledges is a valid persona

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileUserPersonaController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileUserPersonaController.scala
@@ -1,0 +1,56 @@
+package com.keepit.controllers.mobile
+
+import com.google.inject.Inject
+import com.keepit.commanders.UserPersonaCommander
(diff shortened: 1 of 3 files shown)


stkem:
Nice! As I mentioned in your other PR just now, we'll have to include persona metadata (display name, icon url) somewhere. We can talk to @DerekBurch and @thassss tomorrow to figure out what format works best for them.

ahsu1230:
cool! gotcha

stkem:
Doesn't need to stop you from merging this though.

PR #13400, opened 2015-01-29
removed leader-priority routing from scraper

ymatsuda:
@eishay @andrewconner @stkem 

Not sure if this is ok for assignScraperTasks. This may increase a chance of lock time outs and deadlocks.


stkem:
Worth a try. I think we should get this out and see what happens. Can't be much worth than the current situation :-)
@andrewconner @eishay ?

eishay:
yes!

PR #13407, opened 2015-01-29
Re-enable follow pics on mobile logged-out libraries.

yipingliao:
@andrewconner 

Not sure how this has worked before, but I might have broken it when I refactored a directive.


yipingliao:
Will come by and talk to you in person. :)

PR #13412, opened 2015-01-29
Zero, oups.

leogrim:


kifi-backend/modules/search/app/com/keepit/search/engine/LibraryQualityEvaluator.scala
```diff
@@ -31,7 +31,7 @@ class LibraryQualityEvaluator @Inject() (activeShards: ActiveShards) {
   private val mu = Math.log(optimalLibrarySize + Math.pow(sigma, 2))
   private val maxProbability = logNormalDensity(mu, sigma, optimalLibrarySize)
 
-  @inline private def logNormalDensity(mu: Double, sigma: Double, x: Double): Double = {
+  @inline private def logNormalDensity(mu: Double, sigma: Double, x: Double): Double = if (x == 0) 0 else {
     Math.exp(-Math.pow(Math.log(x) - mu, 2) / (2 * Math.pow(sigma, 2))) / (x * sigma * Math.sqrt(2 * Math.PI))
   }
```
yingjieMiao, line 36: If this is intended to be the log of normal density, then the formula seems to be off. e.g. why `log(x)`? the first `Math.exp` should not be there, etc.

leogrim, line 36: Nope it's meant the be the density of the log of a normal variable :)

yingjieMiao, line 36: yeah, i figured it out later. probability terms are not TypeSafe.

PR #13414, opened 2015-01-29
Activity email tracking and refactoring

joshm1:
@stkem


kifi-backend/modules/shoebox/app/com/keepit/commanders/emails/ActivityFeedEmailSender.scala
```diff
@@ -194,10 +159,39 @@ class ActivityFeedEmailSenderImpl @Inject() (
...
+  protected trait BaseActivityEmailDependencies extends ActivityEmailDependencies {
+    private val parent = ActivityFeedEmailSenderImpl.this
+    val airbrake = parent.airbrake
+    val libraryCommander = parent.libraryCommander
+    val libraryRepo = parent.libraryRepo
+    val db = parent.db
+    val membershipRepo = parent.membershipRepo
+    val clock = parent.clock
+    val eliza = parent.eliza
+  }
```
joshm1: not the prettiest solution, but this allows me to bypass guice for dependencies and create user specific instances of classes that are used to different "components" in the email.

the alternative is to create a protected inner class for every component, and I tried that, but it got messy with traits having abstract methods for dependency injected classes that were injected into the outer class

the other alternative is to just put all helper methods in the same class that gets dependencies injected and pass auxiliary objects around (userId, previouslySent emails, etc)

joshm1: this is a less than optimal implementation of the cake pattern. the standard pattern is to wrap each dependency inside a component object and use an additional self type for each dependency. I didn't want to create all of these just for this use case.

eg:

`class ... { self: AirbrakeComponent with ClockComponent with LibraryCommanderComponent =>`

kifi-backend/modules/shoebox/app/com/keepit/commanders/emails/ActivityFeedEmailSender.scala
```diff
@@ -194,10 +159,39 @@ class ActivityFeedEmailSenderImpl @Inject() (
...
-  class UserActivityFeedHelper(val toUserId: Id[User]) {
+  private def persistActivityEmail(activityEmailData: ActivityEmailData) = db.readWrite { implicit rw =>
+    activityEmailRepo.save(ActivityEmail(
+      userId = activityEmailData.userId,
+      libraryRecommendations = Some(activityEmailData.libraryRecos.map(_.libraryId)),
+      otherFollowedLibraries = None,
+      userFollowedLibraries = None))
+  }
 
+  class UserActivityFeedHelper(val toUserId: Id[User], val previouslySent: Seq[ActivityEmail]) extends BaseActivityEmailDependencies with ActivityEmailLibraryHelpers {
```
joshm1: it's a WIP extracting all of these public methods out of this monolith

stkem:
Sorry, I'm a bit late to this party. Looks good overall. Looking forward to your talk on this stuff next week.

PR #13417, opened 2015-01-29
Follow button cta follow up

yipingliao:
@2is10 Here are a couple of UI things for the follow button hover CTA that I didn't have time to get to yesterday:
- When the follow button is stuck to the top header, the CTA should also stick. Previously, the CTA will scroll off on scroll and the bandaid I used is to hide the CTA on scroll. Now the CTA will remain visible even on scroll if the follow button is itself stuck
- Handle better the case when the user has navigated away from the library page tab. Show the CTA automatically only if the tab is visible. When the user returns to the tab, set another timer to show the CTA automatically again.

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/libraryHeaderFollowButtonCta.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/libraryHeaderFollowButtonCta.js
@@ -34,10 +34,18 @@ angular.module('kifi')
               scope.ctaShownAbove = false;
             }
 
+            // If the follow button is "sticky" (i.e., its position is fixed),
+            // similarly fix the CTA's position so that it won't move upon scroll.
+            var position = 'absolute';
+            if (element.css('position') === 'fixed') {
+              position = 'fixed';
+              top = top - angular.element($window).scrollTop();
+            }
+
             // Center the CTA with respect to the follow button.
             var left = elementLeft - Math.round((cta.outerWidth() - element.outerWidth()) / 2);
 
-            cta.css({top: top + 'px', left: left + 'px'});
+            cta.css({position: position, top: top + 'px', left: left + 'px'});
             ctaOuterArrow.css({top: ctaOuterArrowTop + 'px'});
             ctaInnerArrow.css({top: ctaInnerArrowTop + 'px'});
 
@@ -92,12 +100,6 @@ angular.module('kifi')
           trackHover('hover');
         });
         element.on('mouseleave', hideCTA);
-
-        // TODO: make follow CTA sticky when follow button is sticky.
-        $window.addEventListener('scroll', hideCTA);
-        scope.$on('$destroy', function () {
-          $window.removeEventListener('scroll', hideCTA);
-        });
 
 
         //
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/libraryHeaderFollowButtonCta.js
```diff
@@ -93,10 +124,13 @@ angular.module('kifi')
         });
         element.on('mouseleave', hideCTA);
 
-        // TODO: make follow CTA sticky when follow button is sticky.
-        $window.addEventListener('scroll', hideCTA);
+        // Show the CTA automatically when the user returns to the library page.
+        function autoShowCTAQuickly() {
+          autoShowCTA(2000);
```
yipingliao: If the user has been on the page before and is returning to it, it feels like a better experience to not wait as long as 5 seconds to show the CTA. Here I am using just two seconds.

kifi-backend/modules/shoebox/angular/src/libraries/libraryHeaderFollowButtonCta.js
```diff
@@ -129,10 +158,15 @@ angular.module('kifi')
...
-          var autoShowCTAPromise = $timeout(function () {
-            showCTA(true);
-            trackHover('timer');
-          }, 5000);
+          var ctaShown = false;
+          var autoShowCTAPromise = null;
+          autoShowCTA(5000);
+
+          // Show the CTA automatically when the user returns to the library page.
+          document.addEventListener('visibilitychange', autoShowCTAQuickly);
```
2is10, line 166: This line looks a little suspicious—looks like we aren’t checking what kind of visibility change it was. In fact, `autoShowCTAQuickly` does check `document.hidden`, so we’re okay, but doing the check here before calling it or renaming the method slightly (e.g. appending `UnlessHidden`) would help make the code clearer (less likely to give the reader false alarm).

kifi-backend/modules/shoebox/angular/src/libraries/libraryHeaderFollowButtonCta.js
```diff
@@ -129,10 +158,15 @@ angular.module('kifi')
...
+          scope.$on('$destroy', function () {
+            $window.removeEventListener('visibilitychange', autoShowCTAQuickly);
```
2is10, line 168: Need to `.removeEventListener` from same DOM element it was added to—`document`, in this case.

2is10:
> Previously, the CTA will scroll off on scroll and the bandaid I used is to hide the CTA on scroll.

I can see your point and perspective on this. However, as a user, I’m often annoyed by things trying to grab my attention, so I try my best when coding to listen for and properly interpret signals that the user no longer wants to see something we’ve shown to try to grab attention. We can start listening for disinterest signals 1s or more after showing something to make sure we aren’t misinterpreting intent. For a pull-down menu or dialog, clicking somewhere else is a good signal. I think interpreting a click elsewhere as disinterest would apply here as well. Scrolling away from the CTA would be a disinterest signal too, in my opinion. At the very least, after it has scrolled off the page (away from the sticky button), I think we should hide it, rather than allow it to be scrolled back onto the page and reattached to the button. Of course, implementing any of this will mean more work, and we need to balance our priorities, but I couldn’t help mentioning the principle as something to at least consider for this CTA and in the future.

2is10:
Nicely done, btw. Good use of page visibility API.

PR #13418, opened 2015-01-29
Analytics & Tracking for sign up flows on public library pages

ahsu1230:
@andrewconner @yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/library.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/library.js
@@ -20,7 +20,10 @@ angular.module('kifi')
       $scope.platformName = installService.getPlatformName();
       if ($scope.platformName) {
         $scope.thanksVersion = 'installExt';
-        $scope.installExtension = installService.triggerInstall;
+        $scope.installExtension = function() {
+          $analytics.eventTrack('visitor_clicked_page', {type: 'installLibrary', action: 'install'});
+          installService.triggerInstall;
+        }
       } else {
         $scope.thanksVersion = 'notSupported';
       }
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/library.js
```diff
@@ -61,7 +64,11 @@ angular.module('kifi')
         $rootScope.$emit('trackLibraryEvent', 'view', { type: 'installLibrary' });
       }
 
-      $scope.close = modalService.close;
+      $scope.close = function () {
+        $analytics.eventTrack('visitor_clicked_page', {type : 'installLibrary', action: 'close'});
+        return modalService.close;
```
yipingliao, line 69: Same comment here as what we just discussed. Hmm...

kifi-backend/modules/shoebox/angular/src/signup/signupService.js
```diff
@@ -260,7 +288,11 @@ angular.module('kifi')
           } else if (installService.isValidFirefox) {
             $scope.platformName = 'Firefox';
           }
-          $scope.installExtension = installService.triggerInstall;
+          $scope.installExtension = function() {
+            $analytics.eventTrack('visitor_clicked_page', {type: 'installLibrary', action: 'install'});
+            $scope.installTriggered = true;
+            return installService.triggerInstall;
```
yipingliao, line 294: And here. I am curious as to how it works. Maybe it was not working before?!

kifi-backend/modules/shoebox/angular/src/signup/signupService.js
```diff
@@ -126,11 +138,17 @@ angular.module('kifi')
...
+        $rootScope.$emit('trackLibraryEvent', 'click', { type: 'signupLibrary', action: 'clickAuthTwitter' });
+      }
+    };
+
     $scope.fbAuthFromLibrary = function () {
       if ($scope.userData.libraryId) {
         $rootScope.$emit('trackLibraryEvent', 'click', { type: 'signupLibrary', action: 'clickAuthFacebook' });
       }
-      return $scope.fbAuth();
+      //return $scope.fbAuth(); //unused
```
2is10, line 151: @ahsu1230  Is the return value not used, or is the method call redundant? In either case, let’s remove the commented out code.

ahsu1230, line 151: Ok, I was hesitant in removing all the FB auth scripts, but I guess if it'll be saved on Github anyway, we could always return back to the FB Auth popup. @andrewconner

PR #13419, opened 2015-01-29
throttling assignScrapeTasks on the scraper

eishay:


eishay:
@ymatsuda @andrewconner this is timing out again, what do you think about throttling on the scraper side?

ymatsuda:
I think it is better to delay qSize checking until assignTasks is about to be executed by putting it in withLockFuture
- to minimize the time difference between getQueueSize and assignTasks, and
- to avoid calling assignTasks when the queue is full

andrewconner:
Agreed with Yasu

eishay:
fixed with yasu, the new code will come soon

PR #13421, opened 2015-01-30
Do not show hover tooltip on followers in library header for mobile devices

yipingliao:
@andrewconner 
I noticed that on a real phone, when you touch a follower's picture, the tooltip shows up - you have to press twice to get to the follower's profile page. This should fix that.

Another small bug is that on a library card on the user profile page, if you press the curator name and you're already on that curator's profile, an underline will appear. Adding a CSS change to fix that.

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/common/mobile.styl
+++ b/kifi-backend/modules/shoebox/angular/src/common/mobile.styl
@@ -119,8 +119,8 @@
           height: 80px
 
         .kf-keep-lib-additional-followers-text
-          padding-top: 19px
-          font-size: 2.2em
+          padding-top: 25px
+          font-size: 1.5em
 
         .kf-keep-lib-follower-preview
           text-align: center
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/angular/src/common/mobile.styl
```diff
@@ -119,8 +119,8 @@
           height: 80px
 
         .kf-keep-lib-additional-followers-text
-          padding-top: 19px
-          font-size: 2.2em
+          padding-top: 25px
```
yipingliao, line 122: One more: making the text smaller so big numbers will look okay.

2is10, line 122: After years of using padding for vertical centering, I settled on `line-height` as my preferred method for single-line text (setting `line-height` to the element’s inner height and using no vertical padding). One benefit is that no adjustment is needed if the `font-size` changes. Another is that I don’t have to remember to specify an arbitrary `line-height` to ensure consistent vertical positioning across browsers and platforms. (The default value of `line-height` is `normal`, which the CSS spec doesn’t specify precisely how to handle, so browsers vary.)

yipingliao, line 122: Will update. Thanks!

PR #13422, opened 2015-01-30
really distributed the jobs

yingjieMiao:
@eishay 

this `FeatureUpdatePlugin` is an abstract class. All updater inherit this class, and the all Update jobs are distributed to one machine. 

I added the Updater class name, should actually distribute things.


eishay:
good catch!

yingjieMiao:
a tricky one ;-)

PR #13430, opened 2015-01-30
user Logged In sometimes sees 'save'

ahsu1230:
@yipingliao sometimes the Save shows up for logged in users. Just an additional conditional!

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/libraryHeader.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/libraryHeader.js
@@ -142,7 +142,7 @@ angular.module('kifi')
             scope.coverImagePos = formatCoverImagePos(image);
           }
 
-          scope.library.followButtonText = (scope.library.abTestTreatment && !scope.library.abTestTreatment.isControl) ?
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/libraryHeader.js
```diff
@@ -142,7 +142,7 @@ angular.module('kifi')
             scope.coverImagePos = formatCoverImagePos(image);
           }
 
-          scope.library.followButtonText = (scope.library.abTestTreatment && !scope.library.abTestTreatment.isControl) ?
+          scope.library.followButtonText = (scope.userLoggedOut && scope.library.abTestTreatment && !scope.library.abTestTreatment.isControl) ?
```
2is10, line 145: It’s probably safer to use `scope.$root.userLoggedIn` directly, rather than a copied value which can get out of date. We probably ought to do so everywhere.

We don’t need to be cautious about `.userLoggedIn` not being initialized yet because no ui-views are instantiated until then.

No need to fix this now. I’ll get around to it. Just FYI.

PR #13431, opened 2015-01-30
refactoring handling of url pattern rules in order to move it to the scr...

eishay:
...aper and fixing the scraper task fetching per yasu's instructions


eishay:
@andrewconner @ymatsuda

PR #13435, opened 2015-01-30
Fix cta problems

yipingliao:
@2is10

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/libraryHeaderFollowButtonCta.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/libraryHeaderFollowButtonCta.js
@@ -3,8 +3,8 @@
 angular.module('kifi')
 
 .directive('kfLibraryHeaderFollowButtonCta', [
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/libraryHeaderFollowButtonCta.js
```diff
@@ -168,6 +168,14 @@ angular.module('kifi')
             $window.removeEventListener('visibilitychange', autoShowCTAQuickly);
           });
         }
+
+        var deregisterStateChange = $rootScope.$on('$stateChangeSuccess', function (event, toState) {
+          if (!toState.name.match(/^library\.keeps/)) {
```
2is10, line 173: For my own education, what is this check for? To prevent the cancellation of the promise immediately after it’s created? If so, we might consider using `$stateChangeStart` instead, and then maybe this check wouldn’t be necessary.

2is10:
:+1: Looks good. Thanks.

PR #13438, opened 2015-01-30
make URIWanderingTest more stable

ymatsuda:
The probability of failing this test is greater than it looks. The cause was the wandering failed to reach a topic node in preScore phrase.
Increasing the number of trials to 10000. The execution time does not increase too much.

@yingjieMiao @eishay


yingjieMiao:
thanks! It's a bit surprising

PR #13440, opened 2015-01-30
Adding columns to persona table

ahsu1230:
- Add `displayName`, `iconPath`, `activeIconPath`
- assume `iconPath` and `activeIconPath` are not nullable

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/model/Persona.scala
+++ b/kifi-backend/modules/common/app/com/keepit/model/Persona.scala
@@ -10,7 +10,10 @@ case class Persona(
     createdAt: DateTime = currentDateTime,
     updatedAt: DateTime = currentDateTime,
     name: PersonaName,
-    state: State[Persona] = PersonaStates.ACTIVE) extends ModelWithState[Persona] {
+    state: State[Persona] = PersonaStates.ACTIVE,
+    displayName: String,
+    iconPath: String,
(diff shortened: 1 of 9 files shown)


kifi-backend/modules/common/app/com/keepit/model/Persona.scala
```diff
@@ -10,7 +10,10 @@ case class Persona(
     createdAt: DateTime = currentDateTime,
     updatedAt: DateTime = currentDateTime,
     name: PersonaName,
-    state: State[Persona] = PersonaStates.ACTIVE) extends ModelWithState[Persona] {
+    state: State[Persona] = PersonaStates.ACTIVE,
+    displayName: String,
```
yingjieMiao, line 14: could `displayName` reuse the type safe `PersonaName`?
`PersonaName.allPersonas` doesn't have to change.

yingjieMiao:
lgtm

kifi-backend/modules/common/app/com/keepit/model/Persona.scala
stkem, line 14: I don't think that is a good idea. It would make it very easy to mix them up. If anything, it would have to be it's own typed wrapper. I would be ok with this one being just string for the time being though.

stkem:
lgtm

ahsu1230:
thanks for the comments

PR #13441, opened 2015-01-30
add /home marketing page and /parenting updates

joshm1:
@stkem


kifi-backend/modules/shoebox/conf/shoeboxService.routes
```diff
@@ -282,6 +282,7 @@ GET     /getting-started            @com.keepit.controllers.website.HomeControll
 GET     /terms                      @com.keepit.controllers.website.HomeController.route(path = "terms")
 GET     /privacy                    @com.keepit.controllers.website.HomeController.route(path = "privacy")
 GET     /parenting                  @com.keepit.controllers.website.HomeController.route(path = "parenting")
+GET     /home                       @com.keepit.controllers.website.HomeController.route(path = "home")
```
andrewconner, line 285: What is this page?

joshm1, line 285: Same as the landing page but without the sliding images. Sterling is going to run campaigns to this page and use query parameters and optimizely to change messages/images based on the campaign.
On January 30, 2015 at 1:44:58 PM, Andrew Conner (notifications@github.com) wrote:

In kifi-backend/modules/shoebox/conf/shoeboxService.routes:

PR #13443, opened 2015-01-30
eliminating the need to call shoebox with isScrapableP for each url scra...

eishay:
...ped in scraper, using a long term cached list of the URI rules stored in the scraper vm.


eishay:
@ymatsuda @andrewconner @leogrim @stkem

andrewconner:
:+1: Awesome

PR #13446, opened 2015-01-30
Update library search title

yipingliao:
@2is10 @eishay 

Page titles are now:
- Search query, no library (logged-in only)
  - some-query  • Kifi
- Search query inside a library
  - library-name  • some-query  • Kifi
- Tag inside a library
  - library-name  • some-tag  • Kifi
- Tags in general (logged-in only)
  - some-tag  • Kifi

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/layout/header/HeaderCtrl.js
+++ b/kifi-backend/modules/shoebox/angular/src/layout/header/HeaderCtrl.js
@@ -41,36 +41,41 @@ angular.module('kifi')
     //
     // Watchers & Listeners
     //
-    var deregisterLibraryChip = $rootScope.$on('libraryUrl', function (e, library) {
-      $scope.library = library;
+    [
+      $rootScope.$on('libraryUrl', function (e, library) {
+        $scope.library = library;
 
-      if ($scope.library.owner && !$scope.library.owner.picUrl) {
-        $scope.library.owner.picUrl = friendService.getPictureUrlForUser($scope.library.owner);
-      }
+        if ($scope.library.owner && !$scope.library.owner.picUrl) {
+          $scope.library.owner.picUrl = friendService.getPictureUrlForUser($scope.library.owner);
+        }
 
-      if ($scope.library.id) {
-        $scope.search.showName = true;
-      } else {
-        $scope.clearLibraryName();
-      }
+        if ($scope.library.id) {
+          $scope.search.showName = true;
+        } else {
+          $scope.clearLibraryName();
+        }
 
-      $scope.curatorProfileUrl = $scope.library.owner && routeService.getProfileUrl($scope.library.owner.username);
+        $scope.curatorProfileUrl = $scope.library.owner && routeService.getProfileUrl($scope.library.owner.username);
+      }),
+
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/shoebox/angular/src/layout/header/HeaderCtrl.js
```diff
@@ -41,36 +41,41 @@ angular.module('kifi')
...
-    $scope.$on('$destroy', deregisterUpdateSearchText);
+      $rootScope.$on('$stateChangeSuccess', function (event, toState, toParams) {
+        if ((toState.name === 'library.search') || (toState.name === 'search')) {
+          $scope.search.text = toParams.q;
+        } else {
+          $scope.search.text = '';
+        }
+      }),
+
+      $rootScope.$on('newSearch', function (event, params) {
```
yipingliao, line 69: @2is10 Mostly indentation changes to use the better-looking array-of-deregister-methods that you're using. Added a listener on when there is a new search initiated from the search page that was not the result of the search text changing (i.e., clicking on tags on search results page), so that the search text inside the header will be updated. This is a bug that I found while doing this update.

kifi-backend/modules/shoebox/angular/src/libraries/library.js
```diff
@@ -192,61 +192,62 @@ angular.module('kifi')
     //
     // Watches and listeners.
     //
-    [  // TODO: indent two spaces within this array
```
yipingliao, line 195: Indentation only in this file.

kifi-backend/modules/shoebox/angular/src/search/search.js
```diff
@@ -58,9 +70,9 @@ angular.module('kifi')
           return;
         }
 
-        libraryIdPromise = libraryService.getLibraryByUserSlug(userName, librarySlug, authToken, false).then(function (library) {
+        libraryPromise = libraryService.getLibraryByUserSlug(userName, librarySlug, authToken, false).then(function (library) {
           $rootScope.$emit('libraryUrl', library);
-          return library.id;
+          return library;
```
yipingliao, line 75: Previously the poorly named `library` internal variable in this controller was actually populated with a library's id. Updating to let it be a real library object (I needed the library to get its name for populating the title).

kifi-backend/modules/shoebox/angular/src/search/search.js
```diff
@@ -247,6 +259,7 @@ angular.module('kifi')
       // does not propagate to HeaderCtrl when it is injected there.
       // See: http://stackoverflow.com/questions/23081397/ui-router-stateparams-vs-state-params
       _.assign($state.params, $location.search());
+      $rootScope.$emit('newSearch', $state.params);
```
yipingliao, line 262: Let `HeaderCtrl` know that we're doing a new search. The messiness is due to the fact that our global header lives outside the outermost `ui-view` and so is unaffected by any `ui.router` changes (we can't just put a listener on `$stateChangeSuccess` inside `HeaderCtrl`), so we need to emit our own custom signal to it.

2is10:
:+1:

PR #13450, opened 2015-01-31
manually picked recos for personas

yingjieMiao:
@stkem 

curator client, controller hook up in future PR.

### DIFF
--- a/kifi-backend/modules/curator/app/com/keepit/curator/commanders/persona/LibraryPersonaRecommendationPool.scala
+++ b/kifi-backend/modules/curator/app/com/keepit/curator/commanders/persona/LibraryPersonaRecommendationPool.scala
@@ -1,0 +1,35 @@
+package com.keepit.curator.commanders.persona
+
+import com.google.inject.{ Inject, ImplementedBy, Singleton }
+import com.keepit.common.db.Id
+import com.keepit.curator.model.{ LibraryScores, LibraryRecommendation }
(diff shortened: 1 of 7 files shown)


kifi-backend/modules/curator/app/com/keepit/curator/commanders/persona/LibraryPersonaRecommendationPool.scala
```diff
@@ -0,0 +1,35 @@
...
+  val fixedRecos = FixedSetLibraryReco.recos
+
+  def generateRecoItemFromId(userId: Id[User], recoId: Id[Library]): LibraryRecommendation = {
+    FixedSetLibraryReco(userId, recoId)
+  }
+}
+
+object FixedSetLibraryReco {
+
+  private def fakeMasterScore = 100f
```
stkem, line 22: The numbers here we might have to tweak a bit before going to users, but this should probably do to start with.

stkem:
Nice! I like the abstraction you chose.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/persona/LibraryPersonaRecommendationPool.scala
yingjieMiao, line 22: yeah, it's more like a place holder for now.

PR #13452, opened 2015-01-31
removing the three-column layout

2is10:
@yipingliao 

This breaks a lot of stuff, but it’s getting us closer.


kifi-backend/modules/shoebox/angular/src/common/services/mixpanelService.js
```diff
@@ -62,7 +62,7 @@
 
   function trackEvent(action, props) {
     var mixpanel = $window.mixpanel;
-    if (mixpanel) { // TODO: fake implementation for tests
+    if (false && mixpanel) { // TODO: fake implementation for tests
```
2is10, line 65: reverting this file now...

kifi-backend/modules/shoebox/angular/src/keeps/keeps.styl
```diff
@@ -28,7 +28,7 @@
 
 .kf-shadow-dragged-keeps
 .kf-my-keeps
-  padding: 0 25px 20px 30px
+  padding: 0 25px 20px 0
```
2is10, line 31: This left-side space was reserved for editing keeps, but it makes keeps on most libraries (those you can’t edit) look improperly aligned now. Not sure what to do about this yet.

kifi-backend/modules/shoebox/angular/src/layout/main/MainCtrl.js
```diff
@@ -314,43 +314,5 @@ angular.module('kifi')
       $scope.showDelightedSurvey = profileService.prefs && profileService.prefs.show_delighted_question;
     });
     $scope.$on('$destroy', deregisterPrefsChangedListener);
-
-    /**
-     * Make the page "extension-friendly"
-     */
```
2is10, line 320: Hopefully it will be extension-friendly without any complex coercion, going forward. I guess the header will still need help, since it’s `position: fixed`.

kifi-backend/modules/shoebox/angular/src/layout/rightCol/RightColCtrl.js
```diff
@@ -1,225 +0,0 @@
...
-      installService.triggerInstall(function () {
-        modalService.open({
-          template: 'common/modal/installExtensionErrorModal.tpl.html'
-        });
-      });
-    };
-
-    $scope.triggerGuide = function (linkClicked) {
-      $window.postMessage({
-        type: 'start_guide',
```
2is10, line 106: We’ll need to bring this widget for triggering the guide back somewhere else.

PR #13454, opened 2015-01-31
activity email - mention library activity by followers

joshm1:
@stkem 

first dynamic component in the email: "N fellow users have followed <library>"


stkem:
:+1:

PR #13456, opened 2015-01-31
a bit more cleanup and not sending to the scraper uri's that we can't pa...

eishay:
...rse.

for those uri's we'll stop trying to scrape.


eishay:
@andrewconner @ymatsuda @leogrim

kifi-backend/modules/shoebox/app/com/keepit/controllers/internal/ScraperCallbackHelper.scala
```diff
@@ -67,33 +72,6 @@ class ScraperCallbackHelper @Inject() (
...
 
-  def saveImageInfo(info: ImageInfo): Unit = {
-    imageInfoLock.withLock {
-      db.readWrite(attempts = 3) { implicit s =>
-        imageInfoRepo.save(info)
-      }
-    }
-  }
-
-  def savePageInfo(info: PageInfo): Future[PageInfo] = {
```
andrewconner, line 78: Wahoo!

PR #13461, opened 2015-02-02
don't allow a scraper instance to mutate more then one uri at a time. ad...

eishay:
...d java.uri parsing check on the scraper side before sending uris for the shoebox to record. refactoring


eishay:
@andrewconner @ymatsuda @leogrim

PR #13462, opened 2015-02-02
Properly fix the clicking-on-tag search problem.

yipingliao:
@2is10 cc @andrewconner 

Looks like a fix I put in on Friday was not the right way to fix the problem, but introduced the out-of-order problem we saw on Kifi messages this morning.

This should fix the problem correctly, but I'll spend another 15 minutes testing it to check all the cases.

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/header/loggedInHeader.js
+++ b/kifi-backend/modules/shoebox/angular/src/header/loggedInHeader.js
@@ -51,7 +51,7 @@ angular.module('kifi')
 
         if ($scope.library.id) {
           $scope.search.showName = true;
-        } else {
+        } else if ($scope.search.showName) {
           $scope.clearLibraryName();
         }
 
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/angular/src/header/loggedInHeader.js
```diff
@@ -66,10 +66,6 @@ angular.module('kifi')
         }
       }),
 
-      $rootScope.$on('newSearch', function (event, params) {
```
yipingliao, line 69: This reverts Friday's change.

kifi-backend/modules/shoebox/angular/src/search/search.js
```diff
@@ -259,7 +259,6 @@ angular.module('kifi')
       // does not propagate to HeaderCtrl when it is injected there.
       // See: http://stackoverflow.com/questions/23081397/ui-router-stateparams-vs-state-params
       _.assign($state.params, $location.search());
-      $rootScope.$emit('newSearch', $state.params);
```
yipingliao, line 262: This is another reverting of Friday's change.

2is10:
:+1: I’m not entirely sure what this is about, but I will always be a fan of pull requests that do both of 1) eliminate code and 2) avoid message passing.

PR #13463, opened 2015-02-02
making dev.html look more like (production) index.html

2is10:
@andrewconner 

No functional change—this just makes it easier to switch back and forth between the two files in an editor and immediately see the differences.

### DIFF
--- a/kifi-backend/modules/shoebox/angular/dev.html
+++ b/kifi-backend/modules/shoebox/angular/dev.html
@@ -2,30 +2,31 @@
 <meta charset="utf-8">
 <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
 <meta name="viewport" content="width=device-width">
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/angular/dev.html
```diff
@@ -2,30 +2,31 @@
...
+<script>
+  if (location.origin === 'http://dev.ezkeep.com:8080') {
+    document.body.appendChild(function () {
+      var s = document.createElement('script');
+      s.src = '//localhost:35729/livereload.js';
+      s.async = true;
+      return s;
+    }());
+  }
+</script>
```
2is10, line 32: I guess moving this down could be considered a functional change, but doesn’t seem important since it’s loaded async.

andrewconner:
:+1:

PR #13466, opened 2015-02-02
Update deprecated CSS class name to new class name.

yipingliao:
@2is10


2is10:
:+1:  Yikes, I should not have missed those. My bad. Thanks for fixing.

PR #13469, opened 2015-02-02
Basic library color

leogrim:
@2is10 @mrbrown14 @jaredpetker @thassss @DerekBurch 
This requires a full reindexing, so will actually be available ~1 hour from now.


2is10:
:+1: thanks

PR #13470, opened 2015-02-02
styling buttons slightly differently

2is10:
@yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/app.styl
+++ b/kifi-backend/modules/shoebox/angular/src/app.styl
@@ -108,6 +108,33 @@ menu, ol, ul
 .ng-cloak
   display: none !important
 
+.kf-button
+  display: inline-block
+  vertical-align: top
+  height = 32px
+  height: height
+  border: 1px solid #bbb
+  border-radius: 5px
+  padding: 0 15px
+  background: linear-gradient(to top, #f3f3f3, #fff)
+  font: 13px\/height Lato, Helvetica, Arial, sans-serif
+  color: #666
+  a&
+    font-smoothing: auto
+
+    &:focus
+      outline: none
+
+.kf-button-icon
+  display: inline-block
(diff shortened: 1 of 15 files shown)


kifi-backend/modules/shoebox/angular/src/app.styl
```diff
@@ -108,6 +108,33 @@ menu, ol, ul
...
+.kf-button-icon
+  display: inline-block
+  vertical-align: top
+  width: 18px
+  height: inherit
+  background-repeat: no-repeat
+  background-position: center center
+  background-size: contain
+  margin: 0 5px 0 -3px
+
```
2is10, line 137: moved the styles from common.styl to app.styl so they aren’t repeated in our generated stylesheet (for each include)

yipingliao, line 137: Nice!

kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.styl
```diff
@@ -301,15 +301,7 @@ button.manage-lib-delete-button
     width: 100%
 
   .library-share-contact-button
-    // Override a bunch of global dialog button styles.
```
yipingliao, line 304: :)

kifi-backend/modules/shoebox/angular/src/app.styl
2is10, line 137: summary of styling changes:
- using `line-height` for vertical centering of text now
- no more absolute positioning of icons (padding no longer assumes there will be an icon)
- size of icon is now controlled only by its `kf-button-icon` width
- icon position can be adjusted using its margins or by translating or relatively positioning it
- no longer setting any margin on `kf-button`

yipingliao:
:+1:

PR #13471, opened 2015-02-02
kf-sticky no longer adjusting height of sticky element

2is10:
@yipingliao 
I’m not sure why the height was being measured and set explicitly. Our only current functioning use of `kf-sticky` is the follow button, and it hurts there after #13470, making the button too tall.


2is10:
merging since I just merged #13470. feedback appreciated.

PR #13473, opened 2015-02-02
factor out an IdFilter trait into common module

yingjieMiao:
@stkem @ymatsuda 

so that curator can reuse some codes. (The rest of the codes has Lucene I/O dependency)


stkem:
:+1:

PR #13476, opened 2015-02-02
activity email tracking & improved (?) ranking of recent-activity libraries to recommend

joshm1:
@stkem


stkem:
Looks good. Wether it's improved, time will tell :-)

joshm1:
the logic was I didn't want a library with 300% growth from 1 to 3 members or 2 to 6 to outweigh libraries with 200% with 10 to 20 members. i ran a few simulations to see how the ranking would turn out with various members and growths

PR #13482, opened 2015-02-03
a trait for filtering recos using IdFilter

yingjieMiao:
@stkem @ymatsuda 

intended to be shared for URIReco and LibReco.


stkem:
lgtm

PR #13484, opened 2015-02-03
admin controller endpoints for managing personas

ahsu1230:
@stkem @yingjieMiao

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/admin/AdminPersonaController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/admin/AdminPersonaController.scala
@@ -31,7 +31,7 @@ class AdminPersonaController @Inject() (
 
   def createPersona() = AdminUserPage { implicit request =>
     val personaOpt = request.body.asJson.map { json =>
-      val name = (json \ "name").as[PersonaName]
+      val name = (json \ "name").as[PersonaName] // note: can only create personas that are typesafe. Look at PersonaName
       val displayNameOpt = (json \ "displayName").as[Option[String]]
       val iconPath = (json \ "iconPath").as[String]
       val activeIconPath = (json \ "activeIconPath").as[String]
--- a/kifi-backend/modules/shoebox/app/com/keepit/admin/AdminPersonaController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/admin/AdminPersonaController.scala
@@ -1,0 +1,86 @@
+package com.keepit.controllers.admin
+
+import com.google.inject.Inject
+import com.keepit.commanders.UserPersonaCommander
+import com.keepit.common.controller.{ AdminUserActions, UserActionsHelper }
+import com.keepit.common.db.slick.Database
+import com.keepit.model._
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/shoebox/app/com/keepit/admin/AdminUserController.scala
```diff
@@ -805,6 +806,11 @@ class AdminUserController @Inject() (
           keepRepo.getByUser(userId).foreach { bookmark => keepRepo.save(bookmark.withActive(false)) }
           collectionRepo.getUnfortunatelyIncompleteTagsByUser(userId).foreach { collection => collectionRepo.save(collection.copy(state = CollectionStates.INACTIVE)) }
 
+          // Libraries Data
+          libraryInviteRepo.getByUser(userId, Set(LibraryInviteStates.INACTIVE)).foreach { case (invite, _) => libraryInviteRepo.save(invite.withState(LibraryInviteStates.INACTIVE)) } // Library Invites
+          libraryMembershipRepo.getWithUserId(userId).foreach { membership => libraryMembershipRepo.save(membership.withState(LibraryMembershipStates.INACTIVE)) } // Library Memberships
+          libraryRepo.getAllByOwner(userId).foreach { library => libraryRepo.save(library.withState(LibraryStates.INACTIVE)) } // Libraries
```
ahsu1230, line 812: when deactivating users, we should also delete library information. Currently, deactivated users still have active libraries & library memberships =/
@andrewconner @leogrim

stkem:
lgtm

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminPersonaController.scala
```diff
@@ -0,0 +1,86 @@
...
+    request.body.asJson.map { json =>
+      val currentNameOpt = (json \ "name").asOpt[PersonaName]
+      val displayNameOpt = (json \ "displayName").asOpt[String]
+      val iconPathOpt = (json \ "iconPath").asOpt[String]
+      val activeIconPathOpt = (json \ "activeIconPath").asOpt[String]
+      val stateOpt = (json \ "state").asOpt[State[Persona]]
+
+      db.readWrite { implicit s =>
+        val currentPersona = personaRepo.get(id)
+        val newCurrentName = currentNameOpt.getOrElse(currentPersona.name)
```
yingjieMiao, line 75: syntax-wise, these could just be 
`currentNameOpt orElse currentPersona.name`

slightly less crowded.

yingjieMiao:
lgtm

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminPersonaController.scala
ahsu1230, line 75: I think that will change the type of the return value that I want.
From my understanding, `orElse` requires an Option for the second parameter and returns an Option result. For this case, I'd like to just have `PersonaName` =D

ahsu1230:
thanks for the reviews

PR #13485, opened 2015-02-03
Track clicks on profile pages

yipingliao:
@ahsu1230 cc @2is10


kifi-backend/modules/shoebox/angular/src/header/loggedOutHeader.js
```diff
@@ -3,44 +3,62 @@
...
     $scope.clickLogo = function () {
-      $scope.$emit('getCurrentLibrary', { callback: function (lib) {
-        if (lib && lib.id) {
-          libraryService.trackEvent('visitor_clicked_page', lib, {
-            type: 'libraryLanding',
-            action: 'clickedLogo'
-          });
-        }
-      }});
+      if (util.startsWith($state.current.name, 'library')) {
```
yipingliao, line 11: In this file, previously we only tracked clicks for `libraryLanding` because that was the only public page we had. Now that we also have `userProfileLanding`, we need to track that too.

kifi-backend/modules/shoebox/angular/src/signup/signupService.js
```diff
@@ -139,7 +139,7 @@ angular.module('kifi')
       if (!form.$valid) {
         return false;
       }
-      emitTracking('click', 'signup', { action: 'clickSignUpButton' });
+      emitTracking('click', 'signup', { action: 'clickedSignUpButton' });
```
yipingliao, line 142: @ahsu1230 Per discussion with Jen, she actually wanted these to be past tense but used the present tense in the original Asana task.

kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
```diff
@@ -43,52 +30,40 @@ angular.module('kifi')
...
         profileOwnedBy: $scope.viewingOwnProfile ? 'viewer' : ($scope.profile.isFriend ? 'viewersFriend' : 'other'),
         libraryCount: $scope.profile.numLibraries
       }));
     }
 
-    function trackPageClick(/* attributes */) {
-      return;
-
-      /* TODO(yiping): Uncomment this after tracking review with product.
+    function trackPageClick(attributes) {
```
yipingliao, line 47: @ahsu1230 I moved the click-tracking function from `userProfileActionService` into `userProfile.js` because it's not really an action on a user profile, and so that it can live with `trackpageView`.

kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
```diff
@@ -43,52 +30,40 @@ angular.module('kifi')
...
     }
 
 
     //
-    // Scope methods.
+    // Scope methods and data.
     //
 
-    $scope.trackUplCardClick = function (lib) {
-      trackPageClick({
```
yipingliao, line 79: This and up to line 91 is old code that wasn't active because the tracking spec was fully flushed out.

kifi-backend/modules/shoebox/angular/src/signup/signupService.js
```diff
@@ -147,11 +147,11 @@ angular.module('kifi')
     };
 
     $scope.twAuth = function() {
-      emitTracking('click', 'signup', {action: 'clickAuthTwitter'});
+      emitTracking('click', 'signup', {action: 'clickedAuthTwitter'});
```
ahsu1230: lol this is originally me, but how about a spaces before & after `{}`

kifi-backend/modules/shoebox/angular/src/signup/signupService.js
```diff
@@ -147,11 +147,11 @@ angular.module('kifi')
...
     };
 
     $scope.fbAuth = function () {
-      emitTracking('click', 'signup', {action: 'clickAuthFacebook'});
+      emitTracking('click', 'signup', {action: 'clickedAuthFacebook'});
```
ahsu1230: same here

kifi-backend/modules/shoebox/angular/dev.html
```diff
@@ -10,7 +10,7 @@
 
 <!-- HEADER_PLACEHOLDER -->
 
-<div class="kf-cols ng-cloak" ng-controller="AppCtrl">
+<div class="kf-app ng-cloak" ng-controller="AppCtrl">
```
2is10, line 13: looks unintentional, like an artifact leaking in from the `site-relayout-dev` branch

kifi-backend/modules/shoebox/angular/src/signup/signupService.js
2is10: @ahsu1230 @yipingliao I suppose it’s a matter of personal taste, but I prefer no spaces in braces (for one-liner objects), like how we use no spaces in one-liner arrays and one-liner boolean expressions:

``` js
  return x < 0 && (y < 0 || y > 9);  // parentheses
  var numbers = [0, 1, 2, 3, 4];     // brackets
  doSomething('foo', {bar: 'baz'});  // braces
```

I’m happy to go the other way if I’m in the minority.  :smile:

ahsu1230: yea, I'm okay with either way. I was under the impression that we normally use spaces in arrays, booleans, and objects.

```
{ bar: 'baz' }
{ 
 'a': 1,
 'b': 2,
 'c': 3
}
```

As long as we're consistent =D

kifi-backend/modules/shoebox/angular/dev.html
yipingliao, line 13: Yup. This pull was closed and a new one opened because I inadvertently pulled in `site-relayout-dev` into the branch (51 changes files!)

2is10, line 13: I see. When I’m in the office again, I’d be happy to show you how to undo a mishap like that and salvage the pull request.

kifi-backend/modules/shoebox/angular/src/signup/signupService.js
2is10: Yes, spaces after operators like `:` and `,` and `+` and `||`, but not immediately inside and adjacent to parentheses, braces, and brackets for extra padding.

When the contents of an expression are wrapped over multiple lines, of course, we indent those lines two extra spaces.

This is what I meant—not saying this is what we need to accept. Still up for discussion.

PR #13488, opened 2015-02-03
Track profile clicks

yipingliao:
@ahsu1230

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
+++ b/kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
@@ -59,19 +59,11 @@ angular.module('kifi')
     // Scope methods and data.
     //
 
-    // Mapping of library type to origin contexts for tracking.
+    // Mapping of library type to library names for tracking.
     $scope.libraryTypesToNames = {
       'own': 'MyLibraries',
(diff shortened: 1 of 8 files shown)


ahsu1230:
lgtm!

kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
```diff
@@ -43,52 +30,40 @@ angular.module('kifi')
...
-
-    $scope.trackProfileClick = function () {
-      trackPageClick({
-        action: 'clickedProfile'
-      });
+    // Mapping of library type to library names for tracking.
+    $scope.libraryTypesToNames = {
+      'own': 'MyLibraries',
+      'following': 'FollowedLibraries',
+      'invited': 'InvitedLibraries'
```
2is10, line 66: This comment is unrelated to this pull request, but do you think Jen realizes that `MyLibraries` doesn’t always mean the viewer’s libraries? It would probably be clearer to call it `OwnedLibraries` or `OwnLibraries`.

yipingliao, line 66: I think so, but I will check. Thanks for the pointer!

PR #13497, opened 2015-02-03
use idFilter in commanders

yingjieMiao:
@stkem 

Internal changes in curator. return from controller stays the same.

Next PR will address compatibility of curatorClient.


stkem:
lgtm

PR #13502, opened 2015-02-03
client be backward compatible

yingjieMiao:
@stkem 

main changes:
- `CuratorController` changes return format
- `CuratorClient` handles Json with backward support
- Route changes in curator side with default parameters

Deploy order: 1) shoebox 2) curator.

Next PR: call new internal routes within CuratorClient


stkem:
:+1:

PR #13503, opened 2015-02-03
Dealing with Curator Memory Issues

stkem:
-More aggressive and more frequent URI recommendation reaper
-Don’t keep previously recommended URI ids (the alwaysInclude variable)
in memory for a given users if it’s going to be a long time before that
users task is up again.

(The problem was all the `alwaysInclude`s being kept around. We were
basically keeping all the URI ids of all recess in memory)

@yingjieMiao cc @ymatsuda @eishay


yingjieMiao:
hmm… 
the relation of `processSeeds`, `precomputeRecommendationsForUser` is hard to reason about.

They are calling each other, and one of them is recursive.

PR #13506, opened 2015-02-03
Revert "extracting scheduleScrape from the repo, making it async, persis...

eishay:
Reverts kifi/keepit#13489


eishay:
probably stopped scraping of urls, reverting

PR #13508, opened 2015-02-04
client calls with context

yingjieMiao:
@stkem 

Mostly API changes on shoebox side. all methods carry `context` parameter now. 

next PR: new endpoints for web/mobile.


stkem:
lgtm

PR #13514, opened 2015-02-04
Fix library mini card follow on logged out pages.

yipingliao:
Fixes a problem that has been unmasked after public library attributions are working.
@ahsu1230


ahsu1230:
:+1:

PR #13518, opened 2015-02-04
optionally notifying opener and closing window when social login completes

2is10:
@andrewconner 
cc @ahsu1230


2is10:
merging, won’t release extension that uses it until tomorrow or later. feedback appreciated.

ahsu1230:
pretty neat! :+1:

PR #13523, opened 2015-02-04
first persona icons: developer

ahsu1230:
@joshm1


ahsu1230:
here are where the persona icons will live! @andrewconner @stkem

joshm1:
tl;dr

PR #13526, opened 2015-02-04
Library menu (first set of changes)

yipingliao:
@2is10 The first set of changes for the library menu. Current state is that it's not strange looking, handles resize, positioned correctly, and functional (search, filter, scroll). More changes to come from meeting notes and polish.

cc @ahsu1230


kifi-backend/modules/shoebox/angular/src/layout/libraryMenu/libraryMenu.js
```diff
@@ -232,38 +257,43 @@ angular.module('kifi')
...
         //
         // Scrolling.
         //
-
-        // we thought about putting this check into the watch function above,
-        // but even when libraries are enabled, the element is found but the offset is 0
-        // in setLibListHeight(), if the offset is 0, the height of scrollableLibList == window height
-        // and thus no scrolly-bar =[
-        // once the offset is not 0, we know it's in the correct position and we can cancel this interval
-        var lastHeight = 0;
```
yipingliao, line 252: @ahsu1230 This doesn't seem to be needed anymore. Let me know if I am missing something. I'd also be happy to demo what I've got so you can play with it in case if there is some corner condition that I am not checking.

kifi-backend/modules/shoebox/angular/src/layout/libraryMenu/libraryMenu.styl
```diff
@@ -1,8 +1,71 @@
...
+  position: fixed
+  border: 1px solid #e6e6e6
+  padding: 0 15px
+  width: 250px
+  background: #fff
+  box-shadow: 0 0 7px rgba(0, 40, 90, 0.2)
+  overflow: hidden
+  z-index: 1
+
+.kf-nav-item
```
yipingliao, line 11: Renaming to `.kf-lm-` prefixes will come in another pull.

kifi-backend/modules/shoebox/angular/src/layout/libraryMenu/libraryMenu.styl
```diff
@@ -1,8 +1,71 @@
+.kf-library-menu
```
yipingliao, line 1: These styles are mostly old; I've moved them from another `.styl` file to consolidate the style rules into one file.

kifi-backend/modules/shoebox/angular/src/header/loggedInHeader.styl
```diff
@@ -41,6 +42,16 @@ left-logo-width = 92px
     .sprite-burger
       sprite2x($burger-hover)
 
+  .sprite-burger
+    border: 5px solid pageHeaderColor
+
+.kf-lih-toggle-menu-visible
```
yipingliao, line 48: Shows a square (with rounded corners) around the burger when the menu is open to indicate that it's active.

kifi-backend/modules/shoebox/angular/src/layout/libraryMenu/libraryMenu.js
ahsu1230, line 252: I'd love to take a look!

kifi-backend/modules/shoebox/angular/src/header/loggedInHeader.js
```diff
@@ -9,8 +9,8 @@ angular.module('kifi')
     $location, util, keyIndices, modalService, $timeout, $state, routeService) {
 
     $scope.toggleMenu = function () {
-      // navDropdown is inherited from AppCtrl scope.
-      $scope.navDropdown.visible = !$scope.navDropdown.visible;
+      // libraryMenu is inherited from AppCtrl scope.
+      $scope.libraryMenu.visible = !$scope.libraryMenu.visible;
```
2is10, line 13: we might consider moving it out of `AppCtrl` altogether

kifi-backend/modules/shoebox/angular/src/header/loggedInHeader.styl
```diff
@@ -41,6 +42,16 @@ left-logo-width = 92px
     .sprite-burger
       sprite2x($burger-hover)
 
+  .sprite-burger
+    border: 5px solid pageHeaderColor
```
2is10, line 46: better to use the color `transparent` here, I think (assumes less)

kifi-backend/modules/shoebox/angular/src/header/loggedInHeader.tpl.html
```diff
@@ -1,7 +1,7 @@
 <div class="kf-lih" ng-controller="LoggedInHeaderCtrl">
   <div class="kf-lih-inner">
     <div class="kf-lih-left">
-      <button class="kf-lih-toggle-menu" ng-click="toggleMenu()"><span class="sprite sprite-burger"></span></button>
+      <button class="kf-lih-toggle-menu" ng-class="{'kf-lih-toggle-menu-visible': libraryMenu.visible}" ng-click="toggleMenu()"><span class="sprite sprite-burger"></span></button>
```
2is10, line 4: For indicating an element’s state, I’m fine with using short indicator class names, like `kf-active` or `kf-visible` or `kf-open`, assuming they already have a unique identifying class name like `kf-lih-toggle-menu`.

kifi-backend/modules/shoebox/angular/src/layout/libraryMenu/libraryMenu.js
```diff
@@ -36,15 +35,34 @@ angular.module('kifi')
...
 
-
         //
         // Internal methods.
         //
-        function setLibListHeight() {
-          if (scrollableLibList.offset()) {
-            scrollableLibList.height(w.height() - (scrollableLibList.offset().top - w[0].pageYOffset));
+        function getElement(selector) {
+          var cache = {};
```
2is10, line 42: The cache will need to be defined outside of this function in order for it to work as intended.

kifi-backend/modules/shoebox/angular/src/layout/libraryMenu/libraryMenu.js
```diff
@@ -36,15 +35,34 @@ angular.module('kifi')
...
+        function positionMenu() {
+          if (scope.libraryMenu.visible) {
+            element.css({'left': getElement('.kf-lih-toggle-menu').offset().left + 'px'});
+          }
+        }
+
+        function setMenuHeight() {
+          var menuHeight = Math.floor(($window.innerHeight - getElement('.kf-lih').outerHeight()) * 0.9);
+          element.css({'height': menuHeight + 'px'});
+        }
```
2is10, line 60: I think we may be able to size and position using just CSS. I may take a look at it later. For example:

``` less
.kf-lm
  position: absolute   // relative to a container in the header
  left: 0
  top: 100%
  height: 90vh  
```

kifi-backend/modules/shoebox/angular/src/layout/libraryMenu/libraryMenu.js
```diff
@@ -232,38 +257,43 @@ angular.module('kifi')
...
-            setLibListHeight();
-          });
+        scope.$watch('libraryMenu.visible', function (visible) {
+          if (visible) {
+            openMenu();
+          }
         });
 
+        // On window resize, if the library menu is open, close it during the
+        // resize and reopen after resizing has completed.
```
2is10, line 267: I’d be happy to discuss the problems that currently require us to hide on resize. I’m hopeful that we could come up with alternative solutions that wouldn’t have this requirement, but I might just be naïve.

kifi-backend/modules/shoebox/angular/src/layout/libraryMenu/libraryMenu.js
```diff
@@ -422,7 +457,6 @@ angular.module('kifi')
         scope.$on('$destroy', function() {
           $document.off('mousedown', onClick);
```
2is10, line 458: `onDocMouseDown` might be a better name

2is10:
Excellent work, good progress!

kifi-backend/modules/shoebox/angular/src/layout/libraryMenu/libraryMenu.js
yipingliao, line 42: Duh! Thanks for the catch.

PR #13528, opened 2015-02-04
Updating karma, fixing jshint version issue

andrewconner:
This transitions to be very strict on versioning, which hopefully will help us have reliable environments.

Fixes:
- jshint was failing because of a transitive dependency. gulp-jshint is very flexible about versions of jshint it supports, so the same version of gulp-jshint may have a different version of jshint depending on when you updated.
- karma needed to be updated to a new API

This build seems to work on arthur and on my computer. I'd like to merge today. After I do, you'll likely need to delete `node_modules`, run `npm install`, then talk a walk while it installs everything.

@2is10 @yipingliao


yipingliao:
:+1: Nice!

kifi-backend/modules/shoebox/angular/package.json
```diff
@@ -16,55 +16,55 @@
...
+    "gulp-concat": "2.3.4",
+    "gulp-connect": "2.0.6",
+    "gulp-cssmin": "0.1.6",
+    "gulp-if": "1.2.4",
+    "gulp-ignore": "1.2.0",
+    "gulp-jshint": "1.9.2",
+    "gulp-livereload": "2.1.0",
+    "gulp-minify-html": "0.1.8",
+    "gulp-ng-html2js": "0.1.7",
+    "gulp-order": "1.1.1",
```
2is10, line 38: This makes me wish node had Ruby’s `Gemfile.lock` concept, which gives you the best of both worlds—flexible version requirements _and_ 100% reproducible environments.

2is10:
:+1: Thanks!

PR #13529, opened 2015-02-04
Fetch library image urls for library search results

leogrim:
@andrewconner @jaredpetker @mrbrown14


2is10:
Might this add latency to google.com and kifi.com searches? Hope not.

leogrim:
This is only for mobile for now (experimental design), and for library search results, not keeps.

PR #13536, opened 2015-02-05
More work on the library menu in new site layout

yipingliao:
@2is10 I am going to merge this so you can see it and hopefully spot some bugs while you're using it. Will address PR comments and the comments from the previous pull in another pull, along with lots of clean-up/renaming work. 

Known issue: height of menu is static (at 90% of `window.innerHeight`) - this should be less if we're displaying only a few libraries (e.g., when the search filter is being used).

Please see commit messages for the changes being added.


2is10:
:+1:

PR #13540, opened 2015-02-05
Making the sort a bit more deterministic

andrewconner:
@eishay `created_at` had ms resolution, so the test was failing often.


eishay:
good catch

eishay:
i guess we should always do a secondary sort on id if we sort on a date

PR #13544, opened 2015-02-05
Make library menu height dynamic

yipingliao:
@2is10 Getting this small change in there now that we're getting close.


2is10:
:+1:

PR #13545, opened 2015-02-05
using library from parent scope in library search

2is10:
@yipingliao


kifi-backend/modules/shoebox/angular/src/route.js
```diff
@@ -147,7 +150,8 @@ angular.module('kifi')
       .state('library.search', {
         url: '/find?q&f',
         templateUrl: 'search/search.tpl.html',
-        controller: 'SearchCtrl'
+        controller: 'SearchCtrl',
+        reloadOnSearch: false  // controller handles search query changes itself
```
2is10, line 154: seems like this should be here for consistency?

yipingliao:
Nice!

PR #13547, opened 2015-02-05
endpoint for FTUE: given a persona, retrieve a default keep & user's per...

ahsu1230:
...sona-created library

Here's how you would use it:

The Front-end will keep a list of selected personas and then once they're finished (clicked Done), you select any of the persisted personas and call this endpoint.
You will get back a default keep information for that persona and the corresponding user's library Id to keep to.

@2is10 @stkem @andrewconner @yingjieMiao

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/model/Persona.scala
+++ b/kifi-backend/modules/common/app/com/keepit/model/Persona.scala
@@ -2,6 +2,7 @@ package com.keepit.model
 
 import com.keepit.common.db._
 import com.keepit.common.time._
+import com.kifi.macros.json
 import org.joda.time.DateTime
 import play.api.libs.functional.syntax._
 import play.api.libs.json._
@@ -95,7 +96,7 @@ object PersonaName {
     DEEP_THINKER
   )
 
-  val personaLibraries: Map[PersonaName, String] = Map(
+  val personaLibraryNames: Map[PersonaName, String] = Map(
     DEVELOPER -> "Programming",
     TECHIE -> "Computer and Electronics",
     ENTREPRENEUR -> "Startups and Entrepreneurship",
@@ -113,6 +114,39 @@ object PersonaName {
     DEEP_THINKER -> "Interesting Reads"
   )
 
+  val personaKeeps: Map[PersonaName, DefaultPersonaKeep] = Map() // todo (aaron, ashley): pick out a default keep for each persona
+
 }
 
 object PersonaStates extends States[Persona]
+
+@json case class DefaultPersonaKeep( // default keep infos for FTUE
+  url: String,
+  name: Seq[String],
+  site: String,
+  thumbnailPath: String,
+  imagePath: Option[String] = None,
+  imageWidth: Option[Int] = None,
+  imageHeight: Option[Int] = None,
+  noun: String,
+  query: String,
+  title: String,
+  matches: JsObject,
+  track: String)
+
+object DefaultPersonaKeep { // todo (aaron): once every persona has a default keep, remove this!
+  val default = DefaultPersonaKeep(
+    url = "http://www.ted.com/talks/steve_jobs_how_to_live_before_you_die",
+    name = Seq("Steve Jobs", "How to Live", "Before You Die"),
+    site = "ted.com",
+    thumbnailPath = "/img/guide/before_you_die.jpg",
+    imagePath = Some("/img/guide/ted_jobs.jpg"),
+    imageWidth = Some(480),
+    imageHeight = Some(425),
+    noun = "video",
+    query = "steve+jobs",
+    title = "Steve Jobs: How to live before you die | Talk Video | TED.com",
+    matches = Json.obj("title" -> Json.toJson(Seq(Seq(0, 5), Seq(6, 4))), "url" -> Json.toJson(Seq(Seq(25, 5), Seq(31, 4)))),
+    track = "stevenJobsSpeech"
+  )
+}
--- a/kifi-backend/modules/common/app/com/keepit/model/Persona.scala
+++ b/kifi-backend/modules/common/app/com/keepit/model/Persona.scala
@@ -114,17 +114,14 @@ object PersonaName {
     DEEP_THINKER -> "Interesting Reads"
   )
 
-  val personaKeeps: Map[PersonaName, DefaultPersonaKeep] = Map() // todo (aaron, ashley): pick out a default keep for each persona
+  val personaKeeps: Map[PersonaName, PersonaKeep] = Map() // todo (aaron, ashley): pick out a default keep for each persona
 
 }
 
 object PersonaStates extends States[Persona]
 
-@json case class DefaultPersonaKeep( // default keep infos for FTUE
+@json case class PersonaKeep( // default keep infos for FTUE
   url: String,
-  name: Seq[String],
-  site: String,
-  thumbnailPath: String,
   imagePath: Option[String] = None,
   imageWidth: Option[Int] = None,
   imageHeight: Option[Int] = None,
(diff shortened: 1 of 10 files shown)


kifi-backend/modules/shoebox/app/com/keepit/controllers/website/UserPersonaController.scala
```diff
@@ -56,4 +59,30 @@ class UserPersonaController @Inject() (
...
+          db.readOnlyMaster { implicit s =>
+            libraryRepo.getByNameAndUserId(request.userId, libName)
+          }.map { lib =>
+            Library.publicId(lib.id.get)
+          }
+        }.flatten
+
+        Ok(Json.obj(
+          "keep" -> Json.toJson(defaultKeep),
+          "libraryId" -> libraryIdOpt
```
ahsu1230: on second thought, I think I should also include other library information. like library name, probably?

kifi-backend/modules/common/app/com/keepit/model/Persona.scala
```diff
@@ -113,6 +114,39 @@ object PersonaName {
...
+  title: String,
+  matches: JsObject,
+  track: String)
+
+object DefaultPersonaKeep { // todo (aaron): once every persona has a default keep, remove this!
+  val default = DefaultPersonaKeep(
+    url = "http://www.ted.com/talks/steve_jobs_how_to_live_before_you_die",
+    name = Seq("Steve Jobs", "How to Live", "Before You Die"),
+    site = "ted.com",
+    thumbnailPath = "/img/guide/before_you_die.jpg",
```
2is10: Turns out the extension isn’t using this at all. We don’t need a small thumbnail. Just `image`.

kifi-backend/modules/common/app/com/keepit/model/Persona.scala
```diff
@@ -113,6 +114,39 @@ object PersonaName {
...
+  query: String,
+  title: String,
+  matches: JsObject,
+  track: String)
+
+object DefaultPersonaKeep { // todo (aaron): once every persona has a default keep, remove this!
+  val default = DefaultPersonaKeep(
+    url = "http://www.ted.com/talks/steve_jobs_how_to_live_before_you_die",
+    name = Seq("Steve Jobs", "How to Live", "Before You Die"),
+    site = "ted.com",
```
2is10: `name` and `site` are unused, unnecessary

kifi-backend/modules/common/app/com/keepit/model/Persona.scala
```diff
@@ -113,6 +114,39 @@ object PersonaName {
...
+    thumbnailPath = "/img/guide/before_you_die.jpg",
+    imagePath = Some("/img/guide/ted_jobs.jpg"),
+    imageWidth = Some(480),
+    imageHeight = Some(425),
+    noun = "video",
+    query = "steve+jobs",
+    title = "Steve Jobs: How to live before you die | Talk Video | TED.com",
+    matches = Json.obj("title" -> Json.toJson(Seq(Seq(0, 5), Seq(6, 4))), "url" -> Json.toJson(Seq(Seq(25, 5), Seq(31, 4)))),
+    track = "stevenJobsSpeech"
```
2is10: Let’s remove the `n` from `"stevenJobsSpeech"` to match what we have today.

ahsu1230: lol woops :ghost:

kifi-backend/modules/common/app/com/keepit/model/Persona.scala
```diff
@@ -113,6 +114,39 @@ object PersonaName {
     DEEP_THINKER -> "Interesting Reads"
   )
 
+  val personaKeeps: Map[PersonaName, DefaultPersonaKeep] = Map() // todo (aaron, ashley): pick out a default keep for each persona
+
 }
 
 object PersonaStates extends States[Persona]
+
+@json case class DefaultPersonaKeep( // default keep infos for FTUE
```
2is10: Might consider removing the word “default” from this class name, since there’s only one kind of “persona keep”. Using the word “default” kind of makes it seem like there might also be some kind of non-default persona keeps. So maybe `PersonaKeep` or `PersonaGuidedKeep`.

kifi-backend/modules/shoebox/conf/shoeboxService.routes
```diff
@@ -419,9 +419,10 @@ GET     /site/user/abooks           @com.keepit.controllers.website.UserControll
 GET     /site/user/settings         @com.keepit.controllers.website.UserController.getSettings()
 POST    /site/user/settings         @com.keepit.controllers.website.UserController.setSettings()
 
-GET     /site/user/personas         @com.keepit.controllers.website.UserPersonaController.getAllPersonas()
-POST    /site/user/persona/:name    @com.keepit.controllers.website.UserPersonaController.addPersona(name: String)
-DELETE  /site/user/persona/:name    @com.keepit.controllers.website.UserPersonaController.removePersona(name: String)
+GET     /site/user/personas             @com.keepit.controllers.website.UserPersonaController.getAllPersonas()
+POST    /site/user/persona/:name        @com.keepit.controllers.website.UserPersonaController.addPersona(name: String)
+DELETE  /site/user/persona/:name        @com.keepit.controllers.website.UserPersonaController.removePersona(name: String)
+GET     /site/user/persona/keep/:name   @com.keepit.controllers.website.UserPersonaController.getDefaultKeepForPersona(name: String)
```
2is10:
```
GET     /site/user/persona/:name/keep
```

2is10: Actually, all of the `/persona/:name` routes should be `/personas/:name`, to follow REST conventions. Think of `/personas` as a directory on a file system.

2is10:
> The Front-end will keep a list of selected personas and then once they're finished (clicked Done), you select any of the persisted personas and call this endpoint.

What about the case where the user clicks the “Kifi Step-by-step Guide” link on the right-hand side? The web app doesn’t necessarily know what the user’s personas are at that point. Might be better to have a `/site/user/guide` route that returns the data the guide needs. It’s not necessarily related to personas. (Users who haven’t chosen any personas should still be able to start the guide.)

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/UserPersonaController.scala
2is10: That won’t be necessary. The extension will already know it.

kifi-backend/modules/common/app/com/keepit/model/Persona.scala
```diff
@@ -113,6 +114,33 @@ object PersonaName {
...
+
 }
 
 object PersonaStates extends States[Persona]
+
+@json case class PersonaKeep( // default keep infos for FTUE
+  url: String,
+  imagePath: Option[String] = None,
+  imageWidth: Option[Int] = None,
+  imageHeight: Option[Int] = None,
```
2is10, line 127: These will need to be required, because the extension has no alternate way of representing a page on Step 1. Has to be an image. At least that’s how it is today.

kifi-backend/modules/shoebox/app/com/keepit/commanders/UserPersonaCommander.scala
```diff
@@ -14,6 +14,7 @@ trait UserPersonaCommander {
   def addPersonasForUser(userId: Id[User], personas: Set[PersonaName])(implicit context: HeimdalContext): Map[Persona, Option[Library]]
   def removePersonaForUser(userId: Id[User], persona: PersonaName): Option[Persona]
   def removePersonasForUser(userId: Id[User], personas: Set[PersonaName]): Set[Persona]
+  def getPersonaLibraryAndKeep(userId: Id[User]): (PersonaKeep, Option[Library])
```
2is10, line 17: would be good to have the keep and library ordering consistent between the method name and returned tuple

kifi-backend/modules/shoebox/app/com/keepit/commanders/UserPersonaCommander.scala
```diff
@@ -86,4 +87,20 @@ class UserPersonaCommanderImpl @Inject() (
     }
     personasToRemove.values.toSet
   }
+
+  def getPersonaLibraryAndKeep(userId: Id[User]): (PersonaKeep, Option[Library]) = {
+    val personaOpt = db.readOnlyMaster { implicit s =>
+      userPersonaRepo.getPersonasForUser(userId).headOption
```
2is10, line 93: better to filter at the database layer

2is10:
Nice work. Great tests!

PR #13548, opened 2015-02-05
Encode tags that are being added or deleted so that they can contain slashes

yipingliao:
@2is10 This addresses a small bug that @leogrim found where we don't deal well with tags with slashes in them.

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/tags/hashtagService.js
+++ b/kifi-backend/modules/shoebox/angular/src/tags/hashtagService.js
@@ -50,7 +50,7 @@ angular.module('kifi')
 
       var keepIds = _.map(keeps, function (keep) { return keep.id; });
       var deferred = $q.defer();
-      var promise = $http.post(routeService.tagKeeps(tag), { 'keepIds': keepIds });
+      var promise = $http.post(routeService.tagKeeps(encodeURIComponent(tag)), { 'keepIds': keepIds });
 
       if (beOptimistic) {
         addTagToKeeps(keeps);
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/angular/src/tags/hashtagService.js
```diff
@@ -71,7 +71,7 @@ angular.module('kifi')
     };
 
     api.untagKeep = function (keep, tag) {
-      var promise = $http['delete'](routeService.untagKeep(keep.libraryId, keep.id, tag));
+      var promise = $http['delete'](routeService.untagKeep(keep.libraryId, keep.id, encodeURIComponent(tag)));
```
2is10: Okay if we do this inside `routeService` instead? It’s a concern more relevant to the assembly of the URL string.

As you’re probably aware, we need to use `encodeURIComponent` for all user-supplied values that we put in URLs (e.g. usernames, slugs, search queries) unless we restrict valid values to a subset of URL-safe characters (alphanumerics and [a handful of punctuation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent)).

2is10:
Thanks for fixing!

kifi-backend/modules/shoebox/angular/src/tags/hashtagService.js
yipingliao: Will do. Thanks! Slashes are not allowed in usernames and slugs, but let me check search queries.

PR #13549, opened 2015-02-05
Hopefully fix shortener issues

leogrim:
@andrewconner 
(if this works, we can get rid of some of the callback routes / methods)

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/integrity/UriIntegrityPlugin.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/integrity/UriIntegrityPlugin.scala
@@ -386,16 +386,12 @@ class UriIntegrityPluginImpl @Inject() (
 
 @Singleton
 class UriIntegrityHelpers @Inject() (urlRepo: URLRepo, keepRepo: KeepRepo) extends Logging {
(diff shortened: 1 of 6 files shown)


andrewconner:
Should be good — never seen this code before though :)

andrewconner:
Let's try it.

leogrim:
Actually it wasn't, additional commit coming.

kifi-backend/modules/shoebox/app/com/keepit/integrity/UriIntegrityPlugin.scala
```diff
@@ -384,3 +383,26 @@ class UriIntegrityPluginImpl @Inject() (
...
+        case Some(url) => keepWithTitle.copy(url = url.url, urlId = url.id.get)
+      }
+    } else keepWithTitle
+  }
+
+  def improveKeepsSafely(uri: NormalizedURI)(implicit session: RWSession): Unit = {
+    keepRepo.getByUri(uri.id.get).foreach { keep =>
+      val betterKeep = improveKeepSafely(uri, keep)
+      if (betterKeep != keep) { keepRepo.save(betterKeep) }
+    }
```
leogrim, line 402: @andrewconner

PR #13551, opened 2015-02-05
decouple ScrapeInfoRepo from NormalizedURIRepoImpl

ymatsuda:

ScrapeInfo is populated asynchronously using NormalizedURI sequence number rather than transactionally.
This will make it lighter to save NormalizedURI.

next to do: decouple ScrapeInfoRepo from OrphanCleaner and UriIntegrityActor


andrewconner:
Wahoo!

yingjieMiao:
lgtm! Long anticipated change!

PR #13552, opened 2015-02-05
updating the logged out header and logged out library search

2is10:
@yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/friends/friends.js
+++ b/kifi-backend/modules/shoebox/angular/src/friends/friends.js
@@ -3,8 +3,8 @@
 angular.module('kifi')
 
 .controller('FriendsCtrl', [
-  '$scope', '$window', 'friendService', 'socialService', '$rootScope',
-  function ($scope, $window, friendService, socialService, $rootScope) {
+  '$scope', '$window', 'friendService', 'socialService',
+  function ($scope, $window, friendService, socialService) {
     $window.document.title = 'Kifi • Your Friends on Kifi';
 
     $scope.$watch(socialService.checkIfRefreshingSocialGraph, function (v) {
@@ -41,8 +41,6 @@ angular.module('kifi')
 
     friendService.getKifiFriends();
     friendService.getRequests();
-
-    $rootScope.$emit('libraryUrl', {});
   }
 ]);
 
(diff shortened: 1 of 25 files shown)


kifi-backend/modules/shoebox/angular/src/friends/friends.js
```diff
@@ -41,8 +41,6 @@ angular.module('kifi')
 
     friendService.getKifiFriends();
     friendService.getRequests();
-
-    $rootScope.$emit('libraryUrl', {});
```
2is10, line 45: now emitting this only from library.js instead of from every top-level ui-view

yipingliao:
:+1: Merge!

kifi-backend/modules/shoebox/angular/src/header/loggedOutHeader.js
```diff
@@ -24,6 +27,80 @@ angular.module('kifi')
...
+    };
+
+    $scope.onKeydown = function (e) {
+      switch (e.keyCode) {
+        case 27:  // esc
+          $scope.clearInput();
+          break;
+      }
+    };
+
```
2is10, line 103: The above code is taken from loggedInHeader.js, but simplified since less can happen when logged out.

kifi-backend/modules/shoebox/angular/src/libraries/library.tpl.html
```diff
@@ -7,24 +7,20 @@
...
+    <a class="kf-lib-footer-a" target="_top" href="http://blog.kifi.com">Blog</a>
   </div>
-  <span ng-if="!isMobile"> &middot; </span>
-  <div class="second-line">
-    <a class="kf-footer-item-link" target="_top" href="https://www.kifi.com/privacy">Privacy</a> &middot;
-    <a class="kf-footer-item-link" target="_top" href="https://www.kifi.com/terms">Terms</a>
+  <span ng-if="!isMobile">&middot;</span>
+  <div class="kf-lib-footer-2">
+    <a class="kf-lib-footer-a" target="_top" href="/privacy">Privacy</a> &middot;
+    <a class="kf-lib-footer-a" target="_top" href="/terms">Terms</a>
```
2is10, line 24: The class prefix for this template (the library page) is now `kf-lib-`.

PR #13561, opened 2015-02-06
Remove one-keep route

yipingliao:
@2is10 I am going to come back to the animation part later to move on to other tasks first.


kifi-backend/modules/shoebox/angular/src/search/search.js
```diff
@@ -12,7 +14,6 @@ angular.module('kifi')
     var filter;
     var lastResult = null;
     var selectedCount = 0;
-    var authToken = $location.search().authToken || '';
```
2is10, line 15: I fixed the warnings in this file already. I think if you just revert this file, you’ll probably be able to merge this pull request without any conflicts.

2is10:
:+1: Thanks!

PR #13565, opened 2015-02-06
avoiding search box animation when page regains focus

2is10:
@yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/header/loggedOutHeader.js
+++ b/kifi-backend/modules/shoebox/angular/src/header/loggedOutHeader.js
@@ -3,9 +3,9 @@
 angular.module('kifi')
 
 .controller('LoggedOutHeaderCtrl', [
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/shoebox/angular/src/header/loggedOutHeader.js
```diff
@@ -39,22 +39,17 @@ angular.module('kifi')
...
 
-
-    $scope.focusInput = function () {
+    $scope.onInputFocus = function () {
       $scope.search.focused = true;
     };
-    $scope.blurInput = function () {
-      $scope.search.focused = false;
+    $scope.onInputBlur = function ($event) {
+      $scope.search.focused = $window.document.activeElement === $event.target;
```
2is10, line 52: When a page loses focus, the `blur` event fires on the active element, but it remains the active element, and `focus` fires on it when the page regains focus. We want our notion of “focused” to correspond with the document’s active element; we don’t really care about when the page gains and loses focus.

PR #13567, opened 2015-02-06
util.$debounce

2is10:
@yipingliao 

`util.$debounce` is a faster (synchronous) alternative to using `$timeout` inside a debounced function to make sure Angular will notice your updates.

inspired by `safeApply`
https://coderwall.com/p/ngisma/safe-apply-in-angular-js

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/common/util.js
+++ b/kifi-backend/modules/shoebox/angular/src/common/util.js
@@ -59,13 +59,15 @@ angular.module('util', [])
         var _y = rawDom.getBoundingClientRect().top + scrollY;
         return { left: _x, top: _y };
       },
-      isIE: function () {
-        // Feature detection should be preferred to browser detection, so this function should be avoided.
-        return (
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/angular/src/common/util.js
```diff
@@ -59,13 +59,15 @@ angular.module('util', [])
         var _y = rawDom.getBoundingClientRect().top + scrollY;
         return { left: _x, top: _y };
       },
-      isIE: function () {
-        // Feature detection should be preferred to browser detection, so this function should be avoided.
-        return (
-          (navigator.appName === 'Microsoft Internet Explorer') ||
-          ((navigator.appName === 'Netscape') &&
-           (/Trident/.exec(navigator.userAgent) != null))
-        );
```
2is10, line 68: `isIE` was unused

2is10:
I’ve seen a few other places where we could use `util.$debounce`. I’ll gradually convert.

PR #13570, opened 2015-02-06
fixing an issue resulting in a gap before the additional followers count

2is10:
@yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/libraryFollowerPics.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/libraryFollowerPics.js
@@ -3,8 +3,8 @@
 angular.module('kifi')
 
 .directive('kfLibraryFollowerPics', [
-  '$window', 'platformService',
-  function ($window, platformService) {
+  '$window', 'platformService', 'util',
+  function ($window, platformService, util) {
     return {
       restrict: 'A',
       replace: true,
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/libraryFollowerPics.js
```diff
@@ -15,70 +15,54 @@ angular.module('kifi')
...
-            } else if (maxFollowersToShow >= scope.numFollowers) {
-              scope.followersToShow = scope.followers;
-              scope.numAdditionalFollowers = 0;
-            } else {
-              scope.followersToShow = scope.followers.slice(0, maxFollowersToShow);
-              scope.numAdditionalFollowers = scope.numFollowers - maxFollowersToShow;
-            }
-
-            element.find('.kf-lfp-pics')
-              .width(maxFollowersToShow >= 1 ? maxFollowersToShow * widthPerFollowerPic : 0);
```
2is10, line 63: The problem was that this line and the `.numAdditionalFollowers` calculation on line 59 above were using `maxFollowersToShow` instead of the actual number of followers shown, which might be smaller due to a limited number of followers being available in `scope.followers`. Tried to make the new code a little bit simpler and easier to reason about.

kifi-backend/modules/shoebox/angular/src/libraries/libraryFollowerPics.tpl.html
```diff
@@ -1,20 +1,17 @@
...
-        <img ng-src="{{follower.picUrl || 'https://www.kifi.com/assets/img/ghost.200.png'}}"
-          class="kf-lfp-pic"
-          kf-who-tooltip
-          keeper="follower"
-          ng-mouseenter="showTooltip()"
-          ng-mouseleave="hideTooltip()"
-          ng-click="hideTooltip()"
-          alt="{{follower.firstName}} {{follower.lastName}}"
-        ></a>
+    <a ng-repeat="follower in followersToShow" href="{{follower.profileUrl}}" kf-track-origin origin="{{currentPageOrigin}}/follower">
```
2is10, line 3: We can use the loop variable in the `ng-repeat` element itself, eliminating a span wrapper.

kifi-backend/modules/shoebox/angular/src/libraries/libraryFollowerPics.js
yipingliao, line 63: Thanks!

yipingliao:
:+1:

PR #13581, opened 2015-02-06
enable keep image cropping 

joshm1:
@andrewconner any early feedback you can give much appreciated.  this is still a WIP. I still need to persist if an image is from a crop or persist and add an way for a consumer to explicitly request a cropped image

cc @stkem  @eishay


kifi-backend/modules/common/test/com/keepit/common/images/Image4javaWrapperTest.scala
```diff
@@ -140,7 +189,7 @@ class Image4javaWrapperTest extends Specification with CommonTestInjector {
       val resized = im.resizeImage(image, ImageFormat.PNG, 500).get //not really resizing
       resized.getWidth === 500
       resized.getHeight === 333
-      range(imageByteSize(resized, "png"), 225392, 0.2)
+      range(imageByteSize(resized, "png"), 465305, 0.2)
```
joshm1, line 192: @eishay removing the `color(2048)` option from imagemagick makes the "optimized" version of this image the same size as the original

kifi-backend/modules/common/app/com/keepit/commanders/ProcessedImageHelper.scala
```diff
@@ -276,18 +325,49 @@ trait ProcessedImageHelper {
...
-  case object Small extends ProcessedImageSize("small", ImageSize(150, 150))
-  case object Medium extends ProcessedImageSize("medium", ImageSize(400, 400))
-  case object Large extends ProcessedImageSize("large", ImageSize(1000, 1000))
-  case object XLarge extends ProcessedImageSize("xlarge", ImageSize(1500, 1500))
+  val Small = ScaledImageSize.Small
+  val Medium = ScaledImageSize.Medium
+  val Large = ScaledImageSize.Large
+  val XLarge = ScaledImageSize.XLarge
 
   val allSizes: Seq[ProcessedImageSize] = Seq(Small, Medium, Large, XLarge)
```
joshm1, line 381: the methods in this object are unchanged and do not consider cropped images

joshm1:
updated to persist the process operation (scale/crop) and only crop KeepImage

andrewconner:
Have not forgotten about you, review coming!

kifi-backend/modules/common/app/com/keepit/model/BaseImage.scala
```diff
@@ -82,3 +82,18 @@ object ImageHash {
   implicit val format = Json.format[ImageHash]
 }
 
+sealed abstract class ProcessImageOperation(val kind: String, val fileNamePrefix: String)
+object ProcessImageOperation {
+  // None and Scale have the same fileNamePrefix for backwards compatibility
+  object None extends ProcessImageOperation("none", "")
```
andrewconner: Why have this one then?

kifi-backend/modules/common/app/com/keepit/commanders/ProcessedImageHelper.scala
```diff
@@ -66,20 +84,34 @@ trait ProcessedImageHelper {
...
+    def keygen(width: Int, height: Int, label: String = "", operation: ProcessImageOperation) = {
+      baseLabel + "/" + sourceImage.hash.hash + "_" +
+        operation.fileNamePrefix + width + "x" + height + label + "." + outFormat.value
     }
+
     validateAndLoadImageFile(sourceImage.file.file) match {
       case Success(image) =>
-        val resizedImages = calcSizesForImage(image).map { boundingBox =>
+        val sizes = calcSizesForImage(image, scaleCandidates, cropCandidates)
+        val resizedImages = sizes.map { boundingBox =>
```
andrewconner: perhaps call it `operation` or similar?

kifi-backend/modules/common/app/com/keepit/model/BaseImage.scala
joshm1: as long as we're okay classifying both the original image and scaled image the same, I'm fine with dropping this.  This is only for storing the original image.

kifi-backend/modules/common/app/com/keepit/commanders/ProcessedImageHelper.scala
andrewconner: `sizeRequest`?

kifi-backend/modules/common/app/com/keepit/commanders/ProcessedImageHelper.scala
```diff
@@ -109,18 +141,40 @@ trait ProcessedImageHelper {
...
+    val scaleOps = scaleSizes.sorted.flatMap { x =>
       if (x - t > 100) {
         t = x
         Some(x)
       } else None
-    }.filterNot(i => i == Math.max(image.getWidth, image.getHeight)).toSet
+    }.filterNot { i => i == Math.max(image.getWidth, image.getHeight) }.
+      map { x => ScaleRequest(x) }
+
+    val imageRatio = BigDecimal(image.getWidth) / image.getHeight
```
andrewconner: `BigDecimal` needed?

kifi-backend/modules/common/app/com/keepit/commanders/ProcessedImageHelper.scala
joshm1: yes this has been changed in my local dev. i have this among other changes coming soon

kifi-backend/modules/common/app/com/keepit/commanders/ProcessedImageHelper.scala
```diff
@@ -109,18 +141,40 @@ trait ProcessedImageHelper {
...
+    val cropSizes = cropCandidates.filterNot { cropSize =>
+      val imgHeight = image.getHeight
+      val imgWidth = image.getWidth
+      val size = cropSize.idealSize
+
+      // 1) if either the width or height of the actual image is smaller than our crop, abort
+      // 2) or, if the aspect ratio of the image and crop size are the same and there exists
+      //    a scale bounding box close to enough to our crop candidate, we can skip the crop
```
andrewconner, line 183: Not sure I agree with the logic here. Say we need a 200x200 image for an email, and the original is 210x200. This is "close enough" to 200x200, but wouldn't work — 10px still needs to be cut off.

kifi-backend/modules/common/app/com/keepit/commanders/ProcessedImageHelper.scala
```diff
@@ -276,18 +330,43 @@ trait ProcessedImageHelper {
   }
 }
 
-sealed abstract class ProcessedImageSize(val name: String, val idealSize: ImageSize)
+sealed abstract class ProcessedImageSize(val name: String, val idealSize: ImageSize, val operation: ProcessImageOperation) {
+  lazy val aspectRatio = BigDecimal(idealSize.width) / idealSize.height
```
andrewconner: `BigDecimal`?

kifi-backend/modules/common/app/com/keepit/commanders/ProcessedImageHelper.scala
andrewconner: Ah great

kifi-backend/modules/common/app/com/keepit/model/BaseImage.scala
andrewconner: Perhaps we could use "o" for original, and keep it? I _believe_ backwards compatibility isn't needed because we store the key in the db. I can double check if you doubt it.

andrewconner:
Awesome start. It's pretty tricky to do right, and you've done a great job.

kifi-backend/modules/common/app/com/keepit/commanders/ProcessedImageHelper.scala
joshm1: I'm not sure.  Because we're dealing with division we'd have to use float/doubles, and I wasn't sure if we'd run into cases where 2 identical aspect ratios aren't represented as the same.

this works, but are their other examples that don't? `417/303d == 278/202d`

kifi-backend/modules/common/app/com/keepit/commanders/ProcessedImageHelper.scala
joshm1, line 183: the crop 200x200 wouldn't pass the aspect ratio check: `cropSize.aspectRatio == imageRatio`.  the logic is that only scaled images are allowed to be good enough if the original image has the same aspect ratio as the crop candidate

andrewconner, line 183: I see, I missed the first condition: "if the aspect ratio of the image and crop size are the same"

kifi-backend/modules/common/app/com/keepit/model/BaseImage.scala
joshm1: i think you're right, that's what I saw as well.

andrewconner: Cool. In that case, `originalLabel` can be removed.

joshm1:
@andrewconner @stkem this PR has been updated quite a bit since what I sent last Friday, mostly dealing with fetching cropped images for keeps.  

The updates are in this commit: https://github.com/joshm1/keepit/commit/659d867f4c32bba0942f1f3326ee32ea4361499f

Review/feedback much appreciated.

kifi-backend/modules/shoebox/app/com/keepit/commanders/emails/ActivityFeedEmailSender.scala
```diff
@@ -207,39 +183,97 @@ class ActivityFeedEmailSenderImpl @Inject() (
...
+
+  // gets cropped keep images for a library
+  private def fetchLibToKeepImages(libInfoView: LibraryInfoView): Future[LibraryInfoViewWithKeepImages] = {
+    def recur(target: Int, offset: Int, limit: Int): Future[Seq[String]] = {
+      if (target > 0) {
+        val keepsF = libraryCommander.getKeeps(libInfoView.libraryId, offset, limit)
+
+        keepsF flatMap { keeps =>
+          val keepIds = keeps.map(_.id.get).toSet
+          val urls = keepImageCommander.getBestImagesForKeeps(keepIds, CropImageRequest(CroppedImageSize.Medium.idealSize)).collect {
```
joshm1: the problem with this is that it won't find the cropped image for existing keep images and it won't create it either.. i need to use another method that creates the cropped image and waits on it

kifi-backend/modules/common/app/com/keepit/commanders/ProcessedImageHelper.scala
```diff
@@ -109,24 +152,51 @@ trait ProcessedImageHelper {
...
+      }.filterNot { i => i == Math.max(imageSize.width, imageSize.height) }.
+        map { x => ScaleImageRequest(x) }
+    }
+
+    val imgHeight = imageSize.height
+    val imgWidth = imageSize.width
+
+    val cropImageRequests = cropCandidates.filterNot { cropSize =>
+      val size = cropSize.idealSize
+      def isAlmostSameAspectRatio = Math.abs(imgWidth.toFloat / imgHeight - cropSize.aspectRatio) < 0.01
```
andrewconner, line 179: :+1:

andrewconner:
Whew, lots to take in here. Seems good, merge when ready. Grab me when deployed, we'll monitor closely what images start coming through. Adding some temporary logging may help quickly diagnose issues.

PR #13586, opened 2015-02-06
Rework lib menu

yipingliao:
@2is10


2is10:
:+1: Thanks a bunch. Makes a big difference to the experience.

PR #13591, opened 2015-02-07
Add return user FTUE for the library menu.

yipingliao:
@2is10 Not activated yet; needs backend support.


kifi-backend/modules/shoebox/angular/src/app.js
```diff
@@ -122,6 +122,15 @@ angular.module('kifi', [
       $scope.libraryMenu = {
         visible: false
       };
+
+      if (false) {  // Needs condition from backend.
```
yipingliao, line 130: Set this to true for testing. What it should be is something like `if (profileService.me.showReturnLibMenuFTUE)`'.

kifi-backend/modules/shoebox/angular/src/layout/libraryMenu/libraryMenu.js
```diff
@@ -270,6 +270,7 @@ angular.module('kifi')
         scope.$watch('libraryMenu.visible', function (visible) {
           if (visible) {
             openMenu();
+            scope.libraryMenuHelp.visible = false;
```
yipingliao, line 273: Close FTUE on library menu open.

kifi-backend/modules/shoebox/angular/src/layout/libraryMenu/libraryMenuHelp.js
```diff
@@ -0,0 +1,36 @@
...
+      link: function (scope , element /*, attrs*/) {
+        function setLeftPosition() {
+          element.css({'left': angular.element('.kf-lih-toggle-menu').offset().left - 13 + 'px'});
+        }
+
+        scope.$watch('libraryMenuHelp.visible', function (visible) {
+          if (visible) {
+            setLeftPosition();
+          } else {
+            // Let the server know that the user has seen this menu.
```
yipingliao, line 20: Needs backend endpoint here.

kifi-backend/modules/shoebox/angular/src/app.js
```diff
@@ -122,6 +122,15 @@ angular.module('kifi', [
...
+        $timeout(function () {
+          $scope.libraryMenuHelp = {
+            visible: true
+          };
+        }, 2000);
```
yipingliao, line 133: Wait for 2 seconds for the user to realize that something is different before showing the pop-up.

2is10:
Great, thanks.

PR #13594, opened 2015-02-07
bring back some old code supporting goodies

2is10:
@yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/layout/main/MainCtrl.js
+++ b/kifi-backend/modules/shoebox/angular/src/layout/main/MainCtrl.js
@@ -5,66 +5,20 @@ angular.module('kifi')
 .controller('MainCtrl', [
   '$scope', '$element', '$window', '$location', '$timeout', '$rootElement', 'undoService', 'keyIndices',
   'initParams', '$rootScope', '$analytics', 'installService', 'profileService', '$q', 'routeService',
-  'modalService', 'libraryService',
+  'modalService', 'libraryService', 'tagService',
   function ($scope, $element, $window, $location, $timeout, $rootElement, undoService, keyIndices,
     initParams, $rootScope, $analytics, installService, profileService, $q, routeService,
-    modalService, libraryService) {
+    modalService, libraryService, tagService) {
 
     var tooltipMessages = {
       0: 'Welcome back!',
       2: 'Bookmark import in progress. Reload the page to update.'
     };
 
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/shoebox/angular/src/layout/main/MainCtrl.js
```diff
@@ -5,67 +5,21 @@ angular.module('kifi')
...
-      }
-
-      // hacky solution to url event not getting fired
-      $timeout(function () {
-        $scope.$apply();
-      });
-    }
-
-    $scope.onChange = _.debounce(performSearch, 350);
-
```
2is10, line 68: As far as I can tell, all this code was unused, maybe leftover from when this controller handled search instead of the headers.

kifi-backend/modules/shoebox/angular/src/layout/main/MainCtrl.js
```diff
@@ -309,6 +263,29 @@ angular.module('kifi')
...
+          case 'update_keeps':
+          case 'update_tags':
+            $timeout.cancel(refreshTimeout);
+            refreshTimeout = $timeout(function () {
+              tagService.fetchAll(true);
+            }, 1000); // Giving enough time for the services to be updated
+            break;
+        }
+      });
+    });
```
2is10, line 288: Listening for messages from the extension — this needs to live somewhere persistent. Any other ideas?

kifi-backend/modules/shoebox/angular/src/libraries/library.js
```diff
@@ -186,8 +186,12 @@ angular.module('kifi')
...
       $rootScope.$emit('showGlobalModal', 'importBookmarkFile');
     };
-    $scope.callTriggerInstall = function () {
-      $rootScope.$emit('triggerExtensionInstall');
+    $scope.triggerInstall = function () {
+      installService.triggerInstall(function () {
+        modalService.open({
+          template: 'common/modal/installExtensionErrorModal.tpl.html'
+        });
+      });
```
2is10, line 194: can no longer pass a message to the right column, since it’s not always there (only on recommendations page), so just using `installService` directly

kifi-backend/modules/shoebox/angular/src/layout/main/MainCtrl.js
yipingliao, line 68: Agreed.

PR #13597, opened 2015-02-07
Remove reco recency slider; turn refresh into see more.

yipingliao:
@2is10


kifi-backend/modules/shoebox/angular/src/recos/recoStateService.js
```diff
@@ -10,6 +10,14 @@ angular.module('kifi')
       recosList: savedRecos,
 
       populate: function (recos) {
+        _.remove(recos, function (reco) {
```
yipingliao, line 13: Deduplication.

kifi-backend/modules/shoebox/angular/src/recos/recos.styl
```diff
@@ -365,6 +365,21 @@
   label
     margin-left: 5px
 
+.kf-reco-get-more-container
+  margin-bottom: 400px
```
yipingliao, line 369: Setting a huge margin at the bottom results in a nice experience where, when the new recos are loaded, the page does not scroll or shift but the top new reco card replaces the "See more" section and you can read most of it. The downside is that it looks so empty down there! One idea is to ask product to come up with some non-call-to-action asset we can display (like hints and shortcuts, maybe? "Did you know...?") so the space will look nice.

kifi-backend/modules/shoebox/angular/src/recos/recoStateService.js
```diff
@@ -10,6 +10,14 @@ angular.module('kifi')
...
+          return _.some(savedRecos, function (savedReco) {
+            return (
+              reco.recoLib && savedReco.recoLib && (reco.recoLib.url === savedReco.recoLib.url) ||
+              reco.recoKeep && savedReco.recoKeep && (reco.recoKeep.url === savedReco.recoKeep.url)
+            );
+          });
+        });
         savedRecos.push.apply(savedRecos, recos);
```
2is10, line 21: This is O(`m × n`) where `m` is number already showing and `n` is number just loaded. It could be reduced to O(`n`) if we kept a set of URLs of the recommendations currently showing.

``` js
var recoUrls = {};
...
populate: function (recos) {
  for (var i = 0; i < recos.length; i++) {
    var reco = recos[i];
    var url = (reco.recoLib || reco.recoKeep).url;
    if (!recoUrls[url]) {
      recoUrls[url] = true;
      savedRecos.push(reco);
    }
  }
},
```

kifi-backend/modules/shoebox/angular/src/recos/recos.js
```diff
@@ -34,6 +35,24 @@ angular.module('kifi')
...
+          var recos = [];
+          rawRecos.forEach(function (rawReco) {
+            recos.push(recoDecoratorService.newUserRecommendation(rawReco));
+          });
+          recoStateService.populate(recos);
+        } else {
+          $scope.noMoreRecos = true;
+        }
+
+        $scope.moreLoading = false;
```
2is10, line 51: better to put in a `finally`, so page can recover from a server error?

kifi-backend/modules/shoebox/angular/src/recos/recos.styl
2is10, line 369: One idea would be to use a blank recommendation card as a placeholder, kind of how facebook uses feed story placeholders while first loading stories. The “load more” button could be right in the middle of it, or the whole thing.

2is10:
Great, nice work. :+1:

kifi-backend/modules/shoebox/angular/src/recos/recoStateService.js
yipingliao, line 21: Thanks! Will update.

kifi-backend/modules/shoebox/angular/src/recos/recos.js
yipingliao, line 51: Will do. Thanks!

kifi-backend/modules/shoebox/angular/src/recos/recos.styl
yipingliao, line 369: Good idea! Will try.

PR #13613, opened 2015-02-09
avoid outliers in library induced score

yingjieMiao:
@stkem


stkem:
lgtm. Please log it in the algorithm change google doc.

yingjieMiao:
yep.

PR #13615, opened 2015-02-09
making site header look okay when extension pane is open

2is10:
@yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/header/loggedOutHeader.styl
+++ b/kifi-backend/modules/shoebox/angular/src/header/loggedOutHeader.styl
@@ -12,6 +12,12 @@ search-dur = .2s
   z-index: 500
   user-select: none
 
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/angular/src/header/loggedOutHeader.styl
```diff
@@ -12,6 +12,12 @@ search-dur = .2s
   z-index: 500
   user-select: none
 
+  [kifi-pane-parent] &
+    transition: width .25s ease-in
+
+  [kifi-with-pane] &
+    width: calc(100% - 322px)
+
```
2is10, line 20: This generally shouldn’t be needed, since opening the pane requires the user to be authenticated with kifi.com, but there could be weird corner cases, like the user visiting a kifi.com page and then logging in there using the extension’s login form.

PR #13617, opened 2015-02-09
Just a couple of very small changes

yipingliao:
@2is10 

Just wanted to get in a couple of small changes.

Update: up next I am going to go through all the banners I found while grepping for `banner`. Also, `/invite` is too wide so I'll bring that inline with `/friends`.


2is10:
:+1:

PR #13621, opened 2015-02-09
move the ScrapeInfo related logic from OrphanCleaner to ScrapeInfoIntegrityChecker

ymatsuda:
@eishay @andrewconner @stkem


andrewconner:
Great, looks good with me

PR #13622, opened 2015-02-09
Update /invite styles to be like the styles in /friends.

yipingliao:
@2is10 Just a few small style updates to clean up the /invite page.


2is10:
:+1:

PR #13624, opened 2015-02-09
using old callout styles for new callout, hooking up to pref

2is10:
moves the code from #13591 into the logged-in header
@yipingliao


kifi-backend/modules/shoebox/angular/src/profile/profileService.js
```diff
@@ -241,7 +241,6 @@ angular.module('kifi')
       return $http.get(routeService.prefs).then(function (p) {
         var oldPrefs = _.clone(prefs);
         util.replaceObjectInPlace(prefs, p.data);
-        prefs.site_intial_show_library_intro = !!prefs.site_show_library_intro;
```
2is10, line 244: a final remnant from the old callouts

kifi-backend/modules/shoebox/angular/src/header/loggedInHeader.tpl.html
```diff
@@ -1,6 +1,12 @@
 <div class="kf-lih" ng-controller="LoggedInHeaderCtrl">
   <div class="kf-lih-inner">
     <div class="kf-lih-left">
+      <div class="kf-callout-lib-menu kf-callout-bubble kf-callout-bubble-bottom" ng-if="showCallout()">
+        We’ve tidied up a bit.<br>Your libraries are now<br>listed here.
+        <div class="kf-callout-bubble-action">
+          <button class="kf-callout-bubble-button" ng-click="closeCallout()">Got it</button>
+        </div>
+      </div>
```
2is10, line 9: A nice benefit of putting this right next to the button is that we don’t need any JS code to reposition it when the button moves.

kifi-backend/modules/shoebox/angular/src/common/directives/callout/callout.js
```diff
@@ -1,21 +0,0 @@
...
-      templateUrl: 'common/directives/callout/callout.tpl.html',
-      link: function (scope, element, attrs) {
-        if (typeof attrs.relPos !== 'undefined') {
-          var wrap = angular.element('<div>').css('position', 'relative');
-          element.wrap(wrap);
-        }
-      }
-    };
-  }
-]);
```
2is10, line 21: This directive wasn’t doing enough to justify its existence. Now it’s just some reusable styles.

kifi-backend/modules/shoebox/angular/src/header/loggedInHeader.tpl.html
yipingliao, line 9: Nice!

kifi-backend/modules/shoebox/angular/src/layout/libraryMenu/libraryMenu.js
```diff
@@ -259,7 +259,6 @@ angular.module('kifi')
         scope.$watch('libraryMenu.visible', function (visible) {
           if (visible) {
             openMenu();
-            scope.libraryMenuHelp.visible = false;
```
yipingliao, line 262: This logic was not ported over to `loggedInHeader.js`, so if the user clicks the burger sprite the help menu won't go away. Also, if the user happens to click the burger sprite in the first 2.4 seconds (highly unlikely) the help menu will pop above it.

2is10, line 262: Good catch!  Thanks, fixing.

PR #13625, opened 2015-02-09
Fix header on mobile pages

yipingliao:
@2is10 And this takes care of profile pages, but library pages needs more work (unfortunately I misspoke when I said mobile look okay - I was only looking at profile and so my statement is way too optimistic!)


kifi-backend/modules/shoebox/angular/src/header/loggedOutHeader.tpl.html
```diff
@@ -4,7 +4,7 @@
       <a class="kf-loh-logo svg-kifi-clover-dark-text" ui-sref="home" ng-click="clickLogo()"></a>
       <span class="kf-loh-tagline">Connecting people<br>with knowledge</span>
     </div>
-    <div class="kf-loh-center" ng-if="library">
+    <div class="kf-loh-center" ng-if="library && !isMobile">
```
yipingliao, line 7: We don't currently have search on mobile library page, so taking it out is not a regression. We probably need search to look different on mobile to make it friendly to users.

2is10:
:+1:

PR #13626, opened 2015-02-09
showing delighted survey on recos page only

2is10:
@yipingliao


kifi-backend/modules/shoebox/angular/src/delighted/delighted.js
```diff
@@ -68,7 +67,7 @@ angular.module('kifi')
         };
 
         function hideSurvey() {
-          scope.showSurvey = false;
+          scope.hide();
```
yipingliao, line 70: Nice! I like this way of changing the parent's state a lot.

PR #13629, opened 2015-02-10
Fix Claim Account with Twitter on Signup

ahsu1230:
@andrewconner 

In Dev mode, signing up has been looking good.
Hope this fixes the Twitter Account Claim flow

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileAuthController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileAuthController.scala
@@ -166,15 +166,21 @@ class MobileAuthController @Inject() (
   }
 
   def loginWithUserPass(link: String) = MaybeUserAction.async(parse.anyContent) { implicit request =>
-    ProviderController.authenticate("userpass")(request).map {
+    ProviderController.authenticate("userpass")(request).flatMap {
       case res: Result if res.header.status == 303 =>
         authHelper.transformResult(res) { (cookies: Seq[Cookie], sess: Session) =>
-          val newSession = if (link != "") {
-            sess - SecureSocial.OriginalUrlKey + (AuthController.LinkWithKey -> link) // removal of OriginalUrlKey might be redundant
-          } else sess
-          Ok(Json.obj("code" -> "auth_success")).withCookies(cookies: _*).withSession(newSession)
+          if (link != "") {
+            ProviderController.authenticate(link)(request).recover {
+              case _ =>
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileAuthController.scala
```diff
@@ -166,15 +166,21 @@ class MobileAuthController @Inject() (
...
         authHelper.transformResult(res) { (cookies: Seq[Cookie], sess: Session) =>
-          val newSession = if (link != "") {
-            sess - SecureSocial.OriginalUrlKey + (AuthController.LinkWithKey -> link) // removal of OriginalUrlKey might be redundant
-          } else sess
-          Ok(Json.obj("code" -> "auth_success")).withCookies(cookies: _*).withSession(newSession)
+          if (link != "") {
+            ProviderController.authenticate(link)(request).recover {
+              case _ =>
+                log.info(s"[logInWithUserPassMobile] recover from authenticated link for user [${request.userIdOpt} ${request.identityOpt}] through ${link}")
+                Ok(Json.obj("code" -> "auth_success")).withCookies(cookies: _*).withSession(sess)
```
andrewconner, line 176: This can be factored out... you don't actually care about the result here.

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileAuthController.scala
```diff
@@ -166,15 +166,21 @@ class MobileAuthController @Inject() (
...
-    ProviderController.authenticate("userpass")(request).map {
+    ProviderController.authenticate("userpass")(request).flatMap {
       case res: Result if res.header.status == 303 =>
         authHelper.transformResult(res) { (cookies: Seq[Cookie], sess: Session) =>
-          val newSession = if (link != "") {
-            sess - SecureSocial.OriginalUrlKey + (AuthController.LinkWithKey -> link) // removal of OriginalUrlKey might be redundant
-          } else sess
-          Ok(Json.obj("code" -> "auth_success")).withCookies(cookies: _*).withSession(newSession)
+          if (link != "") {
+            ProviderController.authenticate(link)(request).recover {
```
andrewconner, line 173: Did some more thinking... not sure this will work. `request` here doesn't yet have a userId (or at least, shouldn't). This is the login action.

The easiest and cleanest solution I can think of is to manually link the account. ie, find the link social user info record (in our case from yesterday, twitter), and assign the userId.

andrewconner:
Thanks for investigating. Let's try to get this out asap so twitter is not blocked. Grab me and we'll work on it together.

PR #13636, opened 2015-02-10
Make a placeholder reco card to wrap around "Show more" link on recommendation.

yipingliao:
Also, if there are no more recommendations, tell the user "No more recommendations right now."

Conducted UI review with Mark.

@2is10


yipingliao:
screen shot 2015-02-09 at 5 30 18 pm

2is10:
:+1: very nice!

PR #13642, opened 2015-02-10
better contrast logged-in header buttons, svg icons

2is10:
@andrewconner


kifi-backend/modules/shoebox/angular/img/svg/tag-black.svg
```diff
@@ -1,7 +0,0 @@
-<?xml version="1.0" encoding="UTF-8" standalone="no"?>
-<!-- Generator: Avocode 0.4.0 - http://avocode.com -->
-<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="31" height="31" view-box="0 0 31 31">
```
2is10, line 3: FYI, I don’t think `view-box` is a valid attribute name. Avocode seems to generate it. Chrome doesn’t honor it. `viewBox` and `viewbox` work.
http://www.w3.org/TR/SVG11/coords.html#ViewBoxAttribute

PR #13647, opened 2015-02-10
Update logic for deciding how to show/hide CTAs to trigger extension installation and Kifi guide

yipingliao:
@2is10

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/common/services/installService.js
+++ b/kifi-backend/modules/shoebox/angular/src/common/services/installService.js
@@ -19,6 +19,7 @@ angular.module('kifi')
         if (data.type === 'kifi_ext_listening') {
           $rootScope.$apply(function () {
             api.installedVersion = data.version;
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/shoebox/angular/src/common/services/installService.js
```diff
@@ -19,6 +19,7 @@ angular.module('kifi')
         if (data.type === 'kifi_ext_listening') {
           $rootScope.$apply(function () {
             api.installedVersion = data.version;
+            $rootScope.$emit('kifiExt');
```
2is10, line 22: extraneous?

yipingliao, line 22: This is here because when the extension is first installed we want to trigger a check on the minimum version.

2is10, line 22: ah, yes, of course. thanks.

PR #13651, opened 2015-02-10
killing confusing kf-max-version directive

2is10:
@yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/common/directives/minVersion.js
+++ b/kifi-backend/modules/shoebox/angular/src/common/directives/minVersion.js
@@ -9,37 +9,8 @@ angular.module('kifi')
       restrict: 'A',
       link: function (scope, element, attrs) {
         scope.$on('$destroy', $rootScope.$on('kifiExt', function updateNodeDisplay() {
-          if (element.length && element[0].style) {
-            if (!installService.hasMinimumVersion(attrs.kfMinVersion, attrs.minCanary)) {
-              element[0].style.display = 'none';
-            } else {
-              element[0].style.display = 'initial';
-            }
-          }
+          element.css('display', installService.hasMinimumVersion(attrs.kfMinVersion, attrs.minCanary) ? '' : 'none');
         }));
-      }
-    };
-  }
-])
-
-.directive('kfMaxVersion', [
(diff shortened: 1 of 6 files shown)


kifi-backend/modules/shoebox/angular/src/common/directives/minVersion.js
```diff
@@ -9,38 +9,9 @@ angular.module('kifi')
...
       link: function (scope, element, attrs) {
         scope.$on('$destroy', $rootScope.$on('kifiExt', function updateNodeDisplay() {
-          if (element.length && element[0].style) {
-            if (!installService.hasMinimumVersion(attrs.kfMinVersion, attrs.minCanary)) {
-              element[0].style.display = 'none';
-            } else {
-              element[0].style.display = 'initial';
-            }
-          }
+          element.css('display', installService.hasMinimumVersion(attrs.kfMinVersion, attrs.minCanary) ? '' : 'none');
```
2is10, line 12: Shortening, but also changing `'initial'` to `''` here, which clears the inline style. This way regular style rules will again apply to the element.

kifi-backend/modules/shoebox/angular/src/common/directives/minVersion.js
```diff
@@ -9,38 +9,9 @@ angular.module('kifi')
...
-              element[0].style.display = 'initial';
-            }
-          }
-        }
-        // See note above.
-        hideNode();
-        _.map([1, 150, 350, 1000], function (delay) { $timeout (hideNode, delay); });
-      }
-    };
-  }
```
2is10, line 45: This directive was confusing (using `hasMinimumVersion`), used the wrong element attributes (`kfMinVersion` instead of `kfMaxVersion`?), and isn’t updated on install like `kf-min-version` now is. It’s also not easy to intuit that the directive still shows the element if no kifi extension is installed. (One might guess it only shows the element if an old extension is installed.)

kifi-backend/modules/shoebox/angular/src/common/directives/minVersion.js
yipingliao, line 12: Cool!

yipingliao:
:+1:

PR #13652, opened 2015-02-10
Prevent body from scrolling if library menu is short (itself not scrolling)

yipingliao:
@2is10 cc @andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/layout/libraryMenu/libraryMenu.js
+++ b/kifi-backend/modules/shoebox/angular/src/layout/libraryMenu/libraryMenu.js
@@ -288,6 +288,18 @@ angular.module('kifi')
           $window.removeEventListener(hideAndReopenOnResize);
         });
 
+        var preventBodyScrollInMenu = function (event) {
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/angular/src/layout/libraryMenu/libraryMenu.js
```diff
@@ -288,6 +288,18 @@ angular.module('kifi')
           $window.removeEventListener(hideAndReopenOnResize);
         });
 
+        var preventBodyScrollInMenu = function (event) {
+          if (element[0].contains(event.target)) {
+            document.body.style.overflow = 'hidden';
```
2is10, line 293: This approach caused scrollbars to disappear, which moves everything on the page in browsers/OSes that always show scrollbars. Let’s talk about alternatives.

2is10, line 293: I’ve addressed this another way in #13655.

PR #13653, opened 2015-02-10
fix match error

ymatsuda:
@andrewconner


andrewconner:
Thanks! Sad the compiler doesn't fail on this.

PR #13654, opened 2015-02-10
not indexing users with empty profile

eishay:
@andrewconner @2is10

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/common/seo/UserSiteMapGenerator.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/common/seo/UserSiteMapGenerator.scala
@@ -26,7 +26,7 @@ import scala.concurrent.duration._
 import play.api.libs.concurrent.Execution.Implicits.defaultContext
 
 object UserSiteMapGenerator {
-  val BaselineDate = new LocalDate(2015, 1, 28)
+  val BaselineDate = new LocalDate(2015, 2, 10)
 }
 
 @Singleton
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/shoebox/app/com/keepit/common/seo/UserSiteMapGenerator.scala
```diff
@@ -78,10 +79,15 @@ class UserSiteMapGenerator @Inject() (airbrake: AirbrakeNotifier,
...
       userIds.grouped(500) foreach { group =>
         val realUsers = group.filterNot(fakeUsers.contains)
-        db.readOnlyMaster { implicit ro =>
+        db.readOnlyReplica { implicit ro =>
           userRepo.getAllUsers(realUsers.toSeq).values.toSeq
         } filter { user =>
           user.state == UserStates.ACTIVE
+        } filterNot { user =>
+          val countLibraries = db.readOnlyMaster { implicit s =>
+            libraryMembershipRepo.countWithUserIdAndAccess(user.id.get, LibraryAccess.OWNER) + libraryMembershipRepo.countWithUserIdAndAccess(user.id.get, LibraryAccess.READ_ONLY)
```
2is10, line 88: This is a rough approximation of whether the profile will be empty—a good start. A user could have multiple private libraries and/or public libraries hidden on their profiles.

eishay, line 88: True, its more of a gauge checking that the user is somewhat active.

PR #13660, opened 2015-02-10
Fix tags on library pages not going to the tag search results.

yipingliao:
@2is10


kifi-backend/modules/shoebox/angular/src/header/loggedInHeader.js
```diff
@@ -47,13 +47,12 @@ angular.module('kifi')
           $scope.search.libraryChip = true;
         } else if ($scope.search.libraryChip) {
           removeLibraryChip();
-        } else if ($state.params && $state.params.q && util.startsWith($state.params.q, 'tag:')) {
```
yipingliao, line 50: This doesn't seem to be needed here.

PR #13664, opened 2015-02-10
Up next: make searchers type safe :)

leogrim:
@eishay


leogrim:
[Opened](https://app.asana.com/0/14904390416467/24366638193185) on Jan 16th to prevent these mistakes :)

PR #13666, opened 2015-02-10
Do not show libraries with no keeps on profiles

andrewconner:
Sorry for the bad commit history. This makes anonymous people not see empty libraries.

@eishay, there's a bit to do here: See `fixLibraryReferences` below. Unfortunately, this breaks a lot of stuff that isn't expecting random libraries being created.


andrewconner:
@ahsu1230

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -919,21 +919,23 @@ class LibraryCommander @Inject() (
...
-            libraryAnalytics.unfollowLibrary(userId, lib, keepCount, eventContext)
-            searchClient.updateLibraryIndex()
-          }
-          Right()
+          (keepRepo.getCountByLibrary(libraryId), lib)
         }
+        SafeFuture {
+          libraryAnalytics.unfollowLibrary(userId, lib, keepCount, eventContext)
+          searchClient.updateLibraryIndex()
+        } // todo(andrew)!
```
ahsu1230: you can remove that comment now! =D

kifi-backend/modules/shoebox/test/com/keepit/commanders/LibraryCommanderTest.scala
```diff
@@ -1314,17 +1320,22 @@ class LibraryCommanderTest extends TestKitSupport with SpecificationLike with Sh
           libraries(2).map(_.secret().withUser(other)).saved
           val ownerLibs2 = libraries(10).map(_.published().withUser(owner)).saved
           libraries(10).map(_.published().withUser(other)).saved
+
+          libraryRepo.all.take(4).map { lib =>
+            keeps(2).map(_.withLibrary(lib)).saved
+          }
```
ahsu1230, line 1326: doesn't this mean 4 libraries have keeps in them? The anonymous user sees 2. Is it because those libraries are private / hidden? We should have a comment or something to indicate that

ahsu1230:
cool, thanks!

kifi-backend/modules/shoebox/test/com/keepit/commanders/LibraryCommanderTest.scala
andrewconner, line 1326: Yep. First two are private, next two are public. Will comment. Thanks

andrewconner:
Added comments for the test

PR #13676, opened 2015-02-11
Fix search; update CSS styles for cleaned up library menu.

yipingliao:
@2is10 Phew. I hope this is the last of the search fixes for a while. :)

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/header/loggedInHeader.js
+++ b/kifi-backend/modules/shoebox/angular/src/header/loggedInHeader.js
@@ -42,14 +42,7 @@ angular.module('kifi')
         $scope.library = library;
         $scope.libOwnerPicUrl = library && routeService.formatPicUrl(library.owner.id, library.owner.pictureName, 100);
         $scope.libOwnerProfileUrl = library && routeService.getProfileUrl(library.owner.username);
-
-        if (library) {
-          $scope.search.libraryChip = true;
-        } else if ($scope.search.libraryChip) {
-          removeLibraryChip();
-        } else if ($state.params && $state.params.q && util.startsWith($state.params.q, 'tag:')) {
-          $scope.search.text = $state.params.q;
-        }
+        $scope.search.libraryChip = !!library;
       }),
 
       $rootScope.$on('$stateChangeSuccess', function (event, toState, toParams) {
@@ -87,25 +80,19 @@ angular.module('kifi')
     }
 
     function reactToQueryChange() {
-      if ($location.path() === '/find') {
-        $location.search('q', $scope.search.text).replace(); // this keeps any existing URL params
-      } else if ($scope.search.libraryChip) {
-        if ($scope.search.text) {
-          if ($state.params.q) {
-            $location.search('q', $scope.search.text).replace();
-          } else {
-            $location.url($scope.library.url + '/find?q=' + $scope.search.text + '&f=a');
-          }
-        } else {
-          $location.url($scope.library.url);
-        }
-      } else if ($scope.search.text) {
-        $location.url('/find?q=' + $scope.search.text);
+      if ($state.is('search') || $state.is('library.search')) {
+        $rootScope.$emit('searchTextUpdated', $scope.search.text, !!$scope.search.libraryChip && $scope.library.url);
+      } else {
+        $state.go($scope.search.libraryChip ? 'library.search' : 'search', {q: $scope.search.text});
       }
     }
 
     $scope.search.text = $state.params.q || '';
     $scope.onQueryChange = util.$debounce($scope, reactToQueryChange, 250);
+
+    $scope.$on('$destroy', $rootScope.$on('newQueryFromLocation', function (e, query) {
+      $scope.search.text = query;
+    }));
 
     $scope.onSearchBarClicked = function () {
       if ($scope.library) {
--- a/kifi-backend/modules/shoebox/angular/src/search/search.js
+++ b/kifi-backend/modules/shoebox/angular/src/search/search.js
@@ -252,26 +252,52 @@ angular.module('kifi')
       }
     };
 
+    $scope.onClickSearchFilter = function (newSearchFilter) {
+      $location.search({f: newSearchFilter});
+    };
+
 
     //
     // Watches and event listeners.
     //
-    var newSearch = _.debounce(function () {
+    function newSearch() {
       // Use $state.params instead of $stateParams because changes to $stateParams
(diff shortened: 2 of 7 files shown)


kifi-backend/modules/shoebox/angular/src/header/loggedInHeader.js
```diff
@@ -87,26 +80,20 @@ angular.module('kifi')
...
-          }
-        } else {
-          $location.url($scope.library.url);
-        }
-      } else if ($scope.search.text) {
-        $location.url('/find?q=' + $scope.search.text);
+      if ($state.is('search') || $state.is('library.search')) {
+        $rootScope.$emit('searchTextUpdated', $scope.search.text, !!$scope.search.libraryChip && $scope.library.url);
+      } else {
+        $state.go($scope.search.libraryChip ? 'library.search' : 'search', {q: $scope.search.text});
```
2is10, line 86: In the `library.search` case, are the other state parameters (`:username` and `:librarySlug`) automatically inherited from the current state?

kifi-backend/modules/shoebox/angular/src/search/search.js
```diff
@@ -252,27 +252,53 @@ angular.module('kifi')
...
+    };
+
 
     //
     // Watches and event listeners.
     //
-    var newSearch = _.debounce(function () {
+    function newSearch() {
       // Use $state.params instead of $stateParams because changes to $stateParams
       // does not propagate to HeaderCtrl when it is injected there.
```
2is10, line 265: The header controllers no longer inject `$stateParams` — remove this comment?

kifi-backend/modules/shoebox/angular/src/search/search.js
```diff
@@ -252,27 +252,53 @@ angular.module('kifi')
...
+    $scope.onClickSearchFilter = function (newSearchFilter) {
+      $location.search({f: newSearchFilter});
+    };
+
 
     //
     // Watches and event listeners.
     //
-    var newSearch = _.debounce(function () {
+    function newSearch() {
```
2is10, line 263: Thanks for removing this debounce. It’s redundant with the ones in the headers and was sometimes delaying searches unnecessarily.

kifi-backend/modules/shoebox/angular/src/search/search.js
```diff
@@ -252,27 +252,53 @@ angular.module('kifi')
...
       // Use $state.params instead of $stateParams because changes to $stateParams
       // does not propagate to HeaderCtrl when it is injected there.
       // See: http://stackoverflow.com/questions/23081397/ui-router-stateparams-vs-state-params
       _.assign($state.params, $location.search());
```
2is10, line 267: Might be good to make explicit here which parameters we care about and are trying to copy. Just `q` and `f`?

``` js
_.assign($state.params, _.pick($location.search(), 'q', 'f'));
```

A comment explaining that this is here because we set `reloadOnSearch: false` for this controller might also be helpful.

kifi-backend/modules/shoebox/angular/src/search/search.js
```diff
@@ -252,27 +252,53 @@ angular.module('kifi')
...
       init();
-    }, 250);
+    }
+
+    $scope.$on('$destroy', $rootScope.$on('searchTextUpdated', function (e, newSearchText, libraryUrl) {
```
2is10, line 271: Passing `libraryUrl` to this controller from the headers in a `searchTextUpdated` event feels a bit like the tail wagging the dog. This controller already knows its state. It has an injected `library` object (may be `null`) from which we can easily format the URL if necessary, no?

kifi-backend/modules/shoebox/angular/src/search/search.js
```diff
@@ -252,27 +252,53 @@ angular.module('kifi')
...
+            $location.search('q', newSearchText).replace();
+          } else {
+            $location.url(libraryUrl + '/find?q=' + newSearchText + '&f=a');
+          }
+        } else {
+          $location.url(libraryUrl);
+        }
+      } else if (newSearchText) {
+        $location.url('/find?q=' + newSearchText);
+      }
```
2is10, line 286: Do we really need all of this logic? If only the query text has changed, might it suffice to just:

``` js
  if ($state.params.q !== newText) {
    if (newText) {
      $state.params.q = newText;
      init();
      // Alternatively, if the two lines above don’t work:
      // $location.search('q', newText).replace();
      // newSearch();
    } else {
      $state.go(library ? 'library' : 'recommendations');
    }
  }
```

2is10:
This pull request is a huge improvement. Much clearer what’s going on. Thank you!!

My comments are minor and I’m fine with you merging this pull request as is and then you (or me) considering my questions/comments later.

yipingliao:
I am going to merge and then come back to this in just a little bit.

kifi-backend/modules/shoebox/angular/src/header/loggedInHeader.js
yipingliao, line 86: Apparently! I've tested this case and it works.

kifi-backend/modules/shoebox/angular/src/search/search.js
yipingliao, line 265: `$stateParams` was never injected here, but it is used in all the `ui-router` examples, so the comment is trying to explain why we're not using the exemplary way to get state parameters.

kifi-backend/modules/shoebox/angular/src/search/search.js
yipingliao, line 267: Will do. Thanks!

kifi-backend/modules/shoebox/angular/src/search/search.js
yipingliao, line 271: I forgot that `library` was injected, and was looking for it on the scope. :)  Thanks!

kifi-backend/modules/shoebox/angular/src/search/search.js
yipingliao, line 286: Yes, we do. I just tried it and actually had to put back `libraryUrl` instead of using the injected `library`. Both the logic and the `libraryUrl` are needed for the case where the user removes the library chip - in that case `libraryUrl` will be set to false by the header controller and we need the logic to account for this case.

2is10, line 286: I see. We might consider triggering a different event for the removal of the library chip so that each handler could be simpler.

PR #13679, opened 2015-02-11
Twitter->Library sync logic

stkem:
Still missing: Admin page and tests, but this should be workable to start with.
@andrewconner @leogrim


kifi-backend/modules/common/app/com/keepit/common/oauth/OAuth2Configuration.scala
```diff
@@ -9,11 +9,12 @@ case class OAuth1ProviderConfiguration(
...
   authorizationUrl: URL,
-  key: ConsumerKey)
+  key: ConsumerKey,
+  accessToken: ConsumerKey)
 
 object OAuth1ProviderConfiguration {
-  def build(name: String, requestTokenUrl: URL, accessTokenUrl: URL, authorizationUrl: URL) = (key: ConsumerKey) => {
-    OAuth1ProviderConfiguration(name, requestTokenUrl, accessTokenUrl, authorizationUrl, key)
+  def build(name: String, requestTokenUrl: URL, accessTokenUrl: URL, authorizationUrl: URL) = (key: ConsumerKey, accessToken: ConsumerKey) => {
+    OAuth1ProviderConfiguration(name, requestTokenUrl, accessTokenUrl, authorizationUrl, key, null)
```
andrewconner: Whoa. `null`

andrewconner:
Does this elegantly pick up after an archive import?

kifi-backend/modules/common/app/com/keepit/common/oauth/OAuth2Configuration.scala
stkem: Whoops, fixing

stkem:
No, it currently does not pick up elegantly. It's get the last (up to) 3200 tweets again. Smart version is phase 2, if this turns out to be used.

andrewconner:
Hm. Anything the archive importer can do to help? I have the full json... I could grab max ID or something?

stkem:
Yes, technically this isn't very hard, but there are some product questions that need to be figured out, e.g. should a twitter import automatically create a sync?
For now we'll only use this internally for the influencers libraries. Then we can figure out how to give this to regular users, at which point we should tie it in properly with the archive import.

PR #13680, opened 2015-02-11
throttling twitter file import

ymatsuda:
@eishay @andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepInterner.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepInterner.scala
@@ -173,7 +173,7 @@ class KeepInterner @Inject() (
       }
       if (uri.state == ACTIVE || uri.state == INACTIVE) {
         val date = source match {
-          case KeepSource.bookmarkImport => currentDateTime.plus(Random.nextInt(MAX_RANDOM_SCHEDULE_DELAY))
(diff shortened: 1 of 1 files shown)


eishay:
:+1:

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepInterner.scala
```diff
@@ -173,7 +173,7 @@ class KeepInterner @Inject() (
       }
       if (uri.state == ACTIVE || uri.state == INACTIVE) {
         val date = source match {
-          case KeepSource.bookmarkImport => currentDateTime.plus(Random.nextInt(MAX_RANDOM_SCHEDULE_DELAY))
+          case KeepSource.bookmarkImport | KeepSource.twitterFileImport => currentDateTime.plus(Random.nextInt(MAX_RANDOM_SCHEDULE_DELAY))
```
andrewconner, line 176: How about `case s if KeepSource.imports.contains(s)`? Covers more cases.

andrewconner:
Thanks!

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepInterner.scala
ymatsuda, line 176: Good idea.

PR #13687, opened 2015-02-11
image processing backfill for KeepImage

joshm1:
new method will check to see if any KeepImage versions are missing, create them, and wait for processing to complete

@andrewconner 
cc @stkem @eishay


andrewconner:
Looks good. Minor feedback about comments given in person.

PR #13688, opened 2015-02-11
replacing • (dot) with | (pipe) to be more like the other large sites (f...

eishay:
...b, twitter, linkedin)


eishay:
@andrewconner @2is10 should consider merging only after today's releases.

stkem:
merging?

eishay:
too old, will scrap and redo later

PR #13691, opened 2015-02-11
Update Angular tests for the new layout.

yipingliao:
@2is10


kifi-backend/modules/shoebox/angular/test/unit/friends/rightColFriends.spec.js
```diff
@@ -1,51 +0,0 @@
-'use strict';
```
yipingliao, line 1: We no longer have this.

kifi-backend/modules/shoebox/angular/test/e2e/index.spec.js
```diff
@@ -17,17 +17,17 @@ describe('kifi angular sanity suite', function () {
       expect(browser.getTitle()).toEqual('Kifi • Your Recommendation List');
     });
 
-    it('should display the friends module', function () {
-      expect(element(by.css('.kf-rightcol-friends')).isPresent()).toBe(true);
+    it('should display the goodies module', function () {
```
yipingliao, line 20: Instead of checking for friends, check for Goodies.

2is10:
:+1:

PR #13693, opened 2015-02-11
Fixing potential race condition in interning libs

andrewconner:
@ahsu1230 got a couple exceptions. Probably this was called in quick succession from an import or something.


ahsu1230:
huh I see. thanks

PR #13694, opened 2015-02-11
making header width adjust to content width, adding side padding

2is10:
@yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/route.js
+++ b/kifi-backend/modules/shoebox/angular/src/route.js
@@ -30,7 +30,10 @@ angular.module('kifi')
     $stateProvider
       .state('home', {  // Home page.
         url: '/',
-        templateUrl: 'recos/recosView.tpl.html'
+        templateUrl: 'recos/recosView.tpl.html',
+        data: {
+          width: 'kf-866-966'
+        }
       })
       .state('friends', {
         url: '/friends',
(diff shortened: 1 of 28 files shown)


kifi-backend/modules/shoebox/angular/src/route.js
```diff
@@ -30,7 +30,10 @@ angular.module('kifi')
     $stateProvider
       .state('home', {  // Home page.
         url: '/',
-        templateUrl: 'recos/recosView.tpl.html'
+        templateUrl: 'recos/recosView.tpl.html',
+        data: {
+          width: 'kf-866-966'
+        }
```
2is10, line 36: Here’s where we declare that a view/state has a non-standard width.

kifi-backend/modules/shoebox/angular/src/userProfile/userProfileHeader.tpl.html
```diff
@@ -47,17 +47,16 @@
...
-  </nav>
+    <nav class="kf-user-profile-user-nav">
+      <!-- For testing AngularUI routing, using <a> tags. However, will make into divs for V1 release. -->
+      <!-- Switch back to <a> tags in V2. (use: ui-sref="{{userNavLink.routeState}}") -->
+      <div class="kf-user-profile-user-nav-link" ng-repeat="userNavLink in userNavLinks">
+        <span class="kf-user-profile-user-nav-item-count">{{userNavLink.count|number}}</span>
+        <span class="kf-user-profile-user-nav-item-name">{{userNavLink.name}}</span>
+      </div>
+    </nav>
+  </div>
```
2is10, line 61: indentation only, moving within `kf-width-bounded` parent

2is10:
You’ll notice we have the `kf-width-bounded` class scattered throughout views. This gives them the flexibility to have parts that are not width bounded, like the library and profile headers.

PR #13698, opened 2015-02-11
logging for activity email

joshm1:
@stkem


stkem:
go ahead and merge away so you can test in prod. I'll review after the fact.

PR #13699, opened 2015-02-11
Wait until social networks are ready before populating PYMK.

yipingliao:
@2is10 
Previously, the PYMK button was populated once on module load (before the networks are loaded from the backend), and then once again when the networks are updated. Changing it to just populate once, after the networks have been loaded from the backend.

This bug manifested in the UI as a "Facebook flicker" - it actually flickers every time the first selected button is different from the second selected button.

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/friends/peopleYouMayKnow.js
+++ b/kifi-backend/modules/shoebox/angular/src/friends/peopleYouMayKnow.js
@@ -121,25 +121,20 @@ angular.module('kifi')
 
         function getEligibleNetworksCsv() {
           var onTwitterExperiment = _.indexOf(profileService.me.experiments, 'twitter_beta') > -1;
-          return _.compact([
+          var csv = _.compact([
             socialService.facebook && socialService.facebook.profileUrl ? null : 'fb',
             !onTwitterExperiment || (socialService.twitter && socialService.twitter.profileUrl) ? null : 'tw',
             socialService.linkedin && socialService.linkedin.profileUrl ? null : 'li',
             socialService.gmail && socialService.gmail.length ? null : 'gm'
           ]).join(',');
-        }
 
-        //
-        // Watches and listeners
-        //
-        scope.$watch(getEligibleNetworksCsv, function (csv) {
           scope.network = csv ? _.sample(csv.split(',')) : null;
-        });
+        }
 
         //
         // Initialization
         //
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/angular/src/friends/peopleYouMayKnow.js
```diff
@@ -121,25 +121,20 @@ angular.module('kifi')
 
         function getEligibleNetworksCsv() {
           var onTwitterExperiment = _.indexOf(profileService.me.experiments, 'twitter_beta') > -1;
-          return _.compact([
+          var csv = _.compact([
```
2is10, line 124: Hmm... looks like this function always returns `undefined` now.

2is10, line 124: Nevermind. I misread the diff. Despite the name, I now see this is no longer a getter, and is not supposed to return anything.

yipingliao, line 124: Yes. Its return value is no longer being used.

yipingliao, line 124: Maybe I should rename it to `initEligibleNetworks`.

kifi-backend/modules/shoebox/angular/src/friends/peopleYouMayKnow.js
```diff
@@ -121,25 +121,20 @@ angular.module('kifi')
...
           ]).join(',');
-        }
 
-        //
-        // Watches and listeners
-        //
-        scope.$watch(getEligibleNetworksCsv, function (csv) {
           scope.network = csv ? _.sample(csv.split(',')) : null;
-        });
+        }
```
2is10, line 132: I think the reason for the `$watch` was that if the user connects a network, the page should update. Do we always leave the page when connecting a network now?

An alternative fix would just be to move the `$watch` call inside of a `socialService.refresh().then` callback.

2is10:
:+1: Thanks for fixing.

kifi-backend/modules/shoebox/angular/src/friends/peopleYouMayKnow.js
yipingliao, line 132: You're right. Will fix.

PR #13707, opened 2015-02-11
urlencode all library paths

joshm1:
@andrewconner


2is10:
Why? Where are problems occurring, and for which characters? Thanks.

PR #13719, opened 2015-02-12
An idea about sessions?

andrewconner:
@stkem thoughts? Not quite as intrusive as only allowing `MaybeUserPage`s to get a kcid.


stkem:
As discussed in person, I think this is good.

PR #13720, opened 2015-02-12
incrementing user value cache version so users will see library menu callout

2is10:
@andrewconner


2is10:
I’m setting the pref for all existing users in the database now.

PR #13724, opened 2015-02-12
Constrain width of friend request banner

yipingliao:
@2is10 cc @ahsu1230


kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.styl
```diff
@@ -126,6 +126,8 @@
 
 .kf-user-profile-friend-request-header
   position: relative
+  border-bottom-right-radius: 10px
```
yipingliao, line 129: This is to be consistent with the library invite banner. My idea and forgot to check to product. We can take it away if people hate it. :)

2is10:
:+1:

PR #13734, opened 2015-02-12
Twitter Sync gets its own keep source

stkem:
@andrewconner


andrewconner:
Didn't realize the twitter sync stuff was in the controller. Should be factored out, otherwise, looks good.

PR #13735, opened 2015-02-12
System_Persona added

ahsu1230:
`System_Persona` is a new value for `LibraryKind` to denote automatically created libraries from selecting personas. In the BE, persona libraries are treated differently for recos? But, they are treated just like normal `User_Created` Libraries in the FE.

Because of this change, there are a couple of dependencies on the FE side which depend on libraries being `user_created`. I made some changes so that library functionalities don't depend on `user_created`, but rather depend on neither being `system_main` nor `system_secret`

@stkem @yingjieMiao @yipingliao @2is10


kifi-backend/modules/search/app/com/keepit/search/index/graph/library/LibraryIndexable.scala
```diff
@@ -136,7 +138,7 @@ class LibraryIndexable(library: Library, memberships: Seq[LibraryMembershipView]
 
     library.kind match {
       case LibraryKind.SYSTEM_MAIN | LibraryKind.SYSTEM_SECRET => // do not index the name of main/private libraries
-      case LibraryKind.USER_CREATED =>
+      case LibraryKind.USER_CREATED | LibraryKind.SYSTEM_PERSONA =>
```
ahsu1230: made this change in `search` because I was getting warnings about unexhaustive matching on `LibraryKind`.
I made `System_Persona` the same as `User_Created` for now. Is this ok? @leogrim @stkem

yingjieMiao:
thanks for the heads up. I might need to do something on cortex to treat these libraries differently

stkem:
lgtm, but I would like @leogrim to have a quick look at the search change, just to be sure

kifi-backend/modules/search/app/com/keepit/search/index/graph/library/LibraryIndexable.scala
leogrim: Are these libraries public? Will they all share a common set of names?

stkem: They are public, yes, and they will share a common set of names.

leogrim: So we index these names, we're going to see lots of irrelevant / identical library results. On the other end I guess a user must be able to find its own libraries. That last bit is actually true for main and private, which we currently exclude.  I think we should have a consistent approach, here's what I suggest:
- for now, treat it as a system library i.e. do _not_ index its name (keeping the search results clean)
- tomorrow, we can discuss how to bring these system libraries in search (my two cents: just by name, just for yourself)

ahsu1230: changed `system_persona` libraries to not index! Something to consider is what happens once a user changes the name/slug of a system persona name? After the name change, is it important that it's a persona library?

leogrim: No you're right, that's the trouble with these supplied libraries, we need some indicator that a library is "user-created" enough, even if it started as a system library or an import.

kifi-backend/modules/shoebox/angular/src/libraries/libraryService.js
```diff
@@ -143,7 +143,11 @@ angular.module('kifi')
       invitedSummaries: invitedSummaries,
       recentLibraries: recentLibraries,
 
-      isSystemLibrary: function (libraryId) {
+      isSystemLibrary: function (library) {
+        return library.kind === 'system_main' || library.kind === 'system_secret';
+      },
```
2is10, line 148: Having `isSystemLibrary` return `false` for `system_persona` is surprising. This method should probably be called something else, like `isSystemMainOrSecret`.

2is10:
What happens if the user un-checks the persona? Delete the library if it’s empty?

What happens if the user re-checks it? Are we storing which persona a `system_persona` library is for?

Does a `system_persona` library ever change to `user_created`?

I’d love to fully understand the lifecycle of a `system_persona` library.

ahsu1230:
right now, 
- if a user unchecks a persona at anytime, the library is not deleted (even if empty)
- we do have a mapper that maps `persona` -> auto created library name. i.e. `PARENT` -> library name: `I love my kids` (not actually)
- `system_persona` never changes kind

PR #13738, opened 2015-02-12
careful filtering of library kind

yingjieMiao:
@stkem @ahsu1230 

I want to be cautious about the new library kind. We can always let them pass the filter in the future.

### DIFF
--- a/kifi-backend/modules/cortex/app/com/keepit/cortex/models/lda/LDALibraryUpdater.scala
+++ b/kifi-backend/modules/cortex/app/com/keepit/cortex/models/lda/LDALibraryUpdater.scala
@@ -35,6 +35,8 @@ class LDALibraryUpdaterImpl @Inject() (
 
   type Auxiliary = (Option[LDATopic], Option[LDATopic], Option[LDATopic], Option[Float], Option[Float]) // (1st topic, 2nd topic, 3rd topic, 1st topic score, entropy)
 
+  private def acceptLibraryKind(x: LibraryKind): Boolean = (x == LibraryKind.USER_CREATED)
+
   def update(): Unit = {
     representer.versions.foreach { implicit version =>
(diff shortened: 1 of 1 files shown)


ahsu1230:
:+1:

kifi-backend/modules/cortex/app/com/keepit/cortex/models/lda/LDALibraryUpdater.scala
```diff
@@ -35,6 +35,8 @@ class LDALibraryUpdaterImpl @Inject() (
 
   type Auxiliary = (Option[LDATopic], Option[LDATopic], Option[LDATopic], Option[Float], Option[Float]) // (1st topic, 2nd topic, 3rd topic, 1st topic score, entropy)
 
+  private def acceptLibraryKind(x: LibraryKind): Boolean = (x == LibraryKind.USER_CREATED)
```
stkem, line 38: Oh, we should definitely accept the system_persona libraries. They should as much as possible be treated like user_created.

yingjieMiao:
I would like to be cautious at this moment. 

Once these things start to flood in, they will have impact on cortex services. For example, the library-library related graph will grow significantly because of these semi-auto-generated libraries. 

And people might start complaining 'their persona libraries' get recommended (I remember the debate on wether those libraries should be public or private) 

Tightening the filter right now doesn't make it difficult later to include these libraries. However, once these libraries get into library_lda_table, that table doesn't track libary_kind (cortex_library table does that). It's going to be harder to filter these things out.

I will keep the filter tight for now, until products people has thought through it. (After that, I may need to do multiple changes on cortex side)

stkem:
Ok, but we musn't forget about this. We will need to be able to recommend these persona libraries sooner or later and we definitely want to see related libraries _on_ the persona libraries.

PR #13747, opened 2015-02-12
avoiding unwanted propagation of scroll wheel events

2is10:
@yipingliao

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/app.styl
+++ b/kifi-backend/modules/shoebox/angular/src/app.styl
@@ -11,9 +11,6 @@ body
   background: appBackgroundColor
   font-family: Lato, Helvetica, Arial, sans-serif
 
-  &.modal-open
-    overflow: hidden
-
 *, :after, :before
   box-sizing: content-box
 
--- a/kifi-backend/modules/shoebox/angular/src/common/directives/antiscroll.js
+++ b/kifi-backend/modules/shoebox/angular/src/common/directives/antiscroll.js
@@ -2,36 +2,28 @@
 
 angular.module('antiscroll', ['kifi'])
 
+// The parent of 'antiscroll-inner' should have 'position: relative; overflow: hidden'.
+// The 'antiscroll-inner' element should have 'width: calc(100% + 30px)'.
+// Children of 'antiscroll-inner' should have 'width: calc(100% - 30px)'.
 .directive('antiscroll', [
-  '$timeout', 'scrollbar',
-  function ($timeout, scrollbar) {
+  '$timeout',
+  function ($timeout) {
     return {
       restrict: 'A',
       transclude: true,
-      link: function (scope, element, attrs /*, ctrl, transclude*/ ) {
-        var options;
-        if (attrs.antiscroll) {
-          options = scope.$eval(attrs.antiscroll);
-        }
-        scope.scroller = element.antiscroll(options).data('antiscroll');
+      link: function (scope, element, attrs) {
+        var options = attrs.antiscroll ? scope.$eval(attrs.antiscroll) : undefined;
+        var scroller = element.antiscroll(options).data('antiscroll');
+        var refresh = scroller.refresh.bind(scroller);
 
-        scope.refreshScroll = function () {
-          return $timeout(function () {
-            if (scope.scroller) {
-              scope.scroller.refresh();
-            }
-          });
-        };
+        $timeout(refresh);
 
-
-        scope.refreshScroll();
-
-        // http://stackoverflow.com/questions/986937/how-can-i-get-the-browsers-scrollbar-sizes
-        scope.width = 'calc(100% + ' + scrollbar.getAntiscrollWidth() + 'px)';
-
-        scope.$on('refreshScroll', scope.refreshScroll);
+        scope.$on('refreshScroll', function () {
+          $timeout(refresh);
+        });
       },
-      template: '<div class="antiscroll-inner" ng-attr-style="width: {{width}}" ng-transclude></div>'
+
+      template: '<div class="antiscroll-inner" ng-transclude></div>'
     };
   }
 ]);
--- a/kifi-backend/modules/shoebox/angular/src/common/directives/wheelStop.js
+++ b/kifi-backend/modules/shoebox/angular/src/common/directives/wheelStop.js
@@ -1,0 +1,48 @@
+'use strict';
+
+angular.module('kifi')
+
+// for elements on which the mouse wheel should have no effect
+.directive('kfWheelStop', function () {
+  return {
+    restrict: 'A',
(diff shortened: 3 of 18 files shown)


kifi-backend/modules/shoebox/angular/src/app.styl
```diff
@@ -11,9 +11,6 @@ body
   background: appBackgroundColor
   font-family: Lato, Helvetica, Arial, sans-serif
 
-  &.modal-open
-    overflow: hidden
-
```
2is10, line 16: `overflow: hidden` on the body can make the vertical scrollbar disappear, which makes the body wider and makes centered content jump right 10px or so, so eliminating this

kifi-backend/modules/shoebox/angular/src/common/directives/antiscroll.js
```diff
@@ -2,36 +2,28 @@
...
-            }
-          });
-        };
+        $timeout(refresh);
 
-
-        scope.refreshScroll();
-
-        // http://stackoverflow.com/questions/986937/how-can-i-get-the-browsers-scrollbar-sizes
-        scope.width = 'calc(100% + ' + scrollbar.getAntiscrollWidth() + 'px)';
```
2is10, line 30: `getAntiscrollWidth` was returning a constant, so I just moved this to a style rule

kifi-backend/modules/shoebox/angular/src/common/directives/antiscroll.js
```diff
@@ -2,36 +2,28 @@
...
+        scope.$on('refreshScroll', function () {
+          $timeout(refresh);
+        });
       },
-      template: '<div class="antiscroll-inner" ng-attr-style="width: {{width}}" ng-transclude></div>'
+
+      template: '<div class="antiscroll-inner" ng-transclude></div>'
     };
   }
 ]);
```
2is10, line 29: The only use of this directive was the library menu, but I had to stop using it there because I needed to add another directive (`kf-wheel-allow`) on its `antiscroll-inner` element, which this directive doesn’t seem to allow. Still, I’m keeping this directive around because it seems useful.

kifi-backend/modules/shoebox/angular/src/common/directives/wheelStop.js
```diff
@@ -0,0 +1,48 @@
...
+    restrict: 'A',
+    link: function (scope, element) {
+      element.on('wheel', function (event) {
+        var delta = event.originalEvent.deltaY;
+        if (delta) {
+          var scrollTop = this.scrollTop;
+          if (delta > 0) {
+            var scrollTopMax = this.scrollHeight - this.clientHeight;
+            if (scrollTop + delta > scrollTopMax) {
+              this.scrollTop = scrollTopMax;
```
2is10, line 31: Scrolling past the allowed limits can happen on Macs, but we prevent that here because it makes it harder to distinguish which wheel events may have the unwanted side effect of scrolling a container.

kifi-backend/modules/shoebox/angular/src/layout/libraryMenu/libraryMenu.js
```diff
@@ -54,7 +54,7 @@ angular.module('kifi')
         }
 
         function setMenuHeight() {
-          var menuHeight = getElement('.kf-nav-lib-users').outerHeight() + 230;
+          var menuHeight = getElement('.kf-nav-lib-users')[0].scrollHeight + 230;
```
2is10, line 57: This is because I made `.kf-nav-lib-users` the scrollable `.antiscroll-inner` element, to avoid adding a new element and having to indent a bunch of code in the template.

kifi-backend/modules/shoebox/angular/src/layout/libraryMenu/libraryMenu.js
```diff
@@ -63,6 +63,14 @@ angular.module('kifi')
...
           positionMenu();
           scope.changeList();
+          if (!scrollableLibList.data('antiscroll')) {
+            element.on('transitionend', function f(e) {
+              if (e.target === this) {
+                element.off('transitionend', f);
+                scrollableLibList.antiscroll({autoHide: false});
+              }
+            });
+          }
```
2is10, line 73: We only activate antiscroll after the menu is first shown, to avoid having it measure incorrectly and set `width: 0`.

kifi-backend/modules/shoebox/angular/src/layout/libraryMenu/libraryMenu.js
```diff
@@ -119,59 +127,31 @@ angular.module('kifi')
...
-            setPositioning(separators[1], 'fixed', stickToMaxTop - (offset - (secondLimitOverlay - separatorHeight)));
-            setPositioning(separators[2], 'relative', 0);
-
+            sep = separators.eq(1).css({position: 'absolute', top: 0 - (offset - (secondLimitOverlay - separatorHeight))});
           } else {
-            setPositioning(separators[0], 'relative', 0);
-            setPositioning(separators[1], 'relative', 0);
-            setPositioning(separators[2], 'fixed', stickToMaxTop);
-
+            sep = separators.eq(2).css({position: 'absolute', top: 0});
```
2is10, line 147: Using absolute positioning here instead of fixed makes it simpler.

kifi-backend/modules/shoebox/angular/src/layout/libraryMenu/libraryMenu.js
```diff
@@ -356,9 +320,12 @@ angular.module('kifi')
...
 
-          scope.$broadcast('refreshScroll');
-          $timeout(function() {
-            antiscrollLibList.scrollTop(0);
+          var libUsers = getElement('.kf-nav-lib-users').css({visibility: 'hidden'});
+          $timeout(function () {
+            setMenuHeight();
+            libUsers.css({visibility: 'visible'});
+            scrollableLibList.scrollTop(0);
+            (scrollableLibList.data('antiscroll') || {refresh: angular.noop}).refresh();
```
2is10, line 328: Combining what was 3 timeouts into one here.

kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.styl
```diff
@@ -167,10 +167,8 @@ textarea.manage-lib-input
 
 .manage-lib-visibility-description-content
   font-size: 15px
-  line-height = 28px
-  line-height: line-height
-  height: 2 * line-height
-  overflow: auto
+  line-height: 28px
+  height: 48px
```
2is10, line 171: I came across these old styles by searching for uses of `overflow: scroll` or `auto` and thought I’d simplify this while I was here.

yipingliao:
Wow, nice. :+1:

PR #13754, opened 2015-02-13
Use new recommendations endpoint

ahsu1230:
Using new recommendations endpoint.
keeps track of a `uriContext` & `libContext` for every call from 'See More'

@2is10 @yipingliao @yingjieMiao @stkem

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/common/services/routeService.js
+++ b/kifi-backend/modules/shoebox/angular/src/common/services/routeService.js
@@ -107,7 +107,14 @@ angular.module('kifi')
         return route('/recos/adHoc?n=' + howMany);
       },
       recos: function (opts) {
-        return route('/recos/top?more=' + opts.more + '&recency=' + opts.recency);
+        var routeBase = '/recos/topV2?more=' + opts.more + '&recency=' + opts.recency;
+        if (opts.uriContext) {
+          routeBase += '&uriContext=' + opts.uriContext;
+        }
+        if (opts.libContext) {
+          routeBase += '&libContext=' + opts.libContext;
+        }
+        return route(routeBase);
       },
       recosPublic: function () {
         return route('/recos/public');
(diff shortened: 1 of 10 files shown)


kifi-backend/modules/shoebox/angular/src/recos/recoActionService.js
```diff
@@ -15,15 +16,19 @@ angular.module('kifi')
...
 
       return $http.get(routeService.recos(recoOpts)).then(function (res) {
         if (res && res.data) {
           // Save recommendation data from the backend so we don't have to
           // fetch them again on Angular route reload of recommendations.
-          rawRecos = res.data;
-          return res.data;
+          rawRecos = res.data.recos;
+          uriContext = res.data.uctx;
+          libContext = res.data.lctx;
```
2is10, line 30: It’d be nice to use these JSON fields names `uctx` and `lctx` as the query parameter names too, for consistency, but not critical.

2is10:
:+1:

kifi-backend/modules/shoebox/angular/src/common/services/routeService.js
```diff
@@ -107,7 +107,14 @@ angular.module('kifi')
         return route('/recos/adHoc?n=' + howMany);
       },
       recos: function (opts) {
-        return route('/recos/top?more=' + opts.more + '&recency=' + opts.recency);
+        var routeBase = '/recos/topV2?more=' + opts.more + '&recency=' + opts.recency;
+        if (opts.uriContext) {
+          routeBase += '&uriContext=' + opts.uriContext;
+        }
+        if (opts.libContext) {
+          routeBase += '&libContext=' + opts.libContext;
```
2is10, line 115: What kinds of values are these contexts? If they can be anything but alphanumeric then we need to use `encodeURIComponent`.

ahsu1230, line 115: per discussion in person, we're using the same web-safe Base64 as search does. Sooooooooooo should be okay! @yingjieMiao

PR #13777, opened 2015-02-13
No tag for twitter stream imports

andrewconner:
@stkem


stkem:
:+1:

andrewconner:
@eishay Snuck into this PR, I'm adding a few `attempts = 3` for rw transactions that happen during registration. We _really_ shouldn't fail then, so being more safe. Looks like there's occasional deadlocks — any ideas why?

PR #13782, opened 2015-02-13
fix library ordering from profiles endpoint

ahsu1230:
With the introduction of `LibraryKind.System_Persona`, those persona libraries show up before `System_Secret`because of alphabetical ordering lol.
Here's a fix that treats `User_Created` & `System_Persona` the same priority when it comes to `LibraryKind` ordering

@2is10 @andrewconner @stkem


2is10:
:+1:

PR #13784, opened 2015-02-13
using Angular's query string formatter for our ajax URLs

2is10:
@ahsu1230

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/common/util.js
+++ b/kifi-backend/modules/shoebox/angular/src/common/util.js
@@ -3,8 +3,8 @@
 angular.module('util', [])
 
 .factory('util', [
-  '$document', '$window',
-  function ($document, $window) {
+  '$document', '$window', '$location',
+  function ($document, $window, $location) {
     var HTML_ESCAPE_CHARS = /[&<>"'\/]/g;
     var HTML_ESCAPES = {
       '&': '&amp;',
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/shoebox/angular/src/common/util.js
```diff
@@ -111,6 +113,15 @@ angular.module('util', [])
...
         return text == null ? '' : String(text).replace(HTML_ESCAPE_CHARS, htmlEscapeReplace);
       },
+      /**
+       * Behaves like $location.search({...}). Keys whose values are an empty array will be omitted entirely.
+       * https://docs.angularjs.org/api/ng/service/$location#search
+       */
+      formatQueryString: function (params) {
+        queryStringHelper.$$search = params;
+        queryStringHelper.$$compose();
+        return queryStringHelper.$$url;
```
2is10, line 123: This is a little funky. Angular doesn’t expose the function we actually want (called `toKeyValue` in their codebase), so we use `$location.$$compose()`, which calls it. We have a unit test that will fail if the underlying code ever changes in a breaking way (e.g. during a Angular 1.3 or 1.4 upgrade).

ahsu1230:
ooo good thinking. thanks!
Something I thought could be to put the `queryStr()` inside `route()` and pass a parameter to `route()`. That way, you don't have to keep calling `queryStr()`. 
So something like, 

```
def route(url, params) {
  return env.xhrBase + url + queryStr(params);
}

route('/collections/search', {query: query, limit: limit});
```

Would that be better?

2is10:
Great idea.

PR #13785, opened 2015-02-13
Make auto follow button cta go away on scroll and come back on scroll stop.

yipingliao:
@2is10 Thanks for your pointers on this!


kifi-backend/modules/shoebox/angular/src/libraries/libraryHeaderFollowButtonCta.js
```diff
@@ -123,14 +125,25 @@ angular.module('kifi')
...
-        var adjustAutoCTAPosition = function () {
+        var lastScrolled = null;
+        var hideAutoCTA = function () {
+          lastScrolled = Date.now();
+
           if (autoCTAShowing) {
-            positionCTA();
+            hideCTA();
+
+            var showCTAOnScrollStop = $interval(function () {
```
yipingliao, line 135: Using an `$interval` because `scope.$watch` doesn't work - the watches are not evaluated when there is nothing happening on the page, so I had the situation where after I stopped scrolling, I have to click somewhere on the page to kick the watches and then the CTA will re-popup.

2is10:
I’m going to take a closer look at this before deciding what we should do. Closing for now. Thanks Yiping!

PR #13789, opened 2015-02-14
minor route maintenance

2is10:
@ahsu1230

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/common/services/routeService.js
+++ b/kifi-backend/modules/shoebox/angular/src/common/services/routeService.js
@@ -9,8 +9,12 @@ angular.module('kifi')
       return env.xhrBase + path + (params ? queryStr(params) : '');
     }
 
-    function searchRoute(url) {
-      return env.xhrBaseSearch + url;
+    function searchRoute(path) {
+      return env.xhrBaseSearch + path;
+    }
+
+    function navRoute(path, params) {
+      return env.navBase + path + (params ? queryStr(params) : '');
     }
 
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/angular/src/common/services/routeService.js
```diff
@@ -33,12 +37,12 @@ angular.module('kifi')
...
       prefs: route('/user/prefs'),
       importGmail: function (returnPath) {
-        return env.navBase + '/contacts/import' + queryStr({redirectUrl: returnPath || []});
+        return navRoute('/contacts/import', {redirectUrl: returnPath || []});
       },
       networks: route('/user/networks'),
       profileUrl: route('/user/me'),
       profileSettings: route('/user/settings'),
-      logoutUrl: env.navBase + '/logout',
+      logoutUrl: '/logout',
```
2is10, line 45: This only used in the logout link in the header. It doesn’t need to be fully qualified.

kifi-backend/modules/shoebox/angular/src/common/services/routeService.js
```diff
@@ -154,7 +159,7 @@ angular.module('kifi')
       // User registration      //
       ////////////////////////////
       socialSignup: function (provider) {
-        return env.navBase + '/auth/token-signup/' + provider;
+        return route('/auth/token-signup/' + provider);
```
2is10, line 167: This is an ajax call, so should be using our typical ajax subdomain.

PR #13790, opened 2015-02-14
logged start and end of url scrape to stdout (for immediate flushing to ...

eishay:
...file)

we would like to know what was the last url that the scraper attempted to scrape at the time it crashed.
the suspect code is in apache org.apache.http.impl.cookie.BestMatchSpec.formatCookies

Both the jvm crash log and an exception we had in the past are pointing to this method as problematic.


eishay:
@leogrim @andrewconner @stkem

PR #13791, opened 2015-02-14
fix recos See More

ahsu1230:
@2is10 the recommendations `See More` endpoint was breaking from the URI encoding.
The + and / need to stay unless we use a decoder in the backend. @yingjieMiao 

Here's a temporary fix for now.


yingjieMiao:
thanks.

2is10:
Really? Play should be decoding automatically, before giving values to our code. That’s what it does for our search endpoints, where we use `encodeURIComponent` on the context parameter before sending them.

Something’s fishy. I’ll look into it more when I have time. Maybe Angular’s query parameter encoding differs from `encodeURIComponent` in a breaking way for us.

yingjieMiao:
As far as I can tell, production is fine right now.

PR #13792, opened 2015-02-15
wait timeouts is now proportional to the load factor of the current serv...

eishay:
...ice and putting curator on load factor 5 so we'll get less long wait reports


eishay:
@stkem

PR #13800, opened 2015-02-17
deal with race condition on experiment update

eishay:


eishay:
@andrewconner

kifi-backend/modules/shoebox/app/com/keepit/commanders/LocalUserExperimentCommander.scala
```diff
@@ -43,8 +34,15 @@ class LocalUserExperimentCommander @Inject() (
     addDynamicExperiments(userId, staticExperiments)
   }
 
-  def addExperimentForUser(userId: Id[User], experiment: ExperimentType) = {
-    db.readWrite { implicit session => userExperimentRepo.save(UserExperiment(userId = userId, experimentType = experiment)) }
+  def addExperimentForUser(userId: Id[User], experiment: ExperimentType): UserExperiment = {
```
andrewconner, line 37: Great, thanks

PR #13803, opened 2015-02-17
Fixing logs for bm imports

andrewconner:
@eishay


andrewconner:
@stkem Current build has the bug, so silencing it. Left a comment to switch back to error later.

PR #13804, opened 2015-02-17
suggested searches

yingjieMiao:
@stkem @eishay @ahsu1230 
- basic model, repo, evolution
- admin endpoint to show and edit

Will hook up with production LibraryController later. That' minimal work.


yingjieMiao:
the admin js is probably not the best. will continue to play with it a bit.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibrarySuggestedSearchCommander.scala
```diff
@@ -0,0 +1,49 @@
...
+@ImplementedBy(classOf[LibrarySuggestedSearchCommanderImpl])
+trait LibrarySuggestedSearchCommander {
+  def getSuggestedTermsForLibrary(libId: Id[Library], limit: Int): SuggestedSearchTerms
+  def saveSuggestedSearchTermsForLibrary(libId: Id[Library], terms: SuggestedSearchTerms): Unit
+}
+
+@Singleton
+class LibrarySuggestedSearchCommanderImpl @Inject() (
+    val db: Database,
+    val termRepo: LibrarySuggestedSearchRepo) extends LibrarySuggestedSearchCommander {
```
ahsu1230: could the name of the repo match the class name a bit more? Perhaps like `librarySearchRepo`? Just to avoid confusion in the future if we happen to create a Repo with `term` in it.

yingjieMiao: sure.

stkem:
This is going to need a cache, since it will be very hot (on influencers library pages). The cache could cache the whole batch in the right size. lgtm otherwise

stkem:
Cache?

yingjieMiao:
will do that next PR.

PR #13806, opened 2015-02-17
sending ?more=1 instead of just ?more when more is true

2is10:
It turns out the problem #13791 fixed wasn’t with the encoding of the contexts at all. It was a disagreement between Angular and Play about how query parameters with value `true` should be represented.
@ahsu1230 
cc @yingjieMiao


2is10:
Sorry, I knew about this issue, but I didn’t realize the parameter `more` had a boolean value (just from looking at routeService.js). I should have been safer and looked deeper to be sure.

ahsu1230:
oh I see, thanks!

PR #13807, opened 2015-02-17
removing redundant parameter more from latest recos endpoints

2is10:
Also removing recos v1 website endpoint and making `more` optional in old mobile endpoint.
@ahsu1230 
cc @yingjieMiao

### DIFF
--- a/kifi-backend/modules/shoebox/conf/shoeboxService.routes
+++ b/kifi-backend/modules/shoebox/conf/shoeboxService.routes
@@ -236,8 +236,8 @@ POST    /m/1/attribution/appsflyer              @com.keepit.controllers.mobile.M
 
 POST    /m/1/disconnect/:provider               @com.keepit.controllers.mobile.MobileUserController.disconnect(provider)
 
-GET     /m/2/recos/top                          @com.keepit.controllers.mobile.MobileRecommendationsController.topRecosV2(more: Boolean, recency: Float)
-GET     /m/2/recos/topV3                        @com.keepit.controllers.mobile.MobileRecommendationsController.topRecosV3(more: Boolean, recency: Float, uriContext: Option[String] ?= None, libContext: Option[String] ?= None)
+GET     /m/2/recos/top                          @com.keepit.controllers.mobile.MobileRecommendationsController.topRecosV2(recency: Float, more: Boolean ?= false)
+GET     /m/3/recos/top                          @com.keepit.controllers.mobile.MobileRecommendationsController.topRecosV3(recency: Float, uriContext: Option[String] ?= None, libContext: Option[String] ?= None)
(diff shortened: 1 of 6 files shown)


kifi-backend/modules/shoebox/test/com/keepit/controllers/mobile/MobileRecommendationsControllerTest.scala
```diff
@@ -186,12 +186,12 @@ class MobileRecommendationsControllerTest extends TestKitSupport with Specificat
...
 
-        val route = com.keepit.controllers.mobile.routes.MobileRecommendationsController.topPublicRecos().url
+        val route = routes.MobileRecommendationsController.topPublicRecos().url
         route === "/m/1/recos/public"
 
-        val request = FakeRequest("GET", route)
+        val request = FakeRequest("GET", route).withHeaders(USER_AGENT -> "iKeefee/0.0.0 (Device-Type: NA, OS: iOS NA)")
 
-        val controller = inject[RecommendationsController]
+        val controller = inject[MobileRecommendationsController]
```
2is10, line 194: This was testing the wrong controller.

kifi-backend/modules/shoebox/conf/shoeboxService.routes
```diff
@@ -236,8 +236,8 @@ POST    /m/1/attribution/appsflyer              @com.keepit.controllers.mobile.M
 
 POST    /m/1/disconnect/:provider               @com.keepit.controllers.mobile.MobileUserController.disconnect(provider)
 
-GET     /m/2/recos/top                          @com.keepit.controllers.mobile.MobileRecommendationsController.topRecosV2(more: Boolean, recency: Float)
-GET     /m/2/recos/topV3                        @com.keepit.controllers.mobile.MobileRecommendationsController.topRecosV3(more: Boolean, recency: Float, uriContext: Option[String] ?= None, libContext: Option[String] ?= None)
+GET     /m/2/recos/top                          @com.keepit.controllers.mobile.MobileRecommendationsController.topRecosV2(recency: Float, more: Boolean ?= false)
+GET     /m/3/recos/top                          @com.keepit.controllers.mobile.MobileRecommendationsController.topRecosV3(recency: Float, uriContext: Option[String] ?= None, libContext: Option[String] ?= None)
```
2is10, line 240: I checked with both mobile teams. They aren’t using `topV3` yet. I altered the path to follow their versioning convention.

ahsu1230:
:+1:  lgtm

yingjieMiao:
Ah, good catch. Yes, mobile not using the endpoint yet. 

@ahsu1230  The site is OK with the new endpoint, right?

ahsu1230:
@yingjieMiao yup, it should be.

PR #13811, opened 2015-02-17
admin endpoint to get uri ids in a library

yingjieMiao:
@stkem FYI

I guess this is easier, and safer for me to run script


stkem:
Good idea :-)

PR #13813, opened 2015-02-18
positioning follow button using CSS only

2is10:
Fixes #13779, which was doing too much DOM reading+writing in a scroll handler and failed to deregister the scroll handler when the user navigated to a different page.

This is an alternative to the #13785 approach, which hid and then re-showed the promotional callout in a different position if it got to where it didn’t fit. Instead, we now anticipate situations where we wouldn’t be able to confidently show the callout very long (due to active scrolling) and slightly delay showing it in those cases.

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/library.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/library.js
@@ -7,38 +7,6 @@ angular.module('kifi')
   'keepDecoratorService', 'libraryService', 'modalService', 'platformService', 'profileService', 'originTrackingService', 'installService',
   function ($scope, $rootScope, $analytics, $location, $state, $stateParams, $timeout, $window, util, initParams, library,
     keepDecoratorService, libraryService, modalService, platformService, profileService, originTrackingService, installService) {
-    //
-    // A/B Tests.
-    //
-    var abTest = {
-      name: 'exp_follow_popup',
-      salt: 'hgg1dv',
-      treatments: [
-        {
-          name: 'none',
-          isControl: true
-        },
-        {
-          name: 'popupLibrary',
-          data: {
-            buttonText: 'Follow',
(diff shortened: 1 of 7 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/library.js
```diff
@@ -7,38 +7,6 @@ angular.module('kifi')
...
-          data: {
-            buttonText: 'Save',
-            mainText: 'Join Kifi to save this collection.<br/>Discover other collections,<br/>and build your own!',
-            quote: 'From business to personal, Kifi has been<br/>instrumental in my day-to-day life.',
-            quoteAttribution: 'Remy Weinstein, California'
-          }
-        }
-      ]
-    };
-
```
2is10, line 41: just moved this down to where it’s used, below

kifi-backend/modules/shoebox/angular/src/libraries/libraryHeader.js
```diff
@@ -35,7 +35,6 @@ angular.module('kifi')
         //
         // Scope data.
         //
-        scope.isUserLoggedOut = $rootScope.userLoggedIn === false;
```
2is10, line 38: better to use the same scope variable consistently throughout the app

kifi-backend/modules/shoebox/angular/src/libraries/libraryHeader.js
```diff
@@ -133,10 +132,6 @@ angular.module('kifi')
             scope.coverImageUrl = env.picBase + '/' + image.path;
             scope.coverImagePos = formatCoverImagePos(image);
           }
-
-          scope.library.followButtonText = (scope.isUserLoggedOut && scope.library.abTestTreatment && !scope.library.abTestTreatment.isControl) ?
-            scope.library.abTestTreatment.data.buttonText :
-            'Follow Library';
```
2is10, line 139: inlined into template

PR #13816, opened 2015-02-18
cannot curl controller guarded by AdminUserAction

yingjieMiao:
@stkem FYI

security team did a good job.


stkem:
:-)

2is10:
FYI, you can curl an admin endpoint if you “copy as cURL” from your Network tab in Chrome dev tools, paste command into your terminal, them modify the path and anything else you need to before pressing Enter.

yingjieMiao:
thanks @2is10 , forgot this trick.

PR #13818, opened 2015-02-18
Who Are You for website

ahsu1230:
still looking over some stuff & Jo is still testing. Thought I'd get the code out first for review.
[Do Not Merge Yet] since we're waiting on extension & recos testing @yingjieMiao 

@2is10 @andrewconner @stkem


stkem:
This will not show anything to regular users yet, right?

ahsu1230:
Jo has been testing on my dev branch so far =)

kifi-backend/modules/shoebox/angular/src/common/services/routeService.js
```diff
@@ -5,6 +5,9 @@ angular.module('kifi')
 .factory('routeService', [
   'env', 'util',
   function (env, util) {
+    var cdnImageBase = 'https://djty7jcqog9qu.cloudfront.net/'; // for dynamic images (keeps, library images, user profile pics, etc.)
+    var cdnAssetBase = 'https://d1dwdv9wd966qu.cloudfront.net/'; // for static assets (icons)
```
ahsu1230: the cdnAssetBase is for persona icons

kifi-backend/modules/shoebox/angular/src/common/util.js
```diff
@@ -29,6 +29,14 @@ angular.module('util', [])
       endsWith: function (str, suffix) {
         return str === suffix || str.indexOf(suffix, str.length - suffix.length) !== -1;
       },
+      upperCaseFirstLetterOfWords: function (strings) {
```
ahsu1230: this function is used for Tracking Action Names
like (selectedPersonaTechie, unselectedPersonaDeepThinker, etc.)

kifi-backend/modules/shoebox/angular/src/layout/main/MainCtrl.js
```diff
@@ -256,12 +256,27 @@ angular.module('kifi')
...
+    }
+
     $rootScope.$watch(function () {
-      return Boolean(installService.installedVersion && profileService.prefs.auto_show_guide);
+      return profileService.prefs.auto_show_guide;
     }, function (show) {
       if (show) {
-        extensionLiaison.triggerGuide();
-        profileService.savePrefs({auto_show_guide: null});
+        showPersonaModal();
```
ahsu1230: first check if `auto_show_guide` is true, and if it is, show WAY modal.
THEN check if extension is installed to start the guide.

kifi-backend/modules/shoebox/angular/src/recos/recoActionService.js
```diff
@@ -101,4 +99,4 @@ angular.module('kifi')
 
     return api;
   }
-]);
+]);
```
ahsu1230: missed new line, adding...

kifi-backend/modules/shoebox/angular/src/signup/signupService.js
```diff
@@ -78,14 +78,6 @@ angular.module('kifi')
       trackEvent('visitor_clicked_page', 'install', 'getFirefox');
     };
 
-    function createSignupPath(network) {
```
ahsu1230: remove redirect path for signup from public library or profile

kifi-backend/modules/shoebox/angular/src/recos/recos.js
```diff
@@ -125,18 +137,11 @@ angular.module('kifi')
...
-        return reco && reco.recoKeep && reco.recoKeep.isMyBookmark;
-      });
-      $scope.recosState = 'hasRecos';
-    } else {
+    function reloadRecos(invalidate) {
       $scope.loading = true;
 
       recoStateService.empty();
-      recoActionService.get().then(function (rawRecos) {
+      recoActionService.get(invalidate).then(function (rawRecos) {
```
ahsu1230: On page refresh, or Update Interests, we reload the recommendations list.
Otherwise, (if for example, coming from library page, do not show new recos)

kifi-backend/modules/shoebox/angular/src/common/services/routeService.js
2is10: If these are static, why not put them in the `img` directory like our other icons? If they’re not static, why not use the same CDN we use for other “dynamic” images?

I ask because I’m afraid this little change is going to confuse engineers and waste our time going forward.

2is10: Also, the CDN base is already defined as `env.picBase`, since it’s environment dependent. We have a different one for local testing. I’m not sure we should also be defining the CDN base here.

kifi-backend/modules/shoebox/angular/src/common/services/routeService.js
```diff
@@ -158,8 +163,11 @@ angular.module('kifi')
...
-      socialSignup: function (provider) {
-        return route('/auth/token-signup/' + provider);
+      socialSignup: function(provider) {
+        return env.navBase + '/signup/' + provider;
+      },
+      tokenSocialSignup: function (provider) {
+        return env.navBase + '/auth/token-signup/' + provider;
       },
       socialSignupWithRedirect: function (provider, redirectPath, intent) {
         return '/signup/' + provider + queryStr({
```
2is10: Looks like `socialSignupWithRedirect` is no longer used. Remove?

kifi-backend/modules/shoebox/angular/src/common/services/routeService.js
```diff
@@ -158,8 +163,11 @@ angular.module('kifi')
       ////////////////////////////
       // User registration      //
       ////////////////////////////
-      socialSignup: function (provider) {
-        return route('/auth/token-signup/' + provider);
+      socialSignup: function(provider) {
+        return env.navBase + '/signup/' + provider;
+      },
+      tokenSocialSignup: function (provider) {
+        return env.navBase + '/auth/token-signup/' + provider;
```
2is10: It appears that neither of these URLs needs to be absolute. Let’s remove the `env.navBase +`. That’s only needed when we’re passing a fully-qualified return URL to a third-party service.

kifi-backend/modules/shoebox/angular/src/common/services/routeService.js
```diff
@@ -245,6 +253,19 @@ angular.module('kifi')
...
+      // User Personas           //
+      /////////////////////////////
+      getAllPersonas: function() {
+        return route('/user/personas');
+      },
+      addPersona: function(personaName) {
+        return route('/user/personas/' + personaName);
+      },
+      removePersona: function(personaName) {
+        return route('/user/personas/' + personaName);
```
2is10, line 269: for anonymous (nameless) functions, our convention to insert a space between the `function` keyword and the argument list

kifi-backend/modules/shoebox/angular/src/persona/managePersona.js
```diff
@@ -0,0 +1,80 @@
...
+        //
+        scope.selectPersona = function(persona) {
+          persona.selected = true;
+          persona.currentIconSrc = persona.activeIconSrc;
+          scope.selectedPersonaIds.push(persona.id);
+          userPersonaActionService.addPersona(persona.id);
+
+          // to Camel Case Analytics
+          var personaAction = util.upperCaseFirstLetterOfWords(persona.id.split('_')).join('');
+          $analytics.eventTrack('user_clicked_page', {type: persona.id, action: 'selectedPersona' + personaAction});
```
2is10: Let’s use `_.camelCase` and kill `util.upperCaseFirstLetterOfWords`.

``` js
$analytics.eventTrack('user_clicked_page', {
  type: persona.id,
  action: _.camelCase('selectedPersona_' + persona.id)
});
```

2is10: Are you sure `type` is supposed to be `persona.id`? Normally for a `user_clicked_page` event the `type` identifies what page was clicked, which wouldn’t depend on which persona was clicked.

kifi-backend/modules/shoebox/angular/src/common/modal/modal.js
```diff
@@ -29,7 +30,10 @@ angular.module('kifi')
           $document.off('keydown', onDocKeyDown);
         }
 
-        $scope.close = this.close = function (closeAction) {
+        $scope.close = this.close = function (closeAction, forceEscape) {
+          if (_.isBoolean($scope.escapable) && !$scope.escapable && !forceEscape) {
+            return;
+          }
```
2is10: I’m opposed to the idea of unescapable modals, but I understand that you’re probably just doing what product asked. I’d like to understand the flow a little better. When we spoke with Jen at lunch one day, I seem to recall settling on users seeing persona selection in the recommendations page non-modally, and only in a popup if they opened it from a link somewhere, like on the settings page.

kifi-backend/modules/shoebox/angular/src/common/modal/modal.js
```diff
@@ -56,6 +60,7 @@ angular.module('kifi')
 
         scope.backdropStyle.opacity = attrs.kfOpacity || 0.3;
         scope.backdropStyle.backgroundColor = attrs.kfBackdropColor || 'rgba(0, 40, 90, 1)';
+        scope.backdropStyle.top = attrs.kfTopOffset || '0';
```
2is10: Why does this need to be configurable? It looks like, for the persona modal, you’re just overriding the default of `'0'` with `'0px'`.

kifi-backend/modules/shoebox/angular/src/layout/main/MainCtrl.js
```diff
@@ -256,12 +256,27 @@ angular.module('kifi')
...
 
+    function showPersonaModal() {
+      modalService.open({
+        template: 'persona/managePersonaUnescapableModal.tpl.html',
+        modalData: {
+          onClose: function() {
+            if (installService.installedVersion) {
+              extensionLiaison.triggerGuide();
+            }
+            profileService.savePrefs({auto_show_guide: null});
```
2is10: It looks like this new code changes `auto_show_guide` to mean “trigger unescapeable persona selection.” After persona selection, the pref is cleared even if the extension is not installed and the guide was not actually started.

Did you consider letting `auto_show_guide` retain its original meaning and instead introducing a new mechanism to trigger persona selection? That way, even if the user doesn’t have the extension installed the first time they log in to the site and choose a persona, when they do install it, the extension’s guide will still auto-start.

If what I’ve written here is unclear, I’d be happy to discuss it in person.

kifi-backend/modules/shoebox/angular/src/persona/managePersona.js
```diff
@@ -0,0 +1,80 @@
...
+    return {
+      restrict: 'A',
+      require: '^kfModal',
+      templateUrl: 'persona/managePersona.tpl.html',
+      link: function (scope, element, attrs, kfModalCtrl) {
+        //
+        // Scope data.
+        //
+        scope.selectedPersonaIds = [];
+        scope.finishText = scope.modalData && scope.modalData.finishText ? scope.modalData.finishText : 'Update interests';
```
2is10: This can be written a little more concisely here like this:

``` js
scope.finishText = scope.modalData && scope.modalData.finishText || 'Update interests';
```

and even more concisely if inlined into the template:

``` html
{{scope.modalData.finishText || 'Update interests'}}
```

kifi-backend/modules/shoebox/angular/src/persona/managePersona.js
```diff
@@ -0,0 +1,80 @@
...
+            scope.selectPersona(persona);
+          }
+        };
+
+        scope.close = function() {
+          if (scope.selectedPersonaIds.length > 0) {
+            var closeFunc = scope.modalData ? scope.modalData.onClose : undefined;
+            kfModalCtrl.close(closeFunc, true);
+          }
+          $analytics.eventTrack('user_clicked_page', {action: 'closed'});
```
2is10: Seems to me that `type` should be supplied here too, and it should be consistent with the `type` used for the `user_clicked_page` events above.

kifi-backend/modules/shoebox/angular/src/persona/managePersona.tpl.html
```diff
@@ -0,0 +1,21 @@
+<div class="manage-persona dialog-content" ng-if="allPersonas.length > 0">
+  <div class="dialog-body-wrap">
+    <div class="dialog-body">
+      <h1 class="dialog-header"> Who are you? </h1>
+
+      <div class="persona-table">
+        <div ng-repeat="persona in allPersonas" class="persona-icon" ng-class="{active: persona.selected}" ng-click="toggleSelect(persona)">
+          <div class="persona-icon-image-wrapper">
+            <img class="persona-icon-image" src="{{persona.currentIconSrc}}">
```
2is10: This approach is going to cause the icon to temporarily disappear when clicked, until a server roundtrip for the active icon completes. I’d recommend spriting these—if not the whole set, then at least the various states of each icon. If we stick with this individual icon files approach, we’ll need to preload them all.

kifi-backend/modules/shoebox/angular/src/persona/managePersona.js
```diff
@@ -0,0 +1,80 @@
...
+        var cdnBase = routeService.cdnAssetBase;
+        userPersonaActionService.getPersonas().then(function (data) {
+          _.map(data.personas, function(p) {
+            p.displayName = p.displayName.toUpperCase();
+            p.iconSrc = cdnBase + p.iconPath;
+            p.activeIconSrc = cdnBase + p.activeIconPath;
+            p.currentIconSrc = p.selected ? p.activeIconSrc : p.iconSrc;
+            if (p.selected) {
+              scope.selectedPersonaIds.push(p.id);
+            }
```
2is10, line 94: Do we really need three representations of whether a persona is selected or not? Would be nice to reduce to one or two.

kifi-backend/modules/shoebox/angular/src/persona/managePersona.js
```diff
@@ -0,0 +1,80 @@
...
+        };
+
+        //
+        // On link.
+        //
+        // call for List of Personas
+        var cdnBase = routeService.cdnAssetBase;
+        userPersonaActionService.getPersonas().then(function (data) {
+          _.map(data.personas, function(p) {
+            p.displayName = p.displayName.toUpperCase();
```
2is10: This can be done in CSS with `text-transform: uppercase`.

kifi-backend/modules/shoebox/angular/src/persona/managePersona.tpl.html
2is10: For all new templates, we’re following the convention of using a short `kf-xxx-` prefix on all unique class names in the template and just `kf-` for state indicators like `kf-active`. So for this template, managePersona.tpl.html, we might choose the prefix `kf-mp-`. The top-level div would have `class="kf-mp dialog-content"` and the class names you’ve started with `persona-*` or `manage-persona-*` would switch to `kf-mp-*`.

kifi-backend/modules/shoebox/angular/src/profile/profile.tpl.html
```diff
@@ -30,6 +30,10 @@
         <div kf-profile-manage-settings></div>
         <div kf-profile-change-password></div>
         <div kf-profile-email-addresses input-state="addEmailInput" email-list="me.emails" validate-email-action="validateEmail(value)" add-email-action="addEmail(value)" resend-verification-email-action="resendVerificationEmail(value)"></div>
+        <div class="kf-profile-update-personas" ng-click="showPersonaModal()">
+          <div class="profile-setting-arrow"></div>
+          <div class="profile-manage-settings-title">Update your interests</div>
+        </div>
```
2is10: Please use a `<button>` element for this. Thanks.

If you want it to look like a link, you can use:

``` less
.kf-profile-update-personas
  border: none
  padding: 0
  background: transparent
  ...
```

kifi-backend/modules/shoebox/angular/src/recos/recos.js
```diff
@@ -160,6 +165,17 @@ angular.module('kifi')
       });
     }
 
+    // Load a new set of recommendations only on page refresh.
+    // Otherwise, load the recommendations we have previously shown.
+    if ($scope.recos.length > 0) {
+      _.remove($scope.recos, function (reco) {
+        return reco && reco.recoKeep && reco.recoKeep.isMyBookmark;
+      });
```
2is10, line 175: Shouldn’t this `_.remove` be done before the `.length` check?
Also, `reco` couldn’t possibly be falsy, could it? It’s got to be an object of some kind, no?

kifi-backend/modules/shoebox/angular/src/recos/recosView.tpl.html
```diff
@@ -2,9 +2,9 @@
   <div kf-delighted-survey hide="hideDelightedSurvey" ng-if="showDelightedSurvey"></div>
 
   <h1 class="kf-recos-title kf-main-title">Recommendations</h1>
-
   <div class="kf-reco-cols">
     <div class="kf-reco-col-1">
+        <div class="kf-recos-update-personas" ng-click="showPersonaModal()">Update your interests</div>
```
2is10: a `<button>` here too, and it appears that the indentation is off

2is10:
Nice work. I’m excited to try this out. I plan to update the extension tomorrow morning. Hope we can launch this tomorrow.

kifi-backend/modules/shoebox/angular/src/persona/managePersona.js
ahsu1230: This is Jen wanted the naming to be like
selectedPersonaTechie or selectedPersonaDeepThinker (for example)

kifi-backend/modules/shoebox/angular/src/common/modal/modal.js
ahsu1230: woops kept this in, just in case product wanted the top offset to be present (cover the header bar or not)

kifi-backend/modules/shoebox/angular/src/persona/managePersona.js
ahsu1230: hmm turns out `_.camelCase` is undefined. I think it's not defined for lodash build 2.4.1

ahsu1230: ah can use jQuery's `$.camelCase()`

ahsu1230: nvm, that's bad too

kifi-backend/modules/shoebox/angular/src/recos/recos.js
ahsu1230, line 175: Ah yes, your'e right for both things. `reco` is an object and you should check the length of the `$scope.recos` after removing things

kifi-backend/modules/shoebox/angular/src/layout/main/MainCtrl.js
```diff
@@ -256,13 +256,37 @@ angular.module('kifi')
...
+    }
+
+    $rootScope.$watchCollection(function () {
+      return [profileService.prefs.auto_show_guide, profileService.prefs.auto_show_persona];
+    }, function () {
+      if (profileService.prefs.auto_show_persona) {
+        showPersonaModal();
+      } else if (profileService.prefs.auto_show_guide) {
+        startGuide();
+      }
```
ahsu1230, line 267: @2is10 From our discussion before, here's a $watch on `auto_show_guide` and `auto_show_persona`
If `auto_show_persona` is set, it takes precedent and opens the persona modal. Once the user closes, we check to see if need to start the guide.
Otherwise, if `auto_show_persona` is not set, but `auto_show_guide` is (example is installed the extension later), then we can start the guide. Let me know if you have any comments on this.

2is10, line 267: This should work. Long term, I’ll probably separate the watches and terminate them when no longer applicable.

kifi-backend/modules/shoebox/angular/src/recos/recosView.tpl.html
```diff
@@ -1,81 +1,92 @@
...
-        <span class="kf-reco-cta-text">To get started, simply click the follow button on some of the libraries below</span>
-        <img class="kf-reco-cta-image" src="/img/webapp-ftui-follow-button-image.png">
+  <div class="kf-recos-persona-overlay" ng-if="autoShowPersona">
+    <div class="view-overlay">
+      <div class="persona-wrapper">
+        <div kf-manage-persona close-action="closeAutoShowPersonas"></div>
       </div>
+    </div>
+  </div>
+  <div ng-if="!autoShowPersona">
```
ahsu1230, line 10: wrap with an ng-if (load normal recommendations page iff `autoShowPersona` is falsey

kifi-backend/modules/shoebox/angular/src/recos/recosView.tpl.html
```diff
@@ -1,81 +1,92 @@
...
-  <div kf-delighted-survey hide="hideDelightedSurvey" ng-if="showDelightedSurvey"></div>
 
-  <h1 class="kf-recos-title kf-main-title">Recommendations</h1>
-
-  <div class="kf-reco-cols">
-    <div class="kf-reco-col-1">
-      <div class="kf-reco kf-reco-info kf-reco-cta" ng-if="recosState === 'noRecos'" ng-hide="initialCardClosed">
-        <span class="kf-reco-cta-text">To get started, simply click the follow button on some of the libraries below</span>
-        <img class="kf-reco-cta-image" src="/img/webapp-ftui-follow-button-image.png">
+  <div class="kf-recos-persona-overlay" ng-if="autoShowPersona">
```
ahsu1230, line 3: thanks @2is10 for this suggestion. Allowed me to avoid a couple of other issues I had before with the previous implementation

kifi-backend/modules/shoebox/angular/src/recos/recos.js
```diff
@@ -168,6 +189,25 @@ angular.module('kifi')
...
+    });
+
+
+    var unregisterAutoShowPersona = $scope.$watch(function () {
+      return profileService.prefs.auto_show_persona;
+    }, function (newValue, oldValue) {
+      $scope.autoShowPersona = newValue;
+      if (!newValue && oldValue) {
+        // stop listening when autoShowPersona from true -> null (means we're closing the auto-show-persona)
+        unregisterAutoShowPersona();
```
ahsu1230, line 217: terminate $watch after we have closed `autoShowPersona`

kifi-backend/modules/shoebox/angular/src/layout/main/MainCtrl.js
```diff
@@ -256,12 +256,14 @@ angular.module('kifi')
...
-    $rootScope.$watch(function () {
-      return Boolean(installService.installedVersion && profileService.prefs.auto_show_guide);
+
+    var unregisterAutoShowGuide = $rootScope.$watch(function () {
+      return Boolean(installService.installedVersion && profileService.prefs.auto_show_guide && !profileService.prefs.auto_show_persona);
     }, function (show) {
       if (show) {
         extensionLiaison.triggerGuide();
         profileService.savePrefs({auto_show_guide: null});
+        unregisterAutoShowGuide();
```
ahsu1230, line 266: slightly change this $watch to also look out for `auto_show_persona`. The guide should start when `auto_show_persona` is falsey as well since we want personas to show before the guide starts

ahsu1230, line 266: and terminate that watch once we've successfully triggered the guide

2is10, line 266: nice

kifi-backend/modules/shoebox/angular/src/recos/recos.js
```diff
@@ -13,6 +13,7 @@ angular.module('kifi')
     $scope.recosState = 'loading';
     $scope.initialCardClosed = false;
     $scope.noMoreRecos = false;
+    $scope.autoShowPersona = profileService.prefs.auto_show_persona;
```
2is10: We’ll sometimes get here before prefs have loaded. We can make the prefs a data dependency of this view to avoid that. See our existing uses of `require:` in route.js for examples.

ahsu1230:
OK! Things are ready to go! Releasing...

stkem:
:+1:

PR #13825, opened 2015-02-18
formatting large numbers with commas in library hovercard

2is10:
@ahsu1230


ahsu1230:
oh nice! forgot that you could format numbers like that

PR #13827, opened 2015-02-18
Trimming emails that come in.

andrewconner:
@stkem @thassss @jaredpetker @eishay 

This seems like the easiest place to do it.

The additional logging found that it's quite common for social signups to submit a form with a space at the end of the address. Looks like we've lost a few signups this way (people have tried several times, and can't figure out why their signup isn't working). Fortunately, easy fix.


2is10:
test case?

PR #13831, opened 2015-02-18
3.3.41: explcit_social_posting experiment UI

2is10:
@eishay

### DIFF
--- a/packrat/main.js
+++ b/packrat/main.js
@@ -619,7 +619,9 @@ api.port.on({
         og: !tab.usedHistoryApi && data.og || undefined,
         title: !tab.usedHistoryApi && data.ogTitle || data.title,
         guided: data.guided,
-        how: data.how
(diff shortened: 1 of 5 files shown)


packrat/main.js
```diff
@@ -619,7 +619,9 @@ api.port.on({
         og: !tab.usedHistoryApi && data.og || undefined,
         title: !tab.usedHistoryApi && data.ogTitle || data.title,
         guided: data.guided,
-        how: data.how
+        how: data.how,
+        fPost: data.fPost,
+        tweet: data.tweet
```
2is10, line 624: `fPost` and `tweet` will be either `true` or not present.

As you said, let’s be sure to check the experiment and the library visibility on both ends. We may also want to do some rudimentary de-duping on the server.

eishay:
Thanks!

PR #13832, opened 2015-02-18
suggested search cache

yingjieMiao:
@stkem @eishay


stkem:
why did you add it to all the other services cache modules? Only Shoebox needs it, unless I'm missing something. lgtm otherwise.

yingjieMiao:
I guess you are right. some cache is accessed by shoebox client. those 
cache need to be added to every service. this one should not.

stkem:
Let me know when it's deployed. I want to test the API. Could you generate some keywords for library id 30520 and put them in the admin? Just top 10.

yingjieMiao:
Ok. I am working on a batch of things.

You can test with fake data, right?

PR #13836, opened 2015-02-18
Fix internSystemGeneratedLibraries

ahsu1230:
- have `LibraryChecker` only page through active users
- have an extra safety layer for `internSystemGeneratedLibraries()` to also check for active user state

The second change caused a bunch of tests to fail (from nested readWrite operations) and bug out, so would be nice if there was another pair of eyes to check the tests aren't changed significantly. Some calls to `internSystem()` were not used at all either, so removed those from the tests as well.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
@@ -743,6 +743,7 @@ class LibraryCommander @Inject() (
     db.readWrite(attempts = 3) { implicit session =>
       val libMem = libraryMembershipRepo.getWithUserId(userId, None)
       val allLibs = libraryRepo.getByUser(userId, None)
+      val user = userRepo.get(userId)
 
       // Get all current system libraries, for main/secret, make sure only one is active.
       // This corrects any issues with previously created libraries / memberships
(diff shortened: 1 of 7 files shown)


kifi-backend/modules/shoebox/app/com/keepit/integrity/LibraryChecker.scala
```diff
@@ -26,11 +26,11 @@ class LibraryChecker @Inject() (
...
     val hourInd = currentTime.hourOfDay().get
     val minuteInd = currentTime.minuteOfHour().get / 10
+    val index = hourInd * 6 + minuteInd // hour * 6 + minute / 10 (int div)
 
     db.readOnlyMaster { implicit s =>
-      val index = hourInd * 6 + minuteInd // hour * 6 + minute / 10 (int div)
-      val pageSize = userRepo.count / 144 // 144 = 24 (hours per day) * 6 (10 minute intervals per hour)
-      userRepo.page(index, pageSize)
+      val pageSize = userRepo.countIncluding(UserStates.ACTIVE) / 144 // 144 = 24 (hours per day) * 6 (10 minute intervals per hour)
+      userRepo.pageIncluding(UserStates.ACTIVE)(index, pageSize)
```
ahsu1230, line 33: page only active users

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -754,27 +755,33 @@ class LibraryCommander @Inject() (
...
-        case (mem, lib) =>
-          libraryRepo.save(lib)
-          libraryMembershipRepo.save(mem)
+      // save changes for active users only
+      if (sysLibs.nonEmpty && user.state == UserStates.ACTIVE) {
+        sysLibs.map {
+          case (mem, lib) =>
+            libraryRepo.save(lib)
+            libraryMembershipRepo.save(mem)
+        }
```
ahsu1230, line 784: apply changes only for active users

kifi-backend/modules/shoebox/test/com/keepit/controllers/email/EmailRecosControllerTest.scala
```diff
@@ -89,20 +89,21 @@ class EmailRecosControllerTest extends Specification with ShoeboxTestInjector {
...
           val call = com.keepit.controllers.email.routes.EmailRecosController.keepReco(uri1.externalId)
           call.toString === s"/r/e/1/recos/keep?id=${uri1.externalId}"
 
-          inject[FakeUserActionsHelper].setUser(User(Some(Id[User](1L)), firstName = "Foo", lastName = "Bar", username = Username("test"), normalizedUsername = "test"))
+          inject[FakeUserActionsHelper].setUser(user1)
           val result = controller.keepReco(uri1.externalId)(FakeRequest())
           status(result) === SEE_OTHER // redirect
           db.readOnlyMaster { implicit session =>
-            val keeps = keepRepo.getByUser(Id[User](1))
+            val keeps = keepRepo.getByUser(user1.id.get)
```
ahsu1230, line 106: for some reason, a user was not created in `UserRepo` before. anybody know if this was on purpose? @joshm1 @stkem

joshm1, line 106: @ahsu1230 I don't remember specifics, but it doesn't look like a user needed to exist (in the original implementation).  The `User` object was a mock for authentication

ahsu1230, line 106: okay cool. since `internSystemGeneratedLibraries()` needs a valid active user, we'll go ahead and save a user in this test. Thanks @joshm1 !

PR #13838, opened 2015-02-18
temporary to fake-send activity emails to users

joshm1:
@stkem 

going to run the activity email sender once w/o actually sending the email just to populate the activity_emails table with recos these users likely received last Thursday


stkem:
:+1:

eishay:
Welcome back :-)

PR #13839, opened 2015-02-18
minor changes to /ext/guide response format

2is10:
@ahsu1230

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/model/Persona.scala
+++ b/kifi-backend/modules/common/app/com/keepit/model/Persona.scala
@@ -34,6 +34,38 @@ object Persona {
     (__ \ 'iconPath).format[String] and
     (__ \ 'activeIconPath).format[String]
   )(Persona.apply, unlift(Persona.unapply))
+
+  val libraryNames: Map[PersonaName, String] = Map(
+    PersonaName.DEVELOPER -> "Developer Resources",
+    PersonaName.TECHIE -> "Techie Picks",
+    PersonaName.ENTREPRENEUR -> "Entrepreneurial Articles",
+    PersonaName.ARTIST -> "Artistic Ideas",
+    PersonaName.FOODIE -> "Favorite Foods",
+    PersonaName.SCIENCE_BUFF -> "Scientific Picks",
+    PersonaName.FASHIONISTA -> "Fabulous Fashion",
+    PersonaName.HEALTH_NUT -> "Healthy Habits",
+    PersonaName.STUDENT -> "Student Resources",
+    PersonaName.INVESTOR -> "Investing Ideas",
+    PersonaName.TRAVELER -> "Travel Tips",
+    PersonaName.GAMER -> "Gaming News",
+    PersonaName.PARENT -> "Parenting Gems",
+    PersonaName.ANIMAL_LOVER -> "Animal Antics",
+    PersonaName.DEEP_THINKER -> "Deep Thoughts")
+
+  val names: Set[PersonaName] = libraryNames.keySet
+
+  // todo (aaron, ashley): pick out a keep for each persona
+  val keeps: Map[PersonaName, PersonaKeep] = Map()
+
+  // todo (aaron): once every persona has a keep, remove this!
+  val defaultKeep = PersonaKeep(
+    url = "http://www.ted.com/talks/steve_jobs_how_to_live_before_you_die",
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/common/app/com/keepit/model/Persona.scala
```diff
@@ -34,6 +34,38 @@ object Persona {
...
+
+  // todo (aaron): once every persona has a keep, remove this!
+  val defaultKeep = PersonaKeep(
+    url = "http://www.ted.com/talks/steve_jobs_how_to_live_before_you_die",
+    image = PersonaKeepImageInfo("//d1dwdv9wd966qu.cloudfront.net/img/guide/ted_jobs.jpg", 480, 425),
+    noun = "video",
+    query = "steve+jobs",
+    title = "Steve Jobs: How to live before you die | Talk Video | TED.com",
+    matches = Json.obj("title" -> Json.toJson(Seq(Seq(0, 5), Seq(6, 4))), "url" -> Json.toJson(Seq(Seq(25, 5), Seq(31, 4)))),
+    track = "steveJobsSpeech")
```
2is10, line 68: just moved all of the global values up into `Persona` so call sites wouldn’t have to use the longer `PersonaName` or `PersonaKeep` somewhat arbitrarily

2is10, line 68: Also added `//d1dwdv9wd966qu.cloudfront.net` in front of the image path, since that was previously auto-inserted into our production kifi.js file by our angular app’s gulpfile. The less the extension needs to know about our image hosting, the better.

kifi-backend/modules/common/app/com/keepit/model/Persona.scala
```diff
@@ -78,69 +110,20 @@ object PersonaName {
...
+  url: String, // absolute or relative to https://www.kifi.com
+  width: Int, // image's natural dimensions; will be displayed at half size (for hi-res displays)
+  height: Int)
+
+@json case class PersonaKeep( // for keeper FTUE
   url: String,
-  imagePath: String,
-  imageWidth: Int,
-  imageHeight: Int,
+  image: PersonaKeepImageInfo,
```
2is10, line 124: Grouped the image info into an object (as we do with `LibraryImageInfo`) to simplify the top-level object.

kifi-backend/modules/common/app/com/keepit/model/Persona.scala
```diff
@@ -34,6 +34,38 @@ object Persona {
...
+    PersonaName.HEALTH_NUT -> "Healthy Habits",
+    PersonaName.STUDENT -> "Student Resources",
+    PersonaName.INVESTOR -> "Investing Ideas",
+    PersonaName.TRAVELER -> "Travel Tips",
+    PersonaName.GAMER -> "Gaming News",
+    PersonaName.PARENT -> "Parenting Gems",
+    PersonaName.ANIMAL_LOVER -> "Animal Antics",
+    PersonaName.DEEP_THINKER -> "Deep Thoughts")
+
+  val names: Set[PersonaName] = libraryNames.keySet
```
2is10, line 55: using `.keySet` to avoid repeating all of the persona names

2is10:
merging so I can deploy and develop against this. happy to hear and address any comments.

kifi-backend/modules/common/app/com/keepit/model/Persona.scala
ahsu1230, line 68: Thanks, this is good to know. I'm going to start getting default keeps from Ashley, so we'll need to upload the image we want to use in the same place. Does the image size have to be 480x425? Or can it be a different size?

2is10, line 68: They can be different sizes, but we should aim for roughly the same size, for consistency. Please use even image dimensions—not odd like the `425` height we now have.

I think the best place to put them is in our `angular/img/guide/` directory. Then our gulpfile will take care of uploading automatically. Afterwards, we can look at their URLs and hardcode them into this file. That’s necessary because our upload process fingerprints them, since they’re infinitely cacheable.

ahsu1230, line 68: yup sounds good!

ahsu1230, line 68: I think gulp doesn't allow odd dimensions anyway.

PR #13840, opened 2015-02-18
Suggested searches on a library page FE

stkem:
also removes the search header (filters, counts and such) for
in-library search per Marks request.

@2is10 
cc: @yingjieMiao @andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/library.tpl.html
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/library.tpl.html
@@ -9,7 +9,10 @@
 
 <div class="kf-lib-cols">
   <div class="kf-lib-col" ui-view></div>
-  <div kf-related-libraries related-libraries="relatedLibraries" parent-library="library" ng-if="relatedLibraries"></div>
+  <section class="kf-lib-right-col">
+    <div kf-suggested-searches terms="['matrix', 'matlab']" parent-library="library"></div>
+    <div kf-related-libraries related-libraries="relatedLibraries" parent-library="library" ng-if="relatedLibraries"></div>
+  </section>
 </div>
 
 <footer class="kf-lib-footer" ng-if="!$root.userLoggedIn">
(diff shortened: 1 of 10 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/library.tpl.html
```diff
@@ -9,7 +9,10 @@
 
 <div class="kf-lib-cols">
   <div class="kf-lib-col" ui-view></div>
-  <div kf-related-libraries related-libraries="relatedLibraries" parent-library="library" ng-if="relatedLibraries"></div>
+  <section class="kf-lib-right-col">
+    <div kf-suggested-searches terms="['matrix', 'matlab']" parent-library="library"></div>
```
2is10, line 13: `['matrix', 'matlab']` ?

kifi-backend/modules/shoebox/angular/src/libraries/libraryService.js
```diff
@@ -67,6 +67,7 @@ angular.module('kifi')
         if (res.data && res.data.library) {
           res.data.library.access = res.data.membership;
           res.data.library.listed = res.data.listed;
+          res.data.library.suggestedSearches = (res.data.suggestedSearches && res.data.suggestedSearches.terms) || [];
```
2is10, line 70: We should hide this section on mobile devices, no? (since we don’t yet support library search on them)

kifi-backend/modules/shoebox/angular/src/libraries/library.tpl.html
stkem, line 13: Oops. Test data, doesn't actually do anything any more. Will remove.

kifi-backend/modules/shoebox/angular/src/libraries/libraryService.js
stkem, line 70: Ah, you have a point. It actually works, but looks a little crappy. I'll hide it on mobile for now.

PR #13841, opened 2015-02-18
having /ext/guide response include the library path too

2is10:
just for clarity, so that it won’t feel weird on the receiving end that the other fields aren’t used

@ahsu1230


2is10:
closing because I think it might actually be convenient to use this information in the search result on google.com

2is10:
Okay, I’ve decided to keep the info and include the path, too, since search results need it.

ahsu1230:
gotcha :+1:

PR #13842, opened 2015-02-18
using explicit social sharing

eishay:


kifi-backend/modules/shoebox/app/com/keepit/commanders/TwitterPublishingCommander.scala
```diff
@@ -28,7 +28,7 @@ class TwitterPublishingCommander @Inject() (
 
   def publishKeep(userId: Id[User], keep: Keep, library: Library): Unit = {
     require(keep.userId == userId, s"User $userId cannot publish to Twitter a keep by user ${keep.userId}")
-    if (library.visibility == LibraryVisibility.PUBLISHED && hasTwitterExperiment(userId)) {
+    if (library.visibility == LibraryVisibility.PUBLISHED && hasExplicitShareExperiment(userId)) {
```
2is10, line 31: It’s a little unfortunate that we are going back to the database to verify experiment membership when the user’s experiment set is already available in `UserRequest.experiments`.

2is10:
:+1:

PR #13845, opened 2015-02-19
properly update tags when transfering a library

eishay:
reviewed with @leogrim


eishay:
@andrewconner

andrewconner:
great, thanks

PR #13847, opened 2015-02-19
new user pref force_show_persona

ahsu1230:
@2is10


kifi-backend/modules/common/app/com/keepit/model/UserValue.scala
```diff
@@ -78,6 +78,7 @@ object UserValueName {
   val SITE_INTRODUCE_LIBRARY_MENU = UserValueName("site_introduce_library_menu")
 
   val AUTO_SHOW_GUIDE = UserValueName("auto_show_guide")
+  val FORCE_SHOW_PERSONA = UserValueName("force_show_persona") // for new users, show who-are-you modal on page land
```
2is10, line 81: I think `force_choose_persona` would make more sense, since we’re forcing the user to choose one. Or `auto_show_persona` since we’re automatically showing it.

2is10:
:+1:

PR #13853, opened 2015-02-19
shorter tweets

eishay:


kifi-backend/modules/shoebox/app/com/keepit/commanders/TwitterMessages.scala
```diff
@@ -6,20 +6,24 @@ class TwitterMessages {
...
     val contentLength = title.length + libName.length
-    val totalLength = 20 * 2 + 20 + title.length + libName.length
-    if (20 * 2 + 20 + title.length + libName.length <= 140) {
+    val totalLength = UrlLen * 2 + Overhead + title.length + libName.length
+    if (UrlLen * 2 + Overhead + title.length + libName.length <= 140) {
       s"$title $keepUrl kept to $libName $libUrl via @kifi"
     } else {
-      val overtext = 79 - (2 * 3) - contentLength //the 3 stands for the "..."
-      val maxLibName = (libName.size - overtext / 3).min(20)
+      val overtext = contentLength - ContextSpace - (2 * 3) //the 3 stands for the "..."
```
2is10, line 22: We could use `…` (U+2026 horizontal ellipsis), which is one character, instead of three periods.

eishay, line 22: oh, nice! will do

PR #13857, opened 2015-02-19
some users may bypass abuse controll

eishay:


eishay:
@andrewconner @stkem @leogrim

PR #13858, opened 2015-02-19
Ensure we quickly get a summary for twitter syncs

leogrim:
@andrewconner @stkem


andrewconner:
Great! We may be able to do more than concurrency of 1, but let's start there and see.

stkem:
:-)

PR #13862, opened 2015-02-19
fix library kind assertion for social tooltip, demote persona libraries ...

eishay:
...and promote libraries with more followers


eishay:
@2is10 @ahsu1230

2is10:
:+1: thanks!

PR #13865, opened 2015-02-20
15 PERSONA KEEPS!!!

ahsu1230:
@2is10 @stkem 

did my best in getting the pictures to look nice. Some of the styling (like padding & stuff) is the website's fault =P


kifi-backend/modules/shoebox/app/com/keepit/model/PersonaRepo.scala
```diff
@@ -58,7 +58,7 @@ class PersonaRepoImpl @Inject() (
   }
 
   private val getByStateCompiled = Compiled { (state: Column[State[Persona]]) =>
-    (for (r <- rows if r.state === state) yield r)
+    (for (r <- rows if r.state === state) yield r).sortBy(_.createdAt)
```
ahsu1230: @stkem sort by Created At instead of id

stkem: Thanks!

kifi-backend/modules/common/app/com/keepit/model/Persona.scala
```diff
@@ -54,10 +54,152 @@ object Persona {
 
   val names: Set[PersonaName] = libraryNames.keySet
 
-  // todo (aaron, ashley): pick out a keep for each persona
-  val keeps: Map[PersonaName, PersonaKeep] = Map()
+  val keeps: Map[PersonaName, PersonaKeep] = Map(
+    PersonaName.DEVELOPER -> PersonaKeep(
+      title = "Top 10 Professional Sample Code Websites For Programmers",
+      url = "http =//www.makeuseof.com/tag/top-10-professional-sample-code-websites-for-programmers/",
```
2is10: several of these URLs have `http=` instead of `http:` (broken)

kifi-backend/modules/shoebox/app/com/keepit/model/PersonaRepo.scala
2is10: Whenever we sort by a time, we should use `id` as a secondary sort to ensure a stable sort ordering.

``` scala
.sortBy(r => (r.createdAt, r.id))
```

2is10:
:+1: Nice work!

We may want to add padding ourselves manually into some of these page thumbnails. I’ll let Product/QA handle that.

PR #13868, opened 2015-02-20
minor adjustments to suggested searches

2is10:
@stkem

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/library.tpl.html
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/library.tpl.html
@@ -10,7 +10,7 @@
 <div class="kf-lib-cols kf-width-bounded">
   <div class="kf-lib-col" ui-view></div>
   <section class="kf-lib-right-col">
-    <div kf-suggested-searches parent-library="library"></div>
+    <div kf-suggested-searches library="library" ng-if="library.suggestedSearches.length && !isMobile"></div>
     <div kf-related-libraries related-libraries="relatedLibraries" parent-library="library" ng-if="relatedLibraries"></div>
   </section>
 </div>
--- a/kifi-backend/modules/shoebox/angular/src/libraries/suggestedSearches.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/suggestedSearches.js
@@ -3,28 +3,14 @@
 angular.module('kifi')
 
 .directive('kfSuggestedSearches', [
-  '$location', 'platformService',
-  function ($location, platformService) {
+  function () {
     return {
       restrict: 'A',
       replace: true,
       scope: {
-        parentLibrary: '&'
+        library: '='
       },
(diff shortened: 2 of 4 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/library.tpl.html
```diff
@@ -10,7 +10,7 @@
 <div class="kf-lib-cols kf-width-bounded">
   <div class="kf-lib-col" ui-view></div>
   <section class="kf-lib-right-col">
-    <div kf-suggested-searches parent-library="library"></div>
+    <div kf-suggested-searches library="library" ng-if="library.suggestedSearches.length && !isMobile"></div>
```
2is10, line 13: Skipping the whole directive earlier when it doesn’t apply. This runs faster and keeps the directive simpler.

kifi-backend/modules/shoebox/angular/src/libraries/suggestedSearches.js
```diff
@@ -3,28 +3,14 @@
...
-      templateUrl: 'libraries/suggestedSearches.tpl.html',
-      link: function (scope/*, element, attrs*/) {
-        if (!platformService.isSupportedMobilePlatform()) {
-          var parentLibrary = scope.parentLibrary();
-          scope.termsWithActions = _.map(parentLibrary.suggestedSearches, function (term) {
-            return {
-              name: term,
-              action: function () {
-                $location.url(parentLibrary.url + '/find?q=' + term);
-              }
```
2is10, line 23: It’s lighter to define a single handler function (on the scope) and call it from all elements that need it, rather than create one function per element. In this case, we don’t need any handlers, since these are really just links.

kifi-backend/modules/shoebox/angular/src/libraries/suggestedSearches.styl
```diff
@@ -1,20 +1,22 @@
-.suggested-searches
+.kf-lss
   border-radius: 10px
   border: 1px solid #dadae2
   padding: 10px 16px
   background: #fff
   margin-bottom: 15px
 
-.suggested-searches-header
+.kf-lss-title
```
2is10, line 8: We (Yiping, Aaron, Andrew, and I, at least) decided to use a `kf-` prefix on all of our class names.

We also started using a short 2-3 character abbreviation for the directive to keep class names shorter and so that the important part of the name (after the prefix) would stand out more.

stkem:
Come on! The site has many problems and you chose to basically redo a perfectly functional component - incorrect class naming convention aside - that I build _yesterday_ ?

2is10:
I’m sorry if it feels like I’m picking on you or wasting time. That wasn’t my intent. I assure you this was a very quick change. I wasn’t motivated by the state of the code so much as the opportunity to illustrate our newer conventions for you.

PR #13876, opened 2015-02-20
on social graph updates - transfering some load to replica db and reduci...

eishay:
...ng the time a write transaction should be open


eishay:
@andrewconner

PR #13877, opened 2015-02-20
take advantage of cache

yingjieMiao:
@stkem @eishay 

This should already help a lot. Will look at statsd after deploy.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/RelatedLibraryCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/RelatedLibraryCommander.scala
@@ -112,7 +112,7 @@ class RelatedLibraryCommanderImpl @Inject() (
 
   def topicRelatedLibraries(libId: Id[Library]): Future[Seq[RelatedLibrary]] = {
     cortex.similarLibraries(libId, limit = SUB_RETURN_SIZE).map { ids =>
-      db.readOnlyReplica { implicit s =>
+      db.readOnlyMaster { implicit s =>
         ids.map { id => libRepo.get(id) }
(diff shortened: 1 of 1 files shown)


eishay:
what about @stkem recommendations regarding caching the entire method's response?

yingjieMiao:
where does the cache sit in? Cortex returns a seq of ids. What will the cache Key look like? I feel like caching a collection of results is the "last resort" for performance tweaking. And it can make cache invalidation logic weak/complicated. Maybe I missed something. 

Anyhow, I think this should already improve things quite a bit.

stkem:
I doubt this will make a significant difference. If I'm not mistaken, read only sessions on the replica already read from the cache, they just don't write to it and the library by id cache is very hot (virtually no misses according to statsd).

@yingjieMiao we discussed how to do this this morning. Replace the very short lived request consolidator in `suggestedLibraries` (on line 88, I think) with a proper cache, with a short TTL (say, 2 hours on memcached, 30 minutes in memory). No cache invalidation is needed, the entries will just expire. The Key would simply be a `Id[Library]` and the values `Seq[RelatedLibrary]`. The `suggestedLibraries` method currently consistently takes >50ms (we looked at that together). This would bring it down to 2-5ms.

yingjieMiao:
`suggestedLibraries` method consists of 3 parts:
1) cortex return related lib ids. look up shoebox table
2) look up libraries from same owner
3) look up popular libraries from library table

Part 3) alone can take 50ms to 100ms. 

I am under the impression that the original goal is to reduce num of db calls. I haven't heard complains about latency.

stkem:
Yes, and having a cache on `suggestedLibraries` would save _all_ of these calls. I was using latency as a proxy for how expensive these computations are.

PR #13878, opened 2015-02-20
A bit less exception noise from curator

stkem:
@eishay @yingjieMiao

### DIFF
--- a/kifi-backend/modules/curator/app/com/keepit/curator/commanders/AllKeepsSeedIngestionHelper.scala
+++ b/kifi-backend/modules/curator/app/com/keepit/curator/commanders/AllKeepsSeedIngestionHelper.scala
@@ -158,7 +158,7 @@ class AllKeepSeedIngestionHelper @Inject() (
         updateRawSeedItem(rawSeedItem, keep.uriId, keep.keptAt, countChange, discoverable)
       }
     } else {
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/curator/app/com/keepit/curator/commanders/AllKeepsSeedIngestionHelper.scala
```diff
@@ -158,7 +158,7 @@ class AllKeepSeedIngestionHelper @Inject() (
         updateRawSeedItem(rawSeedItem, keep.uriId, keep.keptAt, countChange, discoverable)
       }
     } else {
-      airbrake.notify(s"Missing RSI: keepId ${keepInfo.keepId}, oldUriId ${keepInfo.uriId}, newUriId ${keep.uriId}. Skipping this keep update.")
+      log.info(s"Missing RSI: keepId ${keepInfo.keepId}, oldUriId ${keepInfo.uriId}, newUriId ${keep.uriId}. Skipping this keep update.")
     }
```
yingjieMiao, line 162: these is something odd about this: in the error msgs `oldUriId` always equals to `newUriId`.

Race condition with the reaper?

stkem, line 162: There is no reaper for raw_seed_item. I think there still is a problem with the scheduler where the ingestion sometimes runs on both machines. Still need to investigate that.

PR #13883, opened 2015-02-21
eliminating unnecessary IIFE

2is10:
@andrewconner 

view diff with `?w=` to ignore indentation change


andrewconner:
works for me

PR #13884, opened 2015-02-21
using google uri shortner

eishay:


eishay:
@andrewconner

2is10:
Longer term, perhaps, we could use http://kifi.link/rT7c0 as our shortened
URLs, with a persistent key-to-URL mapping in Voldemort or something.

kifi.link appears to be available for $9.88/month at uniregistry.com. It
seemed the best of the TLDs listed here
http://en.wikipedia.org/wiki/List_of_Internet_top-level_domains.

eishay:
Cool, I just bought:
- kifi.link
- kifi.ink
- kifi.click
- kifi.io

PR #13890, opened 2015-02-22
filtering out foursquare checkins and run keeper activities from tweet imports

eishay:


eishay:
@andrewconner @stkem not sure if it covers the twitter ongoing sync, can you check?

stkem:
Yup, looks like it covers the ongoing sync as well.

PR #13891, opened 2015-02-23
stop using komodia

eishay:
http://arstechnica.com/security/2015/02/ssl-hijacker-behind-superfish-debacle-imperils-big-number-of-users/


eishay:
@andrewconner

PR #13901, opened 2015-02-23
make sure recos are ingested

yingjieMiao:
@stkem @ahsu1230


kifi-backend/modules/shoebox/app/com/keepit/commanders/UserPersonaCommander.scala
```diff
@@ -53,19 +58,21 @@ class UserPersonaCommanderImpl @Inject() (
       }
     }
 
-    curator.ingestPersonaRecos(userId, personasToPersist.values.map { _.id.get }.toSeq.distinct)
+    curator.ingestPersonaRecos(userId, personasToPersist.values.map { _.id.get }.toSeq.distinct).collect {
```
ahsu1230, line 61: :+1:

stkem:
:+1:

PR #13905, opened 2015-02-24
on ExternalMessagingRateLimitException: adding more info about the user ...

eishay:
...and how many messages he tried to send


eishay:
@stkem

PR #13907, opened 2015-02-24
moving /friends to /connections

2is10:
@andrewconner

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
+++ b/kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
@@ -248,7 +248,7 @@ trait UserActions extends Logging { self: Controller =>
           val nRes = Redirect("/login")
           // Less than ideal, but we can't currently test this:
           Play.maybeApplication match {
-            case Some(_) => nRes.withSession(result.session + ("original-url" -> request.uri))
+            case Some(_) => nRes.withSession(result.session + (SecureSocial.OriginalUrlKey -> request.uri))
             case None => nRes
           }
         case result =>
--- a/kifi-backend/modules/shoebox/angular/src/route.js
+++ b/kifi-backend/modules/shoebox/angular/src/route.js
@@ -19,11 +19,7 @@ angular.module('kifi')
           return match[1] + match[2]; // remove trailing slash
         }
       })
-      .when('/friends/invite', '/invite')
-      .when('/friends/requests', '/friends')
-      .when('/friends/requests/:network', '/friends')
       .when('/:username/libraries', '/:username')
-      .when('/recommendations', '/')
       .otherwise('/');  // last resort
 
     // Set up the states.
@@ -32,8 +28,8 @@ angular.module('kifi')
         url: '/',
         templateUrl: 'recos/recosView.tpl.html'
       })
-      .state('friends', {
-        url: '/friends',
+      .state('connections', {
+        url: '/connections',
         templateUrl: 'friends/friends.tpl.html'
       })
       .state('invite', {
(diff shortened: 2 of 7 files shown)


kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
```diff
@@ -248,7 +248,7 @@ trait UserActions extends Logging { self: Controller =>
           val nRes = Redirect("/login")
           // Less than ideal, but we can't currently test this:
           Play.maybeApplication match {
-            case Some(_) => nRes.withSession(result.session + ("original-url" -> request.uri))
+            case Some(_) => nRes.withSession(result.session + (SecureSocial.OriginalUrlKey -> request.uri))
```
andrewconner, line 251: I believe this may cause a test to fail. But if it doesn't, yay.

kifi-backend/modules/shoebox/angular/src/route.js
```diff
@@ -19,11 +19,7 @@ angular.module('kifi')
           return match[1] + match[2]; // remove trailing slash
         }
       })
-      .when('/friends/invite', '/invite')
-      .when('/friends/requests', '/friends')
-      .when('/friends/requests/:network', '/friends')
       .when('/:username/libraries', '/:username')
-      .when('/recommendations', '/')
```
2is10, line 26: routes I removed here should be handled on the server as 301s now

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/KifiSiteRouter.scala
```diff
@@ -56,65 +47,48 @@ object KifiSiteRouter {
...
-    } else if (request.path == "/" && request.userIdOpt.isEmpty) {
-      //should we ever get to this line???
-      Redirect(com.keepit.controllers.website.routes.HomeController.home)
-    } else if (userAgentOpt.exists(_.isMobile) &&
+    } else if (request.userAgentOpt.exists(_.isMobile) &&
       request.queryString.get(KifiMobileAppLinkFlag.key).exists(_.contains(KifiMobileAppLinkFlag.value))) {
       Ok(views.html.mobile.MobileRedirect(request.uri))
+    } else if (request.userOpt.isEmpty && request.identityOpt.isDefined) {
+      // non-authed client, but identity is set. mid-signup, send them there.
+      Redirect(com.keepit.controllers.core.routes.AuthController.signupPage())
```
2is10, line 67: moved this up from line 103 below to simplify the `match`

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/KifiSiteRouter.scala
```diff
@@ -153,38 +142,53 @@ class AngularRouter @Inject() (
...
-      }.getOrElse(RedirectToLogin("/" + path.path)))
-    } else if (path.split.length > 0) {
-      val userOpt = userCommander.getUserByUsernameOrAlias(Username(path.primary))
-
-      userOpt.flatMap {
+      request.userOpt.map { user =>
+        SeeOtherRoute("/" + (user.username.value +: path.segments.drop(1)).map(r => URLEncoder.encode(r, UTF8)).mkString("/"))
+      } orElse {
+        Some(RedirectToLogin(path.path))
+      }
```
2is10, line 167: no substantive change here, just readability

kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
2is10, line 251: All the shoebox tests passed. I’ll run all of them to be sure.

andrewconner:
Looks good.

kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
andrewconner, line 251: Ah okay. We've had to do it before because of something related to dependencies in non-shoebox services, needing an application to be running, or something. If it passes, it's probably fine.

PR #13917, opened 2015-02-24
don't shuffle hand picked with others

yingjieMiao:
@stkem


stkem:
Nice, thanks!

stkem:
lgtm

yingjieMiao:
FYI, I didn't tune up from 5 to 10 libraries. That will affect overall user experience. I would prefer incremental changes here.

stkem:
This does not appear to work correctly. The libraries are in there, but they are not at the top. Please investigate.
Also, with 2 libraries per persona, we'll need to increase the lib count to 10 to have good coverage.

PR #13919, opened 2015-02-24
Adding AdRoll (retargeting) tracker to the angular app

stkem:
@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/angular/index.html
+++ b/kifi-backend/modules/shoebox/angular/index.html
@@ -29,3 +29,20 @@
   s.parentNode.insertBefore(g, s);
   }(document, 'script');
 </script>
+<script type="text/javascript">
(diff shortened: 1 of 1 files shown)


stkem:
Is there a better place to put this? Needs to be on both public and logged in library and profile pages.

stkem:
Getting this out so sterling can test. Happy to move it somewhere else later.

andrewconner:
Ew, do we really need to add it to every angular load?

stkem:
That is what I was told. Or rather "Every profile and library page" which is nearly all of the app (in terms of pages hit)

2is10:
How about only if the viewer is logged out?

andrewconner:
I proposed that, but this PR got reverted because there's a few issues with this implementation. Sterling will decide what's next.

PR #13920, opened 2015-02-24
twitter waitlist basic backend support

ahsu1230:
@andrewconner 
- Twitter Waitlist table only has one constraint: unique pairs for `(userId, twitterHandle)`
- Twitter Waitlist Commander has two methods: `add() & edit()`
- Twitter Waitlist Repo also has a `countActiveEntriesBeforeDateTime()` which counts both `active` and `accepted` entry states before some given time


kifi-backend/modules/shoebox/app/com/keepit/commanders/TwitterWaitlistCommander.scala
```diff
@@ -0,0 +1,57 @@
...
+      }
+    }
+    entryOpt.right.map { entry =>
+      db.readWrite { implicit s =>
+        twitterWaitlistRepo.save(entry)
+      }
+    }
+  }
+
+  def editEntry(entryId: Id[TwitterWaitlistEntry], stateOpt: Option[State[TwitterWaitlistEntry]] = None): TwitterWaitlistEntry = {
```
ahsu1230: kept this function pretty broad so it's easy to add other parameters in case we want to edit other things in future.

kifi-backend/modules/common/conf/evolutions/shoebox/298.sql
```diff
@@ -0,0 +1,18 @@
...
+CREATE TABLE if not exists twitter_waitlist (
+    id bigint(20) NOT NULL AUTO_INCREMENT,
+    created_at datetime NOT NULL,
+    updated_at datetime NOT NULL,
+    user_id bigint(20) NOT NULL,
+    twitter_handle varchar(16) NOT NULL,
+    state varchar(20) NOT NULL,
+
+    PRIMARY KEY (id),
+    UNIQUE INDEX twitter_waitlist_u_user_id_twitter_handle (user_id, twitter_handle)
```
andrewconner: How about just on user_id?

ahsu1230: no need to check for unique `(user_id, handle)` pairs?

kifi-backend/modules/shoebox/app/com/keepit/commanders/TwitterWaitlistCommander.scala
andrewconner: Not sure it's necessary yet. This will eventually have code that will start a twitter sync (using Stephen's twitter stuff), and update the row. Until then, we won't be editing.

Also, we likely would never deal directly with `Id[TwitterWaitlistEntry]`s. More likely just userIds.

kifi-backend/modules/common/conf/evolutions/shoebox/298.sql
andrewconner: If you have a unique on user_id, handle part doesn't really matter. Since users can only auth with one twitter account, they couldn't end up waitlisted twice (and if they manage to do so, we'd only sync one of them).

andrewconner:
Great thanks

kifi-backend/modules/shoebox/app/com/keepit/commanders/TwitterWaitlistCommander.scala
ahsu1230: hm okay. it can be taken out

kifi-backend/modules/common/conf/evolutions/shoebox/298.sql
ahsu1230: I see, sounds good.

PR #13924, opened 2015-02-25
setting disk space alerts to higher numbers

eishay:


eishay:
@andrewconner @stkem @leogrim fyi

PR #13929, opened 2015-02-25
no longer suggesting users keep to libraries they follow

2is10:
@stkem

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/app.js
+++ b/kifi-backend/modules/shoebox/angular/src/app.js
@@ -90,7 +90,7 @@ angular.module('kifi', [
         profileService.fetchMe().then(function () {
           if ($rootScope.userLoggedIn) {
             profileService.fetchPrefs();
-            libraryService.fetchLibrarySummaries(true);
+            libraryService.fetchLibraryInfos(true);
           }
         });
       });
--- a/kifi-backend/modules/shoebox/angular/src/common/directives/keepWho/keepWhoLibsAndPics.js
+++ b/kifi-backend/modules/shoebox/angular/src/common/directives/keepWho/keepWhoLibsAndPics.js
@@ -34,7 +34,7 @@ angular.module('kifi')
 
         // For keep cards that search results on the search page, add and remove user library attribution
         // on the client side when the user keeps or unkeeps.
-        var deregisterKeepAddedListener = $rootScope.$on('keepAdded', function (e, libSlug, keeps, library) {
+        var deregisterKeepAddedListener = $rootScope.$on('keepAdded', function (e, keeps, library) {
           var visibleLibraryIds = _.pluck(scope.visibleKeepLibraries, 'id');
           _.each(keeps, function (keep) {
             if (!scope.keep.id &&                      // No scope.keep.id if the keep is not on a library page.
--- a/kifi-backend/modules/shoebox/angular/src/keep/addKeep.js
+++ b/kifi-backend/modules/shoebox/angular/src/keep/addKeep.js
@@ -60,19 +60,9 @@ angular.module('kifi')
           $document.off('keydown', processKey);
         });
 
-        var writeableLibs = _.filter(libraryService.librarySummaries, function (lib) {
-          return lib.access !== 'read_only';
-        });
-
-        var defaultLib = _.find(writeableLibs, { 'kind': 'system_main' });
-        if (_.isEmpty(scope.modalData.currentLib)) {
-          scope.selectedLibrary = defaultLib;
-        } else {
-          scope.selectedLibrary = _.find(writeableLibs, { 'id': scope.modalData.currentLib.id });
-          if (!scope.selectedLibrary) {
-            scope.selectedLibrary = defaultLib;
-          }
-        }
+        scope.selectedLibrary =
+          scope.modalData.selectedLibId && _.find(libraryService.getOwnInfos(), {id: scope.modalData.selectedLibId}) ||
+          libraryService.getSysMainInfo();
 
         scope.onLibrarySelected = function (selectedLibrary) {
           scope.selectedLibrary = selectedLibrary;
@@ -88,7 +78,7 @@ angular.module('kifi')
                   template: 'common/modal/genericErrorModal.tpl.html'
                 });
               } else {
-                libraryService.fetchLibrarySummaries(true);
+                libraryService.fetchLibraryInfos(true);
                 if (scope.selectedLibrary.visibility === 'secret') {
                   scope.keptPrivate = true;
                 } else {
@@ -108,7 +98,7 @@ angular.module('kifi')
                     var keep = new keepDecoratorService.Keep(fullKeep);
                     keep.buildKeep(keep);
                     keep.makeKept();
-                    $rootScope.$emit('keepAdded', libraryService.getSlugById(scope.selectedLibrary.id), [keep]);
+                    $rootScope.$emit('keepAdded', [keep], scope.selectedLibrary);
                   });
                 }
               }
--- a/kifi-backend/modules/shoebox/angular/src/keep/keep.js
+++ b/kifi-backend/modules/shoebox/angular/src/keep/keep.js
@@ -63,7 +63,7 @@ angular.module('kifi')
 
         scope.isMyLibrary = false;
         scope.$emit('getCurrentLibrary', { callback: function (lib) {
-          scope.isMyLibrary = lib.access === 'owner';
+          scope.isMyLibrary = lib.isMine;
         }});
 
         // test data:
@@ -351,15 +351,10 @@ angular.module('kifi')
         });
 
         scope.$watch(function () {
-          return libraryService.librarySummaries.length;
+          return libraryService.getOwnInfos().length;
         }, function (newVal) {
           if (newVal) {
-            scope.keep.libraryInfo = libraryService.getLibraryInfoById(scope.keep.libraryId);
-
-            scope.libraries = _.filter(libraryService.librarySummaries, function(library) {
-              return (library.access !== 'read_only') && (library.id !== scope.keep.libraryId);
-            });
-            scope.data = {};
+            scope.libraries = _.reject(libraryService.getOwnInfos(), {id: scope.keep.libraryId});
           }
         });
 
@@ -373,8 +368,9 @@ angular.module('kifi')
         });
 
         element.on('mousemove', function (e) {
-          mouseX = e.pageX - util.offset(element).left;
-          mouseY = e.pageY - util.offset(element).top;
+          var offset = util.offset(element);
+          mouseX = e.pageX - offset.left;
+          mouseY = e.pageY - offset.top;
         })
         .on('dragstart', function (e) {
           scope.$apply(function () {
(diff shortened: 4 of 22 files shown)


kifi-backend/modules/shoebox/angular/src/app.js
```diff
@@ -90,7 +90,7 @@ angular.module('kifi', [
         profileService.fetchMe().then(function () {
           if ($rootScope.userLoggedIn) {
             profileService.fetchPrefs();
-            libraryService.fetchLibrarySummaries(true);
+            libraryService.fetchLibraryInfos(true);
```
2is10, line 93: what the angular app calls a “library summary” corresponds to `LibraryInfo` on the server. renaming for consistency.

kifi-backend/modules/shoebox/angular/src/common/directives/keepWho/keepWhoLibsAndPics.js
```diff
@@ -34,7 +34,7 @@ angular.module('kifi')
 
         // For keep cards that search results on the search page, add and remove user library attribution
         // on the client side when the user keeps or unkeeps.
-        var deregisterKeepAddedListener = $rootScope.$on('keepAdded', function (e, libSlug, keeps, library) {
+        var deregisterKeepAddedListener = $rootScope.$on('keepAdded', function (e, keeps, library) {
```
2is10, line 37: `libSlug` was an unnecessary argument to this event. removed it everywhere the event is fired or handled.

kifi-backend/modules/shoebox/angular/src/keep/addKeep.js
```diff
@@ -60,19 +60,9 @@ angular.module('kifi')
           $document.off('keydown', processKey);
         });
 
-        var writeableLibs = _.filter(libraryService.librarySummaries, function (lib) {
-          return lib.access !== 'read_only';
-        });
+        var writeableLibs = libraryService.getOwnInfos();
```
2is10: This is where the bug was manifesting. There are at least three endpoints that return a `LibraryInfo`. Only one inserts the extra `"access"` property (not part of the `LibraryInfo` case class). Yet the JS code here assumes the `.access` property is 100% reliable. My solution is to stop setting the `.access` property altogether, since it would be impossible or impractical to set it everywhere `LibraryInfo` is used on the server. Instead, we distinguish between libraries the user owns and those he/she is following by keeping them in different arrays inside `libraryService`.

kifi-backend/modules/shoebox/angular/src/keep/keep.js
```diff
@@ -351,15 +351,10 @@ angular.module('kifi')
         });
 
         scope.$watch(function () {
-          return libraryService.librarySummaries.length;
+          return libraryService.getOwnInfos().length;
         }, function (newVal) {
           if (newVal) {
-            scope.keep.libraryInfo = libraryService.getLibraryInfoById(scope.keep.libraryId);
```
2is10, line 357: There are no other occurrences of `libraryInfo` in the codebase, so I assume this line was superfluous.

kifi-backend/modules/shoebox/angular/src/keep/keep.js
```diff
@@ -351,15 +351,10 @@ angular.module('kifi')
...
-
-            scope.libraries = _.filter(libraryService.librarySummaries, function(library) {
-              return (library.access !== 'read_only') && (library.id !== scope.keep.libraryId);
-            });
-            scope.data = {};
```
2is10, line 362: I could not find any uses of this `scope.data` object, so also removing it.

kifi-backend/modules/shoebox/angular/src/keep/keepToLibraryWidget.js
```diff
@@ -167,43 +167,23 @@ angular.module('kifi')
         }
 
         function groupWidgetLibraries(libraries) {
-          // Libraries are divided into two possible groupings:
-          // (1) "Kept In" libraries followed by "My Libraries"; and
-          // (2) "Recent Libraries" followed by "Other Libraries".
-          //
-          // If there are any kept-in libraries, the first grouping is displayed.
```
2is10, line 174: This grouping logic does not match what the extension keeper does. I adjusted it to do so while I was updating how it retrieves library info.

kifi-backend/modules/shoebox/angular/src/keep/keepToLibraryWidget.js
```diff
@@ -306,9 +286,7 @@ angular.module('kifi')
           //
           // Group widget libraries.
           //
-          allLibraries = _.filter(libraryService.librarySummaries, function (lib) {
-            return lib.access !== 'read_only';
-          });
+          allLibraries = libraryService.getOwnInfos();
```
2is10, line 289: This is another example of eliminating all uses of `.access` on library infos (summaries).

kifi-backend/modules/shoebox/angular/src/keep/keepToLibraryWidget.styl
```diff
@@ -98,173 +98,171 @@
...
+  font-size: 12px
+  placeholder secondaryTextGray
 
-  .keep-to-library-create-general-error
-    padding: 10px
-    background-color: redColor
-    color: #fff
-    text-align: center
+.keep-to-library-create-name-error
+  pop-over 40px 18px redColor
```
stkem, line 224: `:`?

stkem:
Oh, wow. Bit more involved than I thought. Thanks!

kifi-backend/modules/shoebox/angular/src/keep/keepToLibraryWidget.styl
```diff
@@ -98,173 +98,171 @@
...
     color: #777
 
-  .library-select-list
-    max-height: 305px
-    overflow-x: hidden
-    overflow-y: auto
+.library-select-list
+  max-height: 305px
+  overflow-x: hidden
+  overflow-y: auto
```
2is10, line 104: mostly just removing unnecessary indentation here and below to simplify the generated CSS selectors (improves page performance) and so that I could use some of the constructs I needed, like `.kf-kept-to>&`. That change was to make sure that the Unkeep button only shows on libraries in the “Kept to” section. I encountered a corner case where it was showing up on a library in the Recent Libraries section, which felt quite wrong.

You can view this diff with `?w=` in the location bar URL to ignore whitespace only changes to this file.

kifi-backend/modules/shoebox/angular/src/keep/keepToLibraryWidget.tpl.html
```diff
@@ -9,36 +9,28 @@
...
       </ul>
 
-      <div ng-if="widgetKeptInLibraries.length && widgetMyLibraries.length" class="library-group-header my-libraries-group-header">My Libraries</div>
-      <ul ng-if="widgetKeptInLibraries.length && widgetMyLibraries.length">
-        <li ng-repeat="library in widgetMyLibraries" class="library-select-option"
-          ng-class="{'selected': library.selected, 'keptTo': library.keptTo}"
-          ng-mouseenter="onHover(library)" ng-mouseleave="onUnhover(library)"
-          ng-include="'keep/keepToLibraryWidgetLibraryItem.tpl.html'">
-        </li>
-      </ul>
```
2is10, line 27: The “My Libraries” section is really the same as the “Other Libraries” section. We just use the “My Libraries” heading instead when the “Recent Libraries” heading is not showing above it.

kifi-backend/modules/shoebox/angular/src/layout/main/MainCtrl.js
```diff
@@ -58,7 +58,7 @@ angular.module('kifi')
       fileInput.replaceWith(fileInput = fileInput.clone(true));
 
       // Display the Main Library as the default option.
-      $scope.importLibrary = _.find(libraryService.librarySummaries, { 'kind': 'system_main' });
+      $scope.importLibrary = libraryService.getSysMainInfo();
```
2is10, line 61: Using getters now whenever we need to access library infos (summaries) from `libraryService`. This will make refactoring easier next time we decide to change its internal representation.

kifi-backend/modules/shoebox/angular/src/libraries/libraryHeader.js
```diff
@@ -54,33 +54,6 @@ angular.module('kifi')
...
-          // Figure out whether this library is a library that the user has been invited to.
-          function getInvitePromise() {
-            var promise = null;
-
-            if (libraryService.invitedSummaries.length) {
-              promise = $q.when(libraryService.invitedSummaries);
-            } else {
-              promise = libraryService.fetchLibrarySummaries(true).then(function () {
-                return libraryService.invitedSummaries;
-              });
```
2is10, line 67: This was an extremely heavy way of determining whether the user has been invited to the library he/she is viewing. It was force-reloading library infos for all libraries a user owns, follows, and has been invited to. Instead we now just passively check `libraryService.getInvitedInfos()`. It works because we eagerly load invited library infos. Longer term I think the endpoint for loading the full library info should also return the invite info for the page’s invite banner.

kifi-backend/modules/shoebox/angular/src/keep/keepToLibraryWidget.styl
2is10, line 224: `pop-over` is apparently a [mixin](http://learnboost.github.io/stylus/docs/mixins.html) we define in common.styl. Using parentheses would probably make that clearer. None of us are very experienced with Stylus.

kifi-backend/modules/shoebox/angular/src/libraries/libraryService.js
```diff
@@ -5,84 +5,63 @@ angular.module('kifi')
...
 
-        util.replaceArrayInPlace(librarySummaries, libs);
-        util.replaceArrayInPlace(invitedSummaries, invites);
-
-        return res.data;
+    var libraryInfosClutch = new Clutch(function () {
+      return $http.get(routeService.getLibraryInfos).then(function (res) {
+        var libraries = res.data.libraries;
+        var following = res.data.following || _.remove(libraries, {access: 'read_only'});  // TODO: remove fallback when server sends "following"
+        libraries.concat(following).forEach(function (lib) { delete lib.access; });
```
2is10, line 25: I have a server change coming that will begin sending the `following` list separately. Then we can remove these two lines. Didn’t do it all at once to keep server backwards compatible with existing open web app sessions for a while.

2is10:
> Oh, wow. Bit more involved than I thought. Thanks!

Yeah, I probably spent more time than necessary on this, but the bug arose because of an inconsistent data situation, so I thought it would be best to fix the situation rather than just patch data coming from various sources to make it look alike. That felt clumsy and more likely to reintroduce bugs like this one you found.

Thanks for reviewing.

stkem:
Yeah, even if slightly overkill for the specific problem at hand, this was a good cleanup!

kifi-backend/modules/shoebox/angular/src/keep/keepToLibraryWidget.styl
stkem, line 224: ah, got it

PR #13930, opened 2015-02-25
sending libraries user is following down in a separate list

2is10:
and no longer mixing `"access"` into each `LibraryInfo`
@stkem

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
@@ -146,34 +146,43 @@ class LibraryController @Inject() (
   }
 
   def getLibrarySummariesByUser = UserAction { request =>
-    val (librariesWithMemberships, librariesWithInvites) = libraryCommander.getLibrariesByUser(request.userId)
-    // rule out invites that are not duplicate invites to same library (only show library invite with highest access)
-    val invitesToShow = librariesWithInvites.groupBy(x => x._2).map { lib =>
-      val invites = lib._2.unzip._1
-      val highestInvite = invites.sorted.last
-      (highestInvite, lib._1)
-    }.toSeq
+    val (libsWithMemberships, libsWithAllInvites) = libraryCommander.getLibrariesByUser(request.userId)
+    val libsWithInvites = for ((lib, invites) <- libsWithAllInvites.groupBy(_._2).mapValues(_.map(_._1))) yield {
+      (invites.sorted.last, lib) // only show one invite per library - the one with highest access
+    }
 
-    val libsFollowing = for ((mem, library) <- librariesWithMemberships) yield {
-      val (owner, numKeeps) = db.readOnlyMaster { implicit s =>
-        (basicUserRepo.load(library.ownerId), keepRepo.getCountByLibrary(library.id.get))
+    val (libInfosWithMemberships, libInfosWithInvites) = db.readOnlyMaster { implicit s =>
+      val basicUsers = basicUserRepo.loadAll((libsWithMemberships.map(_._2.ownerId) ++ libsWithInvites.map(_._2.ownerId)).toSet)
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
```diff
@@ -146,34 +146,43 @@ class LibraryController @Inject() (
...
-      val (owner, numKeeps) = db.readOnlyMaster { implicit s =>
-        (basicUserRepo.load(library.ownerId), keepRepo.getCountByLibrary(library.id.get))
-      }
-      val info = LibraryInfo.fromLibraryAndOwner(library, owner, numKeeps, None)
-      val memInfo = if (mem.lastViewed.nonEmpty) Json.obj("access" -> mem.access, "lastViewed" -> mem.lastViewed) else Json.obj("access" -> mem.access)
-      Json.toJson(info).as[JsObject] ++ memInfo
+    val (libsWithMemberships, libsWithAllInvites) = libraryCommander.getLibrariesByUser(request.userId)
+    val libsWithInvites = for ((lib, invites) <- libsWithAllInvites.groupBy(_._2).mapValues(_.map(_._1))) yield {
+      (invites.sorted.last, lib) // only show one invite per library - the one with highest access
     }
```
2is10, line 152: These four lines are equivalent to 149-155 above—just renaming and trying to improve readability.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
```diff
@@ -146,34 +146,43 @@ class LibraryController @Inject() (
...
-    val invitesToShow = librariesWithInvites.groupBy(x => x._2).map { lib =>
-      val invites = lib._2.unzip._1
-      val highestInvite = invites.sorted.last
-      (highestInvite, lib._1)
-    }.toSeq
-
-    val libsFollowing = for ((mem, library) <- librariesWithMemberships) yield {
-      val (owner, numKeeps) = db.readOnlyMaster { implicit s =>
-        (basicUserRepo.load(library.ownerId), keepRepo.getCountByLibrary(library.id.get))
-      }
```
2is10, line 160: This old code was opening a database session per library (owned, followed, and invited). New code below uses only one database session and uses `basicUserRepo.loadAll` to make a bulk memcache call.

2is10:
I’ll probably wait a couple of hours, at least, to merge this, to avoid breaking existing open long-lived browser sessions. In reality, that’s probably just Danny, but still worthwhile.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
stkem, line 160: :+1:

stkem:
:-) lgtm
Makes me wonder if we should have some kind of force reload build into kifi.com if your session is older than some amount of time?

2is10:
That’s a good idea. Or a way to signal old apps to reload just when we know we’re making a backwards incompatible change.

kifi-backend/modules/shoebox/test/com/keepit/controllers/website/LibraryControllerTest.scala
```diff
@@ -657,8 +658,7 @@ class LibraryControllerTest extends Specification with ShoeboxTestInjector {
                     |"numKeeps":0,
                     |"numFollowers":0,
                     |"kind":"user_created",
-                    |"inviter":{"id":"${user2.externalId}","firstName":"Baron","lastName":"B","pictureName":"0.jpg","username":"bhsu"},
-                    |"access":"read_insert"
+                    |"inviter":{"id":"${user2.externalId}","firstName":"Baron","lastName":"B","pictureName":"0.jpg","username":"bhsu"}
```
2is10, line 661: @andrewconner And this even had test coverage! Unfortunately, the inviter was also another library’s owner, which gave me a false sense of security.

PR #13931, opened 2015-02-25
make sure special libraries are not affected by scoring algorithms

yingjieMiao:
@stkem @ahsu1230 

Special library is exempted from complicated scoring algorithm, only if it has not been delivered yet. 
This should fix the persona library reco problem.

### DIFF
--- a/kifi-backend/modules/curator/app/com/keepit/curator/commanders/LibraryRecommendationGenerationCommander.scala
+++ b/kifi-backend/modules/curator/app/com/keepit/curator/commanders/LibraryRecommendationGenerationCommander.scala
@@ -66,10 +66,18 @@ class LibraryRecommendationGenerationCommander @Inject() (
     def scoreReco(reco: LibraryRecommendation) =
       LibraryRecoScore(scoringStrategy.scoreItem(reco.masterScore, reco.allScores, reco.delivered, reco.clicked, None, false, 0f), reco)
 
+    // MAKE DANNY HAPPY. REMOVE THIS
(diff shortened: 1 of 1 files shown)


yingjieMiao:
To be more precise, they are still scored, but after final sorting, they will be promoted to the top of the list.

ahsu1230:
thanks!

kifi-backend/modules/curator/app/com/keepit/curator/commanders/LibraryRecommendationGenerationCommander.scala
```diff
@@ -66,10 +66,18 @@ class LibraryRecommendationGenerationCommander @Inject() (
     def scoreReco(reco: LibraryRecommendation) =
       LibraryRecoScore(scoringStrategy.scoreItem(reco.masterScore, reco.allScores, reco.delivered, reco.clicked, None, false, 0f), reco)
 
+    // MAKE DANNY HAPPY. REMOVE THIS
+    def hackySort(scored: Seq[LibraryRecoScore]): Seq[LibraryRecoScore] = {
```
stkem, line 70: My new favorite sorting algorithm :-)

stkem:
:+1:

kifi-backend/modules/curator/app/com/keepit/curator/commanders/LibraryRecommendationGenerationCommander.scala
ahsu1230, line 70: we should call it `dannySort()`

PR #13935, opened 2015-02-25
perturbation for special libraries

yingjieMiao:
@stkem @ahsu1230 

Make sure libraries have unique scores. I am worrying any Map or groupBy operation will shadow the libraries with same score.


stkem:
:+1:

PR #13941, opened 2015-02-25
One more more static lib reco

stkem:
@yingjieMiao FYI


stkem:
It's a journalist we want to make feel good.

PR #13943, opened 2015-02-25
add persist personas endpoint (like mobile)

ahsu1230:
@stkem @yingjieMiao

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
@@ -842,13 +842,17 @@ class LibraryCommander @Inject() (
             val access = if (targetLib.ownerId != inviterId) LibraryAccess.READ_ONLY else inviteAccess
             recipient match {
               case Left(userId) =>
-                libraryMembershipRepo.getWithLibraryIdAndUserId(libraryId, userId) match {
-                  case Some(mem) if mem.access == access =>
-                    None
-                  case _ =>
-                    val newInvite = LibraryInvite(libraryId = libraryId, inviterId = inviterId, userId = Some(userId), access = access, message = msgOpt)
-                    val inviteeInfo = (Left(invitedBasicUsersById(userId)), access)
-                    Some(newInvite, inviteeInfo)
+                if (userId == targetLib.ownerId) {
+                  None
+                } else {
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -842,13 +842,17 @@ class LibraryCommander @Inject() (
...
             recipient match {
               case Left(userId) =>
-                libraryMembershipRepo.getWithLibraryIdAndUserId(libraryId, userId) match {
-                  case Some(mem) if mem.access == access =>
-                    None
-                  case _ =>
-                    val newInvite = LibraryInvite(libraryId = libraryId, inviterId = inviterId, userId = Some(userId), access = access, message = msgOpt)
-                    val inviteeInfo = (Left(invitedBasicUsersById(userId)), access)
-                    Some(newInvite, inviteeInfo)
+                if (userId == targetLib.ownerId) {
```
ahsu1230, line 845: prevent library invites to that library owner from persisting to database =P

kifi-backend/modules/shoebox/conf/shoeboxService.routes
```diff
@@ -405,6 +405,7 @@ GET     /site/user/settings         @com.keepit.controllers.website.UserControll
 POST    /site/user/settings         @com.keepit.controllers.website.UserController.setSettings()
 
 GET     /site/user/personas             @com.keepit.controllers.website.UserPersonaController.getAllPersonas()
+POST    /site/user/personas              @com.keepit.controllers.website.UserPersonaController.selectPersonas()
 POST    /site/user/personas/:name        @com.keepit.controllers.website.UserPersonaController.addPersona(name: String)
 DELETE  /site/user/personas/:name        @com.keepit.controllers.website.UserPersonaController.removePersona(name: String)
```
ahsu1230, line 410: keeping the other endpoints until we're sure whether they'll be used or not

ahsu1230:
the idea of this is on first sight of WAY, we want to persist personas only when "Done" is pressed, so libraries are not created /recos do not appear for unselected personas.

It will take a little longer for recos to appear, but thankfully the FTUE appears after the WAY (for extension users only) so it buys us some time.

ahsu1230:
merging so can work on FE

stkem:
:+1:

PR #13950, opened 2015-02-25
insta-scrolling to top of page upon navigation

2is10:
@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/app.js
+++ b/kifi-backend/modules/shoebox/angular/src/app.js
@@ -75,10 +75,10 @@ angular.module('kifi', [
 ])
 
 .controller('AppCtrl', [
-  '$scope', 'profileService', '$rootScope', 'libraryService', '$timeout', '$log',
-  'platformService', '$rootElement', '$analytics', '$location', 'util',
-  function ($scope, profileService, $rootScope, libraryService, $timeout, $log,
-      platformService, $rootElement, $analytics, $location, util) {
+  '$scope', '$rootScope', '$rootElement', '$window', '$timeout', '$log', '$analytics', '$location', '$state',
+  'profileService', 'platformService', 'libraryService',
+  function ($scope, $rootScope, $rootElement, $window, $timeout, $log, $analytics, $location, $state,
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/angular/src/app.js
```diff
@@ -94,21 +94,27 @@ angular.module('kifi', [
...
-      if (!$analytics.settings.pageTracking.autoTrackVirtualPages) {
-        $rootScope.$on('$stateChangeSuccess', function (event, toState) {
-          $scope.errorStatus = $scope.errorParams = null;
+      $rootScope.$on('$stateChangeSuccess', function (event, toState, toParams, fromState, fromParams) {
+        $scope.errorStatus = $scope.errorParams = null;
 
+        var toStateParts = toState.name.split('.');
+        if (fromState.name && (toStateParts[0] !== fromState.name.split('.')[0] || paramsDiffer(toStateParts[0], fromParams, toParams))) {
+          $window.document.body.scrollTop = 0;
+        }
```
2is10, line 103: The logic here is that we want to go to the top of the page if we’re going from one library to another or from one user profile to another, but not when navigating between sub-states of the _same_ top-level state.

kifi-backend/modules/shoebox/angular/src/search/search.js
```diff
@@ -78,9 +79,10 @@ angular.module('kifi')
 
       $timeout(function () {
         if (library) {
-          var pxToScroll = angular.element('.kf-lib-cols')[0].getBoundingClientRect().top - angular.element('.kf-lih,.kf-loh')[0].offsetHeight;
-          if (pxToScroll > 0) {
-            scrollDown(pxToScroll);
+          var cols = angular.element('.kf-lib-cols');
+          var header = angular.element('.kf-lih,.kf-loh');
+          if (cols.length && header.length) {
+            smoothScroll(cols[0].getBoundingClientRect().top - header[0].offsetHeight);
```
2is10, line 85: now also smooth scrolling _up_ to the first search result when necessary

PR #13953, opened 2015-02-25
3.3.46: refreshing library list later in guide to avoid race condition

2is10:
@ahsu1230 

I saw your pull request go by that delays the creation of persona libs until the Done button is clicked, which is right when the guide is triggered and the extension loads the library list. Delaying to avoid a race condition.


ahsu1230:
oh sorry, shoulda told you. thanks for the fix!

PR #13956, opened 2015-02-26
maintaining scroll position while navigating between user profile library tabs

2is10:
@andrewconner


andrewconner:
Great thanks. Simple.

PR #13958, opened 2015-02-26
Start of new signup flow

andrewconner:
Committing because it's been too long. Will continue with it.


2is10:
I think we want to use Google-hosted fonts where possible. Did we get to the bottom of the invisible text issue reported on Twitter a couple of weeks back?

andrewconner:
I think Google's fonts fixed it. I'll switch it over.

PR #13962, opened 2015-02-26
Accept normalization at 0.99 similarity

leogrim:


eishay:
:+1:

leogrim:
Note: looks like the similarity is often exactly 0.99, not sure where the truncation is happening but we may have been rejecting urls unwillingly.

PR #13965, opened 2015-02-26
hook up the connections profile list to a route

eishay:


2is10:
Nice! A few things:
- Responses will need to include the **viewer’s** connection status with each user.
- Need a way to load a list of connections given their IDs.
- To optimize page load time, the web app should be able to make one request and get the first page of connections, without knowing their IDs first. The complete list of IDs could also be returned in the same response, but it’s probably simpler (and faster for the viewer) for the web app to pre-load it separately, then remove the IDs it’s already showing.

eishay:
ok

PR #13974, opened 2015-02-26
Disable alerts on Rover cluster being too small

leogrim:
@2is10 @andrewconner @stkem


2is10:
:clap:

2is10:
@leogrim What needs to be deployed to stop the alerts?​

stkem:
All services that listen to Rover, I think. That should be just Shoebox.

PR #13975, opened 2015-02-26
Fixing no max

andrewconner:
@stkem


kifi-backend/modules/shoebox/app/com/keepit/admin/AdminBookmarkController.scala
```diff
@@ -313,4 +313,11 @@ class AdminBookmarksController @Inject() (
     }
   }
 
+  def www$youtube$com$watch$v$otCpCn0l4Wo(keepId: Id[Keep]) = UserAction {
+    db.readWrite { implicit session =>
```
yingjieMiao, line 317: what is this method???? 

```
www$youtube$com$watch$v$otCpCn0l4Wo
```

stkem:
:+1:

PR #13977, opened 2015-02-26
minor refactoring to make the profile connections usable

2is10:
@eishay

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/UserController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/UserController.scala
@@ -36,7 +36,7 @@ import play.api.libs.json._
 
 import scala.concurrent.Future
 import scala.concurrent.duration._
-import scala.util.{ Failure, Success }
+import scala.util.{ Failure, Success, Try }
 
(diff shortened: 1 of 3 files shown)


2is10:
Merging so I can begin using it. I’ll be happy to address any comments, or feel free to make alterations yourself.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/UserController.scala
```diff
@@ -631,6 +614,53 @@ class UserController @Inject() (
...
+          Try(userExtIds.split(',').map(ExternalId[User])) match {
+            case Success(userIds) =>
+              val users = db.readOnlyMaster { implicit s =>
+                userIds.map(userRepo.getOpt).flatten
+              }
+              Future.successful(Ok(Json.obj("users" -> users.map(BasicUser.fromUser))))
+            case _ =>
+              Future.successful(BadRequest("ids invalid"))
+          }
+        }
```
2is10, line 643: I realize this is essentially two endpoints in one, but I thought it made sense b/c they’re both pages of the same data.

PR #13979, opened 2015-02-26
optimize reco ingestion performance for new user

yingjieMiao:
@stkem @ahsu1230 

avoid unnecessary table look ups for new user. should improve speed quite a bit.


stkem:
:+1:

ahsu1230:
thanks for this! lgtm

PR #13981, opened 2015-02-26
more mobile notifications fields (social_user_joined)

ahsu1230:
@jaredpetker @andrewconner


jaredpetker:
LGTM thanks! Should be all we need for the moment I hope!

PR #13985, opened 2015-02-26
killing keepWhoService, defining a user pic filter

2is10:
cc @ahsu1230 @stkem @andrewconner


andrewconner:
Embracing the angular!

PR #13989, opened 2015-02-27
[Rover] Typed Article Structure

leogrim:
@stkem @andrewconner @eishay @2is10 @yingjieMiao 
More on the way, but you get the idea. Suggestions and critics welcome :)


kifi-backend/modules/common/app/com/keepit/rover/article/Article.scala
```diff
@@ -0,0 +1,154 @@
...
+  }
+}
+
+object ArticleKind {
+  val all: Set[ArticleKind[_ <: Article]] = CompanionTypeSystem[Article, ArticleKind[_ <: Article]]("A")
+  val byTypeCode: Map[String, ArticleKind[_ <: Article]] = {
+    require(all.size == all.map(_.typeCode).size, "Duplicate Article type codes.")
+    all.map { articleKind => articleKind.typeCode -> articleKind }.toMap
+  }
+}
```
yingjieMiao, line 50: `CompanionTypeSystem` stuff is cool, however,  I am afraid it's an overkill for this use case. we only have 2 types of content for now. I understand the number of types may grow over time. IMHO, trait or abstract class could just do the job. 

Personally I am not comfortable (or understand) this magic. If other engineers are fine, then I am good. Otherwise, this means risk for future maintenance.

kifi-backend/modules/common/app/com/keepit/rover/article/YoutubeTrack.scala
```diff
@@ -0,0 +1,19 @@
...
+case class YoutubeTrackInfo(
+    id: Int,
+    name: String,
+    langCode: Lang,
+    langOriginal: String,
+    langTranslated: String,
+    isDefault: Boolean,
+    kind: Option[String]) {
+  def isAutomatic = (kind == Some("asr"))
+}
```
yingjieMiao: Is 'asr' some industry standard code?

yingjieMiao: automatic speech recognition ?

kifi-backend/modules/common/app/com/keepit/rover/article/Article.scala
```diff
@@ -0,0 +1,154 @@
...
+    alternateUrls: Set[String],
+    shortUrl: Option[String],
+    // Video Info
+    videoTitle: Option[String],
+    videoDescription: String,
+    videoKeywords: Seq[String],
+    channel: String,
+    tracks: Seq[YoutubeTrack],
+    viewCount: Int) extends Article with NormalizationInfo {
+  type A = YoutubeArticle
```
yingjieMiao, line 91: 20 parameters for a constructor sounds a lot. maybe it could be organized in some sub case classes?

kifi-backend/modules/common/app/com/keepit/rover/article/Article.scala
yingjieMiao, line 50: Are we going to store different format of json files under one store?

leogrim, line 50: Yes, we already have 5+ types of articles, this is nowhere near done. The
purpose of this is to make it dead easy to add more specific articles,
without having to worry about storage, serialization across services and so
on. It provides simple APIs for generic consumption while storing
arbitrarily specific data, so it should be very flexible for future
evolutions (e.g. full page backup with history :)

yingjieMiao, line 50: OK. I understand your points. I need to rewrite my python scripts now. They directly talk to stores.

leogrim, line 50: No hurry, we'll have both working in parallel with a transition period.

kifi-backend/modules/common/app/com/keepit/rover/article/YoutubeTrack.scala
leogrim: Hum that's a good guess, never thought about it. No it's not industry standard, merely reverse engineering of the Youtube internal APIs.

kifi-backend/modules/common/app/com/keepit/rover/article/Article.scala
leogrim, line 91: Thanks, I had similar concerns, now they're confirmed ;)

leogrim:
@yingjieMiao I have largely refactored the article structure.
The separation between `ArticleContent` and `ArticleContext` is somewhat arbitrary, so it may be unnecessary. What do you think?

stkem:
lgtm

leogrim:
@stkem

stkem:
:+1:

PR #13994, opened 2015-02-27
mobile endpoints for setting notifs read

ahsu1230:
@stkem 
you can use it like so:
`/m/1/eliza/notifications/markAsRead?kind=system` marks only system notifications as read
`/m/1/eliza/notifications/markAsRead?kind=message` marks only message notifs as read
`/m/1/eliza/notifications/markAsRead` marks ALL notifs as read

cc: @jaredpetker @thassss @mrbrown14 @DerekBurch


andrewconner:
Thanks Aaron!

PR #13997, opened 2015-02-27
loading and caching user profile numbers

eishay:
@2is10


eishay:
just saw a conflict, pulling back, i'll fix the code first

kifi-backend/modules/common/app/com/keepit/social/BasicUser.scala
```diff
@@ -92,6 +92,15 @@ case class BasicUserUserIdKey(userId: Id[User]) extends Key[BasicUser] {
...
 
+case class BasicUserUserExternalIdKey(userId: ExternalId[User]) extends Key[BasicUser] {
+  override val version = 1
+  val namespace = "basic_user_ext_id"
+  def toKey(): String = userId.id
+}
+
+class BasicUserUserExternalIdCache(stats: CacheStatistics, accessLog: AccessLog, innermostPluginSettings: (FortyTwoCachePlugin, Duration), innerToOuterPluginSettings: (FortyTwoCachePlugin, Duration)*)
+  extends ImmutableJsonCacheImpl[BasicUserUserExternalIdKey, BasicUser](stats, accessLog, innermostPluginSettings, innerToOuterPluginSettings: _*)
+
```
2is10, line 103: I’m not sure we need this, since we have an `ExternalId[User]` → `User` cache already, and `BasicUser` is easy to create from `User`. I updated the new connections endpoint yesterday to use that cache. I’m not sure how much bigger the JSON is for `User` vs `BasicUser`. Feel free to add this cache if you feel it’s worth it.

2is10:
> just saw a conflict, pulling back, i'll fix the code first

@eishay I’m sorry to have surprised you. I did #13977 around lunchtime yesterday so I could begin using the endpoints. Your initial ones used internal IDs, which the webapp doesn’t have. I would have asked you before touching the code, but I didn’t see you for a couple of hours so figured you weren’t working on it and we probably wouldn’t conflict.

PR #14003, opened 2015-02-27
add library image into LibraryInfo and colors into library related notif...

ahsu1230:
...ications

@andrewconner cc: @mrbrown14

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/model/LibraryView.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/model/LibraryView.scala
@@ -97,13 +97,13 @@ case class LibraryInfo(
   shortDescription: Option[String],
   url: String,
   color: Option[LibraryColor], // system libraries have no color
+  image: Option[LibraryImageInfo],
   owner: BasicUser,
   numKeeps: Int,
   numFollowers: Int,
   kind: LibraryKind,
   lastKept: Option[DateTime],
-  inviter: Option[BasicUser],
-  image: Option[LibraryImageInfo])
+  inviter: Option[BasicUser])
 
 object LibraryInfo {
   implicit val format = (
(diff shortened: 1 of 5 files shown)


ahsu1230:
I was about to switch out `LibraryInfo` for `BasicLibraryCardInfo`, but I'm not comfortable doing that because it might break something on mobile side (esp older builds) since there are a few field differences (like `url` vs. `path`, `shortDescription` vs. `description`, etc.)
But would love to start thikning about refactoring sometime

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryView.scala
```diff
@@ -102,7 +102,8 @@ case class LibraryInfo(
   numFollowers: Int,
   kind: LibraryKind,
   lastKept: Option[DateTime],
-  inviter: Option[BasicUser])
+  inviter: Option[BasicUser],
+  image: Option[LibraryImageInfo])
```
2is10: How about putting `image` right after `color` as we do for the other library types?

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryView.scala
```diff
@@ -117,10 +118,11 @@ object LibraryInfo {
     (__ \ 'numFollowers).format[Int] and
     (__ \ 'kind).format[LibraryKind] and
     (__ \ 'lastKept).formatNullable[DateTime] and
-    (__ \ 'inviter).formatNullable[BasicUser]
+    (__ \ 'inviter).formatNullable[BasicUser] and
+    (__ \ 'image).formatNullable[LibraryImageInfo]
   )(LibraryInfo.apply, unlift(LibraryInfo.unapply))
 
-  def fromLibraryAndOwner(lib: Library, owner: BasicUser, keepCount: Int, inviter: Option[BasicUser])(implicit config: PublicIdConfiguration): LibraryInfo = {
+  def fromLibraryAndOwner(lib: Library, owner: BasicUser, keepCount: Int, inviter: Option[BasicUser] = None, image: Option[LibraryImage] = None)(implicit config: PublicIdConfiguration): LibraryInfo = {
```
2is10: Here, I think the argument ordering would make more sense if `image` came right after `lib` since it’s a core attribute of the library, and to keep the argument ordering here consistent with the property ordering that I suggested in my other comment.

Also, I don’t think `image` should have a default value of `None`. That suggests that a library’s image may or may not be included in a `LibraryInfo`, when in fact I think it should always be included if the represented library in fact has an image. Fields that are sometimes present and other times not are confusing and tend to introduce bugs. For example, I just fixed #13929 earlier this week.

2is10:
I’m excited to have library images available in more places. Thanks for working on this.

ahsu1230:
Thanks @2is10 for the review. Some of these places, I purposely left out including a library image in the response because the endpoint is already really slow and unused/doesn't really make sense (i.e. `getLibrarySummariesByUser` used for library list dropdown on website).
I think once we rethink that dropdown list & see where designs / product is going for library info, we should refactor =)

2is10:
:+1:

PR #14005, opened 2015-02-27
add statsd and timing log

yingjieMiao:


yingjieMiao:
@stkem @ahsu1230 

FYI, I tested two mobile sign ups (both with 3+ personas selected). Timing in the controller was 0.242 seconds and 0.413 seconds. 

Not too bad. (Considering it's about 100 database writes)

Any more latency we experience on mobile/site could be internal network latency or just wifi, or some other factors I am not aware of.

yingjieMiao:
Stress test: 

selected all 15 personas. Took 1.461 seconds.

yingjieMiao:
other latencies include a second call to curator and render all library/keep info. Since we are loading 10 libraries now, it potentially cost more time as before.

PR #14007, opened 2015-02-28
loading and caching user profile numbers

eishay:


kifi-backend/modules/shoebox/marketing
```diff
@@ -1 +1 @@
-Subproject commit 480d96a10ce2acca196e4ef21518f9d91c3601b3
+Subproject commit ca8dfe2ad8b4459fbc35c42707195223f1ab0a76
```
2is10, line 1: intentional?

2is10:
:+1: thanks!

PR #14010, opened 2015-02-28
preloading FB and Twitter APIs less eagerly since FB API affects tile

2is10:
Preloading the FB api on the recos page results in an `fb_root` div being appended to the page body—sometimes immediately after the guide has begun. As a result, the tile would move forward to the front of the stacking order, in front of both the facebook div and the guide. But we want it to stay behind the guide, hence this change.
cc @ahsu1230


andrewconner:
Hm, does it behave the same way? When I worked on it, if it wasn't loaded when the link was clicked, it wouldn't work reliably.

2is10:
Yes, it works reliably in my tests. Please let me know if you can make it fail—you’re an expert at that.

PR #14012, opened 2015-02-28
limiting access to scrape info update to a single thread on a single mac...

eishay:
...hine


eishay:
@andrewconner @leogrim

PR #14013, opened 2015-02-28
Cache on hot query

andrewconner:
@eishay


kifi-backend/modules/shoebox/app/com/keepit/common/cache/ShoeboxCacheModule.scala
```diff
@@ -343,6 +343,11 @@ case class ShoeboxCacheModule(cachePluginModules: CachePluginModule*) extends Ca
 
   @Singleton
   @Provides
+  def libraryMembershipCountByLibIdAndAccessCache(stats: CacheStatistics, accessLog: AccessLog, innerRepo: InMemoryCachePlugin, outerRepo: FortyTwoCachePlugin) =
+    new LibraryMembershipCountByLibIdAndAccessCache(stats, accessLog, (innerRepo, 2 minutes), (outerRepo, 7 days))
```
eishay, line 347: i would always do 30 days if you have a good cache invalidation

PR #14014, opened 2015-02-28
beginning of the public /:username/connections page

2is10:


2is10:
@eishay Nothing is linking to `/:username/connections` yet. Okay if I allow it to work publicly to allow the team to see and QA it?

eishay:
sure

PR #14015, opened 2015-02-28
disable url

eishay:


andrewconner:
Different from https://admin.kifi.com/admin/data/patterns ?

eishay:
i wasn't sure it worked

PR #14026, opened 2015-03-01
tmp disable twitter status url normalization update

yingjieMiao:
@eishay @leogrim @andrewconner 

tried to put a url rule in the admin. doesn't seem to stop these crazy links.


kifi-backend/modules/shoebox/app/com/keepit/normalizer/NormalizationService.scala
```diff
@@ -30,18 +30,26 @@ class NormalizationServiceImpl @Inject() (
     priorKnowledge: PriorNormalizationKnowledge,
     airbrake: AirbrakeNotifier) extends NormalizationService with Logging {
 
+  private val tmpDisable = """https:\/\/twitter\.com\/.*""".r
```
eishay, line 33: ok for the night but even for temp solution it seems a bit harsh. thought we'll disable a sub pattern

yingjieMiao, line 33: ah, you are right. out of my mind.

PR #14027, opened 2015-03-01
sub pattern

yingjieMiao:
@eishay

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/normalizer/NormalizationService.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/normalizer/NormalizationService.scala
@@ -30,7 +30,7 @@ class NormalizationServiceImpl @Inject() (
     priorKnowledge: PriorNormalizationKnowledge,
     airbrake: AirbrakeNotifier) extends NormalizationService with Logging {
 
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/app/com/keepit/normalizer/NormalizationService.scala
```diff
@@ -30,7 +30,7 @@ class NormalizationServiceImpl @Inject() (
     priorKnowledge: PriorNormalizationKnowledge,
     airbrake: AirbrakeNotifier) extends NormalizationService with Logging {
 
-  private val tmpDisable = """https:\/\/twitter\.com\/.*""".r
+  private val tmpDisable = """https:\/\/twitter\.com\/.*\/status\/.*""".r
```
eishay, line 33: the problematic urls look like this:
https://twitter.com/aworks/status/552952502879268864?screen_name=aworks&amp;screen_name=aworks&amp;lang=sv&amp;id=552952502879268864&amp;id=552952502879268864
Where we actually want only this:
https://twitter.com/aworks/status/552952502879268864
Thought the regex would look something like 

```
https:\/\/twitter.com\/.*\/status/.*\?screen_name=.*
```

yingjieMiao:
true. your is less restrictive.

PR #14028, opened 2015-03-01
trimming memberships sent to search in DPL to let the ingestion happen, this is a tmp solution.

eishay:


eishay:
@leogrim

leogrim:
To be clear, all followers after the 1000th won't be able to search the library. This may look bad since the popular libraries are the ones we push for people to test the product with. Maybe we should take out your Ted library from the featured libraries for now.

Also this doesn't change the load in terms of calls to Shoebox does it? The ingestion rate to Shoebox is still driven by the sequence number update frequency. Then we go to the database only once anyway, whether we get all followers or not. 

I think slowing down the indexing plugin is the best solution for now (if necessary, we can slow it down even more than you did), it naturally leads to a larger batching of updates. More indexing latency but we preserve correctness / eventual consistency.

The best long term solution I can think of is to have a separate index for library memberships. I don't think that's too much work, ~1 day or so.

eishay:
You're right, this pull request was a temporary solution for couple hours until i came up with something a bit better. see
https://github.com/kifi/keepit/pull/14030
https://github.com/kifi/keepit/pull/14029
One is doing slower ingestion like you suggested and the other is returning only membership ids and having search load the full data from memcache. The solution is still sub optimal, working on few things that may help. Anything that would help us maintain large libraries is very important.

PR #14029, opened 2015-03-01
using cache for library membership ingestion.

eishay:


eishay:
@leogrim its still sub optimal, but should at least be correct and not timeout on the shoebox

PR #14033, opened 2015-03-02
splitting shoebox to "offline" and "online" capability services where on...

eishay:
...line is hooked with the ELB and "offline" is not.

Offline services "Data Pipeline" using a client side routing strategy calls and online services don't (unless all offline are gone)


eishay:
@andrewconner @stkem @leogrim @yingjieMiao

PR #14034, opened 2015-03-02
Introduce LibraryMembership Index

leogrim:
@eishay @andrewconner I may be able to pull this off tonight
Once this index is built, I'll switch search over and we can stop reindexing all library memberships with every new one.

### DIFF
--- a/kifi-backend/modules/search/app/com/keepit/search/index/graph/library/membership/LibraryMembershipIndexable.scala
+++ b/kifi-backend/modules/search/app/com/keepit/search/index/graph/library/membership/LibraryMembershipIndexable.scala
@@ -1,0 +1,57 @@
+package com.keepit.search.index.graph.library.membership
+
+import com.keepit.common.db.Id
+import com.keepit.model.view.LibraryMembershipView
+import com.keepit.model._
+import com.keepit.search.index._
+import com.keepit.search.index.graph.library.LibraryIndexable
+import com.keepit.search.util.LongArraySet
+import org.apache.lucene.index.Term
+
+object LibraryMembershipFields {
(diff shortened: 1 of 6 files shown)


eishay:
great!

yingjieMiao:
I assume we need to change the index query? test?

yingjieMiao:
I am not clear about the search performance impact of index structure change. let's keep that in mind.

leogrim:
No you'll see my next pull request later tonight, pretty minimal change as far as queries are concerned :)

andrewconner:
Amazing job, that was fast

kifi-backend/modules/search/app/com/keepit/search/index/graph/library/membership/LibraryMembershipIndexable.scala
```diff
@@ -0,0 +1,57 @@
...
+
+    doc.add(buildKeywordField(userField, membership.userId.id.toString))
+    doc.add(buildIdValueField(userIdField, membership.userId))
+    if (membership.showInSearch) {
+      doc.add(buildKeywordField(searcherField, membership.userId.id.toString))
+    }
+
+    doc.add(buildKeywordField(libraryField, membership.libraryId.id.toString))
+    doc.add(buildIdValueField(libraryIdField, membership.libraryId))
+
```
yingjieMiao, line 54: what are the differences of these two fields? are they going to be queried differently?

leogrim, line 54: One is indexed, the other one is stored.

PR #14040, opened 2015-03-02
User LibraryMembership Index instead of LibraryIndex

leogrim:

### DIFF
--- a/kifi-backend/modules/search/app/com/keepit/search/index/graph/library/LibraryIndexable.scala
+++ b/kifi-backend/modules/search/app/com/keepit/search/index/graph/library/LibraryIndexable.scala
@@ -1,7 +1,6 @@
 package com.keepit.search.index.graph.library
 
 import com.keepit.common.db.Id
-import com.keepit.model.view.LibraryMembershipView
 import com.keepit.model._
 import com.keepit.search.index._
 import com.keepit.search.LangDetector
@@ -19,8 +18,6 @@ object LibraryFields {
   val kindField = "k"
   val ownerField = "o"
   val ownerIdField = "oid"
-  val usersField = "u"
-  val allUsersField = "a"
   val allUsersCountField = "ac"
   val recordField = "rec"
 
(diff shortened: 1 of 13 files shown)


kifi-backend/modules/shoebox/marketing
```diff
@@ -1 +1 @@
-Subproject commit 480d96a10ce2acca196e4ef21518f9d91c3601b3
+Subproject commit ca8dfe2ad8b4459fbc35c42707195223f1ab0a76
```
yingjieMiao, line 1: marketing commit

kifi-backend/modules/search/app/com/keepit/search/index/graph/library/LibraryIndexable.scala
```diff
@@ -95,6 +92,7 @@ object LibraryIndexable {
     librarySearcher.getStringDocValue(LibraryFields.nameValueField, libId)
   }
 
+  // todo(Léo): we could use LibraryMembership Index to get this count instead
   def getMemberCount(librarySearcher: Searcher, libId: Long): Option[Long] = {
```
yingjieMiao, line 96: about the todo: isn't current impl more efficient?

yingjieMiao:
:+1:

kifi-backend/modules/shoebox/marketing
leogrim, line 1: Oh, I'm not sure what I'm supposed to do here?

kifi-backend/modules/search/app/com/keepit/search/index/graph/library/LibraryIndexable.scala
leogrim, line 96: Maybe, but Lucene probably knows the length of its postings lists so it may be just as fast. And then we no longer need memberCount in the library index, which may enable us to remove `memberCount` from `Library`. Currently we update `Library` every time we update `LibraryMembership`, I meant to remove search's reliance on that assumption.

PR #14054, opened 2015-03-02
not looking up library owner friend status for every user profile library card

2is10:
I don’t remember why I added this in #12989 but mobile and website don’t seem to be using it.

@andrewconner cc @eishay


2is10:
Hopefully this will reduce load a little bit from logged-in user profile page views.

PR #14057, opened 2015-03-02
num followers + num connections to profile endpoint

ahsu1230:
@2is10

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/model/LibraryMembership.scala
+++ b/kifi-backend/modules/common/app/com/keepit/model/LibraryMembership.scala
@@ -107,7 +107,7 @@ class LibraryMembershipCountCache(stats: CacheStatistics, accessLog: AccessLog, 
   extends PrimitiveCacheImpl[LibraryMembershipCountKey, Int](stats, accessLog, innermostPluginSettings, innerToOuterPluginSettings: _*)
 
 case class FollowersCountKey(userId: Id[User]) extends Key[Int] {
-  override val version = 1
+  override val version = 2
   val namespace = "followers_count"
   def toKey(): String = s"$userId"
 }
--- a/kifi-backend/modules/shoebox/app/com/keepit/model/LibraryMembershipRepo.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/model/LibraryMembershipRepo.scala
@@ -43,7 +43,9 @@ trait LibraryMembershipRepo extends Repo[LibraryMembership] with RepoWithDelete[
   def mostMembersSinceForUser(count: Int, since: DateTime, ownerId: Id[User])(implicit session: RSession): Seq[(Id[Library], Int)]
   def countWithUserIdAndAccess(userId: Id[User], access: LibraryAccess)(implicit session: RSession): Int
   def countsWithUserIdAndAccesses(userId: Id[User], accesses: Set[LibraryAccess])(implicit session: RSession): Map[LibraryAccess, Int]
+  def countFollowersFromAnonymous(userId: Id[User])(implicit session: RSession): Int
   def countFollowersWithOwnerId(ownerId: Id[User])(implicit session: RSession): Int
+  def countFollowersForOtherUser(ownerId: Id[User], viewerId: Id[User])(implicit session: RSession): Int
 }
 
 @Singleton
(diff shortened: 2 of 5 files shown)


kifi-backend/modules/common/app/com/keepit/model/LibraryMembership.scala
```diff
@@ -107,7 +107,7 @@ class LibraryMembershipCountCache(stats: CacheStatistics, accessLog: AccessLog,
   extends PrimitiveCacheImpl[LibraryMembershipCountKey, Int](stats, accessLog, innermostPluginSettings, innerToOuterPluginSettings: _*)
 
 case class FollowersCountKey(userId: Id[User]) extends Key[Int] {
-  override val version = 1
+  override val version = 2
```
ahsu1230, line 110: the query was changed, so I think we have to increase the version so old values are not retrieved

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryMembershipRepo.scala
```diff
@@ -322,10 +324,21 @@ class LibraryMembershipRepoImpl @Inject() (
   def countFollowersWithOwnerId(ownerId: Id[User])(implicit session: RSession): Int = {
     followersCountCache.getOrElse(FollowersCountKey(ownerId)) {
       import StaticQuery.interpolation
-      sql"select count(distinct lm.user_id) from library_membership lm, library lib where lm.library_id = lib.id and lib.owner_id = $ownerId and lib.state = 'active' and lm.state = 'active'".as[Int].firstOption.getOrElse(0)
+      sql"select count(distinct lm.user_id) from library_membership lm, library lib where lm.library_id = lib.id and lib.owner_id = $ownerId and lib.state = 'active' and lm.access != 'owner' and lm.state = 'active'".as[Int].firstOption.getOrElse(0)
```
ahsu1230, line 327: before, it used to also count owners as followers.

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryMembershipRepo.scala
```diff
@@ -322,10 +324,21 @@ class LibraryMembershipRepoImpl @Inject() (
...
     }
   }
 
+  def countFollowersFromAnonymous(ownerId: Id[User])(implicit session: RSession): Int = {
+    import StaticQuery.interpolation
+    sql"select count(distinct lm.user_id) from library_membership lm, library lib where lm.library_id = lib.id and lib.owner_id = $ownerId and lib.state = 'active' and lm.access != 'owner' and lm.state = 'active' and lib.visibility = 'published'".as[Int].firstOption.getOrElse(0)
+
+  }
```
2is10, line 335: The anonymous viewer count ought to be cached too, yes?

It would require less space, I think, if we cached both counts in a single memcache entry, but it’s up to you. Both can be read from the database at once using this SQL query (substitute `$ownerId` for the `3`):

``` sql
mysql> select sum(any), sum(published)
       from (select 1 any, max(lib.visibility = 'published') published
             from library_membership lm, library lib
             where lm.library_id = lib.id
               and lib.owner_id = 3
               and lib.state = 'active'
               and lm.access != 'owner'
               and lm.state = 'active'
             group by lm.user_id) booleans;
+----------+----------------+
| sum(any) | sum(published) |
+----------+----------------+
|      128 |            125 |
+----------+----------------+
```

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryMembershipRepo.scala
```diff
@@ -43,7 +43,9 @@ trait LibraryMembershipRepo extends Repo[LibraryMembership] with RepoWithDelete[
   def mostMembersSinceForUser(count: Int, since: DateTime, ownerId: Id[User])(implicit session: RSession): Seq[(Id[Library], Int)]
   def countWithUserIdAndAccess(userId: Id[User], access: LibraryAccess)(implicit session: RSession): Int
   def countsWithUserIdAndAccesses(userId: Id[User], accesses: Set[LibraryAccess])(implicit session: RSession): Map[LibraryAccess, Int]
+  def countFollowersFromAnonymous(userId: Id[User])(implicit session: RSession): Int
   def countFollowersWithOwnerId(ownerId: Id[User])(implicit session: RSession): Int
+  def countFollowersForOtherUser(ownerId: Id[User], viewerId: Id[User])(implicit session: RSession): Int
```
2is10, line 48: To reduce the risk of misuse, how about renaming `countFollowersWithOwnerId` → `countFollowersForOwner`?

2is10:
:+1: Thanks!

PR #14065, opened 2015-03-02
profile followers endpoint

ahsu1230:
@2is10


ahsu1230:
just saw the change for `loadFullConnectionsUser()` - change accordingly! =D

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/UserController.scala
```diff
@@ -77,7 +77,7 @@ class UserController @Inject() (
       //global or mutual
       val libCount = viewer.map(u => libraryRepo.countLibrariesForOtherUser(ownerId, u)).getOrElse(libraryRepo.countLibrariesOfUserFromAnonymous(ownerId)) //not cached
       //global
-      val followersCount = libraryMembershipRepo.countFollowersWithOwnerId(ownerId) //cached
+      val followersCount = libraryMembershipRepo.countFollowersForOwner(ownerId) //cached
```
2is10: Which method we use here should depend on who the viewer is, no?

ahsu1230: yup you're right! was just about to fix that!

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/UserController.scala
```diff
@@ -135,8 +135,8 @@ class UserController @Inject() (
         log.warn(s"can't find username ${username.value}")
         Future.successful(NotFound(s"username ${username.value}"))
       case Some(user) =>
-        val viewerId = request.userIdOpt.getOrElse(user.id.get)
-        userConnectionsCommander.getConnectionsSortedByRelationship(viewerId, user.id.get) map { connections =>
+        val viewerIdOpt = request.userIdOpt.getOrElse(user.id.get)
+        userConnectionsCommander.getConnectionsSortedByRelationship(viewerIdOpt, user.id.get) map { connections =>
```
2is10, line 148: The value you’ve renamed is not an `Option`. Perhaps you meant:

``` scala
val viewerIdOpt = request.userIdOpt
userConnectionsCommander.getConn...(viewerIdOpt.orElse(user.id), user.id.get) ...
```

ahsu1230:
uhoh looks like I messed up somewhere with this branch

2is10:
Eishay just committed a cache that may be helpful in #14069.

PR #14071, opened 2015-03-03
profile followers endpoint

ahsu1230:
@2is10 (cf. #14065 which was merging to the wrong branch)


kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -1455,6 +1455,17 @@ class LibraryCommander @Inject() (
     }
   }
 
+  def getFollowersByViewer(userId: Id[User], viewer: Option[Id[User]]): Seq[Id[User]] = db.readOnlyMaster { implicit s =>
```
2is10, line 1458: Let’s please avoid opening a database connection on the same line as the method declaration b/c it’s important but really easy for readers to miss over there.

ahsu1230, line 1458: sounds good, my bad.

PR #14077, opened 2015-03-03
remove unnecessary save

yingjieMiao:
@andrewconner @eishay @leogrim


leogrim:
Nice!

leogrim:
As I mentioned on Slack, the presence (and maintenance) of 'memberCount' on
'Library' means that we're still broadcasting 'Library' updates on every
new 'LibraryMembership' (it's probably happening at the commander level).
If we can get rid of 'memberCount' (I can do away with it in Search using
the new index from last night), we may further reduce load and traffic from
there.

PR #14078, opened 2015-03-03
refactoring user profile connections/followers for fewer requests, shared code

2is10:
This puts the list of ids in the original response for the first page of connections/followers.
It also combines the endpoints for loading more connections or followers by ID.
@ahsu1230

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/UserController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/UserController.scala
@@ -8,6 +8,7 @@ import com.keepit.commanders.emails.EmailSenderProvider
 import com.keepit.commanders.{ ConnectionInfo, _ }
 import com.keepit.common.controller._
 import com.keepit.common.db.slick._
+import com.keepit.common.db.slick.DBSession.RSession
 import com.keepit.common.db.{ ExternalId, Id }
 import com.keepit.common.healthcheck.AirbrakeNotifier
 import com.keepit.common.http._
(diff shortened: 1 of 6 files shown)


kifi-backend/modules/shoebox/app/com/keepit/controllers/website/UserController.scala
```diff
@@ -71,132 +72,99 @@ class UserController @Inject() (
     libraryMembershipRepo: LibraryMembershipRepo,
     fortytwoConfig: FortyTwoConfig) extends UserActions with ShoeboxServiceController {
 
-  private def loadConnectionUsers(userIds: Set[Id[User]], viewerIdOpt: Option[Id[User]], friendIds: Set[Id[User]] = Set.empty) = {
```
2is10, line 74: Passing `friendStatusCommander.augmentUsers` an empty set of friend IDs would be incorrect if the user is in fact friends with some of the users. There’s another method by the same name that doesn’t take a set of friend IDs, and it looks them up. That’s the one we need to call when we don’t have a friend ID set handy. Sorry, I suppose the `FriendStatusCommander` API could be clearer.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/UserController.scala
```diff
@@ -71,132 +72,99 @@ class UserController @Inject() (
...
-      //mutual
-      viewerIdOpt.map { viewerId =>
-        val followingLibCount = libraryRepo.countLibrariesOfOwnerUserFollow(userId, viewerId) //not cached
-        val mutualConnectionCount = userConnectionRepo.getMutualConnectionCount(userId, viewerId) //cached
-        jsonWithGlobalCounts +
-          ("mConnections" -> JsNumber(mutualConnectionCount)) +
-          ("mLibraries" -> JsNumber(followingLibCount))
-      } getOrElse jsonWithGlobalCounts
-    }
-  }
```
2is10, line 109: I didn’t change this method at all except to use an existing DB connection. You can view the diff with `?w=` in the location bar to verify.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/UserController.scala
ahsu1230, line 74: Ah, I see - sorry, was confused =D

ahsu1230:
Thanks Jared, I think it looks good.
I think we'll probably want to start thinking about how we should cache follower counts or followers, which might be a bit trickier since it depends on the viewer?

ahsu1230:
also what do you think about putting these profile endpoints into a `ProfileController`?
I think `UserController` makes sense to only contain user settings and preferences while `LibraryController` has more specific library actions

`ProfileController` often depends on who's viewing (anonymous, friend, owner, etc.). For followers & libraries at least, it contains similar logic.
Just a thought for when things settle down and we have time to refactor.

2is10:
> also what do you think about putting these profile endpoints into a ProfileController?

I had the same idea today. Great minds think alike. :smile:

PR #14083, opened 2015-03-03
3.3.48: guide step 0 now tolerates wider variety of page thumbnail sizes

2is10:
cc @ahsu1230


ahsu1230:
thanks! can't wait to wrap up everything!

PR #14093, opened 2015-03-03
optionally suppress tracking

yingjieMiao:
@ahsu1230 @stkem 

the parameter is only useful (and necessary) for site FTUE. mobile doesn't need this. 

as a by product, I noticed a typo. my bad. wondering how it worked.


stkem:
:+1:

PR #14095, opened 2015-03-03
adding connection count to each mutual friend

2is10:
@eishay

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/UserCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/UserCommander.scala
@@ -701,18 +701,22 @@ class UserCommander @Inject() (
         }
 
         val mutualFriends = recommendedUsers.map { recommendedUserId =>
-          recommendedUserId -> (friends.get(userId).getOrElse(Set.empty) intersect friends.get(recommendedUserId).getOrElse(Set.empty))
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/UserCommander.scala
```diff
@@ -701,21 +701,25 @@ class UserCommander @Inject() (
...
         FriendRecommendations(basicUsers, mutualFriendConnectionCounts, recommendedUsers, mutualFriends)
       }
     }
   }
 
+  def loadBasicUsersAndConnectionCounts(idsForUsers: Set[Id[User]], idsForCounts: Set[Id[User]]): (Map[Id[User], BasicUser], Map[Id[User], Int]) = {
+    db.readOnlyReplica { implicit session =>
+      val basicUsers = basicUserRepo.loadAll(idsForUsers)
+      val connectionCounts = userConnectionRepo.getConnectionCounts(idsForCounts)
+      (basicUsers, connectionCounts)
```
2is10, line 719: I switched the existing code above (for the People You May Know widget) to the bulk call `userConnectionRepo.getConnectionCounts`, which still uses the cache, and factored out the DB session for reuse. Not sure how reusable this method will be, but we have two callers now.

PR #14100, opened 2015-03-03
Throtteling Scraper calls to Shoebox

stkem:
So when there is a flood for some reason, it’s scraper that will blow up instead of shoebox.
Good Idea/Not Good Idea? Should hopefully prevent most of these remaining timeouts we are seeing.

@andrewconner @eishay


andrewconner:
Not sure this is the best. Scraper makes a request, it waits on the reactive lock, scraper times out, so the scrape task goes back to the queue, the lock frees up, so the task gets run. We end up duplicating a lot of work, and causing online scrape tasks to fail.

We should slow down scraper if we're not able to keep up.

edit: Bad feedback, didn't read it well.

andrewconner:
OH I misread it, sorry. This is the client. Let me think a bit more.

andrewconner:
Yeah this is probably good. We should tune the parallelism factor, taking into account we have these offline machines. Could be best rate is actually lower.

andrewconner:
Talked with @eishay. I actually think this should be baked into all service client calls.

i.e., `call` has an added parameter, `maxParallelism`, which defaults to a constant per service.

``` scala
trait ParallelismSetting { def c: Double }
case object NoLimit extends ParallelismSetting  { val c = 999.0 }
case class HighLimit extends ParallelismSetting { val c = 4.0 }
case class DefaultLimit extends ParallelismSetting { val c = 2.0 }
case class LowLimit extends ParallelismSetting { val c = 1.0 }
case class CustomLimit(val c: Double) extends ParallelismSetting
```

We could make it `c * remoteCpusAvailable`. In effect, it's a lot higher because there's multiple of each service, but this does a few things:
1. Rate limits out-of-control services. If scraper starts moving too fast, it gets slowed down.
2. Allows back pressure to build, so clients deal with it.
3. Gives us insight into the spikes we have. We should collect metrics on the queue size over time.

Thoughts? @leogrim @yingjieMiao

kifi-backend/modules/common/app/com/keepit/model/Persona.scala
```diff
@@ -85,7 +85,7 @@ object Persona {
     PersonaName.ARTIST -> PersonaKeep(
       title = "Psychedelic Science",
       url = "https://www.ted.com/talks/fabian_oefner_psychedelic_science",
-      image = PersonaKeepImageInfo("//d1dwdv9wd966qu.cloudfront.net/img/guide/louvre_480x510.06af404.png", 480 ,510),
+      image = PersonaKeepImageInfo("//d1dwdv9wd966qu.cloudfront.net/img/guide/louvre_480x510.06af404.png", 480, 510),
```
stkem: Somebody didn't run scalariform

stkem:
Updated with slightly better abstraction. Feel free to merge.

PR #14109, opened 2015-03-04
	sigh more image swaps

ahsu1230:


kifi-backend/modules/common/app/com/keepit/model/Persona.scala
```diff
@@ -58,7 +58,7 @@ object Persona {
     PersonaName.DEVELOPER -> PersonaKeep(
       title = "Good Tech Lead, Bad Tech Lead",
       url = "https://medium.com/@jliszka/good-tech-lead-bad-tech-lead-948b2b806d86",
-      image = PersonaKeepImageInfo("//d1dwdv9wd966qu.cloudfront.net/img/guide/techlead_480x446.27a8c4a.png", 480, 446),
+      image = PersonaKeepImageInfo("//d1dwdv9wd966qu.cloudfront.net/img/guide/techlead_960×892.57da042.jpg", 480, 446),
```
2is10, line 61: These are 4x images intentionally?

ahsu1230, line 61: yea we tried `480x446` but apparently the images came out blurry

PR #14111, opened 2015-03-04
signup / login links on top right corner

ahsu1230:
@andrewconner


kifi-backend/modules/shoebox/public/ml/css/authentication.css
```diff
@@ -142,6 +142,21 @@ p {
...
+  height: 100%;
+  margin: 0 auto;
+}
+
+.header-link {
+  position: absolute;
+  right: 0px;
+  top: 50%;
+  transform: translateY(-50%);
+  visibility: hidden;
```
ahsu1230: start off as hidden

kifi-backend/modules/shoebox/public/ml/js/authentication.js
```diff
@@ -1,7 +1,13 @@
 $(function() {
 
-  //Simple demonstration of error messages/animations
+  // show links on /signup and /login
+  if (window.location.pathname === "/new/login") {
+    $('.header-link.signup')[0].style.visibility = 'visible';
+  } else if (window.location.pathname === '/new/signup') {
+    $('.header-link.login')[0].style.visibility = 'visible';
+  }
```
ahsu1230: depending on path, make em visible

andrewconner: It may not always be that simple. The template itself could take a flag, pass it to layout, and layout will know how to render it.

andrewconner:
Thanks!

kifi-backend/modules/shoebox/app/views/minimal/layout.scala.html
```diff
@@ -36,7 +37,21 @@
...
+        @if(headerLink == "login") {
+          <div class="header-link login">
+            Already a member?
+            <a href="/login">Log In</a>
+          </div>
+        }
+        @if(headerLink == "signup") {
+          <div class="header-link signup">
+            Not a member yet?
+            <a href="/signup">Signup</a>
```
2is10, line 51: Signup → Sign up

It’s two words when used as a verb phrase (like “Log In” above), one word or hyphenated when used as a noun or adjective (e.g. “our sign-up flow”, “I completed signup”).

PR #14112, opened 2015-03-04
update ordering of followers and connections tabs

eishay:
@2is10


kifi-backend/modules/shoebox/app/com/keepit/commanders/UserProfileCommander.scala
```diff
@@ -0,0 +1,127 @@
...
+}
+
+class UserConnectionRelationshipCache(stats: CacheStatistics, accessLog: AccessLog, inner: (FortyTwoCachePlugin, Duration), outer: (FortyTwoCachePlugin, Duration)*)
+  extends JsonCacheImpl[UserConnectionRelationshipKey, Seq[ConnectedUserId]](stats, accessLog, inner, outer: _*)
+
+case class UserFollowerRelationshipKey(viewerIdOpt: Option[Id[User]], ownerId: Id[User]) extends Key[Seq[FollowerUserId]] {
+  val namespace = "user_fol_rel"
+  override val version = 1
+  def toKey(): String = ownerId.id.toString + ":" + viewerIdOpt.map(_.id.toString).getOrElse("ANON")
+}
```
2is10, line 41: This key makes the viewer optional, but `UserConnectionRelationshipKey` above doesn’t. Why the difference?

kifi-backend/modules/shoebox/app/com/keepit/commanders/UserProfileCommander.scala
```diff
@@ -0,0 +1,127 @@
...
+
+case class FollowerUserId(userId: Id[User], connected: Boolean)
+object FollowerUserId {
+  implicit val formatter = Json.format[FollowerUserId]
+}
+
+case class UserConnectionRelationshipKey(viewerId: Id[User], ownerId: Id[User]) extends Key[Seq[ConnectedUserId]] {
+  val namespace = "user_con_rel"
+  override val version = 2
```
2is10, line 30: I think we need to increment this version now that the sort ordering is fixed.

eishay, line 30: will do

kifi-backend/modules/shoebox/app/com/keepit/commanders/UserProfileCommander.scala
eishay, line 41: Because the default viewer for `UserConnectionRelationshipKey` is "self" where it can't be true for the followers cache. It may be more readable to have it optional.

kifi-backend/modules/shoebox/marketing
```diff
@@ -1 +1 @@
-Subproject commit 480d96a10ce2acca196e4ef21518f9d91c3601b3
+Subproject commit 46f4fb1c350f3da8f6a153d1640eaac0a71d174c
```
2is10, line 1: intentional?

kifi-backend/modules/shoebox/test/com/keepit/commanders/UserConnectionsCommanderTest.scala
```diff
@@ -97,11 +99,11 @@ class UserConnectionsCommanderTest extends Specification with ShoeboxTestInjecto
       inject[FakeGraphServiceClientImpl].setSociallyRelatedEntities(viewer.id.get, relationship)
       Await.result(inject[FakeGraphServiceClientImpl].getSociallyRelatedEntities(viewer.id.get), Duration.Inf).get === relationship
       Await.result(inject[GraphServiceClient].getSociallyRelatedEntities(viewer.id.get), Duration.Inf).get === relationship
-      val commander = inject[UserConnectionsCommander]
+      val commander = inject[UserProfileCommander]
```
2is10, line 102: Move this test case to UserProfileCommanderTest.scala?

kifi-backend/modules/shoebox/marketing
eishay, line 1: never

kifi-backend/modules/shoebox/app/com/keepit/commanders/UserProfileCommander.scala
2is10, line 41: Hmm, I still don’t understand why the default viewer can’t be handled the same way for both.

kifi-backend/modules/shoebox/marketing
2is10, line 1: fixed in #14120

PR #14114, opened 2015-03-04
Introduce Rover DB + ArticleInfo

leogrim:
@stkem @andrewconner 
Not fully happy with my naming, especially `fetchSeq: Option[SequenceNumber[ArticleInfo]]`, perhaps not the best type either.


kifi-backend/modules/common/conf/evolutions/shoebox/300.sql
```diff
@@ -0,0 +1,33 @@
...
+    state varchar(20) NOT NULL,
+    seq bigint(20) NOT NULL,
+    uri_id bigint(20) NOT NULL,
+    kind varchar(64) NOT NULL,
+    url varchar(3072) NOT NULL,
+    fetch_seq bigint(20) NULL,
+    last_queued_at datetime NULL,
+    last_fetched_at datetime NULL,
+    next_fetch_at datetime NULL,
+    fetch_interval double NOT NULL,
```
yingjieMiao: how is this `fetch_interval` calculated? Based on signature? Do we need double precision here?

yingjieMiao: is `fetch_seq` derived from other seqNum? do we need an index for that?

yingjieMiao: are we going to maintain version number in this table? or elsewhere? or at all?

leogrim: `fetch_interval` will work like the old one, though how we pick it is independent of how we store it :)
`fetch_seq` is incremented every time a new article is uploaded to S3, and will be used in the S3 key
Which version number are you thinking of? The `ArticleKind` version? You're right, maybe I should have it in here for incremental invalidation.

yingjieMiao: I thought the article s3 key is something like uriId/source/version  
If that's the case, we need to store the latest version somewhere for fast look up

leogrim: Yes the version is `fetchSeq`. I'm currently refactoring some of this with your remarks :)

yingjieMiao: oh. thanks

kifi-backend/modules/common/conf/evolutions/shoebox/300.sql
```diff
@@ -0,0 +1,34 @@
...
+    created_at datetime NOT NULL,
+    updated_at datetime NOT NULL,
+    state varchar(20) NOT NULL,
+    seq bigint(20) NOT NULL,
+    uri_id bigint(20) NOT NULL,
+    url varchar(3072) NOT NULL,
+    kind varchar(64) NOT NULL,
+    major int NULL,
+    minor int NULL,
+    last_queued_at datetime NULL,
```
yingjieMiao, line 15: for `major` and `minor`, `smallint unsigned` may be enough. Again, I think `fetch_interval` doesn't need double precision.

leogrim, line 15: `VersionNumber` takes an Int, not sure it's worth the optimization. I can switch interval to a `float`

leogrim:
@stkem @andrewconner @yingjieMiao 
Ready to merge, what do you think?

kifi-backend/modules/eliza/app/com/keepit/common/cache/ElizaCacheModule.scala
```diff
@@ -71,6 +71,16 @@ case class ElizaCacheModule(cachePluginModules: CachePluginModule*) extends Cach
 
   @Singleton
   @Provides
+  def searchIdCache(stats: CacheStatistics, accessLog: AccessLog, innerRepo: InMemoryCachePlugin, outerRepo: FortyTwoCachePlugin) =
```
stkem, line 74: Why does the Eliza Cache module need these?

leogrim, line 74: I refactored `ElizaStoreModule` to follow the pattern of our other services (most of the logic is in `ProdStoreModule` and `DevStoreModule`). Because of `ArticleSearchResult` and its store being in common, these caches are required.

leogrim:
Note: the `Eliza` and `Shoebox` commits are unrelated to `Rover`, I just did some cleanup along the way.

kifi-backend/modules/rover/app/com/keepit/rover/model/ArticleInfo.scala
```diff
@@ -0,0 +1,95 @@
...
+    id: Option[Id[ArticleInfo]] = None,
+    createdAt: DateTime = currentDateTime,
+    updatedAt: DateTime = currentDateTime,
+    state: State[ArticleInfo] = ArticleInfoStates.ACTIVE,
+    seq: SequenceNumber[ArticleInfo] = SequenceNumber.ZERO,
+    uriId: Id[NormalizedURI],
+    url: String,
+    kind: String, // todo(Léo): make this kind: ArticleKind[_ <: Article] with Scala 2.11, (with proper mapper, serialization is unchanged)
+    major: Option[VersionNumber[Article]] = None,
+    minor: Option[VersionNumber[Article]] = None,
```
stkem, line 21: maybe `minorVersion` etc., for more clarity (and yes I know it's in the type, but the type will rarely be spelled out anywhere)

stkem:
lgtm overall

kifi-backend/modules/rover/app/com/keepit/rover/model/ArticleInfo.scala
leogrim, line 21: Yeah I actually hesitated with these. I'm going to merge for now, I think I'm going to rework the available versions a bit anyway. I'm actually thinking having 4 versions:
- `latestMajorVersion` & `latestMinorVersion`
- `bestMajorVersion` & `bestMinorVersion`
  By default, the best versions would be the latest ones, but then we can have an admin page for each uri in order to review all available articles of a specific kind and potentially mark a specific version as _the best_.

leogrim, line 21: And I'd rather have a single `ArticleVersion` nested case class containing all four versions than four options but I would need to figure out how to map one field to several columns (I think we do it somewhere @andrewconner)

andrewconner, line 21: Yeah, the best way to do this is to write your own apply and unapply to convert a db row to case class. See `Keep.applyFromDbRow`, `Keep.unapplyToDbRow`, and `KeepRepo.KeepTable#*`. This pattern has worked well, and makes complex migrations a lot easier, because you have two hooks in code before and after the database.

andrewconner, line 21: In fact, I kind of want to do that pattern for all new tables, even if it's not immediately needed. It makes compile errors a lot better, and makes it very clear how db rows are mapped to classes and vice versa.

leogrim, line 21: Perfect, I'll do that :)

yingjieMiao:
lgtm

leogrim:
Thanks for the feedback guys!

PR #14115, opened 2015-03-04
minor code cleanup

2is10:


kifi-backend/modules/shoebox/app/com/keepit/commanders/UserConnectionsCommander.scala
```diff
@@ -22,6 +22,8 @@ import play.api.libs.json.Json
 import scala.concurrent.Future
 import scala.concurrent.duration.Duration
 
+case class ConnectionInfo(user: BasicUser, userId: Id[User], unfriended: Boolean, unsearched: Boolean)
+
```
2is10, line 26: This class is only used in one method signature, and it’s a method in this commander.

PR #14124, opened 2015-03-04
using real X's for image names

ahsu1230:
@2is10 sorry, I guess I was rushing & getting impatient yesterday so overlooked this part.


ahsu1230:
the hash all stay the same because the images are hashed by content

2is10:
:+1: Thanks!

PR #14133, opened 2015-03-04
break into smaller sessions

yingjieMiao:
@andrewconner @ahsu1230


ahsu1230:
:+1:

andrewconner:
Hm, what's the benefit of using readWriteBatch and a very small manually derived batch?

yingjieMiao:
you suggest just use `db.readWrite()` ?  the original call is expensive, sometimes.

PR #14134, opened 2015-03-04
if curator not responding, try again async

yingjieMiao:
@stkem @ahsu1230 

noticed a 10 seconds timeout in the log. this PR handles exception better.


stkem:
:+1: :+1:

PR #14140, opened 2015-03-05
I thought we didn't have to worry about nulls anymore.

andrewconner:
@leogrim FYI.

Made smallest change that fixes the NPE, but no clue why we do it this way to begin with.


kifi-backend/modules/scraper/app/com/keepit/scraper/extractor/JsoupBasedExtractor.scala
```diff
@@ -14,7 +14,7 @@ import java.util.regex.Pattern
 import scala.util.Try
 
 abstract class JsoupBasedExtractor(url: URI, maxContentChars: Int) extends Extractor with Logging {
-  protected var doc: Document = null
+  protected var doc: Document = null // why!?
```
leogrim, line 17: I don't know, maybe to make it work with the Extractor API. I'm going to work on this a bit anyway. Rather that the current stateful pattern, I aim for extractor to be something like Function[URI, Future[Article]]. Maybe I'll be able to get rid of these nulls.

leogrim:
:+1:

kifi-backend/modules/scraper/app/com/keepit/scraper/extractor/JsoupBasedExtractor.scala
leogrim, line 17: Actually yeah, `doc` is set by side-effect when the `HTTPInputStream` is processed. I'm rewriting that stuff functional style.

PR #14141, opened 2015-03-05
update nginx conf files

joshm1:
@stkem @andrewconner 

i'm updating the AMI for these; conf files for each service pulled from their running server


stkem:
lgtm although my build in nginx config parser is not the best :-)

tools/log_stats.py
```diff
@@ -25,8 +25,10 @@ def report(stats, topK = 20):
     rep_items = report.items()
     rep_items.sort(key = lambda x: -x[1][0])
     lines = []
+    lines.append(' ' * 89 + '{:<5s} | {:<5s} | {:<5s} | {:5s}'.format('min', '50%', '90%', '99%', 'max'))
```
joshm1: @yingjieMiao @stkem 

i'll update latest new servers manually, but the next AMI i build will include this

PR #14142, opened 2015-03-05
query mutual libraries

eishay:


kifi-backend/modules/shoebox/app/com/keepit/model/LibraryRepo.scala
```diff
@@ -309,6 +311,18 @@ class LibraryRepoImpl @Inject() (
...
+    import StaticQuery.interpolation
+    val query = sql"""select lm1.library_id from library_membership lm1 inner join library_membership lm2 on lm1.library_id = lm2.library_id where lm1.user_id = $user1 and lm1.access != 'owner' and lm1.state = 'active' and lm2.user_id = $user2 and lm2.access != 'owner' and lm2.state = 'active';"""
+    query.as[Id[Library]].list
+  }
+
+  def getOwnerLibrariesOtherFollow(owner: Id[User], other: Id[User])(implicit session: RSession): Seq[Id[Library]] = {
+    import StaticQuery.interpolation
+    val query = sql"""select lm1.library_id from library_membership lm1 inner join library_membership lm2 on lm1.library_id = lm2.library_id where lm1.user_id = $owner and lm1.access = 'owner' and lm1.state = 'active' and lm2.user_id = $other and lm2.access != 'owner' and lm2.state = 'active';"""
+    query.as[Id[Library]].list
+  }
```
2is10, line 325: With this approach, it looks like we’ll need to run 3 queries to respond to the request for libraries in common. An alternative approach would be to run one query (like these but without any `access` clauses) and then organize the results into three groups in memory.

eishay, line 325: right, i should have thought of that

PR #14171, opened 2015-03-06
counting only non empty libraries when the viewer is not self and adding...

eishay:
... some tests


eishay:
@2is10

kifi-backend/modules/shoebox/app/com/keepit/model/UserConnectionRepo.scala
```diff
@@ -211,7 +218,10 @@ class UserConnectionRepoImpl @Inject() (
           friendRequestRepo.save(friendRequest.copy(state = FriendRequestStates.IGNORED))
         }
 
-      (users + userId) foreach invalidateCache
+      users foreach { friend =>
+        deleteCacheForUsers(userId, friend)
+      }
```
2is10, line 223: This pattern will repeat 5 calls specifically for just `userId` for each friend (those in `invalidateCache`). We could optimize that repetition away if it’s worth it.

kifi-backend/modules/shoebox/test/com/keepit/controllers/website/UserProfileControllerTest.scala
```diff
@@ -322,6 +339,7 @@ class UserProfileControllerTest extends Specification with ShoeboxTestInjector {
         status(result3) must equalTo(OK)
         contentType(result3) must beSome("application/json")
         val resultJson3 = contentAsJson(result3)
+        println(s"resultJson3=$resultJson3")
```
2is10, line 342: remove?

2is10:
:+1: Great find, thanks for fixing.

kifi-backend/modules/shoebox/app/com/keepit/model/UserConnectionRepo.scala
eishay, line 223: right

kifi-backend/modules/shoebox/test/com/keepit/controllers/website/UserProfileControllerTest.scala
eishay, line 342: sure

PR #14173, opened 2015-03-06
Introduce EmbedlyArticleFetcher

leogrim:
@yingjieMiao @stkem @andrewconner 
Trying to get the best of both worlds i.e. storing Embedly's JSON in full while properly validating the subset that we're using.

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/common/net/WebService.scala
+++ b/kifi-backend/modules/common/app/com/keepit/common/net/WebService.scala
@@ -1,7 +1,13 @@
 package com.keepit.common.net
 
+import java.util.concurrent.atomic.AtomicInteger
+
 import com.google.inject.ImplementedBy
-import play.api.libs.ws.WSRequestHolder
+import com.keepit.common.concurrent.RetryFuture
+import com.keepit.common.logging.Logging
+import play.api.libs.ws.{ WSResponse, WSRequestHolder }
+
+import scala.concurrent.Future
 
 @ImplementedBy(classOf[PlayWS])
 trait WebService {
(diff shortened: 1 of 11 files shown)


kifi-backend/modules/common/app/com/keepit/common/net/WebService.scala
```diff
@@ -14,3 +20,18 @@ class PlayWS extends WebService {
 
   def url(url: String): WSRequestHolder = WS.url(url)(current)
 }
+
+object WebServiceUtils extends Logging {
+  def getWithRetry(request: WSRequestHolder, attempts: Int): Future[WSResponse] = {
+    val count = new AtomicInteger()
+    val resolver: PartialFunction[Throwable, Boolean] = {
+      case t: Throwable =>
+        count.getAndIncrement
```
yingjieMiao, line 29: is this atomic `count` used for flow control? Seems not. If this is just for logging, maybe put the logging code inside `RetryFuture`?

kifi-backend/modules/common/app/com/keepit/rover/article/EmbedlyContent.scala
```diff
@@ -1,32 +1,36 @@
...
-
-  def rawContent = (json \ "content").asOpt[String]
-  def isSafe = (json \ "safe").asOpt[Boolean]
-  def language = (json \ "language").asOpt[String]
-  def images = (json \ "images").as[Seq[EmbedlyImage]]
-  def embedlyKeywords = (json \ "keywords").as[Seq[EmbedlyKeyword]]
-  def entities = (json \ "entities").as[Seq[EmbedlyEntity]]
-  def faviconUrl = (json \ "favicon_url").asOpt[String]
-  def media = (json \ "media").asOpt[EmbedlyMedia]
+  private val parsed = json.as[EmbedlyContent.ParsedFields]
```
yingjieMiao, line 12: The old code parses lazily. could be a lazy val here? or maybe not worth it.

yingjieMiao:
lgtm

kifi-backend/modules/common/app/com/keepit/common/net/WebService.scala
leogrim, line 29: Ah, that's actually your code form EmbedlyClient ;)
Yeah I thought about adding to `RetryFuture`, maybe.

kifi-backend/modules/common/app/com/keepit/rover/article/EmbedlyContent.scala
leogrim, line 12: Yeah that was my intention not to parse lazily, so that validation happens despite keeping the original json around.

kifi-backend/modules/common/app/com/keepit/common/net/WebService.scala
yingjieMiao, line 29: LOL

leogrim, line 29: Oh I know, the issue with having the logging in `RetryFuture` is that we cannot provide any context.

PR #14181, opened 2015-03-06
Fix embedly client

leogrim:
@eishay


leogrim:
I factored things out for Rover Yesterday and introduced the typo, I should have caught the exceptions earlier, apology.

PR #14182, opened 2015-03-06
persona library ingestion filter out user's libs

yingjieMiao:
@stkem 

if user owns/follows a lib, filter it out when ingest persona libs


stkem:
:+1:

PR #14185, opened 2015-03-06
tracking for clicking on signupEmail and signupName

ahsu1230:
@andrewconner


ahsu1230:
next steps: all error messages -> tracking on viewed_page

PR #14186, opened 2015-03-06
Disabling Heisentest for now

stkem:
@ahsu1230 @eishay @andrewconner


andrewconner:
Will investigate later, thanks.

PR #14192, opened 2015-03-06
new signup all error handling

ahsu1230:
@andrewconner


ahsu1230:
created error functions that will highlight the error tip at some object, display a message, and then `.focus().select()` on the desired input object for easy re-input

PR #14194, opened 2015-03-06
tracking for all errors in reg flow

ahsu1230:
@andrewconner 

tracking is included inside the error functions


andrewconner:
Great, thanks

PR #14205, opened 2015-03-07
fixing broken connections link

ahsu1230:
@2is10 @eishay


kifi-backend/modules/shoebox/angular/src/header/loggedInHeader.tpl.html
```diff
@@ -31,7 +31,7 @@
       <button class="kf-lih-add-link sprite sprite-add-link" ng-click="addKeeps(library)">Add a link</button>
       <a href="/tags/manage" class="kf-lih-tags"><span class="svg-tag-white"></span>Manage tags</a>
       <span class="kf-lih-dropdown-parent" kf-hover-menu>
-        <a href="/connections" class="kf-lih-friends"><span class="svg-friends-white"></span>Connections</a>
+        <a href="{{me|profileUrl:'connections'}}" class="kf-lih-friends"><span class="svg-friends-white"></span>Connections</a>
```
eishay, line 34: thanks!

2is10, line 34: whoops, I updated the one two lines down from here, but forgot that the menu button itself was also a link. thanks!

PR #14208, opened 2015-03-07
new install page

ahsu1230:
@andrewconner 

at `/new/install`


kifi-backend/modules/shoebox/app/views/authMinimal/install.scala.html
```diff
@@ -0,0 +1,52 @@
...
+        <a class="browser-button chrome" href="https://www.google.com/chrome" data-track-action="getChrome">Get Chrome</a>
+        <a class="browser-button firefox" href="https://www.mozilla.org/firefox" data-track-action="getFirefox">Get Firefox</a>
+      </div>
+
+
+    </div>
+  </div>
+}
+@scripts = {
+  <script src="/assets/ml/js/authentication.js"></script>
```
andrewconner: Needed?

PR #14217, opened 2015-03-08
api for friend recommendations, inspired by UserCommander#getFriendRecom...

eishay:
...mendations and fitted to the format of new user profile cards of profile pages.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/UserProfileController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/UserProfileController.scala
@@ -1,6 +1,7 @@
 package com.keepit.controllers.website
 
 import com.google.inject.Inject
+import com.keepit.abook.ABookServiceClient
 import com.keepit.commanders._
 import com.keepit.common.controller._
 import com.keepit.common.db.slick._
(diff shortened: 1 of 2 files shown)


eishay:
@2is10 @leogrim

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/UserProfileController.scala
```diff
@@ -60,6 +62,32 @@ class UserProfileController @Inject() (
     }
   }
 
+  /**
+   * @param fullInfoLimit trying to get at least this number of full card info
+   * @param maxExtraIds if available, returning external id of additional maxExtraIds ids that the frontend can iterate on as it does with getProfileConnections
+   */
+  def getFriendRecommendations(fullInfoLimit: Int, maxExtraIds: Int) = UserAction.async { request =>
+    abookServiceClient.getFriendRecommendations(request.userId, 0, fullInfoLimit + maxExtraIds, true) map {
+      case None => Ok(Json.obj("users" -> JsArray()))
```
2is10, line 71: @eishay Why would this be `None`? Should I handle it like a 500 (temporary server error)?

2is10:
:+1: Thanks!

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/UserProfileController.scala
leogrim, line 71: I think it''s handling the case where the graph is late and hasn't ingested
the user yet

PR #14220, opened 2015-03-08
reverting cron jobs that run on a single machine to ignore offline.

eishay:
I suspect the code is not working right, I'll add tests before re-introducing it.


eishay:
@stkem

PR #14224, opened 2015-03-09
pointing out tests that tries to init db tables after the scope of the t...

eishay:
...est is closed


eishay:
This PR is also demonstrating the right pattern to use `withDb() { implicit injector =>`.
If we aggregate few test cases with in the same injector context then we potentially spill state from one test to the other.
While doing it some tests broke, I fixed some and some which I didn't know how to quickly fix I commented out.

eishay:
Please check it out @andrewconner @stkem @yingjieMiao @leogrim

stkem:
- @ahsu1230

2is10:
good find

ahsu1230:
nice! thanks!

PR #14235, opened 2015-03-09
always sending people to profile page to connect

2is10:
@andrewconner

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/common/mail/package.scala
+++ b/kifi-backend/modules/common/app/com/keepit/common/mail/package.scala
@@ -112,7 +112,7 @@ package object template {
 
     def kifiFriendsUrl(content: String) = htmlUrl(s"$baseUrl/friends?", content, openInAppIfMobile = true)
 
-    def acceptFriendUrl(id: Id[User], content: String) = htmlUrl(s"$baseUrl/friends?friend=${userExternalId(id)}&", content, openInAppIfMobile = true)
+    def acceptFriendUrl(id: Id[User], content: String) = htmlUrl(Tag1(tags.profileUrl, id) + "?intent=connect&", content, openInAppIfMobile = true)
 
     private def connectNetworkUrl(network: String, content: String): Html = Html {
       s"$baseUrl/link/$network?${EmailTrackingParam.paramName}=${trackingParam(content)}"
@@ -122,10 +122,10 @@ package object template {
     def connectLinkedInUrl(content: String) = connectNetworkUrl("linkedin", content)
 
     def inviteContactUrl(id: Id[User], content: String) =
-      htmlUrl(s"$baseUrl/invite?friend=${userExternalId(id)}&subtype=contactJoined&", content, openInAppIfMobile = true)
+      htmlUrl(Tag1(tags.profileUrl, id) + "?intent=connect&", content, openInAppIfMobile = true)
 
     def inviteFriendUrl(id: Id[User], index: Int, subtype: String) =
-      htmlUrl(s"$baseUrl/invite?friend=${userExternalId(id)}&subtype=$subtype&", "pymk" + index, openInAppIfMobile = true)
+      htmlUrl(Tag1(tags.profileUrl, id) + "?intent=connect&", "pymk" + index, openInAppIfMobile = true)
 
     def invitedLibrariesUrl(id: Id[User], content: String) = htmlUrl(Tag1(tags.profileUrl, id) + s"/libraries/invited?", content, openInAppIfMobile = true)
 
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
@@ -894,7 +894,7 @@ class LibraryCommander @Inject() (
       userIds = Set(lib.ownerId),
       title = "New Library Follower",
       body = s"${follower.firstName} ${follower.lastName} is now following your Library ${lib.name}",
-      linkText = s"See ${follower.firstName}'s profile",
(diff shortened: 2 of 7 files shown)


kifi-backend/modules/common/app/com/keepit/common/mail/package.scala
```diff
@@ -112,7 +112,7 @@ package object template {
 
     def kifiFriendsUrl(content: String) = htmlUrl(s"$baseUrl/friends?", content, openInAppIfMobile = true)
 
-    def acceptFriendUrl(id: Id[User], content: String) = htmlUrl(s"$baseUrl/friends?friend=${userExternalId(id)}&", content, openInAppIfMobile = true)
+    def acceptFriendUrl(id: Id[User], content: String) = htmlUrl(Tag1(tags.profileUrl, id) + "?intent=connect&", content, openInAppIfMobile = true)
```
2is10, line 115: The prior destination for accepting a friend request was /friends, which redirects to the viewer’s own connections page. I’m changing it to always go to the specific friend’s profile page. If the friend request is still pending, there is a banner at the top. If the users are not connected and there is no friend request, there is (as of today) a green callout pointing to the Connect button.

kifi-backend/modules/common/app/com/keepit/common/mail/package.scala
```diff
@@ -122,10 +122,10 @@ package object template {
     def connectLinkedInUrl(content: String) = connectNetworkUrl("linkedin", content)
 
     def inviteContactUrl(id: Id[User], content: String) =
-      htmlUrl(s"$baseUrl/invite?friend=${userExternalId(id)}&subtype=contactJoined&", content, openInAppIfMobile = true)
+      htmlUrl(Tag1(tags.profileUrl, id) + "?intent=connect&", content, openInAppIfMobile = true)
 
     def inviteFriendUrl(id: Id[User], index: Int, subtype: String) =
-      htmlUrl(s"$baseUrl/invite?friend=${userExternalId(id)}&subtype=$subtype&", "pymk" + index, openInAppIfMobile = true)
+      htmlUrl(Tag1(tags.profileUrl, id) + "?intent=connect&", "pymk" + index, openInAppIfMobile = true)
```
2is10, line 128: These will now go to the new user’s profile page with a green callout on the Connect button.

The invite-friend-to-connect banner on the /invite page is about to die.

kifi-backend/modules/common/app/com/keepit/common/mail/package.scala
andrewconner, line 115: Hm, looks like this broke mobile, which tries to parse the links. Can I revert this line? Alternatively, we need to come up with a way both clients are happy.

2is10, line 115: The mobile app doesn’t have user profiles?

2is10, line 115: I’d like to know exactly what the mobile app _can_ handle before suggesting a solution.

If the link needs to be reverted to exactly what it was before, then we could add something in KifiSiteRouter to redirect that old link to this new one.

2is10, line 115: We may want to introduce a layer of abstraction for email links so that they don’t have to change with each site redesign or logic change requested by product. Instead, they could just have an email template name and an argument list of relevant public/external IDs. For example:

www.kifi.com/email/contactJoined/{user-external-id}
www.kifi.com/email/messageReceived/{message-external-id}
www.kifi.com/email/invitedToConnect/{user-external-id}
www.kifi.com/email/invitedToFollow/{library-public-id}/{sender-external-id}

We kind of started down this path with the “Email Link API” section of our shoebox routes, but then I guess we never encouraged Josh to follow this pattern.

andrewconner, line 115: Agreed, I think that's the long term solution.

Mobile _does_ have profiles, but the mobile apps are parsing the URL that the user came from. They were expecting the old one, not this one.

2is10, line 115: Sorry, I’m not totally following you. Are you saying that no other emails link to user profiles, so they don’t properly handle user profile URLs yet?

andrewconner, line 115: Mobile apps are passed in what URL was originally attempted. They currently, because we do not have a better mechanism, parse the URL, and then do whatever they need to from it. For this case, if I am remembering the details from my discussion with them, they are looking for `?friend=$externalId`. Now, they instead don't see that field, and see the profile URL with `?intent=connect`. This could work, but breaks backwards compatibility with existing apps.

andrewconner, line 115: Either way, the existing mobile clients do not do the right thing. I think they go to a friends page for accepting invites still, need to check.

2is10, line 115: Thanks, that helps, but I still don’t have the full picture. Are they getting more than just a URL? Otherwise, how would they know to look for `?friend=$externalId` if the URL path is not even `/friends`? I didn’t expect them to be looking for `?intent=connect` but I did expect them to correctly handle user profile URLs since I thought I saw other emails linking there, and a connection should be easy to request or accept from a profile page—at least that was my thinking. I’ll sync with them tomorrow. Sorry for the trouble.

mrbrown14, line 115: Right now on mobile it takes you to the profile screen but isnt aware of
the connection/invite, since it doesnt parse the intent param yet.

When I did deep linking some weeks ago accepting friend requests still went
to the "connections" list. So this is new to us.

PR #14243, opened 2015-03-09
explicit lower bound on scrape interval

yingjieMiao:
@eishay @stkem @andrewconner 

In `prod.conf`, there is a configuration of min scrape interval, set to 4 weeks. 
50% items (over 1 million) in the scrape_info table actually have much smaller scrape intervals. 

I put the cap explicitly in the code. (The old code only works if we can assume every existing scrape_interval is already above min value)

If we think 4 weeks is an overkill, we can tweak it shorter. The point is to make sure the configuration really in effect.


stkem:
lgtm. keep at it.

PR #14246, opened 2015-03-09
removing seqNum from profileService.me

2is10:
@andrewconner I don’t see this being used anywhere. Okay to remove?


andrewconner:
I suppose the original use case was removed. It's an easy way for a watcher to get notified if `me` updated. Yeah if we don't need it, feel free to remove.

2is10:
Thanks. I’d just like to simplify for now and make this easier to refactor.

PR #14255, opened 2015-03-09
error message for image uploads!

ahsu1230:
@andrewconner


kifi-backend/modules/shoebox/public/ml/css/authentication.css
```diff
@@ -742,6 +742,15 @@ a, .avatars li, img, .tabs_nav li {
 	80% { -webkit-transform: translate(3px, 0px) rotate(0deg); }
 }
 
+#error-fp {
```
ahsu1230, line 745: shifts over forgot password error message (like email)

PR #14256, opened 2015-03-09
avoiding race condition, not augmenting library objects with .isMine

2is10:
@ahsu1230

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/header/loggedInHeader.js
+++ b/kifi-backend/modules/shoebox/angular/src/header/loggedInHeader.js
@@ -3,9 +3,9 @@
 angular.module('kifi')
 
 .controller('LoggedInHeaderCtrl', [
-  '$scope', '$window', '$rootElement', '$rootScope', '$document', 'profileService',
+  '$scope', '$window', '$rootElement', '$rootScope', '$document', 'profileService', 'libraryService',
   '$location', 'util', 'keyIndices', 'modalService', '$timeout', '$state',
-  function ($scope, $window, $rootElement, $rootScope, $document, profileService,
+  function ($scope, $window, $rootElement, $rootScope, $document, profileService, libraryService,
     $location, util, keyIndices, modalService, $timeout, $state) {
 
     $scope.toggleMenu = function () {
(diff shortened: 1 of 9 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/library.js
```diff
@@ -185,9 +185,7 @@ angular.module('kifi')
           if (existingKeep && !_.find($scope.keeps, { id: keep.id })) {
             existingKeep.keeps.push({
               id: keep.id,
-              isMine: true,
               libraryId: library.id,
-              mine: true,
```
2is10, line 190: As far as I can tell, neither of these fields is used.

kifi-backend/modules/shoebox/angular/src/libraries/libraryMiniCard.js
```diff
@@ -48,18 +48,16 @@ angular.module('kifi')
...
+          return !scope.invite && !scope.isMyLibrary() && scope.isFollowing;
         };
 
         scope.toggleFollow = function () {
-          if (scope.library.isMine) {
-            modalService.openGenericErrorModal({
-              modalData: {
-                genericErrorMessage: 'You cannot follow your own libraries!'
-              }
-            });
```
2is10, line 61: I think there’s never a case where it’s good to show a user this message. If they click Follow on their own library, that’s our fault for showing a Follow button on it.

kifi-backend/modules/shoebox/angular/src/header/loggedInHeader.js
```diff
@@ -131,7 +131,7 @@ angular.module('kifi')
     $scope.addKeeps = function (library) {
       modalService.open({
         template: 'keeps/addKeepsModal.tpl.html',
-        modalData: {selectedLibId: library && library.isMine && library.id}
+        modalData: {selectedLibId: library && libraryService.isMyLibrary(library) && library.id}
```
2is10, line 134: This is one place where the race condition was manifesting. The library augmentation code sometime ran before the webapp figured out who the logged-in user was (`profileService.me` was not yet initialized) and so `library.isMine` would be `false` here instead of `true` if the user ever clicked the Add keeps button on the page, and so we’d suggest they keep to their main library instead of the one they were actually looking at.

PR #14262, opened 2015-03-10
rate limit on content check

yingjieMiao:
@eishay @leogrim @stkem @andrewconner 

To see the problem, see `failed_content_check` table, and query `url1 like %twitter.com%`

This should reduce load both on shoebox and scraper. I think the change is safe.


stkem:
let's give it a try. second (or rather third) pair of eyes for quick review might be good.

PR #14268, opened 2015-03-10
new 500 page

stkem:
@andrewconner 
We should replace the svg with a png. This is far too big (and doesn't compress well), but it will do for now.


andrewconner:
Could you add the exception external ID back in? It helps me quite a bit when trying to find one specific exception, and lets users report something to us.

(The current wording doesn't work since it's a 500, but I'll let product figure that one out. The info is very useful.)

PR #14277, opened 2015-03-10
returning library image URL for use in library hovercards on site

2is10:
@ahsu1230


kifi-backend/modules/shoebox/test/com/keepit/commanders/PageMetaTagsCommanderTest.scala
```diff
@@ -1,7 +1,6 @@
 package com.keepit.commanders
 
 import com.keepit.abook.FakeABookServiceClientModule
-import com.keepit.commanders.UserProfileTab
```
2is10, line 4: fixes compile warning

PR #14278, opened 2015-03-10
new filter for library image inline styles, using it in library hovercards

2is10:
@ahsu1230


ahsu1230:
nice, these filter usages are awesome!

andrewconner:
Thanks

PR #14279, opened 2015-03-10
reset password page

ahsu1230:
@andrewconner 

feel free to merge to start using & QA-ing
still needs specific tracking & correct success case handling

### DIFF
--- a/kifi-backend/modules/shoebox/public/ml/css/install.css
+++ b/kifi-backend/modules/shoebox/public/ml/css/install.css
@@ -14,6 +14,7 @@
 
 .install-title {
   margin-top: 40px;
+  margin-bottom: 40px;
   font-size: 1.7em;
 }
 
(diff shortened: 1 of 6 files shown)


kifi-backend/modules/shoebox/public/ml/js/authentication.js
```diff
@@ -282,6 +282,49 @@ $(function() {
...
+        window.location = '/'; // todo: best location for success?
+      }
+    }).fail(function (xhr) {
+      var body = xhr.responseJSON || {};
+      var $firstPw = $('.form-input').eq(0);
+      errorUnknown($('#error-reset-password1'), $firstPw, 'resetPassword');
+    });
+    return false;
+  };
+  $('.reset-password-button').click(kifi.resetPassword);
```
ahsu1230, line 327: should actually be a `submit`

kifi-backend/modules/shoebox/public/ml/css/install.css
```diff
@@ -21,22 +22,22 @@
   margin: 0 auto;
   width: 300px;
   line-height: 50px;
-  background-color: white;
-  color: #3E4954;
-  border: 1px solid #aaa;
+  background-color: #2CB4E9;
```
ahsu1230, line 25: better call to actions

PR #14280, opened 2015-03-10
Many minor fixes and improvements

andrewconner:
@2is10 This covers roughly 2/3 of your suggestions. More coming.


2is10:
:+1: thanks

PR #14292, opened 2015-03-10
verifying eishay@kifi.com so he can be the root of the delegation chain

2is10:
@eishay

### DIFF
--- a/tools/etc/nginx/conf.d/shoebox.conf
+++ b/tools/etc/nginx/conf.d/shoebox.conf
@@ -173,8 +173,9 @@ server {
     return 200 '<!doctype html>';
   }
 
-  location = /google6de4ad16957106ef.html {
-    return 200 'google-site-verification: google6de4ad16957106ef.html';
+  # site verification for eishay@kifi.com
+  location = /google25ae05cb8bf5b064.html {
+    return 200 'google-site-verification: google25ae05cb8bf5b064.html';
   }
 }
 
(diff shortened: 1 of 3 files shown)


tools/etc/nginx/conf.d/shoebox.conf
```diff
@@ -173,8 +173,9 @@ server {
     return 200 '<!doctype html>';
   }
 
-  location = /google6de4ad16957106ef.html {
-    return 200 'google-site-verification: google6de4ad16957106ef.html';
+  # site verification for eishay@kifi.com
+  location = /google25ae05cb8bf5b064.html {
+    return 200 'google-site-verification: google25ae05cb8bf5b064.html';
```
2is10, line 178: @stkem Will this change automatically make it into our next shoebox AMI?

stkem, line 178: No, the configuration files need to be updated manually on the AMI and all the running shoeboxen. Hopefully we'll have that issue addressed in the next 3-4 weeks, but for now that's what it is.

stkem, line 178: I.e. yes, it will make it into the next AMI, but only when somebody decides to make a new one.

2is10, line 178: Great, thanks. I’ve updated the six current shoeboxen. I hope that any new shoeboxen we spin up will come from a new AMI. :pray:

PR #14293, opened 2015-03-10
set ScrapeAgent receive timeout when busy

yingjieMiao:
@eishay @stkem @andrewconner 

When context switched to busy, a ScrapeAgent can get stuck and never becomes available again.

This breaks the circuit after 60 seconds. Should be pretty lenient.


yingjieMiao:
never mind, this PR is flawed.

PR #14294, opened 2015-03-10
set ScrapeAgent receive timeout when busy

yingjieMiao:
@eishay @stkem @andrewconner 

This prevents a ScrapeAgent to get stuck, and eventually non of the agent is available for scraping work. 

In addition, I removed the broadcast router, partly to make sure the ReceiveTimeout works. (Also reduce waste)

An easier solution is to use an Await for `worker.safeProcess` to break circuit.


yingjieMiao:
will give this a try and babysit.

PR #14298, opened 2015-03-10
including numFollowedLibraries in user profile responses

2is10:
@eishay 

This is so that we can quickly decide to show the followed libraries tab instead of the own libraries tab in a user profile if the user owns none but is following some.
https://app.asana.com/0/20849692840229/28604317893859

In the future, we might also decide to display the counts on the tabs.

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
+++ b/kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
@@ -35,17 +35,17 @@ angular.module('kifi')
       var url = $analytics.settings.pageTracking.basePath + $location.url();
       $analytics.pageTrack(url, originTrackingService.applyAndClear({
         type: $rootScope.userLoggedIn ? 'userProfile' : 'userProfileLanding',
-        profileOwnerUserId: $scope.profile.id,
-        profileOwnedBy: $scope.viewingOwnProfile ? 'viewer' : ($scope.profile.isFriend ? 'viewersFriend' : 'other'),
-        libraryCount: $scope.profile.numLibraries
+        profileOwnerUserId: profile.id,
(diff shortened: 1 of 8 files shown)


kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
```diff
@@ -35,17 +35,17 @@ angular.module('kifi')
       var url = $analytics.settings.pageTracking.basePath + $location.url();
       $analytics.pageTrack(url, originTrackingService.applyAndClear({
         type: $rootScope.userLoggedIn ? 'userProfile' : 'userProfileLanding',
-        profileOwnerUserId: $scope.profile.id,
-        profileOwnedBy: $scope.viewingOwnProfile ? 'viewer' : ($scope.profile.isFriend ? 'viewersFriend' : 'other'),
-        libraryCount: $scope.profile.numLibraries
+        profileOwnerUserId: profile.id,
+        profileOwnedBy: $scope.viewingOwnProfile ? 'viewer' : (profile.isFriend ? 'viewersFriend' : 'other'),
+        libraryCount: profile.numLibraries
```
2is10, line 40: no functional change, just less code

kifi-backend/modules/shoebox/angular/src/userProfile/userProfile.js
```diff
@@ -69,23 +69,23 @@ angular.module('kifi')
...
+      $rootScope.$on('getCurrentLibrary', function (e, args) {
+        args.callback({});
+      }),
+      $rootScope.$on('trackUserProfileEvent', function (e, eventType, attributes) {
+        if (eventType === 'click') {
+          trackPageClick(attributes);
+        } else if (eventType === 'view') {
+          trackPageView(attributes);
+        }
+      })
```
2is10, line 88: indentation only

PR #14301, opened 2015-03-11
Waitlist endpoint

andrewconner:
And some bad equals fixes! @leogrim


kifi-backend/modules/shoebox/app/com/keepit/controllers/website/TwitterWaitlistController.scala
```diff
@@ -3,22 +3,39 @@ package com.keepit.controllers.website
...
+import scala.util.Try
 
 class TwitterWaitlistController @Inject() (
     commander: TwitterWaitlistCommander,
+    socialRepo: SocialUserInfoRepo,
+    userEmailAddressRepo: UserEmailAddressRepo,
+    db: Database,
     val userActionsHelper: UserActionsHelper) extends UserActions with ShoeboxServiceController {
 
   //DO NOT USE THE WORD *FAKE* IN THE ROUTE FOR THIS!!!
```
2is10, line 20: ha ha

PR #14265, opened 2015-03-10
ApacheHttpFetcher Spring cleanup (part 1)

leogrim:
@eishay @andrewconner @stkem @yingjieMiao 

Part 2 is moving away from side-effecting to set extractors.

PR #14308, opened 2015-03-11
HttpFetcher Spring Cleanup (Part 2)

leogrim:
@andrewconner @yingjieMiao @stkem @eishay 
Follow up to part 1 @ https://github.com/kifi/keepit/pull/14265
Now in better shape to help Rover :)


kifi-backend/modules/common/app/com/keepit/rover/fetcher/HttpFetcher.scala
```diff
@@ -6,31 +6,33 @@ import com.keepit.model.HttpProxy
...
+    case FetchContentExtractionError(_) => None
+  }
+}
 
 class HttpInputStream(input: InputStream, val contentType: Option[String] = None) extends FilterInputStream(input)
 
 trait HttpFetcher {
-  def fetch[A](request: FetchRequest)(f: HttpInputStream => A): Future[FetchResponse[A]]
+  def fetch[A](request: FetchRequest)(f: HttpInputStream => A): Future[FetchResult[A]]
 }
```
leogrim, line 38: Mostly interested in comments about this API, though we can always change it later on. I'll merge :)

PR #14309, opened 2015-03-11
Support library search filter in the new website import

leogrim:
@2is10 
Note that I have changed the filter "f" parameter to "u". The reasoning is that our current filter is just about filtering by user, and searching within a library is actually about filtering by library. So this is an attempt to have a more symmetric API where it's just "u"sers and "l"ibrary. I'd like to move towards this more consistently.

### DIFF
--- a/kifi-backend/modules/search/app/com/keepit/search/controllers/ext/ExtSearchController.scala
+++ b/kifi-backend/modules/search/app/com/keepit/search/controllers/ext/ExtSearchController.scala
@@ -72,7 +72,7 @@ class ExtSearchController @Inject() (
     extVersion: Option[KifiExtVersion],
     debug: Option[String] = None) = UserAction { request =>
 
-    val libraryContextFuture = getLibraryContextFuture(None, None, request)
+    val libraryContextFuture = getLibraryFilterFuture(None, None, request)
     val acceptLangs = getAcceptLangs(request)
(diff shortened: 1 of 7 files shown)


2is10:
Okay, sounds good to me.

What would it take for “My Main Library” and “My Private Library” to show up in results? I could implement it in the browser, but I wouldn’t have a good idea of where they should be ranked among other matching libraries.

leogrim:
We have not been indexing the names of system libraries in order to reduce noise in keep search. But now that search is becoming navigational for libraries and people I think we should. It's an easy change on the backend, we'll just have to reindex libraries.

leogrim:
I've made the change.

2is10:
> I've made the change.

Wow, that was fast. Thanks. Will they begin showing up on google.com now too? I’m not sure that’s desirable.

2is10:
To answer my own question, I imagine this won’t affect how we do attribution for keeps.

leogrim:
Nope it won't.

leogrim:
Also, the user filter 'u' now accepts a user's external id in addition to the regular parameters for all, friends and mine. Results are then restricted to that user's keeps from libraries you have access to (for that reason it won't include results from main, a technical limitation in the current implementation, I only spent a few hours on this). This means we could have a search bar on user profile pages, very much like we do on library pages. :)

PR #14315, opened 2015-03-11
related library. partial update	

yingjieMiao:
@stkem @eishay 

this avoid the unnecessary full graph computation and persistence every 3 hours. 
the online update part deals with:
- create adjacency list for new library
- remove edges of an inactive library (both directions)
- all other edges and weights remain unchanged


yingjieMiao:
will give this a try.

PR #14318, opened 2015-03-12
rendering library keeps nicely, one at a time

2is10:
@andrewconner 

I helped Yiping do this for search results a while back. I think it really helps with perceived page load time. I’m doing it now because it also fixes a UI glitch I’ve encountered while working on the new search suggestions feature.

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/library.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/library.js
@@ -91,27 +91,41 @@ angular.module('kifi')
         return;
       }
       $scope.loading = true;
-      return libraryService.getKeepsInLibrary(library.id, offset, $stateParams.authToken).then(function (res) {
+      libraryService.getKeepsInLibrary(library.id, offset, $stateParams.authToken).then(function (res) {
         var rawKeeps = res.keeps;
-        rawKeeps.forEach(function (rawKeep) {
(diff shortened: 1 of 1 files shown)


andrewconner:
Cool. It's a speed thing?

eishay:
nice

2is10:
> Cool. It's a speed thing?

The blocking DOM manipulations done to compute the optimal layout for each keep card really add up. For 10 keeps, it takes about 1 second on my MacBook Pro. This can freeze other things going on in the page, sometimes very noticeably. I was seeing the search bar background-color transition pause halfway between dark and light, for example. I disabled that transition during page navigations, but then still saw the bar staying light for 1½ sec longer than it should have. This change avoids those issues.

andrewconner:
Oh interesting, I never saw it that slowly. I'm looking forward to the more simple, static designs coming down the pipeline.

2is10:
Me too.

I think one reason we don’t often see it that slow is because I was specifically testing navigating back and forth between two libraries, which becomes much easier to do with the new empty search box suggestions. In this case, I think we have been synchronously rendering 20 or 30 keeps instead of just the initial 10, since prefetching was done using `.$evalAsync` (which runs before rendering) and because `Clutch` invokes callbacks synchronously when data is cached. I’m going on my memory of the code here and so could be wrong, but this explanation would be consistent with the behavior I was observing.

PR #14324, opened 2015-03-12
fixing randomly failing test by injecting defaultContext and draining it...

eishay:
... before testing result.

I wonder why this test didn't fail more: http://localhost:6789/job/all-quick-s3/849/consoleFull


eishay:
@leogrim @stkem

PR #14326, opened 2015-03-12
using library filter in new site search endpoint

2is10:
@leogrim


leogrim:
Oh so basically I stopped half-way on Wednesday after renaming the query parameters, that's great.

PR #14327, opened 2015-03-12
Better scraping, minor site changes, waitlist

andrewconner:
@leogrim @2is10 @ahsu1230

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/core/AuthController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/core/AuthController.scala
@@ -444,11 +444,14 @@ class AuthController @Inject() (
             if (cookieIntent.isDefined) {
               cookieIntent.get.value match {
                 case "waitlist" =>
-                  socialRepo.getByUser(ur.user.id.get).find(_.networkType == SocialNetworks.TWITTER).flatMap {
-                    _.getProfileUrl.map(url => url.substring(url.lastIndexOf('/') + 1))
+                  db.readOnlyReplica { implicit session =>
+                    socialRepo.getByUser(ur.user.id.get).find(_.networkType == SocialNetworks.TWITTER).flatMap {
+                      _.getProfileUrl.map(url => url.substring(url.lastIndexOf('/') + 1))
+                    }
                   }.map { handle =>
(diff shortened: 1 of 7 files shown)


kifi-backend/modules/shoebox/app/com/keepit/controllers/core/AuthHelper.scala
```diff
@@ -224,6 +225,14 @@ class AuthHelper @Inject() (
         val result = if (isFinalizedImmediately) {
           Redirect(uri)
         } else if (cookieIntent.isDefined && cookieIntent.get.value == "waitlist") {
+          db.readOnlyMaster { implicit session =>
+            socialRepo.getByUser(user.id.get).find(_.networkType == SocialNetworks.TWITTER).flatMap {
+              _.getProfileUrl.map(url => url.substring(url.lastIndexOf('/') + 1))
```
ahsu1230, line 230: seems like this line is used in multiple places to retrieve twitter handles.
It would be helpful if we could factor this logic out into a `getTwitterHandle` function. Could be now or later, whatever works.

kifi-backend/modules/shoebox/app/com/keepit/controllers/core/AuthController.scala
```diff
@@ -443,6 +444,14 @@ class AuthController @Inject() (
             if (cookieIntent.isDefined) {
               cookieIntent.get.value match {
                 case "waitlist" =>
+                  db.readOnlyReplica { implicit session =>
```
ahsu1230, line 447: I see `AuthController` uses `readOnlyReplica` while `AuthHelper` uses `readOnlyMaster` is there a reason for the difference?

kifi-backend/modules/shoebox/app/com/keepit/controllers/core/AuthHelper.scala
andrewconner, line 230: Good idea

kifi-backend/modules/shoebox/app/com/keepit/controllers/core/AuthController.scala
andrewconner, line 447: I prefer replica, but right after a signup, it's possible that the replica doesn't have it. Slight, but thought it was worth it.

PR #14330, opened 2015-03-12
Fixing bad mysql query

andrewconner:
@2is10 @eishay 

Mysql doesn't like to do `count(table.*)`


andrewconner:
Will see why a test didn't catch this.

2is10:
:+1: Thanks, sorry. I thought I saw test coverage for these queries at the commander level. Could this be a difference between H2 and MYSQL?

Lesson learned: Always run a raw query in a MYSQL console before merging.

2is10:
Yes, FYI, I just verified that `count(table.*)` works in H2. It’s a dialect difference.

PR #14331, opened 2015-03-12
Deploy new library index

leogrim:
@2is10


2is10:
Sounds good. I think I began seeing my main library showing up in search results last night, btw.

leogrim:
I think library sequence numbers get bumped up on each keep (because of the `lastKeptTo` field), so it probably got reindexed lazily because of that.

PR #14332, opened 2015-03-12
Hamburger Icon instead of Menu in /install page

ahsu1230:
@andrewconner @2is10


andrewconner:
Nice! Thanks, much better. Thanks for the suggestion Jared.

PR #14338, opened 2015-03-12
housekeeping in the mobile app redirect page

2is10:
no functional changes
@eishay 

We should update the logo and tagline in this page too. It currently looks like this:

image


2is10:
I verified that the page compiles and that the HTML and JS code looks how I expect. The changes from the old page are minor.

PR #14341, opened 2015-03-12
clarifying comment

2is10:
@andrewconner


kifi-backend/modules/shoebox/app/com/keepit/controllers/website/KifiSiteRouter.scala
```diff
@@ -30,11 +30,10 @@ private case class SeeOtherRoute(url: String) extends Routeable
 private case class RedirectToLogin(originalUrl: String) extends Routeable
 private case object Error404 extends Routeable
 
-@Singleton // holds state for performance reasons
+@Singleton // for performance
```
2is10, line 33: “holds state” concerned me that this class was _stateful_, which I think we agree isn’t the case, since it’s thread-safe.

andrewconner, line 33: Correct, thanks. Just cache stuff, not stateful.

PR #14342, opened 2015-03-12
set of invalid slugs

ahsu1230:
@andrewconner @2is10 

This will reject an add library or modify library request by returning a "invalid_slug" error to the FE

I think this would be a decent fix until we decide what to do with auto-generated slugs. Or maybe we'll need better error messages on the FE for all platforms?

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/model/Library.scala
+++ b/kifi-backend/modules/common/app/com/keepit/model/Library.scala
@@ -154,8 +154,10 @@ object LibrarySlug {
     Format(__.read[String].map(LibrarySlug(_)), new Writes[LibrarySlug] { def writes(o: LibrarySlug) = JsString(o.value) })
 
   val MaxLength = 50
+  val invalidSlugSet = Set("libraries", "connections", "followers", "keeps", "tags")
+
   def isValidSlug(slug: String): Boolean = {
-    slug != "" && !slug.contains(' ') && slug.length <= MaxLength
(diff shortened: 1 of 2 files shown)


2is10:
It’d be good to fix the slug-generation code at the same time. How about just capitalizing the first letter?

2is10:
We should probably differentiate reserved slugs from (syntactically) invalid slugs in the UI. I guess it doesn’t matter what the server returns as long as clients are smart enough to handle reserved slugs properly. Still, we might consider a `"reserved_slug"` error code instead, for clarity.

2is10:
Thanks for working on this.

kifi-backend/modules/common/app/com/keepit/model/Library.scala
```diff
@@ -154,10 +154,16 @@ object LibrarySlug {
     Format(__.read[String].map(LibrarySlug(_)), new Writes[LibrarySlug] { def writes(o: LibrarySlug) = JsString(o.value) })
 
   val MaxLength = 50
+  val invalidSlugSet = Set("libraries", "connections", "followers", "keeps", "tags", "main", "secret")
```
ahsu1230, line 157: Here are the list (7) of reserved slugs that we should prevent users from creating/modify libraries with these urls.
@jaredpetker @DerekBurch @mrbrown14 @2is10 @thassss

2is10:
:+1:

PR #14345, opened 2015-03-13
beginning to unwind/simplify KifiSiteRouter

2is10:
@andrewconner 

As `KifiSiteRouter` has become more complex, it strikes me that the Play framework is better at routing clearly and efficiently than we are. Here I’ve factored out some of the redirects into the shoebox routes file and introduced private action filters for cross-cutting concerns.

I intend to remove a bit more of the branching logic and then add a mechanism for handling email links differently for web and mobile.

### DIFF
--- a/kifi-backend/modules/shoebox/test/com/keepit/controllers/website/KifiSiteRouterTest.scala
+++ b/kifi-backend/modules/shoebox/test/com/keepit/controllers/website/KifiSiteRouterTest.scala
@@ -6,7 +6,7 @@ import com.keepit.commanders.{ LibraryCommander, UserCommander }
 import com.keepit.common.concurrent.FakeExecutionContextModule
 import com.keepit.common.controller.{ FakeUserActionsHelper, UserRequest, NonUserRequest }
 import com.keepit.common.crypto.FakeCryptoModule
-
+import com.keepit.common.db.Id
 import com.keepit.common.mail.{ EmailAddress, FakeMailModule }
 import com.keepit.common.social.FakeSocialGraphModule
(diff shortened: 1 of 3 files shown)


andrewconner:
Interesting. Yeah it had grown a lot. I'll take a close look at this, but like the idea.

kifi-backend/modules/shoebox/test/com/keepit/controllers/website/KifiSiteRouterTest.scala
```diff
@@ -133,36 +176,29 @@ class KifiSiteRouterTest extends Specification with ShoeboxTestInjector {
...
-        val request = FakeRequest("GET", "/some/path?kma=1").withHeaders("user-agent" -> "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.8; rv:20.0) Gecko/20100101 Firefox/20.0")
-        val result = inject[KifiSiteRouter].app("some/path?kma=1")(request)
-        status(result) must equalTo(404);
-      }
-    }
+        { // ignoring flag with no mobile
+          val request = FakeRequest("GET", "/some/path?kma=1").withHeaders("user-agent" -> "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.8; rv:20.0) Gecko/20100101 Firefox/20.0")
+          val result = inject[KifiSiteRouter].app("some/path?kma=1")(request)
+          status(result) must equalTo(404)
+        }
```
2is10, line 194: I moved these into the larger test method above because `withDb` no longer works now that this is a `ShoeboxApplication` test and I don’t want to create an entire application for each of these small test cases.

PR #14347, opened 2015-03-13
more progess on KifiSiteRouter: friends, connections, invite pages

2is10:
@andrewconner 

Only user profile pages and libraries remain (still handled by `AngularRouter`). I may stop here for now.


kifi-backend/modules/shoebox/test/com/keepit/controllers/website/KifiSiteRouterTest.scala
```diff
@@ -101,6 +101,20 @@ class KifiSiteRouterTest extends Specification with ShoeboxApplicationInjector {
             } getOrElse result(false, "", "request did not match any routes", x)
           }
         }
+        object beWebApp extends Matcher[Option[Future[Result]]] {
```
andrewconner, line 104: Nice, great test tool

PR #14348, opened 2015-03-13
join Library now takes in optional authToken for joining a private libra...

ahsu1230:
...ry (from email)

@andrewconner


ahsu1230:
merging so can work on in the FE

PR #14352, opened 2015-03-13
add authToken to calling join library (if defined - will be defined if o...

ahsu1230:
...pening up library link from an email)

@andrewconner


kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -955,9 +955,11 @@ class LibraryCommander @Inject() (
   def joinLibrary(userId: Id[User], libraryId: Id[Library], authToken: Option[String] = None, hashedPassPhrase: Option[HashedPassPhrase] = None)(implicit eventContext: HeimdalContext): Either[LibraryFail, Library] = {
     val (lib, listInvites) = db.readOnlyMaster { implicit s =>
       val lib = libraryRepo.get(libraryId)
-      val listInvites = if (authToken.isDefined) {
+      val listInvites = if (lib.visibility != LibraryVisibility.PUBLISHED && authToken.isDefined) { // private library & auth token (opened by email) requires checking of pass phrase
         getValidLibInvitesFromAuthTokenAndPassPhrase(libraryId, authToken, hashedPassPhrase)
```
ahsu1230, line 959: only need to verify hashed pass phrase for private libraries & opened up by email (auth token).
Otherwise, if auth token is defined, the invite may be to an email. In this case, look for invites with that auth token. If auth token is not defined, the invite is opened in kifi. In that case, look for invites by user Id.

andrewconner:
Looks good. Will help you test in prod if you'd like.

PR #14353, opened 2015-03-13
Fixing non-interpolated string

andrewconner:
@2is10


2is10:
Yikes! Thanks.

2is10:
Maybe we should have a build warning or test that examines source code for these.

andrewconner:
Yes! I have a slew of dev improvements I want to do. Scala actually has several great linting libraries that "fix" rough edges of the language.

PR #14355, opened 2015-03-13
allowing functions to be defined below where they are used

2is10:
I find it useful to sometimes declare “private” helper functions lower in a scope, so they can be motivated by some higher-level code that calls them before the reader encounters them and (perhaps) feels compelled to read them.

@andrewconner @stkem @ahsu1230


andrewconner:
Works for me

PR #14356, opened 2015-03-13
showing the search subtitle in library searches (e.g. number of results, no results)

2is10:
number of results, no results message, and bulk edit link + commands


2is10:
I’m pretty sure these used to be shown. I may have removed them at some point by accident.

PR #14364, opened 2015-03-16
more efficient filtering on libraries, using last_kept value, less db ca...

eishay:
...lls and better use of cache


eishay:
@leogrim @andrewconner @stkem  since we use more of `library.last_kept` and I think that keep count is a good idea, should we have an integrity check once a day on those numbers?

PR #14369, opened 2015-03-16
Defaulting when no connections returned (?), ?!? utility

andrewconner:
@eishay Not sure if this is the best solution, but I see there's been quite a few user-facing exceptions since it was refactored. https://fortytwo.airbrake.io/projects/91268/groups/1155225456510258989

Let me know what you think.


andrewconner:
Merging to stop the exceptions, but feel free to comment. This seems hacky, but I can't tell why we'd get a NoSuchElementException otherwise.

PR #14370, opened 2015-03-16
keep attribution

yingjieMiao:
@andrewconner 

some init work. comments/suggestions?


kifi-backend/modules/common/app/com/keepit/model/KeepSourceAttribution.scala
```diff
@@ -0,0 +1,55 @@
...
+import com.keepit.common.time._
+import org.joda.time.DateTime
+import play.api.libs.json._
+
+case class KeepSourceAttribution(
+    id: Option[Id[KeepSourceAttribution]] = None,
+    createdAt: DateTime = currentDateTime,
+    updatedAt: DateTime = currentDateTime,
+    attributionType: KeepAttributionType,
+    attributionJson: Option[JsValue],
```
andrewconner: Why not type this directly?

andrewconner: ie, instead of `attributionType` and `attributionJson`, just have a `attribution: SourceAttribution` field.

andrewconner:
Left a couple comments. Besides that, good start :+1:

kifi-backend/modules/common/app/com/keepit/model/KeepSourceAttribution.scala
yingjieMiao: i put the type as a column so one can do query about it. Since the attribution json has different structures, if i type it directly, the type has to go into the stored json as well. kind of duplicated.

yingjieMiao:
@andrewconner 

updated and added evolution script. `KeepSourceAttribution` is revised. 
Please let me know if this works better.

andrewconner:
Nice!

andrewconner:
Looks good to me. Next up, integration with the importer :)

kifi-backend/modules/common/app/com/keepit/model/KeepSourceAttribution.scala
```diff
@@ -0,0 +1,75 @@
...
+import com.keepit.common.db.{ States, ModelWithState, State, Id }
+import com.keepit.common.time._
+import org.joda.time.DateTime
+import play.api.libs.json._
+
+case class KeepSourceAttribution(
+    id: Option[Id[KeepSourceAttribution]] = None,
+    createdAt: DateTime = currentDateTime,
+    updatedAt: DateTime = currentDateTime,
+    attribution: SourceAttribution,
```
eishay, line 12: As with the `Keep#keptAt: DateTime` we may want the original time when the attribution happened. E.g. when the tweet actually happened which is not the `createdAt` time.
Not sure if if there's a case for us to query by it, thoughts? @yingjieMiao @andrewconner @stkem

andrewconner, line 12: Hm, wouldn't this be just keptAt?

eishay, line 12: yes

andrewconner, line 12: Mind grabbing me and @yingjieMiao? Don't understand, the time is already in `Keep`, and there's a direct connection from KeepSourceAttribution to Keep.

PR #14373, opened 2015-03-17
Introduce FetchedDocument, RoverDocumentFetcher, non-blocking YoutubeArticleFetcher

leogrim:
@stkem @andrewconner @eishay @yingjieMiao


kifi-backend/modules/rover/app/com/keepit/rover/article/YoutubeArticleFetcher.scala
```diff
@@ -0,0 +1,149 @@
...
+    }
+  }
+
+  private def getVideo(doc: JsoupDocument): Future[YoutubeVideo] = {
+    val futureTracks = getTracks(doc)
+    val headline = Option(doc.doc.getElementById("watch-headline-title")).map(_.text).filter(_.nonEmpty)
+    val description = doc.doc.select("#watch-description-text .content").text
+    val channel = doc.doc.select("#watch7-user-header .yt-user-name").text()
+    val tags = doc.doc.select("#watch-description-extras .content .watch-info-tag-list").map(_.text()).filter(_.nonEmpty)
+    val viewCount = doc.doc.select("#watch7-views-info .watch-view-count").text().trim.replaceAllLiterally(",", "").toInt
```
yingjieMiao, line 71: `toInt` may not be safe overtime.

yingjieMiao:
would be nice if we could have a simple test for `YoutubeArticleFetcher`.

lgtm overall. :+1:  But I didn't scrutinize.

kifi-backend/modules/rover/app/com/keepit/rover/article/YoutubeArticleFetcher.scala
leogrim, line 71: True, though none of this is really safe overtime since it relies on Youtube's internal HTML and APIs :)

2is10, line 71: Still, tests help document the original programmer’s intent. If there’s a problem someday, for example, it would be nice to be able to compare the HTML we’re getting live from YouTube versus what’s in a test, to help us figure out what’s different and to help ensure backwards compatibility (or cross-compatibility with various A/B tests) is maintained.

To be clear, I’m not pressuring you to write a test right now. You know your own tasks and priorities and the time it would cost to write one much better than I.

leogrim, line 71: No you're both right we very much need a test, given how edgy this is. I'll add a task to the Rover project so that I don't forget about it.

PR #14384, opened 2015-03-17
adapting shoebox cluster size alerts to the new reality

stkem:
Turns out we lost three out of five shoeboxen this morning! (Likely related to the too many db connections alerts we got this morning) 

The cluster size alerts were still configured for a cluster of 2. I've changed it to reflect the new cluster size.


stkem:
@andrewconner @eishay

PR #14386, opened 2015-03-17
generate keepSourceAttribution from RawKeep

yingjieMiao:
@andrewconner @stkem 

note I add a column to bookmark table. Cache version is not touched.


kifi-backend/modules/common/conf/evolutions/shoebox/304.sql
```diff
@@ -0,0 +1,11 @@
+# SHOEBOX
+
+# --- !Ups
+
+alter table bookmark
+    add column source_attribution_id bigint(20) NULL;
```
stkem, line 6: foreign key?

stkem:
lgtm. we need to be careful with this table alter, do to the size. Better do it in the evening.

kifi-backend/modules/common/conf/evolutions/shoebox/304.sql
yingjieMiao, line 6: good catch

yingjieMiao:
I agree with you on the size concern. On the other hand, the table has some manual written sql queries plus  db-row-to-object mapping, i.e., things like (`keptAt = r.<<[DateTime]`). If this kind of stuff goes wrong in the night...

Anyway, I will do some localhost test to be double sure.

kifi-backend/modules/common/app/com/keepit/model/Keep.scala
```diff
@@ -30,7 +30,8 @@ case class Keep(
     kifiInstallation: Option[ExternalId[KifiInstallation]] = None,
     seq: SequenceNumber[Keep] = SequenceNumber.ZERO,
     libraryId: Option[Id[Library]],
-    keptAt: DateTime = currentDateTime) extends ModelWithExternalId[Keep] with ModelWithState[Keep] with ModelWithSeqNumber[Keep] {
+    keptAt: DateTime = currentDateTime,
+    sourceAttributionId: Option[Id[KeepSourceAttribution]] = None) extends ModelWithExternalId[Keep] with ModelWithState[Keep] with ModelWithSeqNumber[Keep] {
```
andrewconner, line 34: Thanks for adding it to the end of the class. Makes it easier.

kifi-backend/modules/shoebox/app/com/keepit/model/RawKeep.scala
```diff
@@ -27,6 +27,17 @@ case class RawKeep(
   def withUpdateTime(now: DateTime) = this.copy(updatedAt = now)
 }
 
+object RawKeep {
+  def extractKeepSourceAttribtuion(keep: RawKeep): Option[KeepSourceAttribution] = {
+    keep.source match {
+      case KeepSource.twitterFileImport | KeepSource.twitterSync =>
+        val attrOpt = keep.originalJson.flatMap(js => TwitterAttribution.fromRawTweetJson(js))
+        attrOpt.map { attr => KeepSourceAttribution(attribution = attr) }
```
andrewconner, line 37: Mind logging if it fails? No alert needed, but we'd like to have some mechanism of knowing if it's not working.

andrewconner:
Thanks for adding the tests. Looks good to me.

yingjieMiao:
added foreign key and logging

PR #14393, opened 2015-03-17
increasing reco generation throughput

stkem:
@yingjieMiao


yingjieMiao:
hope it helps. 

FYI, user 273's seqNum is still at 20K , very far behind.

PR #14394, opened 2015-03-17
launching search suggestions for web app

2is10:
@andrewconner


kifi-backend/modules/common/app/com/keepit/model/UserValue.scala
```diff
@@ -25,7 +25,7 @@ case class UserValue(
 }
 
 case class UserValueKey(userId: Id[User], key: UserValueName) extends Key[String] {
-  override val version = 4
+  override val version = 5
```
2is10, line 28: b/c I’m using the database to set `site_notify_libraries_in_search` for everyone

andrewconner:
Wahoo!

PR #14402, opened 2015-03-18
addedAt field from chrome & firefox should be set for RawKeep.createdDat...

ahsu1230:
...e

@andrewconner 
From my understanding, the JSON the Extension is already sending a json with field `addedAt`. We simply look for that field in `RawKeepFactory.toRawKeep`, set it for `createdDate`. `createdDate` gets set for `RawBookmarkRepresentation.keptAt` in `RawKeepImporterPlugin.processBatch`, which eventually translates to `Keep.keptAt`

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/model/RawKeep.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/model/RawKeep.scala
@@ -46,9 +46,10 @@ class RawKeepFactory @Inject() (airbrake: AirbrakeNotifier) {
     val title = (json \ "title").asOpt[String]
     val url = (json \ "url").asOpt[String].getOrElse(throw new Exception(s"json $json did not have a url"))
     val isPrivate = (json \ "isPrivate").asOpt[Boolean].getOrElse(true)
+    val addedAt = (json \ "addedAt").asOpt[DateTime]
     val canonical = (json \ Normalization.CANONICAL.scheme).asOpt[String]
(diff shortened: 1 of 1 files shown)


ahsu1230:
@stkem

stkem:
lgtm

kifi-backend/modules/shoebox/app/com/keepit/model/RawKeep.scala
```diff
@@ -46,9 +46,10 @@ class RawKeepFactory @Inject() (airbrake: AirbrakeNotifier) {
     val title = (json \ "title").asOpt[String]
     val url = (json \ "url").asOpt[String].getOrElse(throw new Exception(s"json $json did not have a url"))
     val isPrivate = (json \ "isPrivate").asOpt[Boolean].getOrElse(true)
+    val addedAt = (json \ "addedAt").asOpt[DateTime]
```
andrewconner, line 49: Does Firefox use "ADD_DATE" instead? (re: your asana post)

ahsu1230, line 49: chrome and firefox both use ADD_DATE in the <a> tag for every exported link

2is10, line 49: This code looks correct to me. The browser extension sends `"addedAt"`.

andrewconner, line 49: Great

PR #14405, opened 2015-03-18
hook up

yingjieMiao:
@andrewconner @stkem 

To my surprise, `RawBookmarkRepresentation` is used in a few controllers and need json serialization. The serialization of the trait `SourceAttribution` is a bit annoying and perhaps not really needed. So I ignored that part.


andrewconner:
Hm, we should clean that up. Specifically, it's supposed to be an input class only — no writes. Not sure how it got elsewhere, but we should fight against it. We've had problems with it before.

yingjieMiao:
the writes is only used in `LibraryControllerTest`

andrewconner:
Oh, let's remove it then, and just manually create that object if needed.

PR #14406, opened 2015-03-18
darker search suggestion menu item highlight

2is10:


kifi-backend/modules/shoebox/angular/src/header/searchSuggest.styl
```diff
@@ -44,7 +44,7 @@ iconAreaWidth = sidePad + 54px
   position: relative
 
   &.kf-selected
-    background-color: #f3faf2
+    background-color: tint(libraryGreen, 84%)
```
2is10, line 47: 84% is 2 × 42%

PR #14417, opened 2015-03-18
adding signature logic to registering mobile devices

ahsu1230:
@eishay @stkem @andrewconner


kifi-backend/modules/common/conf/evolutions/shoebox/305.sql
```diff
@@ -0,0 +1,11 @@
+# ELIZA
+
+# --- !Ups
+
+alter table device add column signature varchar(40) NULL;
+
+create unique index device_u_user_id_device_type_signature on device (user_id, device_type, signature);
```
ahsu1230, line 7: creating an unique index on a nullable column should be able to accept multiple null columns
http://stackoverflow.com/questions/3712222/does-mysql-ignore-null-values-on-unique-constraints

I also put the constraint on `device_type` just to be super sure that ios & android stay separated. JP actually mentioned that it could be possible for a user to change their signature somehow (by editing their mobile device name or something)

kifi-backend/modules/eliza/app/com/keepit/eliza/controllers/mobile/MobileDevicesController.scala
```diff
@@ -9,9 +9,13 @@ import play.api.libs.json._
...
   def registerDevice(deviceType: String) = UserAction(parse.tolerantJson) { implicit request =>
-    (request.body \ "token").asOpt[String] map { token =>
-      val isDev: Boolean = (request.body \ "dev").asOpt[Boolean].exists(x => x)
-      val device = urbanAirship.registerDevice(request.userId, token, DeviceType(deviceType), isDev)
+    val jsonBody = request.body
+    val tokenOpt = (jsonBody \ "token").asOpt[String]
+
+    tokenOpt map { token =>
+      val isDev: Boolean = (jsonBody \ "dev").asOpt[Boolean].exists(x => x)
+      val signatureOpt = (jsonBody \ "signature").asOpt[String].orElse((jsonBody \ "deviceId").asOpt[String]) // todo (aaron & mobile): right now iOS sends "deviceId", would be better & clear to make it "signature"
```
ahsu1230, line 17: @jaredpetker @DerekBurch if you guys could send `signature`, that would be great.
@thassss @mrbrown14 I know iOS is sending `deviceId` right now, but it would be clearer on the backend if you guys eventually switched over to `signature` =D

kifi-backend/modules/eliza/app/com/keepit/realtime/UrbanAirship.scala
```diff
@@ -59,16 +59,45 @@ class UrbanAirshipImpl @Inject() (
     clock: Clock,
     scheduler: Scheduler) extends UrbanAirship with Logging {
 
-  def registerDevice(userId: Id[User], token: String, deviceType: DeviceType, isDev: Boolean): Device = synchronized {
-    log.info(s"Registering device: $token (user $userId)")
-    val device = db.readWrite { implicit s =>
-      deviceRepo.get(token, deviceType).map { d =>
-        if (d.userId != userId) deviceRepo.save(d.copy(state = DeviceStates.INACTIVE))
+  def registerDevice(userId: Id[User], token: String, deviceType: DeviceType, isDev: Boolean, signatureOpt: Option[String]): Device = synchronized {
```
ahsu1230, line 62: it would be great if I had some review on this logic:
When a user registers a device (ios for example)....
if a signature is provided, 
- deactivate all user's ios devices without a signature. 
- if `(user, signature)` exists, update device token. 
- If `(user, signature)` does not exist, new device row for user!

if a signature is not provided, use most of the old logic.
- deactivate any `(token, ios)` that's not associated with the current user
- find device with `(user, token, ios)`. If exists, update with new token!
- if does not exist, create new device row.

Hopefully, eventually we can stop using the latter if block (no signature provided)

ahsu1230:
reviewed with @eishay !

PR #14418, opened 2015-03-18
remove deprecated fields

yingjieMiao:
@andrewconner @eishay @stkem 

`KeepInfo` has too many fields and Play Json macro has problem with it if I add one more field.


yingjieMiao:
confirmed with mobile team. they are not using these.

andrewconner:
Thanks Yingjie for checking with them!

eishay:
great!

PR #14422, opened 2015-03-18
include source attribution in keepInfo

yingjieMiao:
@andrewconner @stkem


stkem:
lgtm. @andrewconner should probably have a quick look

andrewconner:
Looks good. We may want TwitterAttribution to store (or serialize) the id as `id` instead of `idString` (it's typed String, there's no confusion)

PR #14426, opened 2015-03-18
restoring edit controls after undoing a bulk unkeep of all keeps in a library

2is10:


kifi-backend/modules/shoebox/angular/src/keeps/keeps.js
```diff
@@ -132,29 +132,26 @@ angular.module('kifi')
...
+            undoService.add(selectedKeeps.length > 1 ? selectedKeeps.length + ' keeps deleted' : 'keep deleted', function () {
+              keepActionService.keepToLibrary(_.map(selectedKeeps, function (keep) { return _.pick(keep, 'url', 'title'); }), libraryId).then(function () {
+                _.forEach(selectedKeeps, function (keep) {
+                  keep.makeKept();
+                });
+
+                libraryService.addToLibraryCount(libraryId, selectedKeeps.length);
+                scope.selection.selectAll(selectedKeeps);
+
+                scope.availableKeeps = _.reject(scope.keeps, {unkept: true});
```
2is10, line 152: This line was the fix. Just moved the rest of the code so the parallel structure of the do/undo parts is more apparent.

PR #14430, opened 2015-03-18
allow unkeep any keep that is not inactive (active or duplicate)

ahsu1230:
@andrewconner @stkem 

the simple fix for our duplicates problem. The more complex, we can put in lower priority.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
@@ -344,7 +344,7 @@ class KeepsCommander @Inject() (
         val (keeps, invalidKeepIds) = db.readWrite { implicit s =>
           val (keeps, invalidKeepIds) = keepIds.map { kId =>
             keepRepo.getByExtIdandLibraryId(kId, libId, excludeState = None) match {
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
```diff
@@ -344,7 +344,7 @@ class KeepsCommander @Inject() (
         val (keeps, invalidKeepIds) = db.readWrite { implicit s =>
           val (keeps, invalidKeepIds) = keepIds.map { kId =>
             keepRepo.getByExtIdandLibraryId(kId, libId, excludeState = None) match {
-              case Some(k) if (k.isActive) =>
+              case Some(k) if k.state != KeepStates.INACTIVE =>
                 keepsToFinalize = k +: keepsToFinalize
                 Left(setKeepStateWithSession(k, KeepStates.INACTIVE, userId))
```
ahsu1230, line 349: `setKeepStateWithSession` is the part that saves the keep as `INACTIVE`

stkem:
lgtm

stkem:
please make sure it works on kifi.com when this is out

ahsu1230:
yup will do

PR #14432, opened 2015-03-19
only active keeps & keep counts show up for a library or user

ahsu1230:
@andrewconner @stkem

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/model/Keep.scala
+++ b/kifi-backend/modules/common/app/com/keepit/model/Keep.scala
@@ -167,7 +167,7 @@ object KeepUriAndTime {
 }
 
 case class KeepCountKey(userId: Id[User]) extends Key[Int] {
-  override val version = 3
+  override val version = 4
   val namespace = "bookmark_count"
   def toKey(): String = userId.toString
 }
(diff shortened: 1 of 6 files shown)


kifi-backend/modules/common/app/com/keepit/model/Keep.scala
```diff
@@ -194,7 +194,7 @@ class KeepUriUserCache(stats: CacheStatistics, accessLog: AccessLog, innermostPl
   extends JsonCacheImpl[KeepUriUserKey, Keep](stats, accessLog, innermostPluginSettings, innerToOuterPluginSettings: _*)
 
 case class CountByLibraryKey(id: Id[Library]) extends Key[Int] {
-  override val version = 0
+  override val version = 1
```
ahsu1230, line 197: counts in some libraries and for some users have changed. caches need to be invalidated

kifi-backend/modules/shoebox/app/com/keepit/model/KeepRepo.scala
```diff
@@ -44,14 +44,13 @@ trait KeepRepo extends Repo[Keep] with ExternalIdColumnFunction[Keep] with SeqNu
   def getByTitle(title: String)(implicit session: RSession): Seq[Keep]
   def exists(uriId: Id[NormalizedURI])(implicit session: RSession): Boolean
   def getSourcesByUser()(implicit session: RSession): Map[Id[User], Seq[KeepSource]]
-  def oldestKeep(userId: Id[User], excludeState: Option[State[Keep]] = Some(KeepStates.INACTIVE))(implicit session: RSession): Option[Keep]
```
ahsu1230, line 47: unused

PR #14443, opened 2015-03-19
Importing bookmarks with tags (by path and optional param)

ahsu1230:
@andrewconner @2is10


kifi-backend/modules/shoebox/app/com/keepit/model/RawKeep.scala
```diff
@@ -52,14 +56,41 @@ class RawKeepFactory @Inject() (airbrake: AirbrakeNotifier) {
...
-    val canonical = (json \ Normalization.CANONICAL.scheme).asOpt[String]
-    val openGraph = (json \ Normalization.OPENGRAPH.scheme).asOpt[String]
-    RawKeep(userId = userId, title = title, url = url, isPrivate = isPrivate, importId = importId, source = source, originalJson = Some(json), installationId = installationId, libraryId = libraryId, createdDate = addedAt)
+  def toRawKeep(userId: Id[User], source: KeepSource, value: JsValue, importId: Option[String] = None, installationId: Option[ExternalId[KifiInstallation]] = None, libraryId: Option[Id[Library]])(implicit context: HeimdalContext): Seq[RawKeep] = {
+    getBookmarkJsonObjects(value) map { json =>
+      val title = (json \ "title").asOpt[String]
+      val url = (json \ "url").asOpt[String].getOrElse(throw new Exception(s"json $json did not have a url"))
+      val isPrivate = (json \ "isPrivate").asOpt[Boolean].getOrElse(true)
+      val addedAt = (json \ "addedAt").asOpt[DateTime]
+      val pathOpt = (json \ "path").asOpt[String]
```
2is10: I’m sending the path as an array of (non-empty) segments, so less work is needed here.

kifi-backend/modules/shoebox/app/com/keepit/model/RawKeep.scala
```diff
@@ -40,7 +42,9 @@ object RawKeep extends Logging {
   }
 }
 
-class RawKeepFactory @Inject() (airbrake: AirbrakeNotifier) {
+class RawKeepFactory @Inject() (
+    keepCommander: KeepsCommander,
```
2is10: It feels weird to be injecting a commander into a `com.keepit.model` class. Can we refactor to avoid this?

I recommend changing `tagIds: String` to `tags: Set[Hashtag]`. It looks like `RawKeepImporterPlugin` is the only consumer of `RawKeep` and I’m guessing it would be pretty straightforward to switch it from using `Id` to using `Hashtag`. (Database lookups aren’t very “raw”.)

2is10:
Nice work. Users will love this.

kifi-backend/modules/shoebox/app/com/keepit/model/RawKeep.scala
ahsu1230: It seems like there is a `RawKeepRepo` that has `tagIds` field as an array of collection ids and there are a bunch in it already.

andrewconner: Yeah, it was an old oversight on my part. We currently persist tag ids, and apply the tag to the keep — the original idea was to make keep creation faster, since we wouldn't need to look up if they have a tag already for every tag. However, Jared's right, it's better if we persisted a set of hashtags. I've wanted to fix this, but it's been pretty low priority.

Go for it if you want. It'll speed up loading urls into rawkeep quite a bit. You'll likely need to create a new column in rawkeep for it to work, though.

2is10: Ugh. This change alone probably doesn’t warrant that much work, but it definitely needs to be done.

It looks like `keepCommander.getOrCreateTag` is using `readOnlyReplica` and so won’t ever result in new tags getting cached. Imports will likely be resulting in a lot of tag creation, and a lot of tag reuse. We need to make sure we’re both writing to and reading from an in-memory tag cache either here (a simple `Map[String, Id[Collection]]`) or at the repo level.

andrewconner: Yep, that's why BookmarkImporter uses only sets of tags, instead of duplicating work. Doing that is pretty much necessary for large imports.

kifi-backend/modules/shoebox/app/com/keepit/model/RawKeep.scala
```diff
@@ -52,14 +56,41 @@ class RawKeepFactory @Inject() (airbrake: AirbrakeNotifier) {
...
+      val isPrivate = (json \ "isPrivate").asOpt[Boolean].getOrElse(true)
+      val addedAt = (json \ "addedAt").asOpt[DateTime]
+      val pathOpt = (json \ "path").asOpt[Seq[String]]
+      val tagsOpt = (json \ "tags").asOpt[Seq[String]]
+
+      // add tags to bookmark if it has a path or pre-tagged
+      val tagSet = scala.collection.mutable.Set.empty[String]
+      tagsOpt.map { tags =>
+        tags.foreach { t =>
+          tagSet.add(t)
```
andrewconner: Actually (and BookmarkImporter has this problem too), mind only creating tags in one case variant? ie, "JavaScript" is the same as "javascript". You can do this by making it a Map, keyed by t.toLowerCase.

yingjieMiao: could be something like the following?

```
val tagSet = tagsOpt.map{ _.toSet} getOrElse Set()
```

Probably not much performance impact.

kifi-backend/modules/eliza/app/com/keepit/realtime/DeviceRepo.scala
```diff
@@ -28,8 +28,8 @@ class DeviceRepoImpl @Inject() (val db: DataBaseComponent, val clock: Clock) ext
     def token = column[String]("token", O.NotNull)
     def deviceType = column[DeviceType]("device_type", O.NotNull)
     def isDev = column[Boolean]("is_dev", O.NotNull)
-    def signature = column[Option[String]]("signature", O.Nullable)
-    def * = (id.?, userId, token, deviceType, state, createdAt, updatedAt, isDev, signature) <> ((Device.apply _).tupled, Device.unapply _)
+    def signature = column[String]("signature", O.Nullable)
+    def * = (id.?, userId, token, deviceType, state, createdAt, updatedAt, isDev, signature.?) <> ((Device.apply _).tupled, Device.unapply _)
```
ahsu1230, line 32: our arthur building has been failing on DeviceRepo sporatically, hoping to fix it with some changes here and there

kifi-backend/modules/shoebox/app/com/keepit/model/RawKeep.scala
```diff
@@ -52,14 +57,35 @@ class RawKeepFactory @Inject() (airbrake: AirbrakeNotifier) {
...
+        }
+      }
+      pathOpt.map { pathSegments =>
+        pathSegments.filter(_.nonEmpty).map(tagSet.add(_))
+      }
+      val hashTags = if (tagSet.nonEmpty) {
+        Some(tagSet.mkString("::"))
```
ahsu1230, line 80: this could be better, but currently hashtags can allow `,` and many other kinds of punctuations. So I decided with a character combo that is probably not used very often.

kifi-backend/modules/shoebox/app/com/keepit/commanders/RawKeepImporterPlugin.scala
```diff
@@ -134,9 +134,23 @@ private class RawKeepImporterActor @Inject() (
...
           successes.foreach { keep =>
             val allTagIdsForThisKeep = rawKeepByUrl.get(keep.url).flatMap { rk =>
-              rk.tagIds.map { tags =>
+
+              val keepTagIds = rk.tagIds.map { tags =>
                 tags.split(",").toSeq.filter(_.length > 0).map { c => Try(c.toLong).map(Id[Collection]).toOption }.flatten
               }
+
+              val keepHashtagIds = rk.hashtags.map { hashtags =>
+                hashtags.split("::").toSeq.filter(_.nonEmpty).map { hashtag =>
```
ahsu1230, line 143: look at notes below

kifi-backend/modules/shoebox/app/com/keepit/model/RawKeep.scala
andrewconner, line 80: Why not just use a `JsArray`? Just throw the strings into that.

kifi-backend/modules/shoebox/app/com/keepit/model/RawKeep.scala
```diff
@@ -52,14 +57,35 @@ class RawKeepFactory @Inject() (airbrake: AirbrakeNotifier) {
...
+      val isPrivate = (json \ "isPrivate").asOpt[Boolean].getOrElse(true)
+      val addedAt = (json \ "addedAt").asOpt[DateTime]
+      val pathOpt = (json \ "path").asOpt[Seq[String]]
+      val tagsOpt = (json \ "tags").asOpt[Seq[String]]
+
+      // add tags to bookmark if it has a path or pre-tagged
+      val tagSet = scala.collection.mutable.Set.empty[String]
+      tagsOpt.map { tags =>
+        tags.foreach { t =>
+          tagSet.add(t.toLowerCase)
```
andrewconner, line 73: Hm, I didn't really mean this. Someone may want mixed case tags. The point was only allow one (so "Javascript" and "JavaScript" get consolidated into one)

kifi-backend/modules/shoebox/app/com/keepit/model/RawKeep.scala
```diff
@@ -52,14 +57,35 @@ class RawKeepFactory @Inject() (airbrake: AirbrakeNotifier) {
...
+        }
+      }
+      pathOpt.map { pathSegments =>
+        pathSegments.filter(_.nonEmpty).map(seg => tagSet.add(seg.toLowerCase))
+      }
+      val hashTags = if (tagSet.nonEmpty) {
+        Some(tagSet.mkString("::"))
```
andrewconner, line 80: Same as before. Why not just use JsArray? (I think we have a serializer from Seq[String] already, but if not, it's not hard to write)

kifi-backend/modules/common/conf/evolutions/shoebox/306.sql
```diff
@@ -0,0 +1,9 @@
+# SHOEBOX
+
+# --- !Ups
+
+alter table raw_keep add column hashtags text default null;
```
andrewconner, line 5: By the way, this is very expensive. Keep an eye on things, it'll take a few minutes probably.

kifi-backend/modules/shoebox/app/com/keepit/commanders/RawKeepImporterPlugin.scala
```diff
@@ -134,9 +134,23 @@ private class RawKeepImporterActor @Inject() (
...
+                  val tag = bookmarksCommanderProvider.get.getOrCreateTag(userId, hashtag)(context)
```
andrewconner, line 144: You will definitely want to have this cached. On the other side, in BookmarkImporter, caching happens because of set de-duplication. Here, it's iterating per keep.

kifi-backend/modules/eliza/app/com/keepit/realtime/DeviceRepo.scala
andrewconner, line 32: What's the failure?

kifi-backend/modules/common/conf/evolutions/shoebox/305.sql
```diff
@@ -2,7 +2,7 @@
 
 # --- !Ups
 
-alter table device add column signature varchar(40) NULL;
+alter table device add column signature varchar(64) NULL;
```
andrewconner, line 5: Is this already applied? You'll need to alter the current column (with `modify`)

kifi-backend/modules/shoebox/app/com/keepit/commanders/RawKeepImporterPlugin.scala
andrewconner, line 144: One lightweight strategy is to fetch / create all the tags you'll need for an entire `rawKeepGroup`. That's why there's that big (and unusual) `rawKeeps.groupBy` above. It's to take fields that are similar, and make it so we only have to query it once.

kifi-backend/modules/common/conf/evolutions/shoebox/305.sql
ahsu1230, line 5: yea its already applied. i used modify when making signature bigger

kifi-backend/modules/eliza/app/com/keepit/realtime/DeviceRepo.scala
ahsu1230, line 32: it looks like this, and happens every now and then
SQLException: : error persisting Device[id:None,user:1,type:android,dev:false](DbRepo.scala:85)
[error] com.keepit.common.db.slick.DbRepo$class.save(DbRepo.scala:85)
[error] com.keepit.realtime.DeviceRepoImpl.save(DeviceRepo.scala:20)
[error] com.keepit.realtime.UrbanAirshipImpl$$anonfun$5.apply(UrbanAirship.scala:85)
[error] com.keepit.realtime.UrbanAirshipImpl$$anonfun$5.apply(UrbanAirship.scala:84)

andrewconner, line 32: Weird. An issue where the table doesn't exist anymore? Typically we do options like you changed it to, but that would surprise me if it fixes it. I'll take a look too.

PR #14451, opened 2015-03-19
A few more curator performance tweaks

stkem:
See individual commits. This should hopefully do the trick for now.


stkem:
@eishay @yingjieMiao

kifi-backend/modules/curator/app/com/keepit/curator/model/LibraryRecommendation.scala
```diff
@@ -58,4 +60,45 @@ object LibraryRecommendationStates extends States[LibraryRecommendation]
        |si:$sizeScore%1.2f-
        |c:$contentScoreOrDefault%1.2f
      """.stripMargin.replace("\n", "").trim()
+
+  private def reducePrecision(x: Float): Float = {
+    (x * 10000).toInt.toFloat * 10000
+  }
```
yingjieMiao, line 66: did you mean /10000 in the end?

PR #14454, opened 2015-03-19
Filter out corrupted data

stkem:
I accidentally introduced some corrupted “allScores” data into the DB
with a very short lived previous PR (rolled back immediately and cause is fixed). This
will make _really_ sure it’s not surface to users.
@yingjieMiao @eishay

### DIFF
--- a/kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationRetrievalCommander.scala
+++ b/kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationRetrievalCommander.scala
@@ -111,8 +111,19 @@ class RecommendationRetrievalCommander @Inject() (
     def scoreReco(reco: UriRecommendation) =
       UriRecoScore(scoringStrategy.scoreItem(reco.masterScore, reco.allScores, reco.delivered, reco.clicked, reco.vote, more, recencyWeight), reco)
 
+    def badData(reco: UriRecommendation) = { //temporary because I corrupted a small fraction of the data. Can be removed after April 19th 2015. Data will have been reaped by then. -Stephen
+      reco.allScores.socialScore < 1000 &&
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationRetrievalCommander.scala
```diff
@@ -111,8 +111,19 @@ class RecommendationRetrievalCommander @Inject() (
...
+      reco.allScores.socialScore < 1000 &&
+        reco.allScores.popularityScore < 1000 &&
+        reco.allScores.overallInterestScore < 1000 &&
+        reco.allScores.recentInterestScore < 1000 &&
+        reco.allScores.recencyScore < 1000 &&
+        reco.allScores.priorScore < 1000 &&
+        reco.allScores.rekeepScore < 1000 &&
+        reco.allScores.discoveryScore < 1000
+    }
+
```
yingjieMiao, line 124: hmm, your change was `x  => x * 10000 * 10000`. 
If all original subscores are less than `1e-5`, they are going to pass this guard. 
It's probably not critical.

yingjieMiao, line 124: And looks like this method is really testing `IsGoodData`?

stkem, line 124: Ha, yeah, it's not the best. I was moving very quickly :-)
It's a stop gap. I'll make it better.

stkem, line 124: The name was mostly so I can write "filter badData" :-)

PR #14456, opened 2015-03-20
Introducing SQSPriorityQueue

leogrim:
@yingjieMiao @stkem @eishay @andrewconner 
Not really a Priority Queue but sort of, feel free to come up with a better name :)


kifi-backend/modules/common/app/com/keepit/common/queue/SQSPriorityQueue.scala
```diff
@@ -0,0 +1,28 @@
...
+
+import com.keepit.common.concurrent.FutureHelpers
+import com.keepit.common.math.ProbabilityDensityBuilder
+import com.kifi.franz.{ SQSMessage, SQSQueue }
+
+import scala.concurrent.{ Future, ExecutionContext }
+import scala.concurrent.duration.FiniteDuration
+import scala.util.Random
+
+class SQSPriorityQueue[T](weights: Map[SQSQueue[T], Double]) {
```
stkem: If you extend `SQSQueue` you'll get all the other retrieval methods for free (based on `nextBatchWithLock`)

stkem:
`ProbabilisticMultiQueue`?

kifi-backend/modules/common/app/com/keepit/common/queue/SQSPriorityQueue.scala
```diff
@@ -0,0 +1,28 @@
...
+  def nextBatchWithLock(n: Int, lockTimeout: FiniteDuration)(implicit ec: ExecutionContext): Future[Seq[SQSMessage[T]]] = {
+    val offset = offsetDensity.sample(Random.nextDouble()).get
+    // not using foldLeftWhile on purpose: it doesn't return the last accumulator, which may lock tasks without returning them
+    FutureHelpers.foldLeft(sortedQueues.drop(offset))(Seq.empty[SQSMessage[T]]) {
+      case (tasks, nextQueue) =>
+        val numberOfMissingTasks = n - tasks.length
+        if (numberOfMissingTasks <= 0) Future.successful(tasks)
+        else nextQueue.nextBatchWithLock(numberOfMissingTasks, lockTimeout).map { moreTasks => tasks ++ moreTasks }
+    }
+  }
```
yingjieMiao: `sortedQueue` is sorted from high weight to low weight. When fetch jobs, offset is sampled according to the weight, and it seems that the head of the queue is dropped more often than the tail. (Thus, a queue with heaviest weight will have least probability to sample from?)

Overall I am not very clear about the intend of this method (on a quick glance).

kifi-backend/modules/common/app/com/keepit/common/queue/SQSPriorityQueue.scala
```diff
@@ -0,0 +1,28 @@
...
+class SQSPriorityQueue[T](weights: Map[SQSQueue[T], Double]) {
+  private val sortedQueues = weights.keySet.toSeq.sortBy(-weights(_))
+  private val offsetDensity = {
+    val builder = new ProbabilityDensityBuilder[Int]()
+    sortedQueues.zipWithIndex.foreach { case (queue, index) => builder.add(index, weights(queue)) }
+    builder.build()
+  }
+  def nextBatchWithLock(n: Int, lockTimeout: FiniteDuration)(implicit ec: ExecutionContext): Future[Seq[SQSMessage[T]]] = {
+    val offset = offsetDensity.sample(Random.nextDouble()).get
+    // not using foldLeftWhile on purpose: it doesn't return the last accumulator, which may lock tasks without returning them
```
yingjieMiao: If the weights are not probabilities, `sample` could return `None` with a positive probability. Cause exceptions.

leogrim: I think the builder normalizes them.

kifi-backend/modules/common/app/com/keepit/common/queue/SQSPriorityQueue.scala
leogrim: Queues are first sampled according to their weight (I may use a `SortedSet` to make that clearer than manipulating array offsets). If that queue is exhausted, move to the next most important queue and so on.

kifi-backend/modules/common/app/com/keepit/common/queue/SQSPriorityQueue.scala
yingjieMiao: I see.

kifi-backend/modules/common/app/com/keepit/common/queue/SQSPriorityQueue.scala
leogrim: It's also asking for `fromString` and `toString`

leogrim: and you can't really send, it's more of a reader

leogrim:
`ProbabilisticMultiQueue` sounds pretty good

kifi-backend/modules/common/app/com/keepit/common/queue/SQSPriorityQueue.scala
yingjieMiao: `sortedQueues.drop(offset)`. The first queue is included only when offset = 0. The tail is almost always included. 

Did I miss anything?

leogrim: But you only get to the tail if the previous queues haven't filled the bucket first.
Let's say we have two queues with (90%, 10%).
10% of the time we'll hit the second queue and grab whatever we can from it.
90% of the time we'll hit the first queue, and only if it's exhausted go to the second queue.

Assuming queues are always full, we get the expected (90%, 10%) distribution. Only if the most important queue is empty does the less important one benefit from a probability "spillover". Basically the actual retrieval distribution depends on the underlying queue load distribution but more important queues do prevail at full load.

stkem:
lgtm

PR #14458, opened 2015-03-20
import bookmarks with tags

ahsu1230:
closed https://github.com/kifi/keepit/pull/14443 because I was confusing myself.

@andrewconner @2is10


kifi-backend/modules/shoebox/app/com/keepit/model/RawKeepRepo.scala
```diff
@@ -36,7 +36,8 @@ class RawKeepRepoImpl @Inject() (val db: DataBaseComponent, val clock: Clock) ex
     def tagIds = column[String]("tag_ids", O.Nullable) // Comma separated list of references to `collection.id`
     def libraryId = column[Id[Library]]("library_id", O.Nullable)
     def createdDate = column[DateTime]("created_date", O.Nullable)
-    def * = (id.?, userId, createdAt, updatedAt, url, title.?, isPrivate, importId.?, source, kifiInstallationId.?, originalJson, state, tagIds.?, libraryId.?, createdDate.?) <> ((RawKeep.apply _).tupled, RawKeep.unapply _)
+    def hashtags = column[JsArray]("hashtags", O.Nullable)
```
ahsu1230: a nullable `JsArray` which will be NULL if the array is empty

kifi-backend/modules/shoebox/app/com/keepit/model/RawKeep.scala
```diff
@@ -57,9 +58,30 @@ class RawKeepFactory @Inject() (airbrake: AirbrakeNotifier) {
...
+        tagMap += (seg.toLowerCase -> seg.trim)
+      )
+    }
+    tagsOpt.map { tags =>
+      tags.map { tag =>
+        tagMap += (tag.toLowerCase -> tag.trim)
+      }
+    }
+    val hashtags = if (tagMap.nonEmpty) {
+      Some(JsArray(tagMap.values.toSeq.map(JsString(_))))
```
ahsu1230: When using `JsString` (and I've tested with `Json.toJson`) quotes become persisted to the database, so I Get
["aa","bb","cc"] when I was hoping for [aa,bb,cc] so i don't have to remove the extra double quotes when reading from the DB. but looks like I have to?

kifi-backend/modules/shoebox/app/com/keepit/commanders/RawKeepImporterPlugin.scala
```diff
@@ -87,10 +87,22 @@ private class RawKeepImporterActor @Inject() (
...
-      case ((userId, importIdOpt, source, installationId, isPrivate, libraryId), rawKeepGroup) =>
-
+    rawKeeps.groupBy(rk => (rk.userId, rk.importId, rk.source, rk.installationId, rk.isPrivate, rk.libraryId, rk.hashtags)).map {
+      case ((userId, importIdOpt, source, installationId, isPrivate, libraryId, hashtags), rawKeepGroup) =>
         val context = importIdOpt.map(importId => getHeimdalContext(userId, importId)).flatten.getOrElse(HeimdalContext.empty)
+
+        val hashtagSet = rawKeepGroup.map { rk =>
+          rk.hashtags.map { tagArray =>
+            val tagNames = tagArray.value.map { j =>
+              Json.stringify(j).replace("\"", "")
```
ahsu1230: extra quotes need to be parsed

kifi-backend/modules/shoebox/app/com/keepit/commanders/RawKeepImporterPlugin.scala
```diff
@@ -134,9 +146,21 @@ private class RawKeepImporterActor @Inject() (
 
           successes.foreach { keep =>
             val allTagIdsForThisKeep = rawKeepByUrl.get(keep.url).flatMap { rk =>
-              rk.tagIds.map { tags =>
+              val tagIdsFromTags = rk.tagIds.map { tags =>
                 tags.split(",").toSeq.filter(_.length > 0).map { c => Try(c.toLong).map(Id[Collection]).toOption }.flatten
               }
+
+              val tagIdsFromHashtags = rk.hashtags.map { tagArray =>
+                tagArray.value.map(Json.stringify(_).replace("\"", "")).map { tagName =>
```
ahsu1230: same here

kifi-backend/modules/shoebox/app/com/keepit/commanders/RawKeepImporterPlugin.scala
```diff
@@ -87,10 +87,22 @@ private class RawKeepImporterActor @Inject() (
   private def processBatch(rawKeeps: Seq[RawKeep], reason: String): Unit = {
     log.info(s"[RawKeepImporterActor] Processing ($reason) ${rawKeeps.length} keeps")
 
-    rawKeeps.groupBy(rk => (rk.userId, rk.importId, rk.source, rk.installationId, rk.isPrivate, rk.libraryId)).map {
-      case ((userId, importIdOpt, source, installationId, isPrivate, libraryId), rawKeepGroup) =>
-
+    rawKeeps.groupBy(rk => (rk.userId, rk.importId, rk.source, rk.installationId, rk.isPrivate, rk.libraryId, rk.hashtags)).map {
+      case ((userId, importIdOpt, source, installationId, isPrivate, libraryId, hashtags), rawKeepGroup) =>
```
ahsu1230: big groupBy for faster tag retrieving/creating

kifi-backend/modules/shoebox/app/com/keepit/model/RawKeep.scala
andrewconner: Yep, quotes are fine. Otherwise, you'd have to deal with escaping commas anyways.

kifi-backend/modules/shoebox/app/com/keepit/commanders/RawKeepImporterPlugin.scala
andrewconner: This isn't necessary and probably introduces bugs. Just use normal JSON, so the 3 tags "JavaScript, Scala, Programming" become `["JavaScript", "Scala", "Programming"]` in the database, so `JsArray(Seq(JsString("JavaScript"), JsString("Scala"), JsString("Programming")))` (automatic, no need to do that manually). The overhead here is tiny, especially considering we already persist large json objects (`originalJson`). Then, just parse it as `tagArray.as[Seq[String]]`

``` scala
scala> Json.parse(""" ["a", "b"] """)
res0: play.api.libs.json.JsValue = ["a","b"]

scala> res0.as[Seq[String]]
res1: Seq[String] = List(a, b)
```

kifi-backend/modules/shoebox/app/com/keepit/commanders/RawKeepImporterPlugin.scala
andrewconner: Same as before

kifi-backend/modules/shoebox/app/com/keepit/commanders/RawKeepImporterPlugin.scala
andrewconner: Not sure if this is a good idea for `hashtags: Option[JsArray]`. All the `None`s are grouped together, which is good, but any keep with any set of tags has to match exactly (including order) to another keep to get grouped. I would expect this to slow the existing code down quite a bit. Compare to `importIdOpt`, `source`, etc that are constant through an import (if you import 500 keeps, they'll all be the same).

The grouping here is for performance; I think I stopped mid-idea earlier when mentioning this. For hashtags, we'd want the same batching idea. One way is to take your hashtags across many keeps, add to a set, and get/create the tags there. Then, you can reference the objects keyed by the hashtag string.

It seems that you actually do most of this below. Just make sure to only do it once per hashtag, instead of per hashtag in a keep.

kifi-backend/modules/shoebox/app/com/keepit/model/RawKeepRepo.scala
andrewconner: Thanks!

kifi-backend/modules/shoebox/app/com/keepit/model/RawKeep.scala
```diff
@@ -21,9 +21,10 @@ case class RawKeep(
     installationId: Option[ExternalId[KifiInstallation]] = None,
     originalJson: Option[JsValue] = None,
     state: State[RawKeep] = RawKeepStates.ACTIVE,
-    tagIds: Option[String] = None,
+    tagIds: Option[String] = None, // deprecated - use hashtags instead!
     libraryId: Option[Id[Library]],
-    createdDate: Option[DateTime] = None) extends Model[RawKeep] {
+    createdDate: Option[DateTime] = None,
+    hashtags: Option[JsArray] = None) extends Model[RawKeep] {
```
2is10: Thanks for tackling this.

Personally, I’d prefer we call this field `tags`. That’s what our product calls them, and how users think of them. Even if we someday decide to make users type `#` to apply them, these are conceptually quite different from hashtags on, say, Twitter or Instagram. I’d like to rename the `Hashtag` type to `Tag` as well, once obsolete code is gone and we think we can do it without causing any confusion. Or maybe `KeepTag`. I’d love to hear other opinions on this.

I mention this here and now because classes are cheap to rename, but database columns are much harder.

ahsu1230: I'm kind of leaning on `KeepTag` because it's the most explicit on the backend (tags for keeps). Later on, if we want to these tags to do more stuff and have different functionalities, we could use `Tag` or `Hashtag`

But, I think if we name how the tags work now as `KeepTag` this name probably won't change in the future.

kifi-backend/modules/shoebox/app/com/keepit/commanders/RawKeepImporterPlugin.scala
```diff
@@ -89,8 +89,17 @@ private class RawKeepImporterActor @Inject() (
 
     rawKeeps.groupBy(rk => (rk.userId, rk.importId, rk.source, rk.installationId, rk.isPrivate, rk.libraryId)).map {
       case ((userId, importIdOpt, source, installationId, isPrivate, libraryId), rawKeepGroup) =>
```
ahsu1230, line 91: kept the grouping as before. The `getOrCreateTag` requires a context based on userId & importId, which I would expect span over lots of keeps at a time. This is where we'll create the tagset & retrieve tagIds later

kifi-backend/modules/shoebox/app/com/keepit/commanders/RawKeepImporterPlugin.scala
```diff
@@ -89,8 +89,17 @@ private class RawKeepImporterActor @Inject() (
...
-
         val context = importIdOpt.map(importId => getHeimdalContext(userId, importId)).flatten.getOrElse(HeimdalContext.empty)
+
+        val keepTagSet = rawKeepGroup.map { rk =>
```
andrewconner: This is a Map?

kifi-backend/modules/shoebox/app/com/keepit/commanders/RawKeepImporterPlugin.scala
```diff
@@ -89,8 +89,17 @@ private class RawKeepImporterActor @Inject() (
...
+          rk.keepTags.map { tagNames =>
+            tagNames.as[Seq[String]].map { tag =>
+              val collection = bookmarksCommanderProvider.get.getOrCreateTag(userId, tag)(context)
```
andrewconner: This has the same inefficiency. For each tag in each keep, looking up tags. Since tags are often reused, it is probably better to a) get the set of tags (strings), b) look them up. Grab me if you'd like to pair it.

kifi-backend/modules/shoebox/app/com/keepit/model/RawKeep.scala
```diff
@@ -57,9 +58,30 @@ class RawKeepFactory @Inject() (airbrake: AirbrakeNotifier) {
     val url = (json \ "url").asOpt[String].getOrElse(throw new Exception(s"json $json did not have a url"))
     val isPrivate = (json \ "isPrivate").asOpt[Boolean].getOrElse(true)
     val addedAt = (json \ "addedAt").asOpt[DateTime]
+    val pathOpt = (json \ "path").asOpt[Seq[String]]
+    val tagsOpt = (json \ "tags").asOpt[Seq[String]]
+
```
andrewconner, line 63: How about something like:

`(pathOpt.getOrElse(Seq.empty) ++ tagsOpt.getOrElse(Seq.empty)).groupBy(_.toLowerCase).map(_._2.head)`

kifi-backend/modules/shoebox/test/data/bookmarks_small.json
```diff
@@ -1 +1 @@
...
\ No newline at end of file
+{"bookmarks":[{"id":"4","url":"http://www.hackcollege.com/blog/2011/11/23/infographic-get-more-out-of-google.html","title":"Infographic: Get More Out Of Google - HackCollege","tags":["hack","coLLege"],"path":["college","stuff"]},{"id":"5","url":"http://www.findsounds.com/types.html","title":"FindSounds - Sound Types"},{"id":"6","url":"http://www.justbeamit.com/","title":"JustBeamIt - file transfer made easy"},{"id":"7","url":"http://en.wikipedia.org/wiki/Web_2.0_Suicide_Machine/","title":"Web 2.0 Suicide Machine/ - Wikipedia, the free encyclopedia"},{"id":"8","url":"http://en.wikipedia.org/wiki/Web_2.0_Suicide_Machine","title":"Web 2.0 Suicide Machine - Wikipedia, the free encyclopedia"}],"source":"INIT_LOAD"}
```
ahsu1230, line 1: this also checks "college" vs. "coLLege" - the test only comes with "college"

ahsu1230:
thanks for all the help & patience. @andrewconner @2is10

kifi-backend/modules/shoebox/app/com/keepit/commanders/RawKeepImporterPlugin.scala
```diff
@@ -89,8 +89,23 @@ private class RawKeepImporterActor @Inject() (
...
+          rk.keepTags.map { tagNames =>
+            tagNames.as[Seq[String]].foreach { tagName =>
+              keepTagNamesMap += (tagName.toLowerCase -> tagName)
+            }
+          }
+        }
+        // map tagNames to keepTags
+        val keepTagMap = keepTagNamesMap.values.toSet.map { tagName: String =>
+          val keepTag = bookmarksCommanderProvider.get.getOrCreateTag(userId, tagName)(context)
+          (tagName.toLowerCase, keepTag)
```
ahsu1230, line 106: basically, JavaScript and Javascript should map to the same tag

PR #14460, opened 2015-03-20
should not do IO on a controller class initialization.

eishay:
Note that this class is re-initiated each time its used (its not a singleton) and its being loaded on tests.
Its a good idea to have a class level `val` in controller


eishay:
@yingjieMiao @stkem

yingjieMiao:
Ok. It calls fake cortex client to get version number. I wrote that way so that in production when I hit the admin page, it's less likely to timeout -- I hope.

eishay:
It actually tried to do a real IO in one of the tests (ShoeboxModuleTest)

PR #14469, opened 2015-03-20
hopefully fix sporadic test build fails

ahsu1230:
@eishay @stkem


kifi-backend/modules/common/conf/evolutions/shoebox/101.sql
```diff
@@ -6,7 +6,8 @@ CREATE INDEX device_i_user_id ON device (user_id);
 
 insert into evolutions (name, description) values('101.sql', 'fixing devices table indices');
 
-# --- !Downs
-
 DROP INDEX device_u_token ON device;
 DROP INDEX device_u_user_id ON device;
+
+# --- !Downs
+
```
ahsu1230, line 10: this sql is to remove indices defined in 71.sql
But because it's in Downs, it's actually not being applied when building the dev test environment
- is my theory

PR #14472, opened 2015-03-20
Notes field on keeps

andrewconner:
@stkem


andrewconner:
Evolution applied.

stkem:
:+1:

yingjieMiao:
`longtext`? we really need to support a note size up to 4GB? 

I would prefer to keep prefix bytes as few as possible (while leave enough room for future). Saving 1 or 2 bytes per row might sound trivial, but it's equally trivial to use a column type like `text` or `medium text`.

andrewconner:
We'll enforce size limitations in code. This is not a full feature, just letting mobile play with it. There is the possibility for additional markup inside of notes (like look heres), so I modeled after Eliza's messages.

PR #14473, opened 2015-03-20
[BE] adding user bio support

ahsu1230:
so there seems to be a `UserValue.USER_DESCRIPTION` value already in place with several users already having them (last updated in June 2014).

It's in UserValue probably because of the idea that biographies would be sparse.

@andrewconner


ahsu1230:
can edit biographies from these endpoints:

```
 * POST /site/user/me/biography       {"biography" : "asdf"}
 * POST /m/1/user/me                       {"biography" : "asdf"}
```

You should be able to find the `biography` field when you call the profile endpoint:
`/site/user/:name/profile`
`/m/1/user/:username/profile`

@2is10 @jaredpetker @DerekBurch @thassss @mrbrown14

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileUserController.scala
```diff
@@ -298,17 +297,24 @@ class MobileUserController @Inject() (
...
+          (numConnections, userBio)
         }
+
+        val jsonFriendInfo = Json.toJson(profile.basicUserWithFriendStatus).as[JsObject]
+        val jsonProfileInfo = Json.toJson(UserProfileStats(
+          numLibraries = numLibraries,
+          numFollowedLibraries = numFollowedLibs,
+          numKeeps = profile.numKeeps,
+          numConnections = numConnections,
+          numFollowers = libraryCommander.countFollowers(profile.userId, viewer.map(_.id.get)),
```
ahsu1230, line 312: mobile guys, you get number of user's # of connections & followers now =D

ahsu1230:
merging to start FE work

PR #14474, opened 2015-03-20
faster logged-out page loads (one less roundtrip—for a 403)

2is10:
@andrewconner 

This is another baby step towards including data in the page. The first data I’d like to include is basic information about the logged-in user. Confusingly, we’ll need to make the app continue to load the information via XHR in the dev environment.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/KifiSiteRouter.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/KifiSiteRouter.scala
@@ -51,7 +51,7 @@ class KifiSiteRouter @Inject() (
     redirectUserToProfileToConnect(friend, request) getOrElse redirUserToOwnProfile("/connections", request)
   }
   def handleInvitePage(friend: Option[String]) = WebAppPage { implicit request =>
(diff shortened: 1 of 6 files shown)


kifi-backend/modules/shoebox/app/com/keepit/controllers/website/KifiSiteRouter.scala
```diff
@@ -106,7 +108,7 @@ class KifiSiteRouter @Inject() (
               val status = if (!isLibraryAlias || userRedirectStatusOpt.exists(_ == 303)) 303 else 301
               Redirect(uri, status)
             } else {
-              serveWebApp(request.userAgentOpt.orElse(Some(UserAgent.UnknownUserAgent)).filter(_.possiblyBot).map(_ => libMetadata(library)))
+              AngularApp.app(() => libMetadata(library))
```
2is10, line 111: @eishay, I’ve taken this `possiblyBot` logic we were using to decide whether to embed library metadata and generalized it, so now user profile pages won’t get metadata either, unless `possiblyBot`.

andrewconner:
Nice.

PR #14478, opened 2015-03-23
Upgrade to Play 2.3.8, sbt 0.13.8

andrewconner:
Better incremental compiles, cached dependency sub-graphs.

@stkem @eishay

### DIFF
--- a/kifi-backend/.sbtopts
+++ b/kifi-backend/.sbtopts
@@ -6,4 +6,3 @@
 -J-XX:+HeapDumpOnOutOfMemoryError
 -J-XX:+DoEscapeAnalysis
 -J-XX:+UseCodeCacheFlushing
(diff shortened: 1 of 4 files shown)


kifi-backend/.sbtopts
```diff
@@ -6,4 +6,3 @@
 -J-XX:+HeapDumpOnOutOfMemoryError
 -J-XX:+DoEscapeAnalysis
 -J-XX:+UseCodeCacheFlushing
--J-XX:+UseConcMarkSweepGC
```
stkem, line 9: ?

stkem:
:+1:

kifi-backend/.sbtopts
andrewconner, line 9: Java 8 doesn't have that.

PR #14479, opened 2015-03-23
ng-src to prevent spurious (failing) HTTP requests

2is10:
@ahsu1230

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/persona/managePersona.tpl.html
+++ b/kifi-backend/modules/shoebox/angular/src/persona/managePersona.tpl.html
@@ -7,8 +7,8 @@
       <div class="kf-mp-table">
         <div ng-repeat="persona in allPersonas" class="kf-mp-icon" ng-class="{active: persona.selected}" ng-click="toggleSelect(persona)">
           <div class="kf-mp-icon-image-wrapper">
-            <img class="kf-mp-icon-image" src="{{persona.iconSrc}}" ng-if="!persona.selected">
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/angular/src/persona/managePersona.tpl.html
```diff
@@ -7,8 +7,8 @@ <h1 class="dialog-header"> Who are you? </h1>
       <div class="kf-mp-table">
         <div ng-repeat="persona in allPersonas" class="kf-mp-icon" ng-class="{active: persona.selected}" ng-click="toggleSelect(persona)">
           <div class="kf-mp-icon-image-wrapper">
-            <img class="kf-mp-icon-image" src="{{persona.iconSrc}}" ng-if="!persona.selected">
-            <img class="kf-mp-icon-image" src="{{persona.activeIconSrc}}" ng-if="persona.selected">
+            <img class="kf-mp-icon-image" ng-src="{{persona.iconSrc}}" ng-if="!persona.selected">
+            <img class="kf-mp-icon-image" ng-src="{{persona.activeIconSrc}}" ng-if="persona.selected">
```
2is10, line 11: Angular instantiates the DOM before processing the templates, so it can’t stop the browser from making requests for images if they have a `src` attribute. So use `ng-src` if there’s an embedded expression that needs evaluating.

ahsu1230, line 11: ohhh I see. Thanks! good to know.

PR #14483, opened 2015-03-23
edit user biography web FE (on profile & settings page)

ahsu1230:
@2is10 @andrewconner


ahsu1230:
Look at this one: https://github.com/kifi/keepit/pull/14484 (so Jo can test)

PR #14484, opened 2015-03-23
edit user biography web FE (on profile & settings page)

ahsu1230:
@andrewconner @2is10


kifi-backend/modules/shoebox/angular/src/userProfile/userProfileHeader.styl
```diff
@@ -193,6 +194,81 @@
...
+.kf-uph-bio
+  width: 400px
+  margin: 0 auto 10px auto
+  color: white
+  font-size: 0.9em
+  line-height: 1.6em
+  font-weight: 100
+  word-break: break-word
+
+.kf-bio-input
```
2is10, line 227: Let’s put these `kf-bio` styles in their own file (editUserBio.styl, to go with editUserBio.tpl.html and editUserBio.js).

PR #14487, opened 2015-03-24
Rover Ingestion from Shoebox

leogrim:
@stkem @eishay @andrewconner @yingjieMiao 
This is running late, met more hiccups :(

### DIFF
--- a/kifi-backend/modules/rover/app/com/keepit/rover/model/ArticleInfo.scala
+++ b/kifi-backend/modules/rover/app/com/keepit/rover/model/ArticleInfo.scala
@@ -19,6 +19,7 @@ case class ArticleInfo(
     kind: String, // todo(Léo): make this kind: ArticleKind[_ <: Article] with Scala 2.11, (with proper mapper, serialization is unchanged)
     bestVersion: Option[ArticleVersion] = None,
     latestVersion: Option[ArticleVersion] = None,
+    oldestVersion: Option[ArticleVersion] = None,
     lastQueuedAt: Option[DateTime] = None,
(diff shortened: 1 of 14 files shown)


kifi-backend/modules/rover/app/com/keepit/rover/model/ArticleInfo.scala
```diff
@@ -19,13 +19,24 @@ case class ArticleInfo(
     kind: String, // todo(Léo): make this kind: ArticleKind[_ <: Article] with Scala 2.11, (with proper mapper, serialization is unchanged)
     bestVersion: Option[ArticleVersion] = None,
     latestVersion: Option[ArticleVersion] = None,
+    oldestVersion: Option[ArticleVersion] = None,
     lastQueuedAt: Option[DateTime] = None,
     lastFetchedAt: Option[DateTime] = None,
     nextFetchAt: Option[DateTime] = None,
     fetchInterval: Float = 24.0f) extends ModelWithState[ArticleInfo] with ModelWithSeqNumber[ArticleInfo] with ArticleKeyHolder {
```
yingjieMiao, line 26: this default interval is perhaps unnecessarily small (current system has a config at about 2 weeks)

leogrim, line 26: Yes you're probably right, I'll make it larger.

yingjieMiao, line 26: thank you. i know it's old code.

PR #14489, opened 2015-03-24
Scala 2.11

andrewconner:
Release information:
http://scala-lang.org/news/2.11.0
http://docs.scala-lang.org/scala/2.11/

Tests pass, a few extra warnings from newly depricated classes that we'll knock out.

@eishay @stkem @ahsu1230 @2is10 @yingjieMiao @leogrim


stkem:
:-)

PR #14492, opened 2015-03-24
using @media queries instead of kf-mobile class for more responsive experience

2is10:
cc @andrewconner @eishay @ahsu1230 @stkem


2is10:
See Google’s [Multi-device layouts](https://developers.google.com/web/fundamentals/layouts/) page for more background.

PR #14500, opened 2015-03-24
fix marking notifications as read

ahsu1230:
@stkem 

thanks for the find!


ahsu1230:
let's hope it workS!

2is10:
gud fix

PR #14510, opened 2015-03-25
edit profile dropdown menu

ahsu1230:
@2is10


kifi-backend/modules/shoebox/angular/src/profile/editUserBioModal.styl
```diff
@@ -0,0 +1,44 @@
...
+  font-size: 1.1em
+  line-height: 1.6em
+
+.kf-bio-chars
+  margin-top: 10px
+  color: accentGreen
+  float: right
+
+  &.over
+    color: #d50000
```
2is10, line 18: Let’s use a `kf-` prefix even on “state” classes. Otherwise lame extensions may interfere with us.

kifi-backend/modules/shoebox/angular/src/userProfile/userProfileHeader.styl
```diff
@@ -11,6 +11,9 @@ uphSmallMaxDim = 736px // for @media max-height and max-width
 .kf-uph-top
   position: relative
   padding-top: 20px
+  width: 100%
```
2is10: `width: 100%` seems unnecessary here since that’s what a `div` gets automatically. am I missing something?

kifi-backend/modules/shoebox/angular/src/userProfile/userProfileHeader.styl
```diff
@@ -11,6 +11,9 @@ uphSmallMaxDim = 736px // for @media max-height and max-width
...
+  max-width: 906px
```
2is10: let’s use the `appMaxWidth` constant here

kifi-backend/modules/shoebox/angular/src/userProfile/userProfileHeader.tpl.html
```diff
@@ -35,10 +35,17 @@
...
+        <span class="kf-uph-dropdown-parent" kf-hover-menu ng-switch-when="self">
+          <a ui-sref="settings" ng-click="trackClickedSettings()">
+            <span class="kf-uph-settings-icon svg-settings-gear"></span>
+            <span class="kf-uph-action-text">Settings</span>
+          </a>
+          <menu class="kf-dropdown-menu">
+            <button class="kf-dropdown-menu-item" ng-click="showBiographyModal()">Edit your biography</button>
+            <a class="kf-dropdown-menu-item" ui-sref="settings" ng-click="trackClickedSettings()">More settings</a>
+          </menu>
+        </span>
```
2is10: It looks like `<span>` isn’t technically allowed to contain `<menu>` elements. So some browsers may auto-close a `<span>` when the `<menu>` open tag is encountered. Better to use a `<div>` for the wrapper.
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/span (see “Permitted Content”)
https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#Phrasing_content

2is10:
Nice work, thanks! :+1:

kifi-backend/modules/shoebox/angular/src/userProfile/userProfileHeader.styl
ahsu1230: nope you're right, not sure why I felt compelled to explicitly state this

kifi-backend/modules/shoebox/angular/src/userProfile/userProfileHeader.styl
```diff
@@ -219,6 +259,7 @@ uphSmallMaxDim = 736px // for @media max-height and max-width
   font-size: 0.9em
   font-weight: 100
   font-smoothing: antialiased
+  -moz-osx-font-smoothing: grayscale;
```
2is10, line 262: Nice! We should make `font-smoothing: antialiased` add that automatically.

PR #14525, opened 2015-03-25
reco feedback from mobile support

stkem:
@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/conf/shoeboxService.routes
+++ b/kifi-backend/modules/shoebox/conf/shoeboxService.routes
@@ -251,6 +251,7 @@ GET     /m/2/recos/top                          @com.keepit.controllers.mobile.M
 GET     /m/3/recos/top                          @com.keepit.controllers.mobile.MobileRecommendationsController.topRecosV3(recency: Float, uriContext: Option[String] ?= None, libContext: Option[String] ?= None, uctx: Option[String] ?= None, lctx: Option[String] ?= None)
 GET     /m/1/recos/public                       @com.keepit.controllers.mobile.MobileRecommendationsController.topPublicRecos()
 POST    /m/1/recos/trash                        @com.keepit.controllers.mobile.MobileRecommendationsController.trash(id: ExternalId[NormalizedURI])
+POST    /m/1/recos/feedback                     @com.keepit.controllers.website.RecommendationsController.updateUriRecommendationFeedback(id: ExternalId[NormalizedURI])
 
 GET     /m/1/users/:userStr/libraries/:slug     @com.keepit.controllers.mobile.MobileLibraryController.getLibraryByPath(userStr:String, slug: String, is: Option[String] ?= None)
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/shoebox/conf/shoeboxService.routes
```diff
@@ -251,6 +251,7 @@ GET     /m/2/recos/top                          @com.keepit.controllers.mobile.M
 GET     /m/3/recos/top                          @com.keepit.controllers.mobile.MobileRecommendationsController.topRecosV3(recency: Float, uriContext: Option[String] ?= None, libContext: Option[String] ?= None, uctx: Option[String] ?= None, lctx: Option[String] ?= None)
 GET     /m/1/recos/public                       @com.keepit.controllers.mobile.MobileRecommendationsController.topPublicRecos()
 POST    /m/1/recos/trash                        @com.keepit.controllers.mobile.MobileRecommendationsController.trash(id: ExternalId[NormalizedURI])
+POST    /m/1/recos/feedback                     @com.keepit.controllers.website.RecommendationsController.updateUriRecommendationFeedback(id: ExternalId[NormalizedURI])
```
stkem, line 254: @jaredpetker @mrbrown14 
For keep recos: one query parameter `id` with the id of the reco and a post body with a json object with the following fields:

```
    clicked: Boolean,
    kept: Boolean,
    trashed: Boolean //,
    vote: Boolean //true -> more like this, false -> no more like this,
    comment: String //for additional feedback (no mobile UI afaik)
```

all fields are optional. In most calls only one will be set.

kifi-backend/modules/shoebox/conf/shoeboxService.routes
```diff
@@ -269,6 +270,8 @@ POST    /m/1/libraries/:id/invite               @com.keepit.controllers.mobile.M
 GET     /m/1/libraries/:id/members              @com.keepit.controllers.mobile.MobileLibraryController.getLibraryMembers(id: PublicId[Library], offset: Int ?= 0, limit: Int ?= 20)
 GET     /m/1/libraries/:id/members/suggest      @com.keepit.controllers.mobile.MobileLibraryController.suggestMembers(id: PublicId[Library], q: Option[String] ?= None, n: Int ?= 5)
 GET     /m/1/libraries/:id/inviteInfo           @com.keepit.controllers.website.InviteController.getLibraryInviteInfo(id: PublicId[Library])
+POST    /m/1/libraries/recos/feedback      @com.keepit.controllers.website.LibraryRecommendationsController.updateLibraryRecommendationFeedback(id: PublicId[Library])
```
stkem, line 273: @jaredpetker @mrbrown14  
For library recos: One query parameter `id` with the publicId of the library and a post body with a json object with the following fields:

```
clicked: Boolean,
    followed: Boolean,
    trashed: Boolean,
    vote: Boolean,
    comment: String
```

all fields are optional. In most calls only one will be set.

PR #14528, opened 2015-03-25
send email for twitter waitlist

ahsu1230:
@andrewconner


kifi-backend/modules/shoebox/test/com/keepit/commanders/TwitterWaitlistCommanderTest.scala
```diff
@@ -45,18 +46,21 @@ class TwitterWaitlistCommanderTest extends TestKitSupport with ShoeboxTestInject
...
           userRepo.count === 1
           twitterWaitlistRepo.count === 0
+          emailRepo.count === 0
         }
 
         // add a new entry
         twitterWaitlistCommander.addEntry(user1.id.get, "therealcaptainfalcon").isRight === true
         db.readOnlyMaster { implicit s =>
           twitterWaitlistRepo.getByUserAndHandle(user1.id.get, "therealcaptainfalcon").nonEmpty === true
+          emailRepo.count === 1
```
ahsu1230, line 62: woops, need to do an `Await`...

PR #14529, opened 2015-03-25
some statsd

yingjieMiao:
@stkem


stkem:
:+1: :+1:

yingjieMiao:
i don't think data are coming in. In fact, non of the statsd data.

I guess we didn't migrate statsd from marvin to arthur (or elsewhere)

stkem:
marvin is still alive, so that should not be affected. _All_ statsd is broken?

yingjieMiao:
i think so. just look at existing admin pages.

PR #14534, opened 2015-03-25
twitter library header + keep card

ahsu1230:
@andrewconner @stkem @2is10


kifi-backend/modules/shoebox/angular/img/svg/twitter-logo.svg
```diff
@@ -0,0 +1,14 @@
...
+            </g>
+        </g>
+    </g>
```
2is10: maybe call this twitter-blue.svg to differentiate it from the other three twitter logos in this directory? :smile:

kifi-backend/modules/shoebox/angular/src/keep/keepCard.tpl.html
```diff
@@ -12,6 +12,11 @@
       </div>
       <div class="kf-keep-subtitle">
         <span class="kf-keep-url" ng-attr-title="{{keep.url}}" ng-bind-html="keep.descHtml"></span>
+        <span class="kf-keep-twitter-attr" ng-if="keep.sourceAttribution.twitter.screenName">
+          <a href="https://www.twitter.com/{{keep.sourceAttribution.twitter.screenName}}/status/{{keep.sourceAttribution.twitter.idString}}" target="_blank">
+            <div class="kf-keep-twitter-logo svg-twitter-logo"></div>@{{keep.sourceAttribution.twitter.screenName}}
```
2is10: `<div>` cannot go inside `<span>`, so use `<span>`

kifi-backend/modules/shoebox/angular/src/libraries/libraryHeader.styl
```diff
@@ -444,6 +444,43 @@ invite-banner-orange = #ff7c3c
 .kf-lh-owner-name
   color: #939393
 
+.kf-lh-twitter-name
+  display: inline-block
+  margin-left: 15px
+
+  &:hover .kf-lh-twitter-img
+    background-color: #e0e0e0
```
2is10: I prefer putting overrides as close as possible to what they’re overriding. This rule could be moved into the `.kf-lh-twitter-img` section and written this way:

``` stylus
  :hover>&
    background-color: #e0e0e0
```

kifi-backend/modules/shoebox/angular/src/libraries/libraryHeader.tpl.html
```diff
@@ -45,6 +45,17 @@ <h1 class="kf-lh-name" itemtype="name">{{library.name}}</h1>
               <div class="kf-lh-owner-label">Curator:</div>
               <div class="kf-lh-owner-name" itemprop="name">{{library.owner.firstName}} {{library.owner.lastName}}</div>
             </div>
+
+            <a class="kf-lh-twitter-name" href="https://twitter.com/{{library.attr.twitter.screenName}}" target="_blank" ng-if="library.attr.twitter.screenName">
+              <div class="kf-lh-twitter-img">
+                <div class="kf-lh-twitter-logo svg-twitter-logo"></div>
+              </div>
+              <div class="kf-lh-owner-text">
+                <div class="kf-lh-owner-name">
```
2is10: Let’s please not re-use the `-owner-` classes here. If they have rules with property values that should stay in sync, you can define those values as stylus constants.

2is10:
nicely done!

PR #14536, opened 2015-03-25
reco processing priority to certain users

stkem:
@yingjieMiao


yingjieMiao:
this only separates the locks. Do we want to include him for every update cycle?

stkem:
All users are in every update cycle, just the lock prevents more then 8 (now 6) from running.

PR #14537, opened 2015-03-26
measuring space for follower pics accurately

2is10:


kifi-backend/modules/shoebox/angular/src/recos/recoLibraryCard.styl
```diff
@@ -133,10 +133,13 @@ rcl-side-pad = 30px
...
 
-  .kf-lfp
+  .kf-rcl-followers
     display: table-cell
     width: 99%
-    vertical-align: middle
+    padding: 15px 0
+
+    &>.kf-lfp
+      position: absolute
```
2is10, line 142: using `position: absolute` here to get the follower pics out of the flow, so that the container’s width can be measured uninfluenced by the number of follower pics currently showing

PR #14545, opened 2015-03-26
only consider keeper keeps

yingjieMiao:
@stkem


kifi-backend/modules/cortex/app/com/keepit/cortex/dbmodel/URILDATopicRepo.scala
```diff
@@ -176,18 +176,19 @@ class URILDATopicRepoImpl @Inject() (
...
   }
 
-  def getUserURIFeatures(userId: Id[User], version: ModelVersion[DenseLDA], min_num_words: Int)(implicit session: RSession): Seq[LDATopicFeature] = {
+  def getUserURIFeatures(userId: Id[User], version: ModelVersion[DenseLDA], min_num_words: Int, keeperOnly: Boolean = false)(implicit session: RSession): Seq[LDATopicFeature] = {
     import com.keepit.common.db.slick.StaticQueryFixed.interpolation
     import scala.slick.jdbc.GetResult
     implicit val getLDATopicFeature = getResultFromMapper[LDATopicFeature]
-
-    val q = sql"""select tp.feature from cortex_keep as ck inner join uri_lda_topic as tp on ck.uri_id = tp.uri_id where ck.user_id = ${userId.id} and ck.state = 'active' and tp.version = ${version.version} and ck.source != 'default' and tp.state = 'active' and tp.num_words > ${min_num_words}"""
+    val sourceConstrain = if (keeperOnly) """and ck.source = 'keeper' """ else ""
```
stkem: This is not quite right. It should explicitly exclude `bookmarkImport`, not explicitly include `keeper`. Otherwise it will also excluce things like mobile and site keeps, we should be included.

yingjieMiao:
ok, I will just use source in (keeper, site, mobile)

there are tons of other imports, e.g. pocket, instapaper

stkem:
There are many more that are generally good. Twitter, email, emailReco, userCopied, ..... I would argue that Instapaper and pocket are also good since they are curated. The only one that is problematic is bookmarkImport because we prompt user during signup to do that.

yingjieMiao:
I would love to confirm this with product. Technical detail is trivial here.

yingjieMiao:
@stkem  updated query constrain:

`ck.source in ('keeper', 'mobile', 'site', 'twitterSync')`

This is confirmed with Mark.

PR #14549, opened 2015-03-26
fix grouping over twitter state syncing

ahsu1230:
@andrewconner


kifi-backend/modules/shoebox/app/com/keepit/common/social/TwitterSocialGraph.scala
```diff
@@ -174,20 +174,21 @@ class TwitterSocialGraphImpl @Inject() (
...
-          syncStateMap.map { case (userId, twitterSyncState) =>
-            db.readWrite { implicit s =>
-              val libraryId = twitterSyncState.libraryId
-              libraryMembershipRepo.getWithLibraryIdAndUserId(libraryId, userId, None) match {
-                case None =>
-                  log.info(s"[fetchSocialUserInfo(${socialUserInfo.socialId})] auto-joining user ${userId} twitter_sync library ${libraryId}")
-                  libraryMembershipRepo.save(LibraryMembership(libraryId = libraryId, userId = userId, access = LibraryAccess.READ_ONLY))
-                case _ =>
+        log.info(s"[fetchSocialUserInfo(${socialUserInfo.socialId})] fetching twitter_syncs for ${friendIds.length} friends...")
+        friendIds.map(Id[User](_)).grouped(100) foreach { userIds =>
```
ahsu1230, line 178: woops my b. This should be better. Group the whole set of `friendIds` by 100 and then search twitter syncs (by 100)

PR #14551, opened 2015-03-26
much longer caching of social scores

stkem:
cache is useless right now because it takes longer to run through all users once than the cache TTL.
@yingjieMiao


yingjieMiao:
good catch

PR #14562, opened 2015-03-27
bigger curator batch size

stkem:
@yingjieMiao


yingjieMiao:
what does this do? user in the reactive lock queue will wait longer?

stkem:
No, more URIs per time a user is processed (i.e. we will now process 350 uris for a user until they go back to the end of the line)

yingjieMiao:
That's what I meant: each user now costs more time, so increasing wait time for other guys in the queue.

stkem:
Yeah, but that is minimal. A batch takes less than 1 second on average. I'm hoping larger batches will reduce overhead.

PR #14564, opened 2015-03-27
Bio Modal tweeks

ahsu1230:
@2is10 

ignore commit history names - they aren't very correct

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/profile/editUserBio.tpl.html
+++ b/kifi-backend/modules/shoebox/angular/src/profile/editUserBio.tpl.html
@@ -5,10 +5,16 @@
 
     <div class="dialog-body dialog-centered">
       <textarea class="kf-bio-input" type="text" placeholder="Edit your biography here... Limit to 160 characters" ng-model="bioText" ng-change="checkOverLimit()">{{bioText}}</textarea>
-      <div class="kf-bio-chars" ng-class="{'kf-over': charactersLeft < 0}">
+
+      <div class="kf-bio-chars" ng-if="charactersLeft >= 0">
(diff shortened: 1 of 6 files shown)


kifi-backend/modules/shoebox/angular/src/profile/editUserBio.tpl.html
```diff
@@ -5,10 +5,16 @@ <h1 class="dialog-header">Edit your profile biography</h1>
...
       <textarea class="kf-bio-input" type="text" placeholder="Edit your biography here... Limit to 160 characters" ng-model="bioText" ng-change="checkOverLimit()">{{bioText}}</textarea>
-      <div class="kf-bio-chars" ng-class="{'kf-over': charactersLeft < 0}">
+
+      <div class="kf-bio-chars" ng-if="charactersLeft >= 0">
         <ng-pluralize count="charactersLeft" when="{'0': 'No characters left', '1': '1 character left', 'other': '{} characters left'}">
         </ng-pluralize>
       </div>
+      <div class="kf-bio-chars kf-over" ng-if="charactersLeft < 0">
+        <ng-pluralize count="absoluteVal(charactersLeft)" when="{'1': '1 character over', 'other': '{} characters over'}">
+        </ng-pluralize>
```
2is10, line 15: The `absoluteVal` function isn’t necessary. We know the sign and can use unary `-` to negate it:
`count="-charactersLeft"`

2is10, line 15: Not critical, but FYI, we can also use `ng-pluralize` as an attribute (e.g. on the parent `<div>`) for more concise markup and to avoid creating an extra DOM node.

2is10:
:+1:

PR #14565, opened 2015-03-27
more edit Bio weeks

ahsu1230:
@2is10 thanks for the review on last PR


kifi-backend/modules/shoebox/angular/src/libraries/libraryHeader.styl
```diff
@@ -458,7 +458,7 @@ invite-banner-orange = #ff7c3c
     margin-left: 8px
 
     @media (max-width: 480px)
-      margin-top: -10px
+      margin-top: -5px
```
ahsu1230, line 461: for some reason shows up weird on iPhone6+, but on web, even in small windows, it looks fine =/

2is10:
:+1:

PR #14569, opened 2015-03-27
tidying up a bit

2is10:


kifi-backend/modules/shoebox/app/views/authMinimal/install.scala.html
```diff
@@ -81,7 +81,6 @@ <h1 class="install-title">Get the browser extension<br>to start using Kifi</h1>
   </div>
 }
 @scripts = {
-  <script type="text/javascript" src="//www.googleadservices.com/pagead/conversion.js"></script>
```
2is10, line 84: This was in the page twice—see old (red) line 127.

kifi-backend/modules/shoebox/app/views/authMinimal/install.scala.html
```diff
@@ -121,23 +123,10 @@ <h1 class="install-title">Get the browser extension<br>to start using Kifi</h1>
...
-  <script type="text/javascript" src="//www.googleadservices.com/pagead/conversion.js">
-  </script>
-
-  <script language='JavaScript'>
-    var OB_ADV_ID=25665;
-    var scheme = (("https:" == document.location.protocol) ? "https://" : "http://");
-    var str = '<script src="'+scheme+'widgets.outbrain.com/obtp.js" type="text/javascript"><\/script>';
-    document.write(str);
   </script>
-
```
2is10, line 136: I removed the blank lines between the various components of each tracker so that we’d be less likely to unintentionally separate them by inserting new code in the middle again in the future.

kifi-backend/modules/shoebox/app/views/authMinimal/install.scala.html
```diff
@@ -121,23 +123,10 @@ <h1 class="install-title">Get the browser extension<br>to start using Kifi</h1>
...
-  </script>
-
-  <script type="text/javascript" src="//www.googleadservices.com/pagead/conversion.js">
-  </script>
-
-  <script language='JavaScript'>
-    var OB_ADV_ID=25665;
-    var scheme = (("https:" == document.location.protocol) ? "https://" : "http://");
-    var str = '<script src="'+scheme+'widgets.outbrain.com/obtp.js" type="text/javascript"><\/script>';
-    document.write(str);
```
2is10, line 134: avoiding `document.write`, which slows down the page, and clobbering/leaking global vars `scheme` and `str`

PR #14571, opened 2015-03-27
content flagging for mobile endpoint

ahsu1230:
@andrewconner @stkem @yingjieMiao 

```
POST /m/1/content/flagUrl 
{
  "reason" -> "adult",
  "url" -> "http://www.nsfw.com"
}
```

You will get a `BadRequest` if the `normalizedURI` is not found, the `reason` is unknown, or if `reason` & `url` aren't in the json body.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileUriController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileUriController.scala
@@ -27,7 +27,7 @@ class MobileUriController @Inject() (
             normalizedUriRepo.updateURIRestriction(uri.id.get, Some(Restriction.ADULT))
             NoContent
           case _ =>
-            BadRequest(Json.obj("error" -> "reason_not_found"))
+            BadRequest(Json.obj("error" -> "reason_unknown"))
         }
       }
     }.getOrElse(BadRequest(Json.obj("error" -> "uri_not_found")))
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileUriController.scala
```diff
@@ -0,0 +1,35 @@
...
+  def flagContent() = UserAction(parse.tolerantJson) { request =>
+    val jsonBody = request.body
+    val reason = (jsonBody \ "reason").as[String]
+    val url = (jsonBody \ "url").as[String]
+
+    db.readWrite { implicit s =>
+      normalizedUriInterner.getByUri(url).map { uri =>
+        reason match {
+          case "adult" =>
+            normalizedUriRepo.updateURIRestriction(uri.id.get, Some(Restriction.ADULT))
```
andrewconner: Can you add emailing in here as well? Email me, Eishay, and Mark with the URL and uriId. Thanks!

andrewconner:
Awesome.

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileUriController.scala
ahsu1230: right missed that, thanks!

stkem: Wait, this should _not_ actually update the URI restriction, but only email. Otherwise any user can hide any content from other peoples search results.
It should only notify us. Please include me in the email recipients as well.

ahsu1230: ohhhh crap. Ok, my bad!

andrewconner: Ok talked with Stephen. Let's not automark it, but make the email include a link to the place where we can easily mark it: https://admin.kifi.com/admin/scraped/:uriId

PR #14573, opened 2015-03-27
related keep info

yingjieMiao:
@eishay @stkem @2is10 

basic data pipe and experiment support. 

There could be a few backend algorithm tweaks, but not in this version.


kifi-backend/modules/shoebox/app/com/keepit/commanders/PageCommander.scala
```diff
@@ -114,9 +117,24 @@ class PageCommander @Inject() (
       }
       (nUri, nUriStr)
     }
+
+    val relatedUriF = uriInfoF.flatMap {
+      case (uri, _) =>
+        if (uri.isDefined && experiments.contains(ExperimentType.RELATED_PAGE_INFO)) {
+          relatedPageCommander.getRelatedPageInfo(uri.get.id.get).recover{
+            case _ =>
+              log.error(s"error in getting related page info for uri: ${uri.get.id.get}")
```
eishay: not airbrake?

kifi-backend/modules/shoebox/app/com/keepit/commanders/PageCommander.scala
```diff
@@ -114,9 +117,24 @@ class PageCommander @Inject() (
...
+              Seq[RelatedPageInfo]()
```
eishay: `Seq.empty` ?

kifi-backend/modules/shoebox/app/com/keepit/commanders/PageCommander.scala
yingjieMiao: Could be. I didn't want to create noise, but since this is under experiment protection, I can add airbrake.

kifi-backend/modules/shoebox/app/com/keepit/commanders/PageCommander.scala
yingjieMiao: yeah

eishay:
not sure i saw the image part in this PR. 
we should make sure only urls with images are passed and the image is small (need to check with design)

yingjieMiao:
image check is there:
`val uriWithImages = (uris zip summaries).filter(_._2.imageUrl.isDefined)`

kifi-backend/modules/shoebox/app/com/keepit/commanders/PageCommander.scala
```diff
@@ -234,3 +252,47 @@ case class KeeperPagePartialInfo(
...
+      Future.sequence(uris.map { x => uriSummaryCmdr.getDefaultURISummary(x, waiting = false) })
+    }
+
+    for {
+      uris <- urisF
+      summaries <- summariesF
+    } yield {
+      assume(uris.size == summaries.size)
+      val uriWithImages = (uris zip summaries).filter(_._2.imageUrl.isDefined)
+      if (uriWithImages.size >= MIN_POOL_SIZE) {
```
stkem, line 293: Why not show anything if the pool is small?

stkem, line 293: Wouldn't some be better than nothing?

yingjieMiao, line 293: it's related to the 'small bucket problem'.  I actually need to sample a subset to return. Returning 20 pages are not helpful. Will push a fix.

PR #14575, opened 2015-03-27
better grouping for library checker

ahsu1230:
@andrewconner @eishay


kifi-backend/modules/shoebox/app/com/keepit/integrity/LibraryChecker.scala
```diff
@@ -66,8 +66,11 @@ class LibraryChecker @Inject() (
             val numKeeps = keepCountMap(libId)
             if (numKeeps != 0) {
               airbrake.notify(s"Library ${libId} has no last_kept but has $numKeeps active keeps... making them inactive!")
-              db.readWrite { implicit s =>
-                keepRepo.getByLibrary(libId, 0, numKeeps, Set.empty).map { k =>
```
ahsu1230, line 70: could potentially be lotssss of keeps

PR #14576, opened 2015-03-27
a bunch more stats tracking for curator

stkem:
@yingjieMiao

### DIFF
--- a/kifi-backend/modules/curator/app/com/keepit/curator/commanders/AttributionHelper.scala
+++ b/kifi-backend/modules/curator/app/com/keepit/curator/commanders/AttributionHelper.scala
@@ -41,9 +41,11 @@ class SeedAttributionHelper @Inject() (
   }
 
   private def getLibraryAttribution(seeds: Seq[ScoredSeedItem]): Future[Seq[Option[LibraryAttribution]]] = {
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/curator/app/com/keepit/curator/commanders/AttributionHelper.scala
```diff
@@ -91,13 +96,16 @@ class SeedAttributionHelper @Inject() (
...
         cortex.explainFeed(userId, uriIds).map { res =>
+          timer.stopAndReport()
           res.map { keepIds => if (!keepIds.isEmpty) Some(KeepAttribution(keepIds)) else None }
         }
     }
   }
 
   private def getTopicAttribution(seeds: Seq[ScoredSeedItem]): Future[Seq[Option[TopicAttribution]]] = {
+    val timer = new NamedStatsdTimer("SeedAttributionHelper.getTopicAttribution")
     cortex.getTopicNames(seeds.map { _.uriId }, seeds.headOption.map { _.userId }).map { names =>
```
yingjieMiao, line 107: the topic attribution is effectively disabled. (It's taking computation time, but a final guard will throw that away).
I don't think it's needed anymore (people don't miss the feature)

PR #14577, opened 2015-03-28
whoops, only wanted to prevent logging, not everything

stkem:
@yingjieMiao


yingjieMiao:
is scaling used anywhere? why?

stkem:
Yes, for the per user per uri timing. I only have the total time of the batch, so need to scale by batch size.

PR #14592, opened 2015-03-30
save a few bytes

yingjieMiao:
@eishay

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeeperInfo.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeeperInfo.scala
@@ -43,7 +43,7 @@ case class KeeperPageInfo(
   keepersTotal: Int,
   libraries: Seq[JsObject],
   keeps: Seq[KeepData],
-  related: Seq[RelatedPageInfo])
+  related: Option[Seq[RelatedPageInfo]]) // Option[Seq[_]] to save a few bytes
 object KeeperPageInfo {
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/KeeperInfo.scala
```diff
@@ -54,7 +54,7 @@ object KeeperPageInfo {
     (__ \ 'keepersTotal).writeNullable[Int].contramap[Int](Some(_).filter(_ > 0)) and
     (__ \ 'libraries).writeNullable[Seq[JsObject]].contramap[Seq[JsObject]](Some(_).filter(_.nonEmpty)) and
     (__ \ 'keeps).write[Seq[KeepData]] and
-    (__ \ 'related).write[Seq[RelatedPageInfo]]
+    (__ \ 'related).write[Option[Seq[RelatedPageInfo]]]
```
2is10, line 57: Does this actually work? The fields above use `.writeNullable` and `.contramap` to omit them entirely when empty.

yingjieMiao:
thanks, you are right. will change that.

2is10:
Okay, thanks for letting me know. I wasn’t sure. I thought maybe a recent Play release had made it easier to omit keys.

yingjieMiao:
```
case class Foo(x: Int, y: Option[Seq[Int]])
implicit val fw = Json.writes[Foo]
```

This does omit keys. Apparently it needs more care when manually written.

PR #14593, opened 2015-03-30
make sure returned map contains every key

yingjieMiao:
@andrewconner @ahsu1230


yingjieMiao:
this should stop exceptions in library checker.

ahsu1230:
thanks @yingjieMiao. I wonder why libraryIds aren't found...

andrewconner:
Mind investigating Aaron? Could be a bug somewhere else.

PR #14594, opened 2015-03-30
fixing library checker if no last_kept

ahsu1230:
bad logic... 9 libraries were affected, so I'll change them to the right state
@andrewconner @leogrim @eishay @yingjieMiao


yingjieMiao:
OK. It's perhaps better to change the state through some admin endpoint (So that seqNum are correctly processed)

andrewconner:
Yeah, you will probably need code to fix the wrong ones. Keep is heavily cached.

ahsu1230:
yea, I'll merge this now to change the library checker logic and use an admin endpoint later

PR #14600, opened 2015-03-30
Disabling Topic and Library-You-Follow attribution

stkem:
Topic attribution is already not shown to the user and
library-you-follow attribution is extremely rarely shown, not
justifying the computation cost at this time.

@yingjieMiao


yingjieMiao:
:+1:

PR #14601, opened 2015-03-30
allow admin library point to show inactive keeps

ahsu1230:
allow admin to show inactivated keeps. From there you can bulk update keeps to be active/inactive with the update button.

Some of the keeps in libraries were inactivated before the `LibraryChecker` did its thing, so you can differentiate using a keeps `updated_at` field.
It's a bit manual, but thankfully it's actually not many keeps at all. So I don't mind doing it.

Sorry for the bug! T_T @andrewconner @yingjieMiao


kifi-backend/modules/shoebox/app/com/keepit/model/KeepRepo.scala
```diff
@@ -439,14 +439,14 @@ class KeepRepoImpl @Inject() (
...
   }
 
-  def getCountByLibrary(libraryId: Id[Library])(implicit session: RSession): Int = {
-    getCountsByLibrary(Set(libraryId)).getOrElse(libraryId, 0)
+  def getCountByLibrary(libraryId: Id[Library], excludeSet: Set[State[Keep]] = Set(KeepStates.INACTIVE, KeepStates.DUPLICATE))(implicit session: RSession): Int = {
+    getCountsByLibrary(Set(libraryId), excludeSet).getOrElse(libraryId, 0)
   }
 
-  def getCountsByLibrary(libraryIds: Set[Id[Library]])(implicit session: RSession): Map[Id[Library], Int] = {
+  def getCountsByLibrary(libraryIds: Set[Id[Library]], excludeSet: Set[State[Keep]] = Set(KeepStates.INACTIVE, KeepStates.DUPLICATE))(implicit session: RSession): Map[Id[Library], Int] = {
```
andrewconner: I wouldn't worry about this. As-is, this looks like it may cause incorrect totals (totals including everything) to be added to the cache.

ahsu1230: Ok, is it better to just leave the admin endpoint showing only 'active' keep counts then?

andrewconner: Yeah, especially since it's cached, and we don't want to put in bad values into the cache.

PR #14603, opened 2015-03-30
Just a hunch that some emails are erroring instead of sending.

andrewconner:
@stkem User reported that he's not getting password reset emails. It works for me, but I noticed there's no error handling in this — so errors just evaporate away. Seeing if this does anything.


stkem:
:+1:

PR #14604, opened 2015-03-30
make non_sensitive field optional

yingjieMiao:
@stkem 

Some transition logic.

next step is to modify current 'false' values into 'null' (e.g. via SQL query). Then bump up cache version and deploy.

Then, we can start building the blacklist.


stkem:
lgtm

PR #14628, opened 2015-03-31
Modify store article key

leogrim:


kifi-backend/modules/rover/app/com/keepit/rover/store/RoverUnderlyingArticleStore.scala
```diff
@@ -8,7 +8,7 @@ import com.keepit.rover.model.ArticleKey
 import play.api.libs.json.Format
 
 private[store] class ArticleStoreKey(key: ArticleKey[_]) {
-  override def toString = s"${key.uriId.id}/${key.kind.typeCode}/${key.version.major.value}/${key.version.minor.value}"
+  override def toString = s"${key.uriId.id}/${key.kind.typeCode}/${key.version.major.value}.${key.version.minor.value}"
 }
```
leogrim, line 12: @stkem @andrewconner

stkem, line 12: lgtm

PR #14630, opened 2015-03-31
More lax email validation

andrewconner:
@stkem

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/common/mail/EmailAddress.scala
+++ b/kifi-backend/modules/common/app/com/keepit/common/mail/EmailAddress.scala
@@ -34,7 +34,10 @@ object EmailAddress {
     }
   }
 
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/common/app/com/keepit/common/mail/EmailAddress.scala
```diff
@@ -34,7 +34,10 @@ object EmailAddress {
     }
   }
 
-  implicit val formMapping: Mapping[EmailAddress] = Forms.email.verifying("invalid_email_address", validate(_).isSuccess).transform(validate(_).get, _.address)
+  implicit val formMapping: Mapping[EmailAddress] = {
+    import play.api.data.format.Formats._
+    Forms.of[String].verifying("error.email", validate(_).isSuccess).transform(validate(_).get, _.address)
+  }
```
2is10, line 40: How about a simple test case with an example of something that validates now that didn’t before?

PR #14633, opened 2015-03-31
an endpoint to mark domain as porn

yingjieMiao:
@stkem 

2 modes: preview mode & read-write mode.


kifi-backend/modules/shoebox/app/com/keepit/model/NormalizedURIRepo.scala
```diff
@@ -204,4 +205,8 @@ class NormalizedURIRepoImpl @Inject() (
   def getFromId(fromId: Id[NormalizedURI])(implicit session: RSession): Seq[NormalizedURI] = {
     (for (t <- rows if t.id > fromId) yield t).list
   }
+
+  def getByRegex(regex: String)(implicit session: RSession): Seq[NormalizedURI] = {
+    (for (u <- rows if (u.url like regex)) yield u).list
```
stkem, line 210: this does not actually do regex matching, you need to use `RLIKE`.

stkem, line 210: not sure if that is intended or not

stkem:

kifi-backend/modules/shoebox/app/com/keepit/model/NormalizedURIRepo.scala
yingjieMiao, line 210: `urlRepo` has a similar query. In fact, on the admin page, this is a separate input box. User can specify things like "porn.com". This should do it.

yingjieMiao, line 210: I do this before the db query:

```
"%" + regex.trim + "%"
```

PR #14634, opened 2015-03-31
Lower waitlist fudge factor for early users we're inviting

andrewconner:
@stkem We're inviting some users by hand, telling them that they're super early. Might as well make it more believable.


stkem:
:-)

PR #14636, opened 2015-03-31
less logging

stkem:
these two lines seem to make up most of the curator logs. not very useful. Probably left over from josh debugging.
@yingjieMiao


yingjieMiao:
nice

PR #14641, opened 2015-03-31
Context Bankruptcy. Alternate POST endpoint.

andrewconner:
@stkem @mrbrown14 @jaredpetker 

https://www.youtube.com/watch?v=ydID4hKCMD8

For context bankruptcy, you won't need to do anything. However, the preferred format is now using a POST. No query params, send all of them in a request JSON object.

``` javascript
{
  'recency': Float,
  'uriContext': Option[String]
  'libContext': Option[String]
}
```


stkem:
\+ @yingjieMiao

stkem:
lgtm

PR #14646, opened 2015-03-31
remove seen items before sorting

yingjieMiao:
@stkem 

Old method can cause problem: during sorting, uris are grouped by topicIds, and we apply decay within each topic. If a topic contains several seen items (and they are likely ranked high within the topic), remaining items in the topic will receive big penalty. Those items later will lose any competition with uris from other topics, esp. topics having no seen items).


stkem:
good catch!

PR #14649, opened 2015-03-31
remove seen libraries before sorting

yingjieMiao:
@stkem 

Similar as URIs. I don't think it actually suffered from the problem. But it's safe to do this anyway.


stkem:
:+1:

PR #14650, opened 2015-03-31
populating push tasks based on app installations/opens. last activity is...

eishay:
... using opened app as well


eishay:
@andrewconner

PR #14653, opened 2015-04-01
App Boy Integration

ahsu1230:
@andrewconner @stkem 

Big Changes:
- `token` in `Device` object is optional
- `MobilePushNotifier` decides whether to use `UrbanAirship` or `AppBoy` service depending on what's given
- `AppBoy` push notification json & client


kifi-backend/modules/eliza/app/com/keepit/realtime/AppBoy.scala
```diff
@@ -0,0 +1,131 @@
...
+          log.info(s"[AppBoy] sent user $userId push notifications to ${activeDevices.size} active devices out of ${allDevices.size}. $notification")
+          // update device state?
+          activeDevices.size
+
+        case None =>
+          airbrake.notify(s"[AppBoy] user $userId not found to send push notifications! $allDevices devices. notification: $notification)")
+          0
+      }
+    }
+  }
```
ahsu1230, line 66: needs to be a future because of `shoeboxClient.getUser`
which causes a couple of other API functions to be `Future[Int]` instead of just `Int`

kifi-backend/modules/eliza/app/com/keepit/realtime/AppBoy.scala
```diff
@@ -0,0 +1,131 @@
...
+    val json = Json.obj(
+      "app_group_id" -> "4212bbb0-d07b-4109-986a-aac019d8062a",
+      "external_user_ids" -> Json.toJson(Seq(user.externalId)),
+      "messages" -> Json.obj(
+        deviceMsgType -> Json.obj(
+          "sound" -> Json.toJson(notification.sound),
+          "title" -> notification.message,
+          "alert" -> notification.message,
+          "extra" -> extraJson
+        )
```
ahsu1230, line 104: only one `app_group_id`. `deviceMsgType` is either `apple_push` or `android_push`
and the `extraJson` information from UrbanAirship also stays as well

kifi-backend/modules/eliza/app/com/keepit/realtime/MobilePushNotifier.scala
```diff
@@ -0,0 +1,68 @@
...
+      deviceRepo.getByUserId(userId, None)
+    }.partition(d => d.token.isDefined)
+
+    val numSentUrbanAirship = urbanAirship.get.notifyUser(userId, devicesWithToken, notification)
+    val numSentAppBoyF = appBoy.get.notifyUser(userId, devicesNoToken, notification)
+
+    numSentAppBoyF.map { numSentAppBoy =>
+      numSentUrbanAirship + numSentAppBoy
+    }
+  }
```
ahsu1230, line 69: differentiates between `UrbanAirship` or `AppBoy` for `notifyUser()`

kifi-backend/modules/eliza/app/com/keepit/realtime/MobilePushNotifier.scala
```diff
@@ -0,0 +1,68 @@
...
+    deviceRepo: DeviceRepo,
+    implicit val executionContext: ExecutionContext) extends MobilePushNotifier with Logging {
+
+  def registerDevice(userId: Id[User], token: Option[String], deviceType: DeviceType, isDev: Boolean, signature: Option[String]): Either[String, Device] = {
+    token match {
+      case Some(tokenStr) => Right(urbanAirship.get.registerDevice(userId, tokenStr, deviceType, isDev, signature))
+      case None if signature.isDefined => Right(appBoy.get.registerDevice(userId, deviceType, isDev, "ab_" + signature.get))
+      case _ => Left("invalid_params")
+    }
+  }
```
ahsu1230, line 53: differentiates between `UrbanAirship` or `AppBoy` for `registerDevice()`.
Server handles having different signatures for `AppBoy`. Right now, in order to be able to use both UrbanAirship and AppBoy, we change the `signature` for Appboy. This is because there's a unique constraint on `(userId, deviceType, signature)`

andrewconner:
Great, thanks for jumping on this so quickly. Sorry for no review yet, just getting caught up with code changes. I'll go through this tonight/early morning.

ahsu1230:
no probs! Thanks for reviewing. I'm sure that this shouldn't affect any UrbanAirship functionality. Hope to test appboy out tomorrow once it's deployed.

kifi-backend/modules/eliza/app/com/keepit/realtime/AppBoy.scala
```diff
@@ -0,0 +1,131 @@
...
+    shoeboxClient: ShoeboxServiceClient,
+    messagingAnalytics: MessagingAnalytics,
+    airbrake: AirbrakeNotifier,
+    implicit val publicIdConfig: PublicIdConfiguration,
+    implicit val executionContext: ExecutionContext) extends Logging {
+
+  def registerDevice(userId: Id[User], deviceType: DeviceType, isDev: Boolean, signature: String): Device = {
+    log.info(s"[AppBoy] Registering device: $deviceType:(no token) for (user $userId, signature $signature)")
+
+    val updatedDevice = db.readOnlyMaster { implicit s =>
```
andrewconner: Since you need a `readWrite` both ways below, it probably makes sense to do all of this in one transaction.

kifi-backend/modules/eliza/app/com/keepit/realtime/AppBoy.scala
```diff
@@ -0,0 +1,131 @@
...
+
+  private def sendNotification(user: User, device: Device, notification: PushNotification): Unit = {
+    val deviceMsgType = device.deviceType match {
+      case DeviceType.IOS => "apple_push"
+      case DeviceType.Android => "android_push"
+    }
+    val extraJson = addExtraJson(notification, device.deviceType)
+
+    val json = Json.obj(
+      "app_group_id" -> "4212bbb0-d07b-4109-986a-aac019d8062a",
```
andrewconner: Since you have a `AppBoyConfig`, might as well throw this in that.

kifi-backend/modules/eliza/app/com/keepit/realtime/DeviceRepo.scala
```diff
@@ -25,11 +25,11 @@ class DeviceRepoImpl @Inject() (val db: DataBaseComponent, val clock: Clock) ext
   type RepoImpl = DeviceTable
   class DeviceTable(tag: Tag) extends RepoTable[Device](db, tag, "device") {
     def userId = column[Id[User]]("user_id", O.NotNull)
-    def token = column[String]("token", O.NotNull)
+    def token = column[String]("token", O.Nullable)
```
andrewconner: Swap this to be `column[Option[String]]("token", O.Nullable)`, then you can remove to `.?` below. Slick 2.1 cares a bit more about how we do it.

kifi-backend/modules/eliza/app/com/keepit/realtime/MobilePushNotifier.scala
```diff
@@ -0,0 +1,71 @@
...
+
+  def notifyUser(userId: Id[User], notification: PushNotification): Future[Int] = {
+    val (devicesWithToken, devicesNoToken) = db.readOnlyMaster { implicit s =>
+      deviceRepo.getByUserId(userId, None)
+    }.partition(d => d.token.isDefined)
+
+    val numSentUrbanAirshipF = urbanAirship.get.notifyUser(userId, devicesWithToken, notification)
+    val numSentAppBoyF = appBoy.get.notifyUser(userId, devicesNoToken, notification)
```
andrewconner, line 61: Why do we not just filter out inactive devices above? Then we could avoid the partitions in each of the clients. Did I miss something they're doing?

kifi-backend/modules/eliza/test/com/keepit/realtime/AppBoyTest.scala
```diff
@@ -0,0 +1,150 @@
...
+    FakeElizaServiceClientModule(),
+    FakeHeimdalServiceClientModule(),
+    FakeActorSystemModule(),
+    FakeHttpClientModule(),
+    FakeElizaStoreModule(),
+    FakeAppBoyModule(),
+    FakeExecutionContextModule()
+  )
+
+  "AppBoy" should {
```
andrewconner, line 36: Awesome. I appreciate the thorough tests so much.

andrewconner:
Minor comments. Great job, push when you're happy with it.

kifi-backend/modules/eliza/app/com/keepit/realtime/DeviceRepo.scala
ahsu1230: ah right forgot about 2.1

kifi-backend/modules/eliza/app/com/keepit/realtime/MobilePushNotifier.scala
ahsu1230, line 61: so `AppBoy` doesn't do anything with it. but `UrbanAirship` does something with inactive devices. It has this `updateDeviceStates` function. I'm not sure what problems that function is trying to solve exactly, but didn't want to mess with the logic.

andrewconner, line 61: I see. It's fine then.

ahsu1230:
merging, evos, deploy

PR #14658, opened 2015-04-01
updating activity push based on app open

eishay:


eishay:
@andrewconner look at the updatedActivity placeholder please

kifi-backend/modules/shoebox/app/com/keepit/shoebox/cron/ActivityPushSchedualer.scala
```diff
@@ -89,6 +89,10 @@ class ActivityPusher @Inject() (
     implicit val executionContext: ExecutionContext,
     clock: Clock) extends Logging {
 
+  def updatedActivity(kifiInstallation: KifiInstallation): Unit = {
+    //todo(andrew): placeholder
```
andrewconner, line 93: Works for me.

PR #14667, opened 2015-04-01
reduce graph ingestion concurrency

stkem:
hopefully reducing some of the lock contention on the raw seed item table. I'll keep an eye on it to make sure it doesn't fall behind.
@yingjieMiao


yingjieMiao:
ok

PR #14668, opened 2015-04-01
some logging

yingjieMiao:
@stkem


stkem:
What are you trying to track down? This is going to generate _a lot_ of log lines (as in several per second). Will be hard to get through.

yingjieMiao:
Trying to see what's going on there (esp. those recursive calls). At least I can track per user. 
I can remove them after a while.

PR #14676, opened 2015-04-01
do not send welcome email when signed up for twitter waitlist

ahsu1230:
@andrewconner @sterlingz


kifi-backend/modules/shoebox/app/com/keepit/controllers/core/AuthHelper.scala
```diff
@@ -190,17 +190,25 @@ class AuthHelper @Inject() (
...
-      SafeFuture { userCommander.sendWelcomeEmail(user, withVerification = true, Some(unverifiedEmail)) }
-    } else {
-      db.readWrite { implicit session =>
-        emailAddressRepo.getByAddressOpt(emailAddress) map { emailAddr =>
-          userRepo.save(user.copy(primaryEmail = Some(emailAddr.address)))
+
+    val cookieTargetLibId = request.cookies.get("publicLibraryId")
+    val cookieIntent = request.cookies.get("intent")
+    val discardedCookies = Seq(cookieTargetLibId, cookieIntent).flatten.map(c => DiscardingCookie(c.name)) :+ DiscardingCookie("inv")
+    val fromTwitterWaitlist = cookieIntent.exists(_.value == "waitlist")
```
ahsu1230, line 197: check intent cookie for waitlist.

we call the signup part (`/signup/twitter?intent=waitlist`) from AuthController.requestToTwitterWaitlist

PR #14678, opened 2015-04-02
inactivate urban airship rows (only with same signature)

ahsu1230:
prevents double pushes + tests!

@andrewconner @eishay


ahsu1230:
and a lot of scalariform

PR #14689, opened 2015-04-02
Trying to track down which task contribute how much DB load

stkem:
this will be reverted after one hour
@yingjieMiao


yingjieMiao:
nice hack. The problem may be more prominent when multiple tasks are running concurrently. 

BTW, in `RecommendationGenerationCommander`, multiple threads are competing to `saveScoredSeedItems`. An alternative is 'scatter-and-gather', i.e. threads concurrently calls remote services for data, and use only 1 thread to `saveScoredSeedItems` (e.g. with a separate lock, or just use plain scala like `Future.sequence.oncompete(save all scored items in this batch)`).

PR #14698, opened 2015-04-02
SecureSocialUserPluginImpl refactor, cleanup

andrewconner:
@eishay @stkem I would appreciate a very thorough code review.

This refactor was initiated by several issues in `internUser`, including a persisting race (see: "We fairly often hit race conditions here"). Much of the code is now annotated, so ideally we can start to remove parts.


kifi-backend/modules/shoebox/app/com/keepit/commanders/AuthCommander.scala
```diff
@@ -145,8 +146,8 @@ class AuthCommander @Inject() (
       allowSignup = true
     )
 
-    val savedIdentity = UserService.save(newIdentity) // Kifi User is created here if it doesn't exist
-    if (!isComplete) { // fix-up: with UserIdentity.isComplete gone, UserService.save creates user in ACTIVE state by default
+    val savedIdentity = socialService.save(newIdentity) // Kifi User is created here if it doesn't exist
```
andrewconner: There's no reason to use SecureSocial's delegate in any injected class. We can inject our implementation directly, and (eventually) not have to follow their trait.

kifi-backend/modules/shoebox/app/com/keepit/commanders/UserCommander.scala
```diff
@@ -642,15 +642,17 @@ class UserCommander @Inject() (
       val filler = Seq.fill(4 - name.length)(0)
       s"$name-$filler"
     } else name
-    def randomNumber = scala.util.Random.nextInt(999)
+    def randomNumber(max: Int) = scala.util.Random.nextInt(max)
```
andrewconner, line 659: This was just to avoid the random alphanumeric usernames that people with short enough names have. Numbers look better to me.

stkem:
lgtm. hard to parse though.

andrewconner:
This has been open too long. Not merged because a test doesn't pass, can't figure out why. I suppose it shows how bad this class is, when it can't be refactored after careful examination + a friendly typechecker.

PR #14702, opened 2015-04-02
Tracking (recos & library page - sharing via email, fb, or twitter)

ahsu1230:
@2is10 @stkem


kifi-backend/modules/shoebox/angular/src/libraries/libraryShareSearch.js
```diff
@@ -62,6 +62,8 @@ angular.module('kifi')
           shareMenu.show();
           contactList.scrollTop(0);
 
+          trackShareEvent('user_clicked_page', { action: 'clickedShareEmail' });
```
2is10, line 65: This is a list of Kifi connections and imported contacts, and the user is allowed to type in any email address. Also, if it’s the user’s own library, this widget is re-used to invite followers. `clickedShareEmail` doesn’t quite capture all of that. It might be worth double-checking with Jen to see if she meant for those cases to all fall under this one umbrella.

2is10:
:+1:

PR #14717, opened 2015-04-03
allow curator to be offline

eishay:


eishay:
@stkem @andrewconner

stkem:
thanks

PR #14726, opened 2015-04-03
Fixing infinite loop?

andrewconner:
@eishay 

Not sure what's wrong, replica lag or if the counter logic is bad.


eishay:
fixed it?

andrewconner:
Yeah, the `count <= 0` logic was a bit off. I'll try to replace it soon, but for now, we'll just send 100 per 30 seconds (still 12,000 an hour).

PR #14733, opened 2015-04-05
relax lock waiting size

yingjieMiao:
@eishay


yingjieMiao:
i see it got stuck when waiting size = 3. Let's see if this helps.

PR #14745, opened 2015-04-06
Rover throttling

leogrim:
@stkem


kifi-backend/modules/rover/app/com/keepit/rover/model/ArticleInfoRepo.scala
```diff
@@ -164,6 +166,12 @@ class ArticleInfoRepoImpl @Inject() (
       }
     }
   }
+
+  def setDomains(p: Int, s: Int)(implicit session: RWSession): Int = {
+    page(p, s).map { info =>
+      (for (r <- rows if r.id === info.id) yield r.domain).update(URI.parse(info.url).toOption.flatMap(_.host).map(_.name))
+    }.sum
+  }
```
yingjieMiao, line 174: `sum` exception when page is empty.

leogrim, line 174: I think they fixed that (as far as the REPL has been recently saying :)

yingjieMiao, line 174: ah, nice.

stkem:
:+1:

PR #14758, opened 2015-04-06
filter out small libraries from sitemap generator and making sure we're ...

eishay:
...not loading empty libraries


eishay:
@andrewconner fixing the bug you sent me few minutes ago

PR #14772, opened 2015-04-07
scraper tweak

yingjieMiao:
@eishay 

Here is my understanding:
scraper stucks at the reactive lock queue size limit, partly b/c the aggressive pull frequency. It's pulling 8 tasks per second, and when we have large imports/rescrape, the lock's queue size soon becomes maxed, and seems to stuck there forever.


eishay:
thanks! lets see how it works :-)

PR #14780, opened 2015-04-07
fixing library attribution hovercards

2is10:
@andrewconner 

Apparently the problem had something to do with using an isolate scope in a transcluded directive. I don’t fully understand, but this seems to work around it.

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/libraryMiniCard.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/libraryMiniCard.js
@@ -8,15 +8,8 @@ angular.module('kifi')
     return {
       restrict: 'A',
       replace: true,
-      scope: {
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/libraryMiniCard.js
```diff
@@ -8,15 +8,8 @@ angular.module('kifi')
...
       replace: true,
-      scope: {
-        refLibrary: '&library',
-        invite: '='
-      },
       templateUrl: 'libraries/libraryMiniCard.tpl.html',
       link: function (scope) {
-        scope.library = _.cloneDeep(scope.refLibrary());
-        scope.showMiniCard = true;
-
```
2is10, line 19: Note: No longer showing card until all of its data has loaded.

PR #14786, opened 2015-04-07
retire a json field

yingjieMiao:
@stkem 

metaData is not used anymore. We could just remove the field entirely, but would love to sync with mobile guys later.

Removing this just makes the transition clearer.


stkem:
:+1:

PR #14788, opened 2015-04-07
fixing cover image uploading in production

2is10:
@andrewconner


kifi-backend/modules/shoebox/angular/src/app.js
```diff
@@ -53,7 +53,7 @@ angular.module('kifi', [
   '$compileProvider', '$FBProvider', 'env',
   function ($compileProvider, $FBProvider, env) {
     // ng-perf.com/2014/10/24/simple-trick-to-speed-up-your-angularjs-app-load-time/
-    $compileProvider.debugInfoEnabled(env.dev);
+    // $compileProvider.debugInfoEnabled(env.dev);
```
andrewconner, line 56: Whoa. Why?

2is10, line 56: `.$scope` is considered “debug info” I guess. I’m going to write a directive that will let us handle `<input type=file>` changes properly, and then we can change this back.

Full story:
https://github.com/angular/angular.js/issues/1375

PR #14789, opened 2015-04-07
introducing kf-file-change directive and re-disabling debug info in production

2is10:
@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/layout/main/MainCtrl.js
+++ b/kifi-backend/modules/shoebox/angular/src/layout/main/MainCtrl.js
@@ -117,20 +117,18 @@ angular.module('kifi')
 
     $scope.disableBookmarkImport = true;
 
-    $scope.allowUpload = function (elem) {
-      $scope.$apply(function () {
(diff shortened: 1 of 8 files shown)


kifi-backend/modules/shoebox/angular/src/layout/main/MainCtrl.js
```diff
@@ -179,11 +177,11 @@ angular.module('kifi')
...
 
           var tooSlowTimer = $timeout(function () {
-            $scope.importFileStatus = 'Your bookmarks are still uploading... Hang tight!';
+            $scope.importFileStatus = 'Uploading... Hang tight!';
             $scope.disableBookmarkImport = false;
-          }, 8000);
+          }, 3000);
 
-          $scope.importFileStatus = 'Uploading! May take a bit, especially if you have a lot of links.';
+          $scope.importFileStatus = '';
```
2is10, line 184: Showing that message instantly (in dark orange) and only very briefly, in the success case, looked like an error. Now only showing a message after 3 seconds.

PR #14791, opened 2015-04-08
not responding to JSON request with a 301 permanent redirect to HTML page

2is10:
@leogrim


2is10:
We should also fix KifiSiteRouter, as we discussed.

leogrim:
Thanks!

PR #14792, opened 2015-04-08
Shoebox Rover ingestion

leogrim:
@stkem @eishay @andrewconner @yingjieMiao 
I still have a couple of small Asana tasks left in that area but this is in a pretty good state :)

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/shoebox/rover/ArticleUpdateNormalizer.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/shoebox/rover/ArticleUpdateNormalizer.scala
@@ -8,9 +8,11 @@ import com.keepit.common.logging.Logging
 import com.keepit.common.net.{ WebService, URI }
 import com.keepit.model._
 import com.keepit.normalizer._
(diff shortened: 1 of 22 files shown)


kifi-backend/modules/shoebox/app/com/keepit/shoebox/rover/ArticleUpdateNormalizer.scala
```diff
@@ -0,0 +1,176 @@
...
+  private def removeRedirectRestriction(uri: NormalizedURI): NormalizedURI = uri.restriction match {
+    case Some(restriction) if Restriction.redirects.contains(restriction) => uri.copy(restriction = None)
+    case _ => uri
+  }
+
+  private def updateRedirectRestriction(uri: NormalizedURI, restriction: Restriction): NormalizedURI = {
+    if (Restriction.redirects.contains(restriction)) uri.copy(restriction = Some(restriction)) else removeRedirectRestriction(uri)
+  }
+
+}
```
yingjieMiao: Please consider adding some comments/tests to this class. It's very hard to follow and maintain.

PR #14793, opened 2015-04-08
Different way of assigning Seq Nums for Raw Seed Items

stkem:
Instead of being a separate plugins it’s now done synchronously as part
of data ingestions, which is basically the only part that writes to
this table. This should stop the db lock contention issues.
Not a big change, but hopefully the right one.

@yingjieMiao @eishay

### DIFF
--- a/kifi-backend/modules/curator/app/com/keepit/curator/commanders/SeedIngestionCommander.scala
+++ b/kifi-backend/modules/curator/app/com/keepit/curator/commanders/SeedIngestionCommander.scala
@@ -28,6 +28,7 @@ class SeedIngestionCommander @Inject() (
     keepInfoRepo: CuratorKeepInfoRepo,
     libMembershipRepo: CuratorLibraryMembershipInfoRepo,
     experimentCommander: RemoteUserExperimentCommander,
+    rawSeedSeqNumAssigner: RawSeedItemSequenceNumberAssigner,
     db: Database) extends Logging {
 
   val INGESTION_BATCH_SIZE = 50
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/curator/app/com/keepit/curator/commanders/SeedIngestionCommander.scala
```diff
@@ -50,8 +51,11 @@ class SeedIngestionCommander @Inject() (
 
       val ingestKeepsF = ingestAllKeeps().flatMap { _ =>
         log.info("XYZ: ingest all keeps future completed")
+        rawSeedSeqNumAssigner.assignSequenceNumbers()
         val userIds = usersToIngestGraphDataFor()
-        FutureHelpers.sequentialExec(userIds)(ingestTopUris)
+        val resFut = FutureHelpers.sequentialExec(userIds)(ingestTopUris)
+        rawSeedSeqNumAssigner.assignSequenceNumbers()
```
yingjieMiao, line 57: Do you need to wait for `resFut` completes, then call `assignSeq`?

kifi-backend/modules/curator/app/com/keepit/curator/commanders/SeedIngestionCommander.scala
```diff
@@ -50,8 +51,11 @@ class SeedIngestionCommander @Inject() (
 
       val ingestKeepsF = ingestAllKeeps().flatMap { _ =>
         log.info("XYZ: ingest all keeps future completed")
+        rawSeedSeqNumAssigner.assignSequenceNumbers()
```
yingjieMiao, line 54: why call this twice?

stkem, line 54: to reduce latency. The graph ingestion below can take quite a while and I want new keeps to get a sequence number quickly.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/SeedIngestionCommander.scala
stkem, line 57: I do indeed! Good find. Fixing it now.

yingjieMiao:
Let's see how it works. 

As I commented in an earlier PR, an alternative is to gather scored items for 8 (or whatever `locksize`) users concurrently, and persist all data in 1 thread. In that case, the db sessions on that table is roughly reduced from `8 + 1` to `1 + 1`, 1 being the seq assigner. 

I am not fully sure about my theory though.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/SeedIngestionCommander.scala
yingjieMiao, line 54: ah, I see, it's actually for the previous step. Makes sense.

PR #14800, opened 2015-04-08
Fixing double send of AppBoy push notifications.

andrewconner:
@eishay


kifi-backend/modules/eliza/app/com/keepit/realtime/AppBoy.scala
```diff
@@ -56,18 +56,17 @@ class AppBoy @Inject() (
     updatedDevice
   }
 
+  // When this is the only push provider, refactor this to just take userId and notification. No need to get list of devices.
   def notifyUser(userId: Id[User], allDevices: Seq[Device], notification: PushNotification): Future[Int] = {
     log.info(s"[AppBoy] Notifying user: $userId with $allDevices")
-    val (activeDevices, _) = allDevices.partition(d => d.state == DeviceStates.ACTIVE)
+    val deviceTypes = allDevices.filter(_.state == DeviceStates.ACTIVE).groupBy(_.deviceType).keys.toList
```
andrewconner, line 62: This is the fix. Previously, we were sending one per active device. Now we're sending only one, total.

PR #14809, opened 2015-04-09
Exclude Foursquare, Instagram, Periscope from Twitter imports

leogrim:
@andrewconner @eishay


andrewconner:
Looks good, thanks!

PR #14813, opened 2015-04-10
Backfilling Twitter sync

andrewconner:
@stkem 

Strategy is:
1. Given `syncState: TwitterSyncState`, fetch all until the current max. This is typically just one batch. With no previous fetches, this is always one batch.
2. Fetch before `syncState`, by setting the fetch upper bound to existing lower bound. Recurse.

Infinite loops could potentially occur in two ways:
1. We fetch the same or similar batches over and over, and don't progress.
2. We don't stop when twitter returns no tweets.

Could you verify both of these?

Seems to work, I fetched 245 links over 500+ tweets from my account.


andrewconner:
I'm reasonably confident in it. Trying it out.

PR #14814, opened 2015-04-10
model and repo to support feedback

yingjieMiao:
@stkem 

I took out the topic ids, as they can be queried from Cortex later on (if we want to retrain model). During online update, there is no need to store them. Stored values can be problematic during model transition. 

View data will be persisted as mobile supports that.

### DIFF
--- a/kifi-backend/modules/curator/app/com/keepit/curator/model/UriRecoFeedback.scala
+++ b/kifi-backend/modules/curator/app/com/keepit/curator/model/UriRecoFeedback.scala
@@ -11,20 +11,21 @@ case class UriRecoFeedback(
     updatedAt: DateTime = currentDateTime,
     userId: Id[User],
     uriId: Id[NormalizedURI],
-    viewed: Option[Boolean],
-    clicked: Option[Boolean],
-    kept: Option[Boolean],
-    liked: Option[Boolean],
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/curator/app/com/keepit/curator/model/UriRecoFeedback.scala
```diff
@@ -0,0 +1,30 @@
...
+case class UriRecoFeedback(
+    id: Option[Id[UriRecoFeedback]],
+    createdAt: DateTime = currentDateTime,
+    updatedAt: DateTime = currentDateTime,
+    userId: Id[User],
+    uriId: Id[NormalizedURI],
+    viewed: Option[Boolean],
+    clicked: Option[Boolean],
+    kept: Option[Boolean],
+    liked: Option[Boolean],
```
stkem: This has the same problem as the `uri_recommendation` table, namely that we would need to alter the table whenever we add a kind of feedback for the user (and possibly when we remove one). We talked about this yesterday, the feedback should be a single column with an enum (or something like that) in it, similar to how we deal with states.

yingjieMiao:
@stkem 

updated.

stkem:
lgtm

PR #14833, opened 2015-04-10
this seems to have memcached issue

yingjieMiao:
@eishay @andrewconner @stkem 

tmp disable this (doesn't cause incorrect logic). See if this helps.


yingjieMiao:
I think I trigger some false alarms earlier. Bottom line is uriRepo.get(uriId) is causing memcached problem.

stkem:
Why would this particular call have memcached issues? Isn't it likely just the most frequent one?

yingjieMiao:
this is what I see from the log. Will disable this for now see if things get better.

PR #14836, opened 2015-04-10
restartable memcached client

yingjieMiao:
@eishay

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/common/cache/MemcachedClientProvider.scala
+++ b/kifi-backend/modules/common/app/com/keepit/common/cache/MemcachedClientProvider.scala
@@ -1,5 +1,7 @@
 package com.keepit.common.cache
 
+import java.util.concurrent.TimeUnit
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/common/app/com/keepit/common/cache/MemcachedClientProvider.scala
```diff
@@ -0,0 +1,25 @@
...
+      new MemcachedClient(AddrUtil.getAddresses(endpoint))
+    }.getOrElse(throw new RuntimeException("Bad configuration for memcached: missing host(s)"))
+  }
+
+  def get(): MemcachedClient = {
+    client
+  }
+
+  def recreate(): Unit = {
+    client = create()
```
eishay, line 26:
``` scala
val old = client
client = create()
old.kill()//or something like this...
```

yingjieMiao:
merging, will babysit this.

PR #14851, opened 2015-04-11
Removing logging, silently failing on a deadlock.

andrewconner:
This is a bad place to do intensive lookups. Imports have gotten really slow, need to figure out a better way.


andrewconner:
@eishay @stkem fyi

PR #14855, opened 2015-04-12
trimming down of libraries and keepers that we need to filter and get me...

eishay:
...tadata for.

tracked it down when noticed we're pulling from cache hundreds of entities (like library count) for this call. its overloads the cache


eishay:
@2is10

2is10:
:+1:

PR #14856, opened 2015-04-12
reduce noise

yingjieMiao:
@eishay


eishay:
what's the threshold where you think gets cached into a bad state?

yingjieMiao:
most bulkget has size < 400. 500 is just a heuristic, so it's not very noisy. LibraryChecker is the only one that reaches a size > 3000.

PR #14863, opened 2015-04-13
more friendly bulkget

yingjieMiao:
@eishay @stkem @andrewconner @leogrim @ahsu1230 

huge bulkget tends to crash the memcached client, and it seems to take ~ 10 minutes for the client to recover. Any cache access during this time window is likely to throw timeout exceptions.


yingjieMiao:
any concern with latency (when key size > 200) ?

eishay:
great!

stkem:
:+1:

PR #14864, opened 2015-04-13
Using the default font-weight (400) instead of 100 (hairline) in emails

2is10:
@andrewconner @stkem 

It looks like maybe someone thought `100` means “normal” font-weight. These occurrences are on font sizes of 10 and 18, which are too small for a hairline font weight.


andrewconner:
Thanks

PR #14865, opened 2015-04-13
a font-smoothing mixin that sets -moz-osx-font-smoothing too

2is10:
@ahsu1230 introduced me to this property, added in Firefox 25 just for Macs, not yet documented in MDN.

In most places we were already using an apparently built-in `font-smoothing` mixin that outputs `-webkit-font-smoothing`. Now we have our own mixin that outputs both properties, and we’re using it consistently.

cc @andrewconner @stkem


andrewconner:
Nice

PR #14872, opened 2015-04-14
Destination-based Rover throttling

leogrim:
@stkem @eishay


kifi-backend/modules/common/app/com/keepit/common/time/RandomDelay.scala
```diff
@@ -0,0 +1,9 @@
+package com.keepit.common.time
+
+import scala.concurrent.duration._
+import scala.util.Random
+
+class RandomDelay(maxDelay: Duration) extends (() => Duration) {
```
stkem: A minor point, and you certainly don't need to change that if you don't want to, but I generally find it easier to read if you don't use the syntactic sugar function type, but rather the full type in situations like this, i.e. `... extends Function0[Duration]`

kifi-backend/modules/rover/app/com/keepit/rover/document/FetchThrottlingException.scala
```diff
@@ -0,0 +1,86 @@
...
+import com.keepit.common.logging.AccessLog
+import com.keepit.common.net.URI
+import com.keepit.common.time._
+import org.joda.time.DateTime
+import scala.concurrent.ExecutionContext
+import scala.concurrent.duration._
+import play.api.libs.json.{ JsValue, Json, Format }
+
+case class FetchThrottlingException(url: String, destinationUrl: String, nextFetch: DateTime)
+  extends Throwable(s"Throttling fetch of $url with destination $destinationUrl. Try again at $nextFetch")
```
stkem: Extending `Throwable` can be dangerous in scala since it is also used for control flow (specifically return statements). Better extends `Exception`, especially since that is what this is :-)

PR #14873, opened 2015-04-14
use denormalized keep counts

yingjieMiao:
@eishay @leogrim 

reduce cache load


yingjieMiao:
oops, looks like @eishay was working on this.

PR #14874, opened 2015-04-14
using the denormalized library keep count instead of pounding on the db ...

eishay:
...and cache like crazy


eishay:
@andrewconner @leogrim @yingjieMiao @ahsu1230

PR #14900, opened 2015-04-15
semi fixing broken private invitations

eishay:


eishay:
DO NOT MERGE

eishay:
@andrewconner @ahsu1230 please review this one.
I did make some of the test pass but its still broken and I'm not sure how the logic should work.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -1031,23 +1031,24 @@ class LibraryCommander @Inject() (
...
       (lib, listInvites)
     }
+    println(lib.visibility != LibraryVisibility.PUBLISHED)
+    println(listInvites.isEmpty)
 
-    if (lib.kind == LibraryKind.SYSTEM_MAIN || lib.kind == LibraryKind.SYSTEM_SECRET)
+    if (lib.kind == LibraryKind.SYSTEM_MAIN || lib.kind == LibraryKind.SYSTEM_SECRET) {
       Left(LibraryFail(FORBIDDEN, "cant_join_system_generated_library"))
-    else if (lib.visibility != LibraryVisibility.PUBLISHED && listInvites.isEmpty)
+    } else if (lib.visibility != LibraryVisibility.PUBLISHED && !listInvites.exists(_.userId.nonEmpty)) {
```
eishay, line 1049: I used `!listInvites.exists(_.userId.nonEmpty)` instead of `isEmpty` and it made the test line #657 pass, but I'm not sure its the right things todo. 
Anyway the test is still broken few lines down.
I'll be happy to co-work on it but it wold help to have someone who understand the logic better ( @ahsu1230 or @andrewconner )

ahsu1230, line 1049: I don't think line 657 should pass in the test. That tests that the `hashedPassPhrase` is incorrect even though the `authToken` is correct. Can I ask in what situation does following a private library break (open from email, find on kifi.com, etc.)? Sorry I wasn't around earlier to help T_T
I'll be happy to help look at it tomorrow though.

eishay, line 1049: Invite a user to a private library, have the user use the link in the email, clicking on "follow" returns a 404

2is10, line 1049: I think Eishay is on the right track. I think the `authToken` should only be considered if the logged-in user doesn’t have an invite.

ahsu1230, line 1049: @2is10 here's a PR in the works. #14903

PR #14901, opened 2015-04-15
Read replica for curator

stkem:
\+ less aggressive seq num assignment (to make ingestion faster)

@eishay


eishay:
nice :-)
it caught up

[image: Inline image 1]

PR #14903, opened 2015-04-15
fixing following a private library

ahsu1230:
@eishay @andrewconner @2is10 

Broken case today:
Friend B invites me (kifi user, not email) to his private library. I open up the library by email and try to follow. It'll be broken =(

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
@@ -1030,12 +1030,12 @@ class LibraryCommander @Inject() (
     }
   }
 
-  def joinLibrary(userId: Id[User], libraryId: Id[Library], authToken: Option[String] = None)(implicit eventContext: HeimdalContext): Either[LibraryFail, Library] = {
+  def joinLibrary(userId: Id[User], libraryId: Id[Library], authToken: Option[String] = None, hashedPassPhrase: Option[HashedPassPhrase] = None)(implicit eventContext: HeimdalContext): Either[LibraryFail, Library] = {
     val (lib, inviteList) = db.readOnlyMaster { implicit s =>
       val lib = libraryRepo.get(libraryId)
-      val tokenInvites = authToken.map { token =>
-        libraryInviteRepo.getByLibraryIdAndAuthToken(libraryId, token)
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -1030,25 +1030,23 @@ class LibraryCommander @Inject() (
...
-      } else { // public or private library (called from kifi.com)
-        libraryInviteRepo.getWithLibraryIdAndUserId(libraryId, userId)
-      }
-      (lib, listInvites)
+      val tokenInvites = authToken.map { token =>
+        libraryInviteRepo.getByLibraryIdAndAuthToken(libraryId, token)
+      } getOrElse Seq.empty
+      val libInvites = libraryInviteRepo.getWithLibraryIdAndUserId(libraryId, userId)
+      val allInvites = tokenInvites ++ libInvites
+      (lib, allInvites)
```
ahsu1230, line 1041: @eishay I think the idea behind your original PR was about right =) thanks for looking into this

2is10:
I’m not convinced it’s okay to use `authToken` without the passphrase here. An uninvited “attacker” with an `authToken` can easily construct this follow request and thereby gain access to the private library.

ahsu1230:
so we don't always give a user the `hashedPassPhrase`. If friend B invites me (kifi user, not email) to his private library, the email I receive does not have the pass phrase. If I open the library by email, I can see the library without having to input a pass phrase. Is this right?

2is10:
Yes. In that case, we should ignore the `authToken` allow the follow because the logged-in user was invited to the library. This is the case that is broken today.

2is10:
I don’t think we should ever be giving out an `authToken` without a passphrase, btw.

ahsu1230:
take back what I said about `hashedpassphrase`. It's still in, so the new code should also look for library invites with logged in userId.

If the library is private, but there are no library invites (with userId or with correct authToken/passphrase), then we return `Forbidden`

2is10:
This looks good to me. Thanks.

If convenient, it’d be good to add a test case that passes now but would fail with the old code, to help avoid a regression.

ahsu1230:
Ok, I'll see what cases I can fit in. Thanks for looking it over!

PR #14905, opened 2015-04-15
Check for escaped query parameters in canonical urls (from Rover)

leogrim:


2is10:
Thanks for porting. What about the test? I think it really helps clarify.

leogrim:
Planning on porting over all tests from Scraper: https://app.asana.com/0/27930653682008/29229947277076

PR #14906, opened 2015-04-15
mobile endpoint for editing keep notes

ahsu1230:
@andrewconner @DerekBurch @thassss @mrbrown14 

use `POST    /m/1/keeps/:id` to edit keep info. Send a body with

```
{
   "title" : "new title",
   "note" : "new comment"
}
```

to update the keep. I know title isn't part of the spec, but wouldn't hurt to have it now =)

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileKeepsController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileKeepsController.scala
@@ -180,6 +180,26 @@ class MobileKeepsController @Inject() (
     }
   }
 
+  def editKeepInfo(id: ExternalId[Keep]) = UserAction(parse.tolerantJson) { request =>
+    val keepOpt = db.readOnlyMaster { implicit s => keepRepo.getOpt(id).filter(_.isActive) }
+    keepOpt match {
+      case None =>
+        NotFound(Json.obj("error" -> "not_found"))
+      case Some(keep) =>
+        val json = request.body
+        val titleOpt = (json \ "title").asOpt[String]
+        val noteOpt = (json \ "note").asOpt[String]
+
+        val newTitle = titleOpt map { _.trim } filterNot { _.isEmpty }
+        val newNote = noteOpt map { _.trim } filterNot { _.isEmpty }
+
+        db.readWrite { implicit s =>
+          keepRepo.save(keep.copy(title = newTitle, note = newNote))
+        }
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileLibraryController.scala
```diff
@@ -159,7 +159,7 @@ class MobileLibraryController @Inject() (
           keepDataList.map { keepData =>
             val keep = keepRepo.get(keepData.id)
             val keepImageUrl = keepImageCommander.getBestImageForKeep(keep.id.get, ScaleImageRequest(MobileLibraryController.defaultKeepImageSize)).flatten.map(keepImageCommander.getUrl)
-            val keepObj = Json.obj("id" -> keep.externalId, "title" -> keep.title, "imageUrl" -> keepImageUrl, "hashtags" -> Json.toJson(collectionRepo.getTagsByKeepId(keep.id.get)))
+            val keepObj = Json.obj("id" -> keep.externalId, "title" -> keep.title, "note" -> keep.note, "imageUrl" -> keepImageUrl, "hashtags" -> Json.toJson(collectionRepo.getTagsByKeepId(keep.id.get)))
```
ahsu1230, line 162: you should now see a "note" field in keep objects (if it has a note) when calling `POST /m/1/libraries`
Use this to see what your previous note was for a particular url

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileKeepsController.scala
```diff
@@ -180,6 +180,26 @@ class MobileKeepsController @Inject() (
...
+      case Some(keep) =>
+        val json = request.body
+        val titleOpt = (json \ "title").asOpt[String]
+        val noteOpt = (json \ "note").asOpt[String]
+
+        val newTitle = titleOpt map { _.trim } filterNot { _.isEmpty }
+        val newNote = noteOpt map { _.trim } filterNot { _.isEmpty }
+
+        db.readWrite { implicit s =>
+          keepRepo.save(keep.copy(title = newTitle, note = newNote))
```
andrewconner, line 197: Make sure mobile knows that they _have_ to send a note and title, or it'll be removed. An alternative is to fall back to the existing title or note if the field is omitted from the request JSON.

andrewconner:
Looks great

PR #14917, opened 2015-04-16
ByteArrayCounter, 	feedback tracking commander

yingjieMiao:
@stkem 

this deals with storage/update. 
Inference/boosting part comes later.


kifi-backend/modules/curator/app/com/keepit/curator/feedback/ByteArrayCounter.scala
```diff
@@ -0,0 +1,73 @@
...
+    val arr = toArray()
+    (indexes zip values).foreach { case (i, x) => arr(i) = x }
+    ByteArrayCounter.fromArray(arr)
+  }
+
+  def toArray(): Array[Int] = Array.tabulate(bytes.size) { i => get(i) }
+}
+
+object ByteArrayCounter {
+  val MAX_COUNT = 255
```
stkem: Why is this configurable? Won't it always be 255? (being a byte, after all)

kifi-backend/modules/curator/app/com/keepit/curator/feedback/ByteArrayCounter.scala
```diff
@@ -0,0 +1,73 @@
+package com.keepit.curator.feedback
+
+import play.api.libs.json._
+
+// Immutable operations. Optimized for thread-safe.
+case class ByteArrayCounter(bytes: Array[Byte]) {
```
stkem: Why a `case class` instead of a regular `class`?

stkem:
lgtm overall, but how will we know if rescaling happened? Seems important.

kifi-backend/modules/curator/app/com/keepit/curator/feedback/ByteArrayCounter.scala
yingjieMiao: more like a syntatic sugar for the constant 255. Named constants are easier to search, and prevents typo. As you can see, the constant is used in a few places.

kifi-backend/modules/curator/app/com/keepit/curator/feedback/ByteArrayCounter.scala
yingjieMiao: You are right. I will change it to regular class to restrict access to those bytes.

stkem:
:+1:

PR #14927, opened 2015-04-16
Twitter accept, set library pic and description

andrewconner:


kifi-backend/modules/shoebox/app/com/keepit/common/social/twitter/RawTweet.scala
```diff
@@ -62,7 +62,7 @@ object RawTweet {
 
   object RawRetweet {
     implicit def format: Reads[RawRetweet] = (
-      (__ \ 'id_str).read[TwitterId] and
+      (__ \ 'id).read[TwitterId] and
```
andrewconner, line 65: This fixes the retweet parsing error. Added a test. @eishay

kifi-backend/modules/shoebox/app/com/keepit/commanders/TwitterWaitlistCommander.scala
```diff
@@ -121,5 +155,68 @@ class TwitterWaitlistCommanderImpl @Inject() (
         Left(s"Couldn't accept $userId. $entryOpt, $suiOpt, $syncOpt")
     }
   }
+
+  private def updateUserShow(sui: SocialUserInfo, handle: String): Future[Option[TwitterUserShow]] = {
```
andrewconner, line 159: I should probably move this elsewhere. Will give it some thought, we have a few places that do twitter API calls.

PR #14933, opened 2015-04-16
more logging, and do not shutdown mc client

yingjieMiao:
@eishay 

shutdown client causes bugs, other running threads are trying to get stuff out of it. 

GC will eventually kill those clients, does it?


eishay:
it wont, if the connection is open we need to close it

yingjieMiao:
OK. I will add it back.

eishay:
we can consider closing it in a future 5 seconds form `now()` to let current clients get done with their queries.

PR #14938, opened 2015-04-17
lock during recreation

yingjieMiao:
@eishay 

I think the 'compare and create()' operation is not safe. Multiple creates can happen, and they overwrite each other.


eishay:
great catch!

PR #14939, opened 2015-04-17
maybe fixing curator slave db issues

stkem:
@yingjieMiao @eishay

### DIFF
--- a/kifi-backend/modules/curator/app/com/keepit/curator/CuratorSlickModule.scala
+++ b/kifi-backend/modules/curator/app/com/keepit/curator/CuratorSlickModule.scala
@@ -2,6 +2,7 @@ package com.keepit.curator
 
 import scala.slick.jdbc.JdbcBackend.{ Database => SlickDatabase }
 
+import scala.util._
 import com.google.inject.{ Provides, Singleton }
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/curator/app/com/keepit/curator/CuratorSlickModule.scala
```diff
@@ -12,7 +13,14 @@ import play.api.db.DB
...
   def masterDatabase = SlickDatabase.forDataSource(DB.getDataSource("shoebox")(Play.current))
 
-  override def slaveDatabase = None
+  override def slaveDatabase = Try(SlickDatabase.forDataSource(DB.getDataSource("shoeboxslave")(Play.current))) match {
+    case Success(db) =>
+      println("loaded slave db")
+      Some(db)
+    case Failure(e) =>
+      println(s"could not load slave db for: $e")
+      None
```
yingjieMiao, line 22: println?

stkem, line 22: I stole this from the shoebox module. I'm not sure but I think this runs before the logger is initialized.

PR #14940, opened 2015-04-17
admin flow for twitter sync

ahsu1230:
@andrewconner 

few things
- switched admin twitter stuff out of the website controller and into an admin controller
- both the waitlist and the acceptUser have an admin page
- admin page has userId, libraryId, user email, twitter handle, link to both admin library page and real kifi library page
- has an error message

future work: better error messages. Maybe actions to do when you see an error message?


andrewconner:
Awesome, great job.

andrewconner:
After you push, we'll sync with Mark about what he needs next.

PR #14942, opened 2015-04-17
more stable tests

eishay:
@ahsu1230 the test randomly failed on me, not sure why. i think this may help


ahsu1230:
huh that's so weird. Thanks though

PR #7904, opened 2014-04-16
Safer simple graph persistence mechanism

leogrim:

PR #14948, opened 2015-04-17
Revamp S3AsyncHelper to use explicit inbox directory in all image stores

leogrim:
@andrewconner @stkem @eishay


andrewconner:
Did you figure out why Yasu hated regular temp directories?

leogrim:
Not exactly, per his words "FileUtils.getTempDirectory returns the systems temp directory, right. Writing a big data file to the systems temp directory is often problematic." https://github.com/kifi/keepit/pull/7904

Doesn't add much to make a separate directory, most of the work here is unrelated to that.

PR #14950, opened 2015-04-17
adding numFriends to top level user PYMK

ahsu1230:
@leogrim @andrewconner @jaredpetker


kifi-backend/modules/common/app/com/keepit/model/FriendRecommendations.scala
```diff
@@ -5,18 +5,6 @@ import com.keepit.social.BasicUser
 import com.keepit.social.BasicUser.{ mapUserIdToInt, mapUserIdToBasicUser }
 
 case class FriendRecommendations(basicUsers: Map[Id[User], BasicUser],
-  mutualFriendConnectionCounts: Map[Id[User], Int],
+  userConnectionCounts: Map[Id[User], Int],
   recommendedUsers: Seq[Id[User]],
   mutualFriends: Map[Id[User], Seq[Id[User]]])
-
-object FriendRecommendations {
```
ahsu1230, line 12: unused

PR #14953, opened 2015-04-17
a few trivial tests and use unique index

yingjieMiao:
@stkem


stkem:
lgtm. What is the status of the rest? In particular actually recording feedback.

PR #14962, opened 2015-04-19
drop throwing airbrakes on {"message":"Missing recipients"} 4xx errors f...

eishay:
...rom appboy since its too noise and we record it in the db


eishay:
@andrewconner @ahsu1230

ahsu1230:
oo thanks for this!

PR #14971, opened 2015-04-20
A few changes for easier QAing

stkem:
-The fastlane is now an experiment instead of being hard coded
-The subsampling and the new scoring are split into separate experiments
-experiments are looked up on the fly in each batch instead of once on start (there is an in memory cache
-new scores are shifted up to be comparable to old scores (also important for persona recos)
-No longer use non-linear scoring on retrieval for user with the next gen scoring
@yingjieMiao


kifi-backend/modules/curator/app/com/keepit/curator/commanders/RecommendationGenerationCommander.scala
```diff
@@ -241,7 +238,7 @@ class RecommendationGenerationCommander @Inject() (
         db.readWrite(attempts = 2) { implicit s =>
           items foreach {
             case (item, masterScore) =>
-              if (masterScore > 0.5) {
+              if (masterScore > 16.0f) {
```
yingjieMiao, line 241: this is shifted by 15.5, above was shifted by 15. Not sure if this is intended. In any case, it's perhaps helpful to have a named constant for that.

stkem, line 241: Yup, this was intended. The previous threshold was too soft (new threshold is an effective master score of `1`, was `0.5` previously.
And you are right, a named constant would be good. I'll add that.

PR #14975, opened 2015-04-20
sending Keep.note down to extension

2is10:
@ahsu1230

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeeperInfo.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeeperInfo.scala
@@ -6,7 +6,7 @@ import com.keepit.common.db.ExternalId
 import com.keepit.common.time.DateTimeJsonFormat
 import com.keepit.model._
 import com.kifi.macros.json
-import play.api.libs.json.{ __, JsObject, Writes }
+import play.api.libs.json.{ __, JsObject, JsValue, Writes }
 import play.api.libs.functional.syntax._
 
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/KeeperInfo.scala
```diff
@@ -80,13 +80,27 @@ object KeepData {
...
     (__ \ 'image).writeNullable[String] and
+    (__ \ 'note).writeNullable[String] and
     (__ \ 'tags).writeNullable[Seq[String]].contramap[Seq[String]](Some(_).filter(_.nonEmpty))
   )(unlift(MoarKeepData.unapply))
+
+  // temporarily auto-populating note field with tags if necessary
+  implicit val writes = new Writes[MoarKeepData] {
+    def writes(o: MoarKeepData): JsValue = {
+      val o2 = if (o.tags.nonEmpty && o.note.isEmpty) {
+        o.copy(note = Some(o.tags.map(t => '#' + t.replace(' ', '\u00a0')).mkString(" ")))
```
ahsu1230, line 97: nice, this is the non-breaking space character?

ahsu1230:
:+1: ty!

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeeperInfo.scala
2is10, line 97: yes

PR #14978, opened 2015-04-20
Sharding reco generation

stkem:
Code is getting to be a bit messy. I’ll clean that up soon.

@eishay


andrewconner:
Nice. Is this all that's needed to scale horizontally for recos?

stkem:
no, not quite. This is the groundwork, but there are still some database related bottle necks to be removed. Otherwise adding more machines will not actually make it substantially faster.

PR #14988, opened 2015-04-21
Refactor Rover actors (+ Introduce RoverImageInfo, ArticleImage)

leogrim:
@stkem @andrewconner


leogrim:
This is worth taking a look at, refactored all actors in Rover to rely on the same underlying trait `ConcurrentTaskProcessingActor`. In production, all is well :) 

About to use it for image stuff next.

kifi-backend/modules/rover/app/com/keepit/rover/manager/ConcurrentTaskProcessingActor.scala
```diff
@@ -0,0 +1,136 @@
...
+
+import com.keepit.common.akka.{ UnsupportedActorMessage, FortyTwoActor }
+import com.keepit.common.healthcheck.AirbrakeNotifier
+import com.keepit.common.logging.Logging
+import com.keepit.common.concurrent.ExecutionContext
+
+import scala.concurrent.{ Future }
+import scala.util.{ Failure, Success }
+import com.keepit.common.core._
+
```
leogrim, line 11: @stkem @andrewconner @eishay @yingjieMiao DRYing up :)

PR #15002, opened 2015-04-22
mobile mutual user connections -> fix empty map entry (key -> value pair...

ahsu1230:
...)

Sorry, seems like we were getting airbrakes on this endpoint.
@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/conf/shoeboxService.routes
+++ b/kifi-backend/modules/shoebox/conf/shoeboxService.routes
@@ -608,9 +608,6 @@ GET     /admin/search/performance/viewPerformance @com.keepit.controllers.admin.
 
 GET     /admin/users/reindex        @com.keepit.controllers.admin.AdminUserIndexController.reindex()
 GET     /admin/users                @com.keepit.controllers.admin.AdminUserController.allUsersView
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/conf/shoeboxService.routes
```diff
@@ -689,6 +686,9 @@ GET     /admin/explainUserResult        @com.keepit.controllers.admin.AdminSearc
 
 GET     /admin/search/results/:id   @com.keepit.controllers.admin.AdminSearchController.articleSearchResult(id: ExternalId[ArticleSearchResult])
 
+POST    /admin/users/createPushActivityEntities     @com.keepit.controllers.admin.AdminUserController.createPushActivityEntities
+POST    /admin/users/:id/pushLibraryActivity        @com.keepit.controllers.admin.AdminUserController.pushLibraryActivity(id: Id[User])
+POST    /admin/users/:id/pushPersonaActivity        @com.keepit.controllers.admin.AdminUserController.pushPersonaActivity(id: Id[User])
```
ahsu1230, line 691: changing url so you don't need to impersonate & can instead just use `userId` in url @eishay

PR #15004, opened 2015-04-22
All social fetching has failed on Twitter

andrewconner:
@eishay @stkem I don't know how long this has failed, but it's broken now. We have received several hundred reports of this, so I assume because of the constant flood of exceptions, we all missed it.

This is the current call:

`https://api.twitter.com/1.1/followers/ids.json?user_id=TwitterId%2878157234%29&cursor=TwitterId%28-1%29&count=5000`

Notice the case class `toString` bug.


andrewconner:
Looks like it broke on April 10. Obviously this should have been tested, but I wouldn't assign too much fault here since we _were_ notified. Perhaps this illustrates the state we're in?

PR #15005, opened 2015-04-22
queue activity emails to SQS

joshm1:
the motivation of using SQS instead of the existing implementation is to prevent pending emails from getting dropped if the server goes down (or is restarted) while there is an in-memory queue of emails to process and send

@stkem should we do a fake run of this to internal emails (like we did with this email originally) before it's live? The only risk I can imagine is if there's a production-environment bug that would halt either queuing or dequeuing messages to SQS and prevent emails from being sent. I included a test that ensures the correct # of emails (messages) get queued and consumed in `ActivityFeedEmailQueueHelper` and changed the existing test to use those methods

### DIFF
--- a/kifi-backend/modules/common/test/com/keepit/common/queue/FakeQueue.scala
+++ b/kifi-backend/modules/common/test/com/keepit/common/queue/FakeQueue.scala
@@ -1,0 +1,61 @@
+package com.keepit.common.queue
+
+import java.util.UUID
+
+import com.kifi.franz.{ FakeSQSQueue, MessageId, SQSMessage }
+
+import scala.collection.mutable
+import scala.concurrent.{ ExecutionContext, Future }
+import scala.concurrent.duration.FiniteDuration
+
+class FakeQueue[T] extends FakeSQSQueue[T] {
+
+  // mutable objects are public so test cases can check the state of the fake queue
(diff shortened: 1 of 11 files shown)


eishay:
wowowowow!
Welcome back Josh!

kifi-backend/modules/common/test/com/keepit/common/queue/FakeQueue.scala
```diff
@@ -0,0 +1,61 @@
...
+
+import java.util.UUID
+
+import com.kifi.franz.{ FakeSQSQueue, MessageId, SQSMessage }
+
+import scala.collection.mutable
+import scala.concurrent.{ ExecutionContext, Future }
+import scala.concurrent.duration.FiniteDuration
+
+class FakeQueue[T] extends FakeSQSQueue[T] {
```
joshm1, line 11: this wraps the fake queue provided by franz so the queuing/consumption of messages is testable. i extracted this out of curator so I could use it for both activity and feed digest emails

stkem:
lgtm! And yes, I think we should do a test, but not with all emails (it's a lot now). Maybe we can do that together this weekend?
Let's hold off with merging it until then, since the actual email is going out tomorrow morning.

PR #15014, opened 2015-04-23
invalidating RelatedLibrariesCache on library update

eishay:
https://app.asana.com/0/20849692840229/31926040278494/f

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/model/LibraryRepo.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/model/LibraryRepo.scala
@@ -1,7 +1,7 @@
 package com.keepit.model
 
 import com.google.inject.{ ImplementedBy, Inject, Singleton, Provider }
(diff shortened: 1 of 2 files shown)


eishay:
@yingjieMiao

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryRepo.scala
```diff
@@ -122,6 +123,7 @@ class LibraryRepoImpl @Inject() (
     library.id.map { id =>
       libraryMetadataCache.remove(LibraryMetadataKey(id))
       idCache.remove(LibraryIdKey(id))
+      relatedLibrariesCache.remove(RelatedLibariesKey(id))
```
2is10, line 126: I think what we really want here is to invalidate the cache for all libraries that claim this library is related. We’d need a reverse mapping of related libraries to identify them.

eishay:
the commit was bad.

PR #15017, opened 2015-04-23
remove dead code

yingjieMiao:
@stkem 

explainFeed can be expensive. This should reduce load on cortex by quite a bit.


stkem:
:+1:

PR #15019, opened 2015-04-23
checking keep count integrity

eishay:


eishay:
https://app.asana.com/0/20849692840229/31926040278494/f

PR #15020, opened 2015-04-23
change note field when tagging from current mobile endpoints

ahsu1230:
@2is10 @andrewconner 

This PR does the following:
- `hashtagCommander` handles logic of finding hashtags in a String (like a note), appending hashtags to the end of the String, and removing hashtags from a String. For easy API, these functions can take both `Set[String]` or `Set[Hashtag]`
- Wherever `keep.note` appears in a mobile response, take out all `[#...]`. In order to do that, we'll need the keep's hashtags, so we remove the correct `[#...]`

next step: update a keep's note field from any client when tagged/untagged
one time step: backfill every keep's note field with hashtags.
Thankfully, with `hashtagCommander`, this will be a bit easier.


kifi-backend/modules/shoebox/test/com/keepit/commanders/HashtagCommanderTest.scala
```diff
@@ -0,0 +1,60 @@
...
+        commander.removeHashtagNamesFromString("I like [#turtles]", Set("turtles")) === "I like"
+        commander.removeHashtagNamesFromString("I like [#turtles]", Set("turtles", "lizards")) === "I like"
+        commander.removeHashtagNamesFromString("[#turtles] I like [#lizards]", Set("turtles", "lizards")) === "I like"
+        commander.removeHashtagNamesFromString("[#turtles] I like [#lizards]", Set("lizards")) === "[#turtles] I like"
+        commander.removeHashtagNamesFromString("[#turtles] I like [#lizards]", Set("turtles")) === "I like [#lizards]"
+        commander.removeHashtagNamesFromString("[#turtles] I like [#reptiles] [#lizards]", Set("lizards")) === "[#turtles] I like [#reptiles]"
+        commander.removeHashtagNamesFromString("[#turtles] I like [#reptiles] [#lizards]", Set("turtles", "reptiles")) === "I like [#lizards]"
+        commander.removeHashtagNamesFromString("[#a] [#b] [#c]", Set("a", "b", "c")) === ""
+        commander.removeHashtagNamesFromString("[#a] [#a] [#a]", Set("a", "b", "c")) === ""
+      }
```
ahsu1230, line 68: feedback appreciated so I know I'm doing things right.
Refer to these test cases to see how it works! I'd be happy to add more test cases.

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileKeepsController.scala
```diff
@@ -193,13 +202,23 @@ class MobileKeepsController @Inject() (
...
         val newNote = noteOpt orElse keep.note map { _.trim } filterNot { _.isEmpty }
-        tagsOpt.map { tagNames =>
+
+        val newNoteWithTags = tagsOpt.map { tagNames =>
           implicit val context = heimdalContextBuilder.withRequestInfoAndSource(request, KeepSource.mobile).build
           keepsCommander.persistHashtagsForKeep(request.userId, keep.id.get, tagNames)
-        }
+
+          // append any new tags to note, remove any tags not in tagNames
+          val origNote = newNote getOrElse ""
```
ahsu1230, line 211: this mobile endpoint takes in a `tags` field which is an array of Strings. This array are all the tags that need to persist, not what tags to add.

kifi-backend/modules/shoebox/app/com/keepit/commanders/HashtagCommander.scala
```diff
@@ -0,0 +1,91 @@
+package com.keepit.commanders
+
+import com.keepit.model.Hashtag
+
+class HashtagCommander {
+
+  // look for [
```
ahsu1230: woops, will fix up comments =)

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileLibraryController.scala
```diff
@@ -156,11 +158,26 @@ class MobileLibraryController @Inject() (
         Json.obj("error" -> error)
       case Right(keepDataList) if keepDataList.nonEmpty =>
         val completeKeepData = db.readOnlyMaster { implicit s =>
+          val allKeepExtIds = keepDataList.map(_.id).toSet
+          val keepMap = keepRepo.getByExtIds(allKeepExtIds)
+          val allKeeps = keepMap.values.flatten.toSeq
+          val keepToHashtagNamesMap = collectionRepo.getTagsByKeepIds(allKeeps.map(_.id.get).toSet)
+
           keepDataList.map { keepData =>
```
ahsu1230, line 164: This endpoint serves to give back a list of libraries and which of those libraries did you already keep a url to.
So, you get `keepData` for those libraries that have that url. Because of this new note thing, we'll have to search through the `collection` table as well. Might be a pretty big endpoint =/

2is10:
Looking now, sorry for the delay.

andrewconner:
Thanks for the great tests!

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileKeepsController.scala
```diff
@@ -175,7 +176,15 @@ class MobileKeepsController @Inject() (
             h <- idealImageHeight
           } yield ImageSize(w, h)
         } getOrElse ProcessedImageSize.Large.idealSize
-        keepDecorator.decorateKeepsIntoKeepInfos(request.userIdOpt, false, Seq(keep), idealImageSize, withKeepTime = true).imap { case Seq(keepInfo) => Ok(Json.toJson(keepInfo)) }
+        keepDecorator.decorateKeepsIntoKeepInfos(request.userIdOpt, false, Seq(keep), idealImageSize, withKeepTime = true).imap {
+          case Seq(keepInfo) =>
+            val editedNote = (keepInfo.note, keepInfo.hashtags) match {
+              case (Some(note), Some(hashtags)) =>
+                Some(hashtagCommander.removeHashtagsFromString(note, hashtags))
```
2is10: With the markdown format we decided on, there is no ambiguity between user-typed hashtags (which we escape) and kifi-inserted hashtags, so we no longer need to pass the hashtags in order to strip them out.

We do need to add code to escape `"[#"` as `"[\#"` (and probably `"[@"` too for good measure) before persisting notes. In practice, I don’t think users will type that in notes often, if ever.

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileKeepsController.scala
```diff
@@ -175,7 +176,15 @@ class MobileKeepsController @Inject() (
...
+              case (note, _) => note
+            }
+            Ok(Json.toJson(keepInfo.copy(note = editedNote))) // todo (aaron): remove editedNote once tags are gone!
```
2is10: Regarding the TODO comment, I don’t think it’s necessary, since this endpoint will never change. Instead, we’ll add a new endpoint that doesn’t strip hashtags.

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileKeepsController.scala
```diff
@@ -193,13 +202,23 @@ class MobileKeepsController @Inject() (
...
+
+        val newNoteWithTags = tagsOpt.map { tagNames =>
           implicit val context = heimdalContextBuilder.withRequestInfoAndSource(request, KeepSource.mobile).build
           keepsCommander.persistHashtagsForKeep(request.userId, keep.id.get, tagNames)
-        }
+
+          // append any new tags to note, remove any tags not in tagNames
+          val origNote = newNote getOrElse ""
+          val existingTags = hashtagCommander.findAllHashtagNames(origNote)
+          val tagsToAdd = tagNames.toSet.filterNot(t => existingTags.contains(t))
```
2is10: Will the `.toSet` here potentially change the order of the tags? Would be best to preserve order, if possible.

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileKeepsController.scala
```diff
@@ -193,13 +202,23 @@ class MobileKeepsController @Inject() (
...
+          val tagsToRemove = existingTags.filterNot(t => tagNames.contains(t))
```
2is10: `tagNames.contains` is linear time. Might pay to make a `tagNamesSet` and then use it here.

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileKeepsController.scala
ahsu1230: yea, I'll go ahead and make `appendHashtags` take in a `Seq` so it allows order

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileKeepsController.scala
ahsu1230: just spotted that too! thanks

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileKeepsController.scala
```diff
@@ -193,13 +202,23 @@ class MobileKeepsController @Inject() (
...
+          val noteWithTagsAdded = hashtagCommander.appendHashtagNamesToString(origNote, tagsToAdd)
+          hashtagCommander.removeHashtagNamesFromString(noteWithTagsAdded, tagsToRemove)
+        } orElse newNote
 
         db.readWrite { implicit s =>
-          keepRepo.save(keep.copy(title = newTitle, note = newNote))
+          if (newTitle != keep.title || newNoteWithTags != keep.note)
+            keepRepo.save(keep.copy(title = newTitle, note = newNoteWithTags))
```
2is10, line 223: It’s generally a good practice to “canonicalize” the empty string to `null` in a nullable database field. All existing keeps have a `null` note, so besides keeping the database table more uniform, doing this canonicalization here would prevent writes that just change `null` to the empty string and alter `updated_at` unnecessarily.

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileLibraryController.scala
2is10, line 164: Good news: we don’t need to look up each keep’s hashtags any more in order to remove them (see my earlier comment).

kifi-backend/modules/shoebox/app/com/keepit/model/CollectionRepo.scala
```diff
@@ -167,6 +168,20 @@ class CollectionRepoImpl @Inject() (
...
+      val query = sql"select DISTINCT kc.bookmark_id, c.name from keep_to_collection kc, collection c where kc.bookmark_id in #$idSet and c.id = kc.collection_id and c.state='#${CollectionStates.ACTIVE}' and kc.state='#${KeepToCollectionStates.ACTIVE}'"
+      query.as[(Id[Keep], String)].list.groupBy(_._1).map {
+        case (keepId, pairs) =>
+          keepId -> pairs.map { pair => Hashtag(pair._2) }.toSet
+      }
+    } else {
+      Map.empty[Id[Keep], Set[Hashtag]]
+    }
+  }
+
```
2is10: This new method may be unnecessary now.

kifi-backend/modules/shoebox/app/com/keepit/model/KeepRepo.scala
```diff
@@ -170,6 +171,15 @@ class KeepRepoImpl @Inject() (
     }
   }
 
+  def getByExtIds(extIds: Set[ExternalId[Keep]])(implicit session: RSession): Map[ExternalId[Keep], Option[Keep]] = {
+    val keepMap = (for (b <- rows if b.externalId.inSet(extIds) && b.state === KeepStates.ACTIVE) yield b).list.map { keep =>
+      (keep.externalId, keep)
+    }.toMap
+    extIds.map { extId =>
+      extId -> (keepMap.get(extId) orElse None)
+    }.toMap
```
2is10: The common case by a wide margin is a singleton set. I think it’s worth optimizing this case here to delegate to `get(ExternalId[Keep])` or `getOpt`, which use a pre-compiled query.

(The old code used `.get` so I think it may be okay to just return `Map[ExternalId[Keep], Keep]`—not `Option[Keep]`.)

kifi-backend/modules/shoebox/app/com/keepit/commanders/HashtagCommander.scala
```diff
@@ -0,0 +1,95 @@
...
+          }
+        } else {
+          idx = editedStr.length // no closing ']', there should not be any more hashtags in string
+        }
+      } else {
+        idx += 1
+      }
+    }
+    editedStr.trim
+  }
```
2is10, line 28: A regular expression would probably help here.

``` scala
val hashTagRe = """\[#((?:\\[\\\]]|[^\]])*)\] ?""".r
```

``` scala
def removeHashTags(s: String): String = hashTagRe.replaceAllIn(s, "").trim
```

ahsu1230, line 28: doh. Thanks. I'm not very good at regex - something I really need to get better at.

andrewconner, line 28: http://regex.learncodethehardway.org/book/

2is10, line 28: Ha! I love it. From [the introduction](http://regex.learncodethehardway.org/book/introduction.html):

> You cannot learn these concepts without actually doing the work and practicing them. Otherwise you might as well just save your money and continue to suck at programming.

ahsu1230, line 28: Oh something I noticed. What do you think about collapsing spaces when removing hashtags?
i.e.

```
removeHashtags("asdf [#123a] [#qwer] asdf")
res17: String = asdf   asdf
```

in which case then we could either look for 1 space or endofline after `]`
and include that into the set to replace

kifi-backend/modules/shoebox/app/com/keepit/commanders/HashtagCommander.scala
```diff
@@ -0,0 +1,95 @@
...
+
+
+  // append hashtags to the end of a string, separated by spaces
+  def appendHashtagsToString(str: String, hashtags: Set[Hashtag]): String = {
+    val hashtagNames = hashtags.map(_.tag)
+    appendHashtagNamesToString(str, hashtagNames)
+  }
+
+  def appendHashtagNamesToString(str: String, hashtagNames: Set[String]): String = {
+    val tagsStr = hashtagNames.filter(_.nonEmpty).map("[#" + _ + "]").mkString(" ")
```
2is10: be sure to backslash-escape `]` and `\` in the hashtag names

``` scala
val backslashEscapeRe = """[\]\\]""".r
```

``` scala
backslashEscapeRe.replaceAllIn(_, """\\$0""")
```

kifi-backend/modules/shoebox/app/com/keepit/commanders/HashtagCommander.scala
```diff
@@ -0,0 +1,95 @@
...
+          idx = closingIdx + 1
+        } else {
+          idx = str.length // no closing ']' in string, so no more potential hashtags
+        }
+      } else {
+        idx += 1
+      }
+    }
+    hashtags.toSet
+  }
```
2is10: This method can use the same `hashTagRe` as the `removeHashTags` method:

``` scala
val backslashUnescapeRe = """\\(.)""".r
```

``` scala
def findAllHashTagNames(s: String): Set[String] = {
  hashTagRe.findAllMatchIn(s).map(_ group 1).map { escapedName =>
    backslashUnescapeRe.replaceAllIn(escapedName, "$1")
  } toSet
}
```

kifi-backend/modules/shoebox/app/com/keepit/commanders/HashtagCommander.scala
2is10, line 28: Adding ` ?` at the end of the regular expression will take care of one optional trailing space. I’d rather not remove newlines. We’ll probably use them as paragraph separators.

2is10, line 28: Multiple consecutive whitespace characters collapse to one when HTML is rendered, so it doesn’t really matter.

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileKeepsController.scala
ahsu1230, line 223: the `filterNot` on line 219 should turn `Some("")` into a `None`

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepDecorator.scala
```diff
@@ -186,4 +186,15 @@ class KeepDecorator @Inject() (
...
+  val escapeMarkupsRe = """\[([#@])""".r
+  def escapeMarkupNotes(str: String): String = {
+    escapeMarkupsRe.replaceAllIn(str, """[\\$1""")
+  }
+
+  val unescapeMarkupsRe = """\[\\([#@])""".r
+  def unescapeMarkupNotes(str: String): String = {
+    unescapeMarkupsRe.replaceAllIn(str, """[$1""")
+  }
+
```
ahsu1230, line 201: placed this logic in KeepDecorator for now since it applies to Keep notes.
Once we start marking up library descriptions, user biographies, etc. we should place it somewhere else =)

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileKeepsController.scala
2is10, line 223: Awesome, I didn’t read carefully enough.

kifi-backend/modules/shoebox/app/com/keepit/commanders/HashtagCommander.scala
```diff
@@ -0,0 +1,38 @@
...
+
+class HashtagCommander {
+
+  // matches '[#...]'.
+  // A literal '[#' indicates the start of a hashtag
+  // A literal ']' indicates the end of the hashtag
+  // inside that (...), all ']' & '\' will be escaped to indicate they are part of the hashtag
+  val hashTagRe = """\[#((?:\\[\\\]]|[^\]])+)\]""".r
+  val backslashEscapeRe = """[\]\\]""".r
+  val backslashUnescapeRe = """\\(.)""".r
```
2is10: These should go in an `object` so they only have to be compiled once. Maybe just make `HashtagCommander` an `object`?

2is10:
Very nice. :+1: Merge away!

ahsu1230:
thanks for the reviewing!!!!!!

PR #15030, opened 2015-04-23
fix library keep count (during copy/move keeps)

yingjieMiao:
@ahsu1230 @eishay

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
@@ -1212,8 +1212,8 @@ class LibraryCommander @Inject() (
           case Right(successKeep) => goodKeeps += successKeep
         }
       }
-      if (badKeeps.size != keeps.size)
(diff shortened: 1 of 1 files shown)


eishay:
Thanks!
Does that really take care of all the edge cases?
Shouldn't we do it always at the same place we're updating the `Library.lastKept` field?

yingjieMiao:
It's consistent here. It appears not in the same place b/c I want to make sure we have cleared the cache.

ahsu1230:
the logic looks good. `LibraryRepo.updateLastKept` is also used in `KeepInterner.internRawBookmark` & `internRawBookmarksWithStatus`. Both of which are eventually called by the private function `internKeep` that updates a library's `keep_count`. 

Note: There's a comment in `KeepInterner.internKeep` that recommends us to change how we're counting keeps in a library since it's not only super expensive, but also prone to deadlocks @stkem

yingjieMiao:
The situation is better here (I hope). In keepInterner, we update libraryKeepCount for every keep we intern. Here, we update only after a batch operation is done. (and those batch operations from users are low frequency anyway)

yingjieMiao:
merging. will babysit for a while.

PR #15032, opened 2015-04-24
admin endpoint to backfill keep notes with currently tagged hashtags

ahsu1230:
@andrewconner @2is10

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/admin/AdminBookmarkController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/admin/AdminBookmarkController.scala
@@ -339,13 +339,13 @@ class AdminBookmarksController @Inject() (
         collectionRepo.getHashtagsByKeepIds(keepIds)
       }
       val keepNotesMap = keepGroup.map { k =>
-        // given a keep note, escape characters: [#...] -> [\#...] and [@...] -> [\@...]
-        val escapedNote = k.note.map(keepDecorator.escapeMarkupNotes(_))
         // look for all hashtags for a keep
         val newNote = keepHashtagsMap.get(k.id.get) match {
           case Some(tags) if tags.nonEmpty =>
-            val noteStr = escapedNote getOrElse ""
-            Some(hashtagCommander.appendHashtagsToString(noteStr, tags).trim) filterNot (_.isEmpty)
+            val noteStr = k.note getOrElse ""
+            val existingTags = hashtagCommander.findAllHashtagNames(noteStr)
+            val tagsToAppend = tags.map(_.tag) diff existingTags.toSeq
+            Some(hashtagCommander.appendHashtagNamesToString(noteStr, tagsToAppend).trim) filterNot (_.isEmpty)
           case _ =>
             k.note
         }
--- a/kifi-backend/modules/shoebox/app/com/keepit/admin/AdminBookmarkController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/admin/AdminBookmarkController.scala
@@ -1,7 +1,7 @@
 package com.keepit.controllers.admin
 
 import com.google.inject.Inject
-import com.keepit.commanders.{ CollectionCommander, KeepsCommander, LibraryCommander, RichWhoKeptMyKeeps, URISummaryCommander }
+import com.keepit.commanders._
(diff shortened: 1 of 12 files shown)


kifi-backend/modules/shoebox/test/com/keepit/controllers/admin/AdminPersonaControllerTest.scala
```diff
@@ -156,6 +159,42 @@ class AdminPersonaControllerTest extends Specification with ShoeboxApplicationIn
         }
       }
     }
+
+    "temporary keep note backfill" in {
```
ahsu1230, line 163: sorry it's a little lazy, but I put this `AdminBookmarksController` function inside a `AdminPersonaControllerTest` so I don't have to create a new test class & then later remove the one test it will have (since it's a temp endpoint) =P

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminBookmarkController.scala
```diff
@@ -326,4 +328,37 @@ class AdminBookmarksController @Inject() (
     Ok
   }
 
+  def populateKeepNotesWithTag(page: Int = 0, size: Int = 500, grouping: Int = 500) = AdminUserAction {
```
ahsu1230, line 331: paginated so we can control how we're doing

kifi-backend/modules/shoebox/test/com/keepit/controllers/admin/AdminPersonaControllerTest.scala
andrewconner, line 163: Good with me

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminBookmarkController.scala
```diff
@@ -326,4 +328,37 @@ class AdminBookmarksController @Inject() (
...
+    db.readOnlyMaster { implicit s =>
+      keepRepo.page(page, size, Set.empty)
+    }.grouped(grouping).foreach { keepGroup =>
+      val keepIds = keepGroup.map(_.id.get).toSet
+      // look for hashtags for every keep
+      val keepHashtagsMap = db.readOnlyMaster { implicit s =>
+        collectionRepo.getHashtagsByKeepIds(keepIds)
+      }
+      val keepNotesMap = keepGroup.map { k =>
+        // given a keep note, escape characters: [#...] -> [\#...] and [@...] -> [\@...]
```
2is10: We should remove all of the `...]` from this comment, since that part is inaccurate.
Actually, I’d recommend not doing this escaping here. The database currently contains 7427 non-null notes, and none of them contain `[#` or `[@`. Escaping here may be more likely to cause problems than avoid them (e.g. if this code runs twice for the same keep).

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminBookmarkController.scala
```diff
@@ -326,4 +328,37 @@ class AdminBookmarksController @Inject() (
...
+        val escapedNote = k.note.map(keepDecorator.escapeMarkupNotes(_))
+        // look for all hashtags for a keep
+        val newNote = keepHashtagsMap.get(k.id.get) match {
+          case Some(tags) if tags.nonEmpty =>
+            val noteStr = escapedNote getOrElse ""
+            Some(hashtagCommander.appendHashtagsToString(noteStr, tags).trim) filterNot (_.isEmpty)
```
2is10: How about checking whether the tags are already in the note before appending them? That would make this endpoint idempotent, yes? Data migrations are safer when idempotent.

2is10:
LGTM

2is10:
By the way, we should have all code in place that will maintain tags in the notes field—when a tag is added or removed using any endpoint—before doing the migration. Otherwise, we’d have to redo the migration for some keeps later.

ahsu1230:
Yup, I'll definitely change those client endpoints to update notes on tagging/untagging. I thought this endpoint would be a bit easier to do, so I decided to go ahead and get it out of the way for now =)

ahsu1230:
will call this admin endpoint once the client endpoints are correct

PR #15034, opened 2015-04-24
Killing this

andrewconner:
@eishay I believe this notification is crashing my app every time. We probably didn't notice it because Twitter auto-follow has been broken.

Revert if you disagree. I'm going to go ahead and try to delete all existing ones.


andrewconner:
Actually this doesn't fix it. Please hold...

eishay:
ok

eishay:
should check tomorrow

PR #15044, opened 2015-04-25
marking social user info as expired when needed

eishay:
DONT MERGE YET!!!


eishay:
@andrewconner 
when merging and deploying we need to do extensive tests (login/out) immediately.

PR #15050, opened 2015-04-25
we have lots of timeouts or long wait on these calls and it seems like w...

eishay:
...e can be more relaxed there

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/curator/CuratorServiceClient.scala
+++ b/kifi-backend/modules/common/app/com/keepit/curator/CuratorServiceClient.scala
@@ -35,7 +35,7 @@ class CuratorServiceClientImpl(
     implicit val defaultContext: ExecutionContext,
     val airbrakeNotifier: AirbrakeNotifier) extends CuratorServiceClient {
 
-  val longTimeout = CallTimeouts(responseTimeout = Some(30000), maxWaitTime = Some(3000), maxJsonParseTime = Some(10000))
+  val longTimeout = CallTimeouts(responseTimeout = Some(30000), maxWaitTime = Some(10000), maxJsonParseTime = Some(10000))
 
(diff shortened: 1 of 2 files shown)


eishay:
@stkem @leogrim

kifi-backend/modules/common/app/com/keepit/curator/CuratorServiceClient.scala
```diff
@@ -119,7 +119,7 @@ class CuratorServiceClientImpl(
 
   def ingestPersonaRecos(userId: Id[User], personaIds: Seq[Id[Persona]], reverseIngestion: Boolean = false): Future[Unit] = {
     val payload = Json.obj("personaIds" -> personaIds)
-    call(Curator.internal.ingestPersonaRecos(userId, reverseIngestion), payload).map { _ => Unit }
+    call(Curator.internal.ingestPersonaRecos(userId, reverseIngestion), payload, callTimeouts = longTimeout).map { _ => Unit }
```
yingjieMiao, line 122: this is typically fast, i.e. under 1 second. Increase this means client side (e.g. mobile) `freezes` during sign up.  I see the timeouts are more frequent after we 'scaled out' curator. Likely to be a contention problem.

eishay, line 122: so you thing we should revert this one?

yingjieMiao, line 122: I think this is better for user experience. Let user waiting for 30 seconds is not fun. 

We used to have an async fashion, i.e. shoebox tells curator to inject and forget about it (without curator confirm it's done). The problem was client will immediately finish the sign up process and call curator for recos. Sometimes ingestion hasn't even started by then, which leaves a blank page.

eishay:
ok, will revert that one

PR #15054, opened 2015-04-26
Since I'm not an expert there and it seems like the error is fixing itse...

eishay:
...lf I'm converting airbrake to log until we can get Aaron figure it out. No point in sending 700 errors in a blast.


eishay:
from the email:

Bookmark 3270988 has inconsistent visibility with library 502815. Expected: PUBLISHED Actual: SECRET com.keepit.common.healthcheck.DefaultAirbrakeException: null 
  Cause: [No Cause]

Total Errors since start: 706
Total Errors since last alert: 705
Server: shoebox-spot-1
Server started at: 2015-04-25T19:56:45.401Z
Full Details

2015-04-26T00:49:49.724Z: [2a51bde9-5a0c-4565-a279-e891e61e53c0]
error message [Bookmark 3270988 has inconsistent visibility with library 502815. Expected: PUBLISHED Actual: SECRET]
Exception stack trace: 
Cause:

eishay:
@ahsu1230 @andrewconner please figure it out

PR #15059, opened 2015-04-27
better error reporting on twitter issues

eishay:


eishay:
it seems like @ahsu1230 is actually on the actual cause of these exceptions (good job!), but this will help with future investigations

ahsu1230:
:+1:

PR #15060, opened 2015-04-27
data integrity runs 10 minutes to align with library checker

ahsu1230:
@eishay @yingjieMiao

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/integrity/LibraryChecker.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/integrity/LibraryChecker.scala
@@ -138,8 +138,8 @@ class LibraryChecker @Inject() (
     val currentTime = clock.now()
     val hourInd = currentTime.hourOfDay().get
     val minuteInd = currentTime.minuteOfHour().get
-    val index = hourInd * 4 + (minuteInd / 15) // hour * 4 + minute / 15 (int div)
-    val numIntervals = 96 // 96 = 24 (hours per day) * 4 (15 minute intervals per hour)
+    val index = hourInd * 6 + (minuteInd / 10) // hour * 6 + minute / 10 (int div)
+    val numIntervals = 144 // 144 = 24 (hours per day) * 6 (10 minute intervals per hour)
     (index, numIntervals)
   }
 
--- a/kifi-backend/modules/shoebox/app/com/keepit/integrity/LibraryChecker.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/integrity/LibraryChecker.scala
@@ -6,8 +6,9 @@ import com.keepit.common.db.Id
 import com.keepit.common.db.slick.Database
 import com.keepit.common.healthcheck.AirbrakeNotifier
 import com.keepit.common.logging.{ NamedStatsdTimer, Logging }
+import com.keepit.common.time.Clock
 import com.keepit.model._
-import org.joda.time.{ Minutes, DateTime }
+import com.keepit.common.time._
 
 class LibraryChecker @Inject() (
     db: Database,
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/app/com/keepit/integrity/LibraryChecker.scala
```diff
@@ -133,11 +135,11 @@ class LibraryChecker @Inject() (
...
   private def getIndex(): (Int, Int) = {
-    val currentTime = DateTime.now()
+    val currentTime = clock.now()
     val hourInd = currentTime.hourOfDay().get
-    val minuteInd = currentTime.minuteOfHour().get / 10 // 10 times per hour
-    val index = hourInd * 6 + minuteInd // hour * 6 + minute / 10 (int div)
-    val numIntervals = 144 // 144 = 24 (hours per day) * 6 (10 minute intervals per hour)
+    val minuteInd = currentTime.minuteOfHour().get
+    val index = hourInd * 4 + (minuteInd / 15) // hour * 4 + minute / 15 (int div)
+    val numIntervals = 96 // 96 = 24 (hours per day) * 4 (15 minute intervals per hour)
```
ahsu1230: DataChecker has a frequency interval of 13 minutes.
I figured if we have intervals of 15 instead of 10, while there may be overlap between library checking iterations, at least we'll hit all the keeps. A concern is going from 144 intervals to 96 could be a lot, so every library check is bigger.

kifi-backend/modules/shoebox/app/com/keepit/integrity/LibraryChecker.scala
```diff
@@ -16,7 +17,8 @@ class LibraryChecker @Inject() (
     libraryMembershipRepo: LibraryMembershipRepo,
     keepRepo: KeepRepo,
     libraryCommander: LibraryCommander,
-    airbrake: AirbrakeNotifier) extends Logging {
+    airbrake: AirbrakeNotifier,
+    clock: Clock) extends Logging {
```
eishay, line 21: :+1:

kifi-backend/modules/shoebox/app/com/keepit/integrity/LibraryChecker.scala
```diff
@@ -32,7 +34,7 @@ class LibraryChecker @Inject() (
     val timer = new NamedStatsdTimer("LibraryChecker.checkSystemLibraries")
     val (index, numIntervals) = getIndex()
 
-    db.readOnlyMaster { implicit s =>
+    db.readOnlyReplica { implicit s =>
```
eishay, line 37: :+1:

kifi-backend/modules/shoebox/app/com/keepit/integrity/LibraryChecker.scala
eishay: @yingjieMiao can you check it out? I'm sure you'll be a better reviewer of the getIndex then me.

yingjieMiao: I am Ok with that. But note that here we could have 2 consecutive checks which are identical (assuming libraries number do not change). i.e. If the minutes of a day are 0 to 1440 -1, the first 2 checks happen at minute 0 and 13, which is exactly the same page of library table.

A simple check shows that with current config, the plugin will run at 111 checkpoints in the day, and only 96 of these are unique. Note that the overlap/duplicate is not random, i.e. it always happen for certain pages.

I think it's Ok to change the plugin's frequency to 10 minutes. and stick with original 144 checks. This makes reasoning easier.

ahsu1230: Ok, I'm unsure why the plugin's frequency is 13 minutes. Hopefully changing it to 10 doesn't affect the other data integrity checkers

ahsu1230: 13's just a bit of an odd number. no pun intended :octopus:

yingjieMiao:
Minor suggestion: how about factor out the `10` minutes as a constant that can be reused in both classes? Otherwise, the library checker can get broken.

PR #15073, opened 2015-04-27
keepCommander fixes keep note when keep is tagged by client endpoints

ahsu1230:
@2is10 @andrewconner 

A couple changes around `KeepCommander` to fix up a keep's notes when tags are involved.
Currently, tagging by clients is used by the following:

`LibraryController.tagKeep`  calls `KeepCommander.addToCollection()`
`LibraryController.untagKeep` -> `KeepCommander.removeFromCollection()`
`KeepsController.tagKeeps` -> `KeepCommander.addToCollection()`

`ExtLibraryController.tagKeep`-> `KeepCommander.addToCollection()`
`ExtLibraryController.untagKeep` -> `KeepCommander.removeFromCollection()`

`MobileLibraryController.keepToLibrary` -> `KeepCommander.keepWithSelectedTags()`
`MobileKeepsController.editKeepInfo` -> `KeepCommander.persistHashtagsForKeep()`

so the places to fix are in KeepCommander are:
- `addToCollection()`
- `removeFromCollection()`
- `keepWithSelectedTags()`
- `persistHashtagsForKeep()`

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
@@ -503,9 +503,9 @@ class KeepsCommander @Inject() (
         val targetKeep = keepsById(ktc.keepId)
         val editedNote = targetKeep.note.map { noteStr =>
           hashtagCommander.removeHashtagsFromString(noteStr, Set(collection.name))
-        } filterNot (_.isEmpty)
-        keepRepo.save(keepsById(ktc.keepId).copy(note = editedNote)) // notify keep index
-        libraryAnalytics.untaggedPage(collection, keepsById(ktc.keepId), context, removedAt)
+        }.filterNot(_.isEmpty)
+        val updatedKeep = keepRepo.save(targetKeep.copy(note = editedNote)) // notify keep index
+        libraryAnalytics.untaggedPage(collection, updatedKeep, context, removedAt)
       }
       removed.toSet
     } tap { _ =>
@@ -610,15 +610,16 @@ class KeepsCommander @Inject() (
   // Changes a keep's notes based on the hashtags to persist!
   def persistHashtagsForKeep(userId: Id[User], keep: Keep, selectedTagNames: Seq[String])(implicit session: RWSession, context: HeimdalContext) = {
 
-    val selectedCollections = selectedTagNames.map { getOrCreateTag(userId, _) }
-    val selectedTagIds = selectedCollections.map(_.id.get).toSet
+    // get all tags from HT list
+    val selectedTags = selectedTagNames.map { getOrCreateTag(userId, _) }
+    val selectedTagIds = selectedTags.map(_.id.get).toSet
+    // get all active tags for keep
     val activeTagIds = keepToCollectionRepo.getCollectionsForKeep(keep.id.get).toSet
 
     val tagIdsToAdd = selectedTagIds.filterNot(activeTagIds.contains(_))
     val tagIdsToRemove = activeTagIds.filterNot(selectedTagIds.contains(_))
-    val hashtagsToRemove = selectedCollections.filter(c => tagIdsToRemove.contains(c.id.get)).map(_.name.tag)
-    val hashtagsToPersist = selectedCollections.filter(c => !tagIdsToRemove.contains(c.id.get) || activeTagIds.contains(c.id.get)).map(_.name.tag)
 
+    // fix k2c for tagsToAdd & tagsToRemove
     tagIdsToAdd.map { tagId =>
       keepToCollectionRepo.getOpt(keep.id.get, tagId) match {
         case None => keepToCollectionRepo.save(KeepToCollection(keepId = keep.id.get, collectionId = tagId))
@@ -631,12 +632,23 @@ class KeepsCommander @Inject() (
       collectionRepo.collectionChanged(tagId, false, inactivateIfEmpty = true)
     }
 
-    val origNote = keep.note getOrElse ""
-    val noteWithHashtagsRemoved = hashtagCommander.removeHashtagNamesFromString(origNote, hashtagsToRemove.toSet)
-    val editedNote = keepDecorator.escapeMarkupNotes(noteWithHashtagsRemoved)
-    val noteWithHashtagsAppended = hashtagCommander.appendHashtagNamesToString(editedNote, hashtagsToPersist)
+    // go through note and find all HT
+    val keepNote = keep.note.getOrElse("")
+    val hashtagsInNote = hashtagCommander.findAllHashtagNames(keepNote)
+    val hashtagsToPersistSet = selectedTagNames.toSet
+
+    // find tags to remove & to append
+    val hashtagsToRemove = hashtagsInNote.filterNot(hashtagsToPersistSet.contains(_))
+    val hashtagsToAppend = selectedTagNames.filterNot(hashtagsInNote.contains(_))
+    val noteWithHashtagsRemoved = hashtagCommander.removeHashtagNamesFromString(keepNote, hashtagsToRemove.toSet)
+    val markedNote = keepDecorator.escapeMarkupNotes(noteWithHashtagsRemoved)
+    val noteWithHashtagsAppended = hashtagCommander.appendHashtagNamesToString(markedNote, hashtagsToAppend)
     val finalNote = Some(noteWithHashtagsAppended.trim).filterNot(_.isEmpty)
+
+    // save keep
     keepRepo.save(keep.copy(note = finalNote)) // notify keep index
+    // todo: update search index
+    // todo: report to analytics
   }
 
   private def postSingleKeepReporting(keep: Keep, isNewKeep: Boolean, library: Library, socialShare: SocialShare): Unit = SafeFuture {
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
@@ -467,16 +467,13 @@ class KeepsCommander @Inject() (
           val taggingAt = currentDateTime
           tagged.foreach { ktc =>
             val targetKeep = keepsById(ktc.keepId)
-            val editedNote = targetKeep.note.map { noteStr =>
-              val existingTags = hashtagCommander.findAllHashtagNames(noteStr)
-              if (existingTags.contains(collection.name.tag)) {
-                noteStr
-              } else {
-                hashtagCommander.appendHashtagsToString(noteStr, Seq(collection.name))
-              }
-            } filterNot (_.isEmpty)
-            keepRepo.save(targetKeep.copy(note = editedNote)) // notify keep index
-            libraryAnalytics.taggedPage(updatedCollection, keepsById(ktc.keepId), context, taggingAt)
+            val noteStr = targetKeep.note.getOrElse("")
+            val existingTags = hashtagCommander.findAllHashtagNames(noteStr)
+            if (!existingTags.contains(collection.name.tag)) {
+              val editedNote = Some(hashtagCommander.appendHashtagsToString(noteStr, Seq(collection.name))).filterNot(_.isEmpty)
+              val updatedKeep = keepRepo.save(targetKeep.copy(note = editedNote)) // notify keep index
+              libraryAnalytics.taggedPage(updatedCollection, updatedKeep, context, taggingAt)
+            }
           }
           tagged
         }
@@ -602,7 +599,7 @@ class KeepsCommander @Inject() (
       case Failure(e) => Left(e.getMessage)
       case Success((keep, isNewKeep)) =>
         val tags = db.readWrite { implicit s =>
-          persistHashtags(userId, keep, selectedTagNames)(s, context)
+          persistHashtagsForKeep(userId, keep, selectedTagNames)(s, context)
           keepToCollectionRepo.getCollectionsForKeep(keep.id.get).map { id => collectionRepo.get(id) }
         }
         postSingleKeepReporting(keep, isNewKeep, library, socialShare)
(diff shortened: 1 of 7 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
```diff
@@ -464,7 +466,16 @@ class KeepsCommander @Inject() (
...
+            val editedNote = targetKeep.note.map { noteStr =>
+              val existingTags = hashtagCommander.findAllHashtagNames(noteStr)
+              if (existingTags.contains(collection.name.tag)) {
+                noteStr
+              } else {
+                hashtagCommander.appendHashtagsToString(noteStr, Seq(collection.name))
+              }
+            }
+            keepRepo.save(targetKeep.copy(note = editedNote)) // notify keep index
             libraryAnalytics.taggedPage(updatedCollection, keepsById(ktc.keepId), context, taggingAt)
```
2is10: This line could use the result of the `keepRepo.save` from the previous line instead of looking up the old keep in `keepsById`, yes?

Also, these two lines could go in the `else` block above, no?

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
```diff
@@ -596,28 +610,45 @@ class KeepsCommander @Inject() (
...
-    tagsToRemove.map { tagId =>
-      keepToCollectionRepo.remove(keepId, tagId)
+    val collectionsRemoved = tagsToRemove.map { tagId =>
+      keepToCollectionRepo.remove(keep.id.get, tagId)
       collectionRepo.collectionChanged(tagId, false, inactivateIfEmpty = true)
     }
+
+    // first turn all '[#' -> '[\#'
+    val origNote = keep.note getOrElse ""
+    val editedNote = keepDecorator.escapeMarkupNotes(origNote)
```
2is10: [revised] This line will destroy any existing hashtags. I think it would be clearer and safer to only use it in specific endpoints (at the controller level) instead of using it here.

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
```diff
@@ -596,28 +610,45 @@ class KeepsCommander @Inject() (
...
+
+    // remove all hashtags. Anything user typed in with [#...] has already been converted to [\#...]
+    val noteWithHashtagsRemoved = hashtagCommander.removeHashtagsFromString(editedNote, collectionsRemoved.map(_.name).toSet)
```
2is10: This `.removeHashtagsFromString` call will never have any effect, because the previous operation (`.escapeMarkupNotes`) destroys all hashtags.

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileKeepsController.scala
```diff
@@ -200,28 +200,22 @@ class MobileKeepsController @Inject() (
...
+            collectionRepo.getHashtagsByKeepId(keep.id.get).map(_.tag).toSeq
+          }
+        )
 
-          // append all the correct hashtags (in order)
-          hashtagCommander.appendHashtagNamesToString(noteWithHashtagsRemoved, tagNames).trim
-        } orElse newNote filterNot { _.isEmpty }
+        val updatedKeep = if (titleToPersist != keep.title || noteToPersist != keep.note) {
+          db.readWrite { implicit s =>
+            keepRepo.save(keep.copy(title = titleToPersist, note = noteToPersist))
```
2is10: `noteToPersist` doesn’t contain any hashtags (correct?), so I don’t think it should be persisted here due to the possibility of losing user data. Instead it could be passed in the keep object given to `.persistHashtagsForKeep`.

Also, may want to call `keepsCommander.persistHashtags` directly (make it public) and reuse this `readWrite` session for it.

kifi-backend/modules/shoebox/test/com/keepit/controllers/mobile/MobileKeepsControllerTest.scala
```diff
@@ -909,6 +909,26 @@ class MobileKeepsControllerTest extends Specification with ShoeboxTestInjector w
...
+        status(testEditTags3) must equalTo(NO_CONTENT)
+        db.readOnlyMaster { implicit s =>
+          val currentKeep = keepRepo.get(keep.externalId)
+          currentKeep.title === Some("a real keep")
+          currentKeep.note === Some("a real [\\#note] thing [#a] [#b] [#d] [#e]")
+          keepToCollectionRepo.count === 5
+          collectionRepo.count(user.id.get) === 5
+        }
+
+        val testEditTags4 = editKeepInfo(user, keep, Json.obj("note" -> "a real [#note] thing", "tags" -> Json.toJson(Seq(): Seq[String])))
```
2is10: I think a test case that starts with a keep that already has tags persisted would also be helpful.

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
```diff
@@ -464,8 +466,14 @@ class KeepsCommander @Inject() (
           val tagged: Set[KeepToCollection] = (activated ++ newK2C).toSet
           val taggingAt = currentDateTime
           tagged.foreach { ktc =>
-            keepRepo.save(keepsById(ktc.keepId)) // notify keep index
-            libraryAnalytics.taggedPage(updatedCollection, keepsById(ktc.keepId), context, taggingAt)
+            val targetKeep = keepsById(ktc.keepId)
+            val noteStr = targetKeep.note.getOrElse("")
+            val existingTags = hashtagCommander.findAllHashtagNames(noteStr)
+            if (!existingTags.contains(collection.name.tag)) {
```
2is10: Should this be a case-insensitive check? I think we’ve historically ignored case differences in tags.

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
```diff
@@ -596,29 +607,44 @@ class KeepsCommander @Inject() (
...
+    // find hashtags to remove & to append
+    val hashtagsToRemove = hashtagsInNote.filterNot(hashtagsToPersistSet.contains(_))
+    val hashtagsToAppend = selectedTagNames.filterNot(hashtagsInNote.contains(_))
+    val noteWithHashtagsRemoved = hashtagCommander.removeHashtagNamesFromString(keepNote, hashtagsToRemove.toSet)
+    val noteWithHashtagsAppended = hashtagCommander.appendHashtagNamesToString(noteWithHashtagsRemoved, hashtagsToAppend)
+    val finalNote = Some(noteWithHashtagsAppended.trim).filterNot(_.isEmpty)
+
+    keepRepo.save(keep.copy(note = finalNote))
+    // todo: update search index
+    // todo: report to analytics
```
2is10: How about putting these 3 lines in an `if` block?

``` scala
if (finalNote != keep.note) {

}
```

Regarding the new TODO comments—are you sure those aren’t already being done somehow? If not, why did the old line 591 have the comment `// notify keep index` and why does this method take an implicit `HeimdalContext`? I’m sure you have a better idea of how these things are done than I do. I’m just asking to make sure these comments won’t be possibly misleading.

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileKeepsController.scala
```diff
@@ -200,27 +200,21 @@ class MobileKeepsController @Inject() (
         val noteOpt = (json \ "note").asOpt[String]
         val tagsOpt = (json \ "tags").asOpt[Seq[String]]
 
-        val newTitle = titleOpt orElse keep.title map { _.trim } filterNot { _.isEmpty }
-        val newNote = noteOpt orElse keep.note map { _.trim } filterNot { _.isEmpty }
+        val titleToPersist = (titleOpt orElse keep.title) map (_.trim) filterNot (_.isEmpty)
+        val noteToPersist = (noteOpt orElse keep.note).map { note =>
+          keepDecorator.escapeMarkupNotes(note).trim
+        }.filter(_.nonEmpty)
```
2is10: We don’t want to escape `keep.note`, since that would already have been escaped. Something like:

``` scala
val noteToPersist = noteOpt map { note =>
  keepDecorator.escapeMarkupNotes(note).trim
} orElse keep.note filter (_.nonEmpty)
```

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
ahsu1230: I'll go ahead and add some logic into `hashtagCommander` to append hashtags (case-insensitive check) that do not exist in a note string. It seems to be used often, so it will be helpful

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileKeepsController.scala
```diff
@@ -200,27 +200,21 @@ class MobileKeepsController @Inject() (
...
+          }
+        )
+        val updatedKeep = keep.copy(title = titleToPersist, note = noteToPersist)
 
+        implicit val context = heimdalContextBuilder.withRequestInfoAndSource(request, KeepSource.mobile).build
         db.readWrite { implicit s =>
-          if (newTitle != keep.title || newNoteWithTags != keep.note)
-            keepRepo.save(keep.copy(title = newTitle, note = newNoteWithTags))
+          keepsCommander.persistHashtagsForKeep(request.userId, updatedKeep, tagsToPersist)(s, context)
         }
```
2is10, line 218: Can this whole `db.readWrite` block be skipped if `noteOpt` and `tagsOpt` are both `None`?
Nevermind. I guess not, since the updated title is no longer persisted earlier.

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
ahsu1230: `HeimdalContext` is being used by some of the tag functions like `getOrCreateTags` at the very least (mainly for when a tag is created). 
And you're right, it seems search index and many other things are updated by another function `keepWithSelectedTags` which also calls this function. 

Maybe I could have a wrapper function that calls this function `persistHashtagsForKeep`, then update searchIndex

kifi-backend/modules/shoebox/test/com/keepit/controllers/mobile/MobileKeepsControllerTest.scala
```diff
@@ -906,9 +909,42 @@ class MobileKeepsControllerTest extends Specification with ShoeboxTestInjector w
...
+
+        val testEditTags3 = editKeepInfo(user, keep, Json.obj("note" -> "a real [#note] thing"))
+        status(testEditTags3) must equalTo(NO_CONTENT)
+        db.readOnlyMaster { implicit s =>
+          val currentKeep = keepRepo.get(keep.externalId)
+          currentKeep.title === Some("a real keep")
+          currentKeep.note === Some("a real [\\#note] thing [#a] [#b] [#d] [#e]")
+        }
+
+        val testEditTags4 = editKeepInfo(user, keep, Json.obj("note" -> "a real [#note] thing", "tags" -> Json.toJson(Seq(): Seq[String])))
```
2is10: If you like, you can replace:
`Json.toJson(Seq(): Seq[String])`
with:
`Seq.empty[String]`

kifi-backend/modules/shoebox/test/com/keepit/controllers/mobile/MobileKeepsControllerTest.scala
```diff
@@ -906,9 +909,42 @@ class MobileKeepsControllerTest extends Specification with ShoeboxTestInjector w
...
+
+        // test a keep that already has tags persisted
+        val testEditExistingTags1 = editKeepInfo(user, oldKeep, Json.obj("note" -> "this keep already has [#tags]"))
+        status(testEditExistingTags1) must equalTo(NO_CONTENT)
+        db.readOnlyMaster { implicit s =>
+          val currentKeep = keepRepo.get(oldKeep.externalId)
+          currentKeep.title === Some("default1")
+          currentKeep.note === Some("this keep already has [\\#tags] [#tag1] [#tag2]")
+        }
+        val testEditExistingTags2 = editKeepInfo(user, oldKeep, Json.obj("note" -> "this keep already has [#tags]", "tags" -> JsArray(Seq(JsString("tag1")))))
```
2is10: Similarly,
`JsArray(Seq(JsString("tag1")))`
is shorter as:
`Seq("tag1")`

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileKeepsController.scala
ahsu1230, line 218: would it be weird stylistically to have something like this:

```
val updatedKeep = keep.copy(title = titleToPersist, note = noteToPersist)
...
db.readWrite { implicit s =>
   if (noteOpt.isEmpty && tagsOpt.isEmpty) {
          keepsCommander.persistHashtagsForKeep(request.userId, updatedKeep, tagsToPersist)(s, context) // persists updated keep to database
   } else {
      keepRepo.save(updatedKeep)
   }
}
```

kifi-backend/modules/shoebox/test/com/keepit/controllers/mobile/MobileKeepsControllerTest.scala
```diff
@@ -832,19 +832,22 @@ class MobileKeepsControllerTest extends Specification with ShoeboxTestInjector w
...
+          val tag2 = collectionRepo.save(Collection(userId = user.id.get, name = Hashtag("tag2")))
+          keepToCollectionRepo.save(KeepToCollection(keepId = oldKeep.id.get, collectionId = tag1.id.get))
+          keepToCollectionRepo.save(KeepToCollection(keepId = oldKeep.id.get, collectionId = tag2.id.get))
 
-          keepToCollectionRepo.count === 0
+          collectionRepo.count(user.id.get) === 2
+          keepToCollectionRepo.count === 2
 
-          (user, keep, keepInactive)
+          (user, keep, oldKeep, keepInactive)
```
2is10: The name `oldKeep` doesn’t really help me understand how it’s different from `keep`. Both are pre-existing keeps. The difference seems to be that the second one also has tags. Calling them `keep1` and `keep2` might be simpler/clearer. Or `keep` and `keepWithTags`.

2is10:
Good test cases. Nice work. I’m happy with it when you are.

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileKeepsController.scala
2is10, line 218: This codepath isn’t particularly important to optimize. I’m fine with it as is.

ahsu1230:
Thanks for the reviews!! @2is10

PR #15074, opened 2015-04-27
fetch twitter page not found!

ahsu1230:
@eishay @andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/common/social/TwitterSocialGraph.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/common/social/TwitterSocialGraph.scala
@@ -443,9 +443,16 @@ class TwitterSocialGraphImpl @Inject() (
               }
             }
             Seq.empty
-          } else if (response.status == 404) { // {"errors":[{"code":34,"message":"Sorry, that page does not exist."}]} ->
-            //something in twitter is bad or our api is broken, happen a lots recentrly, in check
+          } else if (response.status == 404) {
             log.warn(s"Failed to get users $handle timeline, status ${response.status}, msg: ${response.json.toString}, social user info $socialUserInfoOpt , signature $sig", stackTrace)
+            val errorCode = (response.json \\ "code").map(_.as[Int])
+            if (errorCode.contains(34)) { // {"errors":[{"code":34, "message":"Sorry, that page does not exist."}]}
+              socialUserInfoOpt.map { sui =>
+                db.readWrite { implicit s =>
+                  socialUserInfoRepo.save(sui.copy(state = SocialUserInfoStates.FETCH_PAGE_NOT_FOUND))
+                }
+              }
+            }
             Seq.empty
           } else {
             airbrake.notify(s"Failed to get [$endpoint] users $handle timeline, status ${response.status}, msg: ${response.json.toString}, social user info $socialUserInfoOpt , signature $sig", stackTrace)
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/shoebox/app/com/keepit/common/social/TwitterSocialGraph.scala
```diff
@@ -443,9 +443,16 @@ class TwitterSocialGraphImpl @Inject() (
...
             Seq.empty
-          } else if (response.status == 404) { // {"errors":[{"code":34,"message":"Sorry, that page does not exist."}]} ->
-            //something in twitter is bad or our api is broken, happen a lots recentrly, in check
+          } else if (response.status == 404) {
             log.warn(s"Failed to get users $handle timeline, status ${response.status}, msg: ${response.json.toString}, social user info $socialUserInfoOpt , signature $sig", stackTrace)
+            val errorCode = (response.json \\ "code").map(_.as[Int])
+            if (errorCode.contains(34)) { // {"errors":[{"code":34, "message":"Sorry, that page does not exist."}]}
+              socialUserInfoOpt.map { sui =>
+                db.readWrite { implicit s =>
+                  socialUserInfoRepo.save(sui.copy(state = SocialUserInfoStates.FETCH_PAGE_NOT_FOUND))
```
ahsu1230, line 452: seems like some twitter page is not found for whatever reason. Set the sui state to `FETCH_PAGE_NOT_FOUND`. But, I'm not sure if this will actually stop trying to fetch for this socialuser. I'm not sure where that part could be...

kifi-backend/modules/shoebox/app/com/keepit/integrity/LibraryChecker.scala
```diff
@@ -138,8 +138,11 @@ class LibraryChecker @Inject() (
     val currentTime = clock.now()
     val hourInd = currentTime.hourOfDay().get
     val minuteInd = currentTime.minuteOfHour().get
-    val index = hourInd * 6 + (minuteInd / 10) // hour * 6 + minute / 10 (int div)
-    val numIntervals = 144 // 144 = 24 (hours per day) * 6 (10 minute intervals per hour)
+
+    val frequency = DataIntegrityScheduling.frequency
+    val freqPerHour = 60 / frequency // assuming frequency is 10
+    val index = hourInd * freqPerHour + (minuteInd / frequency) // hour * 6 + minute / 10 (int div)
+    val numIntervals = 24 * freqPerHour // 144 = 24 (hours per day) * 6 (10 minute intervals per hour)
```
ahsu1230, line 145: @yingjieMiao you're right about the minor refactoring for `DataIntegrity` constants. thanks for the reminder!

eishay:
lgtm

andrewconner:
Don't merge, in a meeting.

andrewconner:
This doesn't fix it and will mess up the SUI record

eishay:
interesting, why?

andrewconner:
This error happens when we try to fetch the tweets from `handle` using the token in `socialUserInfoOpt` (`None` means use the system token). If twitter says the `handle` is bad, it doesn't say much about the validity of `socialUserInfo` — in fact, the sui itself is likely _known_ to be good / valid. An invalid handle just means Twitter sync is broken.

A 404 here should potentially update `twitter_sync_state`, but probably not here (separation of concerns). We could potentially salvage a handle change, but it's heuristic.

PR #15076, opened 2015-04-28
Falling back on upserts for persona recos to avoid integrity constraint exception

stkem:
@yingjieMiao


stkem:
@yingjieMiao

yingjieMiao:
lgtm. Some notes:

on shoebox side, we first fire a request to curator, hoping it to finish ingestion soon. However, sometimes the call time-out, and shoebox fires a new request (fire-and-forget). SQL problem happens if the old thread and new thread are concurrent. 

From what I noticed so far: 1) even when exception happens, the user's persona recos are there 2) usually the timeout exception and SQL violation happens one after the other, which suggest where this duplication come from.

stkem:
Yup, it indeed looks like duplicate request from the client in some cases. Not a big deal and at least this will prevent exceptions (without suppressing them)

PR #15077, opened 2015-04-28
Don't get stuck if a batch is completely empty after filtering

stkem:
I recently broke this. Thankfully it is pretty rare.
@yingjieMiao


yingjieMiao:
thanks

PR #15079, opened 2015-04-28
Introduce RoverImageStore, ImagePath

leogrim:


leogrim:
@andrewconner @eishay @stkem

PR #15085, opened 2015-04-28
trying the other direction: way more db throughput

stkem:
@eishay @yingjieMiao


yingjieMiao:
This might work.

FYI, i just deployed cortex with reverted PR. Expect to see increase in latency from cortex.

PR #15096, opened 2015-04-29
website & ext endpoints for updating keep note

ahsu1230:
@2is10 @andrewconner 

These are endpoints for the extension & website that will edit a keep note & any hashtags in the note will be persisted as tags for that keep.

You can use them like this:

```
POST    /ext/libraries/:id/keeps/:k/note
POST    /site/libraries/:id/keeps/:k/note

where :id is libraryId and :k is keepId
```


kifi-backend/modules/shoebox/test/com/keepit/controllers/website/LibraryControllerTest.scala
```diff
@@ -1379,6 +1379,50 @@ class LibraryControllerTest extends Specification with ShoeboxTestInjector {
...
+          collectionRepo.getHashtagsByKeepId(keep.id.get) === Set.empty
+        }
+
+        // test adding a note (with hashtags)
+        val result3 = libraryController.editKeepNote(pubId1, keep1.externalId)(FakeRequest("POST", testPath).withBody(Json.obj("note" -> "thwip! #spiders [#avengers] [#tonysucks] blah")))
+        status(result3) must equalTo(NO_CONTENT)
+        db.readOnlyMaster { implicit s =>
+          val keep = keepRepo.getOpt(keep1.externalId).get
+          keep.note === Some("thwip! #spiders [#avengers] [#tonysucks] blah")
+          collectionRepo.getHashtagsByKeepId(keep.id.get).map(_.tag) === Set("tonysucks", "avengers")
```
ahsu1230, line 1419: @2is10 I hope I'm understanding this correctly. In these new editNote endpoints, the front-end clients should be doing the mark-up (otherwise we won't know the difference between a real kifi hashtag vs. a not-really hashtag). So whatever the client sends as a note, we will persist that note as is in the database. From there, we pick out the hashtags so we can tag the keep.

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
```diff
@@ -431,6 +432,19 @@ class KeepsCommander @Inject() (
     }
   }
 
+  def updateKeepNote(userId: Id[User], oldKeep: Keep, newNote: String)(implicit context: HeimdalContext) = {
+    val noteToPersist = Some(newNote.trim).filter(_.nonEmpty)
+    if (noteToPersist != oldKeep.note) {
+      val updatedKeep = db.readWrite { implicit s =>
+        keepRepo.save(oldKeep.copy(note = noteToPersist))
+      }
```
2is10, line 441: This persisting of the updated keep is also done by `persistHashtagsForKeep` below. I think it’d be better to do it just once, and in the same database session as the `Collection` updates.

kifi-backend/modules/shoebox/test/com/keepit/commanders/KeepsCommanderTest.scala
```diff
@@ -19,6 +20,10 @@ import org.joda.time.DateTime
 import org.specs2.mutable.Specification
 import com.keepit.shoebox.{ FakeShoeboxServiceModule, FakeKeepImportsModule }
 import com.keepit.common.store.FakeShoeboxStoreModule
+import com.keepit.model.UserFactory._
+import com.keepit.model.UserFactoryHelper._
+import com.keepit.model.KeepFactory._
+import com.keepit.model.KeepFactoryHelper._
```
2is10: These are already on lines 4-5 above.

kifi-backend/modules/shoebox/test/com/keepit/controllers/website/LibraryControllerTest.scala
2is10, line 1419: Yes, exactly. Nice work.

2is10:
Minor comments, looks good.

PR #15104, opened 2015-04-29
updating extension endpoint that temporarily backfills tags into note

2is10:
Also making HashtagCommander a simple `object` so that `MoarKeepData` can use it.

I think an `object` is appropriate since it’s just a collection of string processing functions—not a commander in our conventional sense (a layer between controller and repo).

@ahsu1230


kifi-backend/modules/shoebox/app/com/keepit/commanders/KeeperInfo.scala
```diff
@@ -94,7 +94,7 @@ object MoarKeepData {
   implicit val writes = new Writes[MoarKeepData] {
     def writes(o: MoarKeepData): JsValue = {
       val o2 = if (o.tags.nonEmpty && o.note.isEmpty) {
-        o.copy(note = Some(o.tags.map(t => '#' + t.replace(' ', '\u00a0')).mkString(" ")))
+        o.copy(note = Some(Hashtags.appendHashtagNamesToString("", o.tags)))
```
2is10, line 97: This line is all I set out to accomplish. Sorry for the big change.

kifi-backend/modules/shoebox/test/com/keepit/commanders/HashtagCommanderTest.scala
```diff
@@ -1,104 +0,0 @@
...
-import org.specs2.mutable.Specification
-
-class HashtagCommanderTest extends Specification with ShoeboxTestInjector {
-  def modules = Nil
-
-  "HashtagCommander" should {
-
-    "find hashtags in string" in {
-      withDb(modules: _*) { implicit injector =>
-        val commander = inject[HashtagCommander]
```
2is10, line 14: I noticed that this test wasn’t using the database. Now it runs lightning fast.

PR #15106, opened 2015-04-29
Revert "open up a little"

yingjieMiao:
Reverts kifi/keepit#15105

@stkem 

things eventually blow up after a while (e.g. 1 hour).


eishay:
what do you mean blow up?

yingjieMiao:
many long running sessions begin to cumulate. Reduce concurrency helps to keep the system stable (my guess).

PR #15113, opened 2015-04-30
Site and ext Facebook API update. Super simple.

andrewconner:
@2is10 Tested locally, this seems to be it for the clients.

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/common/services/angular-facebook-api.js
+++ b/kifi-backend/modules/shoebox/angular/src/common/services/angular-facebook-api.js
@@ -141,7 +141,7 @@ angular.module('jun.facebook', [])
 				}
 				js = d.createElement(s);
 				js.id = id;
-				js.src = '//connect.facebook.net/en_US/all.js';
+				js.src = '//connect.facebook.net/en_US/sdk.js';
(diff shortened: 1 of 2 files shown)


andrewconner:
Fortunately we don't use any parts of the API that changed client side anymore (we used to for login pages, but it's not used now).

andrewconner:
Facebook is actually leaving the old script up for now, looks like, so I don't think we're rushed to get this out for the extension.

2is10:
> Fortunately we don't use any parts of the API that changed client side anymore (we used to for login pages, but it's not used now).

Are you referring to the bit that shows the Facebook picture if the user is logged in to Facebook and has authed previously? The extension still does that. I thought it was removed from the site in haste for the redesign. I was hoping we’d put it back there too.

andrewconner:
I misspoke, we used more for the angular signup, but now it just redirects directly to /signup/facebook. The picture when user is logged in works still. In extension dev mode:

2is10:
Oh, I just merged this and tested it locally myself. It seems that, in the extension, it still works. But will it stop working for new users when FB starts giving us fake user IDs? Or will their graph API accept those fake user IDs?
https://graph.facebook.com/FAKE-USER-ID/picture?width=50&height=50

PR #15121, opened 2015-04-30
new mobile endpoints for editing keep note, and retrieving keep notes

ahsu1230:
@2is10


ahsu1230:
actually hold on this, it's possible we might be delaying mobile launch for notes... :octopus:

PR #15127, opened 2015-05-01
stop refetching twitter handles of missing users

eishay:


eishay:
@andrewconner @ahsu1230

PR #15128, opened 2015-05-01
handling escaped backslashes properly in future code

2is10:
@ahsu1230

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/Hashtags.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/Hashtags.scala
@@ -2,7 +2,7 @@ package com.keepit.commanders
 
 import com.keepit.model.Hashtag
 
-import scala.util.matching.Regex.Match
+import scala.util.matching.Regex
 
 object Hashtags {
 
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/Hashtags.scala
```diff
@@ -44,24 +44,24 @@ object Hashtags {
...
     note.map { noteStr =>
       val noteWithoutHashtags = Hashtags.removeAllHashtagsFromString(noteStr)
-      /* // todo: add this logic in once hashtags are supported!
-      val note2 = if (noteWithoutHashtags.nonEmpty && noteWithoutHashtags != noteStr) {
+
+      // todo(jared|aaron): remove false below once notes + hashtags are launched on any platform
+      val note2 = if (false && noteWithoutHashtags.nonEmpty && noteWithoutHashtags != noteStr) {
         hashTagRe.replaceAllIn(noteStr, m => {
-          "#" + backslashUnescapeRe.replaceAllIn(m.group(1), "$1")
+          Regex.quoteReplacement("#" + backslashUnescapeRe.replaceAllIn(m.group(1), "$1"))
```
2is10, line 58: The fix is this `Regex.quoteReplacement` call. It’s needed because `.replaceAllIn` interpolates the string that its second argument (our match function) returns. From the scaladoc:

> In the replacement String, a dollar sign (`$`) followed by a number will be interpreted as a reference to a group in the matched pattern, with numbers 1 through 9 corresponding to the first nine groups, and 0 standing for the whole match. Any other character is an error. The backslash (`\`) character will be interpreted as an escape character and can be used to escape the dollar sign. Use `Regex.quoteReplacement` to escape these characters.

PR #15129, opened 2015-05-01
ArticleContentExtractor

leogrim:
@yingjieMiao @stkem makes sense?


yingjieMiao:
lgtm. nice use of stream.

PR #15131, opened 2015-05-01
more aggressive reco clean frequency

yingjieMiao:
@stkem 

sweeps all users in 2 weeks. each batch has fewer users though, because of higher frequency.


stkem:
:+1:

PR #15133, opened 2015-05-01
new mobile endpoint for editing keeps (title & note -> persists hashtags...

ahsu1230:
... in notes as is)

@2is10


2is10:
:+1:

ahsu1230:

This is the new endpoint to edit your keep (takes in title & note, but no tags)

```
POST    /m/2/keeps/:id
```

Based on the markup process Jared described yesterday, we will look for any hashtags in this form `[#...]` in the note to persist as a tag. So, if you want a hashtag in your note, you want something like this: `I love [#sanfrancisco]`

The old one is still 

```
POST    /m/1/keeps/:id
```

PR #15135, opened 2015-05-01
Collaborator fields added

andrewconner:
Should be safe and backwards compatible everywhere. Tests updated.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtLibraryController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtLibraryController.scala
@@ -44,14 +44,17 @@ class ExtLibraryController @Inject() (
     extends UserActions with LibraryAccessActions with ShoeboxServiceController {
 
   def getLibraries() = UserAction { request =>
-    val datas = libraryCommander.getLibrariesUserCanKeepTo(request.userId) map { lib =>
-      val owner = db.readOnlyMaster { implicit s => basicUserRepo.load(lib.ownerId) }
-      LibraryData(
-        id = Library.publicId(lib.id.get),
-        name = lib.name,
-        color = lib.color,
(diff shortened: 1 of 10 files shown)


kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtLibraryController.scala
```diff
@@ -44,14 +44,17 @@ class ExtLibraryController @Inject() (
...
   def getLibraries() = UserAction { request =>
-    val datas = libraryCommander.getLibrariesUserCanKeepTo(request.userId) map { lib =>
-      val owner = db.readOnlyMaster { implicit s => basicUserRepo.load(lib.ownerId) }
-      LibraryData(
-        id = Library.publicId(lib.id.get),
-        name = lib.name,
-        color = lib.color,
-        visibility = lib.visibility,
-        path = Library.formatLibraryPath(owner.username, lib.slug))
+    val datas = libraryCommander.getLibrariesUserCanKeepTo(request.userId) map {
```
yingjieMiao, line 47: what happens if a user is keeping a page that is already in the library? (for user's own keeps, keeper already shows kept, so no problem like this)

andrewconner, line 47: It'll show as kept, as well. Here's what I see in Stephen's library, I didn't keep it (he did).

andrewconner, line 47: The larger point here is that collaborators are treated equal to owners in most cases. They can unkeep, keep, etc at will. Think of it like a Dropbox shared folder.

yingjieMiao, line 47: Great!

PR #15137, opened 2015-05-01
Prototype library updates in the reco feed for mobile (new endpoint)

stkem:
Still missing some smartness about when to show, but the API should
stay like this (or at least not change much), so mobile can start
testing.

@andrewconner


kifi-backend/modules/shoebox/conf/shoeboxService.routes
```diff
@@ -248,6 +248,10 @@ GET     /m/1/recos/public                       @com.keepit.controllers.mobile.M
 POST    /m/1/recos/trash                        @com.keepit.controllers.mobile.MobileRecommendationsController.trash(id: ExternalId[NormalizedURI])
 POST    /m/1/recos/feedback                     @com.keepit.controllers.website.RecommendationsController.updateUriRecommendationFeedback(id: ExternalId[NormalizedURI])
 
+POST     /m/1/feed                              @com.keepit.controllers.mobile.MobileRecommendationsController.feedV1Post()
```
stkem, line 251: @mrbrown14 @jaredpetker @DerekBurch @thassss new feed endpoint for you. Post body is the same as before, but this one has a third kind in it which is the library updates (from libs you follow)

PR #15139, opened 2015-05-01
BookmarkImporter use kept names instead of tagIds

ahsu1230:
@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/BookmarkImporter.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/BookmarkImporter.scala
@@ -103,8 +103,8 @@ class BookmarkImporter @Inject() (
     }.toMap
     val taggedKeeps = parsed.map {
       case Bookmark(t, h, tagNames, createdDate, originalJson) =>
-        val keepTags = tagNames.map(tags.get).flatten.map(_.id.get)
-        BookmarkWithTagIds(t, h, keepTags, createdDate, originalJson)
+        val keepTags = tagNames.map(tags.get).flatten
+        BookmarkWithTags(t, h, keepTags, createdDate, originalJson)
     }
 
     val (importId, rawKeeps) = createRawKeeps(userId, sourceOpt, taggedKeeps, libraryId)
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/app/com/keepit/controllers/website/BookmarkImporter.scala
```diff
@@ -227,7 +229,7 @@ class BookmarkImporter @Inject() (
           source = source.getOrElse(KeepSource.bookmarkFileImport),
           originalJson = originalJson,
           installationId = None,
-          tagIds = tags,
+          keepTags = hashtagsArray,
```
ahsu1230, line 231: create `JsArray` of hashtag names which get turned into `RawKeep` and processed by `KeepInterner.persistRawKeeps()`. Later RawKeepImporterPlugin will process batches of `RawKeep` and handle the `keepTags` field accordingly.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/BookmarkImporter.scala
```diff
@@ -237,4 +239,4 @@ class BookmarkImporter @Inject() (
 }
 
 case class Bookmark(title: Option[String], href: String, tags: List[String], createdDate: Option[DateTime], originalJson: Option[JsValue])
-case class BookmarkWithTagIds(title: Option[String], href: String, tags: List[Id[Collection]], createdAt: Option[DateTime], originalJson: Option[JsValue])
+case class BookmarkWithTags(title: Option[String], href: String, tags: List[Collection], createdAt: Option[DateTime], originalJson: Option[JsValue])
```
andrewconner: Do you need a `Collection` at all here? Why not just the array of strings?

ahsu1230: I did think about that, but I thought it would be more general having a `List[Collection]`

ahsu1230: On second thought, if we're trying to move off of Collections, we should replace as much as possible with Strings (hashtag names) instead

PR #15146, opened 2015-05-02
update library membership endpoint (site & mobile)

ahsu1230:
@andrewconner 

front-enders: @2is10 @mrbrown14 @thassss @jaredpetker @DerekBurch 
You can edit library memberships now using this:

```
POST    /m/1/libraries/:id/members/:uId/access/:access
POST    /site/libraries/:id/members/:uId/access/:access
:id is libraryId
:uId is userId
:access is a string for what access you would like to persist
```

For `access`, you can have the following: `read_write`, `read_insert`, `read_only`, and `none`
If access is none, the membership is deactivated.
You cannot change owners nor can you go from owner to a follower/collaborator.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
@@ -1682,4 +1682,35 @@ class LibraryCommander @Inject() (
     }
   }
 
+  ///////////////////
+  // Collaborators!
+  ///////////////////
+
+  def updateLibraryMembershipAccess(libraryId: Id[Library], targetUserId: Id[User], newAccess: Option[LibraryAccess]): Either[LibraryFail, LibraryMembership] = {
+    if (newAccess.isDefined && newAccess.get == LibraryAccess.OWNER) {
+      Left(LibraryFail(BAD_REQUEST, "cannot_change_access_to_owner"))
+    } else {
+      db.readOnlyMaster { implicit s =>
+        libraryMembershipRepo.getWithLibraryIdAndUserId(libraryId, targetUserId)
+      } match {
+        case None =>
+          Left(LibraryFail(NOT_FOUND, "membership_not_found"))
+        case Some(targetMembership) =>
+          if (targetMembership.access == LibraryAccess.OWNER) {
+            Left(LibraryFail(BAD_REQUEST, "cannot_change_owner_access"))
+          } else {
+            val updatedMembership = db.readWrite { implicit s =>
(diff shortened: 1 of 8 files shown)


andrewconner:
Whoops, somehow I missed this PR. Reviewing, sorry about that.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -1682,4 +1682,35 @@ class LibraryCommander @Inject() (
...
+      db.readOnlyMaster { implicit s =>
+        libraryMembershipRepo.getWithLibraryIdAndUserId(libraryId, targetUserId)
+      } match {
+        case None =>
+          Left(LibraryFail(NOT_FOUND, "membership_not_found"))
+        case Some(targetMembership) =>
+          if (targetMembership.access == LibraryAccess.OWNER) {
+            Left(LibraryFail(BAD_REQUEST, "cannot_change_owner_access"))
+          } else {
+            val updatedMembership = db.readWrite { implicit s =>
```
andrewconner, line 1702: This lets anyone who is a follower become a collaborator.

andrewconner:
It may be more convenient to make the APIs role-based, so clients talk in terms of "followers", "collaborators", "owners", etc. This way, clients don't have to know that we have a "read_insert" permission (that technically isn't and won't be supported), and "none" isn't a valid existing access anyways. It'll also match the UI well.

`POST /site/libraries/:id/members/:uId/access` with a body `{"access": "follower"}`?

Also, make sure to do permissions checks. Owners can set this for any existing follower/collaborator. Followers can set "none". Collaborators can demote themselves to followers, or "none".

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
ahsu1230, line 1702: Hm, I was under the impression that an owner or a collaborator could promote a follower to have `read_write` access. So rather than having to invite somebody, wait until they accept, I could just simply promote their access level. Similarly, I could also demote the access level as well. Is that right?

andrewconner, line 1702: Sorry, rewording: this lets someone who is a follower to set themselves to be a collaborator. ie, "take over" a library

ahsu1230, line 1702: I see you're right. I didn't add a `canWrite` check for changing access.

PR #15149, opened 2015-05-04
use timeSlicer in libraryChecker

yingjieMiao:
@ahsu1230 @eishay 

spread checks over a week to reduce system load and prepare for future scalability.


ahsu1230:
thanks yingjie!

PR #15151, opened 2015-05-04
added a guard so only SuperAdmins can assign or deassign admins

camhashemi:
@stkem @andrewconner @eishay


camhashemi:
tests coming soon

andrewconner:
Great first push!

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminUserController.scala
```diff
@@ -76,6 +76,8 @@ object UserViewTypes {
 }
 import UserViewTypes._
 
+case class SuperAdminAccessDenied() extends Exception
```
andrewconner, line 79: In Scala, in general, exceptions are reserved for particularly exceptional behavior (i.e., divide by zero, connection failures, etc) and not control flow. The reason is that exceptions break the function signature contract (if you have a `def f(a: Int): String`, throwing an exception breaks the intended type return value).

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminUserController.scala
```diff
@@ -499,21 +501,40 @@ class AdminUserController @Inject() (
...
+  def addExperimentAction(userId: Id[User], experiment: String) = AdminUserAction { request =>
+    try {
+      addExperiment(request.userId, userId, experiment)
+      Ok(Json.obj(experiment -> true))
+    } catch {
+      case SuperAdminAccessDenied() => Forbidden
+    }
+  }
+
+  def addExperiment(requesterUserId: Id[User], userId: Id[User], experiment: String) {
```
andrewconner, line 520: We also try to avoid returning `Unit` from Scala methods. In Java, void is pretty common; but in Scala, it's unusual (and nearly always avoidable) to return nothing. In this case, `addExperiment` can either succeed (perhaps by returning the typed experiment that was applied), or fail (perhaps this isn't a valid experiment? or the user doesn't have permissions to do this?). That could be represented a couple ways, but one idea is `Either[String, ExperimentType]`.

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminUserController.scala
```diff
@@ -499,21 +501,40 @@ class AdminUserController @Inject() (
...
+  def isSuperAdmin(userId: Id[User]) = {
+    val superAdminSet: Set[Id[User]] = Set(Id[User](1))
+    superAdminSet contains userId
+  }
+
+  def isAdminExperiment(expType: ExperimentType) = expType == ExperimentType.ADMIN
+
+  def addExperimentAction(userId: Id[User], experiment: String) = AdminUserAction { request =>
+    try {
+      addExperiment(request.userId, userId, experiment)
```
andrewconner, line 513: If addExperiment did return `Either[String, ExperimentType]`, this could be rewritten as this with no try/catch:

``` scala
addExperiment(request.userId, userId, experiment) match {
  case Right(s) => Forbidden(s)
  case Left(exp) => Ok(Json.obj(exp.value -> true))
}
```

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminUserController.scala
andrewconner, line 520: Another possibility is just `Option[ExperimentType]`, where failures are just swallowed if we truly didn't care.

andrewconner:
Left a few comments. Some idiomatic Scala takes a bit to learn, so no worries. Awesome job.

PR #15158, opened 2015-05-05
higher read timeout for library image uploads

2is10:
@andrewconner I’m not sure this solves [the problem](https://twitter.com/UnsungHero97/status/595343005105541120) the guy on Twitter had, but it’s my best first stab at fixing it.

He said he was uploading a 2.4MB file. I suspect he was on a slower connection than ours. Nginx [describes](http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_read_timeout) `proxy_read_timeout` this way:

> Defines a timeout for reading a response from the proxied server. The timeout is set only between two successive read operations, not for the transmission of the whole response. If the proxied server does not transmit anything within this time, the connection is closed.

I’m not sure how the mechanics of uploads work between nginx and Play. Do you think nginx waits until the entire file is uploaded (written to temp file?) before passing the request on to Play? If so, that should rule out connection speed as a potential culprit, and would suggest image processing time as the culprit. If nginx streams bytes on to Play as they arrive, and Play writes to a temp file, then we may need to determine whether Play initiates the response 1) before the upload is complete or 2) as soon as the upload is complete and before processing or 3) after processing is done. Or maybe the `proxy_read_timeout` also applies to the first byte of the response, in which case the total upload time + image processing time counts.

Regardless of the specifics, I raised the value for this endpoint from 20 sec to 3 min. I guess we could check the nginx access logs for uploads—specifically long ones—to see if this makes any difference.

### DIFF
--- a/tools/etc/nginx/conf.d/shoebox.conf
+++ b/tools/etc/nginx/conf.d/shoebox.conf
@@ -126,7 +126,7 @@ server {
   }
 
   location / {
-    if ($http_origin ~* (https?://[^/]*\.(ezkeep|kifi)\.com(:[0-9]+)?)) {
+    if ($http_origin ~* ^https?://[^/]*\.(ezkeep|kifi)\.com(:[0-9]+)?$) {
         set $cors "true";
     }
     if ($request_method = 'OPTIONS') {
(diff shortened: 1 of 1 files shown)


andrewconner:
Great. I'm not sure how nginx does it.

PR #15161, opened 2015-05-05
Only include "updates to libraries you follow" in reco feed once per 12 hours (and only when there are any)

stkem:
@andrewconner @eishay 
Don't merge yet, because this is going to make it really difficult to test on the clients.


andrewconner:
Perhaps an experiment could make it deliver every time?

stkem:
@jaredpetker going live in a few minutes

PR #15163, opened 2015-05-05
added AdminUserControllerTest.scala

camhashemi:
test cases assert that only superAdmins can add/remove admins, non-superAdmins cannot add/remove admins

edit:
removed Exception throwing in favor of an Either[String, ExperimentType] return for addExperiment/removeExperiment

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/admin/AdminUserController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/admin/AdminUserController.scala
@@ -76,8 +76,6 @@ object UserViewTypes {
 }
 import UserViewTypes._
 
-case class SuperAdminAccessDenied() extends Exception
-
 class AdminUserController @Inject() (
     val userActionsHelper: UserActionsHelper,
     db: Database,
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/app/com/keepit/admin/AdminUserController.scala
```diff
@@ -509,18 +507,16 @@ class AdminUserController @Inject() (
   def isAdminExperiment(expType: ExperimentType) = expType == ExperimentType.ADMIN
 
   def addExperimentAction(userId: Id[User], experiment: String) = AdminUserAction { request =>
-    try {
-      addExperiment(request.userId, userId, experiment)
-      Ok(Json.obj(experiment -> true))
-    } catch {
-      case SuperAdminAccessDenied() => Forbidden
+    addExperiment(requesterUserId = request.userId, userId, experiment) match {
```
andrewconner, line 510: Nice! Very idiomatic Scala.

andrewconner:
@camhashemi Looks good.

PR #15169, opened 2015-05-05
LibraryOwnerAction - allow certain library functions to be only for owne...

ahsu1230:
...r

@andrewconner


ahsu1230:
LibraryOwnerAction only allows owners to edit library (name, visibility, image, etc.)
all other LibraryWriteActions are left as is for keeping, unkeeping, etc.

andrewconner:
Excellent!

PR #15174, opened 2015-05-06
change keep ownership when collaborator leaves (or gets kicked out)

ahsu1230:
@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
@@ -1725,14 +1725,10 @@ class LibraryCommander @Inject() (
             db.readWrite { implicit s =>
               newAccess match {
                 case None =>
-                  if (targetMem.access == LibraryAccess.READ_WRITE) {
-                    convertKeepOwnershipToLibraryOwner(targetMem.userId, library)
-                  }
+                  SafeFuture { convertKeepOwnershipToLibraryOwner(targetMem.userId, library) }
                   Right(libraryMembershipRepo.save(targetMem.copy(state = LibraryMembershipStates.INACTIVE)))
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -1699,32 +1708,39 @@ class LibraryCommander @Inject() (
...
-        case (Some(mem), Some(targetMem)) =>
+        case (Some(mem), Some(targetMem), library) =>
           if ((mem.isOwner && !targetMem.isOwner) || // owners can edit anyone except themselves
             (mem.isCollaborator && mem.userId == targetMem.userId) || // a collaborator can only edit herself
             (mem.isFollower && mem.userId == targetMem.userId)) { // a follower can only edit herself
             db.readWrite { implicit s =>
               newAccess match {
                 case None =>
+                  if (targetMem.access == LibraryAccess.READ_WRITE) {
+                    convertKeepOwnershipToLibraryOwner(targetMem.userId, library)
```
andrewconner: We probably want to call `convertKeepOwnershipToLibraryOwner` in all cases, for the case of a collaborator who becomes a follower, and then leaves.

andrewconner:
Nice job.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
ahsu1230: ohh I see. nice catch, thanks!

andrewconner:
Good idea, I remember you mentioning this. @ahsu1230 can you add this when you add the `order` field for ordering keeps? (ie no need to do it immediately, but soon)

ahsu1230:
ok gotcha! sounds good. for `original_keeper_id`, I will backfill them as the current keep's userId

PR #15188, opened 2015-05-06
Subscribing to library update notifications

stkem:
Still missing from this: Push Notifications

@andrewconner @ahsu1230 
cc: @jaredpetker @2is10

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/model/LibraryMembership.scala
+++ b/kifi-backend/modules/common/app/com/keepit/model/LibraryMembership.scala
@@ -25,7 +25,8 @@ case class LibraryMembership(
     listed: Boolean = true, // whether library appears on user's profile
     lastViewed: Option[DateTime] = None,
     lastEmailSent: Option[DateTime] = None,
-    lastJoinedAt: Option[DateTime] = None) extends ModelWithState[LibraryMembership] with ModelWithSeqNumber[LibraryMembership] {
+    lastJoinedAt: Option[DateTime] = None,
+    subscribedToUpdates: Boolean = false) extends ModelWithState[LibraryMembership] with ModelWithSeqNumber[LibraryMembership] {
 
   def withId(id: Id[LibraryMembership]): LibraryMembership = this.copy(id = Some(id))
   def withUpdateTime(now: DateTime): LibraryMembership = this.copy(updatedAt = now)
@@ -55,7 +56,8 @@ object LibraryMembership {
     (__ \ 'listed).format[Boolean] and
     (__ \ 'lastViewed).formatNullable[DateTime] and
     (__ \ 'lastEmailSent).formatNullable[DateTime] and
-    (__ \ 'lastJoinedAt).formatNullable[DateTime]
+    (__ \ 'lastJoinedAt).formatNullable[DateTime] and
+    (__ \ 'subscribedToUpdates).format[Boolean]
   )(LibraryMembership.apply, unlift(LibraryMembership.unapply))
 }
 
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
@@ -147,15 +147,17 @@ class LibraryCommander @Inject() (
     (libInfo.copy(image = imageOpt), memOpt)
   }
 
-  def getLibraryWithOwnerAndCounts(libraryId: Id[Library], viewerUserId: Id[User]): Either[LibraryFail, (Library, BasicUser, Int, Option[Boolean])] = {
+  def getLibraryWithOwnerAndCounts(libraryId: Id[Library], viewerUserId: Id[User]): Either[LibraryFail, (Library, BasicUser, Int, Option[Boolean], Boolean)] = {
     db.readOnlyReplica { implicit s =>
       val library = libraryRepo.get(libraryId)
       val mine = library.ownerId == viewerUserId
-      val following = if (mine) None else Some(libraryMembershipRepo.getWithLibraryIdAndUserId(libraryId, viewerUserId).isDefined)
+      val memOpt = libraryMembershipRepo.getWithLibraryIdAndUserId(libraryId, viewerUserId)
+      val following = if (mine) None else Some(memOpt.isDefined)
+      val subscribedToUpdates = memOpt.map(_.subscribedToUpdates).getOrElse(false)
       if (library.visibility == LibraryVisibility.PUBLISHED || mine || following.get) {
         val owner = basicUserRepo.load(library.ownerId)
         val followerCount = libraryMembershipRepo.countWithLibraryIdByAccess(library.id.get).readOnly
-        Right(library, owner, followerCount, following)
+        Right(library, owner, followerCount, following, subscribedToUpdates)
       } else {
         Left(LibraryFail(FORBIDDEN, "library_access_denied"))
       }
@@ -689,9 +691,9 @@ class LibraryCommander @Inject() (
     }
   }
 
-  def getLibrariesUserCanKeepTo(userId: Id[User]): Seq[(Library, Boolean)] = {
+  def getLibrariesUserCanKeepTo(userId: Id[User]): Seq[(Library, Boolean, Boolean)] = {
     db.readOnlyMaster { implicit s =>
-      libraryRepo.getByUserWithCollab(userId, excludeAccess = Some(LibraryAccess.READ_ONLY)).map(r => (r._2, r._3))
+      libraryRepo.getByUserWithCollab(userId, excludeAccess = Some(LibraryAccess.READ_ONLY)).map(r => (r._2, r._3, r._1.subscribedToUpdates))
     }
   }
 
@@ -1007,7 +1009,7 @@ class LibraryCommander @Inject() (
       if (newKeep.libraryId.get != library.id.get) { throw new IllegalArgumentException(s"Keep ${newKeep.id.get} does not belong to expected library ${library.id.get}") }
     }
     val (relevantFollowers, usersById) = db.readOnlyReplica { implicit session =>
-      val relevantFollowers = libraryMembershipRepo.getWithLibraryId(library.id.get).map(_.userId).filter(experimentCommander.userHasExperiment(_, ExperimentType.NEW_KEEP_NOTIFICATIONS)).toSet
+      val relevantFollowers: Set[Id[User]] = libraryMembershipRepo.getWithLibraryId(library.id.get).filter(_.subscribedToUpdates).map(_.userId).toSet
       val usersById = userRepo.getUsers(newKeeps.map(_.userId) :+ library.ownerId)
       (relevantFollowers, usersById)
     }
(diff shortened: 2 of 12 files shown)


kifi-backend/modules/common/app/com/keepit/model/LibraryMembership.scala
```diff
@@ -25,7 +25,8 @@ case class LibraryMembership(
     listed: Boolean = true, // whether library appears on user's profile
     lastViewed: Option[DateTime] = None,
     lastEmailSent: Option[DateTime] = None,
-    lastJoinedAt: Option[DateTime] = None) extends ModelWithState[LibraryMembership] with ModelWithSeqNumber[LibraryMembership] {
+    lastJoinedAt: Option[DateTime] = None,
+    subscribedToUpdates: Boolean = false) extends ModelWithState[LibraryMembership] with ModelWithSeqNumber[LibraryMembership] {
```
andrewconner, line 29: I think it was Eishay that suggested staying away from boolean db fields. Strings are cheap enough where this could be `subscription: Option[String] = None` with possible value `"subscribed"` for now. Naming may be bad. I can see a future where users may want to chose how they get notified.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -147,15 +147,17 @@ class LibraryCommander @Inject() (
...
       val mine = library.ownerId == viewerUserId
-      val following = if (mine) None else Some(libraryMembershipRepo.getWithLibraryIdAndUserId(libraryId, viewerUserId).isDefined)
+      val memOpt = libraryMembershipRepo.getWithLibraryIdAndUserId(libraryId, viewerUserId)
+      val following = if (mine) None else Some(memOpt.isDefined)
+      val subscribedToUpdates = memOpt.map(_.subscribedToUpdates).getOrElse(false)
       if (library.visibility == LibraryVisibility.PUBLISHED || mine || following.get) {
         val owner = basicUserRepo.load(library.ownerId)
         val followerCount = libraryMembershipRepo.countWithLibraryIdByAccess(library.id.get).readOnly
-        Right(library, owner, followerCount, following)
+        Right(library, owner, followerCount, following, subscribedToUpdates)
```
andrewconner: I know this predates this PR, but accidental auto-tupling. Being removed in the next Scala. `Right((library, owner, followerCount, following, subscribedToUpdates))`

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -1689,6 +1691,20 @@ class LibraryCommander @Inject() (
     }
   }
 
+  def updatedLibraryUpdateSusbcription(userId: Id[User], libraryId: Id[Library], subscribedToUpdatesNew: Boolean): Either[LibraryFail, LibraryMembership] = {
+    db.readOnlyMaster { implicit s =>
+      libraryMembershipRepo.getWithLibraryIdAndUserId(libraryId, userId)
+    } match {
+      case None => Left(LibraryFail(NOT_FOUND, "Can't subscribe to updaes from a library you are not following"))
```
andrewconner: These errors should be for clients, since it's an API endpoint. ie, human readability may be a negative since someone could accidentally change it. Prefer to stick to tokens that _look_ machine-readable, like `need_to_follow_to_subscribe` or something.

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryMembershipRepo.scala
```diff
@@ -76,7 +76,8 @@ class LibraryMembershipRepoImpl @Inject() (
     def lastEmailSent = column[Option[DateTime]]("last_email_sent", O.Nullable)
     def listed = column[Boolean]("listed", O.NotNull)
     def lastJoinedAt = column[Option[DateTime]]("last_joined_at", O.Nullable)
-    def * = (id.?, libraryId, userId, access, createdAt, updatedAt, state, seq, showInSearch, listed, lastViewed, lastEmailSent, lastJoinedAt) <> ((LibraryMembership.apply _).tupled, LibraryMembership.unapply)
+    def subscribedToUpdates = column[Boolean]("subscribed_to_updates", O.Nullable)
```
andrewconner: Type is `Boolean` but column is nullable?

kifi-backend/modules/common/app/com/keepit/model/LibraryMembership.scala
2is10, line 29: or `updates` with value `"subscribed"`

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -1689,6 +1691,20 @@ class LibraryCommander @Inject() (
     }
   }
 
+  def updatedLibraryUpdateSusbcription(userId: Id[User], libraryId: Id[Library], subscribedToUpdatesNew: Boolean): Either[LibraryFail, LibraryMembership] = {
```
2is10: spelling of `Susbcription`

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileLibraryController.scala
```diff
@@ -126,8 +127,9 @@ class MobileLibraryController @Inject() (
           val idealSize = imageSize.flatMap { s => Try(ImageSize(s)).toOption }.getOrElse(MobileLibraryController.defaultLibraryImageSize)
           libraryCommander.createFullLibraryInfo(request.userIdOpt, false, library, idealSize).map { libInfo =>
             val memOpt = libraryCommander.getMaybeMembership(request.userIdOpt, library.id.get)
+            val subscribedToUpdates = memOpt.map(_.subscribedToUpdates).getOrElse(false)
             val accessStr = memOpt.map(_.access.value).getOrElse("none")
-            Ok(Json.obj("library" -> Json.toJson(libInfo), "membership" -> accessStr))
+            Ok(Json.obj("library" -> Json.toJson(libInfo), "membership" -> accessStr, "subscribedToUpdates" -> subscribedToUpdates))
```
2is10, line 132: This new value is a property of the membership. It might be clearer to use a membership object containing the access level and subscription preference. I suppose this would require a new endpoint version, or else returning the object keyed by a different name, keeping `"membership"` too, as is, for now for backwards compatibility. Worth it?

kifi-backend/modules/shoebox/conf/shoeboxService.routes
```diff
@@ -138,6 +138,8 @@ GET     /ext/libraries/:id/keeps/:k @com.keepit.controllers.ext.ExtLibraryContro
 POST    /ext/libraries/:id/keeps/:k @com.keepit.controllers.ext.ExtLibraryController.updateKeep(id: PublicId[Library], k: ExternalId[Keep])
 DELETE  /ext/libraries/:id/keeps/:k @com.keepit.controllers.ext.ExtLibraryController.removeKeep(id: PublicId[Library], k: ExternalId[Keep])
 POST    /ext/libraries/:id/keeps/:k/note  @com.keepit.controllers.ext.ExtLibraryController.editKeepNote(id: PublicId[Library], k: ExternalId[Keep])
+POST    /ext/libraries/:id/subscribedToUpdates @com.keepit.controllers.ext.ExtLibraryController.setSubscribedToUpdates(id: PublicId[Library], subscribedToUpdatesNew: Boolean)
+
```
2is10, line 142: We’ve aimed to use a single word in each path segment in our outward-facing (non-admin) APIs. How about choosing either `/updates` or `/subscription`?

2is10:
What does it mean to subscribe to notifications? Will I get a Kifi notification (extension and/or mobile) and an email for every new keep? Should we provide a way to disable one mechanism or the other? Maybe that should only be controlled with a higher-level setting, and not per library.

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileLibraryController.scala
```diff
@@ -111,10 +111,11 @@ class MobileLibraryController @Inject() (
     libraryCommander.getLibraryById(userIdOpt, false, libraryId, idealSize) map { libInfo =>
       val memOpt = libraryCommander.getMaybeMembership(userIdOpt, libraryId)
       val accessStr = memOpt.map(_.access.value).getOrElse("none")
+      val subscribedToUpdates = memOpt.map(_.subscribedToUpdates).getOrElse(false)
       val editedLibInfo = libInfo.copy(keeps = libInfo.keeps.map { k =>
         k.copy(note = Hashtags.formatMobileNote(k.note, v1))
       })
-      Ok(Json.obj("library" -> Json.toJson(editedLibInfo), "membership" -> accessStr))
+      Ok(Json.obj("library" -> Json.toJson(editedLibInfo), "membership" -> accessStr, "subscribedToUpdates" -> subscribedToUpdates))
```
ahsu1230, line 118: extending @2is10 's comment below. `v2` of this endpoint is not being used in production yet.
So you could use a `"membership"` object here to contain the access & subscriptionToUpdates. Will be super useful later on! Just make sure to change the object only for v2 and not for v1

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryMembershipRepo.scala
stkem: Woops. Copy paste error. Fixing.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
stkem: Ah, good catch

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileLibraryController.scala
andrewconner, line 118: That's a good idea.

kifi-backend/modules/common/app/com/keepit/model/LibraryMembership.scala
stkem, line 29: I'm not a huge fan of this. The product use case (you get notified or you don't) maps very clearly to true/false. A string field with just a single valid value just makes it harder to understand. _If_ in the future we allow for more control for the user here it would be easy to migrate.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
stkem: Oops. Fixing.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
stkem: Sounds good. Changing it.

kifi-backend/modules/shoebox/conf/shoeboxService.routes
stkem, line 142: Oh, sure. I'm going with `subscription` unless someone feels strongly otherwise.

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileLibraryController.scala
stkem, line 118: Currently `getLibraryById` (both versions) and `getLibraryByPath` (which does not have 2 versions) return the same format. Changing the new version of `getLibraryById` would make them different. Easy change, but I'm not sure that makes thing easier for the clients (unless we change all the calls). @jaredpetker would you like to weigh in on what you would prefer?

stkem:
@2is10 It means getting a kifi & push notification (not implemented yet) for every new keep that is not yours. It's sort of a super-follow.

stkem:
I'm going to merge this to test the actual subscribing. Happy to change the client APIs when feedback from mobile comes.

PR #15190, opened 2015-05-06
Library counts refactor, adding numCollabLibraries

andrewconner:
@ahsu1230 Mind a quick review? Tests pass with the additional field. Merge if you're happy with it.


kifi-backend/modules/shoebox/app/com/keepit/model/LibraryRepo.scala
```diff
@@ -39,8 +39,8 @@ trait LibraryRepo extends Repo[Library] with SeqNumberFunction[Library] {
   def getFollowingLibrariesForSelf(userId: Id[User], page: Paginator)(implicit session: RSession): Seq[Library]
   def getFollowingLibrariesForAnonymous(userId: Id[User], page: Paginator)(implicit session: RSession): Seq[Library]
   // def countFollowingLibrariesForSelf(userId: Id[User])(implicit session: RSession): Int  // use LibraryMembershipRepo.countWithUserIdAndAccess(userId, READ_ONLY) instead (cached)
-  def countFollowingLibrariesForOtherUser(userId: Id[User], viewerId: Id[User])(implicit session: RSession): Int
-  def countFollowingLibrariesForAnonymous(userId: Id[User])(implicit session: RSession): Int
+  def countMemberLibrariesForOtherUser(userId: Id[User], viewerId: Id[User])(implicit session: RSession): Map[LibraryAccess, Int]
+  def countMemberLibrariesForAnonymous(userId: Id[User])(implicit session: RSession): Map[LibraryAccess, Int]
```
ahsu1230, line 43: nice, these methods will definitely be useful

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -1512,31 +1512,38 @@ class LibraryCommander @Inject() (
...
+  def countLibraries(userId: Id[User], viewer: Option[Id[User]]): (Int, Int, Int, Option[Int]) = {
     viewer match {
       case None =>
         db.readOnlyReplica { implicit s =>
-          val numLibsOwned = libraryRepo.countLibrariesOfUserForAnonymous(userId)
-          val numLibsFollowing = libraryRepo.countFollowingLibrariesForAnonymous(userId)
-          (numLibsOwned, numLibsFollowing, None)
+          val counts = libraryRepo.countMemberLibrariesForAnonymous(userId)
+          val numLibsOwned = counts.getOrElse(LibraryAccess.OWNER, 0)
+          val numLibsCollab = counts.getOrElse(LibraryAccess.READ_WRITE, 0) + counts.getOrElse(LibraryAccess.READ_ONLY, 0)
```
ahsu1230: I don't think you want to add the number libraries a user collaborates in with the number libraries a user follows. At least in this method. According to your comment above, the number of libraries user collaborates on (that a viewer can see) should not include libraries the user also follows

PR #15191, opened 2015-05-06
added URI Info Admin Page

camhashemi:
added admin/uriId/:uriId to shoeboxService.routes, getURIInfo() to UrlController, and uriId.scala.html

no tests written.

@leogrim


kifi-backend/modules/shoebox/conf/shoeboxService.routes
```diff
@@ -563,6 +563,7 @@ GET     /admin/pageInfo/:id         @com.keepit.controllers.admin.AdminPageInfoC
 GET     /admin/pageInfos            @com.keepit.controllers.admin.AdminPageInfoController.pageInfos(page:Int ?= 0, size:Int ?= 50)
 
 GET     /admin/scraped/:uriId       @com.keepit.controllers.admin.ScraperAdminController.getScraped(uriId: Id[NormalizedURI])
+GET     /admin/uriId/:uriId         @com.keepit.controllers.admin.UrlController.getURIInfo(uriId: Id[NormalizedURI])
```
2is10: The string `"Id"` is not normally used in paths. Consider just `/admin/uri/:uriId`.

kifi-backend/modules/shoebox/app/com/keepit/admin/UrlController.scala
```diff
@@ -356,6 +359,17 @@ class UrlController @Inject() (
     Redirect(routes.ScraperAdminController.getScraped(uriId))
   }
 
+  def getURIInfo(id: Id[NormalizedURI]) = AdminUserPage.async { implicit request =>
+    val articleInfoMapFuture = roverServiceClient.getArticleInfosByUris(Set(id))
+    val uri = db.readOnlyReplica { implicit s => uriRepo.get(id) }
+    articleInfoMapFuture.map { articleInfoMap =>
+      articleInfoMap.get(id) match {
+        case Some(articleInfoSet) => Ok(html.admin.uriId(uri, articleInfoSet))
+        case None => NotFound("Article Info not found for uri" + id.toString)
```
leogrim: You can probably just use `articleInfoMap.getOrElse(id,Set.empty)` 
(in theory, `getArticleInfosByUris` should always contain all the ids that were passed in, but that's safer :)

kifi-backend/modules/shoebox/app/views/admin/uriId.scala.html
```diff
@@ -0,0 +1,51 @@
...
+    </tr>
+    @for(articleInfo <- articleInfoSet) {
+        <tr>
+            <th>Article Kind</th><td>articleInfo.kind</td>
+        </tr>
+        <tr>
+            <th>Best Version</th><td>articleInfo.bestVersion.getOrElse("Not found")</td>
+        </tr>
+        <tr>
+            <th>Latest Version</th><td>articleInfo.latestVersion.getOrElse("Not found")</td>
```
leogrim: I would say `"None"` instead of `"Not found"`, which sounds like something's wrong.

kifi-backend/modules/shoebox/app/views/admin/uriId.scala.html
```diff
@@ -0,0 +1,51 @@
...
+        <th>Updated at</th><td>@adminHelper.dateTimeDisplay(uri.updatedAt)</td>
+    </tr>
+    @*
+    <tr>
+        <th>Last scrape</th><td>@info.map(scrapeInfo => adminHelper.dateTimeDisplay(scrapeInfo.lastScrape))</td>
+    </tr>
+    <tr>
+        <th>Next scrape</th><td>@info.map(scrapeInfo => adminHelper.dateTimeDisplay(scrapeInfo.nextScrape))</td>
+    </tr>
+    *@
```
leogrim: This is going away, we can get rid of it.

kifi-backend/modules/shoebox/app/com/keepit/admin/UrlController.scala
camhashemi: That way we can still display the URI info, in the rare case that the articleInfoSet is not found. Thank you.

kifi-backend/modules/shoebox/conf/shoeboxService.routes
camhashemi: Changed it. Thank you.

kifi-backend/modules/shoebox/app/views/admin/uriId.scala.html
camhashemi: Okay. Let me know if any more NormalizedURI fields should be included, I wouldn't know.

https://github.com/kifi/keepit/blob/0740053d298b600f4864d04e19b3599d57593a46/kifi-backend/modules/common/app/com/keepit/model/NormalizedURI.scala

camhashemi:
@leogrim git woes are clearing up and it can now merge. there's just this odd ActivityEmailCronPlugin diff left. I made sure to copy directly from the current file at https://github.com/kifi/keepit/blob/master/kifi-backend/modules/shoebox/app/com/keepit/shoebox/cron/ActivityEmailCronPlugin.scala, so not sure why it displays a change.

camhashemi:
@stkem

stkem:
lgtm. Go for it.

PR #15196, opened 2015-05-07
collaborators (num & basic users) in small library cards (website)

ahsu1230:
@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/model/LibraryView.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/model/LibraryView.scala
@@ -162,8 +162,13 @@ object LibraryCardInfo {
     JsObject((Json.toJson(o).as[JsObject].value - "owner").toSeq)
   }
 
-  def showable(members: Seq[BasicUser]): Seq[BasicUser] = {
-    members.filter(_.pictureName != "0.jpg")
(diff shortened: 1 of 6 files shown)


kifi-backend/modules/shoebox/app/com/keepit/model/LibraryView.scala
```diff
@@ -146,18 +150,25 @@ case class LibraryCardInfo(
...
 
-  def showable(followers: Seq[BasicUser]): Seq[BasicUser] = {
-    followers.filter(_.pictureName != "0.jpg").take(3)
+  def makeMembersShowable(members: Seq[BasicUser], filterBadPics: Boolean): Seq[BasicUser] = {
+    if (filterBadPics) {
+      members.filter(_.pictureName != "0.jpg").take(3)
+    } else {
+      val (membersWithPics, membersNoPics) = members.partition(_.pictureName != "0.jpg")
+      membersWithPics ++ membersNoPics
+    }
```
ahsu1230, line 171: keep the original functionality for followers
For collaborators, we don't want to filter out any users, even if they have bad pictures! Simply move them to the back.
On the other hand, once we know the limit of how many collaborators to fit, we can use `take()`

andrewconner, line 171: Good thinking

andrewconner:
Looks good

PR #15199, opened 2015-05-07
consider more signals in feedback loop

yingjieMiao:
@stkem 

if you are not comfortable with the bayesian approach, we can switch to a frequentist approach, whose behavior is bumpier.


stkem:
lgtm to start with. No objections to a Bayesian approach in general.

mrbrown14:
I vote for a bayesfrequent approach :D

PR #15209, opened 2015-05-07
build fix

ahsu1230:


ahsu1230:
@2is10 I think I finally found what's been wrong with the broken build lol

ahsu1230:
@andrewconner

2is10:
Thanks, and sorry. Maybe "kifi.com" was omitted from the mapper intentionally because someone preferred showing "www.kifi.com" when we’re referring to our site—I dunno, there were no explanations in comments or unit tests.

PR #15216, opened 2015-05-08
simple push notification hookup for library subscriptions

stkem:
Note that this is a little naive and when we have libraries with
greater than, say, 5k subscribers, we’ll need to queue these
notifications somewhere so they don’t get lost during deploys and such.

@andrewconner @eishay


eishay:
yey

eishay:
how about checking number of subscribers and if they're greater then 100 send an airbrake or something to warn us that we're going to have to deal with it?

stkem:
Good idea. I'll add that.

stkem:
Merging to test it out (I'm the only one with a subscription right now). Happy to respond to further comments after the fact.

PR #15219, opened 2015-05-08
Sitemap for libraries is over 50k as well, reducing lib count

eishay:
by trimming low quality libraries


eishay:
@andrewconner

andrewconner:
Works for me

PR #15223, opened 2015-05-08
edited formatting for uri.scala.html

camhashemi:
@leogrim minor changes to improve the UI

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/admin/UrlController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/admin/UrlController.scala
@@ -382,7 +382,7 @@ class UrlController @Inject() (
     fBestArticleWithId.map { bestArticleWithId: Map[Id[NormalizedURI], Set[Article]] =>
       val bestArticles = bestArticleWithId.getOrElse(uriId, Set.empty)
       val targetArticle = bestArticles.filter(article => article.kind == kind).head
-      Ok(Json.obj(("content", targetArticle)))
+      Ok(Json.obj(("content" -> targetArticle.content.content.getOrElse("").toString)))
     }
   }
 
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/app/com/keepit/admin/UrlController.scala
```diff
@@ -391,7 +391,7 @@ class UrlController @Inject() (
     fBestArticleWithId.map { bestArticleWithId: Map[Id[NormalizedURI], Set[Article]] =>
       val bestArticles = bestArticleWithId.getOrElse(uriId, Set.empty)
       val targetArticle = bestArticles.filter(article => article.kind == kind).head
-      Ok(Json.obj("content" -> targetArticle.content.toString))
+      Ok(Json.obj("content" -> targetArticle.content.content.getOrElse("").toString))
```
leogrim, line 394: You want to serialize the entire article as a json here, not just the `content` field. 
Just do `Json.toJson(targetArticle)`. This works thanks to an implicit `Format` that defines how to serialize a specific type to json. It's usually defined in the companion object of the type so that the compiler can find it as an implicit - I think Scala For the Impatient has a good paragraph on Implicit resolution).

camhashemi, line 394: I see, that's good to know. Have you looked at the page itself? e.g. https://admin.kifi.com/admin/uri/6429465. There's probably a few improvements I can make for readability.

kifi-backend/modules/shoebox/app/views/admin/uri.scala.html
```diff
@@ -47,35 +47,31 @@
     </tr>
 
     <tr>
-        <th>Aggregate Content</th>
+        <th>Aggregate Content</th><td></td>
```
leogrim, line 50: @camhashemi I think you forgot the content here :)

leogrim:
A few comments:
- could you display each article info on one line with multiple columns?
- could you use `article` instead of `kind` in the path to actually get an article, in order to be consistent with the resource that is actually returned?

PR #15229, opened 2015-05-08
fixed/improved URI Admin page

camhashemi:
serializing targetArticle in UrlController, improving formatting in uri.scala.html, refactoring

### DIFF
--- a/kifi-backend/modules/shoebox/app/views/admin/uri.scala.html
+++ b/kifi-backend/modules/shoebox/app/views/admin/uri.scala.html
@@ -1,6 +1,6 @@
 @(uri: com.keepit.model.NormalizedURI, articleInfos: Set[com.keepit.rover.model.ArticleInfo], aggregateContent: com.keepit.rover.article.content.ArticleContentExtractor)(implicit request: com.keepit.common.controller.UserRequest[_])
 
-@admin(uri.title.getOrElse("Title Not Found")) {
+@admin(aggregateContent.title.getOrElse("Title Not Found")) {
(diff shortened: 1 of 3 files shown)


camhashemi:
@leogrim e.g. https://admin.kifi.com/admin/uri/153423. Let me know what you'd like to add/change. Normalization format will be fixed in the next commit and I'll remove admin/scraped once you give the thumbs up.

kifi-backend/modules/shoebox/app/views/admin/uri.scala.html
```diff
@@ -1,6 +1,6 @@
 @(uri: com.keepit.model.NormalizedURI, articleInfos: Set[com.keepit.rover.model.ArticleInfo], aggregateContent: com.keepit.rover.article.content.ArticleContentExtractor)(implicit request: com.keepit.common.controller.UserRequest[_])
 
-@admin(uri.title.getOrElse("Title Not Found")) {
+@admin(aggregateContent.title.getOrElse("Title Not Found")) {
```
leogrim, line 3: Can we just leave it empty? This page is about inspecting our data, so we want to return it as raw as possible.

leogrim:
:+1:

kifi-backend/modules/shoebox/app/views/admin/uri.scala.html
camhashemi, line 3: I can do that. What about the URI id? it's not on the page, only in the page URL.

PR #15233, opened 2015-05-08
mobile endpoint writeable libraries with url V2 change

ahsu1230:
@andrewconner @DerekBurch @mrbrown14

### DIFF
--- a/kifi-backend/modules/shoebox/test/com/keepit/controllers/mobile/MobileLibraryControllerTest.scala
+++ b/kifi-backend/modules/shoebox/test/com/keepit/controllers/mobile/MobileLibraryControllerTest.scala
@@ -463,7 +463,7 @@ class MobileLibraryControllerTest extends Specification with ShoeboxTestInjector
       }
     }
 
-    "get writeable libraries" in {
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/test/com/keepit/controllers/mobile/MobileLibraryControllerTest.scala
```diff
@@ -516,7 +516,49 @@ class MobileLibraryControllerTest extends Specification with ShoeboxTestInjector
...
+        (response3 \ "error").asOpt[String] === None
+        val keepData = (response3 \ "alreadyKept")
+        keepData === Json.parse(s"""
+            [
+             {
+                "id":"${k1.externalId}",
+                "title":"krustykrab",
+                "note":"[#food1] [#food2]",
+                "imageUrl":null,
+                "libraryId":"${pubLibId1.id}"
```
ahsu1230, line 556: v1 endpoint stays as is.
v2 endpoint puts the `libraryId` into the object. This object also doesn't have `hashtags`

PR #15234, opened 2015-05-08
collaborator libraries in profiles

ahsu1230:
libraries coming back should depend on the viewer
@andrewconner


ahsu1230:
seems like a couple of things are a bit messed up with retrieving profile libraries.
LibraryCommander, LibraryRepo, LibraryMembershipRepo are pretty messy, so will refactor Profile logic into a ProfileCommander or something.

andrewconner:
Sounds good

PR #15238, opened 2015-05-09
Switch search to new article index(er)

leogrim:
@eishay @stkem 
More work than I thought, especially to get the tests to pass: I found a bug / race condition in the way hits from others are counted. I think I know how to fix it but my brain is fried right now. I'll take another look over coffee tomorrow morning and will deploy / test it.


eishay:
:-)

stkem:
:+1:

PR #15239, opened 2015-05-10
dealing with possible race conditions in flaky tests

eishay:


eishay:
@andrewconner

andrewconner:
Ah, great

PR #15240, opened 2015-05-10
more test stability

eishay:
@andrewconner


eishay:
`keptAt` is by default current time which may cause problems in tests where few keeps will have a matching keptAt time so sorting won't be deterministic.

PR #15247, opened 2015-05-11
Original keeper added to `Keep`

andrewconner:
Will merge tonight, may cause a little bit of downtime. @ahsu1230


kifi-backend/modules/common/app/com/keepit/model/Keep.scala
```diff
@@ -142,7 +142,7 @@ object Keep {
         "inDisjointLib" -> k.inDisjointLib,
         "urlId" -> k.urlId,
         "url" -> k.url,
-        "bookmarkPath" -> k.bookmarkPath,
+        "bookmarkPath" -> (None: Option[String]),
```
ahsu1230, line 145: could you just remove this part if you're going to drop the `bookmarkPath` column?

ahsu1230:
is it necessary to have an admin endpoint backfill the `originalKeeperId` field for all current keeps?

andrewconner:
I'm actually going to just do it in SQL

ahsu1230:
:+1:

PR #15248, opened 2015-05-11
Adds UserName and external id to Admin Users page.

Colin-Lane:
@stkem @andrewconner


andrewconner:
Great! Looks good to me.

PR #15259, opened 2015-05-12
adding 'owner' field to OwnLibraryCard (now with collaborators)

ahsu1230:
@andrewconner @2is10

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/UserProfileCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/UserProfileCommander.scala
@@ -64,6 +64,7 @@ class UserProfileCommander @Inject() (
           slug = info.slug,
           kind = lib.kind,
           visibility = lib.visibility,
+          owner = info.owner,
           numKeeps = info.numKeeps,
           numFollowers = info.numFollowers,
           followers = info.followers,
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/UserProfileCommander.scala
```diff
@@ -64,6 +64,7 @@ class UserProfileCommander @Inject() (
           slug = info.slug,
           kind = lib.kind,
           visibility = lib.visibility,
+          owner = info.owner,
```
2is10, line 67: Since the receiving ends already support it, how about continuing to omit the owner in the common case, when it’s the profile user? Maybe:

``` scala
owner = Some(info.owner).filter(_.id != user.externalId)
```

2is10, line 67: May not be worth the complexity. It will gzip well, at least.

ahsu1230, line 67: Yea, I thought about doing something like that - making owner optional. But, heard mobile team want more consistent JSON responses from endpoints. So, I decided to just leave it as is this time.

andrewconner, line 67: Yeah, both mobile clients care a bit more about static models. Obviously wrappers can be used to infer fields, but it's easiest for them to see a `libraryCard` module and be able to treat it like that.

PR #15261, opened 2015-05-12
more clear logging on "Can't connect $socialUserInfo to user" problem…

eishay:
… and stop using suspected cache


eishay:
@andrewconner @stkem

PR #15262, opened 2015-05-12
Introduce RoverServiceClient.getOrElseFetchUriSummary

leogrim:
@stkem @andrewconner @eishay 

Make online calls to Rover possible, switch external messaging to Rover (behind experiment).


eishay:
yey

PR #15269, opened 2015-05-12
fix collaborators fields in LibraryCards

ahsu1230:
@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
@@ -1604,7 +1604,8 @@ class LibraryCommander @Inject() (
         val numFollowers = countMap.readOnly
         val numCollaborators = countMap.readWrite
 
-        val (collaborators, followers) = libraryMembershipRepo.pageWithLibraryIdAndAccess(lib.id.get, 0, 3, Set(LibraryAccess.READ_ONLY, LibraryAccess.READ_WRITE)).partition(mem => mem.canWrite)
+        val collaborators = libraryMembershipRepo.pageWithLibraryIdAndAccess(lib.id.get, 0, 3, Set(LibraryAccess.READ_WRITE, LibraryAccess.READ_INSERT))
+        val followers = libraryMembershipRepo.pageWithLibraryIdAndAccess(lib.id.get, 0, 3, Set(LibraryAccess.READ_ONLY))
         val collabIds = collaborators.map(_.userId).toSet
         val followerIds = followers.map(_.userId).toSet
         val userSample = basicUserRepo.loadAll(followerIds ++ collabIds) //we don't care about the order now anyway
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -1604,7 +1604,8 @@ class LibraryCommander @Inject() (
         val numFollowers = countMap.readOnly
         val numCollaborators = countMap.readWrite
 
-        val (collaborators, followers) = libraryMembershipRepo.pageWithLibraryIdAndAccess(lib.id.get, 0, 3, Set(LibraryAccess.READ_ONLY, LibraryAccess.READ_WRITE)).partition(mem => mem.canWrite)
+        val collaborators = libraryMembershipRepo.pageWithLibraryIdAndAccess(lib.id.get, 0, 3, Set(LibraryAccess.READ_WRITE, LibraryAccess.READ_INSERT))
+        val followers = libraryMembershipRepo.pageWithLibraryIdAndAccess(lib.id.get, 0, 3, Set(LibraryAccess.READ_ONLY))
```
ahsu1230, line 1608: the pagination was done incorrectly before. So we'll have to look in LibraryMembership for both collab access, then again for follower access. Since the FE design separates them

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryRepo.scala
```diff
@@ -285,13 +285,13 @@ class LibraryRepoImpl @Inject() (
 
   def getOwnerLibrariesForOtherUser(userId: Id[User], friendId: Id[User], page: Paginator)(implicit session: RSession): Seq[Library] = {
     import com.keepit.common.db.slick.StaticQueryFixed.interpolation
-    val libsFriendFollow = sql"select lib.id from library_membership lm, library lib where lm.library_id = lib.id and lib.owner_id = $userId and lm.user_id = $friendId and lib.state = 'active' and lm.state = 'active'".as[Id[Library]].list
+    val libsFriendFollow = sql"select lib.id from library_membership lm, library lib where lm.library_id = lib.id and lm.user_id = $friendId and lib.state = 'active' and lm.state = 'active'".as[Id[Library]].list
```
ahsu1230: Right now, I can see Stephen's `Collaborative` library on Stephen's profile and my profile. However, I can't see it on Andrew's profile (even though he's a collaborator). So we can't make the assumption that the library owner is `userId` for collaborators.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
ahsu1230, line 1608: Trying to figure out a better way to do this - not having to look through the membership repo twice. It seems `pageWithLibraryIdAndAccess` is used elsewhere similarly - to page Followers, then to page Collaborators

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryRepo.scala
ahsu1230: Problem with this is that this `libsFriendFollow` can be a pretty big list =/

PR #15271, opened 2015-05-12
Adds validation to updating a username

Colin-Lane:
@stkem @andrewconner


andrewconner:
That admin action actually already exists.

`AdminUserController.setUsername(id: Id[User], username: String, overrideValidityCheck: Boolean ?= false, overrideProtection: Boolean ?= false)`

`overrideValidityCheck` means that we can set someone's username to something forbidden, like "kifi" (which isn't allowed normally). For admins, this can be true (we just need to be careful). `overrideProtection` can safely stay false for now.

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminUserController.scala
```diff
@@ -432,6 +432,22 @@ class AdminUserController @Inject() (
     }
   }
 
+  def retrieveFormFromRequest(request: UserRequest[AnyContent]): Map[String, String] = request.request.body.asFormUrlEncoded match {
+    case Some(request) => request.map(entry => (entry._1 -> entry._2.head))
+    case None => throw new Exception("invalid form entry.")
```
andrewconner: In Scala, we typically don't use exceptions for control flow. [Play actually has Form handling built in](https://www.playframework.com/documentation/2.3.x/ScalaForms), but if you were to implement yourself, returning an empty Map would be just fine (ie, there was no form body).

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminUserController.scala
```diff
@@ -432,6 +432,22 @@ class AdminUserController @Inject() (
     }
   }
 
+  def retrieveFormFromRequest(request: UserRequest[AnyContent]): Map[String, String] = request.request.body.asFormUrlEncoded match {
+    case Some(request) => request.map(entry => (entry._1 -> entry._2.head))
```
andrewconner: Also, use of `.head` is typically discouraged, because it'll throw an exception when there is an empty list. `headOption` is much preferred. In this case, if you don't care about any keys without values, you could just not include them in the Map.

andrewconner:
Here's an idea that's safe to bad inputs, and shows a bit how Scala typically handles optional values:

``` scala
request.body.asFormUrlEncoded.flatMap(_.get("username").flatMap(_.headOption)).map { newUsername =>
  userCommander.setUsername(userId, Username(newUserName.trim)) match {
    case Right(_) => Ok
    case Left(err) => BadRequest(err)
  }
}.getOrElse(BadRequest("No username provided"))
```

Colin-Lane:
Taking your comments into consideration, I have made a few changes.
it is now using the setUsername route, as well as setting overrideValidityCheck to true for the admin page.
Also no user input is still being handled.

andrewconner:
Looks great! Thanks

PR #15276, opened 2015-05-12
members endpoint includes owner

ahsu1230:
@andrewconner @DerekBurch @jaredpetker @thassss @mrbrown14 

Go ahead and start using 
`GET /m/2/libraries/:id/members` instead of `GET /m/1/libraries/:id/members` so owners (then collaborators, then followers) to show up!


andrewconner:
Looks good

PR #15281, opened 2015-05-12
Basically RIP LinkedIn

andrewconner:
@camhashemi Good catch. I knew they were cutting us off, didn't know how soon.


camhashemi:
it's the end of an era...

PR #15286, opened 2015-05-12
Use Rover for Eliza's notifications

leogrim:


kifi-backend/modules/eliza/app/com/keepit/eliza/commanders/NotificationJsonMaker.scala
```diff
@@ -25,33 +20,52 @@ case class NotificationJson(obj: JsObject) extends AnyVal
...
+        val futureUriSummaryJson = futureUriSummary.imap(_.map { summary =>
+          val image = summary.images.get(idealImageSize)
+          Json.obj(
+            "title" -> summary.article.title,
+            "description" -> summary.article.description,
+            "imageUrl" -> image.map(_.path.getUrl),
+            "imageWidth" -> image.map(_.size.width),
+            "imageHeight" -> image.map(_.size.height)
+          )
+        })
```
leogrim, line 59: @stkem @andrewconner 
Rover can now supply images only (instead of full uri summary), not using it in Eliza for now because I don't now if the title and description of the page are actually used.

PR #15293, opened 2015-05-13
Bypass transactional caching when recovering from an exception (since…

leogrim:
… the session is being rolled back)

@eishay @andrewconner @stkem likely never worked :)


andrewconner:
Nice catch @leogrim

PR #15299, opened 2015-05-13
`user` on keeps

andrewconner:


andrewconner:
@2is10

andrewconner:
To recap chat:

```
Jared Jacobs [12:44 PM] 
@ac: maybe `keeper` instead of `user`? It’s more specific and who knows, there may be more than one user associated with a keep, in various roles, in the future.

Andrew Conner [12:51 PM] 
Valid. However, right now, it’s closer semantically to “note author”, which doesn’t really exist either.

Tommy Hass [12:51 PM] 
so why not embed the user object inside the note?

Andrew Conner [12:52 PM] 
We could, but it doesn’t necessarily require the existence of a note

Andrew Conner [12:52 PM]
The contents of `user` is what’s shown on the keep card as the user responsible

Andrew Conner [1:08 PM] 
Everyone okay with it as proposed, or more debate needed?

Tommy Hass [1:09 PM] 
im all good

Andrew Conner [1:10 PM] 
Agreed that notes wasn’t built right. Hacked in to give ability, and then whoops, we’re using it in production clients. We may split it to a separate table or something, where note owner makes sense.
```

2is10:
okay, `user` it is

PR #15301, opened 2015-05-13
More logging in ArticleCommander.fetchAsap

leogrim:


kifi-backend/modules/rover/app/com/keepit/rover/article/ArticleCommander.scala
```diff
@@ -94,7 +93,7 @@ class ArticleCommander @Inject() (
         markAsFetching(info.id.get)
         fetchAndPersist(info)
       }
-      case fetchedArticleOpt: Option[info.A] => Future.successful(fetchedArticleOpt)
+      case fetchedArticleOpt => Future.successful(fetchedArticleOpt)
```
leogrim, line 96: @andrewconner this is what Scalariform didn't like.

andrewconner, line 96: Ah, weird

PR #15303, opened 2015-05-13
Create and transfer library for admins

Colin-Lane:
When viewing a library from admin/libraries/:id if the parameter transfer=true is specified a library’s owner can be changed to a new user with their user id.

When viewing libraries from admin/user/:id/libraries a new library can be created by giving a library name, a slug, and a visibility level.

@stkem @andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/admin/AdminLibraryController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/admin/AdminLibraryController.scala
@@ -55,51 +55,46 @@ class AdminLibraryController @Inject() (
     suggestedSearchCommander: LibrarySuggestedSearchCommander,
     implicit val publicIdConfig: PublicIdConfiguration) extends AdminUserActions {
 
-  def updateLibraryOwner(libraryId: Id[Library], fromUserId: Id[User]) = AdminUserPage { implicit request =>
-    request.body.asFormUrlEncoded.flatMap(_.get("toUserId").flatMap(_.headOption)).map(s => Id[User](s.toLong)) match {
-      case None => BadRequest("No User Id given")
-      case Some(toUserId) => {
-        db.readWrite { implicit session =>
-          val lib = libraryRepo.get(libraryId)
-          if (lib.ownerId != fromUserId) throw new Exception(s"orig user $fromUserId is not matching current library owner $lib")
-          libraryAliasRepo.alias(lib.ownerId, lib.slug, lib.id.get)
-          libraryAliasRepo.reclaim(toUserId, lib.slug) // reclaim existing alias to a former library of toUserId with the same slug
-          val newOwnerLib = lib.copy(ownerId = toUserId)
-          libraryRepo.save(newOwnerLib)
-          val currentOwnership = libraryMembershipRepo.getWithLibraryIdAndUserId(libraryId, fromUserId).getOrElse(throw new Exception(s"no ownership to lib $lib for user $fromUserId"))
-          libraryMembershipRepo.save(currentOwnership.copy(access = LibraryAccess.READ_ONLY))
-          libraryMembershipRepo.getWithLibraryIdAndUserId(libraryId, toUserId) match {
-            case None =>
-              libraryMembershipRepo.save(currentOwnership.copy(id = None, userId = toUserId))
-            case Some(newOwnership) =>
-              libraryMembershipRepo.save(newOwnership.copy(access = LibraryAccess.OWNER, showInSearch = currentOwnership.showInSearch))
-          }
-          var page = 0
-          val pageSize = 100
-          var hasMore = true
-          val keeps = ArrayBuffer[Keep]()
-          while (hasMore) {
-            val from = page * pageSize
-            val chunk: Seq[Keep] = keepRepo.getByLibrary(libraryId, from, from + pageSize, Set.empty) map { keep =>
-              val tags = keepToCollectionRepo.getByKeep(keep.id.get)
-              tags foreach { keepToTag =>
-                val origTag = collectionRepo.get(keepToTag.collectionId)
-                val tag = collectionRepo.getByUserAndName(toUserId, origTag.name, excludeState = None) match {
-                  case None => collectionRepo.save(Collection(userId = toUserId, name = origTag.name))
-                  case Some(existing) if !existing.isActive => collectionRepo.save(existing.copy(state = CollectionStates.ACTIVE, name = origTag.name))
-                  case Some(existing) => existing
-                }
-                keepToCollectionRepo.save(keepToTag.copy(collectionId = tag.id.get))
-              }
-              keepRepo.save(keep.copy(userId = toUserId))
-            }
-            hasMore = chunk.size >= pageSize
-            keeps.appendAll(chunk)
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/shoebox/app/com/keepit/admin/AdminLibraryController.scala
```diff
@@ -55,46 +55,51 @@ class AdminLibraryController @Inject() (
...
-        val chunk: Seq[Keep] = keepRepo.getByLibrary(libraryId, from, from + pageSize, Set.empty) map { keep =>
-          val tags = keepToCollectionRepo.getByKeep(keep.id.get)
-          tags foreach { keepToTag =>
-            val origTag = collectionRepo.get(keepToTag.collectionId)
-            val tag = collectionRepo.getByUserAndName(toUserId, origTag.name, excludeState = None) match {
-              case None => collectionRepo.save(Collection(userId = toUserId, name = origTag.name))
-              case Some(existing) if !existing.isActive => collectionRepo.save(existing.copy(state = CollectionStates.ACTIVE, name = origTag.name))
-              case Some(existing) => existing
+  def updateLibraryOwner(libraryId: Id[Library], fromUserId: Id[User]) = AdminUserPage { implicit request =>
+    request.body.asFormUrlEncoded.flatMap(_.get("toUserId").flatMap(_.headOption)).map(s => Id[User](s.toLong)) match {
```
stkem: Is there a particular reason that you moved the `toUserId` parameter from a query parameter to the body? It would be nice to keep the existing easily curl-able API. Instead of submitting a form with a body you an just grab the `toUserId` out of the form with javascript and put it in the url in javascript (i.e. you can't use the reverse router for that part).

stkem:
made one comment. Rest lgtm! Nice work.

kifi-backend/modules/shoebox/app/views/admin/library.scala.html
```diff
@@ -112,6 +124,18 @@
     document.getElementById("edit_tc").addEventListener("click", startEdit, true);
     disableEdit(true)
 
+    @if(transfer) {
+      $('.transfer-ownership').on("submit", function() {
+        $.post("@com.keepit.controllers.admin.routes.AdminLibraryController.updateLibraryOwner(library.id.get, library.ownerId)", $('.transfer-ownership').serialize())
+        .done(function(data) {
+          alert("Library Owner Updated")
+        }).fail(function(data) {
+          alert("Bad Request: " + data.responseText);
```
andrewconner: Just so you know, this case will almost always be the one that's run for large libraries because of timeouts. It doesn't mean it failed though — transferring large libraries can take several minutes.

kifi-backend/modules/shoebox/app/views/admin/library.scala.html
```diff
@@ -112,6 +124,18 @@
     document.getElementById("edit_tc").addEventListener("click", startEdit, true);
     disableEdit(true)
 
+    @if(transfer) {
+      $('.transfer-ownership').on("submit", function() {
```
andrewconner, line 129: I would disable submitting again until a response comes back, because running two migrations at once ends up causing some weird side effects (also, it's a lot slower because of row locks).

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminLibraryController.scala
```diff
@@ -55,46 +55,51 @@ class AdminLibraryController @Inject() (
...
+          libraryMembershipRepo.getWithLibraryIdAndUserId(libraryId, toUserId) match {
+            case None =>
+              libraryMembershipRepo.save(currentOwnership.copy(id = None, userId = toUserId))
+            case Some(newOwnership) =>
+              libraryMembershipRepo.save(newOwnership.copy(access = LibraryAccess.OWNER, showInSearch = currentOwnership.showInSearch))
+          }
+          var page = 0
+          val pageSize = 100
+          var hasMore = true
+          val keeps = ArrayBuffer[Keep]()
```
andrewconner: I'm not sure the history of this, and obviously predates you, but since we're not doing anything with this arraybuffer (besides getting its size), and knowing this function is already pretty slow, how about just keeping track of the number of keeps moved?

kifi-backend/modules/shoebox/app/views/admin/userLibraries.scala.html
```diff
@@ -13,6 +13,20 @@
...
     </table>
 
+    <div>
+        <p>Create a Library</p>
+        <form action="#" method="POST" class="create-library">
+            <input type="text" name="name" placeholder="Library Name">
+            <input type="text" name="slug" placeholder="slug">
+            <select name="visibility">
+                <option value="published">Published</option>
+                <option value="discoverable">Discoverable</option>
```
andrewconner: We shouldn't actually surface this option. Discoverable is a special permission that the Main library is in, but shouldn't be used elsewhere (the UI doesn't handle this).

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminUserController.scala
```diff
@@ -409,6 +411,26 @@ class AdminUserController @Inject() (
     userStatisticsPage(page, FakeUsersViewType).map { p => Ok(html.admin.users(p, None)) }
   }
 
+  def createLibrary(userId: Id[User]) = AdminUserPage(parse.tolerantFormUrlEncoded) { implicit request =>
+    val name = request.body.get("name").flatMap(_.headOption)
+    val visibility = request.body.get("visibility").flatMap(_.headOption).map(LibraryVisibility(_))
+    val slug = request.body.get("slug").flatMap(_.headOption)
+
+    (name, visibility, slug) match {
+      case (Some(name), Some(visibility), Some(slug)) => {
```
andrewconner, line 420: This is acceptable, but typically when you have several `Option`, `Seq`, etc variables that you want to get to, you can use a for-comprehension:

``` scala
for {
  name <- nameOpt
  visibility <- visibilityOpt
  slug <- slugOpt
} yield {
  // now you have access to name, visibility, and slug
}
```

Again, for this case, matching this way is fine, but you'll see for comprehensions around the place. Looks like you're picking up syntax fast, nice job!

andrewconner, line 420: Also, in this specific case, we try to avoid shadowing variables because it can make the code harder to read. In this block, `name` is a `String` from line 420. However, outside this block, `name` is an `Option[String` from line 415. It's valid Scala to shadow, but we try to stay away. A common pattern is calling the outer one `nameOpt` or something.

kifi-backend/modules/shoebox/app/views/admin/library.scala.html
andrewconner: So, it may be useful to say that the process is running, and please do not resubmit :). It's not perfect (if a server restarts, progress is in a weird state and/or lost), but may help a non-technical admin from submitting it many times because they keep getting errors. I've seen one very large one take 10 minutes.

Colin-Lane:
Updated the code based on the comments above.

stkem:
:+1: Let's get it out.

PR #15311, opened 2015-05-14
change inviteToCollab to whoCanInvite

ahsu1230:
considering changing this up while it's easy to
@andrewconner

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/model/LibraryInvite.scala
+++ b/kifi-backend/modules/common/app/com/keepit/model/LibraryInvite.scala
@@ -74,6 +74,23 @@ object LibraryInvite extends ModelWithPublicIdCompanion[LibraryInvite] {
   }
 }
 
+sealed abstract class LibraryInvitePermissions(val value: String)
+
+object LibraryInvitePermissions {
+  case object COLLABORATORS_ALLOWED extends LibraryInvitePermissions("collaborators_allowed")
+  case object OWNER_ONLY extends LibraryInvitePermissions("owner_only")
+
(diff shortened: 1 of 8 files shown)


kifi-backend/modules/common/app/com/keepit/model/LibraryInvite.scala
```diff
@@ -74,6 +74,23 @@ object LibraryInvite extends ModelWithPublicIdCompanion[LibraryInvite] {
...
+  case object COLLABORATORS_ALLOWED extends LibraryInvitePermissions("collaborators_allowed")
+  case object OWNER_ONLY extends LibraryInvitePermissions("owner_only")
+
+  implicit def format[T]: Format[LibraryInvitePermissions] =
+    Format(__.read[String].map(LibraryInvitePermissions(_)), new Writes[LibraryInvitePermissions] { def writes(o: LibraryInvitePermissions) = JsString(o.value) })
+
+  def apply(str: String): LibraryInvitePermissions = {
+    str match {
+      case COLLABORATORS_ALLOWED.value | "read_write" => COLLABORATORS_ALLOWED
+      case OWNER_ONLY.value | "owner" => OWNER_ONLY
```
ahsu1230, line 89: With this PR, we'll be persisting `owner_only` or `collaborators_allowed`
right now `owner` and `read_write` are in the database, this way we can still read the old values and support new values
once the old values are changed respectively, we can get rid of `owner` and `read_write`

andrewconner, line 89: Very nice handling old data

andrewconner:
:+1:

PR #15312, opened 2015-05-14
checking for case where user has no social user info (happens now in …

eishay:
…prod) and throwing contextual exception.

Also adding twitter SUI as a priority for image load after FB and before LI


eishay:
@Colin-Lane @andrewconner @stkem @ahsu1230

PR #15313, opened 2015-05-14
auto subscribe if collaborator

ahsu1230:
should be like this, I think
@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
@@ -1203,13 +1203,15 @@ class LibraryCommander @Inject() (
       val updatedLib = db.readWrite(attempts = 3) { implicit s =>
         libraryMembershipRepo.getWithLibraryIdAndUserId(libraryId, userId, None) match {
           case None =>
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -1203,13 +1203,15 @@ class LibraryCommander @Inject() (
       val updatedLib = db.readWrite(attempts = 3) { implicit s =>
         libraryMembershipRepo.getWithLibraryIdAndUserId(libraryId, userId, None) match {
           case None =>
-            libraryMembershipRepo.save(LibraryMembership(libraryId = libraryId, userId = userId, access = maxAccess, lastJoinedAt = Some(currentDateTime)))
+            val subscribedToUpdates = if (maxAccess == LibraryAccess.READ_WRITE) true else false
+            libraryMembershipRepo.save(LibraryMembership(libraryId = libraryId, userId = userId, access = maxAccess, lastJoinedAt = Some(currentDateTime), subscribedToUpdates = subscribedToUpdates))
```
ahsu1230, line 1207: `libraryCommander.updatedLibraryUpdateSubscription` looks for the `LibraryMembership`, but we don't need to call the commander function because we already know the `LibraryMembership` at this point. We can just go ahead and just edit.

andrewconner:
Works for me, nice job

PR #15314, opened 2015-05-14
fixed some bugs, added some tests for "plain" emails

camhashemi:
tests make sure that no Scala syntax is left in the emails


camhashemi:
@stkem I forgot to enable production testing for library invites

PR #15315, opened 2015-05-14
covering more possible places where social id could be an empty string

eishay:


eishay:
@andrewconner

andrewconner:
Fingers crossed

PR #15322, opened 2015-05-14
remove passphrase

ahsu1230:
@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
@@ -238,18 +238,12 @@ class LibraryController @Inject() (
       case Failure(ex) =>
         BadRequest(Json.obj("error" -> "invalid_id"))
       case Success(libId) =>
-        val hashedPassPhrase = if (authToken.isDefined) {
-          val existingCookieFields = request.session.get("library_access").flatMap(libraryCommander.getLibraryIdAndPassPhraseFromCookie)
-          existingCookieFields.map(_._2)
-        } else {
-          None
-        }
         implicit val context = heimdalContextBuilder.withRequestInfoAndSource(request, KeepSource.site).build
         db.readOnlyMaster { implicit s =>
           libraryMembershipRepo.getWithLibraryIdAndUserId(libId, request.userId)
         } match {
           case None =>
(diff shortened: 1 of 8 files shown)


kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
```diff
@@ -464,32 +458,11 @@ class LibraryController @Inject() (
...
-          val unhashedPassPhraseOpt = (request.body \ "passPhrase").asOpt[String]
-          unhashedPassPhraseOpt.map { unhashedPassPhrase =>
-            val passPhrase = HashedPassPhrase.generateHashedPhrase(unhashedPassPhrase)
-            if (libraryCommander.canViewLibrary(request.userIdOpt, library, authToken, Some(passPhrase))) {
-              val cookie = ("library_access", s"${Library.publicId(library.id.get).id}/${passPhrase.value}")
-              NoContent.addingToSession(cookie)
-            } else {
-              BadRequest(Json.obj("error" -> "invalid_access"))
-            }
-          }.getOrElse(BadRequest(Json.obj("error" -> "no_passphrase_provided")))
```
ahsu1230, line 492: no need to consider cookie `library_access` or `passPhrase`
just check if the authToken matches!

kifi-backend/modules/shoebox/app/com/keepit/shoebox/controllers/LibraryAccessActions.scala
```diff
@@ -37,8 +37,8 @@ trait LibraryAccessActions {
 
   private def lookupViewAccess[A](libraryPubId: PublicId[Library], input: MaybeUserRequest[A]) = {
     parseRequest(libraryPubId, input) match {
-      case Some((libraryId, userIdOpt, accessToken, hashedPassPhrase)) =>
-        val access = libraryCommander.canViewLibrary(userIdOpt, libraryId, accessToken, hashedPassPhrase)
+      case Some((libraryId, userIdOpt, accessToken)) =>
+        val access = libraryCommander.canViewLibrary(userIdOpt, libraryId, accessToken)
```
ahsu1230, line 41: this is where the logic that actually determines whether you can view the current library or not. Now that we don't consider `hashedPassPhrase` in `canViewLibrary()`, this action should allow the user to view the library, so we should be skipping the "Enter the passphrase" part on the website.

PR #15323, opened 2015-05-14
removed passPhrase from library invite emails - deprecated

camhashemi:
option.get was not called


camhashemi:
@stkem this should fix that passphrase bug with the old email. we were requiring users to enter a passphrase without showing them what it was.

p.s. don't merge yet. I'm going to roll out the new welcome email per Sterling's go-ahead.

stkem:
:+1:

camhashemi:
@andrewconner removed all passphrase components from library invite emails. I've confirmed that all extra stuff is already in kifi/keepit.

PR #15325, opened 2015-05-14
[WIP] heimdal library view tracks

yingjieMiao:
@stkem @eishay 

added `VIEWED_LIBRARY` user event type. Shoebox is not triggering this event type yet. 

This PR gives you a preview of where I am heading.


kifi-backend/modules/heimdal/app/com/keepit/controllers/EventTrackingController.scala
```diff
@@ -15,6 +16,10 @@ import scala.concurrent.ExecutionContext
 import scala.concurrent.duration._
 import scala.util.{ Failure, Success }
 
+trait UserEventHandler {
```
stkem, line 19: Nice new abstraction!

stkem:
Good Direction!

yingjieMiao:
@stkem @eishay 

added `VisitorEventTypes.VIEWED_LIBRARY`. This completes the tracking on Heimdal side, which makes this PR self-contained. 

I would like to open a different PR on shoebox side, really send down these events from `LibraryAnalytics`

stkem:
I would prefer one PR with both parts if possible. Makes it easier to see what is going on. This is self contained code wise, but has no purpose without the other part (and it's not so long as to warrant a split PR)

yingjieMiao:
@stkem   shoebox tracking added.

thank @ahsu1230 for pointing the code path

stkem:
:+1:

PR #15331, opened 2015-05-14
I believe this fixes the off-by-one problem

andrewconner:
@ahsu1230 You may be able to double check my logic.


ahsu1230:
think it works!

PR #15335, opened 2015-05-14
library attribution tweaks

2is10:


kifi-backend/modules/shoebox/angular/src/keep/keepCard.js
```diff
@@ -119,6 +119,7 @@ angular.module('kifi')
               }
             });
           }
+          _.remove(keep.keepers, {pictureName: '0.jpg'});
```
2is10, line 122: seeing our default user picture in attribution is a bad experience

PR #15338, opened 2015-05-15
requesting smaller keep images for search suggestion thumbnails

2is10:
@leogrim Thanks!


kifi-backend/modules/shoebox/angular/src/header/searchSuggest.styl
```diff
@@ -53,7 +53,6 @@ uriHeight = 48px
 
 .kf-ssg-uri-thumb
   position: absolute
-  sidePad = 10px
```
2is10, line 56: a redundant definition

PR #15339, opened 2015-05-15
Kifi collaborators 

ahsu1230:
@2is10 @andrewconner 

In this PR:
- library headers have collaborators
- collab experiment guard for collaborator invite
- library subscription notifications (top left of library header)


kifi-backend/modules/shoebox/angular/src/libraries/libraryHeader.tpl.html
```diff
@@ -52,18 +52,46 @@ <h1 class="kf-lh-name" ng-class="{'kf-long-1': library.name.length > 24, 'kf-lon
...
+      <!--  owner with collaborators -->
+      <div class="kf-lh-owners" ng-if="library.numCollaborators > 0">
+        <div class="kf-lh-owner-image-wrap collab">
+          <a class="kf-lh-owner-image-a" href="{{library.owner|profileUrl}}" itemprop="url" kf-track-origin="libraryPage/curator">
+            <img class="kf-lh-owner-image" ng-src="{{library.owner|pic:200}}" itemprop="image" alt="{{library.owner.firstName}}">
+          </a>
+          <a class="kf-lh-owner-name collab" href="{{library.owner|profileUrl}}" itemprop="name" kf-track-origin="libraryPage/curator">{{library.owner.firstName}}</a>
+        </div>
+        <div class="kf-lh-owner-image-wrap collab" ng-repeat="collab in library.collaborators">
+          <a class="kf-lh-owner-image-a" href="{{collab|profileUrl}}" itemprop="url" kf-track-origin="libraryPage/curator">
```
andrewconner: Should tracking be `libraryPage/curator`? Is this used elsewhere, or just here? If just here, how about `libraryPage/collaborator`?

ahsu1230: just in this file, I'll change things to be `libraryPage/collaborator`

kifi-backend/modules/shoebox/angular/src/libraries/libraryHeader.styl
```diff
@@ -555,6 +566,44 @@ ownerImageDiam2 = 3.6em
...
+  border-radius: 50%
+  vertical-align: top
+
+  &:hover
+    background: rgba(255, 255, 255, 0.3)
+
+  &.next-to-owner
+    position: absolute
+    top: 0px
+    left: calc(50% + 35px)
```
2is10, line 692: Personally, I think this button should push the owner picture left, so the two circles are centered. Or else it should be much fainter, at least until hovered. I’ll take a look at it and propose something.

kifi-backend/modules/shoebox/angular/src/libraries/libraryHeader.tpl.html
```diff
@@ -52,18 +52,46 @@ <h1 class="kf-lh-name" ng-class="{'kf-long-1': library.name.length > 24, 'kf-lon
...
+          </button>
+        </div>
         <div class="kf-lh-owner-name-wrap">
           <a class="kf-lh-owner-name" href="{{library.owner|profileUrl}}" itemprop="name" kf-track-origin="libraryPage/curator">{{library.owner|name}}</a>
           <a class="kf-lh-owner-twitter-a svg-twitter-blue" href="https://www.twitter.com/{{library.attr.twitter.screenName}}" target="_blank" title="@{{library.attr.twitter.screenName}}" ng-if="library.attr.twitter.screenName" ng-click="trackTwitterProfile()"></a>
         </div>
       </div>
+
+      <!--  owner with collaborators -->
+      <div class="kf-lh-owners" ng-if="library.numCollaborators > 0">
```
2is10: I try to avoid using the same identifying class on different elements in templates. It makes styling more difficult, because it’s easy to forget that the class also applies to some other doppelgänger element in an alternate ~~universe~~ code path. I suspect this can be refactored without having an `ng-if` at quite this high level in the template, and thus without duplicating the `kf-lh-owners` div and perhaps other elements like the primary owner’s picture. If you disagree (which may be the case, and for good reason, since you’ve thought more about this than I have), then I’d urge you to avoid reusing CSS classes between the `ng-if` block above and this one, so that changing styles for one won’t affect the other accidentally.

kifi-backend/modules/shoebox/angular/src/libraries/libraryHeader.tpl.html
```diff
@@ -52,18 +52,46 @@ <h1 class="kf-lh-name" ng-class="{'kf-long-1': library.name.length > 24, 'kf-lon
...
+          <a class="kf-lh-owner-image-a" href="{{library.owner|profileUrl}}" itemprop="url" kf-track-origin="libraryPage/curator">
+            <img class="kf-lh-owner-image" ng-src="{{library.owner|pic:200}}" itemprop="image" alt="{{library.owner.firstName}}">
+          </a>
+          <a class="kf-lh-owner-name collab" href="{{library.owner|profileUrl}}" itemprop="name" kf-track-origin="libraryPage/curator">{{library.owner.firstName}}</a>
+        </div>
+        <div class="kf-lh-owner-image-wrap collab" ng-repeat="collab in library.collaborators">
+          <a class="kf-lh-owner-image-a" href="{{collab|profileUrl}}" itemprop="url" kf-track-origin="libraryPage/collaborator">
+            <img class="kf-lh-owner-image" ng-src="{{collab|pic:200}}" itemprop="image" alt="{{collab|name}}">
+          </a>
+          <a class="kf-lh-owner-name collab" href="{{collab|profileUrl}}" itemprop="name" kf-track-origin="libraryPage/collaborator">{{collab.firstName}}</a>
```
2is10: Here, for instance, I’d use a different identifying class instead of a modifier, maybe `kf-lh-collab-name`. To the extent that you want this name to look like a solitary owner’s name, you can use shared style blocks in the stylesheet (a block with multiple selectors) or Stylus constants.

2is10:
Great start!

PR #15343, opened 2015-05-15
migrate to hosted statsd

joshm1:
@stkem @andrewconner  from what I understand reading the docs and the source code for the StatsD library (https://github.com/typesafehub/play-plugins/tree/6ac0e80bb35dd21b5f0b1b1e2f00f3675d7a8585/statsd/src/main/scala/play/modules/statsd/api), this is the place to put this. 

there is some risk to this PR because I haven't been able to test it locally. I tried to enable statsd in application-dev.conf but haven't seen any dev-mode stats added (might just be that it didn't execute any code that sent to statsd).

we'd need to keep an eye on graphite to make sure the metrics are flowing in after a deploy

### DIFF
--- a/kifi-backend/modules/common/conf/application-dev.conf
+++ b/kifi-backend/modules/common/conf/application-dev.conf
@@ -183,6 +183,7 @@ elasticache.config.endpoint="localhost:11211"
 # ~~~~~~~~~~~~~~~~~~
 statsd {
   enabled = false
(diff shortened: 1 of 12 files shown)


stkem:
I say we just roll it out and see what happens. If we loose statsd for a little bit it's really not a big deal. It's happened many times before due to the process on marvin dying or network issues.
Any objections?

joshm1:
sounds good to me

PR #15344, opened 2015-05-15
Easier than the eye sore?

andrewconner:
@stkem Better? Useful whenever interfacing with Java libs that have habits of throwing exceptions and returning null.


stkem:
:+1:

PR #15346, opened 2015-05-15
notifications in header

ahsu1230:
@2is10 @andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/common/services/routeService.js
+++ b/kifi-backend/modules/shoebox/angular/src/common/services/routeService.js
@@ -203,6 +203,9 @@ angular.module('kifi')
       deleteLibrary: function (libraryId) {
         return route('/libraries/' + libraryId + '/delete');
       },
+      updateSubscriptionToLibrary: function (libraryId, subscribed) {
+        return route('/libraries/' + libraryId + '/subscription?subscribed=' + subscribed, {});
+      },
       uploadLibraryCoverImage: function (libraryId, x, y, idealSize) {
         return route('/libraries/' + libraryId + '/image/upload', {x: x, y: y, is: idealSize || []});
       },
--- a/kifi-backend/modules/shoebox/angular/src/libraries/libraryHeader.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/libraryHeader.js
@@ -94,6 +94,12 @@ angular.module('kifi')
           }
         };
 
+        scope.changeSubscription = function () {
+          libraryService.updateSubscriptionToLibrary(scope.library.id, !(scope.library.subscribed)).then(function(res) {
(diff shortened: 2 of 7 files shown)


kifi-backend/modules/shoebox/angular/src/common/services/routeService.js
```diff
@@ -203,6 +203,9 @@ angular.module('kifi')
       deleteLibrary: function (libraryId) {
         return route('/libraries/' + libraryId + '/delete');
       },
+      updateSubscriptionToLibrary: function (libraryId, subscribed) {
+        return route('/libraries/' + libraryId + '/subscription?subscribed=' + subscribed, {});
```
2is10, line 207:
``` js
route('/libraries/' + libraryId + '/subscription', {subscribed: subscribed})
```

kifi-backend/modules/shoebox/angular/src/libraries/libraryHeader.js
```diff
@@ -94,6 +98,12 @@ angular.module('kifi')
           }
         };
 
+        scope.changeSubscription = function () {
+          libraryService.updateSubscriptionToLibrary(scope.library.id, !(scope.library.subscribed)).then(function(res) {
+            scope.library.subscribed = !(scope.library.subscribed)
```
2is10, line 103: missing semicolon

kifi-backend/modules/shoebox/angular/src/libraries/libraryHeader.styl
```diff
@@ -241,9 +241,52 @@ followerPicDiam = 2.2em
...
+    background-color: #fff
+    font-size: 13px
+    color: #3E4955
+    opacity: 0
+    visibility: hidden
+    transition: opacity linear 0.1s, visibility linear 0.1s
+
+    &.enabled::before
+      content: 'Notifications enabled'
+    &.disabled::before
```
2is10, line 277: we’ve run into occasional problems using unprefixed class names like `enabled` and `disabled`. let’s please use `kf-enabled` and `kf-disabled`. thanks.

kifi-backend/modules/shoebox/angular/src/libraries/libraryHeader.styl
```diff
@@ -241,9 +241,52 @@ followerPicDiam = 2.2em
...
+  background: none
+  border: none
+  opacity: 0.9
+
+  &.for-owner
+    margin-left: 0.9em
+
+  .kf-lh-subscribed-icon
+    width: 24px
+    height: 28px
```
2is10, line 257: The library header uses `em` for nearly all measurements because it uses its font size as the key for its responsive layout. Please be sure to test several viewport widths down to, say, 420px or 480px when working on the library header. Thanks.

PR #15350, opened 2015-05-15
Revert "put statsd stats for Hosted StatsD to have the same keys as b…

joshm1:
Reverts kifi/keepit#15349


joshm1:
this put stats under "counters.stats.curator" instead of "stats.counters.curator" (which is where existing stats are)

PR #15364, opened 2015-05-15
Can modify cache as admin

Colin-Lane:
on page /admin/cache
- added 3 new forms to allow for get/delete/set key/value pairs in
  the cache

Looks like remove and set from the cache return Unit, so they only have success scenarios.
I also didn't think limiting the user to one request at a time was necessary as all of these operations should be immediate when done on a cache.

The only drawback to putting these forms on the cache overview page is that the cache overview page takes a while to load the graphs. It might be better to place them on their own page.

@stkem @andrewconner


stkem:
I'm afraid this doesn't work correctly. You are only accessing the local in memory cache, when we really want to do this to memcached.
Also, like you said, I think this would indeed be better off on it's own page.

Colin-Lane:
The cache operations have been moved to their own page. Also, using memcache instead of the local cache.

stkem:
lgtm! make sure test it in prod when it's out (with some very clearly not used in prod cache key prefix, e.g. 'testtesttest' and a short-ish TTL)

PR #15365, opened 2015-05-15
changed Sendgrid username/password (now uses API key)

camhashemi:
@stkem also a minor patch to lib invites for non-users

tested in dev mode, sending welcome + reset password emails properly.


stkem:
Code looks good to me. I'm assuming the key format is correct :-)

camhashemi:
I tested in dev and received the right emails, so it should work. I'll do a few tests in production and roll back if something's wrong.

stkem:
:+1:

PR #15372, opened 2015-05-16
filtering out library descriptions that are too long

camhashemi:
@andrewconner I think this solution from sterling is better. just throw away any long descriptions.


camhashemi:
@stkem

PR #15374, opened 2015-05-16
possible temp solution for the user creation problem and pointing out…

eishay:
… empty email

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/core/AuthController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/core/AuthController.scala
@@ -513,7 +513,7 @@ class AuthController @Inject() (
                 val sfi = SocialFinalizeInfo(
                   firstName = User.sanitizeName(identity.firstName),
                   lastName = User.sanitizeName(identity.lastName),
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/app/com/keepit/controllers/core/AuthController.scala
```diff
@@ -513,7 +513,7 @@ class AuthController @Inject() (
                 val sfi = SocialFinalizeInfo(
                   firstName = User.sanitizeName(identity.firstName),
                   lastName = User.sanitizeName(identity.lastName),
-                  email = EmailAddress(identity.email.getOrElse("")),
+                  email = EmailAddress(identity.email.getOrElse("")), //todo(andrew): is having an empty string for email is the right thing to do at this point???
```
eishay, line 516: @andrewconner please investigate this line.
Is having an empty string instead of email address is the right thing to do here?
That email address translate to empty string social id in `SocialNetworks.FORTYTWO.authProvider` SocialUserInfo later on and all users with empty emails will not be able to be inserted to the db because of unique constraints.

eishay:
@andrewconner

PR #15380, opened 2015-05-18
adding alerts on partial fail/success of db batch ops

eishay:


eishay:
@andrewconner i think its related to the errors with batch inserting keeps, raw keeps update and attaching collections to keeps.

eishay:
main problem i see that if there's a failure the exception is not bubbled up or checked correctly by the caller.
will merge after the normalized uri migration

PR #15384, opened 2015-05-18
heimdal client supports library views query

yingjieMiao:
@stkem @camhashemi


kifi-backend/modules/common/app/com/keepit/heimdal/HeimdalServiceClient.scala
```diff
@@ -92,6 +92,8 @@ trait HeimdalServiceClient extends ServiceClient with KeepDiscoveryRepoAccess wi
   def processSearchHitAttribution(hit: SearchHitReport): Future[Unit]
 
   def processKeepAttribution(userId: Id[User], newKeeps: Seq[Keep]): Future[Unit]
+
+  def getOnwerLibraryViewStats(ownerId: Id[User]): Future[(Int, Map[Id[Library], Int])]
```
camhashemi, line 96: Owner*

yingjieMiao, line 96: thanks!

PR #15403, opened 2015-05-19
some logging in seqAssigner

yingjieMiao:
@stkem 

I noticed that `assignSequenceNumbers` is called on the fly, instead of scheduled. Given the fact there are > 1M items to get seq assigned, the call is going to take forever. 

Introduce `MAX_ITER` to break it eventually. 

This does not solve the problem though.


stkem:
:+1:

PR #15406, opened 2015-05-19
added GratificationEmailSender

camhashemi:
@stkem this isn't finished, need to merge with @yingjieMiao 's work.

initial (newConnections) pass, initial email printouts look good. we still need to create the scheduler plugin.


kifi-backend/modules/shoebox/app/views/email/black/gratification.scala.html
```diff
@@ -0,0 +1,46 @@
...
+@if(totalConnections >= 1) {
+<p><b>You made @totalConnections new connections</b></p>
+<ul>
+@for(userId <- newConnections) {
+    <li><a href="@profileUrl(userId, "")">@fullName(userId)</a></li>
+}
+</ul>
+Share a library with them :)
+}
+
```
yingjieMiao, line 42: It's possible that all three `if` branches return empty content. We might need a guard somewhere (likely outside the html template)

camhashemi, line 42: The class that decides who and when to send the email is probably the right place (yet to be implemented). gratEmailSender.sendToUser() should only be called when an email should be sent to that user, though we could add an extra guard for redundancy

camhashemi:
@stkem @yingjieMiao just added testing and merged with heimdal's apis. the scheduler is next.

kifi-backend/modules/shoebox/app/com/keepit/model/UserConnectionRepo.scala
```diff
@@ -149,6 +151,13 @@ class UserConnectionRepoImpl @Inject() (
     }
   }
 
+  def getConnectionsSince(id: Id[User], since: DateTime)(implicit session: RSession): Set[Id[User]] = {
```
stkem, line 154: You should add a test for this method in particular. (All repo methods with hand written SQL should have simple tests, might just be testing the syntax)

stkem:
looks good to me overall so far. I'll make another detailed pass when all the parts are there.

camhashemi:
@stkem

@yingjieMiao

kifi-backend/modules/heimdal/test/com/keepit/model/LibraryViewTrackingRepoTest.scala
```diff
@@ -0,0 +1,26 @@
+package com.keepit.model
+
+import com.keepit.common.db.Id
+import com.keepit.model.tracking.LibraryViewTrackingRepo
+import com.keepit.test.HeimdalTestInjector
+import org.specs2.mutable.Specification
+import org.joda.time.DateTime.now
+
+class LibraryViewTrackingRepoTest extends Specification with HeimdalTestInjector {
+  // TODO: heimdal test injector is erroring "CreationException: Unable to create injector".
```
camhashemi, line 12: @yingjieMiao

camhashemi:
@yingjieMiao I noticed this, but can't see any obvious differences between the two Test classes.

kifi-backend/modules/common/app/com/keepit/model/UserExperiment.scala
```diff
@@ -66,6 +66,7 @@ object ExperimentType {
   val CURATOR_DIVERSE_TOPIC_RECOS = ExperimentType("curator_diverse_topic_recos")
   val NEW_PUBLIC_FEED = ExperimentType("new_public_feed")
   val PLAIN_EMAIL = ExperimentType("plain_email")
+  val GRATIFICATION = ExperimentType("gratification")
```
stkem: Maybe a bit more descriptive name, e.g. "gratification_email"?

kifi-backend/modules/shoebox/app/com/keepit/shoebox/cron/GratificationEmailCron.scala
```diff
@@ -0,0 +1,52 @@
...
+    emailSender: GratificationEmailSender,
+    protected val airbrake: AirbrakeNotifier) extends FortyTwoActor(airbrake) with Logging {
+
+  val testDestinationEmail = EmailAddress("cam@kifi.com")
+
+  def receive = {
+    case SendEmails =>
+      emailSender.usersToSendEmailTo match {
+        case Left(ids) => ids.foreach { id => emailSender.sendToUser(id, Some(testDestinationEmail)) }
+        case Right(fIds) => fIds.map { _ filter { id => id.id != -1 } map { id => emailSender.sendToUser(id, Some(testDestinationEmail)) } }
```
stkem: It's not immediately clear what is going on here. While it seem functionally correct, overall, I would add type annotations to the Right/Left case just for clarity (i.e. `Left(x: T)` kinds thing) and brake the Right case into multiple lines with an actual argument names. It's a bit too complex to use `_` imho.
Also, what is with the  `id.id != -1` filter. Where might a `-1` id be coming from? Tests?

kifi-backend/modules/shoebox/app/com/keepit/commanders/emails/GratificationEmailSender.scala
```diff
@@ -0,0 +1,84 @@
...
+  val EXPERIMENT_DEPLOY = true
+  val MIN_FOLLOWERS = 1
+  val MIN_VIEWS = 5
+  val MIN_CONNECTIONS = 1
+
+  def apply(userId: Id[User], toAddress: Option[EmailAddress]) = sendToUser(userId, toAddress)
+
+  def sendToUser(userId: Id[User], toAddress: Option[EmailAddress]): Future[ElectronicMail] = {
+
+    val fViewsByLibrary = gratificationCommander.getLibraryViewData(userId)
```
stkem: Not required at all, but actually explicitly typing these variables explicitly can help with understanding.

kifi-backend/modules/shoebox/app/com/keepit/commanders/emails/GratificationEmailSender.scala
```diff
@@ -0,0 +1,84 @@
...
+    if (EXPERIMENT_DEPLOY) {
+      Left(userIds.filter { id =>
+        localUserExperimentCommander.userHasExperiment(id, ExperimentType.GRATIFICATION)
+      })
+    } else {
+      val result: Seq[Future[Id[User]]] = userIds.map { id =>
+        val newConnections = getNewConnections(id)
+        val followersByLibrary = gratificationCommander.getLibraryFollowerCounts(id)
+        val fViewsByLibrary = gratificationCommander.getLibraryViewData(id)
+        fViewsByLibrary.map { viewsByLib => if (newConnections.length >= MIN_CONNECTIONS || followersByLibrary.totalCount >= MIN_FOLLOWERS || viewsByLib.totalCount >= MIN_VIEWS) id else Id[User](-1) }.flatMap { future => SafeFuture(future) }
```
stkem: way too much for one line

camhashemi:
@stkem I moved the logic we were working on out of the sender and into the commander, I think that's accurate. Also see MailSender.

stkem:
:+1:

PR #15410, opened 2015-05-19
temp disabling ingestion in order to quickly clean up the table

stkem:
(there was a lot of lock contention)
@yingjieMiao


yingjieMiao:
makes sense ;-)

PR #15413, opened 2015-05-19
Airbrake upgrade

Colin-Lane:
Updates airbrake to version 3

I tried to keep the functionality the same as the previous version, just upgrading from the deprecated v2 xml notifier to a more current json notifier.

I wasn't quite sure where to inject the apiKey and projectId from as I couldn't find the module that was declaring the AirbrakeSender (so I just hardcoded those two fields :( )

I went with using case classes and Json.format[T] to build my objects and have a rigidly defined structure according to the airbrake v3 docs.

@stkem @andrewconner

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/common/healthcheck/AirbrakeFormatter.scala
+++ b/kifi-backend/modules/common/app/com/keepit/common/healthcheck/AirbrakeFormatter.scala
@@ -17,6 +17,8 @@ import akka.actor._
 
 import java.io._
 import java.net._
+import play.api.libs.json._
+
 import scala.xml._
 
 import play.api.mvc._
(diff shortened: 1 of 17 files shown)


Colin-Lane:
@leogrim

kifi-backend/modules/common/app/com/keepit/common/healthcheck/AirbrakeFormatter.scala
```diff
@@ -55,6 +57,121 @@ object ErrorWithStack {
...
+  object JsonAirbrakeBacktrace { implicit val formatter = Json.format[JsonAirbrakeBacktrace] }
+  object JsonAirbrakeError { implicit val formatter = Json.format[JsonAirbrakeError] }
+  object JsonAirbrakeNotifier { implicit val formatter = Json.format[JsonAirbrakeNotifier] }
+  object JsonAirbrakeContext { implicit val formatter = Json.format[JsonAirbrakeContext] }
+  object JsonAirbrakeEnvironment { implicit val formatter = Json.format[JsonAirbrakeEnvironment] }
+  object JsonAirbrakeMessage { implicit val formatter = Json.format[JsonAirbrakeMessage] }
+
+  private val notifier = JsonAirbrakeNotifier("S42", "0.0.2", "https://admin.kifi.com/admin")
+
+  def format(airbrake: AirbrakeError): JsValue = {
```
leogrim, line 86: You can use `writes` instead to be consistent with `Play! Format` API.

kifi-backend/modules/common/app/com/keepit/common/healthcheck/AirbrakeNotifier.scala
```diff
@@ -74,6 +74,10 @@ class AirbrakeSender @Inject() (
   systemAdminMailSender: SystemAdminMailSender)
     extends Logging {
 
+  // Where are these injected from?
+  val apiKey = "701215c3c9029a56df52250ce2c5750f"
```
stkem: They come from config files. Look at AirbrakeModule.scala line 21 (I think). `val apiKey = Play.current.configuration.getString("airbrake.key").get`

kifi-backend/modules/common/app/com/keepit/common/healthcheck/AirbrakeFormatter.scala
```diff
@@ -55,6 +57,121 @@ object ErrorWithStack {
...
 }
 
+class JsonAirbrakeFormatter(val apiKey: String, val playMode: Mode, service: FortyTwoServices, serviceDiscovery: ServiceDiscovery) {
+  val deploymentMessage: String = {
+    val repo = "https://github.com/kifi/keepit"
+    val version = service.currentVersion
+    s"api_key=$apiKey&deploy[rails_env]=$modeToRailsNaming&deploy[scm_repository]=$repo&deploy[scm_revision]=$version"
+  }
+
+  case class JsonAirbrakeBacktrace(file: String, function: String, line: Option[Int], column: Option[Int])
```
stkem, line 67: Nicely abstracted!

kifi-backend/modules/common/app/com/keepit/common/healthcheck/AirbrakeNotifier.scala
stkem: Note that one of these (I think the project id) should be different for each service.

stkem:
Made some comments regarding keys/configs. Looks good to me otherwise.

stkem:
We need to be a little careful when deploying this (i.e. keep an eye on it to make sure it's working everywhere).
It should be deployed before noon.

kifi-backend/modules/common/app/com/keepit/common/healthcheck/AirbrakeFormatter.scala
```diff
@@ -55,6 +57,121 @@ object ErrorWithStack {
...
+  case class JsonAirbrakeError(`type`: String, message: String, backtrace: Seq[JsonAirbrakeBacktrace])
```
andrewconner, line 68: Hah, you found a rough patch of Scala (`type`).

Colin-Lane:
Updated the airbrake config settings to include the project id.

stkem:
lgtm. let's get it out!

PR #15414, opened 2015-05-19
Admin Cache fix

Colin-Lane:
We currently return {key : value} as a json object, the key being
whatever the actual key is. We were getting undefined for data.key from the admin cache page.
Since we already know what the key is, just return the value.

@stkem @andrewconner


stkem:
:+1:

PR #15415, opened 2015-05-19
invite modal (to collaborate only)

ahsu1230:
getting rid of the dropdown invite search menu (a lot of things taken from `libraryShareSearch`).
Functionality is basically the same, but without logic regarding showing/hiding the menu & scrolling around.

For now, only have this modal popup for collaboration invite, which only appears for anyone on `collaborative` experiment.

@2is10 @andrewconner


kifi-backend/modules/shoebox/angular/src/libraries/libraryInviteSearch.js
```diff
@@ -0,0 +1,302 @@
...
+        scope.showSpinner = false;
+        scope.email = 'Send to any email';
+        scope.emailHelp = 'Type the email address';
+        scope.inviter = profileService.me;
+
+        function init() {
+          if (scope.modalData.inviteType) {
+            scope.library = scope.modalData.library;
+            scope.inviteType = scope.modalData.inviteType;
+            scope.modalTitle = scope.inviteType === 'collaborate' ? 'Who would you like to collaborate with?' : 'Who would you like to invite?';
```
2is10: Angular templates support ternary expressions. Better to put this directly in the template:

``` HTML
<h1>{{inviteType === 'collaborate' ? 'Who would you like to collaborate with?' : 'Who would you like to invite?'}}</h1>
```

kifi-backend/modules/shoebox/angular/src/libraries/libraryInviteSearch.js
```diff
@@ -0,0 +1,302 @@
...
+        //
+        // Scope data.
+        //
+        scope.isOwn = false;
+        scope.results = [];
+        scope.search = {};
+        scope.share = {};
+        scope.showSpinner = false;
+        scope.email = 'Send to any email';
+        scope.emailHelp = 'Type the email address';
```
2is10: I’m of the opinion that messages belong in templates. Consider:

``` js
scope.query = '';
scope.queryIsValidEmailAddr = false;
```

and

``` HTML
<div>{{query || 'Send to any email'}}</div>
<div>{{queryIsValidEmailAddr ? 'An email address' : query ? 'Keep typing the email address' : 'Type the email address'}}</div>
```

2is10:
Nice work.

ahsu1230:
thanks for the review @2is10 (sorry, it was a big one)

PR #15417, opened 2015-05-19
cortex update with new uri format (step 1)

yingjieMiao:
@leogrim @stkem 

keynote:
- sql table augmentation
- updated `CortexURI`
- tmp disabled URI syncing from shoebox. (plan to reset seqNum and re-enable)
- using rover client as article fetcher


leogrim:
:+1:

PR #15423, opened 2015-05-19
fix javascript errors

ahsu1230:
@andrewconner @2is10 sorry!

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/libraryHeader.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/libraryHeader.js
@@ -45,7 +45,7 @@ angular.module('kifi')
         scope.descScrollable = false;
         scope.imagePreview = null;
         scope.followBtnJustClicked = false;
-        scope.onCollabExperiment = profileService.me.experiments.indexOf('collaborative') > -1;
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/libraryHeader.js
```diff
@@ -45,7 +45,7 @@ angular.module('kifi')
         scope.descScrollable = false;
         scope.imagePreview = null;
         scope.followBtnJustClicked = false;
-        scope.onCollabExperiment = profileService.me.experiments.indexOf('collaborative') > -1;
+        scope.onCollabExperiment = (profileService.me.experiments || []).indexOf('collaborative') > -1;
```
2is10, line 48: Thanks. I guess we could also set `me.experiments` to an empty array in `profileService` when a user is logged out, but that feels kind of weird to me. Personally, I think `me` should be `null` when a user is logged out.

Another way this could have been written, btw, is:

``` js
 scope.onCollabExperiment = scope.$root.userLoggedIn && profileService.me.experiments.indexOf('collaborative') > -1;
```

2is10:
:+1:

PR #15424, opened 2015-05-19
Updates plain email to include link to profile

Colin-Lane:
Adds text to include a link to the user profile.

I didn't see "I’m Kifi’s co-founder and CEO." in the last revision of the email chain so I removed it.
I updated the rest of the email to include what was in bold as well.

@andrewconner @eishay


andrewconner:
Looks good to me :+1:

PR #15432, opened 2015-05-20
adding membership to library card infos

ahsu1230:
@andrewconner


ahsu1230:
@jaredpetker @DerekBurch @2is10 @mrbrown14 @thassss 

In endpoints like `/m/1/:username/libraries` & `/site/:username/libraries/` to get profile libraries
library card objects now hold a `membership` field which will tell you the access (`read_only`, `read_write`) of the user viewing the library and `none` if there is no membership.
You can stop using the `following` flag since this would be redundant =)

2is10:
With the new design, the website at least needs to know a few attributes of a `LibraryMembership` for a user profile library card: access level, listed, and subscribed to updates. It might be nice to group those in an object (like `"membership": {"access": "read_only", "listed": true, "subscribed": true}`) to make the JSON more closely resemble our domain model on the server.

ahsu1230:
Thanks @2is10 for the good point, please go ahead and use the `membership` object and stop using the deprecated `listed` or `following` fields. If the `membership` object does not appear, the user does not have a membership. Thanks!

2is10:
:+1:

thassss:
So this is replacing the "access" field?

ahsu1230:
@thassss for this PR, this only affects the profile libraries (GET m/1/:username/libraries)
The `access` field in the (POST m/1/:username/libraries) remains the same. Does that clear things up?

PR #15438, opened 2015-05-20
An attempt at fixing the ID issue

andrewconner:
@eishay Error reporting happens a few lines down. As before, we need to look for "failed to persist"s

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepInterner.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepInterner.scala
@@ -146,10 +146,10 @@ class KeepInterner @Inject() (
   }
 
   private def internUriAndBookmarkBatch(bms: Seq[RawBookmarkRepresentation], userId: Id[User], library: Library, source: KeepSource, installationId: Option[ExternalId[KifiInstallation]]) = {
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepInterner.scala
```diff
@@ -146,10 +146,10 @@ class KeepInterner @Inject() (
   }
 
   private def internUriAndBookmarkBatch(bms: Seq[RawBookmarkRepresentation], userId: Id[User], library: Library, source: KeepSource, installationId: Option[ExternalId[KifiInstallation]]) = {
-    val (persisted, failed) = db.readWriteBatch(bms, attempts = 4) { (session, bm) =>
-      internUriAndBookmark(bm, userId, library, source, installationId)(session)
-    } map {
-      case (bm, res) => bm -> res.flatten
+    val (persisted, failed) = db.readWrite { implicit session =>
```
eishay, line 149: no `attempts`?

andrewconner, line 149: I don't actually think that would help. No exception can get to that block (`internUriAndBookmark` is wrapped in a `try` and I wrap it again — in case there's a semantic change — in a `Try`). I can explicitly retry failed inserts if you'd like.

PR #15439, opened 2015-05-20
Retrying inserting bookmarks

andrewconner:
@eishay If you want retries, here's a way to do it. Not sure it's necessary though — all failed inserts I see recently are bad URLs (like `file:///Users/jannaferrett/Desktop/PROJECTS/AfriqueUnie/themeforest-8645312-clever-course-learning-management-system-theme/Document-Instruction/index.html`). So not sure this helps.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepInterner.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepInterner.scala
@@ -190,7 +190,7 @@ class KeepInterner @Inject() (
           case s if KeepSource.discrete.contains(s) => START_OF_TIME
           case _ => currentDateTime // todo: useful to de-prioritize bulk imports.
         }
-        scraper.scheduleScrape(uri, date)
+        Try(scraper.scheduleScrape(uri, date)) // This throws exceptions, such as MySQLIntegrityConstraintViolationException. Why!? Needed?
       }
 
       if (KeepSource.discrete.contains(source)) {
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepInterner.scala
```diff
@@ -177,7 +190,7 @@ class KeepInterner @Inject() (
           case s if KeepSource.discrete.contains(s) => START_OF_TIME
           case _ => currentDateTime // todo: useful to de-prioritize bulk imports.
         }
-        scraper.scheduleScrape(uri, date)
+        Try(scraper.scheduleScrape(uri, date)) // This throws exceptions, such as MySQLIntegrityConstraintViolationException. Why!? Needed?
```
andrewconner, line 193: @leogrim Thoughts? Seeing a few failed keeps because of this.

leogrim, line 193: Yep I have a good answer today: I'm going to remove this and `ScrapeInfo` altogether.

andrewconner, line 193: !!!!

Colin-Lane, line 193: Since this pull request is about retrying. Found an excellent retry snippet on stackoverflow the other day.

```
  @tailrec
  private final def retry[T](retries: Int)(block: => T): Try[T] = {
    Try(block) match {
      case success: Success[T] => success
      case Failure(e) if !NonFatal(e) => throw e
      case _ if retries > 1 => retry(retries - 1)(block)
      case failure => failure
    }
  }
```

PR #15440, opened 2015-05-20
Removing pass phrase UI (todo: business logic). Adding index to email_opt_out table.

andrewconner:
@Colin-Lane @ahsu1230


ahsu1230:
angular part is good!

PR #15449, opened 2015-05-20
Removing passphrase from our code

andrewconner:
@camhashemi @ahsu1230


camhashemi:
@andrewconner awesome. I only removed passphrase's from templates. thank you for doing the real clean up.

PR #15456, opened 2015-05-20
deprecating routeService, introducing net

2is10:
cc @andrewconner @ahsu1230 @stkem 

The way we’ve been doing things with `routeService` is cumbersome:
1. While URLs are specified in routeService, their HTTP methods are specified elsewhere.
2. Users of `routeService` have to inject `$http` too.

Hence this new `net` service.


stkem:
Nice!

PR #15461, opened 2015-05-21
avoiding 200 with no content (shows error in Firefox console)

2is10:
@ahsu1230


ahsu1230:
is this is a thing for Firefox? Like and Ok status with no response body will always return an error?

2is10:
It’s a technicality of the HTTP spec. Success with no response body is supposed to be indicated using status 204. Other browsers don’t complain. Firefox does. Better to abide by the HTTP spec.
http://en.wikipedia.org/wiki/List_of_HTTP_status_codes#2xx_Success

2is10:
It’s not fatal in Firefox. A distracting and unclear “no element found” error message (warning, really) just shows up in the console. Actually, after a little more reading in [the Mozilla bug](https://bugzilla.mozilla.org/show_bug.cgi?id=884693) and other various commentary online, I’m not even sure using status 204 will make it stop.

PR #15470, opened 2015-05-21
update conf for hosted statsd on all services

joshm1:
@stkem 

looks like stats keys are in the right place for curator


stkem:
:+1:

PR #15471, opened 2015-05-21
Library revoke invitation

Colin-Lane:
Provides Web api to revoke library invitations given an email address or an external id.

@stkem

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
@@ -1001,23 +1001,27 @@ class LibraryCommander @Inject() (
     }
   }
 
-  def revokeInvitationToLibrary(libraryId: Id[Library], inviterId: Id[User], invitee: Either[Id[User], EmailAddress]): Either[String, Tuple2[String, String]] = {
-    val libraryInvite: Option[LibraryInvite] = invitee match {
-      case Left(id) => db.readOnlyMaster { implicit s =>
-        libraryInviteRepo.getLastSentByLibraryIdAndInviterIdAndUserId(libraryId, inviterId, id, Set(LibraryInviteStates.ACTIVE))
+  def revokeInvitationToLibrary(libraryId: Id[Library], inviterId: Id[User], invitee: Either[ExternalId[User], EmailAddress]): Either[String, Tuple2[String, String]] = {
+    val libraryInvite: Either[Option[LibraryInvite], String] = invitee match {
+      case Left(externalId) => db.readOnlyMaster { implicit s =>
+        userRepo.getOpt(externalId) match {
+          case Some(userId) => Left(libraryInviteRepo.getLastSentByLibraryIdAndInviterIdAndUserId(libraryId, inviterId, userId.id.get, Set(LibraryInviteStates.ACTIVE)))
+          case None => Right(s"user with external id $externalId does not exist")
+        }
       }
       case Right(email) => db.readOnlyMaster { implicit s =>
-        libraryInviteRepo.getLastSentByLibraryIdAndInviterIdAndEmail(libraryId, inviterId, email, Set(LibraryInviteStates.ACTIVE))
+        Left(libraryInviteRepo.getLastSentByLibraryIdAndInviterIdAndEmail(libraryId, inviterId, email, Set(LibraryInviteStates.ACTIVE)))
       }
     }
     libraryInvite match {
-      case Some(toDelete) => db.readWrite(attempts = 3) { implicit s =>
+      case Left(Some(toDelete)) => db.readWrite(attempts = 3) { implicit s =>
         libraryInviteRepo.delete(toDelete) match {
           case x if x > 0 => Left("successfully deleted library invite")
           case _ => Right("error" -> "could not delete library invite")
         }
       }
-      case _ => Right("error" -> "could not find library invite to delete")
+      case Left(None) => Right("error" -> "could not find library invite to delete")
+      case Right(error) => Right("error" -> error)
     }
   }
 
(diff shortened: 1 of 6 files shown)


Colin-Lane:
Is there a distinction between mobile and web, or should I just copy/paste my LibraryController code into the MobileLibraryController as well?
@stkem @andrewconner

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -1001,6 +1001,30 @@ class LibraryCommander @Inject() (
...
+          case Some(userId) => Left(libraryInviteRepo.getLastSentByLibraryIdAndInviterIdAndUserId(libraryId, inviterId, userId.id.get, Set(LibraryInviteStates.ACTIVE)))
+          case None => Right(s"user with external id $externalId does not exist")
+        }
+      }
+      case Right(email) => db.readOnlyMaster { implicit s =>
+        Left(libraryInviteRepo.getLastSentByLibraryIdAndInviterIdAndEmail(libraryId, inviterId, email, Set(LibraryInviteStates.ACTIVE)))
+      }
+    }
+    libraryInvite match {
+      case Left(Some(toDelete)) => db.readWrite(attempts = 3) { implicit s =>
```
andrewconner: Convention for "success" vs "failure" `Either` usage is that `Left` is the error case, and `Right` is the success.

Colin-Lane: In that case, I got it backwards everywhere :)

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -1001,6 +1001,30 @@ class LibraryCommander @Inject() (
...
+        libraryInviteRepo.delete(toDelete) match {
+          case x if x > 0 => Left("successfully deleted library invite")
+          case _ => Right("error" -> "could not delete library invite")
+        }
+      }
+      case Left(None) => Right("error" -> "could not find library invite to delete")
```
andrewconner: Avoid magic strings that appear human readable. A client will be looking for these (if it can handle that error case), and would have to match on `"could not find library invite to delete"`. This could easily be changed because it appears to be for human consumption. For error codes, pick something that _looks_ like it's for a machine: `"no_invite_exists"`.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
andrewconner: http://www.scala-lang.org/api/2.11.5/index.html#scala.util.Either

> A common use of Either is as an alternative to scala.Option for dealing with possible missing values. In this usage, scala.None is replaced with a scala.util.Left which can contain useful information. scala.util.Right takes the place of scala.Some. Convention dictates that Left is used for failure and Right is used for success.

It's confusing.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -1001,6 +1001,30 @@ class LibraryCommander @Inject() (
...
+          case None => Right(s"user with external id $externalId does not exist")
+        }
+      }
+      case Right(email) => db.readOnlyMaster { implicit s =>
+        Left(libraryInviteRepo.getLastSentByLibraryIdAndInviterIdAndEmail(libraryId, inviterId, email, Set(LibraryInviteStates.ACTIVE)))
+      }
+    }
+    libraryInvite match {
+      case Left(Some(toDelete)) => db.readWrite(attempts = 3) { implicit s =>
+        libraryInviteRepo.delete(toDelete) match {
```
andrewconner: Usually, we avoid deleting rows because we'll have no record of them existing. Instead, we set the state (which all models should have) as `inactive`. Example: `libraryInviteRepo.save(toDelete.copy(state = LibraryInviteStates.INACTIVE))`. It's a trade-off, because then we have to make sure we're filtering for inactives, but it seems to help us often.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -1001,6 +1001,30 @@ class LibraryCommander @Inject() (
     }
   }
 
+  def revokeInvitationToLibrary(libraryId: Id[Library], inviterId: Id[User], invitee: Either[ExternalId[User], EmailAddress]): Either[String, Tuple2[String, String]] = {
```
andrewconner: `Tuple2` is fine, but you can also write this type as `(String, String)`.

andrewconner:
Re: mobile. Yep. Our pattern typically is to write the business logic in a Commander (returning typed objects instead of JSON), and then have separate controller methods for mobile/extension/site who manage serializing and deserializing data. The reason is APIs change for specific clients, and trying to have a unified controller is a nightmare. We trade code duplication to get this.

Colin-Lane:
Updated the code to be more inline with andrew's suggestions.
Also added in the mobile controller code
- tests all passing locally
  @stkem @andrewconner

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileLibraryController.scala
```diff
@@ -260,6 +260,29 @@ class MobileLibraryController @Inject() (
     Ok(Json.obj("libraries" -> libsFollowing, "invited" -> libsInvitedTo))
   }
 
+  def getUserByIdOrEmail(json: JsValue) = {
```
stkem, line 263: I have a hard time parsing this. I think an explicit return type would help a lot here.

stkem:
lgtm!

PR #15474, opened 2015-05-21
refactor "membership" responses on website - adding membership object to website endpoints

ahsu1230:
@andrewconner @2is10 
To make `membership` objects consistent on all clients, I'd like to have membership info (listed, subscription, access level) inside a membership object
This is the first PR to refactoring. We can add a `membershipObj` field as a response. Step 2 is to only use `membershipObj` on the FE. Step 3 is to remove the unused fields (and maybe rename membershipObj to just membership).


andrewconner:
Good plan. Not a huge fan of `membershipObj`. Perhaps @2is10 can come up with a good name? :)

2is10:
I like this direction. I agree with Andrew that using `"membershipObj"` here is suboptimal. One alternative approach:
1. Before merging this, first make the Angular app forward compatible with `"membership"` being an object instead of string (while also keeping it working with shoebox sending a string there, as it does today).
2. After Angular accepts either format, merge this change—without worrying about backwards compatibility.
3. Then go back and simplify the Angular app, dropping its support for the old format.

PR #15476, opened 2015-05-21
Reclaim Username Aliases on User Delete

Colin-Lane:
When deleting a user, also reclaim their usernames.
@stkem


andrewconner:
Are we already clearing out the username field? That's actually the big one that matters.

stkem:
:+1:

PR #15478, opened 2015-05-21
guard prevents sending test emails to non-kifi addresses

camhashemi:
@stkem


stkem:
Not sure we need that level of safety, but sure, why not :-)

PR #15488, opened 2015-05-22
Being hyper careful about db transactions

andrewconner:
@leogrim @eishay


eishay:
still having problems

andrewconner:
Alright, then I don't know where it would come from. Which exception is it?

eishay:
The errors about can't link tags and keeps because the keep id is missing
in the db.

PR #15499, opened 2015-05-22
limit library features retrieved.

yingjieMiao:
@stkem 

cheaper computation


yingjieMiao:
would like to see if this helps at all

stkem:
I think 25 is too low for this. I thought we talked about making it 100. (And there isn't really a performance problem here)

yingjieMiao:
ok, let me revert.

PR #15508, opened 2015-05-22
Safer youtube parsing

andrewconner:
- We don't trust Jsoup to be nice. It returns nulls and throws exceptions.
- YouTube can always break their selectors. We need to know about it, but shouldn't break because of it. Happy to change these defaults if you have a way to represent "failed scrape"
- In this case, YouTube changed their view count format, from something like "2,333" to "2,333 views"

### DIFF
--- a/kifi-backend/modules/rover/app/com/keepit/rover/article/fetcher/YoutubeArticleFetcher.scala
+++ b/kifi-backend/modules/rover/app/com/keepit/rover/article/fetcher/YoutubeArticleFetcher.scala
@@ -74,7 +74,7 @@ class YoutubeArticleFetcher @Inject() (
     val description = SafeOpt(doc.doc.select("#watch-description-text .content").text()).getOrElse("")
     val channel = SafeOpt(doc.doc.select("#watch7-user-header .yt-user-name").text()).getOrElse("")
     val tags = SafeOpt(doc.doc.select("#watch-description-extras .content .watch-info-tag-list").map(_.text()).filter(_.nonEmpty)).getOrElse(Seq.empty)
-    val viewCount = SafeOpt(doc.doc.select("#watch7-views-info .watch-view-count").text().replaceAll ("\\D", "").toInt).getOrElse(0)
+    val viewCount = SafeOpt(doc.doc.select("#watch7-views-info .watch-view-count").text().replaceAll("\\D", "").toInt).getOrElse(0)
 
     // todo: Add alerting so we know if youtube breaks these selectors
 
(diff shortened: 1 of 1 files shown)


andrewconner:
@leogrim @stkem This okay? Do we have a better way to report that a scrape failed (without blowing up and throwing 5k exceptions)?

leogrim:
Thanks for fixing this, I noticed it this morning, we can easily reschedule refetches for these with a database query. 

The fetcher doesn't do anything to catch exceptions, it merely wraps them. It's up to the caller to decide what to do with it:
- 'official' fetches (i.e. ones that have an `ArticleInfo`) do not Airbrake. So we can just look at the `ArticleInfo` table. Actually, it would be a nice addition to have a plugin that looks at recent rows and notify us if too many are reporting a failure.
- Online calls to fetch urls that we haven't interned yet are throwing these exceptions. These calls are used for normalization (get signature, get recent article) and almost by definition, the urls involved have often not been vetted yet, so there's probably a higher error rate. The fetcher is called in `ArticleCommander` where we already catch a couple of different exceptions, we can add that one NotOkResponseException :)

kifi-backend/modules/rover/app/com/keepit/rover/article/fetcher/YoutubeArticleFetcher.scala
```diff
@@ -70,11 +70,14 @@ class YoutubeArticleFetcher @Inject() (
...
-    val channel = doc.doc.select("#watch7-user-header .yt-user-name").text()
-    val tags = doc.doc.select("#watch-description-extras .content .watch-info-tag-list").map(_.text()).filter(_.nonEmpty)
-    val viewCount = doc.doc.select("#watch7-views-info .watch-view-count").text().trim.replaceAllLiterally(",", "").toInt
+    val headline = SafeOpt(doc.doc.getElementById("watch-headline-title")).map(_.text).filter(_.nonEmpty)
+    val description = SafeOpt(doc.doc.select("#watch-description-text .content").text()).getOrElse("")
+    val channel = SafeOpt(doc.doc.select("#watch7-user-header .yt-user-name").text()).getOrElse("")
+    val tags = SafeOpt(doc.doc.select("#watch-description-extras .content .watch-info-tag-list").map(_.text()).filter(_.nonEmpty)).getOrElse(Seq.empty)
+    val viewCount = SafeOpt(doc.doc.select("#watch7-views-info .watch-view-count").text().replaceAll("\\D", "").toInt).getOrElse(0)
+
+    // todo: Add alerting so we know if youtube breaks these selectors
```
leogrim, line 79: All errors on urls that have been interned `Id[NormalizedURI]` are persisted to the database in `ArticleInfo`. It's probably the best place to have a plugin look for patterns and report to us.

andrewconner:
So, does that mean this is good?

leogrim:
:+1: ;)

PR #15510, opened 2015-05-22
Adds RSS Feeds to libraries

Colin-Lane:
Adds RSS Feeds to a library at /:user/:slug/rss
- guarded behind experiment library_rss_feed

@stkem @andrewconner


kifi-backend/modules/shoebox/app/com/keepit/controllers/website/KifiSiteRouter.scala
```diff
@@ -98,6 +101,28 @@ class KifiSiteRouter @Inject() (
     }
   }
 
+  def serveRSSFeedIfLibraryFound(username: Username, librarySlug: String, authToken: Option[String]) = WebAppPage { implicit request =>
```
stkem, line 104: I think this should get it's own controller. The Kifi Site Router pretty much exclusively deal with actually serving the angular page.

Colin-Lane, line 104: I was wondering about this as well. But I went with it because of the similar routes. I can find a new home for this controller method.

stkem, line 104: Also should not be a WebAppPage, but a regular UserAction, I think.

stkem:
Except for the controller thing everything looks good! (I'm sure there will be tweaks to this after it's out, e.g. the licensing stuff)

kifi-backend/modules/shoebox/app/com/keepit/common/seo/FeedCommander.scala
```diff
@@ -34,6 +40,62 @@ class FeedCommander @Inject() (
...
+      </item>
+    }
+  }
+
+  def libraryFeed(feedUrl: String, library: Library): Elem = {
+    val keepCountToDisplay = 10
+
+    implicit def convert(keep: Keep): RssItem = {
+      val (keepImage, originalKeeper) = db.readOnlyMaster { implicit s =>
+        val keepImages: Seq[KeepImage] = keepImageRepo.getForKeepId(keep.id.get)
```
andrewconner: Getting the best image for a keep is actually a bit more tricky. Check out `KeepImageCommander#getBestImageForKeep`. As an argument, you can include a hint about the target size you want. Judging by the name of this field (`icon`), if you want something small, `ScaleImageRequest(ImageSize(100,100))` may be good.

kifi-backend/modules/shoebox/app/com/keepit/common/seo/FeedCommander.scala
```diff
@@ -34,6 +40,62 @@ class FeedCommander @Inject() (
...
+        (keepImages.headOption, userRepo.getNoCache(keep.originalKeeperId.getOrElse(keep.userId)))
+      }
+
+      RssItem(title = keep.title.getOrElse(""), description = keep.note.getOrElse("None"), link = keep.url,
```
andrewconner: Does RSS have a concept of an empty field? Specifically looking at `description`. Unless spec says make it "None" for no description (which it may, haven't checked), blank may be better. Thoughts?

kifi-backend/modules/shoebox/app/com/keepit/common/seo/FeedCommander.scala
```diff
@@ -34,6 +40,62 @@ class FeedCommander @Inject() (
...
+        guid = keep.externalId.id, pubDate = keep.keptAt, creator = originalKeeper.fullName,
+        icon = keepImage.map(_.imagePath.getUrl(s3ImageConfig)).getOrElse("No Image"))
```
andrewconner: Same as above, if "No Image" isn't the standard, ignoring the field or leaving it blank may be best. Browsers may try to render that as a URL.

kifi-backend/modules/shoebox/app/com/keepit/common/seo/FeedCommander.scala
```diff
@@ -34,6 +40,62 @@ class FeedCommander @Inject() (
...
+    }
+
+    val (libImage, keeps) = db.readOnlyMaster { implicit session =>
+      val image = libraryImageRepo.getActiveForLibraryId(library.id.get)
+      val keeps = keepRepo.getKeepsFromLibrarySince(DateTime.now.withYear(2000), library.id.get, keepCountToDisplay)
```
andrewconner: If I'm reading the repo method right, this seems to order by `keptAt` ascending, so these are the _oldest_ keeps. Wouldn't we want the newest?

kifi-backend/modules/shoebox/app/com/keepit/common/seo/FeedCommander.scala
```diff
@@ -34,6 +40,62 @@ class FeedCommander @Inject() (
...
+        (keepImages.headOption, userRepo.getNoCache(keep.originalKeeperId.getOrElse(keep.userId)))
+      }
+
+      RssItem(title = keep.title.getOrElse(""), description = keep.note.getOrElse("None"), link = keep.url,
+        guid = keep.externalId.id, pubDate = keep.keptAt, creator = originalKeeper.fullName,
+        icon = keepImage.map(_.imagePath.getUrl(s3ImageConfig)).getOrElse("No Image"))
+    }
+
+    val (libImage, keeps) = db.readOnlyMaster { implicit session =>
+      val image = libraryImageRepo.getActiveForLibraryId(library.id.get)
```
andrewconner: Same as `KeepImage`s, check out `LibraryImageCommander`. There's a bit more logic in picking a good image. The reason it matters is this query could return the original 3000x2000 image, which many clients wouldn't be able to handle well.

kifi-backend/modules/shoebox/app/com/keepit/common/seo/FeedCommander.scala
```diff
@@ -34,6 +40,62 @@ class FeedCommander @Inject() (
...
+        <description>{ library.description }</description>
+        <image>
+          <url>{ libImage }</url>
+          <title>{ library.name }</title>
+          <link>{ feedUrl }</link>
+        </image>
+        <copyright>Copyright { currentDateTime.getYear }, FortyTwo Inc.</copyright>
+        <atom:link ref="self" type="application/rss+xml" href={ feedUrl }/>
+        <atom:link ref="hub" href="https://pubsubhubbub.appspot.com/"/>
+        { rssItems(keeps) }{ /* License asking for attribution */ }
```
andrewconner: I am slow on this, sorry! Will get some sort of license for you here.

kifi-backend/modules/shoebox/app/com/keepit/common/seo/FeedCommander.scala
Colin-Lane: It's not completely clear, but I think description may be required. It does not say use "None" however. I could just leave it blank, but I'm not sure that would look right either.

kifi-backend/modules/shoebox/app/com/keepit/common/seo/FeedCommander.scala
Colin-Lane: yes, we would want newest. I can just use getByLibrary which sorts by keptAt descending, and filters by state.

kifi-backend/modules/shoebox/app/com/keepit/common/seo/FeedCommander.scala
andrewconner: Blank is probably better. Clients can have sane defaults. "None" means it'll always say that (no client could plan for a sane default for that).

Colin-Lane:
Moved Library RSS controller code to its own file.
Added tests to verify controller is behaving correctly.
FeedCommander now using best Image for images.
Keeps are ordered by descending keptAt

@stkem @andrewconner

kifi-backend/modules/shoebox/app/com/keepit/common/seo/FeedCommander.scala
```diff
@@ -34,6 +38,65 @@ class FeedCommander @Inject() (
...
+          <media:title type="html">{ item.title }</media:title>
+        </media:content>
+      </item>
+    }
+  }
+
+  def libraryFeed(feedUrl: String, library: Library): Elem = {
+    val keepCountToDisplay = 10
+
+    implicit def convert(keep: Keep): RssItem = {
```
andrewconner: Is there an advantage of using an implicit here? Be wary of this, because it can potentially make it harder to understand what's happening (you had a list of `Keep`s, now you have a list of `RssItem`s magically).

Implicit variables and implicit classes (which enable library augmentation, see `com.keepit.common.core`) do not suffer from this quite as much.

PR #15515, opened 2015-05-22
using the .access property of libraries and reco libraries to determine whether following

2is10:
...instead of the long-lived in-memory list of followed libraries, which I think we can next stop loading altogether.

@ahsu1230


ahsu1230:
:+1:

PR #15521, opened 2015-05-24
Stopgap to help bad migrations

andrewconner:
@leogrim @eishay @stkem See https://fortytwo.airbrake.io/projects/91268/groups/1426344665598673305/notices/1427551881995501671

Migration appears to have stopped because of a bad URI. Eishay tried adding a rule, didn't help. I tried making the URL more friendly, didn't help:

`update normalized_uri set url = "http://www.flaticon.com/categories/animals", should_have_content=0,seq=-1 where id=6951507 limit 1;`

I hope this doesn't break anything, please review this asap.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/normalizer/NormalizationService.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/normalizer/NormalizationService.scala
@@ -58,10 +58,10 @@ class NormalizationServiceImpl @Inject() (
     val recentFailedChecks = db.readOnlyReplica { implicit s => failedContentCheckRepo.getRecentCountByURL(currentReference.url, now.minusMinutes(5)) }
     val somethingBadIsHappening = {
       // Léo: Please improve/fix this properly
-      log.warn(s"[NormalizationService] Stopping normalization for ${currentReference.uriId}. ${currentReference.url}. Candidates: ${candidates.map(_.url).mkString("  ")}")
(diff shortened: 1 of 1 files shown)


eishay:
hope it solves this one, though I guess other things could go wrong in addition to url length.

andrewconner:
This does appear to have fixed it, or at least stopped the infinite loop. Seeing lots of "Better reference"s in the logs again (they had stopped). Many things we need to understand, but primarily: why does one bad host stop normalization system wide?

leogrim:
Sorry about that, I'm on mobile so I can explain more on Tuesday:
- the ingestion on Shoebox keeps retrying on exceptions and exceptions have
  been happening on online calls - get recent article or signature, while
  offline failures are persisted to the DB. The idea is that Rover returns
  Options and isn't supposed to throw any permanent one (i.e. outside of S3
  being temporarily unavailable for instance, in which case the ingestion
  will retry and succeed next time). But this has only been live for a week
  or so and I haven't encountered all the possible exceptions yet, so I'm not
  catching them all in Rover -ArticleCommander.getOrElseFetchRecentArticle -
  if they are "permanent" exceptions e.g. the socket timeout, we can add a
  recover over there.

I would need to take a better look on my laptop to understand whether
there's anything else than that going on here

leogrim:
Also: As I'm about to blog about it, I have been improving
`ConcurrentTaskProcessingActor` to better track errors and alert us when
it's stuck

PR #15522, opened 2015-05-24
getting better clarity on why do we have exceptions on process normal…

eishay:
…ization info.

these exception leads to "Failed to ingest Shoebox Article updates from Rover" and probably infinite loops


eishay:
@andrewconner @leogrim @stkem

PR #15525, opened 2015-05-25
Hopefully stopping bad prenormalizations.

andrewconner:
@leogrim @eishay This may stop the recent one. Hopefully this is safe, but I'm not sure. This probably won't handle everything, but may stop it sooner.

Please review and give some thought about a better way to do it.

User provided data should not cause us to have stability issues or flood airbrake.

### DIFF
--- a/kifi-backend/modules/shoebox/test/com/keepit/commanders/KeepInternerTest.scala
+++ b/kifi-backend/modules/shoebox/test/com/keepit/commanders/KeepInternerTest.scala
@@ -162,13 +162,12 @@ class KeepInternerTest extends Specification with ShoeboxTestInjector {
         raw === deduped
 
         val (bookmarks, _) = bookmarkInterner.internRawBookmarks(raw, user.id.get, library, KeepSource.email)
-        fakeAirbrake.errorCount() === 0
-        bookmarks.size === 3
+        fakeAirbrake.errorCount() === 2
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/test/com/keepit/commanders/KeepInternerTest.scala
```diff
@@ -162,13 +162,12 @@ class KeepInternerTest extends Specification with ShoeboxTestInjector {
...
-        fakeAirbrake.errorCount() === 0
-        bookmarks.size === 3
+        fakeAirbrake.errorCount() === 2
+        bookmarks.size === 2
         db.readWrite { implicit session =>
-          keepRepo.all.size === 3
+          keepRepo.all.size === 2
           keepRepo.all.map(_.url).toSet === Set[String](
             "http://42go.com",
-            ("http://kifi.com/" + List.fill(300)("this_is_a_very_long_url/").mkString).take(URLFactory.MAX_URL_SIZE),
```
andrewconner, line 171: My change broke this test. We were testing that users _can_ keep bad (too long) URLs? Seems bad.

andrewconner:
Because of the test change, would like another pair of eyes before merging.

leogrim:
Looks good to me, I don't know about this test / whether there's a good reason not reject long urls right away.

PR #15533, opened 2015-05-26
refactor website client to handle a membership object

ahsu1230:
@andrewconner @2is10 

change `getLibrary` (by user/slug or id) to handle a `membership` object inside the `library` object instead of separate sibling fields `membership`, `access`, `subscribed`, `listed`.

If a user does not have access to a library, there should be no `membership` object

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/library.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/library.js
@@ -280,7 +280,7 @@ angular.module('kifi')
 
     $rootScope.$emit('libraryOnPage', library);
 
-    if (!libraryService.isLibraryMainOrSecret(library) && library.access !== 'none') {
+    if (!libraryService.isLibraryMainOrSecret(library) && library.membership) {
       $rootScope.$emit('lastViewedLib', library);
     }
 
@@ -297,7 +297,7 @@ angular.module('kifi')
       if (initParams.getAndClear('install') === '1' && !installService.installedVersion) {
         showInstallModal();
       }
-      if (initParams.getAndClear('intent') === 'follow' && $scope.library.access === 'none') {
+      if (initParams.getAndClear('intent') === 'follow' && !$scope.library.membership) {
         libraryService.joinLibrary($scope.library.id);
       }
     });
--- a/kifi-backend/modules/shoebox/angular/src/libraries/libraryHeader.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/libraryHeader.js
@@ -507,7 +507,7 @@ angular.module('kifi')
         }
 
         scope.followingLibrary = function () {
-          return scope.library.access === 'read_only';
+          return scope.library.membership && scope.library.membership.access === 'read_only';
         };
 
         scope.followLibrary = function (btnJustClicked) {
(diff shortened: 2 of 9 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/library.js
```diff
@@ -280,7 +280,7 @@ angular.module('kifi')
 
     $rootScope.$emit('libraryOnPage', library);
 
-    if (!libraryService.isLibraryMainOrSecret(library) && library.access !== 'none') {
+    if (!libraryService.isLibraryMainOrSecret(library) && library.membership) {
```
2is10, line 283: This reads better. Nice bonus.

kifi-backend/modules/shoebox/angular/src/libraries/libraryHeader.js
```diff
@@ -634,9 +634,14 @@ angular.module('kifi')
           }),
           $rootScope.$on('libraryJoined', function (e, libraryId) {
             var lib = scope.library;
-            if (lib && libraryId === lib.id && lib.access === 'none') {
+            if (lib && libraryId === lib.id && !lib.membership) {
+              (lib.invite || {}).actedOn = true;
```
2is10: Harmless, but this `.actedOn` line shouldn’t be added here. I removed it last week.

ahsu1230: oh I see. no wonder.

kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.js
```diff
@@ -74,13 +74,13 @@ angular.module('kifi')
...
             visibility: scope.library.visibility,
-            listed: scope.library.listed,
+            listed: scope.library.membership.listed,
             color: colorNames[scope.library.color]
           }, true).then(function (resp) {
             libraryService.fetchLibraryInfos(true);
 
             var newLibrary = resp.data.library;
-            newLibrary.listed = resp.data.listed;
+            newLibrary.listed = resp.data.listed || (resp.data.library.membership && resp.data.library.membership.listed);
```
2is10, line 83: Odd that this `.listed` is still on a library object. Maybe we should rethink having the value in this dialog. Everything else there is a property of the library, not the owner/collaborator’s membership. Maybe the checkbox in this form should instead control the default or “maximum” value of `.listed` for all collaborators or even all members.

kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.tpl.html
```diff
@@ -48,7 +48,7 @@ <h1 class="dialog-header">
               <div ng-switch-default class="manage-lib-visibility-description-content">
                 Anyone can see and follow this library
                 <div class="manage-lib-listed-visibility">
-                  <input type="checkbox" id="listed-visibility-checkbox" ng-model="library.listed">
+                  <input type="checkbox" id="listed-visibility-checkbox" ng-model="library.membership.listed">
```
2is10, line 51: In `openModifyLibrary` in userProfileLibraries.js, I see:

``` js
library.listed = data.listed;
```

It should now set `.membership.listed` instead, no?

2is10, line 51: There’s another use of `lib.listed` in userProfileLibrariesList.tpl.html.

2is10:
Nice work. This is tricky.

kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.tpl.html
ahsu1230, line 51: yup just edited both!

ahsu1230:
oh yea, btw: with this `membership` will be in the `library` object and PR #15474 will change accordingly

PR #15534, opened 2015-05-26
Updating ng dev dependencies, unbreaking dev for new engineers

andrewconner:
@2is10 People installing `npm` and `node` now and starting on Angular development were consistently having problems, so this PR should address them.

If you installed `node`/`npm` from homebrew, you may want to remove it and install directly from https://nodejs.org. However, if the following works, then don't worry about it.

For reference, my `npm` version is 2.10.1.

If you're running a very old npm, get a new one: `sudo npm install npm -g`

Run `npm install`. You shouldn't see any striking errors. If you do, grab me and we'll debug it.

### DIFF
--- a/kifi-backend/modules/shoebox/angular/package.json
+++ b/kifi-backend/modules/shoebox/angular/package.json
@@ -10,7 +10,7 @@
   },
   "engines": {
     "node": "0.10.x",
-    "npm": "1.3.x"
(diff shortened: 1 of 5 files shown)


andrewconner:
You will likely need to `rm -rf node_modules`

2is10:
:+1:

PR #15536, opened 2015-05-26
Increase size of airbrake exceptions

Colin-Lane:
- Increased the size before trimming airbrake emails.

@stkem @andrewconner @eishay


stkem:
:+1:

PR #15543, opened 2015-05-26
Temporarily taking some load out of the db so I can properly test better queries

stkem:
The query takes currently about 6 seconds to execute. Adding to 2
seconds to give the DB a breather, so I can test a better query. Hard
to accurately benchmark when the CPU is pegged at 100%. Will revert
this afterwards.

@leogrim @andrewconner


kifi-backend/modules/rover/app/com/keepit/rover/manager/RoverFetchSchedulingActor.scala
```diff
@@ -30,6 +30,7 @@ class RoverFetchSchedulingActor @Inject() (
   protected def nextBatch: Future[Seq[RoverArticleInfo]] = {
     SafeFuture {
       log.info(s"Queuing up to $maxBatchSize article fetch tasks...")
+      Thread.sleep(2000)
```
andrewconner, line 33: Awww

PR #15544, opened 2015-05-26
factoring out userProfileLibraryCard, giving it an isolate scope

2is10:
@ahsu1230

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/userProfile/userProfileLibrariesList.tpl.html
+++ b/kifi-backend/modules/shoebox/angular/src/userProfile/userProfileLibrariesList.tpl.html
@@ -3,58 +3,8 @@
     <span class="kf-upl-create-card-add-icon svg-create-library"></span>
     <div class="kf-upl-create-card-text">CREATE A LIBRARY</div>
   </button>
-  <div class="kf-upl-card kf-upl-lib-card" ng-repeat="lib in libraries" ng-style="{'background-color': lib.color}"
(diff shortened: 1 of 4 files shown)


2is10:
seems to work

kifi-backend/modules/shoebox/angular/src/userProfile/userProfileLibrariesList.tpl.html
```diff
@@ -3,58 +3,8 @@
...
-      </div>
-      <div class="kf-upl-updated" ng-switch-when="updated">
-        Updated&nbsp;<time ng-attr-datetime="{{lib.lastKept}}" am-time-ago="lib.lastKept"></time>
-      </div>
-      <div class="kf-upl-unlisted" ng-switch-when="unlisted">
-        Hidden on public profile
-      </div>
-    </div>
+  <div kf-user-profile-library-card="library" ng-repeat="library in libraries"
+    current-page-name="{{::currentPageName}}" current-page-origin="{{::currentPageOrigin}}">
```
ahsu1230, line 7: oooo how do the `::` work?

2is10, line 7: It’s a magical prefix that means “only evaluate this binding once”.
https://docs.angularjs.org/guide/expression#one-time-binding

ahsu1230, line 7: I See. neat!

PR #15546, opened 2015-05-26
remove owner's pic when viewing own profile libraries

ahsu1230:
@2is10


ahsu1230:
currently my image appears next to all the "by Aaron Hsu" when viewing my own libraries on my profile page.

ahsu1230:
this correctly sets `libraryType` so they don't show up!

2is10:
thanks

PR #15548, opened 2015-05-26
Inactive Library

Colin-Lane:
- canViewLibrary and userAccess in LibraryCommander require the library
  to be active.
  - this makes it so that deleted libraries cannot be accessed anymore.

@andrewconner @stkem


andrewconner:
Great!

stkem:
:+1:

PR #15550, opened 2015-05-26
Fixed RSS Description Option

Colin-Lane:
- now uses getOrElse


Colin-Lane:
@andrewconner

Colin-Lane:
Updated the RSS Feed with the latest suggestions from the trello board.
@stkem @andrewconner

PR #15551, opened 2015-05-27
add number of members back into library header

ahsu1230:
@andrewconner @2is10 

and made collab modals a bit longer


2is10:
:+1:

PR #15556, opened 2015-05-27
Handle meta-refresh redirects

leogrim:
@andrewconner @stkem @eishay


leogrim:
It's alive and well :)

https://admin.kifi.com/admin/uri/5879590/article/default
https://admin.kifi.com/admin/uri/3878733/article/default

PR #15566, opened 2015-05-27
Absolute Links for Mobile Library Invite

Colin-Lane:
@stkem

### DIFF
--- a/kifi-backend/modules/shoebox/conf/shoeboxService.routes
+++ b/kifi-backend/modules/shoebox/conf/shoeboxService.routes
@@ -224,7 +224,7 @@ POST    /m/1/invite                             @com.keepit.controllers.mobile.M
 
 # Profile Page
 GET     /m/1/user/:username/profile             @com.keepit.controllers.mobile.MobileUserProfileController.profile(username)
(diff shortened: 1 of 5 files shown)


stkem:
:+1: 
+ @jaredpetker you don't need to do anything on your end for this, right? (See title)

Colin-Lane:
Forgot to change branches before pushing this code. Both are still pretty minor.
@stkem 
I can close this pull and re-open two others with the commits separate, or just leave it how it is.

kifi-backend/modules/shoebox/conf/shoeboxService.routes
```diff
@@ -224,6 +224,7 @@ POST    /m/1/invite                             @com.keepit.controllers.mobile.M
 
 # Profile Page
 GET     /m/1/user/:username/profile             @com.keepit.controllers.mobile.MobileUserProfileController.profile(username)
+GET     /m/1/user/:name/connections             @com.keepit.controllers.mobile.MobileUserProfileController.getProfileConnections(name: Username, limit: Int)
```
stkem: `:username`? to follow the convention in the other routes?

stkem:
Nah, not worth the split effort for this. It's fine like this.

PR #15578, opened 2015-05-27
invite banner below library header

ahsu1230:
@2is10 @andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/libraryHeader.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/libraryHeader.js
@@ -75,10 +75,6 @@ angular.module('kifi')
           scope.numMoreFollowersText = showPlusFollowers ? $filter('num')(numFollowers - numFollowersToShow) : '';
         }
 
-        function updateInvite() {
-          scope.invite = scope.library.invite;
-        }
-
         //
         // Scope methods.
         //
@@ -87,9 +83,9 @@ angular.module('kifi')
         };
 
         scope.ignoreInvitation = function () {
-          if (scope.invite) {
+          if (scope.library && scope.library.invite) {
             libraryService.declineToJoinLibrary(scope.library.id).then(function() {
-              scope.invite = null;
+              scope.library.invite = null;
             });
           }
         };
@@ -629,21 +625,19 @@ angular.module('kifi')
         scope.$watch('library.numFollowers', setFollowersShown);
 
         [
-          $rootScope.$on('invitedLibrariesChanged', updateInvite),
           $rootScope.$on('libraryKeepCountChanged', function (e, libraryId, keepCount) {
             if (libraryId === scope.library.id) {
               scope.library.numKeeps = keepCount;
             }
           }),
           $rootScope.$on('libraryJoined', function (e, libraryId) {
-            var lib = scope.library;
-            if (lib && libraryId === lib.id && !lib.membership) {
+            if (scope.library && libraryId === scope.library.id && !scope.library.membership) {
               scope.library.membership = {
-                access: scope.invite ? scope.invite.access : 'read_only',
+                access: scope.library.invite ? scope.library.invite.access : 'read_only',
                 listed: true,
                 subscribed: false
               };
-              scope.invite = null;
+              scope.library.invite = null;
               var me = profileService.me;
               if (scope.library.membership.access === 'read_only') {
                 scope.library.numFollowers++;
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/libraryHeader.styl
```diff
@@ -25,46 +25,97 @@ colorBarHeight = .4em
...
+
+  @media (max-width: maxWindowWidthForNarrowKeeps)
+    max-width: inviteMaxWidth
+    font-size: 0.8em
+
+  @media (max-width: 536px)
+    width: calc(100% - 24px)
+
+  @media (max-width: 479px)
+    font-size: 0.92em
```
ahsu1230: hopefully there's a better way to try to match the invite banner size to keep content sizes =/

kifi-backend/modules/shoebox/angular/src/libraries/libraryHeader.js
```diff
@@ -82,7 +76,7 @@ angular.module('kifi')
         }
 
         function updateInvite() {
-          scope.inviter = libraryService.getInviter(scope.library.id);
+          scope.invite = scope.library.invite;
         }
```
2is10: This function probably isn’t needed anymore. Simpler just to use `library.invite` everywhere?

kifi-backend/modules/shoebox/angular/src/libraries/libraryHeader.js
```diff
@@ -635,25 +638,40 @@ angular.module('kifi')
           $rootScope.$on('libraryJoined', function (e, libraryId) {
             var lib = scope.library;
             if (lib && libraryId === lib.id && !lib.membership) {
-              lib.membership = {
-                access: 'read_only',
+              scope.library.membership = {
+                access: scope.invite ? scope.invite.access : 'read_only',
```
2is10: This whole code block was using `lib` instead of `scope.library` b/c it’s shorter. If you’d rather use `scope.library` explicitly everywhere, then it’s probably better to remove the `lib` variable altogether to avoid confusion.

kifi-backend/modules/shoebox/angular/src/libraries/libraryHeader.js
```diff
@@ -635,25 +638,40 @@ angular.module('kifi')
...
               var me = profileService.me;
-              if (!_.contains(lib.followers, {id: me.id})) {
-                lib.followers.push(_.pick(me, 'id', 'firstName', 'lastName', 'pictureName', 'username'));
+              if (scope.library.membership.access === 'read_only') {
+                scope.library.numFollowers++;
+                if (!_.contains(scope.library.followers, {id: me.id})) {
+                  scope.library.followers.push(_.pick(me, 'id', 'firstName', 'lastName', 'pictureName', 'username'));
+                }
+              } else if (scope.library.membership.access === 'read_write') {
+                $timeout(function() {
```
2is10: a “why” comment for this `$timeout` would be helpful

kifi-backend/modules/shoebox/angular/src/libraries/libraryMembers.js
```diff
@@ -75,6 +76,15 @@ angular.module('kifi')
...
 
+          // update own membership in library object (change access or remove membership)
+          if (scope.selectedMember.id === scope.me.id) {
+            if (newAccess) {
+              scope.library.membership.access = newAccess;
+            } else {
+              scope.library.membership = null;
+            }
+          }
+
```
2is10, line 87: This is in case a library owner transfers ownership to someone else?

kifi-backend/modules/shoebox/angular/src/libraries/libraryHeader.styl
2is10: As we discussed in person, I think the cleaner long-term solution is to move this component (and all the buttons in this space) out of the library header and into libraryKeeps.js.

2is10:
Nicely done! :+1:

kifi-backend/modules/shoebox/angular/src/libraries/libraryMembers.js
ahsu1230, line 87: This is in the case when you open the manageMembersModal (as a collaborator) and you update/remove your own membership.
Then `scope.library.membership` should change accordingly

2is10, line 87: I see. And if we decide to allow library owners to transfer ownership, this would handle that too. Got it, thanks.

PR #15579, opened 2015-05-28
libraryInviteRepo.getWithLibraryIdAndUserId

ahsu1230:
@andrewconner @2is10 

@2is10 found a bug where you can still see a private library even though you declined all library invites (no active invites).
Before, `getWithLibraryIdAndUserId()` only excluded `LibraryInviteStates.INACTIVE`. This change has it look only for active library invites (so exclude inactive, accepted, and declined invites)


2is10:
Thanks.

PR #15580, opened 2015-05-28
have invite followers buttons use new invite modal (only behind colla…

ahsu1230:
…b experiment)

@andrewconner @2is10


2is10:
:+1:

PR #15591, opened 2015-05-28
clarifying methods for choosing followers and collaborators for display

2is10:
@ahsu1230


kifi-backend/modules/shoebox/app/com/keepit/model/LibraryView.scala
```diff
@@ -170,17 +170,12 @@ case class LibraryCardInfo(
     extends BaseLibraryCardInfo(id, name, description, color, image, slug, owner, numKeeps, numFollowers, followers, numCollaborators, collaborators, lastKept, following, membership, modifiedAt, kind)
 
 object LibraryCardInfo {
-  val writesWithoutOwner = Writes[LibraryCardInfo] { o => // for case when receiving end already knows the owner
-    JsObject((Json.toJson(o).as[JsObject].value - "owner").toSeq)
+  def chooseCollaborators(collaborators: Seq[BasicUser]): Seq[BasicUser] = {
+    collaborators.sortBy(_.pictureName == "0.jpg").take(3) // owner + 2 collaborators shown, 1 extra in case viewer is one and leaves
```
2is10, line 174: This puts collaborators with pictures first, since `false < true`. It also maintains the original ordering of the two partitions: all `Seq` sorting methods are guaranteed to be “stable”.

ahsu1230, line 174: nice. thanks!

PR #15592, opened 2015-05-28
Atom feed

Colin-Lane:
Initial Atom
- atom has a logo field, didn't know where to get the logo from.
- also atom can add contributors as an xml field (didn't add that to keep it the same as RSS)
- still need licensing in the rights field.

@andrewconner @stkem

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/common/seo/AtomCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/common/seo/AtomCommander.scala
@@ -86,7 +86,7 @@ class AtomCommander @Inject() (
     val descriptionsFuture = db.readOnlyMaster { implicit s => rover.getUriSummaryByUris(keeps.map(_.uriId).toSet) }
     descriptionsFuture map { descriptions =>
       val entries = keeps.map { keep =>
-        val (keepImage, keeper) = db.readOnlyMaster { implicit s =>
+        val (keepImageOpt, keeper) = db.readOnlyMaster { implicit s =>
           val image = keepImageCommander.getBestImageForKeep(keep.id.get, ScaleImageRequest(ImageSize(100, 100)))
           (image, userRepo.getNoCache(keep.userId))
         }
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/shoebox/app/com/keepit/common/seo/AtomCommander.scala
```diff
@@ -0,0 +1,107 @@
...
+import com.keepit.model._
+import com.keepit.rover.RoverServiceClient
+import org.joda.time.DateTime
+import scala.concurrent.ExecutionContext.Implicits.global
+
+import scala.concurrent.Future
+import scala.xml.{ Elem, Text }
+
+/**
+ * Created by colinlane on 5/28/15.
```
stkem: We don't usually do this. That is what `git blame` is for :-)

kifi-backend/modules/shoebox/app/com/keepit/common/seo/AtomCommander.scala
```diff
@@ -0,0 +1,107 @@
...
+import com.keepit.commanders.{ KeepImageCommander, ScaleImageRequest, LibraryImageCommander }
+import com.keepit.common.db.slick.Database
+import com.keepit.common.logging.Logging
+import com.keepit.common.store.{ ImageSize, S3ImageConfig }
+import com.keepit.common.time._
+import com.keepit.inject.FortyTwoConfig
+import com.keepit.model._
+import com.keepit.rover.RoverServiceClient
+import org.joda.time.DateTime
+import scala.concurrent.ExecutionContext.Implicits.global
```
stkem: Please use the play execution context `play.api.libs.concurrent.Execution.Implicits._`, not the default scala one. Or, even better, inject an execution context if that is possible here.

stkem: The scala execution context is not very configurable, whereas the play one is configured by us in the application config.

kifi-backend/modules/shoebox/test/com/keepit/common/seo/AtomCommanderTest.scala
```diff
@@ -0,0 +1,128 @@
...
+import com.keepit.test.ShoeboxTestInjector
+import org.joda.time.DateTime
+import org.specs2.mutable.Specification
+
+import scala.concurrent.Await
+import scala.concurrent.duration.Duration
+import scala.xml.NodeSeq
+
+/**
+ * Created by colinlane on 5/28/15.
```
stkem: Ditto, my comment on the commander

stkem:
made a few comments. lgtm otherwise. Don't worry about logo etc. to start with. We'll add those on second pass.
@andrewconner do you know the status on that license?

kifi-backend/modules/shoebox/app/com/keepit/common/seo/AtomCommander.scala
Colin-Lane: Haha, that was autocreated by Intellij upon class creation. I'll mess with the class creation templates.

PR #15593, opened 2015-05-28
responsive collab pics & followers

ahsu1230:
@andrewconner @2is10


ahsu1230:
responsive list of collaborators and responsive followers on upper right of library header

2is10:
:+1:

PR #15595, opened 2015-05-29
Webhook Support

camhashemi:
@stkem


kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryWebhookCommander.scala
```diff
@@ -0,0 +1,53 @@
...
+
+class LibraryWebhookCommander @Inject() (
+    db: Database,
+    httpClient: HttpClient,
+    libraryWebhookRepo: LibraryWebhookRepo,
+    userRepo: UserRepo,
+    implicit val executionContext: ExecutionContext,
+    airbrake: AirbrakeNotifier) extends Logging {
+
+  def sendNewKeepWebhook(bookmark: RawBookmarkRepresentation, userId: Id[User], library: Library) = SafeFuture {
```
andrewconner: How about take in a `Keep`? We only want it when it turned into a real `Keep`, right?

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryWebhookCommander.scala
```diff
@@ -0,0 +1,53 @@
...
+    val webhooks = db.readOnlyReplica { implicit session =>
+      library.id.map { id => libraryWebhookRepo.getByLibraryIdAndTrigger(id, WebhookTrigger.NEW_KEEP) }
+    }.getOrElse(throw new NoSuchFieldException) // not sure what to do in case library.id == None
```
andrewconner: Actually, in this case, don't worry about it. `library.id.get` is fine. This is against all other suggestions about `Option`s, but for our models, all models have an ID, and unpersisted objects don't. If we somehow give an unpersisted object to this function, just fail hard.

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryWebhookRepo.scala
```diff
@@ -0,0 +1,49 @@
...
+class LibraryWebhookRepoImpl @Inject() (val db: DataBaseComponent, val clock: Clock) extends DbRepo[LibraryWebhook] with LibraryWebhookRepo with Logging {
+  import db.Driver.simple._
+
+  type RepoImpl = LibraryWebhookTable
+
+  class LibraryWebhookTable(tag: Tag) extends RepoTable[LibraryWebhook](db, tag, "library_webhook") {
+
+    def trigger = column[WebhookTrigger]("action", O.NotNull)
+    def libraryId = column[Id[Library]]("library_id", O.NotNull)
+    def action = column[JsValue]("dest", O.NotNull)
```
stkem: `JsObject` maybe, for forwards compatibility? Even if what you need is just a string wrapping it in an object can help, because you can just add things later. If you are just storing it as a string and need to add something you'll have problems with old data in the db.
All this is of course possible with `JsValue` as the type, but `JsObject` would enforce it.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryWebhookCommander.scala
andrewconner: (Disclaimer: I started this pattern, and incoming wall of text that hopefully I can reuse elsewhere when I see this). `SafeFuture` has turned into a warning sign for me, and in this case, it's not awful. However, something to think about... There's two reasons why you may want to use one:
1. You want exception reporting for free for your `Future`, because you're not handling it elsewhere or don't expect anyone to handle it. So, your errors don't really change anything.
2. You want to fire-and-forget a block of code, thus probably aren't handling errors (and likely don't really care, it doesn't change anything if it fails, since nothing after depends on it).

For (1), it's a warning sign if you're returning a future and aren't caring about errors. Sometimes that means you don't actually expect the caller to do anything with the future. However, this case is usually the most justified.

For (2), it means that your function side effects (which makes things harder to test and reason about), and that it's doing something that takes long enough that you don't want to run in the calling thread, BUT there's no rate limiting or back pressure. These bite us often. Reactive lock is a decent (better) solution.

Yours is closer to (1) (you just want errors to be handled just in case), but actually, we may want to use a reactive lock to rate limit these outgoing calls.

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryWebhookRepo.scala
andrewconner: `JsObject` is better

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryWebhook.scala
```diff
@@ -0,0 +1,47 @@
+package com.keepit.model
+
+import com.keepit.common.db._
+import com.keepit.common.time._
+import org.joda.time.DateTime
+import play.api.libs.functional.syntax._
+import play.api.libs.json._
+import scala.slick.lifted.MappedTo
+
+case class LibraryWebhook(
```
andrewconner: How about opening this to all actions, not just webhooks?

andrewconner: Very cheaply, we can support future use such as "email me when a keep is added", without actually having to design that.

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryWebhook.scala
```diff
@@ -0,0 +1,47 @@
...
+    id: Option[Id[LibraryWebhook]] = None,
+    createdAt: DateTime = currentDateTime,
+    updatedAt: DateTime = currentDateTime,
+    state: State[LibraryWebhook] = State("active"),
+    trigger: WebhookTrigger,
+    libraryid: Id[Library],
+    action: JsValue) extends ModelWithState[LibraryWebhook] {
```
andrewconner: Type this!

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryWebhookCommander.scala
```diff
@@ -0,0 +1,53 @@
...
+
+    val keeperName = db.readOnlyReplica { implicit session => userRepo.get(userId).fullName }
+
+    val client = httpClient.withTimeout(CallTimeouts(responseTimeout = Some(2 * 60 * 1000), maxJsonParseTime = Some(20000))) // copied from FacebookPublishingCommander, can tweak
+
+    webhooks.map { webhook =>
+      webhook.action.\("sendThrough").toString match { // the "sendThrough" field could be "slack", "email", "eliza", etc... perhaps there's a better name
```
andrewconner: If this was typed, easier to match here.

andrewconner: Also, never `.toString` `JsValue`s. They'll be formatted oddly. Check out `.as[String]`, which can do cool things like `.as[Seq[String]]` for free.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryWebhookCommander.scala
camhashemi: 1. I was initially going to call this from KeepInterner.internRawBookmark. This gets passed a RawBookmark, and it's the first place where we can say "hey guys, somebody kept something". Then we make a call to notify followers (via eliza), so I intended on modeling the call after that.
2. I think it's more like (2), if you look at the (proposed) calling function above, but all I know about SafeFutures is that it offers free exception reporting. I'll look into ReactiveLock.

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryWebhook.scala
```diff
@@ -0,0 +1,47 @@
...
+    (__ \ 'libraryId).format[Id[Library]] and
+    (__ \ 'action).format[JsValue])(LibraryWebhook.apply _, unlift(LibraryWebhook.unapply _)) // json de/serialization
+}
+
+case class WebhookTrigger(value: String) extends MappedTo[String] {
+  override def toString = value
+}
+
+object WebhookTrigger {
+  val NEW_KEEP = WebhookTrigger("new_keep")
```
andrewconner: Any idea what other triggers we may have? May help design the best API.

stkem: `NEW_FOLLOWER`, `NEW_COLLBORATOR`, `INVITATION_SENT`. Possibly so far as `VIEWED`?
When we have it: `NEW_COMMENT`
Just some ideas.

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryWebhook.scala
stkem: It looks like the architecture already supports this (i.e. the free form json action). The name is just lagging behind :-)

camhashemi: refactoring the class LibrarySubscription. "action" has been refactored to "info" of class SubscriptionInfo. let me know if this may be confusing. new PR on the way.

camhashemi:
@stkem updated.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibrarySubscriptionCommander.scala
```diff
@@ -0,0 +1,52 @@
...
+
+class LibrarySubscriptionCommander @Inject() (
+    db: Database,
+    httpClient: HttpClient,
+    librarySubscriptionRepo: LibrarySubscriptionRepo,
+    userRepo: UserRepo,
+    implicit val executionContext: ExecutionContext,
+    airbrake: AirbrakeNotifier) extends Logging {
+
+  val httpLock = new ReactiveLock(5)
```
stkem, line 32: This will only really work if you make the class `@Singleton`. Otherwise Guice will make a new instance (with a new lock) every time it is injected.

stkem, line 32: By class, I mean the `LibrarySubscriptionCommander`.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibrarySubscriptionCommander.scala
```diff
@@ -0,0 +1,52 @@
...
+
+  def sendNewKeepMessage(bookmark: RawBookmarkRepresentation, keeperId: Id[User], library: Library) = Future {
```
stkem, line 34: This is not a hard rule, but I think it is a good idea to have an explicit return type on any function longer than a few lines. Makes it much easier to read the code.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibrarySubscriptionCommander.scala
```diff
@@ -0,0 +1,52 @@
...
+    val keeperName = db.readOnlyReplica { implicit session => userRepo.get(keeperId).fullName }
+
+    val client = httpClient.withTimeout(CallTimeouts(responseTimeout = Some(2 * 60 * 1000), maxJsonParseTime = Some(20000)))
+
+    subscriptions.map { subscription =>
+      subscription.info match {
+        case info: SlackInfo =>
+          val text = keeperName + " just added a <$bookmark.url|keep> to the <$library.url|" + library.name + "> library." // slack hypertext uses the < url | text > format
+          val body = BasicSlackMessage(text)
+          httpLock.withLock(client.postFuture(DirectUrl(info.url), Json.toJson(body)))
```
stkem, line 46: you need `withLockFuture` here. Somewhat confusingly, `withLock` does nothing if the body is a Future.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibrarySubscriptionCommander.scala
stkem, line 34: Looking at it, it seems the return type is `Future[Seq[Future[ClientResponse]`, which I'm guessing is not what you want. Explicit return types help with catching that.

stkem, line 34: Also, as I think @andrewconner suggested before, I think it would be better to take in a `Keep` here, so you know it's fully persisted.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibrarySubscriptionCommander.scala
```diff
@@ -0,0 +1,52 @@
...
+    airbrake: AirbrakeNotifier) extends Logging {
+
+  val httpLock = new ReactiveLock(5)
+
+  def sendNewKeepMessage(bookmark: RawBookmarkRepresentation, keeperId: Id[User], library: Library) = Future {
+    val subscriptions = db.readOnlyReplica { implicit session => librarySubscriptionRepo.getByLibraryIdAndTrigger(library.id.get, SubscriptionTrigger.NEW_KEEP) }
+
+    val keeperName = db.readOnlyReplica { implicit session => userRepo.get(keeperId).fullName }
+
+    val client = httpClient.withTimeout(CallTimeouts(responseTimeout = Some(2 * 60 * 1000), maxJsonParseTime = Some(20000)))
```
stkem, line 39: This might be better off at the class level. No need to do this to the client on every call to this method.

kifi-backend/modules/shoebox/app/com/keepit/model/LibrarySubscriptionRepo.scala
```diff
@@ -0,0 +1,62 @@
...
+    def * = (id.?, createdAt, updatedAt, state, trigger, libraryId, info) <> ((LibrarySubscription.apply _).tupled, LibrarySubscription.unapply _)
+  }
+
+  def table(tag: Tag) = new LibrarySubscriptionTable(tag)
+  initTable()
+
+  def deleteCache(model: LibrarySubscription)(implicit session: RSession): Unit = {}
+  def invalidateCache(model: LibrarySubscription)(implicit session: RSession): Unit = {}
+
+  def getByLibraryId(id: Id[Library])(implicit session: RSession): Seq[LibrarySubscription] = {
```
stkem, line 55: In most cases you probably only want `ACTIVE` entries. Look at KeepRepo.scala and search for `excludeStates` to see how we do this usually.

kifi-backend/modules/shoebox/app/com/keepit/model/LibrarySubscriptionRepo.scala
```diff
@@ -0,0 +1,62 @@
...
+    (for (c <- rows if c.libraryId === id) yield c).list
+  }
+
+  def getByLibraryIdAndTrigger(id: Id[Library], trigger: SubscriptionTrigger)(implicit session: RSession): Seq[LibrarySubscription] = {
```
stkem, line 59: Same here re `ACTIVE`

kifi-backend/modules/shoebox/app/com/keepit/model/LibrarySubscription.scala
```diff
@@ -0,0 +1,76 @@
...
+import play.api.libs.functional.syntax._
+import play.api.libs.json._
+import scala.slick.collection.heterogenous.Zero.+
+import scala.slick.lifted.MappedTo
+
+case class LibrarySubscription(
+    id: Option[Id[LibrarySubscription]] = None,
+    createdAt: DateTime = currentDateTime,
+    updatedAt: DateTime = currentDateTime,
+    state: State[LibrarySubscription] = State("active"),
```
stkem, line 17: You shouldn't deal with explicit strings in these situations. Error prone. User something like `LibrarySubscriptionStates.ACTIVE` (see other subclasses of `ModelWithState`)

kifi-backend/modules/shoebox/app/com/keepit/model/LibrarySubscription.scala
```diff
@@ -0,0 +1,76 @@
...
+    trigger: SubscriptionTrigger,
+    libraryid: Id[Library],
+    info: SubscriptionInfo) extends ModelWithState[LibrarySubscription] {
+  def withId(id: Id[LibrarySubscription]) = this.copy(id = Some(id))
+  def withUpdateTime(now: DateTime) = this.copy(updatedAt = now)
+  def withState(state: State[LibrarySubscription]) = this.copy(state = state)
+  def isActive = this.state.value == "active"
```
stkem, line 24: Looks like this is not used.

kifi-backend/modules/shoebox/app/com/keepit/model/LibrarySubscription.scala
```diff
@@ -0,0 +1,76 @@
...
+    (__ \ 'updatedAt).format[DateTime] and
+    (__ \ 'state).format[State[LibrarySubscription]] and
+    (__ \ 'trigger).format[SubscriptionTrigger] and
+    (__ \ 'libraryId).format[Id[Library]] and
+    (__ \ 'info).format[SubscriptionInfo])(LibrarySubscription.apply _, unlift(LibrarySubscription.unapply)) // json de/serialization
+}
+
+trait SubscriptionInfo
+
+case class SlackInfo(kind: String = "slack", url: String) extends SubscriptionInfo
```
stkem, line 40: The kind here is unnecessary. It's already in the type of the class. (And yes, removing that means you can't just use apply and unapply in the `format`.)

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibrarySubscriptionCommander.scala
```diff
@@ -0,0 +1,52 @@
...
+
+import com.google.inject.Inject
+import com.keepit.common.akka.SafeFuture
+import com.keepit.common.db.Id
+import com.keepit.common.db.slick.Database
+import com.keepit.common.healthcheck.AirbrakeNotifier
+import com.keepit.common.logging.Logging
+import com.keepit.common.net.{ DirectUrl, CallTimeouts, HttpClient }
+import com.keepit.common.concurrent.ReactiveLock
+import com.keepit.model._
```
stkem, line 11: Not a huge fan of `_` imports if you can help it. Makes the code harder to read because it's not immediately obvious where things are coming from.

stkem:
Nicely abstracted! Made some comments on lower level patterns, mostly. Please also add tests.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibrarySubscriptionCommander.scala
camhashemi, line 34: It looks like the first place where we know that a keep is new is KeepInterner[136]. I would prefer to pass a Keep, but this is how the keep notification system seems to currently work.

That being said, we haven't yet specified where sendNewKeepMessage will be called from.

stkem, line 34: Discussed in person. Solution found :-)

camhashemi, line 34: I'm going to work on my "can't do" attitude... thank you Stephen.

stkem, line 34: No worries :-)
And you don't really have a "can't do" attitude.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibrarySubscriptionCommander.scala
andrewconner, line 11: This is a tradeoff though. At some point, it can become ridiculous :)

camhashemi, line 11: No need to import com.keepit.model._ though haha. I'm probably only using a few classes from it.

PR #15597, opened 2015-05-29
toggle function on library invites

ahsu1230:
@andrewconner @2is10 

it wasn't clear enough that you could invite to collaborate or follow. So here's a toggle-like functionality that helps users go in between the two.


2is10:
cool

PR #15601, opened 2015-05-29
moving from Clutch to ExpiringCache for immutable cache values, $http interoperability, and less code

2is10:
@andrewconner @ahsu1230 

Like it? It works, but I haven’t written tests yet. Didn’t want to spend too much time on it given our priorities.

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/libraryService.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/libraryService.js
@@ -3,61 +3,14 @@
 angular.module('kifi')
 
 .factory('libraryService', [
-  '$http', '$rootScope', 'profileService', 'routeService', 'Clutch', '$q', '$analytics',
-  function ($http, $rootScope, profileService, routeService, Clutch, $q, $analytics) {
+  '$http', '$rootScope', 'profileService', 'routeService', '$q', '$analytics', 'net',
+  function ($http, $rootScope, profileService, routeService, $q, $analytics, net) {
     var infos = {
       own: []
     };
     var recentIds = [];  // in-memory cache, length limited, most recent first
 
     // TODO: flush any non-public cached data when a user logs out.
-
-    //
-    // Clutches.
-    //
-
-    var libraryInfosClutch = new Clutch(function () {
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/libraryService.js
```diff
@@ -198,8 +162,10 @@ angular.module('kifi')
         return $http.post(routeService.modifyLibrary(opts.id), opts);
       },
 
-      getLibraryShareContacts: function (libId, opt_query) {
-        return contactSearchClutch.get(libId, opt_query || '');
+      getLibraryShareContacts: function (libId, query) {
+        return net.getLibraryShareSuggest(libId, {q: query || []}).then(function (res) {
+          return res.data.members;
+        });
```
2is10, line 168: I have mixed feelings about requiring `net` method callers to use actual query parameter names (`q` in this case) and to know Angular’s query parameter serialization quirks, such as using `[]` to get a parameter to be omitted. It feels like a leaky abstraction.

One way to solve this would be to extend the route syntax in `net` to handle parameterized query parameter strings, like: `'/libraries/:id/members/suggest?n=30&q=:query'`, so callers could just pass query parameter values directly in the argument list, like `net.getLibraryShareSuggest(libId, query)`.

andrewconner:
Sorry, this has been open for a while. I don't know much about `$cacheFactory`, but trust your judgement here :).

2is10:
I think the main benefit is that this newer pattern removes a layer of indirection (extra code) that’s currently required for creating clutches (see the libraryService.js diff).

In fact, it may allow us to eliminate many of our simpler services entirely, so controllers can directly make cached remote calls via `net`. I think some of our simpler services (“action services”) primarily exist to stitch together `$http`, `routeService`, and sometimes `Clutch`, and now `net` can handle all three responsibilities.

As I explained in an earlier comment, we can tweak our `net` call syntax until we arrive at something we’re entirely happy with. I’m currently thinking something like Scala’s call syntax would be cool, allowing all of a route’s path and query string parameters to be specified either as a simple sequential parameter list _or_ named parameters (as a JS object) _or_ a mix of both, caller’s preference.

PR #15602, opened 2015-05-29
grat emails still not sending, increasing frequency

camhashemi:
@stkem sorry for all the incoming PRs. I think the way we get "users to send to" is not working, playing with it.


yingjieMiao:
may be helpful to add logging in the `GratificationEmailActor`, to see the actor indeed receives cron messages.

camhashemi:
it's receiving cron messages, just got a batch of emails from the experiment condition. this means that GratificationCommander.usersToSendTo (or its caller) is not working.

camhashemi:
will start logging

PR #15603, opened 2015-05-29
removing memberships on the library header

ahsu1230:
@andrewconner @2is10 

Thanks @camhashemi for the suggestions

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/deleteMemberConfirmModal.tpl.html
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/deleteMemberConfirmModal.tpl.html
@@ -1,0 +1,5 @@
+<div kf-modal class="kf-remove-member-confirm-modal">
+  <div kf-basic-modal-content with-cancel="true" title="Are you sure?" with-warning action-text="Remove Member" action="removeMember()">
+    <p>Removing collaborators will prevent them from being able to add, edit or remove keeps in this library</p>
+  </div>
+</div>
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/deleteMemberConfirmModal.tpl.html
```diff
@@ -0,0 +1,5 @@
+<div kf-modal class="kf-remove-member-confirm-modal">
+  <div kf-basic-modal-content with-cancel="true" title="Are you sure?" with-warning action-text="Remove Member" action="removeMember()">
+    <p>Removing collaborators will prevent them from being able to add, edit or remove keeps in this library</p>
+  </div>
+</div>
```
2is10: This file appears to be unused.

ahsu1230:
@2is10 i was trying out the dropdown menu (to both remove or demote a collaborator to a follower), and it seemed really super weird when it was on the library header. I think the best UI is honestly to go with a small `x` on top of a user's image and then a confirmation modal once clicked.

2is10:
:+1:

kifi-backend/modules/shoebox/angular/src/libraries/deleteMemberConfirmModal.tpl.html
ahsu1230: woops you're right. Missed that =/

PR #15606, opened 2015-05-30
Fixing stability issues with calling heimdal too many times

andrewconner:
@stkem @camhashemi I know you're working on this class, just throwing in a fail safe to stop floods of exceptions we're seeing with calls to Heimdal.


stkem:
:+1:

stkem:
Thanks!

PR #15608, opened 2015-05-30
DO NOT MERGE - Teams Backend Models

Colin-Lane:
@leogrim


yingjieMiao:
WOW. that is fast.

mrbrown14:
Incredibly!

On May 29, 2015, at 17:42, yingjieMiao notifications@github.com wrote:

WOW. that is fast.

kifi-backend/modules/common/app/com/keepit/model/Organization.scala
```diff
@@ -0,0 +1,108 @@
...
+      org.state,
+      org.seq,
+      org.name,
+      org.description,
+      org.ownerId,
+      org.slug)
+  }
+}
+
+case class OrganizationSlug(value: String) {
```
leogrim, line 44: `Organization` shares the namespace of `User`, so we need to borrow from `Username` instead of `LibrarySlug` - that especially includes the username normalization logic. Perhaps we should call this `Orgname` for consistency.

kifi-backend/modules/common/app/com/keepit/model/OrganizationInvite.scala
```diff
@@ -0,0 +1,63 @@
...
+    emailAddress: Option[EmailAddress] = None,
+    access: OrganizationAccess,
+    message: Option[String] = None) extends ModelWithPublicId[OrganizationInvite] with ModelWithState[OrganizationInvite] {
+
+  def withId(id: Id[OrganizationInvite]): OrganizationInvite = this.copy(id = Some(id))
+  def withUpdateTime(now: DateTime): OrganizationInvite = this.copy(updatedAt = now)
+  def withState(newState: State[OrganizationInvite]): OrganizationInvite = this.copy(state = newState)
+
+  def isCollaborator = (access == OrganizationAccess.READ_WRITE) || (access == OrganizationAccess.READ_INSERT)
+  def isFollower = (access == OrganizationAccess.READ_ONLY)
```
leogrim: There's no concept of `Follower` vs `Collaborator` for an organization, we can stick with two permission levels at this point: `OWNER` and `READ_WRITE`

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationLogo.scala
```diff
@@ -0,0 +1,60 @@
...
+    (__ \ 'format).format[ImageFormat] and
+    (__ \ 'kind).format[ProcessImageOperation] and
+    (__ \ 'imagePath).format[ImagePath] and
+    (__ \ 'source).format[ImageSource] and
+    (__ \ 'sourceFileHash).format[ImageHash] and
+    (__ \ 'sourceImageURL).formatNullable[String]
+  )(OrganizationLogo.apply, unlift(OrganizationLogo.unapply))
+}
+
+case class ImagePosition(x: Int, y: Int)
```
leogrim, line 51: :+1:

kifi-backend/modules/shoebox/test/com/keepit/model/OrganizationRepoTest.scala
```diff
@@ -0,0 +1,23 @@
...
+          orgRepo.save(Organization(name = "Bob", ownerId = Id[User](1), slug = OrganizationSlug("slug")))
+        }
+
+        db.readOnlyMaster { implicit s =>
+          orgRepo.get(org.id.get) === org
+        }
+      }
+    }
+  }
+}
```
leogrim, line 23: We define evolutions for these repos. Once we do, we need to use an `Application` for these initial tests. Indeed while `ShoeboxTestInjector` makes for faster tests, it relies on Slick to initialize the database tables, which means our evolutions are not actually tested. See `RoverRepoTest` for the latest good practice. I should add something about this somewhere in Confluence :)

leogrim:
Fast start :)

kifi-backend/modules/common/app/com/keepit/model/Organization.scala
```diff
@@ -0,0 +1,108 @@
...
+    seq: SequenceNumber[Organization],
+    name: String,
+    description: Option[String],
+    ownerId: Id[User],
+    slug: OrganizationSlug) = {
+    Organization(id, createdAt, updatedAt, state, seq, name, description, ownerId, slug)
+  }
+
+  def unapplyToDbRow(org: Organization) = {
+    Some(
```
andrewconner: A.A.A.A.T.: Always Avoid Accidental Automatic Tupling. :)

Scala has this weird "feature" if a function takes a single generic parameter (two examples: `def f(a: Any)` and `def f[T](a: T)`), and you pass in multiple arguments, the compiler automatically Tuples them. So instead of `Some(a, b)` — which doesn't make sense because `Some.apply` only takes one argument) — it accepts it and compiles as `Some((a,b))`, which is sugar for `Some(Tuple2(a,b))`.

This feature is the source of quite a few bugs people have, so is being removed in the next Scala version.

kifi-backend/modules/common/app/com/keepit/model/Organization.scala
andrewconner, line 44: Agreed. I haven't finished reviewing, but if it helps, `username` and `slug` are a bit different semantically as well. Usernames happen to show up in URLs, but could be used elsewhere. Library slugs are intended just for URLs.

kifi-backend/modules/common/app/com/keepit/model/Organization.scala
```diff
@@ -0,0 +1,108 @@
...
+}
+
+object OrganizationSlug {
+  implicit def format: Format[OrganizationSlug] =
+    Format(__.read[String].map(OrganizationSlug(_)), new Writes[OrganizationSlug] {
+      def writes(o: OrganizationSlug) = JsString(o.value)
+    })
+
+  private val MaxLength = 50
+  private val ReservedSlugs = Set("libraries", "connections", "followers", "keeps", "tags")
```
andrewconner, line 55: Perhaps we should combine Username restrictions (see `UsernameOps`) and this?

kifi-backend/modules/common/app/com/keepit/model/OrganizationInvite.scala
```diff
@@ -0,0 +1,63 @@
...
+
+  implicit def ord: Ordering[OrganizationInvite] = new Ordering[OrganizationInvite] {
+    def compare(x: OrganizationInvite, y: OrganizationInvite): Int = x.access.priority compare y.access.priority
+  }
+}
+
+object OrganizationInviteStates extends States[OrganizationInvite] {
+  val ACCEPTED = State[OrganizationInvite]("accepted")
+  val DECLINED = State[OrganizationInvite]("declined")
+}
```
andrewconner: We have a convention of new line at the end of a file. Some reasons: http://stackoverflow.com/questions/729692/why-should-files-end-with-a-newline

kifi-backend/modules/common/app/com/keepit/model/OrganizationMembership.scala
```diff
@@ -0,0 +1,116 @@
...
+    })
+
+  implicit def ord: Ordering[OrganizationAccess] = new Ordering[OrganizationAccess] {
+    def compare(x: OrganizationAccess, y: OrganizationAccess): Int = x.priority compare y.priority
+  }
+
+  def apply(str: String): OrganizationAccess = {
+    str match {
+      case READ_ONLY.value => READ_ONLY
+      case READ_INSERT.value => READ_INSERT
```
andrewconner: I don't think `READ_INSERT` should be here. It was probably an over-engineering for libraries. Do we have any expectation of needing this soon?

kifi-backend/modules/common/app/com/keepit/model/OrganizationMembership.scala
```diff
@@ -0,0 +1,116 @@
...
+      case READ_WRITE.value => READ_WRITE
+      case OWNER.value => OWNER
+    }
+  }
+
+  def all: Seq[OrganizationAccess] = Seq(OWNER, READ_WRITE, READ_INSERT, READ_ONLY)
+
+  def collaborativePermissions: Set[OrganizationAccess] = Set(OWNER, READ_WRITE, READ_INSERT)
```
andrewconner: I see references to collaboration in a couple places. Does that make sense for organizations?

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationLogo.scala
andrewconner, line 51: Was this not needed elsewhere?

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationMembershipRepo.scala
```diff
@@ -0,0 +1,33 @@
...
+  override def deleteCache(orgMember: OrganizationMembership)(implicit session: RSession) {}
+  override def invalidateCache(orgMember: OrganizationMembership)(implicit session: RSession) {}
+
+  import db.Driver.simple._
+
+  type RepoImpl = OrganizationMembershipTable
+  class OrganizationMembershipTable(tag: Tag) extends RepoTable[OrganizationMembership](db, tag, "organization_membership") with SeqNumberColumn[OrganizationMembership] {
+    def organizationId = column[Id[Organization]]("organization_id", O.NotNull)
+    def userId = column[Id[User]]("user_id", O.NotNull)
+    implicit val organizationAccessMapper = MappedColumnType.base[OrganizationAccess, String](_.value, OrganizationAccess(_))
```
andrewconner: May be best to separate this away from the column definitions. Makes them harder to read. At the top of the table definition, perhaps?

kifi-backend/modules/common/app/com/keepit/model/Organization.scala
andrewconner, line 44: One difference, regarding naming: Organizations have names ie, "IBM Engineering" and parts in the URL: "ibm-eng". Not sure where this leaves us.

kifi-backend/modules/common/app/com/keepit/model/Organization.scala
leogrim, line 55: Yes, I think we should come up with a name that works for both `Username` and `OrganizationName`. Suggestions? `Handle`? `Alias`?

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationLogo.scala
leogrim, line 51: There wasn't any type for it, `UserPicture` has it denormalized as just `positionX` and `positionY`.

kifi-backend/modules/common/app/com/keepit/model/Organization.scala
Colin-Lane, line 55: Alias sounds like there can be multiple names. I prefer Handle.

kifi-backend/modules/shoebox/test/com/keepit/model/ShoeboxRepoTest.scala
```diff
@@ -0,0 +1,24 @@
...
+        val organizationRepo = inject[OrganizationRepo]
+        db.readWrite { implicit session =>
+          val org = Organization(id = Some(Id(2)), name = "OrgName", ownerId = Id[User](3), slug = OrganizationSlug("slug"))
+          val saved = organizationRepo.save(org)
+          organizationRepo.get(saved.id.get).id.get === 2
+        }
+      }
+    }
+  }
+}
```
leogrim, line 59: @andrewconner @stkem good intern project: add all Shoebox repos in this test. Have a similar test for Eliza, Cortex etc..

kifi-backend/modules/common/app/com/keepit/model/OrganizationInvite.scala
leogrim: @andrewconner would you suggest removing these states? Perhaps having `response: Option[Boolean]`  instead?

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationRepo.scala
```diff
@@ -0,0 +1,74 @@
...
+@ImplementedBy(classOf[OrganizationRepoImpl])
+trait OrganizationRepo extends Repo[Organization] with SeqNumberFunction[Organization] {
+  /*** Accessing Organizations ***/
+  def listLibraries(organizationId: Id[Organization]) = ???
+
+  /*** Modifying Organizations ***/
+  def createOrganization(ownerId: Id[User], teamName: String, description: Option[String], slug: OrganizationSlug) = ???
+  def addLibraries(organizationId: Id[Organization], libraries: Seq[Id[Library]]): Seq[Library] = ???
+  def addLibrary(organizationId: Id[Organization], libraryId: Id[Library]): Library = ???
+  def removeLibrary(libraryId: Id[Library]): Library = ???
```
leogrim, line 21: A repo only manages its model. Libraries are unrelated to the `Organization` model so all of these library-related methods belong elsewhere (most likely, we need to update existing methods in `LibraryCommander`)

kifi-backend/modules/common/app/com/keepit/model/OrganizationInvite.scala
andrewconner: It's fine, you'll have to either always deal with `response`, or if you use states, dealing with discarding `ACCEPTED` and `DECLINED`

kifi-backend/modules/common/app/com/keepit/model/OrganizationInvite.scala
```diff
@@ -0,0 +1,59 @@
...
+  def withUpdateTime(now: DateTime): OrganizationInvite = this.copy(updatedAt = now)
+  def withState(newState: State[OrganizationInvite]): OrganizationInvite = this.copy(state = newState)
+
+  override def toString: String = s"OrganizationInvite[id=$id,organizationId=$organizationId,ownerId=$inviterId,userId=$userId,email=$emailAddress,access=$access,state=$state]"
+
+}
+
+object OrganizationInvite extends ModelWithPublicIdCompanion[OrganizationInvite] {
+
+  protected[this] val publicIdPrefix = "l"
```
andrewconner, line 35: change to something else, like `"o"`

kifi-backend/modules/common/app/com/keepit/model/OrganizationInvite.scala
leogrim: Sure it works either way, but having an explicit field prevents from overloading states. I believe most "leaks" come from not being aware of additional model states (easy to miss, they're not defined under a `sealed trait`).

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationRepo.scala
Colin-Lane, line 21: agreed. I'll remove them from here.

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationLogoRepo.scala
```diff
@@ -0,0 +1,85 @@
...
+import com.keepit.common.db.slick.DBSession.RSession
+import com.keepit.common.db.slick.{ DataBaseComponent, DbRepo, Repo }
+import com.keepit.common.db.{ Id, State }
+import com.keepit.common.store.ImagePath
+import com.keepit.common.time.Clock
+import org.joda.time.DateTime
+
+@ImplementedBy(classOf[OrganizationLogoRepoImpl])
+trait OrganizationLogoRepo extends Repo[OrganizationLogo] {
+  def uploadLogo(organizationId: Id[Organization], position: Option[ImagePosition], width: Int, height: Int,
```
leogrim, line 13: Let's use `ImageSize` in this API

kifi-backend/modules/common/app/com/keepit/model/OrganizationInvite.scala
```diff
@@ -0,0 +1,59 @@
...
+  protected[this] val publicIdIvSpec = new IvParameterSpec(Array(-20, -76, -59, 85, 85, -2, 72, 61, 58, 38, 60, -2, -128, 79, 9, -87))
```
andrewconner, line 36: This should be unique per model that has a public ID. See `ModelWithPublicId.scala:54`:

``` scala
    val sr = new java.security.SecureRandom()
    val arr = new Array[Byte](16)
    sr.nextBytes(arr)
    arr
```

kifi-backend/modules/common/app/com/keepit/model/OrganizationMembership.scala
```diff
@@ -0,0 +1,84 @@
...
+    seq: SequenceNumber[OrganizationMembership] = SequenceNumber.ZERO,
+    organizationId: Id[Organization],
+    userId: Id[User],
+    access: OrganizationAccess) extends ModelWithState[OrganizationMembership] with ModelWithSeqNumber[OrganizationMembership] {
+
+  def withId(id: Id[OrganizationMembership]): OrganizationMembership = this.copy(id = Some(id))
+  def withUpdateTime(now: DateTime): OrganizationMembership = this.copy(updatedAt = now)
+  def withState(newState: State[OrganizationMembership]): OrganizationMembership = this.copy(state = newState)
+
+  override def toString: String = s"OrganizationMembership[id=$id,organizationId=$organizationId,userId=$userId,access=$access,state=$state]"
```
andrewconner, line 24: Why override `toString`? Unexpected. I'm sure this was copied from somewhere, which was copied from somewhere, which may have had a decent reason originally (ie, `User` I can kind of see, since we print `User` instances in logs a ton and we don't care about your profile pic there).

kifi-backend/modules/common/app/com/keepit/model/OrganizationMembership.scala
```diff
@@ -0,0 +1,84 @@
...
+    (__ \ 'organizationId).format[Id[Organization]] and
+    (__ \ 'userId).format[Id[User]] and
+    (__ \ 'access).format[OrganizationAccess]
+  )(OrganizationMembership.apply, unlift(OrganizationMembership.unapply))
+}
+
+object OrganizationMembershipStates extends States[OrganizationMembership]
+
+sealed abstract class OrganizationAccess(val value: String, val priority: Int) {
+  def isHigherAccess(x: OrganizationAccess): Boolean = {
```
andrewconner, line 46: I saw this elsewhere (lib membership?), but sure if it's useful. The comparisons can be made whenever needed directly, and it's not abstracted away (we're comparing _priorities_ — ah!)

kifi-backend/modules/common/app/com/keepit/model/OrganizationMembership.scala
```diff
@@ -0,0 +1,84 @@
...
+    this.priority < x.priority
+  }
+
+  def isLowerOrEqualAccess(x: OrganizationAccess): Boolean = {
+    this.priority <= x.priority
+  }
+}
+
+object OrganizationAccess {
+  case object READ_WRITE extends OrganizationAccess("read_write", 2)
```
andrewconner, line 64: Starting at 2?

kifi-backend/modules/common/app/com/keepit/model/OrganizationMembership.scala
Colin-Lane, line 24: Yeah, I based this model off of LibraryMembership. We don't really need a toString right now.

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationRepo.scala
yingjieMiao, line 21: `LibraryCommander` already contains 1800+ lines of code, 35+ parameters in the constructor.  Maybe it makes sense to have a new commander to handle organization and libraries. If this doesn't make sense, we should still consider refactor the existing `LibraryCommander`.

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationRepo.scala
```diff
@@ -0,0 +1,74 @@
...
+  def changeName(organizationId: Id[Organization], name: String): Organization = ???
+  def changeDescription(organizationId: Id[Organization], description: String): Organization = ???
+  def destroyOrganization(organizationId: Id[Organization]) = ???
```
leogrim, line 24: We do not support organization deletion at this point.

kifi-backend/modules/common/conf/evolutions/shoebox/335.sql
```diff
@@ -0,0 +1,83 @@
...
+	state varchar(20) NOT NULL,
+	seq bigint(20) NOT NULL,
+	name varchar(64) NOT NULL,
+	description text DEFAULT '',
+	owner_id bigint(20) NOT NULL,
+	slug varchar(50) NOT NULL,
+
+	PRIMARY KEY(id),
+	CONSTRAINT `organization_f_user` FOREIGN KEY (`owner_id`) REFERENCES user(`id`),
+	UNIQUE KEY `organization_slug` (slug),
```
andrewconner, line 17: I'm sure you guys are thinking about this, but unique on `slug` isn't strong enough. For example, for users, we have unique on `normalizedUsername`, because we want `Colin-Lane` to own the `colinlane` namespace as well.

kifi-backend/modules/common/conf/evolutions/shoebox/335.sql
```diff
@@ -0,0 +1,83 @@
...
+	x_position smallint unsigned,
+	y_position smallint unsigned,
+	width smallint(5) unsigned NOT NULL,
+	height smallint(5) unsigned NOT NULL,
+	format varchar(16) NOT NULL,
+	kind varchar(32) NOT NULL DEFAULT 'scale',
+	path varchar(64) NOT NULL,
+	source varchar(32) NOT NULL,
+	source_file_hash varchar(32) DEFAULT NULL,
+	source_image_url varchar(3072) DEFAULT NULL,
```
andrewconner, line 69: You think they'll be uploading profile pictures from external URLs? I think I've always set profile pictures using file upload, but could be wrong.

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationMembershipRepo.scala
```diff
@@ -0,0 +1,62 @@
...
+import com.keepit.common.db.slick._
+import com.keepit.common.logging.Logging
+import com.keepit.common.time.Clock
+import org.joda.time.DateTime
+
+@ImplementedBy(classOf[OrganizationMembershipRepoImpl])
+trait OrganizationMembershipRepo extends Repo[OrganizationMembership] with SeqNumberFunction[OrganizationMembership] {
+  def organizationsByUser(userId: Id[User]): Seq[Organization] = ???
+  def listMembers(orgId: Id[Organization]): Seq[Id[User]] = ???
+  def revokeAccess(orgId: Id[Organization], userId: Id[User]) = ???
```
leogrim, line 15: Then `organizationsByUser` should not return a `Seq[Organization]`, otherwise you'd have to join on the `Organization` table which is a different repo. Additionally, we try to name repo methods in a more "REST" style, as in `getByOrganizationId`.

So trying to keep things consistent with `LibraryMembershipRepo`, we can start with:
`def getByOrganizationId(orgId: Id[Organization], excludeState: Option[State[OrganizationMembership]] = Some(OrganizationMembership.INACTIVE))(implicit session: RSession): Seq[OrganizationMembership]
  def getByUserId(userId: Id[User], excludeState: Option[State[OrganizationMembership]] = Some(OrganizationMembershipStates.INACTIVE))(implicit session: RSession): Seq[OrganizationMembership]`

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationInviteRepo.scala
```diff
@@ -0,0 +1,72 @@
...
+import com.keepit.common.db.slick.DBSession.RSession
+import com.keepit.common.db.slick.{ DataBaseComponent, DbRepo, Repo }
+import com.keepit.common.db.{ Id, State }
+import com.keepit.common.mail.EmailAddress
+import com.keepit.common.time.Clock
+import org.joda.time.DateTime
+
+@ImplementedBy(classOf[OrganizationInviteRepoImpl])
+trait OrganizationInviteRepo extends Repo[OrganizationInvite] {
+  def inviteUsers(orgId: Id[Organization], inviterId: Id[Organization],
```
andrewconner, line 13: What would this do? Create the objects and persist? Why not just create them (super simple with case classes), and use `orgInviteRepo.save`?

kifi-backend/modules/common/conf/evolutions/shoebox/335.sql
Colin-Lane, line 17: I wasn't able to find a way to enforce a unique constraint on a column between multiple tables (slug here, and slug from user). But we could do that manually with a before insert trigger to check whether we should actually insert.

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationInviteRepo.scala
```diff
@@ -0,0 +1,72 @@
...
+    inviteUsers: Seq[Either[Id[User], EmailAddress]], message: Option[String]): Seq[OrganizationInvite] = ???
+  def inviteUser(orgId: Id[Organization], inviterId: Id[Organization], userId: Either[Id[User], EmailAddress],
+    message: Option[String]): OrganizationInvite = ???
+  def acceptInvitation(organizationId: Id[Organization], userId: Id[User]): OrganizationMembership = ???
```
andrewconner, line 17: What if there is no membership? In fact, how would `OrganizationMembership` come out here?

Alternatively, you could have something like `orgInviteRepo.getInvite(orgId, userId): Option[OrganizationInvite]`. Marking one as accepted would just be `orgInviteRepo.save(orgInvite.copy(state = OrganizationInviteStates.ACCEPTED))` (and you'd create a `OrganizationMembership` to match).

andrewconner, line 17: For our repos, we prefer to keep them as dumb as possible. Basically, nearly all our hundreds of repo methods are different kinds of getters. Updates are just `save`s. Any business logic, like "checking if there was an invite, accepting it, creating a membership record" should go into Commanders.

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationMembershipRepo.scala
Colin-Lane, line 15: yeah, the most we could get without needing to join is a Seq[Id[Organization]] from the OrganizationMembershipRepo

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationLogoRepo.scala
andrewconner, line 13: Repos shouldn't do anything fancy like this. Create a dumb case class and save it to insert records.

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationInviteRepo.scala
Colin-Lane, line 17: Yeah, accepting and declining belongs is a bit more complex than just a simple update. It will be moved to a commander.

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationRepo.scala
Colin-Lane, line 21: I don't think refactoring LibraryCommander right now makes sense while trying to get an MVP for Teams out the door. Only once this code is in master would it be reasonable to refactor LibraryCommander.

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationRepo.scala
```diff
@@ -0,0 +1,74 @@
...
+import com.keepit.common.time.Clock
+import org.joda.time.DateTime
+
+@ImplementedBy(classOf[OrganizationRepoImpl])
+trait OrganizationRepo extends Repo[Organization] with SeqNumberFunction[Organization] {
+  /*** Accessing Organizations ***/
+  def listLibraries(organizationId: Id[Organization]) = ???
+
+  /*** Modifying Organizations ***/
+  def createOrganization(ownerId: Id[User], teamName: String, description: Option[String], slug: OrganizationSlug) = ???
```
andrewconner, line 18: As before (I won't mention it again), you can get all this specialization for free by just having case classes and saving them with the repo. That concept handles this one, and things like `changeName`, `changeDescription`, even `destroyOrganization`

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationRepo.scala
andrewconner, line 21: Agreed. This deals exclusively with `Library` objects, so shouldn't be here. And if it dealt with `Organization` objects, we still probably would handle this in other ways (ie, we don't need to be returning the whole model).

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationRepo.scala
```diff
@@ -0,0 +1,74 @@
...
+      seq: SequenceNumber[Organization],
+      name: String,
+      description: Option[String],
+      ownerId: Id[User],
+      slug: OrganizationSlug) = {
+      Organization(id, createdAt, updatedAt, state, seq, name, description, ownerId, slug)
+    }
+
+    def unapplyToDbRow(org: Organization) = {
+      Some(
```
andrewconner, line 57: AAAAT: Always Avoid Accidental Auto Tupling :smile:

PR #15614, opened 2015-06-01
Slowing grat email scheduling

camhashemi:
@stkem refactored and made a quick fix to the future handling


kifi-backend/modules/shoebox/app/com/keepit/commanders/emails/GratificationEmailSender.scala
```diff
@@ -12,22 +10,21 @@ import com.keepit.common.logging.Logging
...
-  case class SortedLibraryCountData(totalCount: Int, countByLibrary: List[(Id[Library], Int)]) // represents the total and per-library view counts for a single user
-  case class SenderInfo(firstName: String, lastName: String, title: String, addr: EmailAddress, path: String)
 
-  val SENDER_FIRSTNAME = "Cam"
-  val SENDER_LASTNAME = "Hashemi"
-  val SENDER_ADDR = SystemEmailAddress.CAM
-  val SENDER_PATH = "/cam"
-  val SENDER_TITLE = "Engineer"
-  val senderInfo = SenderInfo(SENDER_FIRSTNAME, SENDER_LASTNAME, SENDER_TITLE, SENDER_ADDR, SENDER_PATH)
+  case class SenderInfo(
```
stkem, line 20: I think what you want here is an `object`.

camhashemi, line 20: I tried that, but I had trouble importing it into gratification.scala.html.
I think because object creates an anonymous class? Is @Singleton class a
workaround here?

camhashemi, line 20: Nevermind, figured it out.

PR #15617, opened 2015-06-01
Meta Tags for feeds.

Colin-Lane:
- also removes Experiment check

@andrewconner @stkem


kifi-backend/modules/shoebox/app/com/keepit/commanders/PublicPageMetaTags.scala
```diff
@@ -85,6 +85,15 @@ case class PublicPageMetaFullTags(unsafeTitle: String, url: String, urlPathOnly:
   val lastName = clean(unsafeLastName)
   val fullName = s"$firstName $lastName"
 
+  private def includeFeedTags: String = includeFeedUrl match {
+    case true =>
+      s"""
+         |<link rel="alternate" type="application/rss+xml" title="RSS Feed for $url" href="$url/rss" />
```
andrewconner: How about a better title? `<name of library> feed on Kifi` or something

kifi-backend/modules/shoebox/app/com/keepit/commanders/PublicPageMetaTags.scala
```diff
@@ -72,10 +72,10 @@ case class PublicPageMetaPrivateTags(urlPathOnly: String) extends PublicPageMeta
     """.stripMargin
 }
 
-case class PublicPageMetaFullTags(unsafeTitle: String, url: String, urlPathOnly: String, unsafeDescription: String,
-    images: Seq[String], facebookId: Option[String], createdAt: DateTime, updatedAt: DateTime,
-    unsafeFirstName: String, unsafeLastName: String, profileUrl: String, noIndex: Boolean,
-    related: Seq[String]) extends PublicPageMetaTags {
+case class PublicPageMetaFullTags(unsafeTitle: String, url: String, urlPathOnly: String,
+    includeFeedUrl: Boolean = false, unsafeDescription: String, images: Seq[String], facebookId: Option[String],
```
andrewconner: Mind removing the default param here? Makes it very easy to skip over this parameter, and we want the API consumer to always think about it.

andrewconner:
Great!

kifi-backend/modules/shoebox/app/com/keepit/commanders/PublicPageMetaTags.scala
Colin-Lane: yeah, I can remove the default param.

kifi-backend/modules/shoebox/app/com/keepit/commanders/PublicPageMetaTags.scala
Colin-Lane: I don't think `PublicPageMetaTags` currently takes in a library, or the name of the library as a parameter. I could modify the `includeFeedUrl` to be an `Option[String] or Option[Library]` to pass in the library's name.

stkem:
:+1:

stkem:
If the experiment is not used elsewhere, you can remove it entirely (i.e. from `UserExperiment.scala`)

Colin-Lane:
Yeah, it is not used anywhere anymore, so I will remove it.

PR #15624, opened 2015-06-02
adding collaborators and followers to related libraries and search result libraries

2is10:
@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/search/search.js
+++ b/kifi-backend/modules/shoebox/angular/src/search/search.js
@@ -186,17 +186,23 @@ angular.module('kifi')
     }
 
     function unpackLibrary(library) {
+      var followers = library.followers || [];
       return {
         id: library.id,
         owner: library.owner,
-        numFollowers: library.memberCount - 1,
-        numKeeps: library.keepCount,
+        collaborators: library.collaborators,
+        numCollaborators: library.numCollaborators,
+        followers: followers,
+        numFollowers: library.numFollowers,
+        numKeeps: library.numKeeps,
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/shoebox/angular/src/search/search.js
```diff
@@ -186,17 +186,23 @@ angular.module('kifi')
     }
 
     function unpackLibrary(library) {
+      var followers = library.followers || [];
       return {
         id: library.id,
         owner: library.owner,
-        numFollowers: library.memberCount - 1,
-        numKeeps: library.keepCount,
```
2is10, line 193: @leogrim `keepCount` and `memberCount` can be removed shortly after this goes out.

kifi-backend/modules/shoebox/angular/src/search/search.js
```diff
@@ -186,17 +186,23 @@ angular.module('kifi')
...
+        numKeeps: library.numKeeps,
         name: library.name,
         description: library.description,
         color: library.color,
         image: library.image,
         path: library.path,
-        reason: 'topic'
+        membership: library.membership,
+        modifiedAt: library.modifiedAt,
+        reason: followers.length || library.modifiedAt ? '' : 'topic'
```
2is10, line 205: @leogrim In a search result library, if we prefer “Updated 3 days ago” over “Related” at the bottom of the card, you can start sending down `modifiedAt` and the UI will start showing it automatically.

leogrim, line 205: Will do!

PR #15626, opened 2015-06-02
hashtag cloud updater

yingjieMiao:
@stkem 
- add a column `kind` in the `library_suggested_search` table. two kinds: `auto` (e.g. machine generated) and `hashtag` (user generated)
- A single scheduler to generate tag tasks: a) for hashtag tasks, compute it on shoebox right away; b) for auto tags, send requests to SQS, waiting for remote service (i.e. Rover) to finish computation. 
- logic on the rover side will be in next PR.
- scheduler is not activated yet.


andrewconner:
Looks good to me, awesome

stkem:
lgtm! We are going to have to keep an eye on the load for shoebox this will generate.

PR #15628, opened 2015-06-02
added an unsubscribe link to grat emails

camhashemi:
@andrewconner


kifi-backend/modules/shoebox/app/views/email/black/gratification.scala.html
```diff
@@ -53,5 +53,7 @@
 <p>Cheers,<br>
 @SenderInfo.FIRSTNAME, @SenderInfo.ROLE at Kifi<br>
 <a href="@htmlUrl(baseUrl + SenderInfo.PATH + "?", "gratSenderProf", true)">www.kifi.com@SenderInfo.PATH</a></p>
+
+<p>P.S. We send these emails to inform you of new library views, followers, and connections. If you don't want to receive these weekly emails, <a href="@unsubscribeUrl">click here</a>.</p>
```
camhashemi, line 57: A good number of these emails are only going to say "1 new connection", and these users are likely to miss the point of the email and unsub. I want to let those users know that there is meant to be more info on the page without rubbing it in their face, maybe there is a better way.

andrewconner:
Looks good to me.

PR #15629, opened 2015-06-02
Teams merge

Colin-Lane:
@leogrim


kifi-backend/modules/common/app/com/keepit/model/OrganizationInvite.scala
```diff
@@ -0,0 +1,59 @@
...
+  def withUpdateTime(now: DateTime): OrganizationInvite = this.copy(updatedAt = now)
+  def withState(newState: State[OrganizationInvite]): OrganizationInvite = this.copy(state = newState)
+
+  override def toString: String = s"OrganizationInvite[id=$id,organizationId=$organizationId,ownerId=$inviterId,userId=$userId,email=$emailAddress,access=$access,state=$state]"
+
+}
+
+object OrganizationInvite extends ModelWithPublicIdCompanion[OrganizationInvite] {
+
+  protected[this] val publicIdPrefix = "o"
```
andrewconner, line 35: This should be different than `"o"`, since that's taken. Do you even need public ids for org invites? @leogrim @Colin-Lane 

Definitely change it before we have consumers, because changing ids later is hard.

leogrim, line 35: Probably not, org / sender / receiver will likely be what we trade with.

Colin-Lane, line 35: We might not even need it. Every use so far has been looking up an invite
by org, user, email, or authtoken

andrewconner, line 35: Yep, cool. Yeah we don't use library ids. It simplifies clients to let the backend to do the heavy lifting, and the authToken is a sort of id. Remove it?

PR #15639, opened 2015-06-03
Fix search user filters

leogrim:


kifi-backend/modules/search/conf/searchService.routes
```diff
@@ -44,7 +44,7 @@ GET         /m/2/search       @com.keepit.search.controllers.mobile.MobileSearch
...
-GET         /site/search                        @com.keepit.search.controllers.website.WebsiteSearchController.search(q: String, u: Option[String] ?= None, l: Option[String] ?= None, maxUris: Int ?= 0, uriContext: Option[String] ?= None, lastUUIDStr: Option[String] ?= None, maxLibraries: Int ?= 0, libraryContext: Option[String] ?= None, maxUsers: Int ?= 0, userContext: Option[String] ?= None, disablePrefixSearch: Boolean ?= false, auth: Option[String], is: Option[ImageSize] ?= None, debug: Option[String] ?= None)
+GET         /site/search                        @com.keepit.search.controllers.website.WebsiteSearchController.search(q: String, f: Option[String] ?= None, l: Option[String] ?= None, maxUris: Int ?= 0, uriContext: Option[String] ?= None, lastUUIDStr: Option[String] ?= None, maxLibraries: Int ?= 0, libraryContext: Option[String] ?= None, maxUsers: Int ?= 0, userContext: Option[String] ?= None, disablePrefixSearch: Boolean ?= false, auth: Option[String], is: Option[ImageSize] ?= None, debug: Option[String] ?= None)
```
2is10, line 47: @leogrim Oh, was that it? I could change `f` to `u` in the web app instead, if you’d prefer. I remember you mentioning this to me once. I must have forgotten.

leogrim, line 47: It's still `f` on mobile, so it's probably better to keep things consistent for now anyway.

PR #15640, opened 2015-06-03
Added a column to display a user's primary email address in the admin view

ryanpbrewster:
@stkem


stkem:
lgtm! You should merge the other pull request from the same branch first though :-) The changes from that are also in here.

PR #15642, opened 2015-06-03
Refactor LibraryCommander

Colin-Lane:
- pulled out a bunch of methods dealing with invites into
  LibraryInviteCommander (new class)
- modified existing tests for libraryCommander to call relevant methods
  from libraryInviteCommander instead.
- updated all previous references of refactored methods to the new
  class (injected the new class where needed too).

@leogrim @andrewconner


leogrim:
:+1:

PR #15644, opened 2015-06-03
LibrarySubscription: added tests, subscription "name" support, and addSubscription()

camhashemi:
@stkem


kifi-backend/modules/shoebox/app/com/keepit/commanders/LibrarySubscriptionCommander.scala
```diff
@@ -0,0 +1,73 @@
...
+import com.keepit.model
+import com.keepit.model._
+import com.kifi.macros.json
+import play.api.libs.json.{ Json, JsValue }
+
+import scala.concurrent.{ Future, ExecutionContext }
+
+@json case class BasicSlackMessage(
+    text: String,
+    destChannel: String = "",
```
stkem: Why does `destChannel` have a default value? What would the behavior be if it is indeed just `""`?

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibrarySubscriptionCommander.scala
```diff
@@ -0,0 +1,73 @@
...
+  def addSubscription(newSub: LibrarySubscription): Either[String, LibrarySubscription] = { // to be called by a controller, perhaps LibraryController
+
+    val (sameNameSubOpt, sameTriggerSubs): (Option[LibrarySubscription], Seq[LibrarySubscription]) = db.readOnlyReplica { implicit session =>
+      val sameNameSubOpt = librarySubscriptionRepo.getByLibraryIdAndName(libraryId = newSub.libraryId, name = newSub.name) // assumming we allow for "inactive" subscriptions that can be reactivated
+      val sameTriggerSubs = librarySubscriptionRepo.getByLibraryIdAndTrigger(libraryId = newSub.libraryId, trigger = newSub.trigger)
+      (sameNameSubOpt, sameTriggerSubs)
+    }
+
+    (sameNameSubOpt, sameTriggerSubs) match {
+      case (Some(sameName), _) => Left("There already exists a subscription to this library with that name.")
```
stkem: In general, try to keep errors like this more machine friendly, so a client can match on it and do appropriate things. E.g. just something like `"name_taken"`.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibrarySubscriptionCommander.scala
```diff
@@ -0,0 +1,73 @@
...
+      case (_, sameTriggers) if sameTriggers.exists(sub => sub.info.hasSameEndpoint(newSub.info)) => Left("There already exists a subscription to this library with that trigger and endpoint.")
```
stkem: Same here.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibrarySubscriptionCommander.scala
camhashemi: When a user creates a Slack integration, they specify a default channel for messages to be received. "" sends to this channel, though another #channel or @user could be specified.

stkem: Ah, I see. Makes sense. In that case making it an `Option[String] = None` I think would be clearer, instead of having `""` as a magic value.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibrarySubscriptionCommander.scala
stkem: Oh, this only applies for error messages intended for a controller (and thus often the client). Other error messages can (and should) be human reading optimized (as you've been doing in other places)

kifi-backend/modules/shoebox/app/com/keepit/model/LibrarySubscription.scala
```diff
@@ -0,0 +1,95 @@
...
+
+}
+
+object SubscriptionInfo {
+  implicit val format: Format[SubscriptionInfo] = new Format[SubscriptionInfo] {
+    def reads(json: JsValue): JsResult[SubscriptionInfo] = {
+      val kind = (json \ "kind").asOpt[String]
+      kind match {
+        case Some("slack") => Json.fromJson[SlackInfo](json)
+        case s: Some[String] => JsError("[LibrarySubscription] tried to read into SubscriptionInfo with invalid kind field")
```
stkem: Minor, but idiomatic scala here would be something like `case Some(unknownKind) => <Error to include the unknown kind>`. The `[String]` is redundant, since `kind` is already know to be `Option[String]`.

stkem:
Very nice! Only made some minor comments. (No need to make a new PR).
You'll also need to make a SQL evolution to actually create the `LibrarySubscription` table. Look for consecutive `<some small integer>.sql` files in the codebase for examples. You are going to want to add a file to that list.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibrarySubscriptionCommander.scala
camhashemi: I see, that makes sense.

kifi-backend/modules/shoebox/app/com/keepit/model/LibrarySubscriptionRepo.scala
```diff
@@ -0,0 +1,67 @@
...
+  )
+
+  type RepoImpl = LibrarySubscriptionTable
+
+  class LibrarySubscriptionTable(tag: Tag) extends RepoTable[LibrarySubscription](db, tag, "library_subscription") {
+    def libraryId = column[Id[Library]]("library_id", O.NotNull)
+    def name = column[String]("name", O.NotNull)
+    def trigger = column[SubscriptionTrigger]("trigger", O.NotNull)
+    def info = column[SubscriptionInfo]("info", O.NotNull) // this will be changed to SubscriptionInfo once I figure out serialization
+
```
yingjieMiao, line 46: remove comments?

camhashemi, line 46: ah, thank you.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibrarySubscriptionCommander.scala
camhashemi: Got it. Confirmed that this works with Slack.

camhashemi:
@stkem updated. wasn't sure how to go about sizing the varchars.

kifi-backend/modules/common/conf/evolutions/shoebox/338.sql
```diff
@@ -0,0 +1,18 @@
...
+create table library_subscription(
+  id bigint(20) NOT NULL AUTO_INCREMENT,
+  created_at datetime NOT NULL,
+  updated_at datetime NOT NULL,
+  state varchar(20) NOT NULL,
+  library_id bigint(20) NOT NULL,
+  name varchar(32) NOT NULL,
+  trigger varchar(32) NOT NULL,
+  info varchar(2048) NOT NULL
+)
```
stkem: You'll need some indices to make this work well. At the least a primary key on `id` and probably an index on `library_id` for efficient lookup. Possibly a unique index on `(libary_id, name)`.

stkem:
Great! Small comment on the SQL, rest looks good to me.

camhashemi:
@stkem @andrewconner added support for adding/editing subscriptions in LibraryCommander, as well as new keep notifications in KeepInterner.

Eishay wants a manual demo asap, so my next step is to manually add an integration before moving to remove subscriptions and front-end support. I need this merge for that purpose.

camhashemi:
Read^: I'm not blocked but I need this to do the manual integration

stkem:
Go for it!

PR #15645, opened 2015-06-04
Invite code cleanup + start of fixing issues

andrewconner:
@stkem Not much substance here, besides passing along authTokens. Should be removed, though, because anyone can edit these cookies (which is our vulnerability). More incoming, just wanted to get it out there.


kifi-backend/modules/common/app/com/keepit/model/LibraryInvite.scala
```diff
@@ -29,7 +29,7 @@ case class LibraryInvite(
     createdAt: DateTime = currentDateTime,
     updatedAt: DateTime = currentDateTime,
     state: State[LibraryInvite] = LibraryInviteStates.ACTIVE,
-    authToken: String = RandomStringUtils.randomAlphanumeric(7),
+    authToken: String = RandomStringUtils.randomAlphanumeric(9),
```
stkem, line 32: :-P

kifi-backend/modules/shoebox/app/com/keepit/commanders/UserCommander.scala
```diff
@@ -314,7 +314,7 @@ class UserCommander @Inject() (
     SafeFuture {
       db.readWrite(attempts = 3) { implicit session =>
         userValueRepo.setValue(newUser.id.get, UserValueName.AUTO_SHOW_GUIDE, true)
-        userValueRepo.setValue(newUser.id.get, UserValueName.AUTO_SHOW_PERSONA, true)
+        userValueRepo.setValue(newUser.id.get, UserValueName.AUTO_SHOW_PERSONA, true) // todo, this shouldn't be true for all users, like when they were invited to a lib
```
stkem, line 320: For what it's worth, forcing every new user onto personas (explicitly including those invited through a library) was a deliberate product decision when we launched this.

stkem:
lgtm, modulo merge conflict

kifi-backend/modules/shoebox/angular/src/libraries/library.js
```diff
@@ -298,6 +298,7 @@ angular.module('kifi')
         showInstallModal();
       }
       if (initParams.getAndClear('intent') === 'follow' && !$scope.library.membership) {
+        // todo wtf why is this here and possible? no good.
         libraryService.joinLibrary($scope.library.id);
```
2is10, line 298: This was added so that clicking a follow button in an email works. Twitter has something similar. Might be a good idea to tie it to a one-time use or expiring token.

kifi-backend/modules/shoebox/app/com/keepit/commanders/UserCommander.scala
andrewconner, line 320: Alright. It's apparently a deliberate decision to not do it now :smile:

kifi-backend/modules/shoebox/angular/src/libraries/library.js
andrewconner, line 298: Ahh, thanks for the context. Yeah, we'll need to do that. This wasn't the big problem I found yesterday, but I did notice it and it seemed a bit out of place (that a link could force a follow).

PR #15647, opened 2015-06-04
returning invite information with each LibraryCardInfo for Invited tab

2is10:
@andrewconner 
This is necessary so that we can tell the difference between follow invites and collaborate invites.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/model/LibraryRepo.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/model/LibraryRepo.scala
@@ -8,6 +8,7 @@ import com.keepit.common.db.slick._
 import com.keepit.common.db._
 import com.keepit.common.healthcheck.AirbrakeNotifier
 import com.keepit.common.logging.Logging
+import com.keepit.common.mail.EmailAddress
 import com.keepit.common.plugin.{ SchedulingProperties, SequencingActor, SequencingPlugin }
 import com.keepit.common.time._
 import com.keepit.common.util.Paginator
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/shoebox/app/com/keepit/model/LibraryRepo.scala
```diff
@@ -291,10 +308,14 @@ class LibraryRepoImpl @Inject() (
     query.as[Library].list
   }
 
-  def getInvitedLibrariesForSelf(userId: Id[User], page: Paginator)(implicit session: RSession): Seq[Library] = {
+  def getInvitedLibrariesForSelf(userId: Id[User], page: Paginator)(implicit session: RSession): Seq[(Library, LibraryInvite)] = {
     import com.keepit.common.db.slick.StaticQueryFixed.interpolation
-    val query = sql"select distinct lib.* from library lib, library_invite li where li.user_id=$userId and li.state='active' and lib.id = li.library_id and lib.state='active' order by lib.member_count desc, lib.last_kept desc, lib.id desc limit ${page.itemsToDrop}, ${page.size}"
-    query.as[Library].list
+    // The technique we use below to ensure we get the desired invite for each library (maximum access level) is from:
```
andrewconner, line 313: Interesting. In other places, we get all invites for a library and just filter in Scala. You think the optimization is worth it?

2is10, line 313: Well, there’s paging going on here too, which would be cumbersome to replicate in Scala. I generally prefer doing filtering and paging in the database. This specific technique should be efficient—one commenter said:

> Incredible! For 20M records it's like 50 times faster than "naive" algorithm (join against a subquery with max())

PR #15649, opened 2015-06-04
folding three LibraryCardInfo types into one

2is10:
@andrewconner


andrewconner:
Aw. At one time, it was elegant! Oh well.

PR #15653, opened 2015-06-04
trying again to return invite info with each LibraryCardInfo for Invited tab

2is10:
@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/model/LibraryRepo.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/model/LibraryRepo.scala
@@ -8,6 +8,7 @@ import com.keepit.common.db.slick._
 import com.keepit.common.db._
 import com.keepit.common.healthcheck.AirbrakeNotifier
 import com.keepit.common.logging.Logging
+import com.keepit.common.mail.EmailAddress
 import com.keepit.common.plugin.{ SchedulingProperties, SequencingActor, SequencingPlugin }
 import com.keepit.common.time._
 import com.keepit.common.util.Paginator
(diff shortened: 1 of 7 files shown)


kifi-backend/modules/shoebox/app/com/keepit/model/LibraryRepo.scala
```diff
@@ -291,10 +297,20 @@ class LibraryRepoImpl @Inject() (
...
     import com.keepit.common.db.slick.StaticQueryFixed.interpolation
-    val query = sql"select distinct lib.* from library lib, library_invite li where li.user_id=$userId and li.state='active' and lib.id = li.library_id and lib.state='active' order by lib.member_count desc, lib.last_kept desc, lib.id desc limit ${page.itemsToDrop}, ${page.size}"
-    query.as[Library].list
+    // The technique we use below to ensure we get the desired invite for each library (maximum access level) is from:
+    // http://stackoverflow.com/questions/12102200/get-records-with-max-value-for-each-group-of-grouped-sql-results#answer-28090544
+    // It works right now because 'read_write' > 'read_only'. We'll need to modify the query (e.g. use a case expression) if we start
+    // to allow sending invitations with additional access levels where lexicographical ordering does not match ordinal ordering.
+    sql"""
+      select lib.*, li.id, li.library_id, li.inviter_id, li.user_id, li.email_address, li.access, li.created_at, li.updated_at, li.state, li.auth_token, li.message
```
2is10, line 307: Listing all of the desired `library_invite` fields in the expected order this time. The production database orders them differently and has an additional field, `pass_phrase`, which is no longer bound to the `LibraryInvite` model.

PR #15655, opened 2015-06-04
Takes the already existing ability to get keep updates from followed …

jaredpetker:
…libraries and pulls it out into it's own request


jaredpetker:
This isn't really the final version that we would use, but something to get us started with, doesn't support paging, etc.

jaredpetker:
@stkem @eishay @andrewconner

stkem:
:+1: :-D

jaredpetker:
I guess ignore this for now, working on support for querying before a time with count (paging)

jaredpetker:
Ok, I think this is ready for merge, as far as I can tell at least - very possible I'm wrong. Doesn't alter any existing endpoints / live code

andrewconner:
Awesome @jaredpetker. Merging

PR #15658, opened 2015-06-04
tag cloud commander

yingjieMiao:
@stkem 

Again, queue job is not activated yet. I have an endpoint that can be tested manually. Would like to try that first. 

The actual algorithm is fairly close to what I wrote in Python. Tried to simply things to a minimum, it may still look cluttered.


yingjieMiao:
will give it a try. feel free to comment afterwards.

PR #15659, opened 2015-06-04
Better library updates

ryanpbrewster:
Made the RecommendationsCommander sample more fairly from a user's followed libraries. It will only allow a certain number of updates from any given library (currently set to 5). @stkem @andrewconner


kifi-backend/modules/shoebox/app/com/keepit/controllers/website/RecommendationsController.scala
```diff
@@ -73,11 +73,26 @@ trait RecommendationControllerHelper extends Logging {
...
 
+  /**
+   * Takes in a list of URI recommendations and a list of Library recommendations
+   * "Mixes" them together (i.e. a random shuffle), with one slight complexity.
+   *
+   * Partitions the libraries based on whether or not they're in the specialBoostSet
+   * Those that are ARE in the specialBoostSet are placed at the front of the list
+   * The rest of the libraries are mixed together with the URIs and placed at the end
+   */
   def mix(uris: Seq[FullUriRecoInfo], libs: Seq[(Id[Library], FullLibRecoInfo)]): Seq[FullRecoInfo] = {
```
yingjieMiao, line 76: we normally don't write excessive comments.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/RecommendationsController.scala
```diff
@@ -86,14 +101,16 @@ trait RecommendationControllerHelper extends Logging {
...
 
     val beta = {
       val cnts = db.readOnlyMaster { implicit s => libMemRepo.countsWithUserIdAndAccesses(userId, Set(LibraryAccess.OWNER, LibraryAccess.READ_ONLY)) }
-      val cntsSum = cnts.values.foldLeft(0)(_ + _)
+      val cntsSum = cnts.values.sum
       val ratio = cntsSum * 1f / 50
       ratio min 1f
     }
     val libBased = ((1 - beta) * 10 + beta * 2).toInt
+    // 10 - 8*min(1, libs/50)
```
yingjieMiao, line 106: commented out codes?

ryanpbrewster, line 106: I added those when I was figuring out what libraryRecoCount did. I'll remove them.

kifi-backend/modules/shoebox/app/com/keepit/model/KeepRepo.scala
```diff
@@ -507,9 +507,10 @@ class KeepRepoImpl @Inject() (
     keeps.list.toSet
   }
 
-  def getRecentKeepsFromFollowedLibraries(userId: Id[User], num: Int)(implicit session: RSession): Seq[Keep] = {
+  def getRecentKeepsFromFollowedLibraries(userId: Id[User])(implicit session: RSession): Seq[Keep] = {
     import com.keepit.common.db.slick.StaticQueryFixed.interpolation
-    sql"""SELECT #$bookmarkColumnOrder FROM bookmark bm WHERE library_id IN (SELECT library_id FROM library_membership WHERE user_id=$userId AND state='active') AND state='active' AND user_id!=$userId ORDER BY kept_at DESC LIMIT $num;""".as[Keep].list
+    val allRecentLibraryUpdates = sql"""SELECT #$bookmarkColumnOrder FROM bookmark bm WHERE library_id IN (SELECT library_id FROM library_membership WHERE user_id=$userId AND state='active') AND state='active' AND user_id!=$userId ORDER BY kept_at DESC;""".as[Keep].list
```
yingjieMiao: maybe helpful to have a constrain like `access != 'owner'` in the inner query.

yingjieMiao: the alias `bm` looks redundant. Also, why remove the limit? This can potentially return over 10K results.

kifi-backend/modules/shoebox/test/com/keepit/model/KeepRepoTest.scala
```diff
@@ -59,7 +59,7 @@ class KeepRepoTest extends Specification with ShoeboxTestInjector {
         db.readWrite { implicit s =>
           val user1 = user().saved
           inject[LibraryMembershipRepo]
-          inject[KeepRepo].getRecentKeepsFromFollowedLibraries(user1.id.get, 10)
+          inject[KeepRepo].getRecentKeepsFromFollowedLibraries(user1.id.get)
```
yingjieMiao: better to have an explicit check here

kifi-backend/modules/shoebox/app/com/keepit/model/KeepRepo.scala
ryanpbrewster: Limit was removed to avoid the case where one library with a lot of recent activity would hog all the recommendations.

yingjieMiao: we could have a large limit, e.g. 500, 1000, 2000, etc, as needed. No limit looks scary for me. we have some libraries having 1000+ keeps (e.g. those twitter libraries). If I follow a few these libraries, the query performance can be bad.

ryanpbrewster: Good idea. Adding that now.

PR #15662, opened 2015-06-04
Rss feed

Colin-Lane:
- Made the rss feeds work with unicode characters (It was mangling Leo's name)
- Added user's name to feed title in Meta Tags

@stkem


stkem:
:+1:

PR #15664, opened 2015-06-04
showing collaborator names in search suggest library hit rows

2is10:
@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/header/searchSuggest.tpl.html
+++ b/kifi-backend/modules/shoebox/angular/src/header/searchSuggest.tpl.html
@@ -5,27 +5,33 @@
     <div class="kf-ssg-query-icon svg-search-icon-gray"></div>
     See all matching keeps <span ng-if="searchInLibrary">in this library</span>
   </a>
-  <a class="kf-ssg-a kf-ssg-uri" href="{{uri.url}}" target="_blank" ng-repeat="uri in uris">
-    <div class="kf-ssg-uri-thumb" ng-style="{'background-image': uri.image ? 'url(' + uri.image.url + ')' : ''}"></div>
-    <div class="kf-ssg-uri-title">{{uri.title}}</div>
-    <div class="kf-ssg-uri-domain">{{uri.url|domain}}</div>
+  <a class="kf-ssg-a kf-ssg-uri" href="{{::uri.url}}" target="_blank" ng-repeat="uri in uris">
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/angular/src/header/searchSuggest.tpl.html
```diff
@@ -5,27 +5,33 @@
...
     <div class="kf-ssg-lib-owner-pic-wrap"><img class="kf-ssg-lib-owner-pic" ng-src="{{library.owner|pic:100}}"></div>
-    <div class="kf-ssg-lib-name">{{library.name}}</div>
-    <div class="kf-ssg-lib-owner-name">{{library.owner|name}}</div>
+    <div class="kf-ssg-lib-name">{{::library.name}}</div>
+    <div class="kf-ssg-lib-owner-name" ng-if="::!library.numCollaborators">{{::library.owner|name}}</div>
+    <div class="kf-ssg-lib-collaborators" ng-if="::library.numCollaborators">
+      <svg class="kf-ssg-lib-collaborators-icon" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 19.3 12.3">
+        <path ng-attr-fill="{{::library.color}}" d="M13.1,7c-0.3,0-0.5,0-0.9,0.1c1,0.7,1.7,1.7,1.7,2.9v2.1h5.2V10C19.1,8,15.1,7,13.1,7z M6.2,7c-2,0-6,1-6,3v2.1h12.1V10C12.2,8,8.2,7,6.2,7z M6.2,5.3c1.5,0,2.6-1.1,2.6-2.6c0-1.5-1.1-2.6-2.6-2.6S3.6,1.2,3.6,2.7C3.6,4.2,4.7,5.3,6.2,5.3z M13.1,5.3c1.5,0,2.6-1.1,2.6-2.6c0-1.5-1.1-2.6-2.6-2.6c-1.5,0-2.6,1.1-2.6,2.6C10.5,4.2,11.6,5.3,13.1,5.3z"/>
+      </svg>
```
2is10, line 26: @stkem Is this how you were planning on doing the library-colored icon in the extension? Seems simplest.

stkem, line 26: Ah, yes. I have something very similar (i.e. using templating to fill in the color in the svg). My code is messier though. I'll steal some of this :-)
(I tried CSS masks with a bakground color first, but couldn't get it to work in FF).

PR #15668, opened 2015-06-05
Library visibility

Colin-Lane:
Adding Organization visibility to Libraries.
@andrewconner @leogrim

### DIFF
--- a/kifi-backend/modules/search/app/com/keepit/search/index/graph/library/LibraryIndexable.scala
+++ b/kifi-backend/modules/search/app/com/keepit/search/index/graph/library/LibraryIndexable.scala
@@ -30,17 +30,20 @@ object LibraryFields {
     val SECRET = 0
     val DISCOVERABLE = 1
     val PUBLISHED = 2
+    val ORGANIZATION = 3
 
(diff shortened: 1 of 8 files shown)


kifi-backend/modules/search/app/com/keepit/search/index/graph/library/LibraryIndexable.scala
```diff
@@ -30,17 +30,20 @@ object LibraryFields {
...
-      if (visibility == SECRET) LibraryVisibility.SECRET
-      else if (visibility == DISCOVERABLE) LibraryVisibility.DISCOVERABLE
-      else LibraryVisibility.PUBLISHED
+    @inline def fromNumericCode(visibility: Long) = visibility match {
+      case SECRET => LibraryVisibility.SECRET
+      case DISCOVERABLE => LibraryVisibility.DISCOVERABLE
+      case ORGANIZATION => LibraryVisibility.ORGANIZATION
+      case _ => LibraryVisibility.PUBLISHED
     }
   }
```
leogrim, line 48: As long as we don't actually create any organization libraries, we're good. But I probably need to review the logic in search to make sure we're not breaking anything.

PR #15671, opened 2015-06-05
Getting richer collaborator info into the extension libs endpoint

stkem:
+fixing logic. If you were the owner of a library hasCollaborators was always false.
@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/model/LibraryMembershipRepo.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/model/LibraryMembershipRepo.scala
@@ -41,6 +41,7 @@ trait LibraryMembershipRepo extends Repo[LibraryMembership] with RepoWithDelete[
   def mostMembersSinceForUser(count: Int, since: DateTime, ownerId: Id[User])(implicit session: RSession): Seq[(Id[Library], Int)]
   def countNonTrivialLibrariesWithUserIdAndAccess(userId: Id[User], access: LibraryAccess, minKeepCount: Int = 1)(implicit session: RSession): Int
   def countsWithUserIdAndAccesses(userId: Id[User], accesses: Set[LibraryAccess])(implicit session: RSession): Map[LibraryAccess, Int]
+  def getUsersWithWriteAccessForLibraries(libIds: Set[Id[Library]], excludeUser: Id[User])(implicit session: RSession): Map[Id[Library], Seq[Id[User]]]
 
   //
   // Profile Library Repo functions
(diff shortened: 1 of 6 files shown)


stkem:
I have the extension side basically ready but want to test with real data before pushing anything there.

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryMembershipRepo.scala
```diff
@@ -387,6 +388,13 @@ class LibraryMembershipRepoImpl @Inject() (
     val q = sql"select lib.id, count(*) cnt from library as lib inner join library_membership as lm on lib.id = lm.library_id where lib.owner_id = ${ownerId} and lm.access != 'owner' and lm.created_at >= ${since} group by lib.id order by cnt desc limit $limit"
     q.as[(Int, Int)].list.map { case (id, cnt) => (Id[Library](id), cnt) }.toMap
   }
+
+  def getUsersWithWriteAccessForLibraries(libIds: Set[Id[Library]], excludeUser: Id[User])(implicit session: RSession): Map[Id[Library], Seq[Id[User]]] = {
+    val access: LibraryAccess = LibraryAccess.READ_WRITE
+    (for { row <- rows if row.libraryId.inSet(libIds) && row.state === LibraryMembershipStates.ACTIVE && row.userId =!= excludeUser && row.access === access } yield (row.userId, row.libraryId)).list.groupBy(_._2).mapValues { seq =>
+      seq.map(_._1).toSeq
+    }
+  }
```
Colin-Lane, line 397: What conditions normally determine if a query should be slick compiled or not, or just plain sql?

yingjieMiao, line 397: we don't have strict rules. Anything before the `.list` is the actual slick query, anything beyond `.list` is just regular scala collection ops. Sometimes we wrote plain sql query, b/c slick can mess things up when query is less straightforward. However, plain sql query losses the benefits of compilation time error catching, thus, a unit test is required. This query seems to be fairly straightforward, although it might be helpful to braking things after `.list`.

stkem, line 397: In general using slick build queries (like this one) is preferrable, because of compile time type checking, but sometimes, as Yingjie said, slick doesn't do a good job or doesn't support a particular sql feature and thus we have to write by hand.

stkem:
@andrewconner mind giving this a quick once over when you get a chance?

PR #15678, opened 2015-06-05
grat email: guard to prevent inactive libraries/connections from sending to users

camhashemi:
@stkem @yingjieMiao 

thank you @ryanpbrewster for catching this


stkem:
:+1:

PR #15684, opened 2015-06-05
RSS validation

Colin-Lane:
- fixes some of the validation errors.
- This should make the RSS feed work better with RSS Readers

@stkem


stkem:
:+1:

PR #15686, opened 2015-06-05
sending collaborators down with user library list cached for nav and keeping

2is10:
@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
@@ -1735,6 +1735,49 @@ class LibraryCommander @Inject() (
     }
   }
 
+  def createLiteLibraryCardInfos(libs: Seq[Library], viewerId: Id[User])(implicit session: RSession): ParSeq[(LibraryCardInfo, MiniLibraryMembership)] = {
+    val memberships = libraryMembershipRepo.getMinisByLibraryIdsAndAccess(
+      libs.map(_.id.get).toSet, Set(LibraryAccess.OWNER, LibraryAccess.READ_WRITE, LibraryAccess.READ_INSERT))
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -1735,6 +1735,49 @@ class LibraryCommander @Inject() (
...
+        collaborators = collabsSample,
+        lastKept = lib.lastKept.getOrElse(lib.createdAt),
+        listed = None, // not needed
+        following = None, // not needed
+        membership = None, // not needed
+        modifiedAt = lib.updatedAt)
+      (info, viewerMem)
+    }
+  }
+
```
2is10, line 1780: Compare this method with `createLibraryCardInfos` above, and I think this one will be “faster enough” to justify its existence. The other is meant for only a small batch of libraries (up to 12?) at a time. This one is meant for loading all libraries a user can keep to at once.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
```diff
@@ -168,32 +168,19 @@ class LibraryController @Inject() (
...
-    val basicUsers = db.readOnlyReplica { implicit session =>
-      basicUserRepo.loadAll(libsWithMemberships.map(_._2.ownerId).toSet)
-    }
-    SafeFuture {
-      db.readOnlyReplica { implicit session =>
-        for ((mem, library) <- libsWithMemberships) yield {
-          val owner = basicUsers(library.ownerId)
-          (LibraryInfo.fromLibraryAndOwner(library, None, owner), mem) // should have library image, but this endpoint doesn't use it and is already slow & heavy.
+    val objs = db.readOnlyMaster { implicit session =>
+      val libs = libraryRepo.getOwnerLibrariesForSelf(request.userId, Paginator.fromStart(200))
```
2is10: Should we use `Int.MaxValue` here? The number 200 is arbitrary.

PR #15688, opened 2015-06-05
Displaying Collaborators in the extension

stkem:
Sill making some small CSS tweaks, but structurally reviewable. 
@andrewconner @2is10

### DIFF
--- a/packrat/scripts/keep_box.js
+++ b/packrat/scripts/keep_box.js
@@ -1098,7 +1098,7 @@ k.keepBox = k.keepBox || (function () {
 
   function setExtraInfo(lib) {
     function smartlyListCollaboratorNames(collabs) {
-      var names = 'Me'; //max 18 characters
+      var names = 'Me'; //max 15 characters
       while (names.length < 15 && collabs.length > 0) {
         var collab = collabs.shift();
         names = names + ', ' + collab.firstName;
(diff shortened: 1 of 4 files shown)


stkem:
Happy to show my local version

packrat/scripts/keep_box.js
```diff
@@ -1096,9 +1096,24 @@ k.keepBox = k.keepBox || (function () {
...
+        var collab = collabs.shift();
+        names = names + ', ' + collab.firstName;
+      }
+      if (collabs.length > 1) {
+        names = names + ', ' + collabs.length + ' others';
+      } else if (collabs.length === 1) {
+        names = names + ', ' + collabs[0].firstName;
+      }
+      return names;
+    }
```
2is10, line 1112: Might want to consider just listing all available collaborator names, comma-separated, and add `text-overflow: ellipsis`. I don’t think the total number is very important in this context. As you know, a character limit is a heuristic that can both overly and “underly” restrict text width.

Also, in other places, the server only sends `BasicUser` info for a sampling of the collaborators, along with the total number, in order to scale nicely to a large set of collaborators. We might want to follow that pattern here too.

2is10:
:+1: nicely done!

stkem:
@andrewconner This is ready to go. Let me know what you would like to do about release.
(I'll likely make some CSS tweaks after the fact, but that is not a blocker)

andrewconner:
Great. Let's leave it open until site is ready.

PR #15690, opened 2015-06-05
CanMoveLibrary

Colin-Lane:
- initial rules for when a user can move a library to/from an
  organization.

@leogrim

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
@@ -696,22 +696,19 @@ class LibraryCommander @Inject() (
     library.state == LibraryStates.ACTIVE && canViewLibrary(userId, library, accessToken)
   }
 
-  private def getLibraryById(id: Id[Library]): Library = db.readOnlyReplica { implicit s => libraryRepo.get(id) }
-
   def canMoveLibrary(userId: Id[User], libId: Id[Library], from: Option[Id[Organization]], to: Option[Id[Organization]]): Boolean = {
     // lib.ownerId = userId && userId is in `from` and `to`
-    val library = getLibraryById(libId)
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -696,6 +696,25 @@ class LibraryCommander @Inject() (
     library.state == LibraryStates.ACTIVE && canViewLibrary(userId, library, accessToken)
   }
 
+  private def getLibraryById(id: Id[Library]): Library = db.readOnlyReplica { implicit s => libraryRepo.get(id) }
+
+  def canMoveLibrary(userId: Id[User], libId: Id[Library], from: Option[Id[Organization]], to: Option[Id[Organization]]): Boolean = {
```
andrewconner: Are we calling this a "move"? May be confused with moving a library from one user to another.

Colin-Lane: I'm not sure what else is more appropriate at the moment. It could be a move or a transfer.
This is specifically for moving a library from user space to org space and back.
We could potentially name it:
canMoveToFromOrg
canTransferToFromOrg

Also just noticed I should move getting the library into the same read-only transaction.

leogrim: Indeed, it's pretty much the same from a product perspective isn't it?

andrewconner: I like `canMoveToFromOrg`, to at least be clear what's happening (and sure, the signature helps)

Colin-Lane: I like that too, I'll rename it to that.

PR #15698, opened 2015-06-06
switching library hovercards to look like our other library cards

2is10:
@andrewconner


leogrim:
@andrewconner @stkem noticed this one?

PR #15703, opened 2015-06-06
Ip tracking

ryanpbrewster:
@stkem @yingjieMiao @eishay 
Added a model, repo, and commander for UserIpAddress which interfaces
with a new database table, "user_ip_addresses". There are hooks for
the website (in KifiSiteRouter), the mobile app (in MobileUserController),
and the extension (in ExtAuthController).

### DIFF
--- a/kifi-backend/modules/common/conf/evolutions/shoebox/339.sql
+++ b/kifi-backend/modules/common/conf/evolutions/shoebox/339.sql
@@ -1,0 +1,19 @@
+# Shoebox
+
+# --- !Ups
+
+CREATE TABLE user_ip_addresses (
+	id bigint(20) NOT NULL AUTO_INCREMENT UNIQUE,
+	created_at datetime NOT NULL,
+	updated_at datetime NOT NULL,
+	state varchar(20) NOT NULL,
+	user_id bigint(20) DEFAULT NULL,
+	ip_address VARCHAR(15) DEFAULT NULL,
+	agent_type VARCHAR(20) DEFAULT NULL,
+
+	PRIMARY KEY(id)
+);
+
+insert into evolutions(name, description) values('339.sql', 'add user ip address table');
+
+# --- !Downs
--- a/kifi-backend/modules/common/conf/evolutions/shoebox/339.sql
+++ b/kifi-backend/modules/common/conf/evolutions/shoebox/339.sql
@@ -7,11 +7,12 @@ CREATE TABLE user_ip_addresses (
 	created_at datetime NOT NULL,
 	updated_at datetime NOT NULL,
 	state varchar(20) NOT NULL,
-	user_id bigint(20) DEFAULT NULL,
-	ip_address VARCHAR(15) DEFAULT NULL,
-	agent_type VARCHAR(20) DEFAULT NULL,
+	user_id bigint(20) NOT NULL,
+	ip_address VARCHAR(15) NOT NULL,
+	agent_type VARCHAR(20) NOT NULL,
 
-	PRIMARY KEY(id)
+	PRIMARY KEY(id),
+	CONSTRAINT ip_address_f_user FOREIGN KEY (user_id) REFERENCES user(id)
 );
 
 insert into evolutions(name, description) values('339.sql', 'add user ip address table');
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/UserIpAddressCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/UserIpAddressCommander.scala
@@ -15,12 +15,18 @@ class UserIpAddressCommander @Inject() (
 
   def simplifyUserAgent(userAgent: UserAgent): String = {
     // TODO: Do we want to know more than this?
-    userAgent.operatingSystemFamily
+    val agentType = userAgent.typeName.toUpperCase()
+    if (!agentType.isEmpty) agentType else "NONE"
   }
-  def logUser(userId: Id[User], ip: IpAddress, userAgent: UserAgent) {
+  def logUser(userId: Id[User], ip: IpAddress, userAgent: UserAgent): Unit = {
     val now = DateTime.now()
     val agentType = simplifyUserAgent(userAgent)
     val model = UserIpAddress(None, now, now, UserIpAddressStates.ACTIVE, userId, ip, agentType)
+    println("[RPB] logUser: " + model)
     db.readWrite { implicit session => userIpAddressRepo.save(model) }
+  }
+
+  def totalNumberOfLogs(): Int = {
+    db.readOnlyReplica { implicit session => userIpAddressRepo.count }
   }
 }
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/UserIpAddressCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/UserIpAddressCommander.scala
@@ -1,6 +1,7 @@
 package com.keepit.commanders
 
 import com.google.inject.Inject
+import com.keepit.common.controller.UserRequest
 import com.keepit.common.db.Id
 import com.keepit.common.db.slick.Database
 import com.keepit.common.logging.Logging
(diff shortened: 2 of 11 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/UserIpAddressCommander.scala
```diff
@@ -0,0 +1,40 @@
...
+
+  def simplifyUserAgent(userAgent: UserAgent): String = {
+    val agentType = userAgent.typeName.toUpperCase()
+    if (agentType.isEmpty) "NONE" else agentType
+  }
+
+  def logUser(userId: Id[User], ip: IpAddress, userAgent: UserAgent): Unit = {
+    val now = DateTime.now()
+    val agentType = simplifyUserAgent(userAgent)
+    val model = UserIpAddress(None, now, now, UserIpAddressStates.ACTIVE, userId, ip, agentType)
```
andrewconner, line 31: Convention is we just skip saying `None` for ids (and instead just use a default value), because it'll always be none when you're constructing the object.

In general, default values can be a bit dangerous, so I'm glad you're not using them indiscriminately

andrewconner, line 31: And actually, if you look at other models, you'll see that the datetime's you have here are handled using default values as well. This is acceptable, since we'll never have the actual time it's persisted.

kifi-backend/modules/shoebox/app/com/keepit/commanders/UserIpAddressCommander.scala
```diff
@@ -0,0 +1,40 @@
...
+    println("[RPB] logUser: " + model)
```
andrewconner: We stay away from `println`, because we can't really control the output (can't pick where it prints to, can't turn it off for tests, etc). Instead, on your class, mix in `with Logging` (from `com.keepit.common.logging`), and you can use `log.info`, `log.warn`, and `log.error`

kifi-backend/modules/shoebox/app/com/keepit/commanders/UserIpAddressCommander.scala
```diff
@@ -0,0 +1,40 @@
...
+    db.readWrite { implicit session => userIpAddressRepo.save(model) }
+  }
+
+  def logUserByRequest[T](request: UserRequest[T]) {
```
andrewconner: An old Scala best practice was to use this syntax for Unit functions. It's actually being removed, because it looks like someone forgot an `=` for the def definition, and isn't explicit. Our preferred way to have a `Unit` function:
`def f(a: Int): Unit = {}`

kifi-backend/modules/shoebox/test/com/keepit/commanders/UserIpAddressCommanderTest.scala
```diff
@@ -0,0 +1,46 @@
...
+          "Mozilla/5.0 (X11; CrOS armv7l 2913.260.0) AppleWebKit/537.11 (KHTML, like Gecko) Chrome/23.0.1271.99 Safari/537.11",
+          "iKeefee/1.0.12823 (Device-Type: iPhone, OS: iOS 7.0.6)"
+        )
+        val answers = Seq(
+          "BROWSER",
+          "BROWSER",
+          "KIFI IPHONE APP"
+        )
+        for ((inp, ans) <- inputs zip answers) {
+          commander.simplifyUserAgent(UserAgent(inp)) === ans
```
andrewconner, line 26: Awesome! Nice tests

kifi-backend/modules/shoebox/test/com/keepit/commanders/UserIpAddressCommanderTest.scala
```diff
@@ -0,0 +1,46 @@
...
+          "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.8; rv:20.0) Gecko/20100101 Firefox/20.0",
+          "Mozilla/5.0 (X11; CrOS armv7l 2913.260.0) AppleWebKit/537.11 (KHTML, like Gecko) Chrome/23.0.1271.99 Safari/537.11",
+          "iKeefee/1.0.12823 (Device-Type: iPhone, OS: iOS 7.0.6)"
+        )
+        val answers = Seq(
+          "BROWSER",
+          "BROWSER",
+          "KIFI IPHONE APP"
+        )
+        for ((inp, ans) <- inputs zip answers) {
```
andrewconner, line 25: And you're picking up Scala syntax very nicely!

andrewconner:
Very nice, minor comments. Merge when you feel it's ready.

kifi-backend/modules/common/conf/evolutions/shoebox/339.sql
```diff
@@ -0,0 +1,19 @@
+# Shoebox
+
+# --- !Ups
+
+CREATE TABLE user_ip_addresses (
+	id bigint(20) NOT NULL AUTO_INCREMENT UNIQUE,
+	created_at datetime NOT NULL,
+	updated_at datetime NOT NULL,
+	state varchar(20) NOT NULL,
+	user_id bigint(20) DEFAULT NULL,
```
eishay: why `user_id` is default null and does not have a foreign key to the user's table?
should we track ip addresses of non users? not sure about that.

kifi-backend/modules/common/conf/evolutions/shoebox/339.sql
```diff
@@ -0,0 +1,19 @@
...
+	ip_address VARCHAR(15) DEFAULT NULL,
```
eishay: `ip_address` and `agent_type`  should not be null as well.

kifi-backend/modules/shoebox/app/com/keepit/commanders/UserIpAddressCommander.scala
```diff
@@ -0,0 +1,40 @@
...
+    val userId = request.userId
+    val userAgent = UserAgent(request.headers.get("user-agent").getOrElse(""))
+    val ip = request.headers.get("X-Forwarded-For").getOrElse(request.remoteAddress)
+    logUser(userId, IpAddress(ip), userAgent)
+  }
+
+  def totalNumberOfLogs(): Int = {
+    db.readOnlyReplica { implicit session => userIpAddressRepo.count }
+  }
+}
```
eishay: new line at end of file

kifi-backend/modules/shoebox/app/com/keepit/commanders/UserIpAddressCommander.scala
```diff
@@ -0,0 +1,40 @@
...
+    val agentType = simplifyUserAgent(userAgent)
+    val model = UserIpAddress(None, now, now, UserIpAddressStates.ACTIVE, userId, ip, agentType)
+    println("[RPB] logUser: " + model)
+    db.readWrite { implicit session => userIpAddressRepo.save(model) }
+  }
+
+  def logUserByRequest[T](request: UserRequest[T]) {
+    val userId = request.userId
+    val userAgent = UserAgent(request.headers.get("user-agent").getOrElse(""))
+    val ip = request.headers.get("X-Forwarded-For").getOrElse(request.remoteAddress)
```
eishay, line 38: I would check here if the ip address starts with `10.` and if so, throw an exception since `10.x.x.x` are internal ips of ec2.

kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtAuthController.scala
```diff
@@ -67,6 +69,7 @@ class ExtAuthController @Inject() (
     log.info(s"start details: $userAgent, $version, $installationIdOpt")
 
     val (libraries, installation, urlPatterns, isInstall, isUpdate) = db.readWrite { implicit s =>
+      userIpAddressCommander.logUserByRequest(request)
```
eishay: This call should not be inside a db session because it overload the session and it actually opens another session within a session, may case a starvation or even deadlock problem.

kifi-backend/modules/shoebox/app/com/keepit/commanders/UserIpAddressCommander.scala
eishay: I think that this function should return a `Future[Unit]` and run its core actions in a non blocking thread. We don't want it to block user online calls, especially if its a table that could be large and inserts may be a bit slow. Also, exceptions from this function should not bubble up to a live request since they are not impairing functionality. We should just airbrake them up.

kifi-backend/modules/shoebox/app/com/keepit/model/UserIpAddress.scala
```diff
@@ -0,0 +1,26 @@
...
+import org.joda.time.DateTime
+
+case class UserIpAddress(
+    id: Option[Id[UserIpAddress]],
+    createdAt: DateTime,
+    updatedAt: DateTime,
+    state: State[UserIpAddress],
+    userId: Id[User],
+    ipAddress: IpAddress,
+    // TODO: Turn agentType into an Enum?
```
eishay, line 15: `Enum` is so java :-)
It should probably be a sealed case class with list of values in an object. See the `State` pattern.

kifi-backend/modules/shoebox/app/com/keepit/commanders/UserIpAddressCommander.scala
stkem: +1

PR #15706, opened 2015-06-06
added remove subscription and polished slack message

camhashemi:
@stkem

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
@@ -612,11 +612,16 @@ class LibraryCommander @Inject() (
           }
 
           def saveSubscriptionChanges(newSubs: Seq[LibrarySubscription]) {
-            newSubs.foreach { newSub =>
-              targetSubs.find { _ equivalent newSub } match { // is newSub brand new or an update?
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -612,13 +612,18 @@ class LibraryCommander @Inject() (
...
-              targetSubs.find { _ equivalent newSub } match { // is newSub brand new or an update?
+            newSubs.foreach { newSub => // find new subs to be updated or added
+              targetSubs.find { _ equivalent newSub } match {
                 case None => subscriptionCommander.saveSubscription(newSub)
-
                 case Some(targetSub) => subscriptionCommander.saveSubscription(targetSub.copy(name = newSub.name, info = newSub.info))
               }
             }
+            targetSubs.foreach { targetSub =>
+              newSubs.find { _ equivalent targetSub } match { // find target subs that are to be removed
```
eishay, line 622: not so sure about this syntax. is it filtering?

camhashemi, line 622: Each modify library request contains the final list of subscriptions. This block looks for equivalent subscriptions (i.e. those with either the same name or URL) in the DB that are not on that list, and inactivates them.

PR #15712, opened 2015-06-07
Cleanup inactive users

leogrim:
@andrewconner @eishay 
We have many inactive users for whom we've never properly cleaned up their data, looks like some of them are still being retrieved via user connections, running a cleanup.


eishay:
LGTM

PR #15714, opened 2015-06-08
Attempt to get rid of zombie inactive users

leogrim:
@eishay @andrewconner @stkem 
Trying to get rid of the few exceptions we've got through the afternoon.


kifi-backend/modules/common/app/com/keepit/model/User.scala
```diff
@@ -132,6 +132,8 @@ object Username {
 
     override def unbind(key: String, username: Username): String = username.value
   }
+
+  case class UndefinedUsernameException(user: User) extends Exception(s"No username for $user")
```
eishay, line 136: nice

PR #15716, opened 2015-06-08
added error logging for SubscriptionCommander and improve slack message

camhashemi:
@stkem please let me know if this is a proper way to handle errors

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibrarySubscriptionCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibrarySubscriptionCommander.scala
@@ -1,23 +1,26 @@
 package com.keepit.commanders
 
 import com.google.inject.{ Singleton, Inject }
-import com.keepit.common.db.Id
 import com.keepit.common.db.slick.Database
 import com.keepit.common.healthcheck.AirbrakeNotifier
 import com.keepit.common.logging.Logging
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/LibrarySubscriptionCommander.scala
```diff
@@ -1,23 +1,26 @@
...
+import play.api.libs.json._
 
 import scala.concurrent.{ Future, ExecutionContext }
 
-@json case class BasicSlackMessage(
-    text: String,
-    destChannel: Option[String] = None,
-    displayedUserName: String = "kifi",
-    iconUrl: String = "https://djty7jcqog9qu.cloudfront.net/assets/black/logo.png") {
+case class BasicSlackMessage(
```
andrewconner, line 14: Why not just `SlackMessage`?

stkem:
Yup, this is fine re error handling. Once you are done debugging you might want to change the `log.error` to an `airbrake.notify`.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibrarySubscriptionCommander.scala
camhashemi, line 14: This format of a message is about as basic as it gets. Should we choose to add a keep image, user thumbnail, etc, which we can, we would call that a "RichSlackMessage".

PR #15717, opened 2015-06-08
fixed bug in grat cron

camhashemi:
@stkem looks like we may have sent the email to the same people mon/friday. also getting reports of 2 - 3 duplicate emails sent at once, not sure the source of that.


camhashemi:
@stkem I initially put the sendNewKeepMessages in the wrong place - think this is the source of the inconsistency. only notified subscribers if the keep was _brand_ new.

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepInterner.scala
```diff
@@ -136,7 +136,6 @@ class KeepInterner @Inject() (
       val bookmark = persistedBookmarksWithUri.bookmark
       if (persistedBookmarksWithUri.isNewKeep) {
         if (library.kind == LibraryKind.USER_CREATED) SafeFuture { libraryCommander.notifyFollowersOfNewKeeps(library, bookmark) }
```
camhashemi, line 138: we should rethink the location of this call. it only notifies followers if the keep is _brand_ new i.e. it hasn't been kept before across Kifi, not just this library.

PR #15720, opened 2015-06-08
[Do Not Merge] Org id keep denormalize merge

Colin-Lane:
@leogrim


kifi-backend/modules/common/app/com/keepit/model/Keep.scala
```diff
@@ -85,45 +86,69 @@ object Keep {
     }
   }
 
+  def applyFromDbRowTuples(firstArguments: KeepFirstArguments, restArguments: KeepRestArguments): Keep = (firstArguments, restArguments) match {
```
andrewconner, line 89: Hah, what a weird world we're in, where this is necessary. I'm sorry.

Colin-Lane, line 89: I checked for alternatives, it was just as hard to write as it is to look at.

kifi-backend/modules/common/conf/evolutions/shoebox/342.sql
```diff
@@ -0,0 +1,12 @@
+# Shoebox
+
+# --- !Ups
+
+ALTER TABLE bookmark
+	ADD COLUMN organization_id bigint(20) DEFAULT NULL;
+ALTER TABLE bookmark
+    ADD CONSTRAINT `bookmark_f_organization_id` FOREIGN KEY (`organization_id`) REFERENCES organization(`id`);
+
```
Colin-Lane, line 9: @andrewconner @leogrim Going to apply this evolution tonight some time unless there are objections

leogrim, line 9: No constraint at least for now, we're not honoring it.

leogrim:
There's now a conflict on evolutions, you can get rid of this one, I have added a `343.sql` and applied it on my own. Closing this PR so that you can fix the conflicts.

PR #15721, opened 2015-06-08
Ip tracking admin view

ryanpbrewster:
@stkem @eishay @andrewconner 

Right now I have it configured to display a user's entire history (could be very long).
I'm not sure what the performance hit will be.


andrewconner:
Nice!

stkem:
lgtm overall! I think it would be prudent to put a limit on the number of IP addresses retrieved, just so you don't blow up memory or something by accident simply be visiting the admin page later on when the IP table is really big. It could be later, maybe 10k? (i.e. in the SQL query, probably)

PR #15725, opened 2015-06-09
Do not cache BasicUser for inactive users, airbrake when returned

leogrim:
@andrewconner @eishay we should not longer be failing calls, but still returning inactive users. At least we'll know about it and can start tracking the underlying issues.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/common/social/BasicUserRepo.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/common/social/BasicUserRepo.scala
@@ -1,19 +1,19 @@
 package com.keepit.common.social
 
-import com.keepit.common.db.{ ExternalId, Id }
-import com.keepit.model.{ UserRepo, User }
+import com.keepit.common.db.{ Id }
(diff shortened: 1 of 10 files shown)


kifi-backend/modules/shoebox/app/com/keepit/common/social/BasicUserRepo.scala
```diff
@@ -23,11 +23,34 @@ class BasicUserRepo @Inject() (
...
+      val activeBasicUsers = loadAllActive(userIds)
+      userIds.map { userId =>
+        userId -> activeBasicUsers.getOrElse(userId, loadInactive(userId))
+      }.toMap
     }
   }
+
+  def loadActive(userId: Id[User])(implicit session: RSession): Option[BasicUser] = {
+    basicUserCache.getOrElseOpt(BasicUserUserIdKey(userId)) {
+      Some(userRepo.get(userId)).filter(_.state == UserStates.ACTIVE).map(BasicUser.fromUser)
```
andrewconner, line 35: Sorry for missing this, but we do do load basicusers for "incomplete_signup" state users (just checking airbrakes), during registration. Not sure why, if we need to, etc.

PR #15727, opened 2015-06-09
Updates in OrganizationRepo

Colin-Lane:
- implemented update methods.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationRepo.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationRepo.scala
@@ -2,15 +2,17 @@ package com.keepit.model
 
 import com.google.inject.{ Provider, ImplementedBy, Inject, Singleton }
 import com.keepit.common.db.{ Id }
-import com.keepit.common.db.slick.DBSession.RSession
+import com.keepit.common.db.slick.DBSession.{ RWSession, RSession }
 import com.keepit.common.db.slick._
 import com.keepit.common.logging.Logging
 import com.keepit.common.time.Clock
 
+import scala.util.{ Success, Failure, Try }
+
 @ImplementedBy(classOf[OrganizationRepoImpl])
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationRepo.scala
```diff
@@ -34,4 +36,29 @@ class OrganizationRepoImpl @Inject() (val db: DataBaseComponent, val clock: Cloc
...
   initTable()
+
+  private[this] val updateNameCompiled = Compiled { (orgId: Column[Id[Organization]]) =>
+    (for { row <- rows if row.id === orgId } yield row.name)
+  }
+
+  class FailedUpdateException extends Exception
+  val failedUpdate = new FailedUpdateException()
+
+  override def updateName(organizationId: Id[Organization], name: String)(implicit session: RWSession): Try[String] = {
```
andrewconner, line 47: You shouldn't need `override`. Why use a direct update? You'll end up with all these tiny functions to edit all the fields, unless there's a specific use case. @leogrim 

By not using save, once you have a cache in here, it won't be invalidated automatically. = potential bugs!

Colin-Lane, line 47: In that case, we really don't need either of the update methods.

leogrim, line 47: I don't think we do, get inspiration from libraries here. Typically a commander method will handle these updates in a convenient way for clients, while using `save` under the hood (and for easy refactoring, it's nice to have convenience methods to update the model itself instead of using `.copy` everywhere).

Colin-Lane, line 47: I will move the methods from being on the repo to the commander and have the commander check permissions logic, then just call save.

PR #15737, opened 2015-06-10
Organization Membership Commander

Colin-Lane:
-get team members by organization

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
@@ -723,10 +723,12 @@ class LibraryCommander @Inject() (
     library.state == LibraryStates.ACTIVE && canViewLibrary(userId, library, accessToken)
   }
 
-  def canMoveToFromOrg(userId: Id[User], libId: Id[Library], from: Option[Id[Organization]], to: Option[Id[Organization]]): Boolean = {
+  def canMoveToFromOrg(userId: Id[User], libId: Id[Library], to: Option[Id[Organization]]): Boolean = {
     // lib.ownerId = userId && userId is in `from` and `to`
     db.readOnlyMaster { implicit s =>
-      (libraryRepo.get(libId).ownerId == userId) &&
+      val library = libraryRepo.get(libId)
+      val from: Option[Id[Organization]] = library.organizationId
+      (library.ownerId == userId) &&
         (from match {
           case Some(fromOrg) => // No Need to check access for MVP, if they are part of an Organization they can move libraries from it.
             organizationMembershipRepo.getByOrgIdAndUserId(fromOrg, userId).nonEmpty
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
@@ -723,7 +723,7 @@ class LibraryCommander @Inject() (
     library.state == LibraryStates.ACTIVE && canViewLibrary(userId, library, accessToken)
   }
 
-  def canMoveToFromOrg(userId: Id[User], libId: Id[Library], to: Option[Id[Organization]]): Boolean = {
+  def canMoveToOrg(userId: Id[User], libId: Id[Library], to: Option[Id[Organization]]): Boolean = {
     // lib.ownerId = userId && userId is in `from` and `to`
     db.readOnlyMaster { implicit s =>
       val library = libraryRepo.get(libId)
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
@@ -1,0 +1,19 @@
+package com.keepit.commanders
+
+import com.google.inject.Inject
+import com.keepit.common.db.Id
+import com.keepit.common.db.slick.Database
+import com.keepit.model._
+
+class OrganizationMembershipCommander @Inject() (db: Database, organizationMembershipRepo: OrganizationMembershipRepo) {
+  // Offset and Count to prevent accidental reversal of arguments with same type.
+  def getByOrgId(orgId: Id[Organization], count: Count, offset: Offset): Seq[OrganizationMembership] = {
+    db.readWrite { implicit session =>
+      organizationMembershipRepo.getbyOrgId(orgId, count, offset)
+    }
+  }
+
+  //  POST    /m/1/teams/:id/members/invite               // takes a list of userId / access tuple's => { members: [ {userId: USER_ID, access: ACCESS}, ...] }
+  //  POST    /m/1/teams/:id/members/modify               // same as above
+  //  POST    /m/1/teams/:id/members/remove               // takes a list of userIds => { members: [ USER_ID1, USER_ID2] }
+}
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
@@ -8,26 +8,34 @@ import com.keepit.common.social.BasicUserRepo
 import com.keepit.model._
 import com.keepit.social.BasicUser
 import org.joda.time.DateTime
+import play.api.http.Status._
 
 class OrganizationMembershipCommander @Inject() (db: Database,
+    organizationRepo: OrganizationRepo,
     organizationMembershipRepo: OrganizationMembershipRepo,
     organizationInviteRepo: OrganizationInviteRepo,
     basicUserRepo: BasicUserRepo) {
 
   final case class MaybeOrganizationMember(member: Either[BasicUser, BasicContact], access: Option[OrganizationAccess], lastInvitedAt: Option[DateTime])
+  final case class OrganizationFail(status: Int, message: String)
 
   // Offset and Count to prevent accidental reversal of arguments with same type (Long).
   def getMembersAndInvitees(orgId: Id[Organization], count: Count, offset: Offset, includeInvitees: Boolean): Seq[MaybeOrganizationMember] = {
     db.readOnlyMaster { implicit session =>
-      val memberships = organizationMembershipRepo.getbyOrgId(orgId, count, offset)
+      val members = organizationMembershipRepo.getbyOrgId(orgId, count, offset)
       val invitees = includeInvitees match {
         case true =>
-          val leftOverCount = Count(Math.max(count.value - memberships.length, 0))
-          val leftOverOffset = Offset(Math.max(offset.value - memberships.length, 0))
+          val leftOverCount = Count(Math.max(count.value - members.length, 0))
+          val leftOverOffset = if (members.isEmpty) {
+            val totalCountForOrg = organizationMembershipRepo.countByOrgId(orgId)
+            Offset(offset.value - totalCountForOrg)
+          } else {
+            Offset(0)
+          }
           organizationInviteRepo.getByOrganization(orgId, leftOverCount, leftOverOffset)
         case false => Seq.empty[OrganizationInvite]
       }
-      buildMaybeMembers(memberships, invitees).take(count.value.toInt)
+      buildMaybeMembers(members, invitees)
     }
   }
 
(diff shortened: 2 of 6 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -723,10 +723,12 @@ class LibraryCommander @Inject() (
     library.state == LibraryStates.ACTIVE && canViewLibrary(userId, library, accessToken)
   }
 
-  def canMoveToFromOrg(userId: Id[User], libId: Id[Library], from: Option[Id[Organization]], to: Option[Id[Organization]]): Boolean = {
+  def canMoveToFromOrg(userId: Id[User], libId: Id[Library], to: Option[Id[Organization]]): Boolean = {
```
yingjieMiao: name of the method is confusing for me. Since the param `from` is removed, I am wondering if the name need some change as well.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
```diff
@@ -0,0 +1,64 @@
...
+      MaybeOrganizationMember(member = Right(contact), access = Some(invitedByAddress.access), lastInvitedAt = Some(invitedByAddress.updatedAt))
+    }
+
+    membersNotIncludingOwner ++ invitedByUserId ++ invitedByEmailAddress
+  }
+
+  //  POST    /m/1/teams/:id/members/invite               // takes a list of userId / access tuple's => { members: [ {userId: USER_ID, access: ACCESS}, ...] }
+  //  POST    /m/1/teams/:id/members/modify               // same as above
+  //  POST    /m/1/teams/:id/members/remove               // takes a list of userIds => { members: [ USER_ID1, USER_ID2] }
+
```
yingjieMiao, line 69: comments?

Colin-Lane, line 69: Yes, to keep track of methods frontend wants. They'll get removed as I fill them in. Still working on them.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
Colin-Lane: Excellent point. I removed the from field because we get `from` for free when we look up the library. The name should probably just be `canMoveToOrg`

kifi-backend/modules/shoebox/test/com/keepit/commanders/OrganizationMembershipCommanderTest.scala
```diff
@@ -0,0 +1,51 @@
...
+            val userId = UserFactory.user().withId(Id[User](i)).saved.id.get
+            orgMemberRepo.save(OrganizationMembership(organizationId = orgId, userId = userId, access = OrganizationAccess.READ_WRITE))
+          }
+        }
+
+        val orgMemberCommander = inject[OrganizationMembershipCommander]
+        // limit by count
+        val membersLimitedByCount = orgMemberCommander.getMembersAndInvitees(orgId, Count(10), Offset(0), true)
+        membersLimitedByCount.length === 10
+
```
yingjieMiao, line 47: since there are 20 rows in the table, these two tests may be fortified. i.e. instead of just checking numbers, you may want to check actual ids.

Colin-Lane, line 47: Agreed, but that makes the tests more difficult as I am not [ordering](http://dba.stackexchange.com/questions/6051/what-is-the-default-order-of-records-for-a-select-statement-in-mysql) results by anything in particular.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
```diff
@@ -0,0 +1,64 @@
...
+  final case class MaybeOrganizationMember(member: Either[BasicUser, BasicContact], access: Option[OrganizationAccess], lastInvitedAt: Option[DateTime])
+
+  // Offset and Count to prevent accidental reversal of arguments with same type (Long).
+  def getMembersAndInvitees(orgId: Id[Organization], count: Count, offset: Offset, includeInvitees: Boolean): Seq[MaybeOrganizationMember] = {
+    db.readOnlyMaster { implicit session =>
+      val memberships = organizationMembershipRepo.getbyOrgId(orgId, count, offset)
+      val invitees = includeInvitees match {
+        case true =>
+          val leftOverCount = Count(Math.max(count.value - memberships.length, 0))
+          val leftOverOffset = Offset(Math.max(offset.value - memberships.length, 0))
```
yingjieMiao: leftover of a count makes sense. leftover of an offset? Not so sure. If this is for paging, we typically have param names such as `pageNum` (starting from 0) and `pageSize`. `Offset` itself is not very clear for me: is it the offset of pageNum, or offset of record number? For example, if pageSize = 5, pageNum = 2, then page offset = 2, record offset = 10.

Colin-Lane: I guess this is something to discuss. I can see it going both ways. Paging like you say for getting specific pages. I would just set `offset = pageNum * pageSize` and `count = pageSize`.
However if it is a more continuous stream, they may display 20 initially and load 5 more at a time as needed, then offset may make more sense.
Once I get more clarity on how this will be used I will make the necessary updates.

@ryanpbrewster

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
yingjieMiao: ah, much clearer!

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
ryanpbrewster: They do some sort of magic infinite-stream thing on mobile. Let me check how it's done for libraries.

kifi-backend/modules/shoebox/test/com/keepit/commanders/OrganizationMembershipCommanderTest.scala
yingjieMiao, line 47: Should we order result by some standard? Timestamp? Alphabet order? 

Put that aside, How about for the second test, make a large enough offset so that num of returned results is less than 10? Otherwise, I get a feeling that offset is not making any difference.

Colin-Lane, line 47: True, I must have accidentally changed this at some point. The offset will not make any difference at the moment. Ordering by Timestamp, or making it so offset actually reduces the number of results will show it is working.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
ryanpbrewster: From looking at the code, I think that "offset" and "count" are both by record. There is no reference to paging anywhere, because mobile does not use static pages.

The code is necessarily complex because we store members and invites in separate places but we want to display them in a seamless stream.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
```diff
@@ -0,0 +1,64 @@
...
+          organizationInviteRepo.getByOrganization(orgId, leftOverCount, leftOverOffset)
+        case false => Seq.empty[OrganizationInvite]
+      }
+      buildMaybeMembers(memberships, invitees).take(count.value.toInt)
```
ryanpbrewster: Should you need the take(count) here? Isn't the sequence correct by construction?

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
yingjieMiao: Ok. But I am still not clear about leftover of an offset. Say, `count = 15`, `offset = 10`. Suppose we retrieved 7 `memberships`. I understand leftover of the count is 15 - 7 = 8. Leftover of the offset is 10 - 7 = 3, what does that mean? Basically, offset and count are two categories of numbers, does it make sense to do subtractions on them?

Colin-Lane: Leftover offset does make sense, but in this case a subtle bug has crept in. We want if offset moves us past all of the members to only show us how many we should see afterwards.
I will need to update the offset logic to know what rows to bring back.

ryanpbrewster: Ah, I see what you're getting at.

@Colin-Lane somehow you need to see how many records you skipped in the repo method which got the members. Maybe have that return (Seq[User], Count) or something.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
leogrim: While the parameter is no longer there because it's possible to retrieve `from` using `libId` , this method will still be used to decide whether you can move a library out of an organization to a user's personal space as well. So `canMoveToOrg` is not ideal either, any suggestion?

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    membersNotIncludingOwner ++ invitedByUserId ++ invitedByEmailAddress
+  }
+
+  def inviteMembers(orgId: Id[Organization], requestorId: Id[User], invitations: Seq[(Id[User], OrganizationAccess)]): Either[OrganizationFail, Seq[OrganizationInvite]] = ???
+  def modifyMemberships(orgId: Id[Organization], requestorId: Id[User], modifications: Seq[(Id[User], OrganizationAccess)]): Either[OrganizationFail, Seq[OrganizationMembership]] = ???
+
+  case class MemberRemovals(failedToRemove: Seq[Id[User]], removed: Seq[Id[User]])
+  def removeMembers(orgId: Id[Organization], requestorId: Id[User], removals: Seq[Id[User]]): MemberRemovals = ???
```
Colin-Lane, line 71: What kind of a return type are we expecting here. Do we want to allow partial success or not?

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
Colin-Lane: The library is changing the location that it lives when you visit it on kifi. We could potentially do `canMoveToLocation`, not sure how I like that one either.
Alternatively we could just keep org off entirely and go with `canMoveLibrary` the move implying we are not changing owners, just changing where the library is located.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
```diff
@@ -0,0 +1,72 @@
...
+      val invitees = includeInvitees match {
+        case true =>
+          val leftOverCount = Count(Math.max(count.value - members.length, 0))
+          val leftOverOffset = if (members.isEmpty) {
+            val totalCountForOrg = organizationMembershipRepo.countByOrgId(orgId)
+            Offset(offset.value - totalCountForOrg)
+          } else {
+            Offset(0)
+          }
+          organizationInviteRepo.getByOrganization(orgId, leftOverCount, leftOverOffset)
```
ryanpbrewster, line 35: What happens if this is called with leftOverCount = 0? Does it still query the database?

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
```diff
@@ -0,0 +1,72 @@
...
+    organizationRepo: OrganizationRepo,
+    organizationMembershipRepo: OrganizationMembershipRepo,
+    organizationInviteRepo: OrganizationInviteRepo,
+    basicUserRepo: BasicUserRepo) {
+
+  final case class MaybeOrganizationMember(member: Either[BasicUser, BasicContact], access: Option[OrganizationAccess], lastInvitedAt: Option[DateTime])
+  final case class OrganizationFail(status: Int, message: String)
+
+  // Offset and Count to prevent accidental reversal of arguments with same type (Long).
+  def getMembersAndInvitees(orgId: Id[Organization], count: Count, offset: Offset, includeInvitees: Boolean): Seq[MaybeOrganizationMember] = {
```
leogrim, line 23: Really like the type safety here, excellent! Naming-wise, could you go for `Offset` and `Limit`? This is to stick with existing conventions, and `Limit` is more accurate than `Count`. Then we should make these two types the default pattern and encourage / enforce their use everywhere in our codebase :)

PR #15746, opened 2015-06-11
Full-stack Subscription support, behind an experiment

camhashemi:
@stkem @andrewconner 

the frontend is rough, but behind an experiment. locally, the controllers are passing subscriptions across the stack, persisting, and displaying them. I can't get it to send the slack message locally, but that still works in production.


kifi-backend/modules/common/app/com/keepit/model/UserExperiment.scala
```diff
@@ -78,6 +78,9 @@ object ExperimentType {
   val RECO_FASTLANE = ExperimentType("reco_fastlane")
   val RECO_SUBSAMPLE = ExperimentType("reco_subsample")
   val APPLY_RECO_FEEDBACK = ExperimentType("apply_reco_feedback")
+  val ROVER_CONTENT = ExperimentType("rover_content")
```
andrewconner: This is actually a bad merge. This and collaborative were removed.

kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.js
```diff
@@ -37,6 +37,9 @@ angular.module('kifi')
         scope.showFollowers = false;
         scope.colors = ['#447ab7','#5ab7e7','#4fc49e','#f99457','#dd5c60','#c16c9e','#9166ac'];
         scope.currentPageOrigin = '';
+        scope.onCollabExperiment = (profileService.me.experiments || []).indexOf('collaborative') > -1;
```
andrewconner: Should not be here

kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.js
```diff
@@ -180,6 +210,9 @@ angular.module('kifi')
           scope.modifyingExistingLibrary = true;
           scope.emptySlug = false;
           scope.modalTitle = scope.library.name;
+          if (!scope.library.subscriptions) {
```
andrewconner: `scope.library.subscriptions = scope.library.subscriptions || [];`

andrewconner: In fact, this can go above, it's needed in both cases.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibrarySubscriptionCommander.scala
```diff
@@ -62,4 +65,62 @@ class LibrarySubscriptionCommander @Inject() (
     db.readWrite { implicit s => librarySubscriptionRepo.save(librarySubscription) }
   }
 
+  def saveSubByLibIdAndKey(libId: Id[Library], subKey: LibrarySubscriptionKey): LibrarySubscription = {
+    saveSubscription(LibrarySubscription(libraryId = libId, name = subKey.name, trigger = SubscriptionTrigger.NEW_KEEP, info = subKey.info))
+  }
+
+  def saveSubsByLibIdAndKey(libId: Id[Library], subKeys: Seq[LibrarySubscriptionKey]): Seq[LibrarySubscription] = {
+    db.readWrite { implicit s =>
+      subKeys.map { key => saveSubByLibIdAndKey(libId, key) }
```
andrewconner: `saveSubscription` already has a db connection. I'd remove it from there, and just use this one.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibrarySubscriptionCommander.scala
```diff
@@ -62,4 +65,62 @@ class LibrarySubscriptionCommander @Inject() (
...
+    }
+  }
+
+  def updateSubsByLibIdAndKey(libId: Id[Library], subKeys: Seq[LibrarySubscriptionKey]): Boolean = {
+
+    def hasSameNameOrEndpoint(key: LibrarySubscriptionKey, sub: LibrarySubscription) = sub.name == key.name || sub.info.hasSameEndpoint(key.info)
+    def saveUpdates(currSubs: Seq[LibrarySubscription], subKeys: Seq[LibrarySubscriptionKey]) {
```
andrewconner: `def saveUpdates(currSubs: Seq[LibrarySubscription], subKeys: Seq[LibrarySubscriptionKey]): Unit =  {`

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibrarySubscriptionCommander.scala
```diff
@@ -62,4 +65,62 @@ class LibrarySubscriptionCommander @Inject() (
...
+        currSubs.find {
+          hasSameNameOrEndpoint(key, _)
+        } match {
+          case None => saveSubByLibIdAndKey(libId, key) // key represents a new sub, save it
+          case Some(equivalentSub) => saveSubscription(equivalentSub.copy(name = key.name, info = key.info)) // key represents an old sub, update it
+        }
+      }
+    }
+    // TODO: refactor these \/ two ^ into one function
+    def removeDifferences(currSubs: Seq[LibrarySubscription], subKeys: Seq[LibrarySubscriptionKey]) {
```
andrewconner: `def removeDifferences(currSubs: Seq[LibrarySubscription], subKeys: Seq[LibrarySubscriptionKey]): Unit = {`

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibrarySubscriptionCommander.scala
```diff
@@ -62,4 +65,62 @@ class LibrarySubscriptionCommander @Inject() (
...
+      currSubs.foreach { currSub =>
+        subKeys.find {
+          hasSameNameOrEndpoint(_, currSub)
+        } match {
+          case None => saveSubscription(currSub.copy(state = LibrarySubscriptionStates.INACTIVE)) // currSub not found in subKeys, inactivate it
```
andrewconner: In fact, I'd remove `saveSubscription` entirely. Since you're often doing batch operations, you should pass it a db session. Then, it's the same as `librarySubscriptionRepo.save`

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibrarySubscriptionCommander.scala
andrewconner: Should take a db session, you're calling it with one.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibrarySubscriptionCommander.scala
andrewconner: As with the other, should take a db session, since you're calling it with one open.

kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.js
camhashemi: scope.library isn't initialized until this conditional, so I think it needs to be in both blocks.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibrarySubscriptionCommander.scala
```diff
@@ -62,4 +65,62 @@ class LibrarySubscriptionCommander @Inject() (
...
+      if (currSubs.isEmpty) {
+        saveSubsByLibIdAndKey(libId, subKeys)
+      } else {
+        saveUpdates(currSubs, subKeys) // save new subs and changes to existing subs
+        removeDifferences(currSubs, subKeys) // remove existing subs that are not in subKeys
+      }
+      getSubsByLibraryId(libId)
+    }
+
+    currSubs != newSubs
```
andrewconner: This function returns `Unit`. Are you intending on returning a boolean?

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibrarySubscriptionCommander.scala
```diff
@@ -62,4 +65,62 @@ class LibrarySubscriptionCommander @Inject() (
...
+
+  }
+
+  def getSubsByLibraryId(id: Id[Library]): Seq[LibrarySubscription] = {
+    db.readOnlyMaster { implicit s => librarySubscriptionRepo.getByLibraryId(id) }
```
andrewconner: Should be able to safely use the replica here, no?

kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.js
andrewconner: Sorry, below it then.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibrarySubscriptionCommander.scala
```diff
@@ -39,12 +40,14 @@ class LibrarySubscriptionCommander @Inject() (
   def sendNewKeepMessage(keep: Keep, library: Library): Seq[Future[ClientResponse]] = {
     val subscriptions: Seq[LibrarySubscription] = db.readOnlyReplica { implicit session => librarySubscriptionRepo.getByLibraryIdAndTrigger(library.id.get, SubscriptionTrigger.NEW_KEEP) }
```
andrewconner: Why so many separate db sessions?

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
```diff
@@ -116,10 +118,12 @@ class LibraryController @Inject() (
       val suggestedSearches = getSuggestedSearchesAsJson(libraryId)
       val membershipOpt = libraryCommander.getViewerMembershipInfo(request.userIdOpt, libraryId)
       val inviteOpt = libraryInviteCommander.getViewerInviteInfo(request.userIdOpt, libraryId)
+      val subscriptions = librarySubscriptionCommander.getSubsByLibraryId(libraryId).map { sub => LibrarySubscriptionKey(sub.name, sub.info) }
 
       val membershipJson = Json.toJson(membershipOpt)
       val inviteJson = Json.toJson(inviteOpt)
-      val libraryJson = Json.toJson(libInfo).as[JsObject] + ("membership" -> membershipJson) + ("invite" -> inviteJson)
+      val subscriptionJson = Json.toJson(subscriptions)
+      val libraryJson = Json.toJson(libInfo).as[JsObject] + ("membership" -> membershipJson) + ("invite" -> inviteJson) + ("subscriptions" -> subscriptionJson)
```
andrewconner: This should be typed. Not a big fan that `membership` and `invite` was just added in. It makes it difficult for clients to understand APIs when there's no solid type associated with the response.

andrewconner:
Minor comments, awesome!

kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.tpl.html
```diff
@@ -59,6 +59,26 @@ <h1 class="dialog-header">
             </label>
           </div>
 
+          <div class="manage-lib-input-wrapper" ng-if="onSubscriptionExperiment">
```
andrewconner, line 62: Why in its own `.manage-lib-input-wrapper`? Seems the one above basically wraps everything? Or did I miss something?

kifi-backend/modules/shoebox/test/com/keepit/model/LibrarySubscriptionTest.scala
```diff
@@ -90,7 +91,7 @@ class LibrarySubscriptionTest extends Specification with ShoeboxTestInjector {
   }
 
   "LibrarySubscriptionCommander" should {
```
andrewconner: Should this be the `LibrarySubscriptionCommanderTest.scala`?

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibrarySubscriptionCommander.scala
camhashemi: Honestly, I just realized what I've been doing haha... refactoring

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibrarySubscriptionCommander.scala
camhashemi: Sorry, the code is crowded from the helper functions, but function is to return a boolean. Moving the logic to the top, so that it's closer to the signature.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
camhashemi: Looks like FullLibraryInfo has reached its 22 case-class parameter limit, and I think that's why `membership` and `invite` are the way they are.

kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.tpl.html
camhashemi, line 62: The one above is `manage-lib-content`, it wraps everything and I need it to be larger to make room for this input-wrapper. Each input has an input-wrapper.

kifi-backend/modules/shoebox/test/com/keepit/model/LibrarySubscriptionTest.scala
camhashemi: yup

camhashemi:
@andrewconner updated

PR #15752, opened 2015-06-11
[DO NOT MERGE] First draft of the org member controller

ryanpbrewster:
MobileOrganizationMembershipController and OrganizationView mostly

@leogrim @Colin-Lane

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/model/Organization.scala
+++ b/kifi-backend/modules/common/app/com/keepit/model/Organization.scala
@@ -20,7 +20,7 @@ case class Organization(
     seq: SequenceNumber[Organization] = SequenceNumber.ZERO,
     name: String,
     description: Option[String] = None,
-    ownerId: Id[User],
+    val ownerId: Id[User],
     handle: Option[PrimaryOrganizationHandle]) extends ModelWithPublicId[Organization] with ModelWithState[Organization] with ModelWithSeqNumber[Organization] {
 
   override def withId(id: Id[Organization]): Organization = this.copy(id = Some(id))
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileOrganizationInviteController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileOrganizationInviteController.scala
@@ -1,0 +1,88 @@
+package com.keepit.controllers.mobile
+
+import com.google.inject.Inject
+import com.keepit.commanders.UserCommander
+import com.keepit.common.controller._
+import com.keepit.common.crypto.{ PublicIdConfiguration, PublicId }
+import com.keepit.common.db.{ ExternalId, Id }
+import com.keepit.common.store.ImageSize
+import com.keepit.heimdal.HeimdalContextBuilderFactory
+import com.keepit.model._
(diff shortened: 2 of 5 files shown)


kifi-backend/modules/common/app/com/keepit/model/Organization.scala
```diff
@@ -20,7 +20,7 @@ case class Organization(
     seq: SequenceNumber[Organization] = SequenceNumber.ZERO,
     name: String,
     description: Option[String] = None,
-    ownerId: Id[User],
+    val ownerId: Id[User],
```
andrewconner, line 23: Was this added intentionally?

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileOrganizationInviteController.scala
```diff
@@ -0,0 +1,88 @@
...
+import com.keepit.heimdal.HeimdalContextBuilderFactory
+import com.keepit.model._
+import com.keepit.shoebox.controllers.LibraryAccessActions
+import play.api.libs.json._
+
+import scala.concurrent.Future
+import scala.util.{ Failure, Success }
+
+class MobileOrganizationInviteController @Inject() ()
+/*
```
andrewconner, line 18: I know you're still going at it, but as a general rule, we prefer not to commit commented code. Either we want it running, or don't. Code you're removing can always be brought back through git.

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationView.scala
```diff
@@ -0,0 +1,18 @@
...
+final case class MaybeOrganizationMember(member: Either[BasicUser, BasicContact], access: Option[OrganizationAccess], lastInvitedAt: Option[DateTime])
+final case class OrganizationFail(status: Int, message: String)
+
+object MaybeOrganizationMember {
+  implicit val writes = Writes[MaybeOrganizationMember] { maybeMember =>
+    val identityFields = maybeMember.member.fold(user => Json.toJson(user), contact => Json.toJson(contact)).as[JsObject]
+    val relatedFields = Json.obj("membership" -> maybeMember.access, "lastInvitedAt" -> maybeMember.lastInvitedAt)
+    json.minify(identityFields ++ relatedFields)
+  }
+}
```
ryanpbrewster, line 18: @Colin-Lane suggested I use "@json" or the "implicit val format" style to get MaybeOrganizationMember to JSON

kifi-backend/modules/common/app/com/keepit/model/Organization.scala
ryanpbrewster, line 23: Yeah, I added this because I am going to need to check if a user is the owner of an Organization.

After the discussion we just had about OrganizationAccess, maybe that needs to be a bit more involved.

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileOrganizationInviteController.scala
ryanpbrewster, line 18: I'll make sure this commit is squashed before I merge anything into the code base.

PR #15755, opened 2015-06-11
invalidating probabilistic experiment generator cache

stkem:
Had to make a manual DB change to create an unconditional probabilistic
experiment. After this is deployed 30% of users (independently, so
there will be overlap) will get the next_gen_recos, reco_subsample, or
apply_reco_feedback experiments.

@yingjieMiao


yingjieMiao:
good luck with the experiment!

PR #15757, opened 2015-06-11
refact keep importer

yingjieMiao:
@andrewconner 

didn't really find the bug yet. but factored quite a bit along the way. This should have no logic change, just wrap things into helper functions (so that we don't have to worry about so many intermediate variables).


andrewconner:
Great, looks good to me

yingjieMiao:
will give it a try and keep an eye on it.

PR #15760, opened 2015-06-12
Fixed a minor bug with the ip cluster cron job

ryanpbrewster:


kifi-backend/modules/shoebox/app/com/keepit/shoebox/cron/UserIpAddressClusterCron.scala
```diff
@@ -42,14 +42,14 @@ class UserIpAddressClusterCronPluginImpl @Inject() (
 
   override def enabled: Boolean = true
   override def onStart() {
-    val every5Minutes = "0 5 0 * * ?"
-    cronTaskOnLeader(quartz, actor.ref, every5Minutes, SearchForClusters(Period.minutes(5), 1))
+    val every5Minutes = "0 5 * * * ?"
+    cronTaskOnLeader(quartz, actor.ref, every5Minutes, SearchForClusters(Period.minutes(5), numClusters = 1))
```
yingjieMiao, line 46: this runs every hour (5 minutes passed the hour). If you want every 5 minutes, you may need something like `0 0/5 * * * ?`

c.f. http://www.quartz-scheduler.org/documentation/quartz-1.x/tutorials/crontrigger

ryanpbrewster, line 46: Ah, I was wrong again. That cron job is just for testing though, so once an hour should be fine. I'll remove that as soon as I'm satisfied it works.

PR #15762, opened 2015-06-12
Creating traits for LibraryCommander and LibraryInviteCommander

andrewconner:
@yingjieMiao @stkem @leogrim @eishay

You wouldn't believe how much faster this makes compilation.

Traits for large classes are great:
1. They serve as a table of contents for a large classes and form a contract for method signatures.
2. Aid in refactoring. Easy to know what my contracts are if I'm locked into a trait.
3. Speeds up compilation, since signatures are locked and the class constructor doesn't change based on dependency injection changes (one of the downsides of Guice).
4. Make it easier to encapsulate public vs private methods.

If everyone agrees, I will be converting our existing commanders to this pattern over time. If anyone wants to help, this makes it a pretty easy process (leaving cleaning up for later):

``` scala
def genTrait(filename: String) = {
  import scala.io.Source
  val name = filename.drop(filename.lastIndexOf('/')+1).dropRight(6)
  val lines = Source.fromFile(filename).getLines().dropWhile(!_.contains(s"class $name"))
  val methods = lines.filter(_.indexOf("  def ") == 0).map(_.reverse.dropWhile(c => c != '=').drop(1).reverse).mkString("\n")
  val t = s"""
@ImplementedBy(classOf[${name}Impl])
trait $name {
  // todo: For each method here, remove if no one's calling it externally, and set as private in the implementation
$methods
}
  """
  println(t)
}
```

### DIFF
--- a/kifi-backend/modules/shoebox/angular/README.md
+++ b/kifi-backend/modules/shoebox/angular/README.md
@@ -4,7 +4,7 @@
 
 You need to [download](http://nodejs.org/download/) and install node.js.
 
-###1.2. Install node dependencies
+###1.2. Install dev dependencies
 
 After successfully installing node.js, type `npm install` in terminal, and it will start downloading all package depenendencies for development under `node_modules/`.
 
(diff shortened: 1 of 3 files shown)


andrewconner:
Starting project for Max? Boring, but...

yingjieMiao:
:+1:

yingjieMiao:
even after refactoring, the table of content is too big for my taste, which can be subjective.

andrewconner:
Yep, it is. And it's easy to see, and perhaps refactor.

PR #15768, opened 2015-06-12
Library Checker

Colin-Lane:
@leogrim @yingjieMiao


kifi-backend/modules/shoebox/app/com/keepit/integrity/LibraryChecker.scala
```diff
@@ -49,91 +57,131 @@ class LibraryChecker @Inject() (
...
+      case failure => failure
+    }
+  }
+
+  private def updateLibrary(id: Id[Library], mutator: Library => Library) = retry(3) {
+    db.readWrite { implicit session =>
+      /* Since we can be processing hundreds of libraries, the library can be out of date by the time we get to actually updating it with our plugin.
+      Do not overwrite new data with old; Instead just refetch the library and update it. */
+      libraryRepo.save(mutator(libraryRepo.get(id)))
+    }
```
yingjieMiao: `retry` seems to be an overkill here. if this batch really failed, next back (in a minute or so) will retry it anyway. And it's very unlikely this block can fail.

Colin-Lane: My goal was not to retry the whole check, but to avoid a single failure from forcing us to have to rerun the entire check. This way we do not fall behind when a single entry fails to update, but rather we try more than once. Thoughts?

kifi-backend/modules/shoebox/app/com/keepit/integrity/LibraryChecker.scala
```diff
@@ -49,91 +57,131 @@ class LibraryChecker @Inject() (
...
-      val allLibraryIds = libraryMap.keySet
-      val latestKeptAtMap = keepRepo.latestKeptAtByLibraryIds(allLibraryIds)
-      val numKeepsByLibraryMap = allLibraryIds.grouped(100).map { keepRepo.getCountsByLibrary(_) }.foldLeft(Map.empty[Id[Library], Int]) { case (m1, m2) => m1 ++ m2 } // grouped to be more friendly with cache bulkget
-      (libraryMap, latestKeptAtMap, numKeepsByLibraryMap)
+  def max(seq: Seq[Long]): Option[Long] = {
+    seq.isEmpty match {
+      case true => None
+      case false => Some(seq.reduceLeft((x, y) => if (x > y) x else y))
     }
+  }
```
yingjieMiao, line 67: you don't really need this customized function. Our seqNum has comparison method. 
c.f. `val highestSeq = keeps.map { _.seq }.max` in the class `SuggestedSearchTermUpdater`.

Colin-Lane, line 67: I did not see that SequenceNumber was Ordered. I will fix this.

kifi-backend/modules/shoebox/app/com/keepit/integrity/LibraryChecker.scala
yingjieMiao: Our current pattern for these plugin batch methods do not bother 'single point failure'. If one point fails, whole batch fails. And I rarely see "single point failures" in these plugins (except `UriIntegrityPlugin` which does too much tricky things). The operations here should be fairly straightforward. We have a few places in our code base have something like

```
db.readWrite(3){   }
```

This will automatically retry for you. In general, we only specify retry number only in desperate cases.

kifi-backend/modules/shoebox/app/com/keepit/integrity/LibraryChecker.scala
```diff
@@ -49,91 +57,131 @@ class LibraryChecker @Inject() (
...
+      (nextSeqNum, librariesNeedingUpdate, newLibraryKeepCount, latestKeptAtMap)
+    }
+    // sync library lastKept with most recent keep.lastKept
+    librariesNeedingUpdate.foreach {
+      case (libraryId, library) =>
+        latestKeptAtMap(libraryId) match {
+          case Some(keepKeptAt) => library.lastKept match {
+            case None =>
+              updateLibrary(libraryId, _.copy(lastKept = Some(keepKeptAt))) match {
+                case Success(success) => // success
```
yingjieMiao: I suggest not use a `try` block. This makes the following code cluttered with pattern matching on `Success` and `Failure`. We usually use a `try` block for some well-known-and-hard-to-eliminate problems. As I mentioned earlier, the operation here is almost trivial. Unless we really have experienced problems in production,  `Try` seems an overkill for me.

kifi-backend/modules/shoebox/app/com/keepit/integrity/LibraryChecker.scala
Colin-Lane: Excellent point. I will just remove the retry logic.

kifi-backend/modules/shoebox/app/com/keepit/integrity/LibraryChecker.scala
```diff
@@ -49,91 +57,104 @@ class LibraryChecker @Inject() (
...
-            log.warn(s"Library ${libId} has inconsistent keep count. Library's keep count is ${lib.keepCount} but there are ${numKeeps} active keeps... update library's keep_count")
-            db.readWrite { implicit s =>
-              libraryRepo.save(lib.copy(keepCount = numKeeps))
+  def syncLibraryLastKeptAndKeepCount() {
+    val timer = new NamedStatsdTimer("LibraryChecker.syncLibraryLastKeptAndKeepCount")
+    val (nextSeqNum, librariesNeedingUpdate, newLibraryKeepCount, latestKeptAtMap) = db.readOnlyMaster { implicit session =>
+      val lastSeq = getLastSeqNum(LAST_KEPT_AND_KEEP_COUNT_NAME)
+
+      val keeps = keepRepo.getBySequenceNumber(lastSeq, KEEP_FETCH_SIZE)
+      val nextSeqNum = keeps.length match {
```
yingjieMiao: could just be `keeps.lastOption.map{_.seq}`. Those keeps are already ordered by seq in increasing order. If you want to be extra sure, could just sort it first. Performance loss is minimal.

kifi-backend/modules/shoebox/app/com/keepit/integrity/LibraryChecker.scala
```diff
@@ -49,91 +57,104 @@ class LibraryChecker @Inject() (
...
+          case None =>
+            log.warn(s"a library $libraryId that has keeps since last sequence number has no keeps??")
         }
+    }
 
+    // snapshot of library state from above.
+    librariesNeedingUpdate.foreach {
+      // sync library keepCount with sum keeps.active
+      case (libraryId, library) =>
+        newLibraryKeepCount.get(libraryId) match {
```
yingjieMiao: equivalent to testing `newLibraryKeepCount.getOrElse(libraryId, 0) == library.keepCount` ?

kifi-backend/modules/shoebox/app/com/keepit/integrity/LibraryChecker.scala
```diff
@@ -49,91 +57,104 @@ class LibraryChecker @Inject() (
...
+
+      val newLibraryKeepCount = keepRepo.getCountsByLibrary(libraryIds)
+      val latestKeptAtMap = keepRepo.latestKeptAtByLibraryIds(libraryIds)
+      val librariesNeedingUpdate = libraryRepo.getLibraries(libraryIds)
+      (nextSeqNum, librariesNeedingUpdate, newLibraryKeepCount, latestKeptAtMap)
+    }
+    // sync library lastKept with most recent keep.lastKept
+    librariesNeedingUpdate.foreach {
+      case (libraryId, library) =>
+        latestKeptAtMap(libraryId) match {
```
yingjieMiao: old code has a pattern matching on the pair `(timeFromLibrary, timeFromKeepRepo)`. It seems clearer for me.

kifi-backend/modules/shoebox/app/com/keepit/integrity/LibraryChecker.scala
```diff
@@ -49,91 +57,104 @@ class LibraryChecker @Inject() (
...
+    // snapshot of library state from above.
+    librariesNeedingUpdate.foreach {
+      // sync library keepCount with sum keeps.active
+      case (libraryId, library) =>
+        newLibraryKeepCount.get(libraryId) match {
+          case Some(count) if library.keepCount != count => updateLibrary(libraryId, _.copy(keepCount = count))
+          case Some(_) => // all good here.
+          case None => log.warn(s"a library $libraryId that has keeps since last sequence number has no keeps??")
+        }
+    }
```
yingjieMiao: How about combine the two checks into one? i.e. we only need one `librariesNeedingUpdate.foreach{}` block. We will mutate a library if any of the two fields are off.

kifi-backend/modules/shoebox/app/com/keepit/integrity/LibraryChecker.scala
```diff
@@ -49,91 +57,104 @@ class LibraryChecker @Inject() (
...
+    nextSeqNum match {
+      case Some(seqNum) => db.readWrite { implicit session =>
```
yingjieMiao: we usually do something like `valOpt.foreach{ val => ... }`, instead of pattern matching, if the `None` branch does nothing.

Colin-Lane: I'm not used to seeing foreach on an Option. I can make this modification.

kifi-backend/modules/shoebox/app/com/keepit/integrity/LibraryChecker.scala
```diff
@@ -49,91 +57,104 @@ class LibraryChecker @Inject() (
...
-      val libraryMap = libraries.map(lib => lib.id.get -> lib).toMap
-      val allLibraryIds = libraryMap.keySet
-      val numMembersMap = libraryMembershipRepo.countByLibraryId(allLibraryIds)
-      (libraryMap, numMembersMap)
+  def syncLibraryMemberCounts() {
+    val timer = new NamedStatsdTimer("LibraryChecker.syncLibraryMemberCounts")
+    val (nextSeqNum, libraries, libraryMemberCounts) = db.readOnlyMaster { implicit session =>
+      val lastSeq = getLastSeqNum(MEMBER_COUNT_NAME)
+      val members = libraryMembershipRepo.getBySequenceNumber(lastSeq, MEMBER_FETCH_SIZE)
+      val nextSeqNum = members.length match {
```
yingjieMiao: could be more succinct.

kifi-backend/modules/shoebox/app/com/keepit/integrity/LibraryChecker.scala
```diff
@@ -49,91 +57,104 @@ class LibraryChecker @Inject() (
...
-        numMembersMap.get(libId).map { numMembers =>
-          if (lib.memberCount != numMembers) {
-            log.warn(s"Library ${libId} has inconsistent member count. Library's member count is ${lib.memberCount} but there are ${numMembers} active memberships... update library's member_count")
-            db.readWrite { implicit s =>
-              libraryRepo.save(lib.copy(memberCount = numMembers))
-            }
-          }
+    libraries.foreach {
+      case (libraryId, library) =>
+        libraryMemberCounts.get(libraryId) match {
```
yingjieMiao, line 131: similar as above. I think the old code can be reused

kifi-backend/modules/shoebox/app/com/keepit/integrity/LibraryChecker.scala
```diff
@@ -49,91 +57,104 @@ class LibraryChecker @Inject() (
...
+          case None => log.warn(s"a library $libraryId that has members since last sequence number has no members??")
+          case Some(count) if count != library.memberCount => updateLibrary(libraryId, _.copy(memberCount = count))
+          case Some(_) => // here be dragons counting your libraries correctly
         }
     }
+    nextSeqNum match {
```
yingjieMiao: consider `.foreach{}`

kifi-backend/modules/shoebox/test/com/keepit/integrity/LibraryCheckerTest.scala
```diff
@@ -0,0 +1,93 @@
...
+    FakeHeimdalServiceClientModule()
+  )
+
+  "library checker" should {
+    "check keep count" in {
+      withDb(modules: _*) { implicit injector =>
+        val libraryChecker = inject[LibraryChecker]
+        val libRepo = inject[LibraryRepo]
+        val library = db.readWrite { implicit session =>
+          val library = LibraryFactory.library().withId(Id[Library](1)).saved
```
yingjieMiao, line 42: could call `.withKeepCount` here

kifi-backend/modules/shoebox/test/com/keepit/integrity/LibraryCheckerTest.scala
```diff
@@ -0,0 +1,93 @@
...
+    "check last kept time" in {
+      withDb(modules: _*) { implicit injector =>
+        val tenYearsAgo = Some(DateTime.now().minusYears(10))
+
+        val libraryChecker = inject[LibraryChecker]
+        val libRepo = inject[LibraryRepo]
+        val (library, keeps) = db.readWrite { implicit session =>
+          val library = LibraryFactory.library().withId(Id[Library](1)).saved
+          val keeps = KeepFactory.keeps(20).map(_.withLibrary(library)).saved
+          // Uh oh, looks like the count is wrong!
```
yingjieMiao: comment is wrong. also, could just call `copy(lastKept = ...)` in line 60.

kifi-backend/modules/shoebox/test/com/keepit/integrity/LibraryCheckerTest.scala
```diff
@@ -0,0 +1,93 @@
...
+          (libRepo.save(library.copy(lastKept = tenYearsAgo)), keeps)
+        }
+        libraryChecker.syncLibraryLastKeptAndKeepCount()
+        val updatedLibrary = db.readOnlyMaster { implicit session => libRepo.get(library.id.get) }
+        library.lastKept === tenYearsAgo
+        updatedLibrary.lastKept must beSome(keeps(19).createdAt)
```
yingjieMiao: `keeps.last` is clearer, and less hard coding. Also, don't you want to check it matches the `keptAt` field?

kifi-backend/modules/shoebox/test/com/keepit/integrity/LibraryCheckerTest.scala
Colin-Lane: KeepFactoryHelpers.saved updates the lastKept time of the library, so I need to be explicit about where I set it. (after keeps are added)

kifi-backend/modules/shoebox/test/com/keepit/integrity/LibraryCheckerTest.scala
```diff
@@ -0,0 +1,93 @@
...
+          libRepo.save(library.copy(memberCount = wrongCount))
+        }
+        inject[LibraryMembershipSequenceNumberAssigner].assignSequenceNumbers()
+        db.readOnlyMaster { implicit session => libRepo.get(library.id.get) }.memberCount === wrongCount
+        libraryChecker.syncLibraryMemberCounts()
+        val updatedLibrary = db.readOnlyMaster { implicit session => libRepo.get(library.id.get) }
+        updatedLibrary.memberCount === 20
+      }
+    }
+  }
```
yingjieMiao, line 101: In these tests, would be nice if you can check the values in `System_Value` table

Colin-Lane, line 101: yeah, that is a good idea. To show that the table is being updated properly.

yingjieMiao:
looks good overall. Most comments are style/convention suggestions.

kifi-backend/modules/shoebox/app/com/keepit/integrity/DataIntegrity.scala
```diff
@@ -21,6 +21,7 @@ class DataIntegrityPluginImpl @Inject() (
   override def enabled: Boolean = true
   override def onStart() {
     scheduleTaskOnOneMachine(actor.system, 7 minutes, EVERY_N_MINUTE minutes, actor.ref, Cron, getClass.getSimpleName)
+    scheduleTaskOnOneMachine(actor.system, 1 minutes, 1 minutes, actor.ref, LibrariesCheck, getClass.getSimpleName)
   }
```
yingjieMiao, line 26: how about send a separate message for `system library check`, with a different time period?  That one's task retrieval logic is different. Note `EVERY_N_MINUTE` is currently used in `libraryChecker` (for hashing purposes). You want it to be propagated properly.

Colin-Lane, line 26: Ahhh, That one wasn't supposed to have its time updated with the others. I will create another task for it and set that back to EVERY_N_MINUTE

yingjieMiao:
`if (members.isEmpty) None else Some(members.map(_.seq).max)` is fine with me.

kifi-backend/modules/shoebox/app/com/keepit/integrity/LibraryChecker.scala
```diff
@@ -49,91 +56,92 @@ class LibraryChecker @Inject() (
...
+      val librariesNeedingUpdate = libraryRepo.getLibraries(libraryIds)
+      (nextSeqNum, librariesNeedingUpdate, newLibraryKeepCount, latestKeptAtMap)
+    }
+    // sync library lastKept with most recent keep.lastKept
+    def updateLibraryLastKeptAt(libraryId: Id[Library], library: Library) {
+      latestKeptAtMap(libraryId).foreach { keepKeptAt =>
+        library.lastKept match {
+          case None =>
+            updateLibrary(libraryId, _.copy(lastKept = Some(keepKeptAt)))
+          case Some(libLastKept) if keepKeptAt.getMillis - libLastKept.getMillis > 30000 => {
```
yingjieMiao: better use `math.abs()` on the diff.

kifi-backend/modules/shoebox/test/com/keepit/integrity/LibraryCheckerTest.scala
```diff
@@ -0,0 +1,103 @@
...
+        val (library, keeps) = db.readWrite { implicit session =>
+          val library = LibraryFactory.library().withId(Id[Library](1)).saved
+          val keeps = KeepFactory.keeps(20).map(_.withLibrary(library)).saved
+          // Uh oh, looks like the lastKept time is wrong!
+          (libRepo.save(library.copy(lastKept = tenYearsAgo)), keeps)
+        }
+        libraryChecker.syncLibraryLastKeptAndKeepCount()
+        val updatedLibrary = db.readOnlyMaster { implicit session => libRepo.get(library.id.get) }
+        library.lastKept === tenYearsAgo
+        updatedLibrary.lastKept must beSome(keeps.last.createdAt)
```
yingjieMiao: should check `last.keptAt`?

Colin-Lane: Yes I should.

kifi-backend/modules/shoebox/test/com/keepit/integrity/LibraryCheckerTest.scala
```diff
@@ -0,0 +1,103 @@
...
+        db.readOnlyMaster { implicit session => libRepo.get(library.id.get) }.memberCount === wrongCount
+        libraryChecker.syncLibraryMemberCounts()
+        val updatedLibrary = db.readOnlyMaster { implicit session => libRepo.get(library.id.get) }
+        updatedLibrary.memberCount === 20
+
+        val systemValueRepo = inject[SystemValueRepo]
+        val (seqNum, memberSeqNum) = db.readOnlyMaster { implicit session =>
+          val seqNum = systemValueRepo.getSequenceNumber(libraryChecker.MEMBER_COUNT_NAME).get
+          val previousSeqNum = SequenceNumber[LibraryMembership](seqNum.value - 1)
+          val memberSeqNum = libraryMembershipRepo.getBySequenceNumber(previousSeqNum, 1).last.seq
```
yingjieMiao: could just use `libraryMembershipRepo.all.map{_.seq}.max`

Colin-Lane:
I can agree with that. There is no guarantee that sequence numbers will come in order. If I modify a previous keep and its sequence number is modified. Are the sequence numbers now out of order?

kifi-backend/modules/shoebox/app/com/keepit/integrity/LibraryChecker.scala
```diff
@@ -49,91 +56,92 @@ class LibraryChecker @Inject() (
...
+      newLibraryKeepCount.get(libraryId) match {
+        case Some(count) if library.keepCount != count => updateLibrary(libraryId, _.copy(keepCount = count))
+        case _ => // all good here.
+      }
+    }
 
+    librariesNeedingUpdate.foreach {
+      case (libraryId, library) =>
+        updateLibraryLastKeptAt(libraryId, library)
+        updateLibraryKeepCount(libraryId, library)
```
yingjieMiao, line 107: I still have some concern about these two mutations. For the second call, we may first read a dirty value from memcached (before the first mutation's value propagated to cache). I think it will be safer and more straightforward if we check the two fields at the same time and mutate only once. Your previous pattern has two separate `.foreach{}` blocks. It reduces the chance of dirty read, but not eliminates the possibility. 

Thoughts? @leogrim

Colin-Lane, line 107: I'm not quite sure on when the cache is updated, but before the write we do a read to get a fresh entry. If there is a delay in writing to cache we may get an old value.
Since we are running this all in the same thread as long as the cache is updated synchronously there should be no problem here.

yingjieMiao, line 107: Ok, the cache call is synchronous. should be good.

yingjieMiao:
looks good to go!

kifi-backend/modules/shoebox/app/com/keepit/integrity/LibraryChecker.scala
Colin-Lane, line 107: If this does become a problem or a performance bottleneck, updating the writes to combine them into a single one wouldn't be all that difficult.

PR #15770, opened 2015-06-12
Org Invite Commander

Colin-Lane:
- stubbed methods for invite/accept/decline/universal link

@leogrim @ryanpbrewster

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
@@ -18,5 +18,5 @@ class OrganizationInviteCommander @Inject() () {
   def declineInvitation(orgId: Id[Organization], userId: Id[User]) = ???
 
   // Creates a Universal Invite Link for an organization and inviter. Anyone with the link can join the Organization
-  def universalInviteLink(orgId: Id[Organization], inviterId: Id[User], access: OrganizationAccess = OrganizationAccess.READ_WRITE) = ???
+  def universalInviteLink(orgId: Id[Organization], inviterId: Id[User], access: OrganizationAccess = OrganizationAccess.READ_WRITE, authToken: Option[String] = None) = ???
 }
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
@@ -11,7 +11,7 @@ import scala.concurrent.Future
 
 class OrganizationInviteCommander @Inject() () {
 
-  def inviteUsersToOrganization(orgId: Id[Organization], inviterId: Id[User], invitees: Seq[(Either[Id[User], EmailAddress], LibraryAccess, Option[String])]): Future[Either[LibraryFail, Seq[(Either[BasicUser, RichContact], LibraryAccess)]]] = ???
+  def inviteUsersToOrganization(orgId: Id[Organization], inviterId: Id[User], invitees: Seq[(Either[Id[User], EmailAddress], LibraryAccess, Option[String])]): Future[Either[LibraryFail, Seq[(Either[BasicUser, RichContact], OrganizationAccess)]]] = ???
 
   def acceptInvitation(orgId: Id[Organization], userId: Id[User], authToken: Option[String] = None): Either[OrganizationFail, (Organization, OrganizationMembership)] = ???
 
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
@@ -13,7 +13,7 @@ class OrganizationInviteCommander @Inject() () {
 
   def inviteUsersToOrganization(orgId: Id[Organization], inviterId: Id[User], invitees: Seq[(Either[Id[User], EmailAddress], LibraryAccess, Option[String])]): Future[Either[LibraryFail, Seq[(Either[BasicUser, RichContact], LibraryAccess)]]] = ???
 
-  def acceptInvitation(orgId: Id[Organization], userId: Id[User], authToken: Option[String] = None, subscribed: Option[Boolean] = None): Either[OrganizationFail, (Organization, OrganizationMembership)] = ???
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
```diff
@@ -0,0 +1,22 @@
...
+import com.keepit.model._
+import com.keepit.social.BasicUser
+
+import scala.concurrent.Future
+
+class OrganizationInviteCommander @Inject() () {
+
+  def inviteUsersToOrganization(orgId: Id[Organization], inviterId: Id[User], invitees: Seq[(Either[Id[User], EmailAddress], LibraryAccess, Option[String])]): Future[Either[LibraryFail, Seq[(Either[BasicUser, RichContact], LibraryAccess)]]] = ???
+
+  def acceptInvitation(orgId: Id[Organization], userId: Id[User], authToken: Option[String] = None, subscribed: Option[Boolean] = None): Either[OrganizationFail, (Organization, OrganizationMembership)] = ???
```
andrewconner: What does `subscribed` mean in this context?

Colin-Lane: This may not actually fit here at the moment. Subscribed made sense for libraries, but not organizations. I can remove this parameter.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
```diff
@@ -0,0 +1,22 @@
...
+class OrganizationInviteCommander @Inject() () {
+
+  def inviteUsersToOrganization(orgId: Id[Organization], inviterId: Id[User], invitees: Seq[(Either[Id[User], EmailAddress], LibraryAccess, Option[String])]): Future[Either[LibraryFail, Seq[(Either[BasicUser, RichContact], LibraryAccess)]]] = ???
+
+  def acceptInvitation(orgId: Id[Organization], userId: Id[User], authToken: Option[String] = None): Either[OrganizationFail, (Organization, OrganizationMembership)] = ???
+
+  def declineInvitation(orgId: Id[Organization], userId: Id[User]) = ???
+
+  // Creates a Universal Invite Link for an organization and inviter. Anyone with the link can join the Organization
+  def universalInviteLink(orgId: Id[Organization], inviterId: Id[User], access: OrganizationAccess = OrganizationAccess.READ_WRITE) = ???
```
ryanpbrewster: We should decide ASAP whether we want to restrict organization invites to Kifi users. That would definitely make our lives easier. I feel like an Organization has higher security requirements than Library follows.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
```diff
@@ -0,0 +1,22 @@
...
+import com.keepit.common.db.Id
+import com.keepit.common.mail.EmailAddress
+import com.keepit.model._
+import com.keepit.social.BasicUser
+
+import scala.concurrent.Future
+
+class OrganizationInviteCommander @Inject() () {
+
+  def inviteUsersToOrganization(orgId: Id[Organization], inviterId: Id[User], invitees: Seq[(Either[Id[User], EmailAddress], LibraryAccess, Option[String])]): Future[Either[LibraryFail, Seq[(Either[BasicUser, RichContact], LibraryAccess)]]] = ???
```
ryanpbrewster: s/Library/Organization/

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
Colin-Lane: I think initially we are restricting to only already existing users to keep the logic simple.
@leogrim

andrewconner: Why not to email addresses? It seems pretty common to add my team to new platforms, before they have registered.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
Colin-Lane: Nothing to see here. No mistakes were made. :100: out of :100:

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
ryanpbrewster: My concern is still security. I guess if we do emails we can insist that the person signs up with that email address.

Colin-Lane: This seems to be in "next" rather than "now" in the team up document.

andrewconner: Interesting, @leogrim. Why not just bake it into the data model? We did it for libraries, it's not much harder. Re: security: they get a one-time-use token that expires after x days. Many services do that. Do you think it's an issue?

ryanpbrewster: Is the authentication code already written somewhere? If so, that seems good to me.

andrewconner: `libraryCommander.canViewLibrary` has the basic idea

leogrim: I don't have any security concern and it's in the data model. It's also in "now" in the specs, see (5). There is a caveat there to specify the on-boarding flow, which I'm not super familiar with. If we can reuse what we have to onboard people invited to libraries, then it should be straightforward isn't it?

Colin-Lane: Interestingly it is also at (16) User can invite Team’s member by email.
We can leave it in for now and see if we can re-use the library code to make it work for organizations

Colin-Lane: I'm pretty sure the universal invite link will need an authToken, so I'm going to add that, then merge the code.

leogrim: Yes you're right, I added it Yesterday to the Team document, it should have the latest models.

PR #15772, opened 2015-06-12
Make implicit recency ranking explicit

leogrim:
@andrewconner this doesn't directly fix our bug but gets me closer.


kifi-backend/modules/search/app/com/keepit/search/controllers/website/WebsiteSearchController.scala
```diff
@@ -286,10 +289,14 @@ class WebsiteSearchController @Inject() (
     val userFilterFuture = getUserFilterFuture(userFilter)
     val libraryFilterFuture = getLibraryFilterFuture(libraryFilter.map(PublicId[Library](_)), libraryAuth, request)
 
+    val parsedOrderBy = orderBy.flatMap(SearchRanking.parse) getOrElse {
+      if (query.contains("tag:")) SearchRanking.recency else SearchRanking.default
+    }
+
```
leogrim, line 295: @andrewconner @stkem @jaredpetker @thassss 
An `"orderBy=recency"` query parameter should now be passed explicitly when calling search to look up keeps with a specific tag and return them in reverse chronological order (`MobileSearchController` has this as well). 
We have backward compatibility with this hack but should start being explicit.

PR #15775, opened 2015-06-13
Simpler filter query logic in search (behind experiment)

leogrim:
@andrewconner let's hope this is a fix


andrewconner:
I would never never been able to come up with this.

leogrim:
To be fair, I don't even know whether this is a fix. And it took me 4 hours to decide how to change 5 lines of code.

PR #15777, opened 2015-06-13
Gratification email: users + their data are now curated on heimdal

camhashemi:
@stkem @andrewconner 

tests pass, the email sends in dev mode but there's no data to populate it. everything scheduled sends to qa.test@kifi while I test it in prod.


kifi-backend/modules/shoebox/app/com/keepit/commanders/emails/GratificationCommander.scala
```diff
@@ -66,42 +48,26 @@ class GratificationCommander @Inject() (
...
-      val fViewsByLibrary: Future[LibraryCountData] = getLibraryViewData(id)
-      fViewsByLibrary.map { x =>
-        (id, x)
-      } recover { case t => (id, LibraryCountData(-1, Map.empty)) } // will filter out as long as MIN_VIEWS >= 0
+  def getGratData(userId: Id[User]): Future[GratificationData] = heimdal.getGratData(userId)
+
+  def getEligibleGratData: Future[Seq[GratificationData]] = {
+
+    def batchGetGratData(userIds: Seq[Id[User]]): Future[Seq[GratificationData]] = {
+      val BATCH_SIZE = 1000
```
stkem: This might be a bit large of a batch. Keep an eye on it when it first runs.

kifi-backend/modules/shoebox/app/com/keepit/commanders/emails/GratificationCommander.scala
```diff
@@ -66,42 +48,26 @@ class GratificationCommander @Inject() (
...
+      val batches = {
+        for (batch <- userIds.grouped(BATCH_SIZE)) yield {
```
stkem: This will make `number_of_users/BATCH_SIZE` _concurrent_ calls to Heimdal which is likely going to bring that to it's knees (you've run into that before). You'll either need a reactive lock (your `remoteCallQueue` is now unused and should be removed unless you use it here) or perhaps consider using `FutureHelpers.chunkySequentialExec` or one of the other sequentializing helpers in there (in `PimpedFuture.scala`)

kifi-backend/modules/common/app/com/keepit/model/GratificationData.scala
```diff
@@ -0,0 +1,30 @@
...
+    rekeeps: KeepCountData) {
+  import GratificationData._
+  def isEligible = libraryViews.totalCount >= MIN_LIB_VIEWS || keepViews.totalCount >= MIN_KEEP_VIEWS || rekeeps.totalCount >= MIN_REKEEPS
+}
+
+object GratificationData {
+  val MIN_KEEP_VIEWS = 5
+  val MIN_LIB_VIEWS = 5
+  val MIN_REKEEPS = 5
+}
```
andrewconner: New line. Your editor can do this for you.

kifi-backend/modules/shoebox/app/com/keepit/commanders/emails/GratificationCommander.scala
andrewconner: Good catch. This is necessary. Batches of 1000 are fine, as long as we do them one at a time.

kifi-backend/modules/shoebox/app/com/keepit/commanders/emails/GratificationCommander.scala
```diff
@@ -66,42 +48,26 @@ class GratificationCommander @Inject() (
...
-  def filterUsersWithoutData(userIds: Seq[Id[User]]): Future[Seq[Id[User]]] = {
-
-    val idAndViewByLib: Seq[Future[(Id[User], LibraryCountData)]] = userIds.map { id =>
-      val fViewsByLibrary: Future[LibraryCountData] = getLibraryViewData(id)
-      fViewsByLibrary.map { x =>
-        (id, x)
-      } recover { case t => (id, LibraryCountData(-1, Map.empty)) } // will filter out as long as MIN_VIEWS >= 0
+  def getGratData(userId: Id[User]): Future[GratificationData] = heimdal.getGratData(userId)
+
+  def getEligibleGratData: Future[Seq[GratificationData]] = {
```
andrewconner: This will be a hotspot as well. It may work fine now, but we'll end up having everyone we email in memory at once. When we start to see issues, emails should be _sent_ in batches as well.

kifi-backend/modules/shoebox/app/com/keepit/commanders/emails/GratificationCommander.scala
```diff
@@ -66,42 +48,26 @@ class GratificationCommander @Inject() (
...
+          } else {
+            heimdal.getEligibleGratData(batch)
+          }
+        }
+      }.toSeq
+      Future.sequence(batches).map { _.flatten }
     }
 
-    val fIdAndViewByLib: Future[Seq[(Id[User], LibraryCountData)]] = Future.sequence(idAndViewByLib)
+    val userIds: Seq[Id[User]] = db.readOnlyReplica { implicit session => userRepo.getAllIds() }.toSeq
```
andrewconner: Usually, with large processing tasks, it's best to think of the entire batch at once. So generate a batch → fetch data → perform action, repeat. This, verses getting all user ids → fetching all eligible user data → sending all emails.

camhashemi: Makes sense! Thank you.

camhashemi:
updated and moved to #15895

PR #15778, opened 2015-06-13
fix misleading error msg

yingjieMiao:
@andrewconner 

it still happens


andrewconner:
Wow. So odd. I'll look at it more closely this weekend. Any hints?

PR #15779, opened 2015-06-13
Drafted up a permissions-based org membership

ryanpbrewster:
Also changed OrganizationAccess to OrganizationRole everywhere.

@leogrim @Colin-Lane @andrewconner 
I'm going to go in and flag the places I'm most concerned about feedback

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/model/OrganizationMembership.scala
+++ b/kifi-backend/modules/common/app/com/keepit/model/OrganizationMembership.scala
@@ -15,7 +15,7 @@ case class OrganizationMembership(
     organizationId: Id[Organization],
     userId: Id[User],
     role: OrganizationRole,
-    permissions: Seq[OrganizationPermission]) extends ModelWithState[OrganizationMembership] with ModelWithSeqNumber[OrganizationMembership] {
+    permissions: Set[OrganizationPermission]) extends ModelWithState[OrganizationMembership] with ModelWithSeqNumber[OrganizationMembership] {
 
   def withId(id: Id[OrganizationMembership]): OrganizationMembership = this.copy(id = Some(id))
   def withUpdateTime(now: DateTime): OrganizationMembership = this.copy(updatedAt = now)
@@ -25,6 +25,17 @@ case class OrganizationMembership(
 
   def isOwner: Boolean = role == OrganizationRole.OWNER
   def hasRole(r: OrganizationRole): Boolean = r == role
+
+  implicit def format[T]: Format[Set[OrganizationPermission]] = new Format[Set[OrganizationPermission]] {
+    def reads(json: JsValue): JsResult[Set[OrganizationPermission]] = json match {
+      case JsArray(ps) =>
+        JsSuccess(ps.map(OrganizationPermission.format.reads(_).get).toSet)
+      case _ => JsError()
+    }
+    def writes(obj: Set[OrganizationPermission]): JsValue = {
+      Json.toJson(obj.toSeq)
+    }
+  }
 }
 
 object OrganizationMembership {
@@ -37,7 +48,7 @@ object OrganizationMembership {
     (__ \ 'organizationId).format[Id[Organization]] and
     (__ \ 'userId).format[Id[User]] and
     (__ \ 'role).format[OrganizationRole] and
-    (__ \ 'permissions).format[Seq[OrganizationPermission]]
+    (__ \ 'permissions).format[Set[OrganizationPermission]]
   )(OrganizationMembership.apply, unlift(OrganizationMembership.unapply))
 }
 object OrganizationMembershipStates extends States[OrganizationMembership]
@@ -62,13 +73,6 @@ object OrganizationPermission {
       case ADD_LIBRARIES.value => ADD_LIBRARIES
       case REMOVE_LIBRARIES.value => REMOVE_LIBRARIES
     }
-  }
-
-  def serialize(ps: Seq[OrganizationPermission]): String = {
-    ps.map(_.value).sorted.mkString(",")
-  }
-  def deserialize(str: String): Seq[OrganizationPermission] = {
-    str.split(",").map(OrganizationPermission(_))
   }
 }
 
--- a/kifi-backend/modules/common/app/com/keepit/model/OrganizationMembership.scala
+++ b/kifi-backend/modules/common/app/com/keepit/model/OrganizationMembership.scala
@@ -1,6 +1,5 @@
 package com.keepit.model
 
-import com.keepit.common.crypto.ModelWithPublicId
 import com.keepit.common.db._
 import com.keepit.common.time._
 import org.joda.time.DateTime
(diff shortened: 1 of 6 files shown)


kifi-backend/modules/common/app/com/keepit/model/OrganizationMembership.scala
```diff
@@ -15,14 +14,17 @@ case class OrganizationMembership(
     seq: SequenceNumber[OrganizationMembership] = SequenceNumber.ZERO,
     organizationId: Id[Organization],
     userId: Id[User],
-    access: OrganizationAccess) extends ModelWithState[OrganizationMembership] with ModelWithSeqNumber[OrganizationMembership] {
+    role: OrganizationRole,
+    permissions: Seq[OrganizationPermission]) extends ModelWithState[OrganizationMembership] with ModelWithSeqNumber[OrganizationMembership] {
```
ryanpbrewster: I want this to be Set[OrganizationPermission], but Set[T] is more work to (de)serialize.

kifi-backend/modules/common/app/com/keepit/model/OrganizationMembership.scala
```diff
@@ -34,33 +36,76 @@ object OrganizationMembership {
...
+object OrganizationPermission {
+  case object EDIT_ORGANIZATION extends OrganizationPermission("edit_organization")
+  case object INVITE_MEMBERS extends OrganizationPermission("invite_members")
+  case object ADD_LIBRARIES extends OrganizationPermission("add_libraries")
+  case object REMOVE_LIBRARIES extends OrganizationPermission("remove_libraries")
+
+  implicit def format[T]: Format[OrganizationPermission] =
+    Format(__.read[String].map(OrganizationPermission(_)), new Writes[OrganizationPermission] {
+      def writes(o: OrganizationPermission) = JsString(o.value)
+    })
```
ryanpbrewster, line 67: In the codebase I see two ways to convert to/from Json. This way, with the explicit Reads[T] and Writes[T], and the other way with the implicit val format and the (__ \ 'field) declarations. Which is preferred?

kifi-backend/modules/common/app/com/keepit/model/OrganizationMembership.scala
```diff
@@ -34,33 +36,76 @@ object OrganizationMembership {
...
 
-object OrganizationAccess {
-  case object READ_WRITE extends OrganizationAccess("read_write", 0)
-  case object OWNER extends OrganizationAccess("owner", 1)
+  def serialize(ps: Seq[OrganizationPermission]): String = {
+    ps.map(_.value).sorted.mkString(",")
+  }
+  def deserialize(str: String): Seq[OrganizationPermission] = {
+    str.split(",").map(OrganizationPermission(_))
+  }
```
ryanpbrewster: These two methods feel very fragile.

After talking to @stkem I realized I should just use the toJson and fromJson methods that already exist. Have to be careful to get the order or the permissions right though?

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
```diff
@@ -47,25 +47,25 @@ class OrganizationMembershipCommander @Inject() (db: Database,
...
       val contact = BasicContact(invitedByAddress.emailAddress.get)
-      MaybeOrganizationMember(member = Right(contact), access = Some(invitedByAddress.access), lastInvitedAt = Some(invitedByAddress.updatedAt))
+      MaybeOrganizationMember(member = Right(contact), role = Some(invitedByAddress.role), lastInvitedAt = Some(invitedByAddress.updatedAt))
     }
 
     membersNotIncludingOwner ++ invitedByUserId ++ invitedByEmailAddress
   }
 
-  def modifyMemberships(orgId: Id[Organization], requestorId: Id[User], modifications: Seq[(Id[User], OrganizationAccess)]): Either[OrganizationFail, Seq[OrganizationMembership]] = ???
+  def modifyMemberships(orgId: Id[Organization], requestorId: Id[User], modifications: Seq[(Id[User], OrganizationRole)]): Either[OrganizationFail, Seq[OrganizationMembership]] = ???
```
ryanpbrewster, line 68: I think these type signatures ought to take Either[OrganizationRole, Seq[OrganizationPermission]] now.

So you can either say "promote this user to owner" or "let this user send invites".

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationMembershipRepo.scala
```diff
@@ -30,11 +30,13 @@ class OrganizationMembershipRepoImpl @Inject() (val db: DataBaseComponent, val c
 
   type RepoImpl = OrganizationMembershipTable
   class OrganizationMembershipTable(tag: Tag) extends RepoTable[OrganizationMembership](db, tag, "organization_membership") with SeqNumberColumn[OrganizationMembership] {
-    implicit val organizationAccessMapper = MappedColumnType.base[OrganizationAccess, String](_.value, OrganizationAccess(_))
+    implicit val organizationRoleMapper = MappedColumnType.base[OrganizationRole, String](_.value, OrganizationRole(_))
+    implicit val organizationPermissionsMapper = MappedColumnType.base[Seq[OrganizationPermission], String](OrganizationPermission.serialize, OrganizationPermission.deserialize)
```
ryanpbrewster: This is where the fragile code I pointed out above interacts with the DB.

kifi-backend/modules/common/app/com/keepit/model/OrganizationMembership.scala
Colin-Lane: If you want this to be a set, you could always just call toSet on the Seq.

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationInviteRepo.scala
```diff
@@ -23,13 +23,13 @@ class OrganizationInviteRepoImpl @Inject() (val db: DataBaseComponent, val clock
...
   class OrganizationInviteTable(tag: Tag) extends RepoTable[OrganizationInvite](db, tag, "organization_invite") {
-    implicit val organizationAccessMapper = MappedColumnType.base[OrganizationAccess, String](_.value, OrganizationAccess(_))
+    implicit val organizationRoleMapper = MappedColumnType.base[OrganizationRole, String](_.value, OrganizationRole(_))
 
     def organizationId = column[Id[Organization]]("organization_id", O.NotNull)
     def inviterId = column[Id[User]]("inviter_id", O.NotNull)
     def userId = column[Option[Id[User]]]("user_id", O.Nullable)
     def emailAddress = column[Option[EmailAddress]]("email_address", O.Nullable)
-    def access = column[OrganizationAccess]("access", O.NotNull)
+    def role = column[OrganizationRole]("role", O.NotNull)
```
Colin-Lane, line 32: When you remove or add a column, you should create an `evolution` to keep track of these changes in the database.
We have our evolutions in `common/conf/evolutions.shoebox`.
You will need to create one, as well as make sure it works by running the server locally, or writing tests to test it (`ShoeboxRepoTest` is where we actually spawn an application and test our evolutions).

kifi-backend/modules/common/app/com/keepit/model/OrganizationMembership.scala
leogrim, line 67: `(__ \ 'field)` works only for Js Object serialization. When you want to serialize something to something else (here, a JS String), you have to use explicit `Reads[T]` and `Writes[T]`

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
leogrim, line 68: I would take both. This is a `modify` endpoint, so the full state of a membership should be submitted - otherwise it's unclear whether these permissions are added or removed. If we don't allow tweaking of individual permissions and the client doesn't send them for now, let's hide this method and expose a simpler one that just takes a role and submits the default.

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationInviteRepo.scala
leogrim, line 32: In this case, you can just modify the existing evolution (`335.sql`) since we haven't applied it yet.

PR #15780, opened 2015-06-13
Admin utility to look up publicIds by ids, and vice versa

andrewconner:
@stkem @leogrim This has slowed me down quite a few times.


andrewconner:
Should be pretty safe. Feedback appreciated.

PR #15781, opened 2015-06-13
Splitting routes into functional groups

andrewconner:
Makes it a bit easier to manage routes, at the slight cost of knowing which file it's in (the groups are pretty straight forward, though). Should help clients find what they're looking for.

The big gain here is compile speed.


andrewconner:
@eishay @leogrim @stkem @yingjieMiao

andrewconner:
Nevermind, it was a good thought, but there are a ton of conflicts for tests (conflicts happen when two route files define the same reverse router — it's usually laziness or some circular reference, but this will take a while to resolve.)

PR #15782, opened 2015-06-15
using the "library_collaborated" for mobile?

eishay:


eishay:
@andrewconner

eishay:
I see the following comment in the code, is their code in the app store now?

``` scala
// This should be changed to library_collaborated but right now iOS skips categories it doesn't know.
```

andrewconner:
No, this is correct right now (don't merge). @mrbrown14 and I discussed the best option, the `follow` is the best we can do right now. He may be able to elaborate, not sure if the fix made the quick release they just did with the new app store assets.

PR #15783, opened 2015-06-15
creates EC2 instance provisioning scripts using ansible

joshm1:
@stkem @andrewconner 

I added documentation in `provisioning/README.md`.  In short, I've tested ansible against clean Ubuntu 14.04 instances and the existing AMI and have been able to launch, provision, and deploy shoebox to new instances using 1 command: `ansible -i ec2.py production.yml`

The inventory list (instances to run ansible against) is dynamic, so any instances with the tag Ansible=v1 will be provisioned when this is run.

A caveat for launching new instances using the production.yml playbook is that the idempotency will not work across different machines because it uses a local cache to determine if an instance has been launched already.  I assume you'll want this on arthur, so that shouldn't be an issue.

Since the inventory list is dynamic, instances do not need to be persisted (committed to git) in the `vars/ec2.yml` file as they are now (which is what I last used for testing).

This PR only supports shoebox, but adding support for the rest of the services is a small amount of additional work (see the kifi.shoebox files). I wanted to show this before I did that.


joshm1:
this PR isn't as big as it looks. most of the LOC are from ec2.ini and ec2.py which I did not write (source: https://github.com/ansible/ansible/tree/devel/plugins/inventory)

andrewconner:
Nice! Will look at this more closely today.

provisioning/roles/kifi.common/tasks/main.yml
```diff
@@ -0,0 +1,13 @@
+---
+- include: users.yml
+- include: packages.yml
+- include: nginx.yml
+  when: setup_nginx
+- include: service.yml
+- include: eddie.yml
+- include: log_stats.yml
+
+# - get_url: url=https://www.yourkit.com/download/yjp-2014-build-14120-linux.tar.bz2 dest=/opt
```
joshm1, line 10: YourKit is 1 feature I haven't added to ansible yet because the download URL for the 2014 version doesn't work anymore.

provisioning/roles/kifi.shoebox/meta/main.yml
```diff
@@ -0,0 +1,5 @@
+---
+dependencies:
+- role: kifi.common
+  service: shoebox
+  setup_nginx: yes
```
joshm1, line 5: this is really all the kifi.shoebox role is.  other roles will be similar.  I put them into roles incase we have servers that require special setup, it'd be easy to extend w/o touching other services because all that'd need to be done is editing that role and not touching kifi.common

provisioning/roles/kifi.common/tasks/main.yml
andrewconner, line 10: Our license works with the 2015 version

joshm1:
@andrewconner, I pushed updates to:

1) ansible roles created for all kifi services
2) set higher soft & hard ulimits: https://github.com/joshm1/keepit/commit/3c86a04c5aab1a520822d3c8ba051fbce629c103

PR #15785, opened 2015-06-15
Permissions-based organization membership

ryanpbrewster:
@leogrim @Colin-Lane 

OrganizationMembership is now defined by a OrganizationRole (basically, a title
within the organization, e.g. "owner" or "member") and a Set[OrganizationPermission].
There is an OrganizationPermission for basically every action a member of an organization
can take (e.g., ADD_LIBRARY, EDIT_ORGANIZATION). Each OrganizationRole comes with
a default set of permissions.


kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationMembershipRepo.scala
```diff
@@ -30,11 +31,21 @@ class OrganizationMembershipRepoImpl @Inject() (val db: DataBaseComponent, val c
 
   type RepoImpl = OrganizationMembershipTable
   class OrganizationMembershipTable(tag: Tag) extends RepoTable[OrganizationMembership](db, tag, "organization_membership") with SeqNumberColumn[OrganizationMembership] {
-    implicit val organizationAccessMapper = MappedColumnType.base[OrganizationAccess, String](_.value, OrganizationAccess(_))
+    implicit val organizationRoleMapper = MappedColumnType.base[OrganizationRole, String](_.value, OrganizationRole(_))
+    implicit val organizationPermissionsMapper = MappedColumnType.base[Set[OrganizationPermission], JsValue](
+      { permissions => Json.toJson(permissions.toSeq) },
+      { json =>
+        Json.fromJson[Seq[OrganizationPermission]](json) match {
+          case success: JsSuccess[Seq[OrganizationPermission]] => success.value.toSet
```
Colin-Lane: Is the type supposed to be `Set[OrganizationPermission]` looks like you are returning a `Set` but the generic type parameter is a `Seq`

kifi-backend/modules/common/app/com/keepit/model/OrganizationMembership.scala
```diff
@@ -34,33 +48,71 @@ object OrganizationMembership {
...
+      OrganizationPermission.INVITE_MEMBERS,
+      OrganizationPermission.ADD_LIBRARIES,
+      OrganizationPermission.REMOVE_LIBRARIES
+    )
+  }
+  case object MEMBER extends OrganizationRole("member", 1) {
+    def defaultPermissions = Set(
+      OrganizationPermission.ADD_LIBRARIES
+    )
+  }
```
leogrim, line 99: Since `OrganizationRole` is sealed, it would be better to implement default as a method on `OrganizationPermission`: `def default(role: OrganizationRole): Set[OrganizationPermission]` relying on a `match` statement on role. This way the compiler will check that you have all roles covered and, more importantly, you can use that "factory" method in an abstract way whenever you get a role without an explicit set of permissions.

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationMembershipRepo.scala
```diff
@@ -30,11 +31,21 @@ class OrganizationMembershipRepoImpl @Inject() (val db: DataBaseComponent, val c
...
   class OrganizationMembershipTable(tag: Tag) extends RepoTable[OrganizationMembership](db, tag, "organization_membership") with SeqNumberColumn[OrganizationMembership] {
-    implicit val organizationAccessMapper = MappedColumnType.base[OrganizationAccess, String](_.value, OrganizationAccess(_))
+    implicit val organizationRoleMapper = MappedColumnType.base[OrganizationRole, String](_.value, OrganizationRole(_))
+    implicit val organizationPermissionsMapper = MappedColumnType.base[Set[OrganizationPermission], JsValue](
+      { Json.toJson(_) },
+      { json =>
+        Json.fromJson[Set[OrganizationPermission]](json) match {
+          case success: JsSuccess[Set[OrganizationPermission]] => success.value
+          case failure: JsError => throw new JsResultException(failure.errors)
+        }
```
leogrim, line 41: If you end up throwing an exception, you can just do `json.as[Set[OrganizationPermission]]`

leogrim, line 41: Additionally, our mappers go between a high-level type (here `Set[OrganizationPermission]`) and a database type (here `String`), so I wouldn't have a mapper between `Set[OrganizationPermission]` and `JsValue`. Actually, that's exactly what a `Format` is.

leogrim:
Just a couple of small things to modify (see comments). Great job!

kifi-backend/modules/common/app/com/keepit/model/OrganizationMembership.scala
```diff
@@ -15,14 +14,29 @@ case class OrganizationMembership(
...
+    def reads(json: JsValue): JsResult[Set[OrganizationPermission]] = json match {
+      case JsArray(ps) =>
+        JsSuccess(ps.map(OrganizationPermission.format.reads(_).get).toSet)
+      case _ => JsError()
+    }
+    def writes(obj: Set[OrganizationPermission]): JsValue = {
+      Json.toJson(obj.toSeq)
+    }
+  }
 }
```
leogrim, line 40: You shouldn't have to define this format.

PR #15787, opened 2015-06-15
added subscriptions to another library API, polished front-end

camhashemi:
@andrewconner should be ready for QA. Will hasn't looked at the front-end yet, but I think it's presentable.


camhashemi:
merging, won't affect production

PR #15788, opened 2015-06-15
Added an index to the user ip address table

ryanpbrewster:
@sktem

Should speed up the most common query for this table


andrewconner:
Great

PR #15790, opened 2015-06-15
OrganizationInviteRepo

Colin-Lane:
- implementing repo methods.

### DIFF
--- a/kifi-backend/modules/shoebox/test/com/keepit/model/OrganizationInviteRepoTest.scala
+++ b/kifi-backend/modules/shoebox/test/com/keepit/model/OrganizationInviteRepoTest.scala
@@ -4,6 +4,8 @@ import com.keepit.common.db.Id
 import com.keepit.test.ShoeboxTestInjector
 import org.specs2.mutable.Specification
 
+import scala.collection.immutable.IndexedSeq
+
 class OrganizationInviteRepoTest extends Specification with ShoeboxTestInjector {
 
   "Organization Invite Repo" should {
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/test/com/keepit/model/OrganizationInviteRepoTest.scala
```diff
@@ -19,5 +21,24 @@ class OrganizationInviteRepoTest extends Specification with ShoeboxTestInjector
...
       }
     }
+
+    "get by inviter id" in {
+      withDb() { implicit injector =>
+        val orgInviteRepo = inject[OrganizationInviteRepo]
+        val inviterId = Id[User](1)
+        val userIds: IndexedSeq[Some[Id[User]]] = 10 to 20 map (i => Some(Id[User](i)))
+        db.readWrite { implicit s =>
+          for { inviteeId <- userIds } yield {
```
ryanpbrewster: I don't think you need to yield anything here.

Colin-Lane: Thought that was required in scala for for-comprehensions. Still getting used to the syntax of them. I'll fix this.

andrewconner: `for (...) { ... }` is the same as `for (...) yield { ... }` except that the one without `yield` side effects only. _Usually_ what you want is `for... yield` because we care about the response. In this case, you're just making assertions, which blow up if they fail.

PR #15795, opened 2015-06-15
Simplified code logic for OrganizationMembership

ryanpbrewster:
@leogrim


kifi-backend/modules/common/app/com/keepit/model/OrganizationMembership.scala
```diff
@@ -77,26 +66,25 @@ object OrganizationPermission {
...
   }
+}
+
+sealed abstract class OrganizationRole(val value: String, val priority: Int)
+
+object OrganizationRole {
+  case object OWNER extends OrganizationRole("owner", 0)
+  case object MEMBER extends OrganizationRole("member", 1)
 
   implicit def format[T]: Format[OrganizationRole] =
```
leogrim: No type parameter, can be a `val` as well :)

leogrim:
lgtm, thanks!

PR #15802, opened 2015-06-15
Removing XSS

andrewconner:
@Colin-Lane Self-merging to get the fix out and make sure it works. Right now, a user who's name is `Colin<script>alert(1)</script>` causes an XSS.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/PublicPageMetaTags.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/PublicPageMetaTags.scala
@@ -87,8 +87,8 @@ case class PublicPageMetaFullTags(unsafeTitle: String, url: String, urlPathOnly:
 
   private def includeFeedTags: String = feedName.map { name =>
     s"""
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/PublicPageMetaTags.scala
```diff
@@ -87,8 +87,8 @@ case class PublicPageMetaFullTags(unsafeTitle: String, url: String, urlPathOnly:
 
   private def includeFeedTags: String = feedName.map { name =>
     s"""
-         |<link rel="alternate" type="application/rss+xml" title="$name feed by $unsafeFirstName $unsafeLastName on Kifi" href="$url/rss" />
-         |<link rel="alternate" type="application/atom+xml" title="$name feed by $unsafeFirstName $unsafeLastName on Kifi" href="$url/atom" />
+         |<link rel="alternate" type="application/rss+xml" title="$name feed by $fullName on Kifi" href="$url/rss" />
+         |<link rel="alternate" type="application/atom+xml" title="$name feed by $fullName on Kifi" href="$url/atom" />
```
Colin-Lane, line 91: I was wondering why they were named $unsafeFirstName and $unsafeLastName. I'm guessing they were being read in from the url.

PR #15806, opened 2015-06-16
subscription instructions, UI changes to front-end. Keep note included in Slack message.

camhashemi:
@andrewconner @stkem


stkem:
code looks good to me overall. Would love to see this in action before it goes live though. My build in CSS renderer is not that good :-)

PR #15811, opened 2015-06-16
Three very small tweaks

andrewconner:
1) PublicId admin page now shows all classes that match the search string. Helps with LibraryInvite vs Library.
2) Higher (lower is better) scoring for typeahead search of users with no interactions
3) Fixing a test that had a race due to sorting by a time.

@Colin-Lane @leogrim


andrewconner:
Would love feedback, merging since it's been open for almost a day.

PR #15814, opened 2015-06-16
Added OrganizationAccessActions

ryanpbrewster:
@leogrim

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
@@ -9,7 +9,7 @@ import com.keepit.model._
 @ImplementedBy(classOf[OrganizationCommanderImpl])
 trait OrganizationCommander {
   def get(orgId: Id[Organization]): Organization
-  def canViewOrganization(userIdOpt: Option[Id[User]], orgId: Id[Organization]): Boolean
+  def canViewOrganization(userIdOpt: Option[Id[User]], orgId: Id[Organization], authToken: Option[String]): Boolean
 }
 
 @Singleton
@@ -20,5 +20,5 @@ class OrganizationCommanderImpl @Inject() (
   def get(orgId: Id[Organization]): Organization = db.readOnlyReplica { implicit session => orgRepo.get(orgId) }
 
   // Right now, Organization's are 100% public
-  def canViewOrganization(userIdOpt: Option[Id[User]], orgId: Id[Organization]): Boolean = true
+  def canViewOrganization(userIdOpt: Option[Id[User]], orgId: Id[Organization], authToken: Option[String]): Boolean = true
 }
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
@@ -9,6 +9,7 @@ import com.keepit.model._
 @ImplementedBy(classOf[OrganizationCommanderImpl])
 trait OrganizationCommander {
   def get(orgId: Id[Organization]): Organization
+  def canViewOrganization(userIdOpt: Option[Id[User]], orgId: Id[Organization]): Boolean
 }
 
 @Singleton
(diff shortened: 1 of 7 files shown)


kifi-backend/modules/shoebox/app/com/keepit/shoebox/controllers/OrganizationAccessActions.scala
```diff
@@ -0,0 +1,77 @@
...
+  private def parseRequest[A](orgPubId: PublicId[Organization], input: MaybeUserRequest[A]): Option[(Id[Organization], Option[Id[User]], Option[String])] = {
+    val userIdOpt: Option[Id[User]] = input match {
+      case userRequest: UserRequest[A] => Some(userRequest.userId)
+      case _ => None
+    }
+
+    val orgIdOpt = Organization.decodePublicId(orgPubId).toOption
+    orgIdOpt.map { orgId =>
+      (orgId, userIdOpt, input.getQueryString("authToken"))
+    }
```
Colin-Lane, line 48: It looks like `authToken` is being ignored by the other method calls.

kifi-backend/modules/shoebox/app/com/keepit/shoebox/controllers/OrganizationAccessActions.scala
```diff
@@ -0,0 +1,77 @@
...
+import com.keepit.model.{ OrganizationPermission, Organization, User }
+import play.api.libs.json.Json
+import play.api.mvc.{ ActionFilter, Controller, Result }
+
+import scala.concurrent.Future
+
+trait OrganizationAccessActions {
+  self: UserActions with Controller =>
+
+  val publicIdConfig: com.keepit.common.crypto.PublicIdConfiguration
```
Colin-Lane: Is there a reason to use the fully qualified class type here rather than importing it?

kifi-backend/modules/shoebox/app/com/keepit/shoebox/controllers/OrganizationAccessActions.scala
ryanpbrewster, line 48: We probably ought to use it eventually, so I kept it in.

kifi-backend/modules/shoebox/app/com/keepit/shoebox/controllers/OrganizationAccessActions.scala
ryanpbrewster: Not that I can tell. Changing it now.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
```diff
@@ -17,4 +18,7 @@ class OrganizationCommanderImpl @Inject() (
     orgRepo: OrganizationRepo) extends OrganizationCommander with Logging {
 
   def get(orgId: Id[Organization]): Organization = db.readOnlyReplica { implicit session => orgRepo.get(orgId) }
+
+  // Right now, Organization's are 100% public
+  def canViewOrganization(userIdOpt: Option[Id[User]], orgId: Id[Organization], authToken: Option[String]): Boolean = true
```
andrewconner, line 23: Wonder if this should take different levels of detail. For example, `LibraryViewAction` / `LibraryWriteAction` / `LibraryOwnerAction` are for instances that are very common (only people who have read write should be able to keep, unkeep, edit a note, etc). I'm curious if organizations have the same thing. (As an aside, what does "viewing" an organization mean? able to see the page? seeing libraries?)

Orgs will likely evolve to have more and more levels of control eventually, and a new action type for each will get super messy. I spoke with Léo about perhaps having something like `OrgAccessAction(permission, orgId)`, where permission is in a set of permissions like "view existence of org", "view / join / leave libraries", "modify members in org", etc.

Not sure the best way to do it, we want to keep it as generic as possible so it's useful. In general, actions help us check permissions that we should check _all over the place_ (see library examples above) just so we don't forget. However, for highly specialized cases (such as joining a library), we do it in the commander so that no one can screw it up accidentally. It's a balance, but I'm sure you guys will find a good solution. Forcing the API engineer to _think_ about it is our goal; it's too easy to forget to check things, so forcing the thought takes us a long way.

PR #15817, opened 2015-06-16
Invitation to Organization

Colin-Lane:
@leogrim @ryanpbrewster


kifi-backend/modules/common/app/com/keepit/model/OrganizationMembership.scala
```diff
@@ -92,7 +95,7 @@ object OrganizationRole {
     })
 
   implicit def ord: Ordering[OrganizationRole] = new Ordering[OrganizationRole] {
-    def compare(x: OrganizationRole, y: OrganizationRole): Int = x.priority compare y.priority
+    def compare(x: OrganizationRole, y: OrganizationRole): Int = y.priority compare x.priority
   }
```
ryanpbrewster: Do we need both of these? We already extend Ordered[T].

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
```diff
@@ -15,25 +24,209 @@ import scala.concurrent.{ ExecutionContext, Future }
...
+  private def canInvite(membership: OrganizationMembership): Boolean = {
+    membership.hasPermission(OrganizationPermission.INVITE_MEMBERS)
+  }
 
   def inviteUsersToOrganization(orgId: Id[Organization], inviterId: Id[User], invitees: Seq[(Either[Id[User], EmailAddress], OrganizationRole, Option[String])]): Future[Either[OrganizationFail, Seq[(Either[BasicUser, RichContact], OrganizationRole)]]] = {
-    ???
+    val inviterMembershipOpt = db.readOnlyMaster { implicit session =>
+      organizationMembershipRepo.getByOrgIdAndUserId(orgId, inviterId)
+    }
+    inviterMembershipOpt match {
```
ryanpbrewster, line 53: Honestly, I'd rather use early-exit "return" statements than have code nest this deeply. As it is, you have to scroll down almost 50 lines to see where the None is for this first match.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
```diff
@@ -15,25 +24,209 @@ import scala.concurrent.{ ExecutionContext, Future }
...
-    userRepo: UserRepo) extends OrganizationInviteCommander with Logging {
+    userRepo: UserRepo,
+    elizaClient: ElizaServiceClient,
+    s3ImageStore: S3ImageStore,
+    userEmailAddressRepo: UserEmailAddressRepo,
+    emailTemplateSender: EmailTemplateSender) extends OrganizationInviteCommander with Logging {
+
+  private def canInvite(membership: OrganizationMembership): Boolean = {
+    membership.hasPermission(OrganizationPermission.INVITE_MEMBERS)
+  }
```
ryanpbrewster: I think I prefer inlining this function. It's almost as clear, and it avoids the mental overhead of another function.

Colin-Lane: can I just append `@inline` ?
I can inline it manually.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
```diff
@@ -15,25 +24,209 @@ import scala.concurrent.{ ExecutionContext, Future }
...
+      organizationMembershipRepo.getByOrgIdAndUserId(orgId, inviterId)
+    }
+    inviterMembershipOpt match {
+      case Some(inviterMembership) =>
+        if (canInvite(inviterMembership)) {
+          val inviteesByAddress = invitees.collect { case (Right(emailAddress), _, _) => emailAddress }
+          val inviteesByUserId = invitees.collect { case (Left(userId), _, _) => userId }
+
+          // contacts by email address
+          val futureContactsByEmailAddress: Future[Map[EmailAddress, RichContact]] = {
```
ryanpbrewster: Is this called future because it is a Future[T] or because they will be contacts in the future? If it's the former, use contactsByEmailAddressFut to follow conventions that I've seen elsewhere.

Colin-Lane: it is a Future[T]

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
ryanpbrewster: I think do it manually. I'm more concerned with programmer effort than computer performance.

ryanpbrewster: Dammit.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
```diff
@@ -15,25 +24,209 @@ import scala.concurrent.{ ExecutionContext, Future }
...
+            aBookClient.internKifiContacts(inviterId, inviteesByAddress.map(BasicContact(_)): _*).imap { kifiContacts =>
+              (inviteesByAddress zip kifiContacts).toMap
+            }
+          }
+          // contacts by userId
+          val contactsByUserId = db.readOnlyMaster { implicit s =>
+            basicUserRepo.loadAll(inviteesByUserId.toSet)
+          }
+
+          futureContactsByEmailAddress map { contactsByEmail =>
```
ryanpbrewster: Same future thing.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
```diff
@@ -15,25 +24,209 @@ import scala.concurrent.{ ExecutionContext, Future }
...
+                  case Right(email) =>
+                    val orgInvite = OrganizationInvite(organizationId = orgId, inviterId = inviterId, emailAddress = Some(email), role = inviteRole, message = msgOpt)
+                    val inviteeInfo = (Right(contactsByEmail(email)), inviteRole)
+                    Some((orgInvite, inviteeInfo))
+                }
+              } else {
+                log.warn(s"Inviter $inviterId attempting to grant role inviter does not have.")
+                None
+              }
+            }
```
ryanpbrewster, line 99: I think this whole block of logic could be broken out into its own function.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
```diff
@@ -15,25 +24,209 @@ import scala.concurrent.{ ExecutionContext, Future }
...
+              }.toMap
+            }
+            val invitesForInvitees = for ((recipient, inviteRole, msgOpt) <- invitees) yield {
+              if (inviterMembership.role >= inviteRole) {
+                recipient match {
+                  case Left(inviteeId) =>
+                    organizationMembersMap.get(inviteeId) match {
+                      case Some(inviteeMember) if (inviteeMember.role < inviteRole && inviteRole != OrganizationRole.OWNER) => // member needs upgrade
+                        db.readWrite { implicit session =>
+                          organizationMembershipRepo.save(inviteeMember.copy(role = inviteRole))
```
ryanpbrewster: Use inviteeMember.withRole(inviteRole)

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
```diff
@@ -15,25 +24,209 @@ import scala.concurrent.{ ExecutionContext, Future }
...
+              organizationMembershipRepo.getByOrgIdAndUserIds(orgId, inviteesByUserId.toSet).map { membership =>
+                membership.userId -> membership
+              }.toMap
+            }
+            val invitesForInvitees = for ((recipient, inviteRole, msgOpt) <- invitees) yield {
+              if (inviterMembership.role >= inviteRole) {
+                recipient match {
+                  case Left(inviteeId) =>
+                    organizationMembersMap.get(inviteeId) match {
+                      case Some(inviteeMember) if (inviteeMember.role < inviteRole && inviteRole != OrganizationRole.OWNER) => // member needs upgrade
```
ryanpbrewster: Are we allowing people to invite current members of the organization to join that organization? Seems like a weird way to go about promoting a member.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
```diff
@@ -15,25 +24,209 @@ import scala.concurrent.{ ExecutionContext, Future }
...
+          val contactsByUserId = db.readOnlyMaster { implicit s =>
+            basicUserRepo.loadAll(inviteesByUserId.toSet)
+          }
+
+          futureContactsByEmailAddress map { contactsByEmail =>
+            val organizationMembersMap = db.readOnlyMaster { implicit session =>
+              organizationMembershipRepo.getByOrgIdAndUserIds(orgId, inviteesByUserId.toSet).map { membership =>
+                membership.userId -> membership
+              }.toMap
+            }
```
ryanpbrewster, line 75: Does this need to be readOnlyMaster?

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
```diff
@@ -15,25 +24,209 @@ import scala.concurrent.{ ExecutionContext, Future }
...
+
+            val (invites, inviteesWithRole) = invitesForInvitees.flatten.unzip
+
+            val (org, owner, inviter) = db.readOnlyMaster { implicit session =>
+              val org = organizationRepo.get(orgId)
+              val owner = basicUserRepo.load(org.ownerId)
+              val inviter = userRepo.get(inviterId)
+              (org, owner, inviter)
+            }
```
ryanpbrewster, line 108: I think it is best if we can use as few sessions as possible. Can we combine these distinct sessions somehow?

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
```diff
@@ -15,25 +24,209 @@ import scala.concurrent.{ ExecutionContext, Future }
...
+    }.map { lastInvite =>
+      // determine whether to resend invitation.
+      lastInvite.role != invite.role || lastInvite.createdAt.plusMinutes(5).isBefore(invite.createdAt)
+    }.getOrElse(true)
+    shouldInsert match {
+      case true => db.readWrite { implicit s =>
+        Some(organizationInviteRepo.save(invite))
+      }
+      case false => None
+    }
```
ryanpbrewster, line 148: I think match on Boolean is a little untraditional. :)

Also, can we combine these multiple sessions into a single one? I'm pretty sure it's better to have one 5ms session than 3 1ms sessions.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
```diff
@@ -15,25 +24,209 @@ import scala.concurrent.{ ExecutionContext, Future }
...
+  }
+
+  def sendInvitationEmails(persistedInvites: Seq[OrganizationInvite], org: Organization, owner: BasicUser, inviter: User): Unit = {
+    val (inviteesById, _) = persistedInvites.partition(_.userId.nonEmpty)
```
ryanpbrewster, line 152: Since you aren't doing anything with the non-users, let's use filter instead of partition.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
```diff
@@ -15,25 +24,209 @@ import scala.concurrent.{ ExecutionContext, Future }
...
+
+    // send notifications to kifi users only
+    if (inviteesById.nonEmpty) {
+      notifyInviteeAboutInvitationToJoinLibrary(org, owner, inviter, inviteesById.flatMap(_.userId).toSet)
```
ryanpbrewster: Library leak

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
```diff
@@ -15,25 +24,209 @@ import scala.concurrent.{ ExecutionContext, Future }
...
+        extraHeaders = Some(Map(PostOffice.Headers.REPLY_TO -> fromAddress)),
+        campaign = Some("na"),
+        channel = Some("vf_email"),
+        source = Some("library_invite")
+      )
+      emailTemplateSender.send(emailToSend) map (Some(_))
+    } getOrElse {
+      airbrake.notify(s"LibraryInvite does not have a recipient: $invite")
+      Future.successful(None)
+    }
```
ryanpbrewster, line 194: Here it's less of an issue since the method is shorter, but I still prefer early-exit for situations like this, where a None means we have failed.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
```diff
@@ -15,25 +24,209 @@ import scala.concurrent.{ ExecutionContext, Future }
...
+      val trimmedInviteMsg = invite.message map (_.trim) filter (_.nonEmpty)
+      val fromUserId = invite.inviterId
+      val fromAddress = db.readOnlyReplica { implicit session => userEmailAddressRepo.getByUser(invite.inviterId).address }
+      val authToken = invite.authToken
+      val emailToSend = EmailToSend(
+        fromName = Some(Left(invite.inviterId)),
+        from = SystemEmailAddress.NOTIFICATIONS,
+        subject = s"Please join our organization ${org.name} on Kifi!",
+        to = toRecipient,
+        category = toRecipient.fold(_ => NotificationCategory.User.LIBRARY_INVITATION, _ => NotificationCategory.NonUser.LIBRARY_INVITATION),
```
ryanpbrewster: Library leak

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
```diff
@@ -15,25 +24,209 @@ import scala.concurrent.{ ExecutionContext, Future }
...
+        htmlTemplate = views.html.email.organizationInvitationPlain(toRecipient.left.toOption, fromUserId, trimmedInviteMsg, org, authToken),
+        textTemplate = Some(views.html.email.organizationInvitationText(toRecipient.left.toOption, fromUserId, trimmedInviteMsg, org, authToken)),
+        templateOptions = Seq(CustomLayout).toMap,
+        extraHeaders = Some(Map(PostOffice.Headers.REPLY_TO -> fromAddress)),
+        campaign = Some("na"),
+        channel = Some("vf_email"),
+        source = Some("library_invite")
```
ryanpbrewster: Library leak

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
```diff
@@ -15,25 +24,209 @@ import scala.concurrent.{ ExecutionContext, Future }
...
+      )
+      emailTemplateSender.send(emailToSend) map (Some(_))
+    } getOrElse {
+      airbrake.notify(s"LibraryInvite does not have a recipient: $invite")
```
ryanpbrewster: Library leak

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
```diff
@@ -15,25 +24,209 @@ import scala.concurrent.{ ExecutionContext, Future }
...
+      Future.successful(None)
+    }
   }
+
+  def notifyInviteeAboutInvitationToJoinLibrary(org: Organization, orgOwner: BasicUser, inviter: User, invitees: Set[Id[User]]) {
```
ryanpbrewster: Library leak

ryanpbrewster:
Monster commit. Nice work!

kifi-backend/modules/common/app/com/keepit/model/OrganizationMembership.scala
Colin-Lane: Not really, we can probably remove `Ordering[OrganizationRole]`

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
Colin-Lane: I agree, it is a little strange. That is how `LibraryInviteCommander` handled it. Do we have a different process in-place for promoting someone? Also, since we only have two roles `MEMBER` and `OWNER` there is never a case when someone can currently be promoted: see the test comments.
I can remove that logic for now and just use `are you a member? yes? do not modify`

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
Colin-Lane: One other thing is that saving a person with a new role doesn't set his permissions. We will need logic somewhere to set their default permissions.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
Colin-Lane: :disappointed: I need to catch these better.

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationInviteRepo.scala
```diff
@@ -12,6 +12,8 @@ import org.joda.time.DateTime
 trait OrganizationInviteRepo extends Repo[OrganizationInvite] {
   def getByOrganization(organizationId: Id[Organization], limit: Limit, offset: Offset, state: State[OrganizationInvite] = OrganizationInviteStates.ACTIVE)(implicit s: RSession): Seq[OrganizationInvite]
   def getByInviter(inviterId: Id[User], state: State[OrganizationInvite] = OrganizationInviteStates.ACTIVE)(implicit s: RSession): Seq[OrganizationInvite]
+  def getLastSentByOrgIdAndInviterIdAndUserId(organizationId: Id[Organization], inviterId: Id[User], userId: Id[User], includeStates: Set[State[OrganizationInvite]])(implicit s: RSession): Option[OrganizationInvite]
+  def getLastSentByOrgIdAndInviterIdAndEmailAddress(organizationId: Id[Organization], inviterId: Id[User], emailAddress: EmailAddress, includeStates: Set[State[OrganizationInvite]])(implicit s: RSession): Option[OrganizationInvite]
```
Colin-Lane, line 16: Looks like I forgot to write tests for these.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
Colin-Lane, line 148: I think we spoke about this just yesterday as well.

```
if (shouldInsert) {
   Some
} else {
   None
}
```

Not sure which I like better because this happens to be the last statement and the `return`, and so I chose not to use an `if` for the `implicit return`.
I can change it to an if, if desired.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
Colin-Lane, line 152: filter would be better here. I was originally going to use it, but then ended up not needing it.

kifi-backend/modules/common/app/com/keepit/model/OrganizationMembership.scala
leogrim: +1

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
leogrim, line 53: We don't use return statements in our code, that would be very unusual and surprising for external readers. Not relying on return statements forces your to be rigorous with your intermediate types, overall I believe this constraint can lead to better structured code.

They are two ways to structure nested code depending on whether the successful cases are listed first or second. 

With successful branches first, it's very easy to follow the successful path as all successful steps are one below the other in the code. This makes it easier to understand the programmer's main intent. Now, it's harder to follow the outcome of each individual step as the successful branch and the exit branch are far away from each other. 

Conversely when the unsuccessful "exit" cases are listed first, as they're usually shorter, it's easier to track the outcome of each branching step ("exit now or continue with more logic"). Thus it's easier to track the branch responsible for a specific error when debugging your implementation. But it's harder to follow the successful path, since an exit case sits in between each successful case, and the general intent of your code. 

As the code's author, there's an editorial decision about your narrative: do you want to make it easy to understand the default successful scenario (focusing on intent) or each individual branching step (focusing on implementation)?

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
leogrim: +1 making permissions explicit in our model get's us that nice syntax :)

kifi-backend/modules/common/app/com/keepit/model/NotificationCategory.scala
```diff
@@ -50,7 +52,7 @@ object NotificationCategory {
 
     val reportToAnalytics = Set(ANNOUNCEMENT, MESSAGE, EMAIL_KEEP, EMAIL_CONFIRMATION, RESET_PASSWORD, FRIEND_REQUEST,
       FRIEND_ACCEPTED, WELCOME, APPROVED, WAITLIST, WHO_KEPT_MY_KEEP, CONTACT_JOINED, CONNECTION_MADE,
-      SOCIAL_FRIEND_JOINED, LIBRARY_FOLLOWING, LIBRARY_INVITATION, DIGEST, ACTIVITY, GRATIFICATION_EMAIL)
+      SOCIAL_FRIEND_JOINED, LIBRARY_FOLLOWING, LIBRARY_INVITATION, ORGANIZATION_INVITATION, DIGEST, ACTIVITY, GRATIFICATION_EMAIL)
```
leogrim, line 55: Sweet, thanks for thinking of tracking!

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
```diff
@@ -15,25 +24,209 @@ import scala.concurrent.{ ExecutionContext, Future }
...
+  }
 
   def inviteUsersToOrganization(orgId: Id[Organization], inviterId: Id[User], invitees: Seq[(Either[Id[User], EmailAddress], OrganizationRole, Option[String])]): Future[Either[OrganizationFail, Seq[(Either[BasicUser, RichContact], OrganizationRole)]]] = {
-    ???
+    val inviterMembershipOpt = db.readOnlyMaster { implicit session =>
+      organizationMembershipRepo.getByOrgIdAndUserId(orgId, inviterId)
+    }
+    inviterMembershipOpt match {
+      case Some(inviterMembership) =>
+        if (canInvite(inviterMembership)) {
```
leogrim: Case statements support filter, so you could actually remove your nested if-then-else with an extra case statement:
`case Some(inviterMembership) if canInvite(inviterMembership) =>`
`case Some(invalidMembership) =>`
`case None =>`

I really like Filters on case statements, they're a powerful way to eliminate nested constructs and make your code shorter and clearer.

In some scenarios, `Some(invalidSomething)` and `None` are actually handled the same way, and you cant just use a catch-all `_`. Filters are also a good way of avoiding mixing `case` statements with `if-then-else`, I feel like nested code is easier to read when sticking with one branching approach or the other.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
leogrim: @andrewconner what do you think? We have a similar situation with libraries.
Regardless, try and forget that we only have two roles for now, we should try to abstract our handling of roles and permissions as much as possible, it will make it easier to introduce more in the future without having to make sure edge cases are handled everywhere.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
leogrim: I guess `OrganizationMembership.withRole` could handle that, adding / removing the difference in default permissions between the two roles. Makes sense?

Colin-Lane: I think @ryanpbrewster  was writing some logic in `OrganizationMembershipCommander` to handle properly setting permissions, but `withRole` seems like a temporary solution.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
```diff
@@ -15,25 +24,209 @@ import scala.concurrent.{ ExecutionContext, Future }
...
+            def trackSentInvitation = ???
+
+            Right(inviteesWithRole)
+          }
+        } else {
+          Future.successful(Left(OrganizationFail(200, "cannot_invite_members")))
+        }
+      case None =>
+        Future.successful(Left(OrganizationFail(401, "inviter_not_a_member")))
+    }
```
leogrim, line 122: You could use `org.apache.http.HttpStatus` for status code constants here.

Colin-Lane, line 122: I prefer the `play.api.http.Status` codes better, `OK` vs `SC_OK` but yes, I will use them.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
```diff
@@ -15,25 +24,209 @@ import scala.concurrent.{ ExecutionContext, Future }
...
+    elizaClient: ElizaServiceClient,
+    s3ImageStore: S3ImageStore,
+    userEmailAddressRepo: UserEmailAddressRepo,
+    emailTemplateSender: EmailTemplateSender) extends OrganizationInviteCommander with Logging {
+
+  private def canInvite(membership: OrganizationMembership): Boolean = {
+    membership.hasPermission(OrganizationPermission.INVITE_MEMBERS)
+  }
 
   def inviteUsersToOrganization(orgId: Id[Organization], inviterId: Id[User], invitees: Seq[(Either[Id[User], EmailAddress], OrganizationRole, Option[String])]): Future[Either[OrganizationFail, Seq[(Either[BasicUser, RichContact], OrganizationRole)]]] = {
```
yingjieMiao: input parameter type and output parameter type are hard to parse. Please consider create (private?) case class or type alias to improve readability.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
ryanpbrewster: So membership.withRole(newRole) automatically gives that user two things:
    role = newRole
    permissions = OrganizationPermission.defaultPermissions(newRole)

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
andrewconner, line 53: Léo's points are spot on. Being able to look at the last expression of a method to see what gets returned gives a lot of clarity. Also, a technical reason why `return` is avoided when possible: Scala implements `return` using exceptions when calling from a lambda (with no stack, for efficiency). You can think of an explicit return as an interrupt in normal control flow, just as an exception is. Consider:

``` scala
def a: Int = 3 // return 3, as expected
def a: Int = {
  try { return 3 } catch { case ex: Throwable => 4 }
} // returns 3, shouldn't be too surprising
def a: Int = {
  try {
    Seq(3).foreach { i => return i };
    0 // what?! Because `a`'s type is `Int`, so we *have* to end with an Int.
  } catch { case ex: Throwable => 4 }
} // returns 4. wtf? It's because an exception is thrown to get out of the `foreach` lambda, and then we catch it.
```

Recall that in the JVM, functions are objects. So here's some meta-code for how foreach works:

``` scala
val f = new Function1[Int, Unit] {
  override def apply(i: Int) = return i
}
Seq.apply(3).foreach(f)
// The exception will be thrown inside the `Function1` block, so if you catch it, you interrupted control flow
```

All avoided by avoiding `return`
This can yield some surprising behavior. Your initial function may do it right, but a small refactor may introduce this surprising bug.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
ryanpbrewster: There is backend logic for modifying memberships (either promoting or demoting). As for frontend...maybe it makes sense on that end to allow users to "invite" someone to promote them.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
Colin-Lane: The ideal situation would be to not handle updating a user's role here at all (during an invitation), but make that a separate operation altogether

ryanpbrewster: You can call orgMembershipCommander.modifyMembership or addMembership directly. That has all the validation in it. And we have already paid the price of opening a new DB session, so it won't even be more expensive.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
ryanpbrewster, line 53: Jesus that's gross. New plan: no returns. :frowning:

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
leogrim: I think you're right, updating a user's role through an invitation is kind of weird, better to separate these if possible.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
leogrim, line 122: Oh you're right :)

leogrim:
Nice piece of work!

Colin-Lane:
Going to merge this for now, rather get the code in than wait too long with a large commit. I'll continue on implementing tracking.

PR #15819, opened 2015-06-16
frontend-rss-tracking

Colin-Lane:
@andrewconner Is this all that is required to track a click event?


andrewconner:
I think that's it. Merge, and we'll check with product.

PR #15823, opened 2015-06-17
Revert "Revert "Revert "updating marketing page for teams"""

andrewconner:
Reverts kifi/keepit#15822


andrewconner:
@stkem It's broken because the router needs to point to the new pages. Some updated (ie, the existing page updated), some didn't (ie, there's a new page that replaces the old one).

PR #15826, opened 2015-06-17
keep-org-id integrity check.

Colin-Lane:
@leogrim @yingjieMiao


kifi-backend/modules/shoebox/app/com/keepit/integrity/LibraryChecker.scala
```diff
@@ -29,12 +30,40 @@ class LibraryChecker @Inject() (val airbrake: AirbrakeNotifier,
...
+      val lastSeq = getLastSeqNum(LIBRARY_MOVED_NAME)
+      val libraries = libraryRepo.getBySequenceNumber(lastSeq, LIBRARY_FETCH_SIZE)
+      val keepsNeedingUpdate = keepRepo.getByLibraryIds((libraries.map(_.id.get).toSet))
+
+      val nextSeqNum = if (libraries.isEmpty) None else Some(libraries.map(_.seq).max)
+      (nextSeqNum, keepsNeedingUpdate, libraries.map(lib => (lib.id.get, lib)).toMap)
+    }
+
+    keepsNeedingUpdate.foreach { keep =>
+      // library id cannot be None here since we got the keep by looking it up by its library id.
```
leogrim: Actually keep.libraryId should no longer be optional at this point, no need to worry about this.

kifi-backend/modules/shoebox/app/com/keepit/model/KeepRepo.scala
```diff
@@ -19,6 +19,7 @@ trait KeepRepo extends Repo[Keep] with ExternalIdColumnFunction[Keep] with SeqNu
   def getByUserAndUriIds(userId: Id[User], uriIds: Set[Id[NormalizedURI]])(implicit session: RSession): Seq[Keep]
   def getByUserIdAndLibraryId(userId: Id[User], libraryId: Id[Library], excludeSet: Set[State[Keep]] = Set(KeepStates.INACTIVE, KeepStates.DUPLICATE))(implicit session: RSession): Seq[Keep]
   def getByLibraryIdsAndUriIds(libraryIds: Set[Id[Library]], uriIds: Set[Id[NormalizedURI]])(implicit session: RSession): Seq[Keep]
+  def getByLibraryIds(libraryIds: Set[Id[Library]])(implicit session: RSession): Seq[Keep]
```
leogrim, line 22: This could be a lot of keeps, I wouldn't encourage this.

Colin-Lane, line 22: I could put a `Limit` and `Offset` and iterate through the results instead.

kifi-backend/modules/shoebox/app/com/keepit/integrity/LibraryChecker.scala
```diff
@@ -29,12 +30,40 @@ class LibraryChecker @Inject() (val airbrake: AirbrakeNotifier,
...
+    syncLibraryMemberCounts() // reads libraryMembership, updates libraries
+    syncOnLibraryMove() // reads libraries, updates keeps. (This is kind of recursive, but not really as I only update the ones that are wrong.)
+  }
+
+  private[integrity] def syncOnLibraryMove(): Unit = {
+    val timer = new NamedStatsdTimer("LibraryChecker.syncOnLibraryMove")
+    val (nextSeqNum, keepsNeedingUpdate, libraries) = db.readOnlyReplica { implicit s =>
+      val lastSeq = getLastSeqNum(LIBRARY_MOVED_NAME)
+      val libraries = libraryRepo.getBySequenceNumber(lastSeq, LIBRARY_FETCH_SIZE)
+      val keepsNeedingUpdate = keepRepo.getByLibraryIds((libraries.map(_.id.get).toSet))
```
leogrim: It's not necessary to load all these keeps at once, that could be a lot and we just need to stream through them. So we should go through libraries one-by-one, and load and fix their keeps. `LibraryCommander` uses `keepRepo.getByLibrary(libraryId, 0, Int.MaxValue, Set.empty)` to load all the keeps from a single library. Now that could also be a lot of keeps (some libraries have 60k+ keeps), so ideally we're paginating through keeps as well.

leogrim: And actually, `keepRepo.getLibrary` could be updated to use your types `Offset` and `Limit` :)

Colin-Lane: thats the exact comment I just left above.

leogrim:
See my comments on specifics, we should be streaming through data when possible. 

@yingjieMiao @Colin-Lane Another thing regarding the plugin structure: we're now doing 5 or 6 checks that all go through the same libraries, keeping track of their own sequence number, and several of them go through all associated keeps. That's a lot of duplicated work (in terms of loading the same data) and potentially multiple updates of the same model within a few seconds (more write load and it gets noisy as these updates get broadcasted to other services like Cortex or Search). 

Should we instead have this plugin goes through all modified libraries once (e.g. we keep track of a single sequence number in `SystemValueRepo`) and have the various checks happen all at once for each? 

That is, instead of having top level checks, we have one `checkIntegrity(library: Library)` method that encapsulates them for that specific library. That same approach would work for other tables, e.g. listen on `Keep.seq` and call `checkIntegrity(keep: Keep)` on each modified keep. Makes sense?

kifi-backend/modules/shoebox/app/com/keepit/integrity/LibraryChecker.scala
leogrim: Yep, just pointing out that the method already exists, but doesn't have your cool types :)

Colin-Lane:
I understand the approach of reducing duplicated effort and combining writes. I think it is a great idea. However, `syncMemberCounts` reads `memberships` by their sequence number, `syncLibraryLastKeptAndKeepCount` reads `keeps` by their sequence number, and `syncOnLibraryMove` reads `libraries` by their sequence number. Then they modify the library or keep.
Each of these has a different starting condition, so It may not be that straight forward to merge them. I'll look into it a bit more.

I do like the approach of streaming the results in a few at a time though.

leogrim:
Great, then we could have three plugins that listen on library, keeps, and library_memberships. @yingjieMiao is taking a look at this, it's a pattern that goes beyond this PR. 
Here, let's paginate through each library's keeps instead of loading everything before we merge this.

Colin-Lane:
Agreed, I'll paginate then merge.

kifi-backend/modules/shoebox/app/com/keepit/integrity/LibraryChecker.scala
```diff
@@ -29,12 +29,68 @@ class LibraryChecker @Inject() (val airbrake: AirbrakeNotifier,
...
-    syncLibraryLastKeptAndKeepCount()
-    syncLibraryMemberCounts()
+    syncLibraryLastKeptAndKeepCount() // reads keeps, updates libraries
+    syncLibraryMemberCounts() // reads libraryMembership, updates libraries
+    syncOnLibraryMove() // reads libraries, updates keeps. (This is kind of recursive, but not really as I only update the ones that are wrong.)
+  }
+
+  // terminates early when it runs out of elements.
+  @tailrec
+  final def stream[A](f: (Limit, Offset) => Iterable[A], limit: Limit, pageSize: PageSize, currentOffset: Offset = Offset(0))(block: A => Unit): Unit = {
```
yingjieMiao: a few comments on this:
1) a few explanations on the arguments could be helpful. e.g., I initially thought `limit` is the max number of results to be retrieved starting from the `currentOffset`. But Looks like it really means max number of results starting from offset 0? My impression is based on `Math.max(limit.value - currentOffset.value, 0)`, this is 0 if `limit == currentOffset`. As an API reader/user, it's a little surprising for me, or at least deserves some clarification.

2) if you think this pattern is reusable, you could consider add this to some util class, and have a more comprehensive test for it.  @andrewconner @leogrim    thoughts?

3) imperative style perhaps gives much cleaner code. Also, you have something like `var nextSeqNum: SequenceNumber[Library] = lastSeq` below. Not a pure function anyway.

4) in what follows, `f` involves open up a new db session. In general, we would prefer

```
db.readWrite{  // loop }
```

over 

```
loop { db.readwrite{} }
```

For a library with over 10K keeps (which we have a lot), the method here will open up 10K / 25 sessions,  -- just for one library.  @andrewconner @leogrim

Colin-Lane: I wasn't able to find a simple alternative to finding `nextSeqNum` without using a `var` but would love to hear one. I think I added a comment in there about that. Since we are streaming results, we aren't expecting a return type (that would ruin the whole point of keeping a low memory footprint), so I ended up using a side effect.

I could use an `Iterator` that calculates items on demand instead of a `tail-recursive method` and leave the use case open to a larger set of uses.

kifi-backend/modules/shoebox/app/com/keepit/integrity/LibraryChecker.scala
```diff
@@ -29,12 +29,68 @@ class LibraryChecker @Inject() (val airbrake: AirbrakeNotifier,
...
+    def getLibraries(limit: Limit, offset: Offset): Iterable[Library] = db.readOnlyReplica { implicit s =>
+      libraryRepo.getBySequenceNumber(lastSeq + offset.value, limit.value.toInt)
+    }
+
+    // wasn't sure of a better way to do this, since we want to get the libraries lazily, we can't really return anything or we end up keeping the memory around.
+    var nextSeqNum: SequenceNumber[Library] = lastSeq
+
+    // iterate over LIBRARY_FETCH_SIZE libraries by 25library pages.
+    stream(getLibraries, limit = Limit(LIBRARY_FETCH_SIZE), pageSize = PageSize(25)) { library =>
+      def getKeeps(limit: Limit, offset: Offset): Iterable[Keep] = db.readOnlyReplica { implicit s =>
```
yingjieMiao, line 65: the semantic of `offset` and `limit` here are consistent with our db paging convention (e.g. get `limit` records starting from `offset`. If I understood your code correctly, the arguments in `stream` have different semantic. That caused me some trouble.

Colin-Lane, line 65: Yeah, I noticed that as well, I'll have to revise it to be more clear. Maybe instead of using `Limit` I could create a new case class to show that it is different. Maybe `Count` or `TotalItems`, or even use `Pages` and just go off of that?

kifi-backend/modules/shoebox/app/com/keepit/integrity/LibraryChecker.scala
leogrim: In any case, you don't need `PageSize` when you already have `Offset` and `Limit`. Let's move towards a consistent pattern everywhere, that is `Offset` and `Limit`

yingjieMiao:
I think the paging logic can be greatly simplified if we take an imperative approach here. Just create an `ArrayBuffer` and do stuff from there. I can see you put lots of efforts on pure functional method, but `var nextSeqNum: SequenceNumber[Library] = lastSeq` is a violation of your intent.

In addition, as mentioned, current code opens up too many db sessions.

kifi-backend/modules/shoebox/app/com/keepit/integrity/LibraryChecker.scala
leogrim, line 65: I would stick with `Offset` and `Limit` rather than introducing different pagination patterns. In my experience, `Offset` and `Limit` are actually easier to reason about with streams (as they match the arguments for `.drop` and `.take`).

kifi-backend/modules/shoebox/test/com/keepit/integrity/LibraryCheckerTest.scala
```diff
@@ -98,5 +101,22 @@ class LibraryCheckerTest extends TestKitSupport with SpecificationLike with Shoe
...
+        val libraryChecker = inject[LibraryChecker]
+        val buffer: ArrayBuffer[Long] = new ArrayBuffer[Long]()
+        val iterator = (limit: Limit, offset: Offset) => {
+          val values = offset.value until (offset.value + limit.value)
+          buffer ++= values
+          values
+        }
+        libraryChecker.stream(iterator, needed = Limit(1000), Limit(10)) { x =>
+          // buffer only increases when we get a new page, because we are only adding to it on new page in our iterator function.
+          buffer.size === ((x + 10) - (x % 10))
```
yingjieMiao: interesting. this is just `10(1 + x/10)`

kifi-backend/modules/shoebox/test/com/keepit/integrity/LibraryCheckerTest.scala
```diff
@@ -99,6 +102,23 @@ class LibraryCheckerTest extends TestKitSupport with SpecificationLike with Shoe
...
+        val libraryChecker = inject[LibraryChecker]
+        val buffer: ArrayBuffer[Long] = new ArrayBuffer[Long]()
+        val iterator = (limit: Limit, offset: Offset) => {
+          val values = offset.value until (offset.value + limit.value)
+          buffer ++= values
+          values
+        }
+        libraryChecker.stream(iterator, needed = Limit(1000), Limit(10)) { x =>
+          // buffer only increases when we get a new page, because we are only adding to it on new page in our iterator function.
+          buffer.size === 10 * (1 + x / 10)
```
yingjieMiao, line 116: ah, didn't expect you change that one, unless you find it's simpler...

kifi-backend/modules/shoebox/marketing
```diff
@@ -1 +1 @@
-Subproject commit ed865de3c54ac14255b5ce8411ade7153bcaa344
+Subproject commit 3a5e5c28af64e6adc0626eeffbdc4bc411a58050
```
yingjieMiao, line 1: for this type of thing, you can go to the directory `/keepit/kifi-backend/modules/shoebox/marketing`
and 
`git checkout ed865de3c54ac14255b5ce8411ade7153bcaa344`

should bring it back.

kifi-backend/modules/shoebox/test/com/keepit/integrity/LibraryCheckerTest.scala
Colin-Lane, line 116: They're both pretty much the same.

yingjieMiao:
lgtm. Thanks for the changes and tests!

kifi-backend/modules/shoebox/marketing
andrewconner, line 1: This should never happen. Please be careful, this reverted the marketing site.

PR #15828, opened 2015-06-17
debug IllegalArgumentException

yingjieMiao:
@leogrim @andrewconner


andrewconner:
So weird! Looks good.

PR #15830, opened 2015-06-17
Adding the tag count as "numTags" on the profile stats

jaredpetker:
@andrewconner @eishay Unsure the best way to actually test this, it compiles, tests pass, but I don't think there are actual profile tests, I can look into adding them if desired.


andrewconner:
So sorry for leaving this standing, I just noticed it was open and tagged me. Looks good, merging so you can use it.

jaredpetker:
Cool beans, no worries.

andrewconner:
Tests failed :). Hah, will fix and tag you. @jaredpetker

jaredpetker:
Interesting, nothing failed at the time I thought, I can pull and run the
tests and see what happens.

PR #15832, opened 2015-06-17
Quick keep into it's on directive + slightly better place on the page

stkem:
@andrewconner


andrewconner:
Perfect. Reviewed in person. Ship it!

PR #15834, opened 2015-06-17
Org member validation

ryanpbrewster:
@leogrim @Colin-Lane 

Relatively clean validation logic and some tests to go along with them.

Take a look specifically at the tests. If that is not the behavior we want, we should fix it
sooner rather than later. To me it seems strange to have multiple owners, but
it is all internally consistent (and we have org.ownerId to identify the "true" original owner, if
for some reason we need that).

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
@@ -97,31 +97,32 @@ class OrganizationMembershipCommanderImpl @Inject() (
     val requesterOpt = organizationMembershipRepo.getByOrgIdAndUserId(request.orgId, request.requesterId)
     val targetOpt = organizationMembershipRepo.getByOrgIdAndUserId(request.orgId, request.targetId)
 
-    if (requesterOpt.isEmpty) return false // a non-member can't request anything
-    val requester = requesterOpt.get
-
-    request match {
-      case OrganizationMembershipAddRequest(_, _, _, newRole) => validAdd(requester, targetOpt, newRole)
-      case OrganizationMembershipModifyRequest(_, _, _, newRole) => validModify(requester, targetOpt, newRole)
-      case OrganizationMembershipRemoveRequest(_, _, _) => validRemove(requester, targetOpt)
+    requesterOpt match {
+      case None => false // a non-member can't request anything
+      case Some(requester) =>
+        request match {
+          case OrganizationMembershipAddRequest(_, _, _, newRole) => validAdd(requester, targetOpt, newRole)
+          case OrganizationMembershipModifyRequest(_, _, _, newRole) => validModify(requester, targetOpt, newRole)
+          case OrganizationMembershipRemoveRequest(_, _, _) => validRemove(requester, targetOpt)
+        }
     }
   }
   private def validAdd(requester: OrganizationMembership, targetOpt: Option[OrganizationMembership], newRole: OrganizationRole): Boolean = {
     targetOpt match {
-      case None => newRole <= requester.role
+      case None => requester.role >= newRole
       case Some(_) => false
     }
   }
   private def validModify(requester: OrganizationMembership, targetOpt: Option[OrganizationMembership], newRole: OrganizationRole): Boolean = {
     targetOpt match {
       case None => false
-      case Some(target) => target.role <= requester.role && newRole <= requester.role
+      case Some(target) => (requester.role >= target.role) && (newRole <= requester.role)
     }
   }
   private def validRemove(requester: OrganizationMembership, targetOpt: Option[OrganizationMembership]): Boolean = {
     targetOpt match {
       case None => false
-      case Some(target) => target.role <= requester.role
+      case Some(target) => requester.role >= target.role
     }
   }
 
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
@@ -94,7 +94,35 @@ class OrganizationMembershipCommanderImpl @Inject() (
   }
 
   private def validRequest(request: OrganizationMembershipRequest)(implicit session: RSession): Boolean = {
-    true
+    val requesterOpt = organizationMembershipRepo.getByOrgIdAndUserId(request.orgId, request.requesterId)
+    val targetOpt = organizationMembershipRepo.getByOrgIdAndUserId(request.orgId, request.targetId)
+
+    if (requesterOpt.isEmpty) return false // a non-member can't request anything
+    val requester = requesterOpt.get
+
+    request match {
+      case OrganizationMembershipAddRequest(_, _, _, newRole) => validAdd(requester, targetOpt, newRole)
+      case OrganizationMembershipModifyRequest(_, _, _, newRole) => validModify(requester, targetOpt, newRole)
+      case OrganizationMembershipRemoveRequest(_, _, _) => validRemove(requester, targetOpt)
+    }
+  }
+  private def validAdd(requester: OrganizationMembership, targetOpt: Option[OrganizationMembership], newRole: OrganizationRole): Boolean = {
+    targetOpt match {
+      case None => newRole <= requester.role
+      case Some(_) => false
+    }
+  }
+  private def validModify(requester: OrganizationMembership, targetOpt: Option[OrganizationMembership], newRole: OrganizationRole): Boolean = {
+    targetOpt match {
+      case None => false
+      case Some(target) => target.role <= requester.role && newRole <= requester.role
+    }
+  }
+  private def validRemove(requester: OrganizationMembership, targetOpt: Option[OrganizationMembership]): Boolean = {
+    targetOpt match {
+      case None => false
+      case Some(target) => target.role <= requester.role
+    }
   }
 
   def addMembership(request: OrganizationMembershipAddRequest): Either[OrganizationFail, OrganizationMembershipAddResponse] = {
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
```diff
@@ -94,7 +94,35 @@ class OrganizationMembershipCommanderImpl @Inject() (
   }
 
   private def validRequest(request: OrganizationMembershipRequest)(implicit session: RSession): Boolean = {
-    true
+    val requesterOpt = organizationMembershipRepo.getByOrgIdAndUserId(request.orgId, request.requesterId)
+    val targetOpt = organizationMembershipRepo.getByOrgIdAndUserId(request.orgId, request.targetId)
+
+    if (requesterOpt.isEmpty) return false // a non-member can't request anything
```
ryanpbrewster: Ah, just saw your comment on Colin's PR. I'll fix this.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
```diff
@@ -94,7 +94,36 @@ class OrganizationMembershipCommanderImpl @Inject() (
   }
 
   private def validRequest(request: OrganizationMembershipRequest)(implicit session: RSession): Boolean = {
-    true
+    val requesterOpt = organizationMembershipRepo.getByOrgIdAndUserId(request.orgId, request.requesterId)
+    val targetOpt = organizationMembershipRepo.getByOrgIdAndUserId(request.orgId, request.targetId)
+
+    if (requesterOpt.isEmpty) false // a non-member can't request anything
+    else {
```
Colin-Lane: consider using `requestorOpt.nonEmpty` instead to show the success case first.
Or even matching on the requestOpt.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
```diff
@@ -94,7 +94,36 @@ class OrganizationMembershipCommanderImpl @Inject() (
...
+      request match {
+        case OrganizationMembershipAddRequest(_, _, _, newRole) => validAdd(requester, targetOpt, newRole)
+        case OrganizationMembershipModifyRequest(_, _, _, newRole) => validModify(requester, targetOpt, newRole)
+        case OrganizationMembershipRemoveRequest(_, _, _) => validRemove(requester, targetOpt)
+      }
+    }
+  }
+  private def validAdd(requester: OrganizationMembership, targetOpt: Option[OrganizationMembership], newRole: OrganizationRole): Boolean = {
+    targetOpt match {
+      case None => newRole <= requester.role
```
Colin-Lane: this looks a bit odd, you could use parenthesis `(newRole <= requester.role)`

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
```diff
@@ -94,7 +94,36 @@ class OrganizationMembershipCommanderImpl @Inject() (
...
+      case Some(_) => false
+    }
+  }
+  private def validModify(requester: OrganizationMembership, targetOpt: Option[OrganizationMembership], newRole: OrganizationRole): Boolean = {
+    targetOpt match {
+      case None => false
+      case Some(target) => target.role <= requester.role && newRole <= requester.role
```
Colin-Lane: complex boolean expressions are much easier to read with parenthesis wrapping each of the conditionals

kifi-backend/modules/shoebox/test/com/keepit/commanders/OrganizationMembershipCommanderTest.scala
```diff
@@ -52,5 +53,90 @@ class OrganizationMembershipCommanderTest extends TestKitSupport with Specificat
...
+        val memberAddUser = OrganizationMembershipAddRequest(orgId, Id[User](2), Id[User](3), OrganizationRole.MEMBER)
+        orgMemberCommander.addMembership(memberAddUser) === Right(OrganizationMembershipAddResponse(memberAddUser))
+
+        val memberAddUserAsOwner = OrganizationMembershipAddRequest(orgId, Id[User](2), Id[User](5), OrganizationRole.OWNER)
+        orgMemberCommander.addMembership(memberAddUserAsOwner) === Left(OrganizationFail.INSUFFICIENT_PERMISSIONS)
+
+        val noneAddUser = OrganizationMembershipAddRequest(orgId, Id[User](42), Id[User](4), OrganizationRole.MEMBER)
+        orgMemberCommander.addMembership(noneAddUser) === Left(OrganizationFail.INSUFFICIENT_PERMISSIONS)
+
+        val memberAddMember = OrganizationMembershipAddRequest(orgId, Id[User](3), Id[User](1), OrganizationRole.MEMBER)
```
Colin-Lane, line 81: I think `Id[User](1)` here is supposed to be `ownerId`

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
ryanpbrewster: I really prefer the unsuccessful-case-first style. Maybe a match is better, though.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
ryanpbrewster: :smile:

kifi-backend/modules/shoebox/test/com/keepit/commanders/OrganizationMembershipCommanderTest.scala
```diff
@@ -52,5 +53,90 @@ class OrganizationMembershipCommanderTest extends TestKitSupport with Specificat
...
+          orgMemberRepo.save(OrganizationMembership(organizationId = orgId, userId = Id[User](1), role = OrganizationRole.OWNER))
+          orgMemberRepo.save(OrganizationMembership(organizationId = orgId, userId = Id[User](2), role = OrganizationRole.MEMBER))
+          orgMemberRepo.save(OrganizationMembership(organizationId = orgId, userId = Id[User](3), role = OrganizationRole.MEMBER))
+          orgMemberRepo.save(OrganizationMembership(organizationId = orgId, userId = Id[User](4), role = OrganizationRole.MEMBER))
+        }
+
+        val orgMemberCommander = inject[OrganizationMembershipCommander]
+
+        val ownerDelMember = OrganizationMembershipRemoveRequest(orgId, Id[User](1), Id[User](2))
+        orgMemberCommander.removeMembership(ownerDelMember) === Right(OrganizationMembershipRemoveResponse(ownerDelMember))
```
Colin-Lane, line 130: Since the request is unmodified, is there a reason to wrap it in a response, or even return it at all?
If only there was an `Either[E, Nothing]` I think no response on the right side would work just fine here.

ryanpbrewster, line 130: By wrapping it we can abuse the typesystem to guarantee certain things. And we could definitely do an Option[OrganizationFail], but I feel like there are use cases where we may want to know what request just succeeded, like telling a user what they just did.

ryanpbrewster, line 130: And it's possible that at some point in the future we may want additional fields in the response. I can't think of any ATM though.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
```diff
@@ -94,7 +94,36 @@ class OrganizationMembershipCommanderImpl @Inject() (
...
+    targetOpt match {
+      case None => false
+      case Some(target) => (requester.role >= target.role) && (newRole <= requester.role)
+    }
+  }
+  private def validRemove(requester: OrganizationMembership, targetOpt: Option[OrganizationMembership]): Boolean = {
+    targetOpt match {
+      case None => false
+      case Some(target) => requester.role >= target.role
+    }
```
leogrim, line 111: When returning `Boolean`, it's nice to have functions named like a question, e.g.  `isValidMembershipRequest`, `isValidAddRequest`, etc...

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
```diff
@@ -94,7 +94,36 @@ class OrganizationMembershipCommanderImpl @Inject() (
...
   }
```
leogrim, line 112: The common check `requester.role >= target.role` could be factored out of the smaller functions.
Also it seems like you could spare some `match` statements with `Option.exists`. At these point, additional functions are so short than we may just inline their code in `validRequest`

kifi-backend/modules/shoebox/test/com/keepit/commanders/OrganizationMembershipCommanderTest.scala
leogrim, line 130: I like this typed `Request / Response` pattern. Very explicit and flexible, while making top-level commander method signatures shorter.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
leogrim, line 111: With examples:
- for the common check: `!targetOpt.exists(_.role > requester.role)`
- for modify: `targetOpt.exists(_.role <= requester.role) && newRole <= requester.role`

ryanpbrewster, line 111: I like the exists() change, and it makes the checks short enough that I'm happy inlining them.

However, I prefer to keep the add/modify/remove checks separate, since they're very short and I feel like it's easier to understand each individually.

PR #15837, opened 2015-06-17
endpoint to fix keep visibility

yingjieMiao:
@leogrim 

the endpoint is likely to be temporary. Actual method could be reused later.


kifi-backend/modules/shoebox/app/com/keepit/integrity/LibraryChecker.scala
```diff
@@ -144,4 +144,20 @@ class LibraryChecker @Inject() (val airbrake: AirbrakeNotifier,
...
+      val LIMIT = 500
+      while (!done) {
+        val keepsToFix = keepRepo.getByLibraryIdAndExcludingVisibility(libId, excludeVisibility = Some(lib.visibility), limit = LIMIT)
+        total += keepsToFix.size
+        keepsToFix.foreach { keep => keepRepo.save(keep.copy(visibility = lib.visibility)) }
+        if (keepsToFix.size < LIMIT) done = true
+      }
+      total
+    }
+  }
```
leogrim, line 162: @yingjieMiao @Colin-Lane can we check `visibility` and `organizationId` in one go?

PR #15850, opened 2015-06-18
[DO NOT MERGE] Orgs with default permissions

ryanpbrewster:
@leogrim @Colin-Lane @andrewconner 

Some of this code is explicitly wrong and I need to figure out how to make it work.

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/model/Organization.scala
+++ b/kifi-backend/modules/common/app/com/keepit/model/Organization.scala
@@ -3,7 +3,7 @@ package com.keepit.model
 import java.net.URLEncoder
 import javax.crypto.spec.IvParameterSpec
 
-import com.keepit.common.crypto.{ ModelWithPublicIdCompanion, ModelWithPublicId }
+import com.keepit.common.crypto.{ ModelWithPublicId, ModelWithPublicIdCompanion }
 import com.keepit.common.db._
 import com.keepit.common.strings._
 import com.keepit.common.time._
@@ -21,11 +21,20 @@ case class Organization(
     name: String,
     description: Option[String] = None,
     ownerId: Id[User],
-    handle: Option[PrimaryOrganizationHandle]) extends ModelWithPublicId[Organization] with ModelWithState[Organization] with ModelWithSeqNumber[Organization] {
+    handle: Option[PrimaryOrganizationHandle],
+    basePermissions: BasePermissions = Organization.defaultBasePermissions) extends ModelWithPublicId[Organization] with ModelWithState[Organization] with ModelWithSeqNumber[Organization] {
 
   override def withId(id: Id[Organization]): Organization = this.copy(id = Some(id))
-
   override def withUpdateTime(now: DateTime): Organization = this.copy(updatedAt = now)
+
+  def nonmemberPermissions = basePermissions.forNonmember
+  def rolePermissions(role: OrganizationRole) = basePermissions.forRole(role)
+
+  def newMembership(userId: Id[User], role: OrganizationRole): OrganizationMembership =
+    OrganizationMembership(organizationId = id.get, userId = userId, role = role, permissions = rolePermissions(role))
+
+  def modifiedMembership(membership: OrganizationMembership, newRole: OrganizationRole): OrganizationMembership =
+    membership.copy(role = newRole, permissions = rolePermissions(newRole))
 }
 
 object Organization extends ModelWithPublicIdCompanion[Organization] {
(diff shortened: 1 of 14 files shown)


kifi-backend/modules/common/app/com/keepit/model/Organization.scala
```diff
@@ -87,3 +114,33 @@ case class OrganizationHandle(value: String) extends AnyVal {
...
+      /*
+      json.validate[JsObject].map { obj =>
+        for( (k,v) <- obj.value ) yield {
+          val roleOpt = if (k == "none") None else Some(OrganizationRole(k))
+          val permissions = v.as[Set[OrganizationPermission]]
+          roleOpt -> permissions
+        }
+      }
+      */
+    }
```
ryanpbrewster, line 137: @leogrim I couldn't get this part to compile. It gave me a pretty indecipherable error.

yingjieMiao, line 137: don't you want to wrap the map with `BasePermissions`?

kifi-backend/modules/common/app/com/keepit/model/OrganizationMembership.scala
```diff
@@ -20,8 +20,6 @@ case class OrganizationMembership(
   def withId(id: Id[OrganizationMembership]): OrganizationMembership = this.copy(id = Some(id))
   def withUpdateTime(now: DateTime): OrganizationMembership = this.copy(updatedAt = now)
   def withState(newState: State[OrganizationMembership]): OrganizationMembership = this.copy(state = newState)
-  def withRole(newRole: OrganizationRole): OrganizationMembership = this.copy(role = newRole, permissions = OrganizationPermission.defaultPermissions(newRole))
-  def withPermissions(newPermissions: Set[OrganizationPermission]): OrganizationMembership = this.copy(permissions = newPermissions)
```
ryanpbrewster, line 24: I removed these because depending on the organization they would need to do different things.

kifi-backend/modules/common/app/com/keepit/model/Organization.scala
```diff
@@ -21,11 +23,15 @@ case class Organization(
...
+    handle: Option[PrimaryOrganizationHandle],
+    basePermissions: BasePermissions) extends ModelWithPublicId[Organization] with ModelWithState[Organization] with ModelWithSeqNumber[Organization] {
 
   override def withId(id: Id[Organization]): Organization = this.copy(id = Some(id))
 
   override def withUpdateTime(now: DateTime): Organization = this.copy(updatedAt = now)
+
+  def nonmemberPermissions = basePermissions.forNonmember
+  def rolePermissions(role: OrganizationRole) = basePermissions.forRole(role)
 }
```
leogrim, line 38: `getBasePermissions(role: OrganizationRole)` and `getGuestPermissions`?

kifi-backend/modules/common/app/com/keepit/model/Organization.scala
```diff
@@ -87,3 +114,31 @@ case class OrganizationHandle(value: String) extends AnyVal {
 
 @json
 case class PrimaryOrganizationHandle(original: OrganizationHandle, normalized: OrganizationHandle)
+
+case class BasePermissions(permissionsMap: Map[Option[OrganizationRole], Set[OrganizationPermission]]) {
+  def forRole(role: OrganizationRole): Set[OrganizationPermission] = permissionsMap(Some(role))
+  def forNonmember: Set[OrganizationPermission] = permissionsMap(None)
```
leogrim, line 123: Symmetrically `for(role: OrganizationRole)` and `forGuest`?

kifi-backend/modules/common/app/com/keepit/model/Organization.scala
```diff
@@ -87,3 +114,31 @@ case class OrganizationHandle(value: String) extends AnyVal {
...
+object BasePermissions {
+  implicit val format: Format[BasePermissions] = new Format[BasePermissions] {
+    def reads(json: JsValue): JsResult[BasePermissions] = {
+      json.validate[JsObject].map { obj =>
+        val permissionsMap = for ((k, v) <- obj.value) yield {
+          val roleOpt = if (k == "none") None else Some(OrganizationRole(k))
+          val permissions = v.as[Set[OrganizationPermission]]
+          roleOpt -> permissions
+        }
+        BasePermissions(permissionsMap.toSeq.toMap)
```
Colin-Lane: Interesting. `permissionsMap.toSeq.toMap` converting from a `Map` back to a `Map`?

kifi-backend/modules/common/app/com/keepit/model/Organization.scala
ryanpbrewster, line 38: I'm concerned about "Guest" because that could conceivably be an OrganizationRole in the future. I like getBasePermissions though.

kifi-backend/modules/common/app/com/keepit/model/Organization.scala
ryanpbrewster: Yeah, I was not happy about this. permissionsMap is an immutable map, and BasePermissions requires a mutable map. Taking it to a sequence and then back reconstructs it as a mutable map. If you can find a better way I'd be thrilled.

Colin-Lane: Is there a reason BasePermissions needs to be mutable? Usually when we mutate, we just immutably mutate, and then copy the new map in and save to the db.

```
val newPermissions = org.permissions += "extreme hiker" -> Seq("view library")
db.save(org.copy(permissions = newPermissions)
```

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
```diff
@@ -37,15 +38,28 @@ trait OrganizationMembershipCommander {
...
+        val memberPermissionsOpt = organizationMembershipRepo.getByOrgIdAndUserId(orgId, userId).map(_.permissions)
+        memberPermissionsOpt match {
+          case Some(permissions) => permissions
+          case None =>
+            val inviteRoleOpt = organizationInviteRepo.getByOrgIdAndUserId(orgId, userId).map(_.role).maxOpt
+            inviteRoleOpt match {
+              case Some(role) => org.rolePermissions(role)
+              case None => org.nonmemberPermissions
+            }
+        }
```
Colin-Lane, line 61: I have a feeling you may be able to leverage `getOrElse` on `Option` here, or extract some of these matches out to a `def`

kifi-backend/modules/common/app/com/keepit/model/Organization.scala
ryanpbrewster: No good reason. I just didn't want to type out `scala.collection.immutable.Map` every time. ((Is there a way around that?))

Colin-Lane: `scala.collection.immutable.Map` comes in `Predef` as `Map`

PR #15851, opened 2015-06-18
Let people provide a company name when signing up with email/password

stkem:
+notify sales@kifi.com when that happens and set appropriate user values.
@andrewconner @eishay 
(Can demo in dev mode if you want to see it)

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/AuthCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/AuthCommander.scala
@@ -82,7 +82,8 @@ case class EmailPassFinalizeInfo(
   picHeight: Option[Int],
   cropX: Option[Int],
   cropY: Option[Int],
-  cropSize: Option[Int])
+  cropSize: Option[Int],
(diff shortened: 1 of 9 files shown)


andrewconner:
Eishay also wants their description to be "Works at <company>"

andrewconner:
Looks good, let me know if you need help testing it in prod.

kifi-backend/modules/shoebox/app/com/keepit/commanders/AuthCommander.scala
```diff
@@ -233,6 +236,30 @@ class AuthCommander @Inject() (
       (user, emailPassIdentity)
     }
 
+  def processCompanyNameFromSignup(user: User, companyNameOpt: Option[String]): Unit = {
+    companyNameOpt.map { companyName =>
+      val trimmed = companyName.trim
+      if (trimmed != "") {
+        userCommander.updateUserBiography(user.id.get, s"Works at $trimmed");
```
stkem, line 243: @andrewconner: Description is here.

PR #15853, opened 2015-06-18
Prototype hack involving default pictures for users with no picture

andrewconner:
@stkem @leogrim 

Getting user picture closer to a state where we can use the existing `ProcessedImages` pipeline. A roadblock is `pictureName` being a bad API. We'd like to move to using real, absolute URLs for pictures. This is testing that we don't break clients if I put in a backwards compatibility hack.

Shouldn't affect anyone but me, Eishay, and three test users.

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/social/BasicUser.scala
+++ b/kifi-backend/modules/common/app/com/keepit/social/BasicUser.scala
@@ -9,6 +9,7 @@ import com.keepit.model._
 import play.api.libs.functional.syntax._
 import play.api.libs.json._
 import scala.concurrent.duration.Duration
+import play.api.Play
 
 trait BasicUserLikeEntity {
   def asBasicUser: Option[BasicUser] = None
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/common/app/com/keepit/social/BasicUser.scala
```diff
@@ -72,19 +73,44 @@ object BasicUser {
...
+  }
+
   def fromUser(user: User): BasicUser = {
     BasicUser(
       externalId = user.externalId,
       firstName = user.firstName,
       lastName = user.lastName,
-      pictureName = user.pictureName.getOrElse(S3UserPictureConfig.defaultName) + ".jpg",
+      pictureName = userToPictureName(user),
       username = user.username
```
leogrim, line 107: Do you plan on introducing new fields and deprecate this one?

andrewconner, line 107: Introducing new fields

leogrim, line 107: :+1:

stkem:
looks save to me

PR #15854, opened 2015-06-18
subscription UI updates

camhashemi:
@andrewconner I'm having trouble getting the manage-lib-input-error class to take precedence when an error is found for that input. Pretty sure that the scoped $error JS variables are right, per looking at the debugger. I would really appreciate your help.


camhashemi:
@andrewconner I need to deploy so that Sterling can evaluate the UX/UI this morning.

General error messages e.g. "Please enter a valid webhook URL for each subscription" are being used. Highlighting the specific fields is my remaining problem.

PR #15856, opened 2015-06-18
Organizations now have base permissions

ryanpbrewster:
Each organization can configure what permissions each role is entitled to.
This includes permissions for non-members, which should allow us to implement
private organizations.

The base permissions are serialized to JSON and stored in the `organization`
database table.

Changing an OrganizationMembership is now more involved, since you need the
organization itself to tell you what the new role's permissions are. For this
reason we always construct OrganizationMembership via org.newMembership
and modify them via org.modifyMembership.

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/model/Organization.scala
+++ b/kifi-backend/modules/common/app/com/keepit/model/Organization.scala
@@ -3,7 +3,7 @@ package com.keepit.model
 import java.net.URLEncoder
 import javax.crypto.spec.IvParameterSpec
 
-import com.keepit.common.crypto.{ ModelWithPublicIdCompanion, ModelWithPublicId }
+import com.keepit.common.crypto.{ ModelWithPublicId, ModelWithPublicIdCompanion }
 import com.keepit.common.db._
 import com.keepit.common.strings._
 import com.keepit.common.time._
@@ -21,11 +21,20 @@ case class Organization(
     name: String,
     description: Option[String] = None,
     ownerId: Id[User],
-    handle: Option[PrimaryOrganizationHandle]) extends ModelWithPublicId[Organization] with ModelWithState[Organization] with ModelWithSeqNumber[Organization] {
+    handle: Option[PrimaryOrganizationHandle],
+    basePermissions: BasePermissions = Organization.defaultBasePermissions) extends ModelWithPublicId[Organization] with ModelWithState[Organization] with ModelWithSeqNumber[Organization] {
 
   override def withId(id: Id[Organization]): Organization = this.copy(id = Some(id))
-
   override def withUpdateTime(now: DateTime): Organization = this.copy(updatedAt = now)
+
+  def getNonmemberPermissions = basePermissions.forNonmember
+  def getRolePermissions(role: OrganizationRole) = basePermissions.forRole(role)
+
+  def newMembership(userId: Id[User], role: OrganizationRole): OrganizationMembership =
+    OrganizationMembership(organizationId = id.get, userId = userId, role = role, permissions = getRolePermissions(role))
+
+  def modifiedMembership(membership: OrganizationMembership, newRole: OrganizationRole): OrganizationMembership =
+    membership.copy(role = newRole, permissions = getRolePermissions(newRole))
 }
 
 object Organization extends ModelWithPublicIdCompanion[Organization] {
@@ -33,6 +42,24 @@ object Organization extends ModelWithPublicIdCompanion[Organization] {
 
   protected val publicIdPrefix = "o"
   protected val publicIdIvSpec = new IvParameterSpec(Array(62, 91, 74, 34, 82, -77, 19, -35, -118, 3, 112, -59, -70, 94, 101, -115))
+
+  val defaultBasePermissions: BasePermissions =
+    BasePermissions(Map(
+      None -> Set(
+        OrganizationPermission.VIEW_ORGANIZATION
+      ),
+      Some(OrganizationRole.MEMBER) -> Set(
+        OrganizationPermission.VIEW_ORGANIZATION,
+        OrganizationPermission.ADD_LIBRARIES
+      ),
+      Some(OrganizationRole.OWNER) -> Set(
+        OrganizationPermission.VIEW_ORGANIZATION,
+        OrganizationPermission.EDIT_ORGANIZATION,
+        OrganizationPermission.INVITE_MEMBERS,
+        OrganizationPermission.ADD_LIBRARIES,
+        OrganizationPermission.REMOVE_LIBRARIES
+      )
+    ))
 
   implicit val format: Format[Organization] = (
     (__ \ 'id).formatNullable[Id[Organization]] and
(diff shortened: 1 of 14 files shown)


kifi-backend/modules/common/app/com/keepit/model/Organization.scala
```diff
@@ -21,11 +21,20 @@ case class Organization(
...
 
   override def withId(id: Id[Organization]): Organization = this.copy(id = Some(id))
-
   override def withUpdateTime(now: DateTime): Organization = this.copy(updatedAt = now)
+
+  def nonmemberPermissions = basePermissions.forNonmember
+  def rolePermissions(role: OrganizationRole) = basePermissions.forRole(role)
+
+  def newMembership(userId: Id[User], role: OrganizationRole): OrganizationMembership =
+    OrganizationMembership(organizationId = id.get, userId = userId, role = role, permissions = rolePermissions(role))
```
yingjieMiao: our convention of modifying db models are:

```
def withMembership(userId: Id[User], role: OrganizationRole) = this.copy(userId = userId, role = role, permissions = rolePermissions(role))
```

Note the `withXYZ` pattern, and `this.copy()`. Otherwise, the original `createdAt` is lost.

kifi-backend/modules/common/app/com/keepit/model/Organization.scala
```diff
@@ -21,11 +21,20 @@ case class Organization(
...
+
+  def modifiedMembership(membership: OrganizationMembership, newRole: OrganizationRole): OrganizationMembership =
+    membership.copy(role = newRole, permissions = rolePermissions(newRole))
```
yingjieMiao: could this be a method on `OrganizationMembership` model?

kifi-backend/modules/common/app/com/keepit/model/Organization.scala
```diff
@@ -34,6 +43,24 @@ object Organization extends ModelWithPublicIdCompanion[Organization] {
...
+        OrganizationPermission.ADD_LIBRARIES
+      ),
+      Some(OrganizationRole.OWNER) -> Set(
+        OrganizationPermission.VIEW_ORGANIZATION,
+        OrganizationPermission.EDIT_ORGANIZATION,
+        OrganizationPermission.INVITE_MEMBERS,
+        OrganizationPermission.ADD_LIBRARIES,
+        OrganizationPermission.REMOVE_LIBRARIES
+      )
+    ))
```
yingjieMiao, line 62: makes sense to `import OrganizationPermission ._` ?

kifi-backend/modules/common/app/com/keepit/model/Organization.scala
yingjieMiao: i see, you need `rolePermissions`

kifi-backend/modules/common/conf/evolutions/shoebox/344.sql
```diff
@@ -0,0 +1,10 @@
+# SHOEBOX
+
+# --- !Ups
+
+ALTER TABLE organization
+	ADD COLUMN base_permissions varchar(1024) NOT NULL;
```
yingjieMiao: FYI, `text` is safer (much bigger size limit).

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
```diff
@@ -136,8 +152,8 @@ class OrganizationMembershipCommanderImpl @Inject() (
   def removeMembership(request: OrganizationMembershipRemoveRequest): Either[OrganizationFail, OrganizationMembershipRemoveResponse] = {
     db.readWrite { implicit session =>
       if (validRequest(request)) {
-        val oldMembership = organizationMembershipRepo.getByOrgIdAndUserId(request.orgId, request.targetId).get
-        organizationMembershipRepo.deactivate(oldMembership.id.get)
+        val membership = organizationMembershipRepo.getByOrgIdAndUserId(request.orgId, request.targetId).get
+        organizationMembershipRepo.deactivate(membership.id.get)
```
yingjieMiao, line 150: `deactivate` could be a method of the commander, instead of repo.

kifi-backend/modules/common/app/com/keepit/model/Organization.scala
yingjieMiao, line 62: is it possible/helpful to group these permissions? e.g. `org_perm`, `lib_perm`, `member_perm`. These could be separate columns in the table. each column manages a modularized function set. 

The permission set will grow 1) because we want richer settings; 2) as organization has more types of resources to manage, each resource need a permission setting.

As an owner, it probably has the most comprehensive permissions, the size will be something like `Actions x ResourceTypes`. This can become quickly unmanageable in a single column. 

In addition, UI-wise, these permission settings are going to be segmented.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
ryanpbrewster, line 150: Don't we always have repos deactivate? It's an inherited method for the repo.

kifi-backend/modules/common/app/com/keepit/model/Organization.scala
ryanpbrewster: I like that convention, but I had a hard time applying it here. Because the Membership needs to know Org information, I found it easier to move those methods here.

kifi-backend/modules/common/app/com/keepit/model/Organization.scala
ryanpbrewster, line 62: I like this thought. I will see if I can come up with a nice partition.

kifi-backend/modules/common/app/com/keepit/model/Organization.scala
leogrim: You could also have a factory method on `OrganizationMembership` that takes an `Organization`.

kifi-backend/modules/common/app/com/keepit/model/Organization.scala
leogrim, line 62: It's nice to be able to just check a set membership, be careful not to over-engineer this.

kifi-backend/modules/common/app/com/keepit/model/Organization.scala
```diff
@@ -21,11 +21,20 @@ case class Organization(
...
-    handle: Option[PrimaryOrganizationHandle]) extends ModelWithPublicId[Organization] with ModelWithState[Organization] with ModelWithSeqNumber[Organization] {
+    handle: Option[PrimaryOrganizationHandle],
+    basePermissions: BasePermissions = Organization.defaultBasePermissions) extends ModelWithPublicId[Organization] with ModelWithState[Organization] with ModelWithSeqNumber[Organization] {
 
   override def withId(id: Id[Organization]): Organization = this.copy(id = Some(id))
-
   override def withUpdateTime(now: DateTime): Organization = this.copy(updatedAt = now)
+
+  def nonmemberPermissions = basePermissions.forNonmember
+  def rolePermissions(role: OrganizationRole) = basePermissions.forRole(role)
```
leogrim: Try and use verbs when defining methods with parameters :)

kifi-backend/modules/common/conf/evolutions/shoebox/344.sql
leogrim: +1

PR #15857, opened 2015-06-19
A little bit more work on default images.

andrewconner:
@leogrim


kifi-backend/modules/common/app/com/keepit/social/BasicUser.scala
```diff
@@ -73,27 +74,84 @@ object BasicUser {
...
+      case 33 => "penguin"
+      case 34 => "sheep"
+      case 35 => "spider"
+      case 36 => "squirrel"
+      case 37 => "starfish"
+      case 38 => "tiger"
+      case 39 => "turtle"
+      case 40 => "wolf"
+      case 41 => "zebra"
+      case _ => "dog" // the cutest
```
leogrim, line 145: No it's not.

PR #15860, opened 2015-06-19
Accept Invitations to Organization

Colin-Lane:
@ryanpbrewster @leogrim @yingjieMiao

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
@@ -226,17 +226,12 @@ class OrganizationInviteCommanderImpl @Inject() (db: Database,
 
     val sortedInvitations = invitations.sortBy(_.role).reverse
     (membershipOpt match {
-      case Some(membership) =>
-        val modifyRequests = sortedInvitations.map(currentInvitation => OrganizationMembershipModifyRequest(orgId, currentInvitation.inviterId, userId, currentInvitation.role))
-        val findFirstSuccess = modifyRequests.toStream.map(organizationMembershipCommander.modifyMembership(_))
-          .find(_.isRight) // stop at first success
-        findFirstSuccess.map(_.right.map(_.membership))
-          .getOrElse(Left(OrganizationFail.NO_VALID_INVITATIONS))
+      case Some(membership) => Left(OrganizationFail.NO_VALID_INVITATIONS)
       case None => // new membership
         val addRequests = sortedInvitations.map(currentInvitation => OrganizationMembershipAddRequest(orgId, currentInvitation.inviterId, userId, currentInvitation.role))
-        val findFirstSuccess = addRequests.toStream.map(organizationMembershipCommander.addMembership(_))
+        val firstSuccess = addRequests.toStream.map(organizationMembershipCommander.addMembership(_))
           .find(_.isRight)
-        findFirstSuccess.map(_.right.map(_.membership))
+        firstSuccess.map(_.right.map(_.membership))
           .getOrElse(Left(OrganizationFail.NO_VALID_INVITATIONS))
     }).right.map { success =>
       // on success accept invitations
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
@@ -5,6 +5,7 @@ import com.keepit.abook.ABookServiceClient
 import com.keepit.abook.model.RichContact
 import com.keepit.commanders.emails.EmailTemplateSender
 import com.keepit.common.core._
+import com.keepit.common.crypto.PublicIdConfiguration
 import com.keepit.common.db.Id
 import com.keepit.common.db.slick.DBSession.RSession
 import com.keepit.common.db.slick.Database
@@ -15,9 +16,10 @@ import com.keepit.common.mail.template.EmailToSend
 import com.keepit.common.mail.template.TemplateOptions._
 import com.keepit.common.social.BasicUserRepo
 import com.keepit.common.store.S3ImageStore
-import com.keepit.eliza.ElizaServiceClient
+import com.keepit.eliza.{ UserPushNotificationCategory, PushNotificationExperiment, ElizaServiceClient }
 import com.keepit.model._
 import com.keepit.social.BasicUser
+import play.api.libs.json.Json
 
 import scala.concurrent.{ ExecutionContext, Future }
 
@@ -45,7 +47,10 @@ class OrganizationInviteCommanderImpl @Inject() (db: Database,
     elizaClient: ElizaServiceClient,
     s3ImageStore: S3ImageStore,
     userEmailAddressRepo: UserEmailAddressRepo,
-    emailTemplateSender: EmailTemplateSender) extends OrganizationInviteCommander with Logging {
+    emailTemplateSender: EmailTemplateSender,
+    kifiInstallationCommander: KifiInstallationCommander,
+    organizationAvatarCommander: OrganizationAvatarCommander,
+    implicit val publicIdConfig: PublicIdConfiguration) extends OrganizationInviteCommander with Logging {
 
   def inviteUsersToOrganization(orgId: Id[Organization], inviterId: Id[User], invitees: Seq[OrganizationMemberInvitation]): Future[Either[OrganizationFail, Seq[(Either[BasicUser, RichContact], OrganizationRole)]]] = {
     val inviterMembershipOpt = db.readOnlyMaster { implicit session =>
@@ -236,10 +241,44 @@ class OrganizationInviteCommanderImpl @Inject() (db: Database,
     }).right.map { success =>
       // on success accept invitations
       db.readWrite { implicit s =>
-        // TODO: notify inviters about invitee accepting invitation.
+        // Notify inviters on organization joined.
+        notifyInviterOnLibraryInvitationAcceptance(invitations, userRepo.get(userId), organizationRepo.get(orgId))
         invitations.map(_.copy(state = OrganizationInviteStates.ACCEPTED)).foreach(organizationInviteRepo.save(_))
       }
       success
+    }
+  }
+
+  def notifyInviterOnLibraryInvitationAcceptance(invitesToAlert: Seq[OrganizationInvite], invitee: User, org: Organization): Unit = {
+    val inviteeImage = s3ImageStore.avatarUrlByUser(invitee)
+    val orgImageOpt = organizationAvatarCommander.getBestImage(org.id.get, ProcessedImageSize.Medium.idealSize)
+    invitesToAlert foreach { invite =>
+      val title = s"${invitee.firstName} has joined ${org.name}"
+      val inviterId = invite.inviterId
+      elizaClient.sendGlobalNotification( //push sent
+        userIds = Set(inviterId),
+        title = title,
+        body = s"You invited ${invitee.fullName} to join ${org.name}.",
+        linkText = s"See ${invitee.firstName}’s profile",
+        linkUrl = s"https://www.kifi.com/${invitee.username.value}",
+        imageUrl = inviteeImage,
+        sticky = false,
+        category = NotificationCategory.User.ORGANIZATION_JOINED,
+        extra = Some(Json.obj(
+          "member" -> BasicUser.fromUser(invitee),
+          "organization" -> Json.toJson(OrganizationNotificationInfo.fromOrganization(org, orgImageOpt))
+        ))
+      ) map { _ =>
+          val canSendPush = kifiInstallationCommander.isMobileVersionEqualOrGreaterThen(inviterId, KifiAndroidVersion("2.2.4"), KifiIPhoneVersion("2.1.0"))
+          if (canSendPush) {
+            elizaClient.sendUserPushNotification(
+              userId = inviterId,
+              message = title,
+              recipient = invitee,
+              pushNotificationExperiment = PushNotificationExperiment.Experiment1,
+              category = UserPushNotificationCategory.NewOrganizationMember)
+          }
+        }
     }
   }
 
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
@@ -242,14 +242,14 @@ class OrganizationInviteCommanderImpl @Inject() (db: Database,
       // on success accept invitations
       db.readWrite { implicit s =>
         // Notify inviters on organization joined.
-        notifyInviterOnLibraryInvitationAcceptance(invitations, userRepo.get(userId), organizationRepo.get(orgId))
+        notifyInviterOnOrganizationInvitationAcceptance(invitations, userRepo.get(userId), organizationRepo.get(orgId))
         invitations.map(_.copy(state = OrganizationInviteStates.ACCEPTED)).foreach(organizationInviteRepo.save(_))
       }
       success
     }
   }
 
-  def notifyInviterOnLibraryInvitationAcceptance(invitesToAlert: Seq[OrganizationInvite], invitee: User, org: Organization): Unit = {
+  def notifyInviterOnOrganizationInvitationAcceptance(invitesToAlert: Seq[OrganizationInvite], invitee: User, org: Organization): Unit = {
     val inviteeImage = s3ImageStore.avatarUrlByUser(invitee)
     val orgImageOpt = organizationAvatarCommander.getBestImage(org.id.get, ProcessedImageSize.Medium.idealSize)
     invitesToAlert foreach { invite =>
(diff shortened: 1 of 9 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
```diff
@@ -212,8 +212,37 @@ class OrganizationInviteCommanderImpl @Inject() (db: Database,
...
+    (membershipOpt match {
+      case Some(membership) =>
+        val modifyRequests = sortedInvitations.map(currentInvitation => OrganizationMembershipModifyRequest(orgId, currentInvitation.inviterId, userId, currentInvitation.role))
+        val findFirstSuccess = modifyRequests.toStream.map(organizationMembershipCommander.modifyMembership(_))
+          .find(_.isRight) // stop at first success
+        findFirstSuccess.map(_.right.map(_.membership))
+          .getOrElse(Left(OrganizationFail.NO_VALID_INVITATIONS))
+      case None => // new membership
+        val addRequests = sortedInvitations.map(currentInvitation => OrganizationMembershipAddRequest(orgId, currentInvitation.inviterId, userId, currentInvitation.role))
+        val findFirstSuccess = addRequests.toStream.map(organizationMembershipCommander.addMembership(_))
```
Colin-Lane: I need a method that can modify a membership and pick the best valid invitation instead of this.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
```diff
@@ -212,8 +212,37 @@ class OrganizationInviteCommanderImpl @Inject() (db: Database,
...
+          .find(_.isRight)
+        findFirstSuccess.map(_.right.map(_.membership))
+          .getOrElse(Left(OrganizationFail.NO_VALID_INVITATIONS))
+    }).right.map { success =>
+      // on success accept invitations
+      db.readWrite { implicit s =>
+        // TODO: notify inviters about invitee accepting invitation.
+        invitations.map(_.copy(state = OrganizationInviteStates.ACCEPTED)).foreach(organizationInviteRepo.save(_))
+      }
```
Colin-Lane, line 251: TODO: On successful acceptance of invitation notify all distinct inviters about the new user.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
ryanpbrewster: What if we change `OrganizationMembershipAddRequest` (and `ModifyRequest`) to take in `roles: OrganizationRole*` and it sorts them by role and finds the first one that succeeds.

I guess then it'd give back the `OrganizationMembershipAddResponse` with just the request that actually passed.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
```diff
@@ -212,8 +212,37 @@ class OrganizationInviteCommanderImpl @Inject() (db: Database,
...
+    }
+
+    val sortedInvitations = invitations.sortBy(_.role).reverse
+    (membershipOpt match {
+      case Some(membership) =>
+        val modifyRequests = sortedInvitations.map(currentInvitation => OrganizationMembershipModifyRequest(orgId, currentInvitation.inviterId, userId, currentInvitation.role))
+        val findFirstSuccess = modifyRequests.toStream.map(organizationMembershipCommander.modifyMembership(_))
+          .find(_.isRight) // stop at first success
+        findFirstSuccess.map(_.right.map(_.membership))
+          .getOrElse(Left(OrganizationFail.NO_VALID_INVITATIONS))
```
Colin-Lane: Do we want to handle promoting existing members here, or just do it automatically on invitation send?

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
Colin-Lane: That may not work. ex.
An `Admin` invites you to `Admin` with him.
A `Member` invites you to `Member` with him.
You accept the `Member` invitation.
If `ModifyRequest` takes in the member inviter and a `Seq[Role]` then it may fail thinking the `Member` tried to invite someone to `Admin`.
In addition to the `Role` it would also need to take in the `inviterId` as a `Seq[(Role, Inviter)]`.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
ryanpbrewster: I think we just do it on invitation send. That way it's basically a convenience for organization admins, but otherwise it behaves exactly the same as normal promotions.

Colin-Lane: Then I will remove this logical branch.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
```diff
@@ -212,8 +212,37 @@ class OrganizationInviteCommanderImpl @Inject() (db: Database,
...
+      val allInvitations = authInvitations ++ userInvitations
+      val existingMembership = organizationMembershipRepo.getByOrgIdAndUserId(orgId, userId)
+      (allInvitations, existingMembership)
+    }
+
+    val sortedInvitations = invitations.sortBy(_.role).reverse
+    (membershipOpt match {
+      case Some(membership) =>
+        val modifyRequests = sortedInvitations.map(currentInvitation => OrganizationMembershipModifyRequest(orgId, currentInvitation.inviterId, userId, currentInvitation.role))
+        val findFirstSuccess = modifyRequests.toStream.map(organizationMembershipCommander.modifyMembership(_))
```
Colin-Lane: going to rename this to `firstSuccess` instead of `findFirstSuccess`

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
leogrim: I don't think that's good coupling. You add someone with a defined role, not multiple. The issue isn't with all the roles here, it seems like it's with the fact that an inviter's rights may change and a valid invitation may no longer lead to a valid `OrganizationMembershipAddRequest`. I think the simple way to handle this is to check the inviter's rights when the invite is sent and forget about it afterwards. Then when an authentic (as in valid `authToken`) invite is accepted, grab the max role from all other invites and create the membership (don't check inviters' rights again). Either go straight to the repo or make `OrganizationMembershipAddRequest.requesterId` optional (I would go with that last one).

ryanpbrewster: So the other options are: we basically move all of this logic into its own method in `MembershipCommander`, or we write a `addMultipleMemberships(requests: Seq[OrganizationMembershipAddRequest]): Seq[OrganizationMembershipAddResponse]`. My concern about the latter is what the behavior would be if you passed in a bunch of requests from different requesters, for different targets, some of which fail and some of which succeed. How do you even return that? Is it worth the complexity?

Colin-Lane: The current implementation will only check the next invitation if the current does not succeed (we are using `toStream` and `find`). This makes the additional overhead minor for 99% of invitations (how often does someone invite and then get demoted?).
The alternative of using an `optional requestorId` would eliminate any overhead, but then we have introduced an extra `Option` into our code which can lead to dangerous behavior when changing someone's role. I think it is a good idea to keep the requestor id required when modifying someone's role.

leogrim: Then you need to handle the case of an invite being no longer valid, and let the invitee know when this happens - we probably need this anyway for invites that are explicitly revoked. Additionally, note that after accepting an invite for a given role, users may end up with a lesser one.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
```diff
@@ -212,8 +218,68 @@ class OrganizationInviteCommanderImpl @Inject() (db: Database,
...
+          .getOrElse(Left(OrganizationFail.NO_VALID_INVITATIONS))
+    }).right.map { success =>
+      // on success accept invitations
+      db.readWrite { implicit s =>
+        // Notify inviters on organization joined.
+        notifyInviterOnOrganizationInvitationAcceptance(invitations, userRepo.get(userId), organizationRepo.get(orgId))
+        invitations.map(_.copy(state = OrganizationInviteStates.ACCEPTED)).foreach(organizationInviteRepo.save(_))
+      }
+      success
+    }
```
ryanpbrewster, line 253: Can we use use an intermediate `val` to split this block of logic into smaller pieces? Right now the entire flow is a little hard to follow.

Colin-Lane, line 253: Yeah, I guess this is what you would call a run-on-sentence for programmers. I can split it.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
```diff
@@ -136,8 +136,8 @@ class OrganizationMembershipCommanderImpl @Inject() (
       if (validRequest(request)) {
         val membership = organizationMembershipRepo.getByOrgIdAndUserId(request.orgId, request.targetId).get
         val org = organizationRepo.get(request.orgId)
-        organizationMembershipRepo.save(org.modifiedMembership(membership, request.newRole))
-        Right(OrganizationMembershipModifyResponse(request))
+        val newMembership = organizationMembershipRepo.save(org.modifiedMembership(membership, request.newRole))
+        Right(OrganizationMembershipModifyResponse(request, newMembership))
```
ryanpbrewster, line 140: I like these. Maybe we should follow the same pattern in `removeMembership`.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
yingjieMiao, line 253: is this just `invitations.foreach{  organizationInviteRepo.save(_.copy(state = OrganizationInviteStates.ACCEPTED)) }` ?

Colin-Lane, line 253: Kind of, but to make it compile it would need to be `invitations.foreach{ invite => organizationInviteRepo.save(invite.copy(state = OrganizationInviteStates.ACCEPTED)) }`

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
```diff
@@ -212,8 +218,68 @@ class OrganizationInviteCommanderImpl @Inject() (db: Database,
...
+        linkUrl = s"https://www.kifi.com/${invitee.username.value}",
+        imageUrl = inviteeImage,
+        sticky = false,
+        category = NotificationCategory.User.ORGANIZATION_JOINED,
+        extra = Some(Json.obj(
+          "member" -> BasicUser.fromUser(invitee),
+          "organization" -> Json.toJson(OrganizationNotificationInfo.fromOrganization(org, orgImageOpt))
+        ))
+      ) map { _ =>
+          val canSendPush = kifiInstallationCommander.isMobileVersionEqualOrGreaterThen(inviterId, KifiAndroidVersion("2.2.4"), KifiIPhoneVersion("2.1.0"))
```
yingjieMiao: does this second push notification HAS to happen after the global notification? Can we fire two of them at the same time? Any dependency I am missing?

Colin-Lane: That is an interesting question. `sendGlobalNotification` is a `Future`, it may have something to do with exceptions being thrown in the first one (being caught as `SafeFuture`), then not sending to the second. I will have to ask why `LibraryInviteCommander` did it this way.

yingjieMiao: does global notification mean the notification on the extension? At its current state, it just enforces no mobile notification can be pushed until extension gets notified. But it doesn't enforce mobile is sent as well.

Colin-Lane: I am not sure. @andrewconner @leogrim  what is the reason behind why `LibraryInviteCommander:100` maps on the future rather than sending both at once, and should we maybe do it differently for `OrganizationInviteCommander`?

andrewconner: Actually not sure. I was involved in a refactor recently but didn't write it. Should be fine to do both at once.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
leogrim, line 253: Indeed, you can use a single `foreach` with your syntax above is correct.
As a rule of thumb, whenever you use `foreach` if should feel fishy to use it with `map` or `flatMap`, because you're side-effecting anyway.
Similarly, if you use `map` and `filter` together, you can often use `collect` nicely instead.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
```diff
@@ -212,8 +218,69 @@ class OrganizationInviteCommanderImpl @Inject() (db: Database,
...
+      val userInvitations = organizationInviteRepo.getByOrgAndUserId(orgId, userId)
+      val existingMembership = organizationMembershipRepo.getByOrgIdAndUserId(orgId, userId)
+      (userInvitations, existingMembership)
+    }
+
+    val sortedInvitations = invitations.sortBy(_.role).reverse
+    val updatedMembership: Either[OrganizationFail, OrganizationMembership] = membershipOpt match {
+      case Some(membership) => Right(membership) // already a member
+      case None => // new membership
+        val addRequests = sortedInvitations.map(currentInvitation => OrganizationMembershipAddRequest(orgId, currentInvitation.inviterId, userId, currentInvitation.role))
```
leogrim: You could inline `sortedInvitations` here, or at least move its declaration in here - it doesn't need to be in the outside scope, away from where it's actually used and clouding the intent of the main flow. (I agree this is small, but it makes for good habits to hide intermediate variables inside inner scopes)

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationView.scala
```diff
@@ -1,9 +1,31 @@
...
 import com.keepit.common.db.Id
 import com.keepit.common.mail.EmailAddress
+import com.keepit.common.store.ImagePath
+import com.kifi.macros.json
 import play.api.http.Status._
 
+@json case class OrganizationImageInfo(path: ImagePath, x: Int, y: Int)
+
+object OrganizationImageInfo {
+  def createInfo(img: OrganizationAvatar) = OrganizationImageInfo(img.imagePath, img.position.map(_.x).getOrElse(50), img.position.map(_.y).getOrElse(50))
```
leogrim: `OrganizationAvatar` doesn't need a position, my mistake when I designed the initial model. I think I updated the Team Up document since then.

kifi-backend/modules/shoebox/test/com/keepit/commanders/OrganizationInviteCommanderTest.scala
```diff
@@ -119,5 +119,86 @@ class OrganizationInviteCommanderTest extends TestKitSupport with SpecificationL
...
+            UserFactory.user().withId(userId).saved
+            val org = inject[OrganizationRepo].save(Organization(name = "kifi", ownerId = inviterId, handle = None))
+            memberRepo.save(org.newMembership(userId = inviterId, role = OrganizationRole.MEMBER))
+            inviteRepo.save(OrganizationInvite(organizationId = org.id.get, inviterId = inviterId, userId = Some(userId), role = OrganizationRole.OWNER))
+            org
+          }
+          inviteCommander.acceptInvitation(org.id.get, userId) === Left(OrganizationFail.NO_VALID_INVITATIONS)
+        }
+      }
+    }
```
leogrim, line 202: Nice testing :+1:

leogrim:
A few comments to look at, good job overall!

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
Colin-Lane, line 253: Excellent point about the end result side-effecting, that makes sense.

PR #15861, opened 2015-06-19
Rolling out default pics. Done for now.

andrewconner:
@leogrim Stopping here for now.

Next change is to add `pic` and `picSmall` to `BasicUser`. That will take a bit of work (fixing tests primarily). Shelving it for now.


leogrim:
lgtm!

PR #15863, opened 2015-06-19
improved slack UI, fixed a bug

camhashemi:
@andrewconner


kifi-backend/modules/shoebox/test/com/keepit/commanders/LibrarySubscriptionCommanderTest.scala
```diff
@@ -56,9 +56,10 @@ class LibrarySubscriptionCommanderTest extends Specification with ShoeboxTestInj
           LibrarySubscriptionKey("competitors3", SlackInfo("http://www.fakewebhook3.com")))
 
         val (newSubs, didSubsChange) = db.readWrite { implicit s =>
```
andrewconner, line 58: This may work, but I'd be careful of this pattern in general. This is a db rw connection (so a single transaction), and it calls `libSubCommander.updateSubsByLibIdAndKey` which provides its own db session. This session may not read into the other one's result in all cases.

camhashemi, line 58: solved in-person. `libSubCommander.updateSubsByLibIdAndKey` takes an implicit session, and is only called by libraryCommander at this point. If it were to be called by a controller, we would need to have it create its own session.

PR #15864, opened 2015-06-19
OrgCommander can create, modify, and delete

ryanpbrewster:

I tried to use a lot of the same conventions that we set out for OrgMemberships. Still haven't written tests. Doing that now.

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/model/Organization.scala
+++ b/kifi-backend/modules/common/app/com/keepit/model/Organization.scala
@@ -26,6 +26,9 @@ case class Organization(
 
   override def withId(id: Id[Organization]): Organization = this.copy(id = Some(id))
   override def withUpdateTime(now: DateTime): Organization = this.copy(updatedAt = now)
+  def withState(newState: State[Organization]): Organization = this.copy(state = newState)
+  def withName(newName: String): Organization = this.copy(name = newName)
+  def withBasePermissions(newBasePermissions: BasePermissions): Organization = this.copy(basePermissions = newBasePermissions)
 
   def getNonmemberPermissions = basePermissions.forNonmember
   def getRolePermissions(role: OrganizationRole) = basePermissions.forRole(role)
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
@@ -2,16 +2,19 @@ package com.keepit.commanders
 
 import com.google.inject.{ ImplementedBy, Inject, Singleton }
 import com.keepit.common.db.Id
-import com.keepit.common.db.slick.DBSession.RSession
+import com.keepit.common.db.slick.DBSession.{ RWSession, RSession }
 import com.keepit.common.db.slick.Database
 import com.keepit.common.logging.Logging
-import com.keepit.model.OrganizationModification.{ NAME_CHANGE, PERMISSIONS_CHANGE }
-import com.keepit.model.OrganizationPermission.EDIT_ORGANIZATION
+import com.keepit.model.OrganizationPermission.{ VIEW_ORGANIZATION, EDIT_ORGANIZATION }
 import com.keepit.model._
 
 @ImplementedBy(classOf[OrganizationCommanderImpl])
 trait OrganizationCommander {
   def get(orgId: Id[Organization]): Organization
+  def isValidRequest(request: OrganizationRequest)(implicit session: RSession): Boolean
+  def createOrganization(request: OrganizationCreateRequest): Either[OrganizationFail, OrganizationCreateResponse]
+  def modifyOrganization(request: OrganizationModifyRequest): Either[OrganizationFail, OrganizationModifyResponse]
+  def deleteOrganization(request: OrganizationDeleteRequest): Either[OrganizationFail, OrganizationDeleteResponse]
 }
 
 @Singleton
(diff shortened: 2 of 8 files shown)


kifi-backend/modules/common/app/com/keepit/model/Organization.scala
```diff
@@ -26,6 +26,9 @@ case class Organization(
 
   override def withId(id: Id[Organization]): Organization = this.copy(id = Some(id))
   override def withUpdateTime(now: DateTime): Organization = this.copy(updatedAt = now)
+  def withState(newState: State[Organization]): Organization = this.copy(state = newState)
+  def withName(newName: String): Organization = this.copy(name = newName)
```
andrewconner, line 30: You'll see this pattern in several old models, but there's not a huge advantage to `withName` to letting someone just do `.copy(name = newName)`. The two exceptions that come to mind: a) when you're doing some data massaging that wouldn't fit in the copy method, and b) state, id, updatedAt all are interacted with automatically, and traits can't (or, at least in Scala 2.9 when we started) access `copy` since that method is added dynamically because of the `case class`

ryanpbrewster, line 30: Okay. I originally had the `.copy(field = newField)` everywhere. I think it's actually more clear that way, but there are some modifications that are more involved. You think I should keep it the way it is, or go back to the `.copy` syntax?

andrewconner, line 30: You can leave it if you prefer — it _is_ in many models. I just wouldn't necessarily add it elsewhere. For more involved modifications, go for it.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
```diff
@@ -14,7 +17,66 @@ trait OrganizationCommander {
...
+        val orgPrototype = orgRepo.save(Organization(ownerId = request.userId, name = request.orgName, handle = None))
+        val org = handleCommander.autoSetOrganizationHandle(orgPrototype) getOrElse {
+          throw new Exception(s"COULD NOT CREATE ORGANIZATION [$request.orgName] SINCE WE DIDN'T FIND A HANDLE!!!")
+        }
+        Right(OrganizationCreateResponse(request, org))
+      }
+    } catch {
+      case e: Exception => Left(OrganizationFail.HANDLE_UNAVAILABLE)
+    }
+  }
```
Colin-Lane, line 67: Interestingly `validRequest` has a `case OrganizationCreateRequest` but `createOrganization` does not seem to call it.
It may be true right now, but it may not always be that way, and I'd rather have it hooked up correctly right now.

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationView.scala
```diff
@@ -33,13 +33,39 @@ case class OrganizationMembershipAddResponse(request: OrganizationMembershipAddR
...
+case class OrganizationModifyRequest(
+  orgId: Id[Organization],
+  requesterId: Id[User],
+  modification: OrganizationModification) extends OrganizationRequest
+case class OrganizationModifyResponse(request: OrganizationModifyRequest, modifiedOrg: Organization)
+
+case class OrganizationDeleteRequest(
+  orgId: Id[Organization],
+  requesterId: Id[User]) extends OrganizationRequest
+case class OrganizationDeleteResponse(request: OrganizationDeleteRequest, deactivatedOrg: Organization)
```
Colin-Lane, line 74: I like that we are actually returning back the `Organization` with the responses.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
ryanpbrewster, line 67: Good point, my bad.

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationView.scala
ryanpbrewster, line 74: You think we should do the same for `OrganizationMembership`s?

Colin-Lane, line 74: The `OrganizationMembership` already comes free on the save whenever one of these requests is executed. I could see use cases where they may want to actually have the `OrganizationMembership` after instead of just being left with an `Id[OrganizationMembership]`

kifi-backend/modules/common/app/com/keepit/model/Organization.scala
leogrim, line 30: Another advantage is easier refactoring, if you want to do some default data massaging at some point.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
```diff
@@ -14,7 +17,66 @@ trait OrganizationCommander {
...
 class OrganizationCommanderImpl @Inject() (
     db: Database,
-    orgRepo: OrganizationRepo) extends OrganizationCommander with Logging {
+    orgRepo: OrganizationRepo,
+    orgMembershipRepo: OrganizationMembershipRepo,
+    handleCommander: HandleCommander) extends OrganizationCommander with Logging {
 
   def get(orgId: Id[Organization]): Organization = db.readOnlyReplica { implicit session => orgRepo.get(orgId) }
+
+  def validRequest(request: OrganizationRequest)(implicit session: RSession): Boolean = {
```
leogrim: Please use verbs for methods taking parameters. For methods returning a `Boolean`, start with `is`, `has`, `must`, `can`, etc...
Here that would be `isValid` or `isValidRequest`

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
```diff
@@ -14,7 +17,66 @@ trait OrganizationCommander {
...
+    request match {
+      case OrganizationCreateRequest(createrId, orgName) => true // TODO: can we check ahead of time that an org handle is available?
+      case OrganizationModifyRequest(orgId, requesterId, modification) =>
+        orgMembershipRepo.getByOrgIdAndUserId(orgId, requesterId).map(_.permissions).exists {
```
leogrim: You could name the argument here, it would probably be easier to read, or perhaps even this could be slightly clearer:
`orgMembershipRepo.getByOrgIdAndUserId(orgId, requesterId).exists(_.permissions.contains { ...`

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
```diff
@@ -14,7 +17,66 @@ trait OrganizationCommander {
...
+
+  def modifyOrganization(request: OrganizationModifyRequest): Either[OrganizationFail, OrganizationModifyResponse] = {
+    db.readWrite { implicit session =>
+      if (validRequest(request)) {
+        val org = orgRepo.get(request.orgId)
+        request.modification match {
+          case NAME_CHANGE(newName) =>
+            Right(OrganizationModifyResponse(request, orgRepo.save(org.withName(newName))))
+          case PERMISSIONS_CHANGE(newBasePermissions) =>
+            Right(OrganizationModifyResponse(request, orgRepo.save(org.withBasePermissions(newBasePermissions))))
```
leogrim: Thinking of the UI ahead, I can see users making a bunch of changes to the organization before hitting some sort of submit button. That means the `OrganizationModifyRequest` will probably hold several changes at once. Perhaps it should look like an `OrganizationCreateRequest` and we diff it with the current model to decide what should be updated. Check out how we do it for libraries.

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationView.scala
leogrim, line 74: In general, without a known use case, I would default to returning less data in order to focus on what's essential and the method's intent. Now if we already know of a good use case for returning more, let's have it.

ryanpbrewster, line 74: I liked having a consistent convention for all three methods, but I see
your point.

leogrim, line 74: Fair enough.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
ryanpbrewster: I like that.

PR #15866, opened 2015-06-19
decline organization invites

Colin-Lane:
@leogrim @ryanpbrewster
Pretty straight-forward, just decline all active invites.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
@@ -24,7 +24,7 @@ import scala.concurrent.{ ExecutionContext, Future }
 trait OrganizationInviteCommander {
   def inviteUsersToOrganization(orgId: Id[Organization], inviterId: Id[User], invitees: Seq[OrganizationMemberInvitation]): Future[Either[OrganizationFail, Seq[(Either[BasicUser, RichContact], OrganizationRole)]]]
   def acceptInvitation(orgId: Id[Organization], userId: Id[User], authToken: Option[String] = None): Either[OrganizationFail, (Organization, OrganizationMembership)]
-  def declineInvitation(orgId: Id[Organization], userId: Id[User]): Unit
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
```diff
@@ -216,8 +216,11 @@ class OrganizationInviteCommanderImpl @Inject() (db: Database,
...
   }
 
-  def declineInvitation(orgId: Id[Organization], userId: Id[User]) {
-    ???
+  def declineInvitation(orgId: Id[Organization], userId: Id[User]): Seq[OrganizationInvite] = {
+    db.readWrite { implicit s =>
+      organizationInviteRepo.getByOrgIdAndUserId(orgId, userId)
+        .map(inv => organizationInviteRepo.save(inv.copy(state = OrganizationInviteStates.DECLINED)))
+    }
   }
```
ryanpbrewster, line 293: Is this actually the right thing to do? What if a user only wants to decline one invitation because they don't know how invitations work?

Colin-Lane, line 293: This is how Libraries handled declining invites.
I would find it odd if they wanted to decline an invite from one person but accept from another.

ryanpbrewster, line 293: Yeah, and worst-case is they have to shoot someone an email and say they need a new invite. I think this is good as-is.

ryanpbrewster:
lgtm

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
leogrim, line 293: Users see multiple notifications but there's no way to decline a specific invite. You accept or decline to join. So this is good!

leogrim:
lgtm!

PR #15867, opened 2015-06-19
convert pending invites to use userId

Colin-Lane:
@leogrim


kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationInviteRepo.scala
```diff
@@ -10,6 +10,7 @@ import org.joda.time.DateTime
 
 @ImplementedBy(classOf[OrganizationInviteRepoImpl])
 trait OrganizationInviteRepo extends Repo[OrganizationInvite] {
+  def getByEmailAddress(emailAddress: EmailAddress)(implicit session: RSession): Seq[OrganizationInvite]
```
ryanpbrewster, line 13: We don't use it now, but I think it'd be a good idea to write this with a `state: State[OrganizationInvite] = OrganizationInviteStates.ACTIVE` default second argument. I can imagine cases where you would want to get all of the inactive invitations to a particular email.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
```diff
@@ -46,6 +47,14 @@ class OrganizationInviteCommanderImpl @Inject() (db: Database,
     userEmailAddressRepo: UserEmailAddressRepo,
     emailTemplateSender: EmailTemplateSender) extends OrganizationInviteCommander with Logging {
 
+  def convertPendingInvites(emailAddress: EmailAddress, userId: Id[User]): Unit = {
+    db.readWrite { implicit s =>
+      organizationInviteRepo.getByEmailAddress(emailAddress) foreach { invitation =>
```
ryanpbrewster: Is it possible that we could ever end up with invites that are labelled with one user's email address but belong to another user? Like `user2` somehow got invited to `org1` and their invitation has `user1`'s email address on it?

Long story short, do we want to have a more restrictive function `organizationInviteRepo.getNonuserInvitesByEmailAddress` that also checks `row.userId == None`?

leogrim: It's tricky, I guess it depends whether the email's ownership has been verified or not. It looks like for libraries, we're converting all invites regardless of whether an email address has been verified or not (cf `LibraryInviteCommander` line line 251)
@andrewconner was that the intended behavior?

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationInviteRepo.scala
leogrim, line 13: Indeed, one way or another, it's better to exclude `Inactive` rows by default.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
Colin-Lane: I think this is dangerous.
A malicious user could sign up multiple accounts with unverified emails hoping to find an invite that is active to join an `Organization` he has no business joining.
Should we limit only verified users to be able to join an `Organization` and/or have their invitations converted from `EmailAddress` to `Id[User]`?

andrewconner: @leogrim Probably not intended, good catch.

leogrim: So there are two parts here:
- Invites should be converted when an email address ownership switches _unverified_ to _verified_ (as opposed to within the sign up flow).
- When a user joins Kifi via an email invite with a valid auth token, the invited email address is implicitly _verified_ for this new user.

@Colin-Lane since you're in there, can you make this happen both for organizations and libraries?

Colin-Lane: I'll take a look at it. The code is quite tricky since there are many ways to signup.

kifi-backend/modules/shoebox/app/com/keepit/controllers/core/AuthHelper.scala
```diff
@@ -376,10 +377,24 @@ class AuthHelper @Inject() (
     implicit val context = heimdalContextBuilder.withRequestInfo(request).build
     authCommander.finalizeEmailPassAccount(efi, request.userId, request.user.externalId, request.identityOpt, inviteExtIdOpt).map {
       case (user, email, newIdentity) =>
+        convertInvitesForEmailSignup(email, libraryPublicId, libAuthToken)
         finishSignup(user, email, newIdentity, emailConfirmedAlready = false, libraryPublicId = libraryPublicId, libAuthToken = libAuthToken, isFinalizedImmediately = false)
```
Colin-Lane: should it be `emailConfirmedAlready = true` instead if we are implicitly setting email signups as `VERIFIED`?
@leogrim

Colin-Lane: Nevermind. I am doing my own `authToken` checking in `convertInvitesForEmailSignup` I could set it to true though if I pulled out the `authToken` checking logic.

kifi-backend/modules/shoebox/app/com/keepit/commanders/UserEmailAddressCommander.scala
```diff
@@ -0,0 +1,26 @@
...
+class UserEmailAddressCommanderImpl @Inject() (db: Database,
+    userEmailAddressRepo: UserEmailAddressRepo,
+    libraryInviteCommander: LibraryInviteCommander,
+    organizationInviteCommander: OrganizationInviteCommander) extends UserEmailAddressCommander with Logging {
+
+  def saveAsVerified(emailAddress: UserEmailAddress)(implicit session: RWSession): UserEmailAddress = {
+    libraryInviteCommander.convertPendingInvites(emailAddress = emailAddress.address, userId = emailAddress.userId)
+    organizationInviteCommander.convertPendingInvites(emailAddress = emailAddress.address, userId = emailAddress.userId)
+    userEmailAddressRepo.save(emailAddress.copy(state = UserEmailAddressStates.VERIFIED, verifiedAt = Some(currentDateTime)))
+  }
```
leogrim, line 25: :+1:

kifi-backend/modules/shoebox/app/com/keepit/controllers/core/AuthHelper.scala
leogrim: It seems like this should be called something like `verifySignupEmail` instead, which would do the right thing (i.e. if there's a auth token from an invite to that email, verify it right away otherwise send an email confirmation request?)

kifi-backend/modules/shoebox/app/com/keepit/commanders/SendgridCommander.scala
```diff
@@ -155,19 +156,16 @@ class SendgridCommander @Inject() (
...
-      }
+  private def verifyEmailAddress(event: SendgridEvent, email: ElectronicMail): Unit = db.readWrite { implicit rw =>
+    for {
+      userEmail <- email.to.headOption
+      emailAddr <- emailAddressRepo.getByAddressOpt(userEmail)
+      if !emailAddr.verified
+    } yield {
+      log.info(s"verifying email($userEmail) from SendGrid event($event)")
+      userEmailAddressCommander.saveAsVerified(emailAddr)
     }
```
ryanpbrewster, line 167: Don't think yield is necessary here

ryanpbrewster, line 167: More specifically, this function returns a Unit. Is it supposed to?

Colin-Lane, line 167: I think it is just supposed to save an email address as `VERIFIED`

ryanpbrewster, line 167: Right, so it doesn't need to `yield` anything.

Colin-Lane, line 167: Yeah, I'll get rid of that.

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationInviteRepo.scala
Colin-Lane, line 13: With invitations, we need to be careful since it is not just a 2 state system. Excluding `Inactive` would then give us results for `Active`, `Accepted`, and `Declined` which probably isn't what we want either. In this case, we don't want to exclude anything, we want to convert `all` invites from a user's email to their userId.

ryanpbrewster, line 13: If there is a specific use-case for it, I think the `excludedStates` or `includedStates` convention is okay.

leogrim, line 13: Could you make it a two-state system and add an `answer: Option[Boolean]` field?

ryanpbrewster, line 13: Or an Option[OrgInviteDecision]

Colin-Lane, line 13: I think that is an excellent idea. I'll add it to the trello board, and have that be a separate PR. I'd like to get this merged before any more time goes by.

leogrim, line 13: Yep it's been open for a while, merge it :)

Colin-Lane, line 13: Ha, it just changed from `can automatically merge` to `can't automatically merge`. Going to resolve merge conflicts then merge.

PR #15868, opened 2015-06-19
slack UX and bug-fix

camhashemi:
@andrewconner when a user provides blank inputs, we should remove them before making the API call. But, this manipulates the DOM-binding right before closing the modal, leading to an uncomfortable experience.

My (buggy and) inaccurate solution was to send blank inputs to the backend and filter them there. This new solution filters on the front-end without manipulating the DOM.


andrewconner:
Looks good

PR #15869, opened 2015-06-20
Marketing fixes

andrewconner:


andrewconner:
@sterlingz

andrewconner:
We went over this in person.

PR #15870, opened 2015-06-20
[DO NOT MERGE] Org avatars

ryanpbrewster:
@leogrim @andrewconner @Colin-Lane @yingjieMiao 

This compiles and passes tests, but if I'm honest I don't really understand the code I've written.
Also I completely ignored analytics.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationAvatarCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationAvatarCommander.scala
@@ -1,0 +1,245 @@
+package com.keepit.commanders
+
+import java.sql.SQLException
+
+import com.google.inject.{ ImplementedBy, Inject, Singleton }
+import com.keepit.common.db.slick.DBSession.RWSession
+import com.keepit.common.db.slick.Database
+import com.keepit.common.db.{ Id, State }
+import com.keepit.common.images.Photoshop
+import com.keepit.common.logging.Logging
+import com.keepit.common.net.WebService
+import com.keepit.common.store.{ ImageSize, RoverImageStore, S3ImageConfig }
+import com.keepit.heimdal.HeimdalContext
+import com.keepit.model.ProcessImageOperation.Original
+import com.keepit.model._
+import play.api.libs.Files.TemporaryFile
+
+import scala.concurrent.{ ExecutionContext, Future }
+
+object OrganizationAvatarSizes {
+  val scaleSizes = ScaledImageSize.allSizes
+  val cropSizes = Seq.empty
+}
+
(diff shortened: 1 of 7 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationAvatarCommander.scala
```diff
@@ -0,0 +1,245 @@
...
+  def getUrl(orgAvatar: OrganizationAvatar): String
+  def getBestImageForOrganization(orgId: Id[Organization], idealSize: ImageSize): Option[OrganizationAvatar]
+  def getBestImageForOrganizations(orgIds: Set[Id[Organization]], idealSize: ImageSize): Map[Id[Organization], OrganizationAvatar]
+  def uploadOrganizationAvatarFromFile(image: TemporaryFile, orgId: Id[Organization], position: ImagePosition, source: ImageSource, userId: Id[User], requestId: Option[Id[OrganizationAvatarRequest]] = None)(implicit content: HeimdalContext): Future[ImageProcessDone]
+  def positionOrganizationAvatar(orgId: Id[Organization], position: ImagePosition): Seq[OrganizationAvatar]
+  def removeImageForOrganization(orgId: Id[Organization], userId: Id[User])(implicit context: HeimdalContext): Boolean // Returns true if images were removed, false otherwise
+}
+
+@Singleton
+class OrganizationAvatarCommanderImpl @Inject() (
```
yingjieMiao: add comment: Copied wholesale from LibraryImageCommander. 

Wondering if we can reuse code (or DRY the code), by factoring out some abstract class or trait. (Maybe not this PR?)

Keep this much copied code synced at two place can become a nightmare. 

It's really hard to tell what's new, what's modified, etc.

Looking at the difference, seems that there are repos of `org` or `library`. I guess they are just used to figure out some image meta data (id, path, size, etc). At least retrieving from `imageStore` could be factored out?

Not really familiar with the codes, throwing some ideas.

yingjieMiao: As a reference: `FixedSetPersonaRecoPool` , `URIFixedSetPersonaRecoPool`, `LibraryFixedSetPersonaRecoPool`

Not necessarily a clean or good example. Something for inspiration.

ryanpbrewster: I definitely think that there is _something_ here that can be factored out. I chatted with @leogrim about it and I have some ideas.

leogrim: Yep, I would probably close this PR and start from scratch looking at the code in `Rover` instead.

PR #15876, opened 2015-06-22
Temporary fix to keep extractors blowing up (it's flooding current users email inboxes).

andrewconner:
@eishay


eishay:
swallow exception? :-(

andrewconner:
Apparently, there's no error handling here, so if we have an exception, we re-process the next batch on the following minute. We ended up sending a confirmation email to a user every minute for many hours. Swallowing is the best right now.

PR #15877, opened 2015-06-22
Fixing two long-standing email bugs

andrewconner:
repeatedly tell people that we kept their links, and handling empty emails sent to us

@eishay @leogrim


andrewconner:
Including a build fix, merging. Feedback appreciated.

leogrim:
:+1:

PR #15880, opened 2015-06-22
Organization create/modify/delete logic

ryanpbrewster:
@Colin-Lane @leogrim @yingjieMiao 

Cleaned up, passes tests. I'll merge unless y'all have objections.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
@@ -2,16 +2,19 @@ package com.keepit.commanders
 
 import com.google.inject.{ ImplementedBy, Inject, Singleton }
 import com.keepit.common.db.Id
-import com.keepit.common.db.slick.DBSession.RSession
+import com.keepit.common.db.slick.DBSession.{ RWSession, RSession }
 import com.keepit.common.db.slick.Database
 import com.keepit.common.logging.Logging
-import com.keepit.model.OrganizationModification.{ NAME_CHANGE, PERMISSIONS_CHANGE }
-import com.keepit.model.OrganizationPermission.EDIT_ORGANIZATION
+import com.keepit.model.OrganizationPermission.{ VIEW_ORGANIZATION, EDIT_ORGANIZATION }
 import com.keepit.model._
 
 @ImplementedBy(classOf[OrganizationCommanderImpl])
 trait OrganizationCommander {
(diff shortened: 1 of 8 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
```diff
@@ -2,19 +2,117 @@ package com.keepit.commanders
...
+        permissions.exists { _.contains(EDIT_ORGANIZATION) } && !hasInvalidModification(modifications)
+      case OrganizationDeleteRequest(orgId, requesterId) =>
+        requesterId == orgRepo.get(orgId).ownerId
+    }
+  }
+  private def hasInvalidModification(modifications: OrganizationModifications): Boolean = {
+    lazy val badName = modifications.newName.exists(_.isEmpty)
+    lazy val badBasePermissions = modifications.newBasePermissions.exists { bps =>
+      // Are there any members that can't even see the organization?
+      !OrganizationRole.all.forall { role => bps.forRole(role).contains(VIEW_ORGANIZATION) }
```
Colin-Lane: the double negative made me think twice about this since we are checking for `invalid` modifications. Would it be better to rename the method to `hasValidModifications` and invert the conditions.

Colin-Lane: We could even name it `isValidModificationRequest`

ryanpbrewster: I originally had it that way, but I inverted it because the logic was kind of hard to follow.

The condition for being invalid is a little easier to think about, since the no-op is automatically valid. Thus, we only care about changes that both EXIST and are INVALID.

leogrim: Did you see my comments about modifications? I would expected users to update the organization and "submit" their changes altogether, so I don't know if individual modifications make sense.

ryanpbrewster: The `modifyOrganization` method takes in a whole list of modifications.

ryanpbrewster: Rather, it takes in an `OrganizationModifications` object, which encompasses all the fields that can be edited in an `Organization`, and checks each modification one-by-one to ensure they are all valid modifications.

PR #15885, opened 2015-06-23
added support for UserIpAddress in the graph

camhashemi:
@leogrim


kifi-backend/modules/shoebox/marketing
```diff
@@ -1 +1 @@
-Subproject commit 3a5e5c28af64e6adc0626eeffbdc4bc411a58050
+Subproject commit 9c0199f8f42751182f1705b493a1fae2863aa3a9
```
camhashemi: @andrewconner 9c0199f matches the hash in the submodule, should I revert this?

kifi-backend/modules/common/conf/evolutions/shoebox/345.sql
```diff
@@ -0,0 +1,18 @@
...
+-- ALTER TABLE user_ip_addresses ADD seq INT NOT NULL DEFAULT 0;
+-- CREATE TABLE ip_address_sequence (id INT NOT NULL);
+-- INSERT INTO ip_address_sequence VALUES (0);
+-- H2:
+CREATE SEQUENCE ip_address_sequence;
+---
+
+ALTER TABLE user_ip_addresses ADD seq INT NOT NULL DEFAULT 0;
+CREATE INDEX ip_address_seq_index ON user_ip_addresses(seq);
+
```
yingjieMiao, line 16: you probably need a shoebox endpoint to set seqNum for existing rows. (in a different PR)

kifi-backend/modules/common/app/com/keepit/common/routes/Router.scala
```diff
@@ -158,6 +159,7 @@ object Shoebox extends Service {
     def getLibrariesWithWriteAccess(userId: Id[User]) = ServiceRoute(GET, "/internal/shoebox/database/getLibrariesWithWriteAccess", Param("userId", userId))
     def getUserActivePersonas(userId: Id[User]) = ServiceRoute(GET, "/internal/shoebox/user/activePersonas", Param("userId", userId))
     def getLibraryURIS(libId: Id[Library]) = ServiceRoute(GET, "/internal/shoebox/database/dumpLibraryURIIds", Param("libId", libId))
+    def getUserIpAddressesChanged(sequenceNumber: SequenceNumber[IngestableUserIpAddress], fetchSize: Int) = ServiceRoute(GET, "/internal/shoebox/database/getUserIpAddressesChanged")
```
leogrim: Moving towards something consistent, `getIngestableUserIpAddresses`

kifi-backend/modules/common/app/com/keepit/common/service/IpAddress.scala
```diff
@@ -5,6 +5,15 @@ import play.api.libs.json._
 object IpAddress {
   val ipPattern = """^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])$"""
   implicit val format: Format[IpAddress] = Format(__.read[String].map(IpAddress(_)), Writes(ip => JsString(ip.ip)))
+
+  implicit def ipToLong(ipAddress: IpAddress): Long = {
+    val octets = ipAddress.ip.split("\\.")
+    Math.round(octets.zipWithIndex.map { case (x, i) => Integer.parseInt(x) * Math.pow(256, 3 - i) }.sum)
```
leogrim: Nice! A couple of comments: 
- you can just reverse the list instead of hardcoding a `3` in there. 
- `Integer.parseInt(x) <=> x.toInt`

Why rounding?

camhashemi: I agree with all three points, taking the last one as a rhetorical question. Thank you.

kifi-backend/modules/common/app/com/keepit/common/service/IpAddress.scala
```diff
@@ -5,6 +5,15 @@ import play.api.libs.json._
...
+  }
+
+  implicit def longToIp(long: Long): IpAddress = {
+    IpAddress(((long >> 24) & 0xFF) + "." + ((long >> 16) & 0xFF) + "." + ((long >> 8) & 0xFF) + "." + (long & 0xFF))
+  }
```
leogrim, line 16: :+1:

kifi-backend/modules/common/conf/evolutions/shoebox/345.sql
```diff
@@ -0,0 +1,18 @@
+# SHOEBOX
+
+# --- !Ups
+
+-- MySQL:
+-- ALTER TABLE user_ip_addresses ADD seq INT NOT NULL DEFAULT 0;
+-- CREATE TABLE ip_address_sequence (id INT NOT NULL);
+-- INSERT INTO ip_address_sequence VALUES (0);
+-- H2:
+CREATE SEQUENCE ip_address_sequence;
```
leogrim: This should match the table name, so `user_ip_addresses_sequence`
I'm not sure why addresses is plural, but might as well go for consistency.

kifi-backend/modules/common/app/com/keepit/common/service/IpAddress.scala
camhashemi: Basically, Math.pow works with doubles and I need to convert/cast to a Long. I think Math.round might be safer but x.toLong is more readable.

kifi-backend/modules/common/conf/evolutions/shoebox/345.sql
leogrim, line 16: Here's the magic query:

`update your_table_sequence set id = 1`
`update your_table set seq = (UNIX_TIMESTAMP(updated_at) * 1000 - 9223372036854775807)`

(and the plugin will pick it up from there)

camhashemi, line 16: awesome

kifi-backend/modules/graph/app/com/keepit/graph/model/VertexDataId.scala
```diff
@@ -44,4 +45,7 @@ object VertexDataId {
 
   implicit def fromLibraryId(libId: Id[Library]): VertexDataId[LibraryReader] = VertexDataId(libId.id)
   implicit def toLibraryId(libReaderId: VertexDataId[LibraryReader]): Id[Library] = Id(libReaderId.id)
+
+  implicit def fromIpAddr(ipAddr: IpAddress): VertexDataId[IpAddressReader] = VertexDataId(ipAddr)
+  implicit def toIpAddr(ipReaderId: VertexDataId[IpAddressReader]): IpAddress = ipReaderId.id
```
leogrim: Could you spell these out? I'm not a fan of abbreviations in general, especially in method names. The more it reads like English, the better :)

leogrim:
Minor comments here and there. Awesome job, that was fast!

kifi-backend/modules/common/app/com/keepit/common/service/IpAddress.scala
leogrim: You don't really want to be safe here though, you want to fail loud and clear if you don't end up with integers. You could also use bit shifting like you do the other way around, the symmetry would be nice.

camhashemi:
Thank you, it's really easy given the architecture. Ok to merge then? And the order is 1) manual MySQL query, 2) merge the evolutions, right?

kifi-backend/modules/common/conf/evolutions/shoebox/345.sql
```diff
@@ -0,0 +1,20 @@
+# SHOEBOX
+
+# --- !Ups
+
+-- MySQL:
+-- ALTER TABLE user_ip_addresses ADD seq INT NOT NULL DEFAULT 0;
```
leogrim: This should be bigint

kifi-backend/modules/common/conf/evolutions/shoebox/345.sql
```diff
@@ -0,0 +1,20 @@
...
+-- CREATE TABLE user_ip_addresses_sequence (id INT NOT NULL);
```
leogrim: This as well, bigint

kifi-backend/modules/common/conf/evolutions/shoebox/345.sql
```diff
@@ -0,0 +1,20 @@
...
+-- INSERT INTO user_ip_addresses_sequence VALUES (0);
+-- UPDATE user_ip_addresses_sequence SET id = 1;
```
leogrim: You can replace these two by just `INSERT INTO user_ip_addresses_sequence VALUES (1);`

kifi-backend/modules/common/conf/evolutions/shoebox/345.sql
```diff
@@ -0,0 +1,20 @@
...
+-- UPDATE user_ip_addresses SET seq = (UNIX_TIMESTAMP(updated_at) * 1000 - 9223372036854775807);
```
leogrim, line 9: `UPDATE user_ip_addresses SET seq = (UNIX_TIMESTAMP(updated_at) * 1000 - 9223372036854775807) where seq = 0`

leogrim:
See my latest comments on the evolution, be careful about bigint, I had missed that.
- Add the new rows / table.
- Apply the manual update query
- Deploy the new code
- Apply the manual update query again to fix the rows that slip through during deploy

PR #15886, opened 2015-06-23
library path commander

yingjieMiao:
@leogrim 

`LibraryPathHelper` trait and `LibraryPathCommander`. 

The commander is only available on shoebox, but I think it's still very helpful for the next refactoring. The commander takes a library and find the user object and org object. Without it, we have to manually add code to retrieve org object in 30+ places.

This PR doesn't involve constructing org related path. Simply a wrapper.


kifi-backend/modules/shoebox/app/com/keepit/admin/AdminTwitterWaitlistController.scala
```diff
@@ -40,7 +41,7 @@ class AdminTwitterWaitlistController @Inject() (
         val owner = userRepo.get(lib.ownerId)
         (lib, owner)
       }
-      val libraryPath = Library.formatLibraryPath(owner.username, lib.slug)
+      val libraryPath = libPathCommander.getPath(lib)
```
Colin-Lane, line 44: why use a different commander here when everywhere else you `extend LibraryPathHelper` and use `formatLibraryPath`? Would it be better to use a single convention everywhere?

kifi-backend/modules/common/app/com/keepit/model/Library.scala
```diff
@@ -146,6 +146,14 @@ object Library extends ModelWithPublicIdCompanion[Library] {
...
 
+  def toLibraryView(lib: Library): LibraryView = LibraryView(id = lib.id, ownerId = lib.ownerId, state = lib.state, seq = lib.seq, kind = lib.kind)
+
+  def toDetailedLibraryView(lib: Library): DetailedLibraryView = DetailedLibraryView(id = lib.id, ownerId = lib.ownerId, state = lib.state,
+    seq = lib.seq, kind = lib.kind, memberCount = lib.memberCount, keepCount = lib.keepCount, lastKept = lib.lastKept, lastFollowed = None, visibility = lib.visibility,
+    updatedAt = lib.updatedAt, name = lib.name, description = lib.description, color = lib.color, slug = lib.slug)
+}
+
+trait LibraryPathHelper {
   def formatLibraryPath(ownerUsername: Username, slug: LibrarySlug): String = {
```
leogrim, line 157: What's the benefit of using inheritance here? Can't we just use an object? Whether it's the `Library` object itself or a new one?

kifi-backend/modules/shoebox/app/com/keepit/commanders/FacebookPublishingCommander.scala
```diff
@@ -35,7 +36,7 @@ class FacebookPublishingCommander @Inject() (
         case None => log.info(s"user $userId is not connected to facebook!")
         case Some(sui) =>
           val libOwner = db.readOnlyMaster { implicit session => userRepo.get(library.ownerId) }
-          val libraryUrl = s"""https://www.kifi.com${Library.formatLibraryPath(libOwner.username, library.slug)}"""
+          val libraryUrl = s"""https://www.kifi.com${libPathCommander.getPath(library)}"""
```
leogrim, line 39: The library owner is already here. As we introduce organizations, we'll probably load it here as well. I think `libPathCommander.getPath` should take optional arguments so that you can provide them if you have them. Then the method checks whether they match the library, if not airbrake and loads the correct ones.

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminTwitterWaitlistController.scala
yingjieMiao, line 44: everywhere else probably means in `search` service. My original plan is to have a commander, so that in the future, any refactor will happen inside that commander, instead of modifying a raw api that is called in 40+ places (as of now).  The commander needs the repos, which are not available outside `shoebox`. I created the trait so search can use it.

kifi-backend/modules/common/app/com/keepit/model/Library.scala
yingjieMiao, line 157: I thought it was helpful during refactoring. I wanted to take the logic out of `Library`, but without touching test files. 

No practical benefits indeed. A separate object may be more appropriate.

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminTwitterWaitlistController.scala
leogrim, line 44: The trait only contains static methods, so it's better for it to be an object. Composition > Inheritance.

kifi-backend/modules/common/app/com/keepit/model/Library.scala
leogrim, line 157: :+1:

kifi-backend/modules/shoebox/app/com/keepit/commanders/FacebookPublishingCommander.scala
yingjieMiao, line 39: true. Out of 30+ places, there are 1 or 2 places we load user object explicitly because it will be used later. The performance loss is minimal as it will be likely read from cache. 

I can add optional argument later.

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminTwitterWaitlistController.scala
yingjieMiao, line 44: yeah, I guess French can hardly go wrong with style.

yingjieMiao:
@leogrim 

If there is no major concern except style, I'd like to merge this to avoid excess merge conflicts. I have another PR in a different branch.

Will rewrite the trait into an object in that PR.

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminTwitterWaitlistController.scala
leogrim, line 44: lol

kifi-backend/modules/shoebox/app/com/keepit/commanders/FacebookPublishingCommander.scala
leogrim, line 39: Ok!

leogrim:
Thanks!

PR #15887, opened 2015-06-23
better precision

yingjieMiao:
@andrewconner @stkem 

revert a change that boosts recall. Since we are doing experiment now, I am in favor of precision over recall. And cleaner code.


andrewconner:
Works for me

PR #15888, opened 2015-06-23
Mobile Invite Controller

Colin-Lane:
@leogrim @ryanpbrewster

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
@@ -108,9 +108,8 @@ class OrganizationMembershipCommanderImpl @Inject() (
     requesterOpt exists { requester =>
       request match {
         case OrganizationMembershipAddRequest(_, _, _, newRole) =>
-          val inviterMembership = organizationMembershipRepo.getByOrgIdAndUserId(request.orgId, request.requesterId)
           targetOpt.isEmpty && (newRole <= requester.role) &&
-            inviterMembership.exists(_.permissions.contains(OrganizationPermission.INVITE_MEMBERS))
+            requester.permissions.contains(OrganizationPermission.INVITE_MEMBERS)
 
         case OrganizationMembershipModifyRequest(_, _, _, newRole) =>
           targetOpt.exists(_.role < requester.role) && (newRole <= requester.role)
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
@@ -108,7 +108,9 @@ class OrganizationMembershipCommanderImpl @Inject() (
     requesterOpt exists { requester =>
       request match {
         case OrganizationMembershipAddRequest(_, _, _, newRole) =>
-          targetOpt.isEmpty && (newRole <= requester.role)
+          val inviterMembership = organizationMembershipRepo.getByOrgIdAndUserId(request.orgId, request.requesterId)
+          targetOpt.isEmpty && (newRole <= requester.role) &&
+            inviterMembership.exists(_.permissions.contains(OrganizationPermission.INVITE_MEMBERS))
 
         case OrganizationMembershipModifyRequest(_, _, _, newRole) =>
           targetOpt.exists(_.role < requester.role) && (newRole <= requester.role)
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileOrganizationInviteController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileOrganizationInviteController.scala
@@ -9,22 +9,33 @@ import com.keepit.common.mail.EmailAddress
 import com.keepit.heimdal.HeimdalContextBuilderFactory
 import com.keepit.inject.FortyTwoConfig
 import com.keepit.model._
+import com.keepit.shoebox.controllers.OrganizationAccessActions
 import play.api.libs.json._
+import play.api.mvc.{ Action, AnyContent }
 
 import scala.concurrent.{ ExecutionContext, Future }
 import scala.util.{ Failure, Success }
 
+@ImplementedBy(classOf[MobileOrganizationInviteControllerImpl])
+trait MobileOrganizationInviteController {
+  def inviteUsers(pubId: PublicId[Organization]): Action[JsValue]
+  def createAnonymousInviteToOrganization(pubId: PublicId[Organization]): Action[JsValue]
+  def acceptInvitation(pubId: PublicId[Organization]): Action[AnyContent]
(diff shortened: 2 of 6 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
```diff
@@ -108,7 +108,8 @@ class OrganizationMembershipCommanderImpl @Inject() (
     requesterOpt exists { requester =>
       request match {
         case OrganizationMembershipAddRequest(_, _, _, newRole) =>
-          targetOpt.isEmpty && (newRole <= requester.role)
+          targetOpt.isEmpty && (newRole <= requester.role) &&
+            requester.permissions.contains(OrganizationPermission.INVITE_MEMBERS)
```
Colin-Lane, line 112: I'm thinking this may be a good idea. We probably also need an OrganizationPermission.REMOVE_MEMBERS

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationView.scala
```diff
@@ -77,18 +79,23 @@ case class OrganizationModifications(
   newName: Option[String],
   newBasePermissions: Option[BasePermissions])
 
-sealed abstract class OrganizationFail(val status: Int, val message: String)
+sealed abstract class OrganizationFail(val status: Int, val message: String) {
+  def asResponse = Status(status)(Json.obj("error" -> message))
```
Colin-Lane: We have many places where we want to return some sort of error code back based on the OrganizationFail.
We won't accidentally typo the json message or http-status when trying to give frontend an error response, additionally it is easy to see what kinds of errors we return back all in one place.
What do you think about having a method like this?

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
ryanpbrewster, line 112: I like adding more permissions when they don't overlap. Maybe even `MODIFY_MEMBERS`.

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileOrganizationInviteController.scala
```diff
@@ -99,7 +110,23 @@ class MobileOrganizationInviteController @Inject() (
...
+      case Failure(_) => OrganizationFail.INVALID_PUBLIC_ID.asResponse
+    }
+  }
 
+  def declineInvitation(pubId: PublicId[Organization]) = UserAction { request =>
+    Organization.decodePublicId(pubId) match {
+      case Success(orgId) =>
+        organizationInviteCommander.declineInvitation(orgId, request.userId)
+        Ok(JsString("success"))
+      case _ => OrganizationFail.INVALID_PUBLIC_ID.asResponse
```
ryanpbrewster: Yeah, I'm a big fan of this. Good thought. Let's start doing this more.

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationView.scala
ryanpbrewster: :+1:

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationView.scala
```diff
@@ -77,18 +79,23 @@ case class OrganizationModifications(
...
+}
 object OrganizationFail {
   case object INSUFFICIENT_PERMISSIONS extends OrganizationFail(UNAUTHORIZED, "insufficient_permissions")
   case object HANDLE_UNAVAILABLE extends OrganizationFail(FORBIDDEN, "handle_unavailable")
   case object NOT_A_MEMBER extends OrganizationFail(UNAUTHORIZED, "not_a_member")
   case object NO_VALID_INVITATIONS extends OrganizationFail(FORBIDDEN, "no_valid_invitations")
+  case object INVALID_PUBLIC_ID extends OrganizationFail(BAD_REQUEST, "invalid_organization_id")
```
ryanpbrewster: I'd say pick "public" or "organization", but make it consistent.

Colin-Lane: Nice find. I'll fix this.

kifi-backend/modules/shoebox/test/com/keepit/commanders/OrganizationMembershipCommanderTest.scala
```diff
@@ -77,8 +77,9 @@ class OrganizationMembershipCommanderTest extends TestKitSupport with Specificat
         val ownerAddUser = OrganizationMembershipAddRequest(orgId, Id[User](1), Id[User](2), OrganizationRole.MEMBER)
         orgMembershipCommander.addMembership(ownerAddUser) must haveClass[Right[OrganizationFail, OrganizationMembershipAddResponse]]
 
+        // members cannot invite users by default.
         val memberAddUser = OrganizationMembershipAddRequest(orgId, Id[User](2), Id[User](3), OrganizationRole.MEMBER)
-        orgMembershipCommander.addMembership(memberAddUser) must haveClass[Right[OrganizationFail, OrganizationMembershipAddResponse]]
+        orgMembershipCommander.addMembership(memberAddUser) === Left(OrganizationFail.INSUFFICIENT_PERMISSIONS)
```
ryanpbrewster, line 82: Nice. I'll add a few more tests to the membership commander today to get more broad coverage.

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileOrganizationInviteController.scala
leogrim: That's a nice idea!

leogrim: One thing to consider though, is that it becomes easier to break APIs. I have come to learn that it is sometimes better to spell out the json response each controller is returning, because they're bound to evolve on their own as clients change (and mobile always has to be backward compatible).

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationView.scala
leogrim: :+1:

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileOrganizationInviteController.scala
ryanpbrewster: But surely we want to return the same json response every time we encounter an invalid organization id. I like @Colin-Lane's idea because it basically enforces a bijection.

leogrim: If there's a bijection here, then there's a bijection between the controllers, which means changing one may involuntarily break another. We usually think about the website where changes are rolled out immediately, but it's easy to forget about mobile APIs.

PR #15890, opened 2015-06-23
Removing ImageDataIntegrityPlugin

andrewconner:
@leogrim @eishay I think it's outlived its usefulness. It only checks user pictures, and is very loud because of the 0.jpg issue. This issue (broken default images) is largely mitigated now with the default pics.


andrewconner:
The use case was making sure images were actually there. This was important because existence on S3 was implicit, rather than explicit. We've moved to a model where we're very sure it's on S3 because we store one db entry per object.

andrewconner:
Reviewed by Eishay

PR #15891, opened 2015-06-23
Org avatars 2

ryanpbrewster:
`OrganizationAvatarCommander` can take a user-uploaded image (as a `TemporaryFile`),
process it to get a range of useful image shapes, persist all of those images to S3 (via
`S3OrganizationAvatarStoreImpl`), and store the relevant S3 image paths in the
`OrganizationAvatarRepo`.

@andrewconner @Colin-Lane @leogrim 
I'll go in and tag places that I'm concerned about.

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/commanders/ProcessedImageHelper.scala
+++ b/kifi-backend/modules/common/app/com/keepit/commanders/ProcessedImageHelper.scala
@@ -153,13 +153,23 @@ trait ProcessedImageHelper {
     }
   }
 
-  protected def filterProcessImageRequests(expectedImages: Set[ProcessImageRequest], existingImages: Set[ProcessImageRequest]): Set[ProcessImageRequest] = {
+  // All the ProcessImageRequests in A that are also in B
+  def intersectProcessImageRequests(A: Set[ProcessImageRequest], B: Set[ProcessImageRequest]): Set[ProcessImageRequest] = {
+    @inline def boundingBox(imageSize: ImageSize): Int = Math.max(imageSize.width, imageSize.height)
+    val Bp = B.collect { case ProcessImageRequest(ProcessImageOperation.Scale, size) => boundingBox(size) }
+    A filter {
+      case ProcessImageRequest(ProcessImageOperation.Scale, size) => Bp.contains(boundingBox(size))
+      case otherRequest => B.contains(otherRequest)
+    }
+  }
+  // All the ProcessImageRequests in A that are NOT in B
+  protected def diffProcessImageRequests(A: Set[ProcessImageRequest], B: Set[ProcessImageRequest]): Set[ProcessImageRequest] = {
     // hack to eliminate expecting scale versions that have the same scaled bounding box
     @inline def boundingBox(imageSize: ImageSize): Int = Math.max(imageSize.width, imageSize.height)
-    val existingScaleBoundingBoxes = existingImages.collect { case ProcessImageRequest(ProcessImageOperation.Scale, size) => boundingBox(size) }
-    expectedImages filterNot {
-      case ProcessImageRequest(ProcessImageOperation.Scale, size) => existingScaleBoundingBoxes.contains(boundingBox(size))
-      case otherRequest => existingImages.contains(otherRequest)
+    val Bp = B.collect { case ProcessImageRequest(ProcessImageOperation.Scale, size) => boundingBox(size) }
+    A filterNot {
+      case ProcessImageRequest(ProcessImageOperation.Scale, size) => Bp.contains(boundingBox(size))
+      case otherRequest => B.contains(otherRequest)
     }
   }
 
--- a/kifi-backend/modules/common/app/com/keepit/commanders/ProcessedImageHelper.scala
+++ b/kifi-backend/modules/common/app/com/keepit/commanders/ProcessedImageHelper.scala
@@ -165,9 +165,9 @@ trait ProcessedImageHelper {
 
   // Returns Set of bounding box sizes
   protected def calcSizesForImage(imageSize: ImageSize, scaleCandidates: Seq[ScaledImageSize], cropCandidates: Seq[CroppedImageSize]): Set[ProcessImageRequest] = {
-    val scaleSizes = scaleCandidates.map { size =>
+    val scaleSizes = scaleCandidates.flatMap { size =>
       calcResizeBoundingBox(imageSize, size.idealSize)
-    }.flatten
+    }
 
     val scaleImageRequests = {
       var t = 0
--- a/kifi-backend/modules/common/app/com/keepit/common/store/ImagePath.scala
+++ b/kifi-backend/modules/common/app/com/keepit/common/store/ImagePath.scala
@@ -1,0 +1,18 @@
+package com.keepit.common.store
+
+import com.keepit.model._
+import play.api.libs.json._
+
+case class ImagePath(path: String) extends AnyVal {
+  override def toString() = path
+  def getUrl(implicit imageConfig: S3ImageConfig): String = imageConfig.cdnBase + "/" + path
+}
+
+object ImagePath {
+  def apply(prefix: String, hash: ImageHash, size: ImageSize, kind: ProcessImageOperation, format: ImageFormat): ImagePath = {
+    val fileName = hash.hash + "_" + size.width + "x" + size.height + kind.fileNameSuffix + "." + format.value
+    ImagePath(prefix + "/" + fileName)
+  }
+
+  implicit val format: Format[ImagePath] = Format(__.read[String].map(ImagePath(_)), Writes(path => JsString(path.path)))
+}
--- a/kifi-backend/modules/common/app/com/keepit/common/store/OrganizationAvatarStore.scala
+++ b/kifi-backend/modules/common/app/com/keepit/common/store/OrganizationAvatarStore.scala
@@ -46,7 +46,8 @@ class S3OrganizationAvatarStoreImpl @Inject() (
   }
 }
 
-class InMemoryOrganizationAvatarStoreImpl() extends OrganizationAvatarStore {
+@Singleton
+class InMemoryOrganizationAvatarStoreImpl @Inject() () extends OrganizationAvatarStore {
 
   require(!(Play.maybeApplication.isDefined && Play.isProd), "Can't have in memory file store in production")
 
--- a/kifi-backend/modules/common/app/com/keepit/common/store/OrganizationAvatarStore.scala
+++ b/kifi-backend/modules/common/app/com/keepit/common/store/OrganizationAvatarStore.scala
@@ -6,28 +6,26 @@ import com.amazonaws.services.s3.model.ObjectMetadata
 import com.amazonaws.services.s3.transfer.TransferManager
 import com.google.inject.{ Inject, Singleton }
 import com.keepit.common.core._
-import com.keepit.model.{ ImageFormat, ImageHash, ProcessImageOperation }
 import play.api.Play
-import play.api.libs.Files.TemporaryFile
-import play.api.libs.json._
 import play.api.Play.current
+import play.api.libs.Files.TemporaryFile
 
 import scala.collection.mutable
 import scala.concurrent.{ ExecutionContext, Future }
 
-trait RoverImageStore {
+trait OrganizationAvatarStore {
   def put(key: ImagePath, is: InputStream, contentLength: Int, mimeType: String): Future[Unit]
   def get(key: ImagePath): Future[TemporaryFile]
 }
 
-case class RoverImageStoreInbox(dir: File) extends S3InboxDirectory
+case class OrganizationAvatarStoreInbox(dir: File) extends S3InboxDirectory
 
 @Singleton
-class S3RoverImageStoreImpl @Inject() (
+class S3OrganizationAvatarStoreImpl @Inject() (
     s3ImageConfig: S3ImageConfig,
-    val inbox: RoverImageStoreInbox,
+    val inbox: OrganizationAvatarStoreInbox,
     implicit val transferManager: TransferManager,
-    implicit val executionContext: ExecutionContext) extends RoverImageStore with S3AsyncHelper {
+    implicit val executionContext: ExecutionContext) extends OrganizationAvatarStore with S3AsyncHelper {
 
   def put(key: ImagePath, is: InputStream, contentLength: Int, mimeType: String): Future[Unit] = {
     val om = new ObjectMetadata()
@@ -48,7 +46,7 @@ class S3RoverImageStoreImpl @Inject() (
   }
 }
 
-class InMemoryRoverImageStoreImpl() extends RoverImageStore {
+class InMemoryOrganizationAvatarStoreImpl() extends OrganizationAvatarStore {
 
   require(!(Play.maybeApplication.isDefined && Play.isProd), "Can't have in memory file store in production")
 
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationAvatarCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationAvatarCommander.scala
@@ -33,15 +33,9 @@ class OrganizationAvatarCommanderImpl @Inject() (
     val webService: WebService,
     private implicit val executionContext: ExecutionContext) extends OrganizationAvatarCommander with ProcessedImageHelper with Logging {
 
-  def getBestImage(orgId: Id[Organization], imageSize: ImageSize): Option[OrganizationAvatar] = {
-    // Get the image that is the closest to the desired ImageSize, by a sum-of-squares metric
-    def sq(x: Int) = x * x
-    def metric(a: OrganizationAvatar) = sq(a.height - imageSize.height) + sq(a.width - imageSize.width)
-
-    db.readOnlyReplica { implicit session =>
-      val options = orgAvatarRepo.getByOrganization(orgId)
-      options.sortBy(metric).headOption
-    }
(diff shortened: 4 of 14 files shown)


kifi-backend/modules/common/app/com/keepit/common/store/ImagePath.scala
```diff
@@ -0,0 +1,18 @@
...
+}
+
+object ImagePath {
+  def apply(prefix: String, hash: ImageHash, size: ImageSize, kind: ProcessImageOperation, format: ImageFormat): ImagePath = {
+    val fileName = hash.hash + "_" + size.width + "x" + size.height + kind.fileNameSuffix + "." + format.value
+    ImagePath(prefix + "/" + fileName)
+  }
+
+  implicit val format: Format[ImagePath] = Format(__.read[String].map(ImagePath(_)), Writes(path => JsString(path.path)))
+}
```
ryanpbrewster, line 18: Moved this out of `RoverImageStore` because it's useful on its own.

kifi-backend/modules/common/app/com/keepit/common/store/OrganizationAvatarStore.scala
```diff
@@ -0,0 +1,73 @@
...
+  def get(key: ImagePath): Future[TemporaryFile] = {
+    cache.get(key) match {
+      case Some(f) => Future.successful(f)
+      case None => Future.failed(new RuntimeException("Key not there!"))
+    }
+  }
+
+  def all: Map[ImagePath, TemporaryFile] = cache.toMap
+
+}
```
ryanpbrewster, line 73: This whole file was shamelessly stolen from `RoverImageStore`. I think there is some opportunity to factor out some common behavior. I don't really understand why we have different image stores.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationAvatarCommander.scala
```diff
@@ -1,17 +1,185 @@
...
-    None // TODO: do this.
+    // Get the image that is the closest to the desired ImageSize, by a sum-of-squares metric
+    def sq(x: Int) = x * x
+    def metric(a: OrganizationAvatar) = sq(a.height - imageSize.height) + sq(a.width - imageSize.width)
+
+    db.readOnlyReplica { implicit session =>
+      val options = orgAvatarRepo.getByOrganization(orgId)
+      options.sortBy(metric).headOption
+    }
+  }
```
ryanpbrewster, line 39: I don't take into account aspect ratio or anything fancy here. Wonder if that will become necessary.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationAvatarCommander.scala
```diff
@@ -1,17 +1,185 @@
...
+    val (necessary, unnecessary) = determineRequiredProcessImageRequests(sourceImageSize, existingAvatars)
+
+    prepareNewImagesToBePersisted(sourceImage, bufferedSourceImage, existingAvatars, necessary) match {
+      case Left(processingError) => Future.successful(Left(processingError))
+      case Right(processedImagesReadyToPersist) =>
+        val uploads = processedImagesReadyToPersist.map { image =>
+          imageStore.put(image.key, image.is, image.bytes, imageFormatToMimeType(image.format)).imap { _ =>
+            ImageProcessState.UploadedImage(image.key, image.format, image.image, image.processOperation)
+          }
+        }
```
ryanpbrewster, line 68: I took all of those code from `RoverImageFetcher`, but it has been refactored into methods that (mostly) fit on one screen. This `imageStore.put` is one that I didn't touch at all. It passes tests, but does it actually make sense here?

kifi-backend/modules/common/app/com/keepit/common/store/OrganizationAvatarStore.scala
```diff
@@ -0,0 +1,73 @@
...
+  def put(key: ImagePath, is: InputStream, contentLength: Int, mimeType: String): Future[Unit]
+  def get(key: ImagePath): Future[TemporaryFile]
+}
+
+case class OrganizationAvatarStoreInbox(dir: File) extends S3InboxDirectory
+
+@Singleton
+class S3OrganizationAvatarStoreImpl @Inject() (
+    s3ImageConfig: S3ImageConfig,
+    val inbox: OrganizationAvatarStoreInbox,
```
Colin-Lane, line 26: `inbox` is unused?

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationAvatarCommander.scala
```diff
@@ -1,17 +1,185 @@
...
+
+      unnecessary.foreach { processOperationRequest =>
+        val alreadyExistingImageInfo = existingAvatars.find(avatar => avatar.kind == processOperationRequest.operation && avatar.imageSize == processOperationRequest.size).get
+        val orgAvatar = OrganizationAvatar(organizationId = orgId, width = alreadyExistingImageInfo.width, height = alreadyExistingImageInfo.height, format = sourceImage.format, kind = alreadyExistingImageInfo.kind, imagePath = alreadyExistingImageInfo.imagePath, source = UserUpload, sourceFileHash = sourceImage.hash, sourceImageURL = None)
+        orgAvatarRepo.save(orgAvatar)
+      }
+
+      Right(sourceImage.hash)
+    }
+  }
```
ryanpbrewster, line 133: This method is kind of complicated because the original image doesn't fit cleanly into any `ImageProcessRequest` (those deal strictly with `Crop` or `Scale` operations).

Also the `RoverImageFetcher` had a comment saying that this code tended to be prone to race conditions. I don't really see why.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationAvatarCommander.scala
```diff
@@ -1,17 +1,185 @@
...
+
+  // All the ProcessImageRequests in A that ARE in B
+  def intersectProcessImageRequests(A: Set[ProcessImageRequest], B: Set[ProcessImageRequest]): Set[ProcessImageRequest] = {
+    @inline def boundingBox(imageSize: ImageSize): Int = Math.max(imageSize.width, imageSize.height)
+    val Bp = B.collect { case ProcessImageRequest(ProcessImageOperation.Scale, size) => boundingBox(size) }
+    A filter {
+      case ProcessImageRequest(ProcessImageOperation.Scale, size) => Bp.contains(boundingBox(size))
+      case otherRequest => B.contains(otherRequest)
+    }
+  }
```
ryanpbrewster: Both of these methods are simple set operations with a slightly modified definition of equality. Is there a clearer way to express that? `A.diffBy` or something?

kifi-backend/modules/shoebox/app/com/keepit/common/store/ShoeboxProdStoreModule.scala
```diff
@@ -59,6 +60,7 @@ case class ShoeboxDevStoreModule() extends DevStoreModule(ShoeboxProdStoreModule
   def configure() {
     bind[ImageDataIntegrityPlugin].to[ImageDataIntegrityPluginImpl].in[AppScoped]
     bind[RoverImageStore].to[InMemoryRoverImageStoreImpl]
+    bind[OrganizationAvatarStore].to[InMemoryOrganizationAvatarStoreImpl]
```
ryanpbrewster: I don't know why this is here. I tried to follow the example set by `RoverImageStore`, but I already bound `InMemoryOrganizationAvatarStoreImpl` in the `FakeStoreModule`.

Colin-Lane: The `FakeStoreModule` is used by tests. `ShoeboxProdStoreModule` is used in production.

kifi-backend/modules/common/app/com/keepit/common/store/OrganizationAvatarStore.scala
ryanpbrewster, line 26: I think it is used by `S3AsyncHelper`, which `S3OrganizationAvatarStoreImpl` extends.

kifi-backend/modules/shoebox/app/com/keepit/common/store/ShoeboxProdStoreModule.scala
ryanpbrewster: So I shouldn't need to bind the `InMemory` image store, right? I'll try removing it.

kifi-backend/modules/common/app/com/keepit/commanders/ProcessedImageHelper.scala
```diff
@@ -165,9 +165,9 @@ trait ProcessedImageHelper {
 
   // Returns Set of bounding box sizes
   protected def calcSizesForImage(imageSize: ImageSize, scaleCandidates: Seq[ScaledImageSize], cropCandidates: Seq[CroppedImageSize]): Set[ProcessImageRequest] = {
-    val scaleSizes = scaleCandidates.map { size =>
+    val scaleSizes = scaleCandidates.flatMap { size =>
       calcResizeBoundingBox(imageSize, size.idealSize)
-    }.flatten
+    }
```
leogrim, line 180: :+1:

kifi-backend/modules/common/app/com/keepit/common/store/ImagePath.scala
leogrim, line 18: Definitely!

kifi-backend/modules/common/app/com/keepit/common/store/OrganizationAvatarStore.scala
leogrim, line 26: Yep.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationAvatarCommander.scala
```diff
@@ -1,17 +1,185 @@
...
+    imageStore: OrganizationAvatarStore,
+    photoshop: Photoshop,
+    val webService: WebService,
+    private implicit val executionContext: ExecutionContext) extends OrganizationAvatarCommander with ProcessedImageHelper with Logging {
+
   def getBestImage(orgId: Id[Organization], imageSize: ImageSize): Option[OrganizationAvatar] = {
-    None // TODO: do this.
+    // Get the image that is the closest to the desired ImageSize, by a sum-of-squares metric
+    def sq(x: Int) = x * x
+    def metric(a: OrganizationAvatar) = sq(a.height - imageSize.height) + sq(a.width - imageSize.width)
```
leogrim: Check out `BasicImage` and `BasicImages`

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationAvatarCommander.scala
leogrim, line 68: I think so.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationAvatarCommander.scala
```diff
@@ -1,17 +1,185 @@
...
+        val uploadedImagesFut = Future.sequence(uploads).imap(Right(_)).recover { case error: Exception => Left(ImageProcessState.CDNUploadFailed(error)) }
+        uploadedImagesFut.imap {
+          case Left(uploadError) => Left(uploadError)
+          case Right(uploadedImages) =>
+            try {
+              saveNewAvatars(orgId, sourceImage, existingAvatars, uploadedImages, unnecessary)
+              Right(sourceImage.hash)
+            } catch {
+              case imageInfoError: Exception =>
+                log.error(s"Failed to update ImageInfoRepo after fetching image from user uploaded file: $imageInfoError")
```
leogrim: `ImageInfoRepo` leak

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationAvatarCommander.scala
leogrim: Why diverge from `ProcessedImageHelper.filterProcessImageRequests` here?

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationAvatarRepo.scala
```diff
@@ -10,7 +10,9 @@ import org.joda.time.DateTime
 
 @ImplementedBy(classOf[OrganizationAvatarRepoImpl])
 trait OrganizationAvatarRepo extends Repo[OrganizationAvatar] {
-  def getByOrganization(organizationId: Id[Organization])(implicit session: RSession): Seq[OrganizationAvatar]
+  def getByOrganization(organizationId: Id[Organization], state: State[OrganizationAvatar] = OrganizationAvatarStates.ACTIVE)(implicit session: RSession): Seq[OrganizationAvatar]
+  def getByImageHash(hash: ImageHash, state: State[OrganizationAvatar] = OrganizationAvatarStates.ACTIVE)(implicit session: RSession): Seq[OrganizationAvatar]
```
leogrim, line 14: Perhaps you should return something closer to a set of unique `BaseImage` here, that may return a lot of duplicate avatars.

leogrim:
See comments, lgtm otherwise.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationAvatarCommander.scala
ryanpbrewster: It's actually the exact same method, I just renamed it so that it would line up with `diffProcessImageRequests`. Rather, `diff` is the same logic as `filter`.

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationAvatarRepo.scala
ryanpbrewster, line 14: I can easily get the set of unique `ImagePath`s or `(kind, size)` tuples if I just make a `distinct` query, but I'm not sure how easy it would be to get the unique `BaseImage`s.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationAvatarCommander.scala
leogrim: Feel free to rename `filter` to `diff`, so we avoid the code duplication.
I'm not sure I understand why you need both `diff` and `intersect`, clue?

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationAvatarRepo.scala
ryanpbrewster, line 14: Oh, I just looked up what a `BasicImage` is.

leogrim, line 14: `BasicImage` has less stuff than `BaseImage`, not sure if that's what you want here. I guess what I meant is use a `distinct` query to return whatever you need to build your set of existing `ProcessImageRequest`.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationAvatarCommander.scala
ryanpbrewster: I have a set `E` of things that already exist, and set `R` of things that are required. I want `U = R intersect E` for things that are unnecessary to recompute, and `N = R diff U` for the things that are necessary to compute.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationAvatarCommander.scala
leogrim: I think I would maybe return `BasicImages` rather than `OrganizationAvatar`. Then let the consumer use the provided methods to pick up the size. 
If you do want to return `OrganizationAvatar`, you can use `ProcessedImageHelper.pickByIdealImageSize` here.

leogrim: I didn't go full refactor when I introduced `KeepImageCommander.getBasicImagesForKeeps`, but that's pretty much all that's need in practice. Also you can cache `BasicImages` (which is immutable) for the organization rather than trying to cache single images and having to be aware of which size got cached somehow.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationAvatarCommander.scala
leogrim: `N = filter(R, E)` will get you the set 'N' of necessary computations , then you can just use a regular Set diff to compute `U = R -- N`

N may actually be wrong here, because `diff` and `intersect` have special semantics and might not really be reverse of each other.

ryanpbrewster: The special semantics are important here, because it's difficult to look things up in the database. To make this more concrete, consider the following example:

We have an image that is 150x100
We require images that are 150x150 (ends up being 150x100) and 300x300 (ends up being 300x200).

Suppose we already have the 150x100 image computed.

In this case, `R = {150x150, 300x300}`, `E = {150x100}`.
I'd like to end up with `U = {150x100}` and `N = {300x200}` (or `N = {300x300}`, it doesn't matter).

I _need_ `U` in that format because I'm doing a table lookup.

PR #15897, opened 2015-06-23
Adding numLibraries, numConnections, numFollowers to /me

andrewconner:
Checked with clients, we're good to remove old help rank stuff from /me. Should notice a performance improvement.

New fields are for Adam's first feature.

@leogrim @eishay

### DIFF
--- a/integration/site-user_spec.js
+++ b/integration/site-user_spec.js
@@ -28,12 +28,7 @@ frisby.create('get user information for site (api.kifi.com/site/user/me)')
       return Array.isArray(val) && val.every( function (exp) {
         return typeof(exp)==='string';
       })
-    },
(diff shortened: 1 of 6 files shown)


andrewconner:
Merging to get this out to make sure it doesn't break any clients, with a Delighted marketing page update. Feedback appreciated.

kifi-backend/modules/shoebox/marketing
```diff
@@ -1 +1 @@
-Subproject commit 3395dd1ac208c1237910074b1a06fe2d0c54df9d
+Subproject commit 987ec6cf5f85832f315dee692a9ced6a245877c3
```
leogrim, line 1: ?

PR #15898, opened 2015-06-24
Mobile Organization Invite Controller

Colin-Lane:
- universal invite link.
- accepting can accept universal invite link too (still needs testing I think)
  @leogrim @ryanpbrewster

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
@@ -30,7 +30,7 @@ trait OrganizationInviteCommander {
   def acceptInvitation(orgId: Id[Organization], userId: Id[User]): Either[OrganizationFail, OrganizationMembership]
   def declineInvitation(orgId: Id[Organization], userId: Id[User]): Seq[OrganizationInvite]
   // Creates a Universal Invite Link for an organization and inviter. Anyone with the link can join the Organization
-  def universalInviteLink(orgId: Id[Organization], inviterId: Id[User], role: OrganizationRole = OrganizationRole.MEMBER, authToken: Option[String] = None): Either[OrganizationFail, (OrganizationInvite, Organization)]
+  def universalInviteLink(orgId: Id[Organization], inviterId: Id[User], role: OrganizationRole = OrganizationRole.MEMBER): Either[OrganizationFail, OrganizationInvite]
 }
 
 @Singleton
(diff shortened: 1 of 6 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
```diff
@@ -27,10 +27,10 @@ import scala.concurrent.{ ExecutionContext, Future }
 @ImplementedBy(classOf[OrganizationInviteCommanderImpl])
 trait OrganizationInviteCommander {
   def inviteUsersToOrganization(orgId: Id[Organization], inviterId: Id[User], invitees: Seq[OrganizationMemberInvitation])(implicit eventContext: HeimdalContext): Future[Either[OrganizationFail, Seq[(Either[BasicUser, RichContact], OrganizationRole)]]]
-  def acceptInvitation(orgId: Id[Organization], userId: Id[User]): Either[OrganizationFail, OrganizationMembership]
+  def acceptInvitation(orgId: Id[Organization], userId: Id[User], authTokenOpt: Option[String] = None): Either[OrganizationFail, OrganizationMembership]
```
andrewconner: No reason to have a default argument here; easy for forget, and it's always wanted.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
```diff
@@ -27,10 +27,10 @@ import scala.concurrent.{ ExecutionContext, Future }
 @ImplementedBy(classOf[OrganizationInviteCommanderImpl])
 trait OrganizationInviteCommander {
   def inviteUsersToOrganization(orgId: Id[Organization], inviterId: Id[User], invitees: Seq[OrganizationMemberInvitation])(implicit eventContext: HeimdalContext): Future[Either[OrganizationFail, Seq[(Either[BasicUser, RichContact], OrganizationRole)]]]
```
andrewconner: How about calling this `inviteToOrganization`, since it accepts emails?

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
```diff
@@ -27,10 +27,10 @@ import scala.concurrent.{ ExecutionContext, Future }
...
-  def acceptInvitation(orgId: Id[Organization], userId: Id[User]): Either[OrganizationFail, OrganizationMembership]
+  def acceptInvitation(orgId: Id[Organization], userId: Id[User], authTokenOpt: Option[String] = None): Either[OrganizationFail, OrganizationMembership]
   def declineInvitation(orgId: Id[Organization], userId: Id[User]): Seq[OrganizationInvite]
   // Creates a Universal Invite Link for an organization and inviter. Anyone with the link can join the Organization
-  def universalInviteLink(orgId: Id[Organization], inviterId: Id[User], role: OrganizationRole = OrganizationRole.MEMBER, authToken: Option[String] = None): Either[OrganizationFail, (OrganizationInvite, Organization)]
+  def createUniversalInviteLink(orgId: Id[Organization], inviterId: Id[User], role: OrganizationRole = OrganizationRole.MEMBER)(implicit eventContext: HeimdalContext): Either[OrganizationFail, OrganizationInvite]
```
andrewconner: Weird that this doesn't return a link, but is named as if you'd get a link. `universal` is a bit confusing too. How about `createGenericInvite`?

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileOrganizationInviteController.scala
```diff
@@ -20,7 +20,7 @@ import scala.util.{ Failure, Success }
 trait MobileOrganizationInviteController {
   def inviteUsers(pubId: PublicId[Organization]): Action[JsValue]
   def createAnonymousInviteToOrganization(pubId: PublicId[Organization]): Action[JsValue]
-  def acceptInvitation(pubId: PublicId[Organization]): Action[AnyContent]
+  def acceptInvitation(pubId: PublicId[Organization], authTokenOpt: Option[String] = None): Action[AnyContent]
```
andrewconner: No default argument here — _every_ consumer of this controller should be aware of `authToken`s.

andrewconner: Any reason this returns `AnyContent`?

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileOrganizationInviteController.scala
```diff
@@ -20,7 +20,7 @@ import scala.util.{ Failure, Success }
 trait MobileOrganizationInviteController {
```
andrewconner: @leogrim Not sure about the advantage of using traits for controllers. They're the end consumers, and should be rather thin. What do you think? (I know I'm probably super late to this)

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
Colin-Lane: I think that is better as well, I'll make this change.

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileOrganizationInviteController.scala
Colin-Lane: That happened while I was extracting the trait from the controller.

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileOrganizationInviteController.scala
Colin-Lane: Since no one really depends on the controller, there's not much of a reason to have a trait for it. I can remove it.

PR #15910, opened 2015-06-24
OrganizationAccessActions

Colin-Lane:
- 2 simple solutions, lets see which one is liked more.
  @leogrim @ryanpbrewster

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/shoebox/controllers/OrganizationAccessActions.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/shoebox/controllers/OrganizationAccessActions.scala
@@ -9,64 +9,35 @@ import play.api.libs.json.Json
 import play.api.mvc._
 
 import scala.concurrent.Future
+import scala.util.{Failure, Success}
 
 trait OrganizationAccessActions {
   self: UserActions with Controller =>
 
   val publicIdConfig: PublicIdConfiguration
   implicit private val implicitPublicId = publicIdConfig
-  val orgCommander: OrganizationCommander
   val orgMembershipCommander: OrganizationMembershipCommander
 
-  case class OrganizationRequest[T](request: MaybeUserRequest[T], orgId: Id[Organization], authToken: Option[String], permissions: Set[OrganizationPermission]) extends WrappedRequest[T](request) {
-    def userIdOpt = request.userIdOpt
-  }
+  case class OrganizationRequest[T](request: MaybeUserRequest[T], orgId: Id[Organization], authToken: Option[String], permissions: Set[OrganizationPermission]) extends WrappedRequest[T](request)
 
-  def OrganizationAction(id: PublicId[Organization], permissions: OrganizationPermission*) = MaybeUserAction andThen new ActionFunction[MaybeUserRequest, OrganizationRequest] {
+  def OrganizationAction(id: PublicId[Organization], requiredPermissions: OrganizationPermission*) = MaybeUserAction andThen new ActionFunction[MaybeUserRequest, OrganizationRequest] {
     override def invokeBlock[A](request: MaybeUserRequest[A], block: (OrganizationRequest[A]) => Future[Result]): Future[Result] = {
-      parseRequest(id, request) match {
-        case Some((orgId, userIdOpt, authToken)) =>
-          // Right now all organizations are publicly visible
+      Organization.decodePublicId(id) match {
+        case Success(orgId) =>
+          val userIdOpt: Option[Id[User]] = request match {
+            case userRequest: UserRequest[A] => Some(userRequest.userId)
+            case _ => None
+          }
+          val requiredPermissionsSet = requiredPermissions.toSet + OrganizationPermission.VIEW_ORGANIZATION
           val memberPermissions = orgMembershipCommander.getPermissions(orgId, userIdOpt)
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/app/com/keepit/shoebox/controllers/OrganizationAccessActions.scala
```diff
@@ -18,22 +18,43 @@ trait OrganizationAccessActions {
...
+          Future.successful(OrganizationFail.INVALID_PUBLIC_ID.asErrorResponse)
+      }
+    }
+  }
+
+  def OrganizationAccessCheck(id: PublicId[Organization], permissions: OrganizationPermission*) = new ActionFilter[MaybeUserRequest] {
+    def filter[A](input: MaybeUserRequest[A]): Future[Option[Result]] = Future.successful(lookupViewable(id, input, permissions))
+  }
+
+  private def lookupViewable[A](orgPubId: PublicId[Organization], input: MaybeUserRequest[A], permissions: Seq[OrganizationPermission]) = {
```
ryanpbrewster: Can we call these `requiredPermissions`?

kifi-backend/modules/shoebox/app/com/keepit/shoebox/controllers/OrganizationAccessActions.scala
```diff
@@ -18,22 +18,43 @@ trait OrganizationAccessActions {
...
     parseRequest(orgPubId, input) match {
       case Some((orgId, userIdOpt, authToken)) =>
         // Right now all organizations are publicly visible
-        val access = true // TODO: this ought to be removed when Leo gets to OrganizationAccessAction.byPermission
-        if (access) {
+        val memberPermissions = orgMembershipCommander.getPermissions(orgId, userIdOpt)
+        if (permissions forall (memberPermissions.contains)) {
```
ryanpbrewster: I think that Scala has a `subsetOf` function, so `permissions subsetOf memberPermissions`. I had a few issues with type-checking the last time I tried to used it, but maybe it'll work here.

Colin-Lane: permissions is a seq. I could convert to a set and then use subsetOf. I considered doing that, but I am not sure I like it more.

ryanpbrewster:
I prefer the `OrganizationAccessCheck` overall.

kifi-backend/modules/shoebox/app/com/keepit/shoebox/controllers/OrganizationAccessActions.scala
leogrim: +1

kifi-backend/modules/shoebox/app/com/keepit/shoebox/controllers/OrganizationAccessActions.scala
```diff
@@ -18,22 +18,43 @@ trait OrganizationAccessActions {
...
+  case class OrganizationRequest[T](request: MaybeUserRequest[T], orgId: Id[Organization], authToken: Option[String], permissions: Set[OrganizationPermission]) extends WrappedRequest[T](request) {
+    def userIdOpt = request.userIdOpt
   }
 
-  private def lookupViewable[A](orgPubId: PublicId[Organization], input: MaybeUserRequest[A]) = {
+  def OrganizationAction(id: PublicId[Organization], permissions: OrganizationPermission*) = MaybeUserAction andThen new ActionFunction[MaybeUserRequest, OrganizationRequest] {
+    override def invokeBlock[A](request: MaybeUserRequest[A], block: (OrganizationRequest[A]) => Future[Result]): Future[Result] = {
+      parseRequest(id, request) match {
+        case Some((orgId, userIdOpt, authToken)) =>
+          // Right now all organizations are publicly visible
```
leogrim: Let's just always add `VIEW_ORGANIZATION` to the set of explicitly required permissions

kifi-backend/modules/shoebox/app/com/keepit/shoebox/controllers/OrganizationAccessActions.scala
leogrim: Better to make it a `Set` anyway, that's the real semantics (unique + unordered)

kifi-backend/modules/shoebox/app/com/keepit/shoebox/controllers/OrganizationAccessActions.scala
```diff
@@ -18,22 +18,43 @@ trait OrganizationAccessActions {
...
+        val memberPermissions = orgMembershipCommander.getPermissions(orgId, userIdOpt)
+        if (permissions forall (memberPermissions.contains)) {
           None
         } else {
-          Some(Forbidden(Json.obj("error" -> "permission_denied")))
+          Some(OrganizationFail.INSUFFICIENT_PERMISSIONS.asErrorResponse)
         }
       case _ =>
-        Some(BadRequest(Json.obj("error" -> "invalid_id")))
+        Some(OrganizationFail.INVALID_PUBLIC_ID.asErrorResponse)
```
leogrim: Can we move this logic into the main method? There's only one, we don't need to factor it out.

PR #15914, opened 2015-06-24
Adds the ability to have before / after time for keepUpdates endpoint

jaredpetker:


jaredpetker:
@andrewconner

kifi-backend/modules/shoebox/app/com/keepit/model/KeepRepo.scala
```diff
@@ -530,10 +530,20 @@ class KeepRepoImpl @Inject() (
...
+    // if so, default to using beforeTime
+    (beforeTime, afterTime) match {
+      case (Some(before), _) =>
+        sql"""SELECT #$bookmarkColumnOrder FROM bookmark bm WHERE library_id IN (SELECT library_id FROM library_membership WHERE user_id=$userId AND state='active') AND state='active' AND user_id!=$userId AND kept_at<=$before ORDER BY kept_at DESC LIMIT $num;""".as[Keep].list
+      case (None, Some(after)) =>
+        sql"""SELECT #$bookmarkColumnOrder FROM bookmark bm WHERE library_id IN (SELECT library_id FROM library_membership WHERE user_id=$userId AND state='active') AND state='active' AND user_id!=$userId AND kept_at>$after ORDER BY kept_at DESC LIMIT $num;""".as[Keep].list
+      case (None, None) =>
+        sql"""SELECT #$bookmarkColumnOrder FROM bookmark bm WHERE library_id IN (SELECT library_id FROM library_membership WHERE user_id=$userId AND state='active') AND state='active' AND user_id!=$userId ORDER BY kept_at DESC LIMIT $num;""".as[Keep].list
+    }
+
```
leogrim, line 546: I'm not sure we should be injecting Strings directly in these queries.

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileRecommendationsController.scala
```diff
@@ -122,8 +122,8 @@ class MobileRecommendationsController @Inject() (
     }
   }
 
-  def keepUpdates(count: Int, beforeTime: String) = UserAction.async { request =>
-    commander.updatesFromFollowedLibraries(request.userId, count, beforeTime)
+  def keepUpdates(count: Int, beforeTime: Option[String], afterTime: Option[String]) = UserAction.async { request =>
+    commander.updatesFromFollowedLibraries(request.userId, count, beforeTime, afterTime)
```
leogrim, line 126: You should be able to use `beforeTime: Option[Datetime], afterTime: Option[Datetime]` here and in the route, they will be automatically parsed and validated as times. Then you can use these in your database queries :)

kifi-backend/modules/shoebox/app/com/keepit/model/KeepRepo.scala
andrewconner, line 546: I think it's fine, `$` in `sql` interpolated strings isn't an injection. `#$` is. But yeah, you can use `DateTime`s. However, not sure if paginating by time is best, since you can have keeps at the same time (for instance, an import or copying action). Previously, we've done a two step process to paginate, and it works decently. You pass in `beforeId` (or, I suppose, `afterId`, though not sure the use case immediately). Query 1, look up the `keep.id` of that `keep.externalId` (`keepRepo.getOpt(externalId)`). Then, do your where using `id < beforeId` or `id > afterId`.

andrewconner, line 546: We originally paginated by datetime way back when, and had problems. I suppose _now_ is worse because we paginate keeps in libraries just by offsets, but if it's a new endpoint, why not do it better (... and this will be higher traffic, so getting out of sequence is easier).

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileRecommendationsController.scala
jaredpetker, line 126: ahh, nice, still wading through the convertible types

kifi-backend/modules/shoebox/app/com/keepit/model/KeepRepo.scala
jaredpetker, line 546: Hm, let's chat in person real quick so I get it. I guess the point of these were to have before be <= and after be strictly >, where on our side we would make sure we didn't somehow get duplicates when querying on the boundaries.  If what your'e saying also takes care of this well then sounds great.

andrewconner:
See: #16070

PR #15917, opened 2015-06-24
Feature/profile widget

adamjgrant:


kifi-backend/modules/shoebox/angular/dev.html
```diff
@@ -30,4 +30,18 @@
       return s;
     }());
   }
+
+  var ngAppElem = angular.element(document.querySelector('[ng-app]') || document.body);
```
andrewconner, line 34: Wahoo! :)

kifi-backend/modules/shoebox/marketing
```diff
@@ -1 +1 @@
-Subproject commit 987ec6cf5f85832f315dee692a9ced6a245877c3
```
andrewconner: Oo, not sure why this happened (sometimes `git add .` can), but we use git submodules to manage our marketing site (we use contractors, so they have their own repo). This overwrote the existing one. The correct hash is `987ec6cf5f85832f315dee692a9ced6a245877c3`. You can get your branch locally back to "good" with: `git checkout 987ec6cf5f85832f315dee692a9ced6a245877c3`

andrewconner: Sorry, you'll need to `cd kifi-backend/modules/shoebox/marketing` first.

kifi-backend/modules/shoebox/angular/src/common/build-css/typography.styl
```diff
@@ -0,0 +1,28 @@
+heading_xl()
```
andrewconner, line 1: Nice, some style standards. I like.

kifi-backend/modules/shoebox/marketing
adamjgrant: If I just cd into that directory, can I just do a `git reset --hard origin/master` ?

andrewconner: Should work

adamjgrant: Or maybe just `git pull origin master`

kifi-backend/modules/shoebox/angular/src/common/build-css/typography.styl
adamjgrant, line 1: Thanks, I've had a lot of success using this technique on projects in the past, but it only works if you restrict these mixins to font-size, line-height, and maybe margin-bottom–modifying with font-weight, color, etc one-off in the implementation.

adamjgrant:
Okay, I think we're ready to go @andrewconner

andrewconner:
Looks good! Feel free to merge.

kifi-backend/modules/shoebox/angular/src/profile/profileWidget.styl
```diff
@@ -0,0 +1,90 @@
...
+  &-badge
+    display: flex
+    align-items: center
+
+  &-profileImg
+    width: 70px
+    min-width: 70px
+    margin-right: 15px
+
+    img
```
andrewconner, line 16: I've read that broad selectors (like tag selectors) on the right-most side are inefficient because of how CSS is queried (right to left). What do you think? http://stackoverflow.com/questions/5797014/why-do-browsers-match-css-selectors-from-right-to-left

kifi-backend/modules/shoebox/angular/src/profile/profileWidget.tpl.html
```diff
@@ -0,0 +1,45 @@
+<div class="kf-pw">
+
+  <div class="kf-pw-badge">
+    <a href="{{me|profileUrl}}" ng-click="registerEvent('Photo')" class="kf-pw-profileImg">
+      <img src="{{me|pic:100}}" />
+    </a>
+    <aside>
+      <h1>
+        <a href="{{me|profileUrl}}" ng-click="registerEvent('Name')">
+          {{me.firstName}}
```
andrewconner: Should this be first and last?

andrewconner:
Nice!

adamjgrant:
Spoke with Jen and made some subtle design changes. Should be good to go now, @andrewconner. Can I have your re-approval?

adamjgrant:
We basically just added a "+" in front of the notification

andrewconner:
Re-approved!

PR #15918, opened 2015-06-24
Adding `pendingFriendRequests` to `me` object.

andrewconner:
@ajkochanowicz Usually I'm wary of adding a lot to `me`, but this is cached so it's light.


adamjgrant:
:+1:

PR #15923, opened 2015-06-24
Use UrlHash instead of Id[NormalizedURI] for S3 article keys

leogrim:
@andrewconner @eishay 
Double-write for newly fetched articles + copying over existing articles to new keys.
When done, will switch get to using the new keys and delete the old articles.
Easier than I thought, and `BatchProcessingActor` takes out most of the boilerplate for that kind of stuff :)

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/rover/model/ArticleInfo.scala
+++ b/kifi-backend/modules/common/app/com/keepit/rover/model/ArticleInfo.scala
@@ -2,7 +2,7 @@ package com.keepit.rover.model
 
 import com.keepit.common.db.{ VersionNumber, SequenceNumber, Id }
 import com.keepit.common.time.DateTimeJsonFormat
-import com.keepit.model.NormalizedURI
+import com.keepit.model.{ UrlHash, NormalizedURI }
 import com.keepit.rover.article.{ ArticleKind, Article }
 import com.kifi.macros.json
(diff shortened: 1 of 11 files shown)


andrewconner:
:+1:

kifi-backend/modules/common/app/com/keepit/rover/model/ArticleInfo.scala
```diff
@@ -14,17 +14,19 @@ case class ArticleVersion(major: VersionNumber[Article], minor: VersionNumber[Ar
...
   def articleKind: ArticleKind[A]
   def uriId: Id[NormalizedURI]
+  def url: String
+  def urlHash: UrlHash
   def bestVersion: Option[ArticleVersion]
   def latestVersion: Option[ArticleVersion]
   def getLatestKey: Option[ArticleKey[A]] = latestVersion.map(toKey)
   def getBestKey: Option[ArticleKey[A]] = bestVersion.map(toKey)
-  private def toKey(version: ArticleVersion): ArticleKey[A] = ArticleKey(uriId, articleKind, version)
+  private def toKey(version: ArticleVersion): ArticleKey[A] = ArticleKey(uriId, urlHash, articleKind, version)
```
yingjieMiao, line 29: i hope the hash doesn't mess with folder structure.

PR #15924, opened 2015-06-24
add evolutions for org/orgmembership sequence

camhashemi:
@leogrim @ryanpbrewster

### DIFF
--- a/kifi-backend/modules/common/conf/evolutions/shoebox/345.sql
+++ b/kifi-backend/modules/common/conf/evolutions/shoebox/345.sql
@@ -3,15 +3,15 @@
 # --- !Ups
 
 -- MySQL:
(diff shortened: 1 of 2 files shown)


ryanpbrewster:
:+1:

leogrim:
:+1:

camhashemi:
remembered last minute to use BIGINT(20) for ids and sequence numbers.

leogrim:
And I missed it :)

yingjieMiao:
it doesn't really matter though. it only affects display, not storage.

camhashemi:
are evolutions just for record-keeping of the manual SQL queries?

leogrim:
And they're applied in some tests (those that use an application)

PR #15926, opened 2015-06-24
OrganizationFactory helper functions for tests

ryanpbrewster:
@leogrim @Colin-Lane

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/model/OrganizationFactory.scala
+++ b/kifi-backend/modules/common/app/com/keepit/model/OrganizationFactory.scala
@@ -1,0 +1,31 @@
+package com.keepit.model
+import java.util.concurrent.atomic.AtomicLong
+
+import com.keepit.common.db.Id
+import com.keepit.common.mail.EmailAddress
+import org.apache.commons.lang3.RandomStringUtils
+
+import scala.util.Random.nextInt
+
+object OrganizationFactory {
+  private[this] val idx = new AtomicLong(10000 + System.currentTimeMillis() % 100)
+
+  def organization(): PartialOrganization = {
+    new PartialOrganization(Organization(name = RandomStringUtils.random(10), ownerId = Id[User](nextInt(10000)), handle = None))
+  }
+
+  def organizations(count: Int): Seq[PartialOrganization] = List.fill(count)(organization())
+
+  case class PartialOrganization private[OrganizationFactory] (
+      org: Organization,
+      members: Seq[User] = Seq.empty[User],
+      invitedUsers: Seq[User] = Seq.empty[User],
+      invitedEmails: Seq[EmailAddress] = Seq.empty[EmailAddress]) {
+
+    def withName(newName: String) = this.copy(org = org.withName(newName))
+    def withOwner(newOwner: User) = this.copy(org = org.copy(ownerId = newOwner.id.get))
+    def withMembers(newMembers: Seq[User]) = this.copy(members = members ++ newMembers)
+    def withInvitedUsers(newInvitedUsers: Seq[User]) = this.copy(invitedUsers = invitedUsers ++ newInvitedUsers)
+    def withInvitedEmails(newInvitedEmails: Seq[EmailAddress]) = this.copy(invitedEmails = invitedEmails ++ newInvitedEmails)
+  }
+}
--- a/kifi-backend/modules/shoebox/test/com/keepit/model/OrganizationFactoryHelper.scala
+++ b/kifi-backend/modules/shoebox/test/com/keepit/model/OrganizationFactoryHelper.scala
@@ -1,0 +1,30 @@
+package com.keepit.model
+
+import com.google.inject.Injector
+import com.keepit.common.db.slick.DBSession.RWSession
+import com.keepit.model.OrganizationFactory.PartialOrganization
+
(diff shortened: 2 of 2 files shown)


kifi-backend/modules/shoebox/test/com/keepit/model/OrganizationFactoryHelper.scala
```diff
@@ -0,0 +1,27 @@
...
+      partialOrganizations.map(u => new OrganizationPersister(u).saved)
+    }
+  }
+
+  implicit class OrganizationMembershipPersister(membership: OrganizationMembership) {
+    def saved(implicit injector: Injector, session: RWSession): OrganizationMembership = {
+      injector.getInstance(classOf[OrganizationMembershipRepo]).save(membership)
+    }
+  }
+}
```
leogrim, line 30: What about having an `OrganizationTestHelper` that contains both these classes an `OrganizationFactory`? This way we only need to import the content of a single object (as opposed to `OrganizationFactory` and `OrganizationFactoryHelper._` today.)

ryanpbrewster, line 30: I knew I was forgetting something...

kifi-backend/modules/common/app/com/keepit/model/OrganizationFactory.scala
```diff
@@ -0,0 +1,36 @@
...
+import org.apache.commons.lang3.RandomStringUtils
+import org.joda.time.DateTime
+
+import scala.util.Random.nextInt
+
+object OrganizationFactory {
+  private[this] val idx = new AtomicLong(10000 + System.currentTimeMillis() % 100)
+
+  def organization(): PartialOrganization = {
+    new PartialOrganization(Organization(id = Some(Id[Organization](idx.incrementAndGet())), name = RandomStringUtils.random(5), ownerId = Id[User](nextInt(10000)), handle = None))
```
Colin-Lane: Is there any reason to set the `Id`? It just gets set back to None with a `.copy(id = None)`
The only way to enforce the `Id` being set would be to save without an `Id` and then update the `Id`.
`DbRepo.save` chooses to use an `update` instead of an `insert` if the `Id` is already defined.

kifi-backend/modules/shoebox/test/com/keepit/model/OrganizationFactoryHelper.scala
```diff
@@ -0,0 +1,27 @@
...
+
+import com.google.inject.Injector
+import com.keepit.common.db.slick.DBSession.RWSession
+import com.keepit.model.OrganizationFactory.PartialOrganization
+
+object OrganizationFactoryHelper {
+  implicit class OrganizationPersister(partialOrganization: PartialOrganization) {
+    def saved(implicit injector: Injector, session: RWSession): Organization = {
+      val org = injector.getInstance(classOf[OrganizationRepo]).save(partialOrganization.get.copy(id = None))
+      injector.getInstance(classOf[OrganizationMembershipRepo]).save(org.newMembership(org.ownerId, OrganizationRole.OWNER))
```
Colin-Lane, line 11: This feels so strange to me. I get why we did it this way, but I would love to see the  `OrganizationMembership` class have a private constructor so that we cannot instantiate a membership without going through `org.newMembership` to get a new instance.

Since our models and repos all live in `com.keepit.model` I'm wondering if `case class OrganizationMembership private[model] (constructor args) { body...}` might be a good idea.

kifi-backend/modules/shoebox/test/com/keepit/model/OrganizationFactoryHelper.scala
```diff
@@ -0,0 +1,27 @@
...
+      org
+    }
+  }
+
+  implicit class OrganizationsPersister(partialOrganizations: Seq[PartialOrganization]) {
+    def saved(implicit injector: Injector, session: RWSession): Seq[Organization] = {
+      partialOrganizations.map(u => new OrganizationPersister(u).saved)
```
Colin-Lane: `OrganizationPersister` is `implicit`, do we need to explicitly use it here?

kifi-backend/modules/shoebox/test/com/keepit/model/OrganizationFactoryHelper.scala
ryanpbrewster, line 11: That's a good idea. RIght now if you try and instantiate an `OrganizationMembership` directly, you have to supply permissions. And honestly, if you can come up with permissions that you want to set, maybe you should be allowed to.

kifi-backend/modules/shoebox/test/com/keepit/model/OrganizationFactoryHelper.scala
ryanpbrewster: Maybe `partialOrganizations.map(_.saved)` would work? I'll try that.

kifi-backend/modules/common/app/com/keepit/model/OrganizationFactory.scala
ryanpbrewster: Huh, good point. I have no idea why the `UserFactory` did that. I'm going to remove it.

PR #15928, opened 2015-06-24
Remove all try / cache from NormalizedURIInterner

leogrim:
@eishay @andrewconner @stkem YOLO

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/normalizer/NormalizedURIInterner.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/normalizer/NormalizedURIInterner.scala
@@ -7,10 +7,7 @@ import com.keepit.common.db.slick.DBSession.{ RSession, RWSession }
 import scala.concurrent.ExecutionContext
 import scala.util._
 import com.keepit.queue.NormalizationUpdateTask
-import java.sql.SQLException
 import com.keepit.common.logging.{ Timer, Logging }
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/app/com/keepit/normalizer/NormalizedURIInterner.scala
```diff
@@ -88,22 +53,7 @@ class NormalizedURIInterner @Inject() (
...
         }
-        case Failure(ex) =>
-          /**
-           * if we can't parse a url we should not let it get to the db.
-           * its scraping should stop as well and it will be removed from cache.
-           * an error should be thrown so we could examine the problem.
-           * in most cases its a user error so the exceptions may be caught upper in the stack.
-           */
-          val uriCandidate = NormalizedURI.withHash(normalizedUrl = url, normalization = None)
-          normalizedURIRepo.deleteCache(uriCandidate)
```
andrewconner, line 99: This may be useful?

leogrim, line 99: Do we have a reason to believe there's something in the cache? I can't think of a reason to think there might be, if none of us can, I'd like to take it out and put it back if something happens. The idea here is to get back to a clean state, and rebuild necessary exception handling from the ground up, see if it helps :)

PR #15929, opened 2015-06-24
slack: removed uniqueness constraint on name or webhook URL

camhashemi:
@andrewconner before, we were using name as a unique identifier for a subscription. now, people can subscribe to #general from two different slack teams and it should work.

### DIFF
--- a/kifi-backend/modules/common/conf/evolutions/shoebox/347.sql
+++ b/kifi-backend/modules/common/conf/evolutions/shoebox/347.sql
@@ -1,0 +1,9 @@
+# SHOEBOX
+
+# --- !Ups
+
+DROP INDEX library_id ON library_subscription;
+
+insert into evolutions (name, description) values('347.sql', 'removing constraint for unique library subscription name');
+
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/common/conf/evolutions/shoebox/347.sql
```diff
@@ -0,0 +1,9 @@
+# SHOEBOX
+
+# --- !Ups
+
+DROP INDEX library_id ON library_subscription;
```
andrewconner, line 5: Why dropping `library_id`?

camhashemi:
it's not dropping the `library_id` column, but the unique index on  `(library_id, name)` that is basically deprecated from our model. `library_id` was the name of the index.

andrewconner:
Ohhh, I see. Usually our index names are `table_name_i_field1_field2`. Works for me.

camhashemi:
I forgot to name it but made sure I had the name right. Duly noted, thank you.

PR #15930, opened 2015-06-24
Convenience methods in org member commander

ryanpbrewster:
Also fixed an edgecase where users could not remove themselves.
The true owner of an org cannot remove themself. They must delete the org.

@leogrim @Colin-Lane

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
@@ -30,7 +30,9 @@ trait OrganizationMembershipCommander {
   def getMembersAndInvitees(orgId: Id[Organization], limit: Limit, offset: Offset, includeInvitees: Boolean): Seq[MaybeOrganizationMember]
 
   def getPermissions(orgId: Id[Organization], userIdOpt: Option[Id[User]]): Set[OrganizationPermission]
+  def validRequest(request: OrganizationMembershipRequest)(implicit session: RSession): Boolean
 
+  def validateRequests(requests: Seq[OrganizationMembershipRequest]): Map[OrganizationMembershipRequest, Boolean]
   def addMembership(request: OrganizationMembershipAddRequest): Either[OrganizationFail, OrganizationMembershipAddResponse]
   def modifyMembership(request: OrganizationMembershipModifyRequest): Either[OrganizationFail, OrganizationMembershipModifyResponse]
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
```diff
@@ -116,12 +119,19 @@ class OrganizationMembershipCommanderImpl @Inject() (
           targetOpt.exists(_.role < requester.role) && requester.permissions.contains(MODIFY_MEMBERS)
 
         case OrganizationMembershipRemoveRequest(_, _, _) =>
-          targetOpt.exists(_.role < requester.role) &&
-            requester.permissions.contains(REMOVE_MEMBERS)
+          val selfRemove = targetOpt.exists(t => t.userId == requester.userId && t.userId != org.ownerId)
+          val otherRemove = targetOpt.exists(_.role < requester.role) && requester.permissions.contains(REMOVE_MEMBERS)
+          selfRemove || otherRemove
```
Colin-Lane: this doesn't take advantage of `short-circuit evaluation`. Not really necessary to change though.

leogrim: As mentioned somewhere else, can we use verbs for boolean when it makes sense?
e.g. here `canLeave` and `canRemoveOther`

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
```diff
@@ -116,12 +119,19 @@ class OrganizationMembershipCommanderImpl @Inject() (
...
       }
     }
   }
 
+  def validateRequests(requests: Seq[OrganizationMembershipRequest]): Map[OrganizationMembershipRequest, Boolean] = {
+    db.readOnlyReplica { implicit session =>
+      requests.map(r => r -> validRequest(r)).toMap
+    }
+  }
```
Colin-Lane, line 135: `requests.filter(validRequest(_)).map(_ -> true).toMap.withDefaultValue(false)`
Saves memory, and if you missed any `OrganizationMembershipRequest` they will default to invalid.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
ryanpbrewster: Do we have any convention to distinguish values from functions? I had fallen into the habit of "functions as verbs" and "values as nouns", but I see your point regarding Booleans.

PR #15931, opened 2015-06-25
Marketing site update

andrewconner:
@sterlingz


andrewconner:
QAed on the Frogs site

PR #15934, opened 2015-06-25
Make sure we're not dealing with a lazy structure in database session

leogrim:
@andrewconner @eishay @yingjieMiao


andrewconner:
Looks good (reviewed on Kifi chat last night)

PR #15935, opened 2015-06-25
Fixing laziness, finding more?!

andrewconner:
@yingjieMiao @leogrim Found one. This is lazy:

``` scala
scala> Seq(1,2,3,4,5).groupBy(_ < 2).values.toSeq
res1: Seq[Seq[Int]] = Stream(List(2, 3, 4, 5), ?)
```

Airbraking so we find more.


yingjieMiao:
good find!

PR #15936, opened 2015-06-25
Adds ability for library lists to have a specific ordering via LibraryOrdering (mobile only at the moment)

jaredpetker:
- Backwards compatible, old endpoints should not be affected, site is untouched
- The new v2 endpoitn for getting profile libraries can now take a LibraryOrdering, which can be:
  - Alphabetical "alphabetical"
  - LastKeptInto "last_kept_into"
  - MemberCount "member_count"
- Tests are accounted for

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/UserProfileCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/UserProfileCommander.scala
@@ -44,8 +44,12 @@ class UserProfileCommander @Inject() (
     implicit val config: PublicIdConfiguration) {
 
   def getOwnLibrariesForSelf(user: User, page: Paginator, idealSize: ImageSize): ParSeq[LibraryCardInfo] = {
+    getOwnLibrariesForSelf(user, page, idealSize, None)
+  }
+
+  def getOwnLibrariesForSelf(user: User, page: Paginator, idealSize: ImageSize, ordering: Option[LibraryOrdering]): ParSeq[LibraryCardInfo] = {
     val (libraryInfos, memberships) = db.readOnlyMaster { implicit session =>
-      val libs = libraryRepo.getOwnerLibrariesForSelf(user.id.get, page)
+      val libs = libraryRepo.getOwnerLibrariesForSelfWithOrdering(user.id.get, page, ordering)
       val libOwnerIds = libs.map(_.ownerId).toSet
       val owners = basicUserRepo.loadAll(libOwnerIds)
       val libraryIds = libs.map(_.id.get).toSet
@@ -78,13 +82,13 @@ class UserProfileCommander @Inject() (
     }
   }
 
-  def getOwnLibraries(user: User, viewer: Option[User], page: Paginator, idealSize: ImageSize): ParSeq[LibraryCardInfo] = {
+  def getOwnLibraries(user: User, viewer: Option[User], page: Paginator, idealSize: ImageSize, ordering: Option[LibraryOrdering] = None): ParSeq[LibraryCardInfo] = {
     db.readOnlyMaster { implicit session =>
       val libs = viewer match {
         case None =>
-          libraryRepo.getOwnerLibrariesForAnonymous(user.id.get, page)
+          libraryRepo.getOwnerLibrariesForAnonymous(user.id.get, page, ordering)
         case Some(other) =>
-          libraryRepo.getOwnerLibrariesForOtherUser(user.id.get, other.id.get, page)
+          libraryRepo.getOwnerLibrariesForOtherUser(user.id.get, other.id.get, page, ordering)
       }
       val libOwnerIds = libs.map(_.ownerId).toSet
       val owners = basicUserRepo.loadAll(libOwnerIds)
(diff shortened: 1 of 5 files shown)


jaredpetker:
@andrewconner think this is done @leogrim I know you worked on the library filtering for "own/following" if I remember as well

andrewconner:
Nice, will review after getting Adam started for the day

kifi-backend/modules/shoebox/app/com/keepit/commanders/UserProfileCommander.scala
```diff
@@ -44,8 +44,12 @@ class UserProfileCommander @Inject() (
     implicit val config: PublicIdConfiguration) {
 
   def getOwnLibrariesForSelf(user: User, page: Paginator, idealSize: ImageSize): ParSeq[LibraryCardInfo] = {
+    getOwnLibrariesForSelf(user, page, idealSize, None)
+  }
+
+  def getOwnLibrariesForSelf(user: User, page: Paginator, idealSize: ImageSize, ordering: Option[LibraryOrdering]): ParSeq[LibraryCardInfo] = {
```
andrewconner, line 50: You can get rid of the above function and use default arguments:

``` scala
def getOwnLibrariesForSelf(user: User, page: Paginator, idealSize: ImageSize, ordering: Option[LibraryOrdering] = None)
```

tada, backwards compatibility

kifi-backend/modules/shoebox/app/com/keepit/commanders/UserProfileCommander.scala
```diff
@@ -79,12 +83,16 @@ class UserProfileCommander @Inject() (
   }
 
   def getOwnLibraries(user: User, viewer: Option[User], page: Paginator, idealSize: ImageSize): ParSeq[LibraryCardInfo] = {
+    getOwnLibraries(user, viewer, page, idealSize, None)
+  }
+
+  def getOwnLibraries(user: User, viewer: Option[User], page: Paginator, idealSize: ImageSize, ordering: Option[LibraryOrdering]): ParSeq[LibraryCardInfo] = {
```
andrewconner: Same

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryRepo.scala
```diff
@@ -278,15 +281,28 @@ class LibraryRepoImpl @Inject() (
...
+           order by #${getOrderBySql(ordering)} limit ${page.itemsToDrop}, ${page.size}"""
     query.as[Library].list
   }
 
-  def getOwnerLibrariesForOtherUser(userId: Id[User], friendId: Id[User], page: Paginator)(implicit session: RSession): Seq[Library] = {
+  def getOwnerLibrariesForOtherUser(userId: Id[User], friendId: Id[User], page: Paginator, ordering: Option[LibraryOrdering])(implicit session: RSession): Seq[Library] = {
     import com.keepit.common.db.slick.StaticQueryFixed.interpolation
-    val query = sql"select lib.* from library_membership lm, library lib where lm.library_id = lib.id and lm.user_id = $userId and (lm.access='owner' or lm.access='read_write') and lib.state = 'active' and lm.state = 'active' and ((lib.keep_count > 0 and lm.listed and lib.visibility = 'published') or (lib.id in (select lm.library_id from library_membership lm where lm.user_id = $friendId and lm.state = 'active'))) order by lib.member_count desc, lib.last_kept desc, lib.id desc limit ${page.itemsToDrop}, ${page.size}"
+    val query =
+      sql"""select lib.* from library_membership lm, library lib
```
andrewconner, line 299: Much more readable, thanks

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryRepo.scala
```diff
@@ -434,6 +488,42 @@ class LibrarySequencingPluginImpl @Inject() (
   override val interval: FiniteDuration = 20.seconds
 }
 
+/*
+    Does as it it says, a class making it more typesafe to refer
+    to a specific ordering of libraries, currently there exist 3:
+    - alphabetical, last kept into, member count
```
andrewconner, line 494: Curious, should this include a `direction` too? Or too many?

andrewconner:
This is awesome, nice job.

andrewconner:
Merge whenever you're comfortable, can help test.

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryRepo.scala
jaredpetker, line 494: Yeaa, we were thinking about it, I don't think it's too many really, but not something we really need at the moment, adding alternate orderings with direction like "reverse_alphabetical" should be straightforward if needed at some point

kifi-backend/modules/shoebox/app/com/keepit/commanders/UserProfileCommander.scala
jaredpetker, line 50: ah, duh

eishay:
wow! another Scala guy on the team!

jaredpetker:
@andrewconner would be good to know when next deploy occurs so we can make sure all is well.

PR #15937, opened 2015-06-25
[Do not merge] Org member commander convenience functions

ryanpbrewster:
@Colin-Lane @leogrim 

May be overengineered. The API is fairly clean, though.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
@@ -33,9 +33,14 @@ trait OrganizationMembershipCommander {
   def isValidRequest(request: OrganizationMembershipRequest)(implicit session: RSession): Boolean
 
   def validateRequests(requests: Seq[OrganizationMembershipRequest]): Map[OrganizationMembershipRequest, Boolean]
+
   def addMembership(request: OrganizationMembershipAddRequest): Either[OrganizationFail, OrganizationMembershipAddResponse]
   def modifyMembership(request: OrganizationMembershipModifyRequest): Either[OrganizationFail, OrganizationMembershipModifyResponse]
   def removeMembership(request: OrganizationMembershipRemoveRequest): Either[OrganizationFail, OrganizationMembershipRemoveResponse]
+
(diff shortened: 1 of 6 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
```diff
@@ -134,39 +140,85 @@ class OrganizationMembershipCommanderImpl @Inject() (
...
-    db.readWrite { implicit session =>
-      if (isValidRequest(request)) {
-        val membership = organizationMembershipRepo.getByOrgIdAndUserId(request.orgId, request.targetId).get
-        val org = organizationRepo.get(request.orgId)
-        val newMembership = organizationMembershipRepo.save(org.modifiedMembership(membership, request.newRole))
-        Right(OrganizationMembershipModifyResponse(request, newMembership))
-      } else {
-        Left(OrganizationFail.INSUFFICIENT_PERMISSIONS)
+  def foldWithFail[A, B, F](f: (A => Either[F, B]), xs: Seq[A]): Map[A, B] = {
+    def loop(ys: Seq[A], accum: Map[A, B]): Map[A, B] = {
```
yingjieMiao: you may be able to have a third 'flag'  parameter, which helps you to exit early from recursion. so you don't have to throw exception to exit early.

ryanpbrewster: Previously I had it that way.

However, I want to take advantage of the TransactionalDB ability to unroll any changes it made, which (I think) requires me to throw an exception. Is that true?

Colin-Lane: It looks like loop calls itself. To prevent potential stack overflows, is this function `@tailrec` optmized?

ryanpbrewster: I think if it's going to throw an exception anyways, I may be able to do this more cleanly using a fold.
`foldWithFail(f: (A => Either[B,F]), xs: Seq[A]): Map[A,B] = {
    def g(accum: Map[A,B], x: A) = f(x) match {
        case Left(_) => throw new Exception
        case Right(y) => accum + (x -> y)
    }
    xs.foldLeft(Map[A,B]())(g)
}`

PR #15938, opened 2015-06-25
Replacing textarea account removal with a link to support.kifi.com

adamjgrant:


kifi-backend/modules/shoebox/angular/src/profile/profile.js
```diff
@@ -272,35 +272,13 @@ angular.module('kifi')
...
-
-            $analytics.eventTrack('user_clicked_page', {
-              'action': 'clickCloseAccount',
-              'path': $location.path()
-            });
-          }, function () {
-            scope.closeAccountStatus = 'error';
+          $analytics.eventTrack('user_clicked_page', {
+            'action': 'clickCloseAccount',
+            'path': $location.path()
```
adamjgrant, line 278: Leaving this in for now, but I'm going to check with Dafna or Jen to see if they still need this. Might be still good to track.

adamjgrant:
Spoke with Dafne and we're good to review, @andrewconner

adamjgrant:
Oops, forgot to bind the ng-click event...please hold @andrewconner

andrewconner:
Looks good after you have the click event

adamjgrant:
Cool. Mergin.

PR #15939, opened 2015-06-25
Invitation Status split from State

Colin-Lane:
- I don’t know how I only needed to change 5 files. and no tests broke.
  We got lucky.

I might be missing a thing or two, but all the tests pass, and everything compiles.

@leogrim @ryanpbrewster


Colin-Lane:
I'll need to apply the evolution first before committing this. Since the table has no entries yet, there should be no overhead.

kifi-backend/modules/common/conf/evolutions/shoebox/335.sql
```diff
@@ -42,6 +42,7 @@ CREATE TABLE organization_invite (
 	created_at datetime NOT NULL,
 	updated_at datetime NOT NULL,
 	state varchar(20) NOT NULL,
+	status varchar(8) DEFAULT NULL,
```
Colin-Lane: we seem to use `varchar`s for a lot of state. I decided to use `varchar` over `boolean` here to be consistent with what's been done elsewhere.

Colin-Lane: I'll also need to apply this manually as this is editing a previous evolution.
`ALTER TABLE "organization_invite" ADD COLUMN status varchar(8) DEFAULT NULL AFTER state`

ryanpbrewster: How about char(8)?

Colin-Lane: yeah, I can make that change.

ryanpbrewster: Also, can we call it "decision_status" or something instead of "status"? "state" and "status" are almost the same word and it's really unclear (to a person reading the code for the first time) what each one is supposed to represent.

Colin-Lane: I think `decision` sounds good as well.

Colin-Lane:
Going to merge this soon if there are no issues.

ryanpbrewster:
:+1:

leogrim:
Or 'response'

PR #15940, opened 2015-06-25
Batch methods for org membership commander

ryanpbrewster:
It can perform a sequence of add/modify/remove operations in a single session
and if it encounters an error it will unroll all the changes and return an
OrganizationFail

Simpler to have separate PRs for this and the controller work

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
@@ -33,9 +33,14 @@ trait OrganizationMembershipCommander {
   def isValidRequest(request: OrganizationMembershipRequest)(implicit session: RSession): Boolean
 
   def validateRequests(requests: Seq[OrganizationMembershipRequest]): Map[OrganizationMembershipRequest, Boolean]
+
   def addMembership(request: OrganizationMembershipAddRequest): Either[OrganizationFail, OrganizationMembershipAddResponse]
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
```diff
@@ -134,39 +140,78 @@ class OrganizationMembershipCommanderImpl @Inject() (
...
+  def modifyMembership(request: OrganizationMembershipModifyRequest): Either[OrganizationFail, OrganizationMembershipModifyResponse] =
+    db.readWrite { implicit session => modifyMembershipHelper(request) }
+
+  def removeMembership(request: OrganizationMembershipRemoveRequest): Either[OrganizationFail, OrganizationMembershipRemoveResponse] =
+    db.readWrite { implicit session => removeMembershipHelper(request) }
+
+  // Accumulates the A -> B results in a Map. If any of the results fail, throw an exception
+  def accumulateWithFail[A, B, F](f: (A => Either[F, B]), xs: Seq[A]): Map[A, B] = {
+    def g(accum: Map[A, B], x: A) = f(x) match {
+      case Left(failure) => throw new Exception
```
Colin-Lane: We have the `OrganizationFail` here, but then we just throw an exception. Is there a way to return this that would make sense so the caller can decide what to do with it rather than just throwing a generic exception?

ryanpbrewster: Trying a few things.

PR #15955, opened 2015-06-25
Better push notification text

andrewconner:
@jaredpetker @eishay


andrewconner:
Wording feedback appreciated. I don't like the "New keep" without context.

PR #15957, opened 2015-06-26
[ok to merge] added support for domain in graph

camhashemi:
@leogrim 

@yingjieMiao dependent on a db evolution to add isEmailProvider to the domain table [edit: evolution has been applied]


kifi-backend/modules/graph/app/com/keepit/graph/manager/GraphUpdater.scala
```diff
@@ -168,16 +168,36 @@ class GraphUpdaterImpl @Inject() () extends GraphUpdater with Logging {
...
 
   private def processNormalizedUriGraphUpdate(update: NormalizedUriGraphUpdate)(implicit writer: GraphWriter) = update.state match {
-    case NormalizedURIStates.INACTIVE | NormalizedURIStates.REDIRECTED => writer.removeVertexIfExists(update.id)
-    case _ => writer.saveVertex(UriData(update.id))
+    case NormalizedURIStates.INACTIVE | NormalizedURIStates.REDIRECTED => {
+      writer.removeVertexIfExists(update.id)
+      update.domainId.foreach { domainId =>
+        writer.removeEdgeIfExists(update.id, domainId, EmptyEdgeReader)
+        writer.removeEdgeIfExists(domainId, update.id, EmptyEdgeReader)
+      }
```
camhashemi, line 176: Since domain updates piggyback on URI and EmailAccounts, there's no place (yet) where I remove a Domain vertex.

In order to do so, I can pass the State[Domain] through to URI and Email updates. Or maybe my design in general could be improved such that domain updates are less coupled with others.

kifi-backend/modules/common/app/com/keepit/shoebox/ShoeboxServiceClient.scala
```diff
@@ -784,4 +786,9 @@ class ShoeboxServiceClientImpl @Inject() (
   def getIngestableUserIpAddresses(seqNum: SequenceNumber[IngestableUserIpAddress], fetchSize: Int): Future[Seq[IngestableUserIpAddress]] = {
     call(Shoebox.internal.getIngestableUserIpAddresses(seqNum, fetchSize)).map { _.json.as[Seq[IngestableUserIpAddress]] }
   }
+
+  def getDomainIdsByDomainNames(domainNames: Seq[String]): Future[Seq[Option[Id[Domain]]]] = {
+    val payload = Json.toJson(domainNames)
+    call(Shoebox.internal.getDomainIdsByDomainNames(), payload).map { _.json.as[Seq[Option[Id[Domain]]]] }
```
yingjieMiao: One thing to consider is add `routingStrategy = offlinePriority` to the call to shoebox. So that our web servers (e.g. those shoeboxes are `up`) don't get overloaded by these expensive and non-user-facing calls.

camhashemi: That's a great idea, wasn't aware of it.

kifi-backend/modules/graph/app/com/keepit/graph/manager/GraphUpdateFetcher.scala
```diff
@@ -64,9 +67,19 @@ class GraphUpdateFetcherImpl @Inject() (
         }
       }
 
-      case NormalizedUriGraphUpdate => shoebox.getIndexableUris(seq.copy(), fetchSize).imap(_.map(NormalizedUriGraphUpdate.apply))
+      case NormalizedUriGraphUpdate => {
+        shoebox.getIndexableUris(seq.copy(), fetchSize).flatMap { indexableUris: Seq[IndexableUri] =>
+          val domainNames: Seq[String] = indexableUris.map { indexableUri => indexableUri.url.split("/")(2) } // fetches domainName for any URI matching http(s?)://{domainName}{/omit/this/stuff/}
```
yingjieMiao: `URI` object has a `parseDomain` method.

kifi-backend/modules/graph/app/com/keepit/graph/manager/GraphUpdateFetcher.scala
```diff
@@ -64,9 +67,19 @@ class GraphUpdateFetcherImpl @Inject() (
...
+          fDomainIds.imap { domainIds => indexableUris.zip(domainIds).map { case (indexableUri, domainId) => indexableUri.copy(domainId = domainId) }.map(NormalizedUriGraphUpdate.apply) }
+        }
+      }
 
-      case EmailAccountGraphUpdate => abook.getEmailAccountsChanged(seq.copy(), fetchSize).imap(_.map(EmailAccountGraphUpdate.apply))
+      case EmailAccountGraphUpdate => abook.getEmailAccountsChanged(seq.copy(), fetchSize).flatMap { emailInfos: Seq[EmailAccountInfo] =>
+        val domainNames: Seq[String] = emailInfos.map { info => info.address.address.split("@")(1) }
+        val fDomainIds = shoebox.getDomainIdsByDomainNames(domainNames)
+        fDomainIds.imap { domainIds => emailInfos.zip(domainIds).map { case (emailInfo, domainId) => emailInfo.copy(domainId = domainId) }.map(EmailAccountGraphUpdate.apply) }
+      }
```
yingjieMiao, line 93: I would be extra cautious here to check `assert(domainIds.size == emailInfos.size)`, or have the shoebox call return a map.

kifi-backend/modules/shoebox/app/com/keepit/controllers/internal/ShoeboxDataPipeController.scala
```diff
@@ -266,4 +268,20 @@ class ShoeboxDataPipeController @Inject() (
       Ok(Json.toJson(ipAddresses))
     }
   }
+
+  def getDomainIdsByDomainNames() = Action.async(parse.json) { request =>
+    SafeFuture {
```
yingjieMiao, line 273: as discussed in person, it's potentially more efficient to look up a set of domain names, instead of seq. (as email domains collide a lot)

kifi-backend/modules/common/app/com/keepit/classify/Domain.scala
```diff
@@ -33,6 +34,7 @@ object Domain {
     (__ \ 'hostname).format[String] and
     (__ \ 'autoSensitive).formatNullable[Boolean] and
     (__ \ 'manualSensitive).formatNullable[Boolean] and
+    (__ \ 'isEmailProvider).format[Boolean] and
```
yingjieMiao, line 39: there is a cache on domain. The cached object (json) does not have the new field. After deploy, on deserialization from cache, this will throw exception. You can bump up the domainKey.version.

kifi-backend/modules/shoebox/app/com/keepit/classify/DomainRepo.scala
```diff
@@ -37,7 +37,8 @@ class DomainRepoImpl @Inject() (
     def autoSensitive = column[Option[Boolean]]("auto_sensitive", O.Nullable)
     def manualSensitive = column[Option[Boolean]]("manual_sensitive", O.Nullable)
     def hostname = column[String]("hostname", O.NotNull)
-    def * = (id.?, hostname, autoSensitive, manualSensitive, state, createdAt, updatedAt) <> ((Domain.apply _).tupled, Domain.unapply _)
+    def isEmailProvider = column[Boolean]("is_email_provider", O.NotNull)
+    def * = (id.?, hostname, autoSensitive, manualSensitive, isEmailProvider, state, createdAt, updatedAt) <> ((Domain.apply _).tupled, Domain.unapply _)
```
yingjieMiao, line 41: is there an evolution script?

kifi-backend/modules/graph/app/com/keepit/graph/manager/GraphUpdater.scala
yingjieMiao, line 176: passing `State[Domain]` won't help. All domain have same state `active`. A more interesting problem is orphan domains that have no connecting edges. I am not sure if graph automatically takes care of that. @leogrim

kifi-backend/modules/shoebox/app/com/keepit/classify/DomainRepo.scala
camhashemi, line 41: not yet.

kifi-backend/modules/graph/app/com/keepit/graph/manager/GraphUpdater.scala
camhashemi, line 176: @leogrim does graph do garbage collection?

leogrim, line 176: No, let's not care about that for domains. It's not like they come and go from the system, the connections between domains, emails, urls are definitive (structural) by definition.

kifi-backend/modules/shoebox/app/com/keepit/controllers/internal/ShoeboxDataPipeController.scala
leogrim, line 273: +1

kifi-backend/modules/common/app/com/keepit/shoebox/ShoeboxServiceClient.scala
```diff
@@ -125,6 +126,7 @@ trait ShoeboxServiceClient extends ServiceClient {
   def getIngestableOrganizations(seqNum: SequenceNumber[Organization], fetchSize: Int): Future[Seq[IngestableOrganization]]
   def getIngestableOrganizationMemberships(seqNum: SequenceNumber[OrganizationMembership], fetchSize: Int): Future[Seq[IngestableOrganizationMembership]]
   def getIngestableUserIpAddresses(seqNum: SequenceNumber[IngestableUserIpAddress], fetchSize: Int): Future[Seq[IngestableUserIpAddress]]
+  def getDomainIdsByDomainNames(domainNames: Seq[String]): Future[Map[String, Option[Id[Domain]]]]
```
leogrim: You're probably interested in whether they're an email provider or not isn't it?
Also you may want to `intern` here, so that if the domain is not in the table yet, it gets created.

camhashemi: It is interning, I just didn't name it that way because the terminology is weird to me. I will refactor.

leogrim: If it is, why return a Map of Options?

camhashemi: If `domain.isEmailProvider == true`, we return None for the domainId. GraphUpdater will ignore appropriately.

leogrim: Then the API is very deceiving, you should be explicit about this: either return a `Boolean` with the domainId or have the API take an explicit parameter `excludeEmailProviders: Boolean` then drop the relevant keys.

camhashemi: Ah, thank you. Makes sense, I'm building selfishly here.

camhashemi:
@leogrim @yingjieMiao updated. new logic is in GraphUpdateFetcher and ShoeboxDataPipeController.

yingjieMiao:
comments on code style. lgtm otherwise.

kifi-backend/modules/common/app/com/keepit/abook/model/RichContact.scala
```diff
@@ -9,7 +10,9 @@ import play.api.libs.functional.syntax._
 
 @json case class RichContact(email: EmailAddress, name: Option[String] = None, firstName: Option[String] = None, lastName: Option[String] = None, userId: Option[Id[User]] = None)
 
-case class EmailAccountInfo(emailAccountId: Id[EmailAccountInfo], address: EmailAddress, userId: Option[Id[User]], verified: Boolean, seq: SequenceNumber[EmailAccountInfo])
+case class EmailAccountInfo(emailAccountId: Id[EmailAccountInfo], address: EmailAddress, userId: Option[Id[User]], verified: Boolean, seq: SequenceNumber[EmailAccountInfo]) extends IngestableWithDomain {
+  def getDomainName = address.address.split("@")(1)
```
leogrim, line 13: `.last.toLowerCase` just in case?

kifi-backend/modules/common/app/com/keepit/shoebox/ShoeboxServiceClient.scala
```diff
@@ -125,6 +126,7 @@ trait ShoeboxServiceClient extends ServiceClient {
   def getIngestableOrganizations(seqNum: SequenceNumber[Organization], fetchSize: Int): Future[Seq[IngestableOrganization]]
   def getIngestableOrganizationMemberships(seqNum: SequenceNumber[OrganizationMembership], fetchSize: Int): Future[Seq[IngestableOrganizationMembership]]
   def getIngestableUserIpAddresses(seqNum: SequenceNumber[IngestableUserIpAddress], fetchSize: Int): Future[Seq[IngestableUserIpAddress]]
+  def internDomainsByDomainNames(domainNames: Seq[String]): Future[Seq[DomainInfo]]
```
leogrim: Use `Set` and `Map` instead?

kifi-backend/modules/common/conf/evolutions/shoebox/349.sql
```diff
@@ -0,0 +1,13 @@
+# SHOEBOX
+
+# --- !Ups
+
+-- MySQL:
+-- ALTER TABLE domain ADD is_email_provider tinyint(1) DEFAULT 0;
+-- UPDATE domain SET is_email_provider = 1 WHERE hostname IN ${ set of email providers };
+
+ALTER TABLE domain ADD is_email_provider tinyint(1) DEFAULT 0;
```
leogrim, line 9: @andrewconner did you mention something like we should use varchar everywhere?

kifi-backend/modules/graph/app/com/keepit/graph/manager/GraphUpdateFetcher.scala
```diff
@@ -64,9 +68,45 @@ class GraphUpdateFetcherImpl @Inject() (
...
+        shoebox.getIndexableUris(seq.copy(), fetchSize).flatMap { indexableUris =>
+          internDomainsByDomainNames(indexableUris).imap { domainInfos: Seq[DomainInfo] =>
+            val domainIdOpts = indexableUris.map { uri =>
+              domainInfos.find { domain =>
+                uri.getDomainName == domain.hostname && !domain.isEmailProvider
+              } match {
+                case Some(domain: DomainInfo) => domain.id
+                case None => None
+              }
+            }
```
leogrim: Here's another reason to use a `Map` in the API, this is not pretty :)

Overall, I'm not a fan of APIs that rely on ordered parameters, better to rely on key-value structures, they're more flexible.

kifi-backend/modules/common/app/com/keepit/classify/Domain.scala
```diff
@@ -44,10 +48,20 @@ object Domain {
...
+@json
+case class DomainInfo(
+  id: Option[Id[Domain]],
+  hostname: String,
+  isEmailProvider: Boolean)
+
+trait IngestableWithDomain {
+  def getDomainName: String
+}
+
```
leogrim: From what I've seen below, I'm not sure there's much benefit to this trait, and it couples unrelated classes.

kifi-backend/modules/graph/app/com/keepit/graph/manager/GraphUpdater.scala
```diff
@@ -168,16 +168,36 @@ class GraphUpdaterImpl @Inject() () extends GraphUpdater with Logging {
...
       writer.saveVertex(UserData(userId))
+
       writer.saveEdge(userId, update.emailAccountId, EmptyEdgeData)
       writer.saveEdge(update.emailAccountId, userId, EmptyEdgeData)
+
+      update.domainId.foreach { domainId =>
+        writer.saveVertex(DomainData(domainId))
+        writer.saveEdge(update.emailAccountId, domainId, EmptyEdgeData)
+        writer.saveEdge(domainId, update.emailAccountId, EmptyEdgeData)
+      }
```
leogrim, line 200: :+1:

kifi-backend/modules/shoebox/app/com/keepit/controllers/internal/ShoeboxDataPipeController.scala
```diff
@@ -266,4 +269,20 @@ class ShoeboxDataPipeController @Inject() (
...
+  def internDomainsByDomainNames() = Action.async(parse.json) { request =>
+    SafeFuture {
+      val domainNames = (request.body \ "domainNames").as[Seq[String]]
+
+      val domainInfos: Seq[DomainInfo] = db.readWrite { implicit session =>
+        domainNames.map { name =>
+          (name, domainRepo.get(name, excludeState = None))
+        }.map {
+          case (_, Some(domain: Domain)) => domain.toDomainInfo
+          case (name: String, None) => domainRepo.save(Domain(hostname = name)).toDomainInfo
```
leogrim: This is bad, you shouldn't be touching the database in a controller, make a proper repo method for it, such as `internByNames`

Note that such a method should also double check the state of the `Domain` model, here you could be returning `Inactive` ones.

camhashemi: I was thinking to do that, thank you.

kifi-backend/modules/graph/app/com/keepit/graph/manager/GraphUpdateFetcher.scala
camhashemi: Makes sense.

camhashemi:
@leogrim @yingjieMiao thank you guys

kifi-backend/modules/shoebox/app/com/keepit/classify/DomainRepo.scala
```diff
@@ -60,4 +64,16 @@ class DomainRepoImpl @Inject() (
     domainIds foreach { id => invalidateCache(get(id)) }
     count
   }
+
+  def internAllByNames(domainNames: Set[String])(implicit session: RWSession): Map[String, Domain] = {
+    val foundDomains = (for (d <- rows if d.hostname.inSet(domainNames))
+      yield (d.hostname -> d)).toMap
+
+    val domainsToSave = (foundDomains.keys.toSet).diff(domainNames)
```
yingjieMiao: `domainNames.diff(foundDomains.keySet)`

camhashemi: updated, thank you.

kifi-backend/modules/shoebox/app/com/keepit/classify/DomainRepo.scala
```diff
@@ -60,4 +64,16 @@ class DomainRepoImpl @Inject() (
...
+    val savedDomains = domainsToSave.map { domainName =>
+      (domainName -> super.save(Domain(hostname = domainName)))
```
yingjieMiao: I don't think it's good to call `super.save`. if we ever overwrite the `save` method, this line will do different things.

kifi-backend/modules/shoebox/app/com/keepit/controllers/internal/ShoeboxDataPipeController.scala
```diff
@@ -266,4 +269,15 @@ class ShoeboxDataPipeController @Inject() (
       Ok(Json.toJson(ipAddresses))
     }
   }
+
+  def internDomainsByDomainNames() = Action.async(parse.json) { request =>
+    SafeFuture {
+      val domainNames = (request.body \ "domainNames").as[Set[String]]
+
+      val domainInfoByName: Map[String, DomainInfo] = db.readWrite { implicit session =>
+        domainRepo.internAllByNames(domainNames).map { case (name: String, domain: Domain) => (name, domain.toDomainInfo) }
```
yingjieMiao: you could call `.mapValues {_.toDomainInfo}`

yingjieMiao:
lgtm!

kifi-backend/modules/shoebox/app/com/keepit/classify/DomainRepo.scala
```diff
@@ -7,12 +7,15 @@ import com.keepit.common.db.{ Id, State }
 import com.keepit.common.db.slick.DBSession.{ RWSession, RSession }
 import com.keepit.common.db.slick.DBSession
 
+import scala.collection.parallel.mutable
```
yingjieMiao: is this used anywhere? and `parallel.mutable`?

camhashemi: no, will remove. thank you.

kifi-backend/modules/shoebox/app/com/keepit/controllers/internal/ShoeboxDataPipeController.scala
camhashemi: sweet

PR #15958, opened 2015-06-26
Introduce LibrarySpace

leogrim:
@andrewconner @Colin-Lane @ryanpbrewster

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/model/Library.scala
+++ b/kifi-backend/modules/common/app/com/keepit/model/Library.scala
@@ -59,6 +59,8 @@ case class Library(
   }
   val isPublished: Boolean = visibility == LibraryVisibility.PUBLISHED
   val isSecret: Boolean = visibility == LibraryVisibility.SECRET
+
+  def space: LibrarySpace = organizationId.map(LibrarySpace.OrganizationSpace(_)) getOrElse LibrarySpace.UserSpace(ownerId)
 }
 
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/common/app/com/keepit/model/Library.scala
```diff
@@ -349,3 +351,18 @@ object LibraryColor {
   }
 
 }
+
+sealed trait LibrarySpace
+
+object LibrarySpace {
+  case class UserSpace(id: Id[User]) extends LibrarySpace
+  case class OrganizationSpace(id: Id[Organization]) extends LibrarySpace
```
Colin-Lane, line 359: should these classes be final?

kifi-backend/modules/common/app/com/keepit/model/Library.scala
```diff
@@ -349,3 +351,18 @@ object LibraryColor {
...
+
+  implicit def fromUserId(userId: Id[User]) = UserSpace(userId)
+  implicit def fromOrganizationId(organizationId: Id[Organization]) = OrganizationSpace(organizationId)
+
+  def prettyPrint(space: LibrarySpace): String = space match {
```
Colin-Lane, line 364: Is this going to be used by something? or just tests?

leogrim, line 364: It's useful in logging

PR #15961, opened 2015-06-26
OrganizationController can create organizations

ryanpbrewster:
@leogrim @Colin-Lane 

I think that once we get OrganizationView working, we should
pass back `Ok(Json.toJson(OrganiztionView(response.newOrg)))`
so the client can immediately see their new organization.


ryanpbrewster:
@Colin-Lane, if you want to merge this before you start OrgView tomorrow we can avoid a merge conflict.

PR #15963, opened 2015-06-26
Fixing keep visibility inefficiency, making EventListener not cause problems later.

andrewconner:
@leogrim @Colin-Lane

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
@@ -30,6 +30,7 @@ import play.api.http.Status._
 import play.api.libs.functional.syntax._
 import play.api.libs.json._
 import views.html.admin.{ libraries, library }
+import com.keepit.common.core._
 
 import scala.collection.parallel.ParSeq
 import scala.concurrent._
@@ -651,18 +652,8 @@ class LibraryCommanderImpl @Inject() (
         val newColor = modifyReq.color.orElse(targetLib.color)
         val newListed = modifyReq.listed.getOrElse(targetMembership.listed)
         val newInviteToCollab = modifyReq.whoCanInvite.orElse(targetLib.whoCanInvite)
-        Future {
-          val keeps = db.readOnlyMaster { implicit s =>
-            keepRepo.getByLibrary(libraryId, 0, Int.MaxValue, Set.empty)
-          }
-          if (keeps.nonEmpty) {
-            db.readWriteBatch(keeps) { (s, k) =>
-              keepRepo.save(k.copy(visibility = newVisibility))(s)
-            }
-            searchClient.updateKeepIndex()
-          }
-        }
 
+        // New library subscriptions
         newSubKeysOpt match {
           case Some(newSubKeys) => db.readWrite { implicit s =>
             librarySubscriptionCommander.updateSubsByLibIdAndKey(targetLib.id.get, newSubKeys)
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -683,6 +674,28 @@ class LibraryCommanderImpl @Inject() (
           libraryRepo.save(targetLib.copy(name = newName, slug = newSlug, visibility = newVisibility, description = newDescription, color = newColor, whoCanInvite = newInviteToCollab, state = LibraryStates.ACTIVE))
         }
 
+        // Update visibility of keeps
+        def updateKeepVisibility(cnt: Int): Future[Unit] = Future {
```
andrewconner: This is less elegant than I would have hoped, but previously, we were making two errors: a) loading _all_ keeps to update their visibility (which could take along time, and introduce races since we're saving objects), b) doing it all at once. This change makes it happen only to the ones who need it, 1000 at a time. The downside is that introduces a potential infinite loop. Feedback appreciated.

kifi-backend/modules/shoebox/app/com/keepit/common/analytics/EventListener.scala
```diff
@@ -51,7 +51,7 @@ class SliderShownListener @Inject() (
     case Event(_, UserEventMetadata(EventFamilies.SLIDER, "sliderShown", externalUser, _, experiments, metaData, _), _, _) =>
       val (user, normUri) = db.readWrite(attempts = 2) { implicit s =>
         val user = userRepo.get(externalUser)
-        val normUri = (metaData \ "url").asOpt[String].map { url =>
+        val normUri = (metaData \ "url").asOpt[String].collect { case url if url.length < URLFactory.MAX_URL_SIZE =>
```
andrewconner, line 54: This stops someone from browsing to a super long URL to cause a normalization exception.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
Colin-Lane: `cnt` is confusing. I thought that was the number of keeps being updated. It is actually the `iteration number of calling this method`

yingjieMiao: agree, maybe `iter` ? And a while loops seems much clearer, but I can see there is a trend in functional programming.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -683,6 +674,28 @@ class LibraryCommanderImpl @Inject() (
...
+          val keeps = db.readOnlyMaster { implicit s =>
+            keepRepo.getByLibraryIdAndExcludingVisibility(libraryId, Some(newVisibility), 1000)
```
Colin-Lane: If two people decide to modify a large library and change the visibility twice in a short period of time. The two threads executing requests may take turns changing the visibility from one visibility to the other until `cnt` hits 100. Let me think about this to see if there is something better that won't leave the result inconsistent in this case.

andrewconner: Sure, agreed. The problem is trying it all at once means a massive library could take a while, thus dirty reads while it's going anyways. Also, two modifications still has this problem: one would get committed over the other, potentially disagreeing with the library visibility.

Though, actually, there is a way to fix it by always re-checking the lib's visibility. Fix incoming. Thanks.

leogrim: Should we just do it withing the same transaction? See how long it takes
for a large library (rare anyway?)
Perhaps use an update statement + manual cache invalidation?
Then add constraints?

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
andrewconner: I actually purposely recurse inside a future, to ideally slow the updates a bit (ie, each batch has to take its turn in the thread pool queue)

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
andrewconner: Alright, fix to this pushed. The two threads will start to work in unison.

andrewconner:
Good?

PR #15964, opened 2015-06-26
Ansible config updates and support to run on existing/older instances

joshm1:
@andrewconner 

cc @stkem


joshm1:
i also made updates to disable the "changed" status of tasks that always report that they changed.  what happens is that the "command" (running raw shell commands) will always say they changed before they have nothing to compare against

andrewconner:
Nice!

PR #15965, opened 2015-06-26
View Organization Info

Colin-Lane:
Basic View data for Organization at GET     /m/1/teams/:id
@leogrim @ryanpbrewster


kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationView.scala
```diff
@@ -22,6 +22,10 @@ case class OrganizationNotificationInfo(
   handle: Option[PrimaryOrganizationHandle],
   image: Option[OrganizationImageInfo])
 
+@json
+case class FullOrganizationInfo(handle: OrganizationHandle, name: String, description: Option[String], avatar: Option[OrganizationAvatar], members: Seq[ExternalId[User]],
+  memberCount: Int, publicLibraries: Int, organizationLibraries: Int, secretLibraries: Int)
```
Colin-Lane, line 27: @jaredpetker @DerekBurch - Let me know if there are any additional fields you may want displayed?

Colin-Lane, line 27: I probably don't want to return the entire `OrganizationAvatar` but really just a link to it.

ryanpbrewster, line 27: Maybe `avatarPath: Option[ImagePath]`?

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
```diff
@@ -22,11 +24,36 @@ class OrganizationCommanderImpl @Inject() (
...
+    val orgHandle = org.getHandle
+    val orgName = org.name
+    val description = org.description
+    val (externalIds, memberCount, libraries) = db.readOnlyReplica { implicit session =>
+      val members = orgMembershipRepo.getByOrgId(orgId, Limit(8), Offset(0)).map(_.userId)
+      val externalIds = userRepo.getUsers(members).values.map(_.externalId).toSeq
+      val memberCount = orgMembershipRepo.countByOrgId(orgId)
+      val libraries = libraryRepo.countLibrariesForOrgByVisibility(orgId)
+      (externalIds, memberCount, libraries)
+    }
```
ryanpbrewster, line 46: We can maybe combine the `get(orgId)` into this session. We may be getting `FullOrganizationInfo` quite often so it's probably worth minimizing sessions.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
```diff
@@ -22,11 +24,36 @@ class OrganizationCommanderImpl @Inject() (
...
+
+    // TODO: how big should avatars be?
+    val avatar = organizationAvatarCommander.getBestImage(orgId, ImageSize(200, 200))
```
ryanpbrewster: Add a `val` to the commander, like `fullOrganizationInfoAvatarSize = ImageSize(150,150)` or something. Then it's easy to change.

Alternately, the `OrganizationAvatarCommander` has a list of sizes that it automatically creates any time a user uploads an avatar. You can choose from one of those.

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationView.scala
leogrim, line 27: Actually clients are not interested in the path (in S3), they need the full url including the CDN prefix. You want to call `ImagePath.getUrl` requiring an implicit `S3ImageConfig` in scope (just inject it).

PR #15966, opened 2015-06-26
Added grat email tests and refactored

camhashemi:
@andrewconner

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/model/GratificationData.scala
+++ b/kifi-backend/modules/common/app/com/keepit/model/GratificationData.scala
@@ -6,7 +6,7 @@ import play.api.libs.functional.syntax._
 import play.api.libs.json._
 
 case class CountData[T](totalCount: Int, countById: Map[Id[T], Int]) {
-  val sortedCountById = countById.toList.sortBy(-_._2)
+  def sortedCountById = countById.toList.sortBy(-_._2)
 }
 
 object CountData {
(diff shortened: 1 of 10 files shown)


kifi-backend/modules/common/app/com/keepit/model/GratificationData.scala
```diff
@@ -2,29 +2,32 @@ package com.keepit.model
...
+import play.api.libs.json._
 
-@json
-case class LibraryCountData(totalCount: Int, countById: Map[Id[Library], Int]) {
-
-  val sortedCountById = countById.toList.sortWith { _._2 > _._2 }
+case class CountData[T](totalCount: Int, countById: Map[Id[T], Int]) {
+  val sortedCountById = countById.toList.sortWith {
+    _._2 > _._2
+  }
```
yingjieMiao: `countById.toArray.sortBy(-_._2)` ?

camhashemi: oo pretty. thank you.

kifi-backend/modules/common/app/com/keepit/model/GratificationData.scala
```diff
@@ -2,29 +2,30 @@ package com.keepit.model
...
 import com.kifi.macros.json
+import play.api.libs.functional.syntax._
+import play.api.libs.json._
 
-@json
-case class LibraryCountData(totalCount: Int, countById: Map[Id[Library], Int]) {
-
-  val sortedCountById = countById.toList.sortWith { _._2 > _._2 }
+case class CountData[T](totalCount: Int, countById: Map[Id[T], Int]) {
+  val sortedCountById = countById.toList.sortBy(-_._2)
```
andrewconner: Make it a def and you save yourself having to sort until you need it. If you plan on calling it more than once, `lazy val`

andrewconner:
Cool abstraction (higher kinded type). Looks good, sans the `val` comment

PR #15967, opened 2015-06-26
versioned graph directory

yingjieMiao:
@leogrim 

on deploy, this should do nothing. 

We need to config elb so that a backup machine will not get traffic (as it can build different graphs). This is the same as versioning in search. (cortex has a different version management)


yingjieMiao:
we don't need to config elb, just internal routing config.

leogrim:
Awesome thank you!

yingjieMiao:
merge this now, as it really does nothing. next PR will address some configurations.

PR #15971, opened 2015-06-26
Added an org experiment guard to access actions

ryanpbrewster:
@leogrim @Colin-Lane

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/shoebox/controllers/OrganizationAccessActions.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/shoebox/controllers/OrganizationAccessActions.scala
@@ -40,7 +40,8 @@ trait OrganizationAccessActions {
               }
             case Failure(e) => Future.successful(OrganizationFail.INVALID_PUBLIC_ID.asErrorResponse)
           }
-        case _ => Future.successful(OrganizationFail.INSUFFICIENT_PERMISSIONS.asErrorResponse)
+        case _ =>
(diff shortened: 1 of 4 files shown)


ryanpbrewster:
This breaks basically all of the org controller tests. We should fix those before we merge this, I guess.

kifi-backend/modules/shoebox/app/com/keepit/shoebox/controllers/OrganizationAccessActions.scala
```diff
@@ -21,22 +21,26 @@ trait OrganizationAccessActions {
...
-    override def invokeBlock[A](request: MaybeUserRequest[A], block: (OrganizationRequest[A]) => Future[Result]): Future[Result] = {
-      Organization.decodePublicId(id) match {
-        case Success(orgId) =>
-          val userIdOpt: Option[Id[User]] = request match {
-            case userRequest: UserRequest[A] => Some(userRequest.userId)
-            case _ => None
+    override def invokeBlock[A](maybeRequest: MaybeUserRequest[A], block: (OrganizationRequest[A]) => Future[Result]): Future[Result] = {
+      maybeRequest match {
+        case request: UserRequest[_] if request.experiments.contains(ExperimentType.ORGANIZATION) =>
+          Organization.decodePublicId(id) match {
```
Colin-Lane, line 27: Looks fine to me, the only problem is that this won't guard someone from creating an `Organization`, `OrganizationAction` takes a `PublicId[Organization]`, which they won't have before creating one.

ryanpbrewster:
Sorry about the merge conflict you're about to deal with, @Colin-Lane

PR #15972, opened 2015-06-26
Feature/keep list view

adamjgrant:


kifi-backend/modules/shoebox/angular/src/common/build-css/grid.styl
```diff
@@ -0,0 +1,144 @@
...
+  display: block
+  margin:
+    left: grid_gutter/2
+    right: grid_gutter/2
+
+measure-columns(columns, prop)
+  if (prop == "width")
+    width: "calc(%s - %s)" % ((columns/grid_columns)*100%) grid_gutter
+
+//   else if (prop == "margin-left")
```
andrewconner: In general, we've tried to avoid committing commented out code, because it still has to be maintained and makes files harder to reason about. I know you're still working on it, so just conventions. Git history makes it easier to nuke code, and later get it if you need it.

adamjgrant: Hey Andrew

You're right. I actually don't intend to commit this as is. I push it every
now and then as more of a means of backup.

As I work through this, I'll be uncommenting, modifying, or deleting and
those changes will be squashed into the existing commits, so you won't
actually see commented-out code in the final PR.

kifi-backend/modules/shoebox/angular/src/libraries/libraryHeader.styl
```diff
@@ -965,168 +965,190 @@ collabImageDiam2 = 3.6em
...
-    padding-right: 234px  // rightColWidth + colGapWidth
+    figure
+      width: buttonHeight
+      height: buttonHeight
+      background-size: cover
+      background-repeat: no-repeat
+      display: inline-block
+      vertical-align: middle
+
+//
```
adamjgrant: Sorry about this (below), I need to see it side-by-side with what I'm working on. This will be removed and squashed into existing commit.

andrewconner: No worries, thanks

PR #15974, opened 2015-06-26
Safer org creation through the controller

ryanpbrewster:
The controller now uses the user request to get the user id instead of
trusting the user to provide it

@Colin-Lane @DerekBurch @jaredpetker @leogrim

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
@@ -56,11 +56,11 @@ class OrganizationCommanderImpl @Inject() (
 
   def isValidRequest(request: OrganizationRequest)(implicit session: RSession): Boolean = {
     request match {
-      case OrganizationCreateRequest(createrId, orgName) => true // TODO: can we check ahead of time that an org handle is available?
-      case OrganizationModifyRequest(orgId, requesterId, modifications) =>
+      case OrganizationCreateRequest(createrId, _, _) => true // TODO: can we check ahead of time that an org handle is available?
+      case OrganizationModifyRequest(requesterId, orgId, modifications) =>
         val permissions = orgMembershipRepo.getByOrgIdAndUserId(orgId, requesterId).map(_.permissions)
         permissions.exists { _.contains(EDIT_ORGANIZATION) } && areAllValidModifications(modifications)
-      case OrganizationDeleteRequest(orgId, requesterId) =>
+      case OrganizationDeleteRequest(requesterId, orgId) =>
         requesterId == orgRepo.get(orgId).ownerId
     }
   }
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileOrganizationController.scala
```diff
@@ -27,9 +28,14 @@ class MobileOrganizationController @Inject() (
   // getOrganizationCard
 
   def createOrganization = UserAction.async(parse.tolerantJson) { request =>
-    request.body.validate[OrganizationCreateRequest] match {
-      case _: JsError => Future.successful(BadRequest)
-      case JsSuccess(createRequest, _) =>
+    println("request.body = " + request.body)
+    (request.body \ "name").validate[String] match {
+      case _: JsError =>
+        println("Couldn't find a name in the json")
```
Colin-Lane, line 34: :-1:

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileOrganizationController.scala
```diff
@@ -27,9 +28,14 @@ class MobileOrganizationController @Inject() (
...
+        Future.successful(BadRequest)
+      case JsSuccess(orgName, _) =>
+        val orgDescriptionOpt = (request.body \ "description").asOpt[String]
+        val createRequest = OrganizationCreateRequest(requesterId = request.userId, orgName, orgDescriptionOpt)
```
leogrim, line 38: Do we check somewhere up the chain that these Strings are `nonEmpty` ?

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileOrganizationController.scala
andrewconner, line 34: Avoid `println`s, they go to the wrong log file. Also, include some context in log messages, like what the `request.body` is

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileOrganizationController.scala
```diff
@@ -27,9 +28,14 @@ class MobileOrganizationController @Inject() (
   // getOrganizationCard
 
   def createOrganization = UserAction.async(parse.tolerantJson) { request =>
-    request.body.validate[OrganizationCreateRequest] match {
-      case _: JsError => Future.successful(BadRequest)
-      case JsSuccess(createRequest, _) =>
+    println("request.body = " + request.body)
+    (request.body \ "name").validate[String] match {
```
andrewconner, line 32: You can do `(request.body \ "name").asOpt[String].map { ... }`

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileOrganizationController.scala
andrewconner, line 38: Probably fine, the commander can handle that, right? Or would you prefer data validation before?

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
```diff
@@ -56,11 +56,11 @@ class OrganizationCommanderImpl @Inject() (
 
   def isValidRequest(request: OrganizationRequest)(implicit session: RSession): Boolean = {
     request match {
-      case OrganizationCreateRequest(createrId, orgName) => true // TODO: can we check ahead of time that an org handle is available?
-      case OrganizationModifyRequest(orgId, requesterId, modifications) =>
+      case OrganizationCreateRequest(createrId, _, _) => true // TODO: can we check ahead of time that an org handle is available?
```
leogrim, line 59: Yes you can (see `HandleCommander.canBeClaimed`) 

I can actually see the front-end sending the handle on each keystroke while the backend responds on whether it's valid or not - before the user actually submits the organization. @jaredpetker @andrewconner

ryanpbrewster, line 59: `canBeClaimed` applies to a specific handle. The commander is actually using `autoSetOrganizationHandle`, and I would need to check whether or not _that_ will succeed. You've said before that it's almost guaranteed to succeed, but I'm going to keep the `try/catch` in for now.

leogrim, line 59: Oh right.

PR #15976, opened 2015-06-26
respect the user

yingjieMiao:
@leogrim @andrewconner 

per @sterlingz 's comments. This is a naive combination of user's hashtags and auto tags. Actual logic is easy to change. 

We can either merge it now, or merge it when search is fixed.


andrewconner:
Yeah let's merge when the site correctly searches.

yingjieMiao:
@andrewconner , do you think client need to convert a #scala query into "tag: scala"?

andrewconner:
Yeah probably. I'll do some testing

leogrim:
Search is fixed re: this

yingjieMiao:
@andrewconner  
is frontend ready to convert #query to "tag: query"?

andrewconner:
Not yet. Still on the list

PR #15977, opened 2015-06-26
Starring a Library

Colin-Lane:
@leogrim 
This is the initial code for adding a star to a library on the commander side.


kifi-backend/modules/common/conf/evolutions/shoebox/347.sql
```diff
@@ -1,19 +1,19 @@
...
---  ADD UNIQUE `library_alias_u_organization_id_slug` (`organization_id`,`slug`);
+-- these constraints leverage mysql's property of not enforcing constraints involving null columns. h2 does...
+--alter table library_alias
+--	add constraint `library_alias_f_user` foreign key (`owner_id`) references user(`id`),
+--	add constraint `library_alias_f_organization` foreign key (`organization_id`) references organization(`id`),
+--  drop index library_alias_i_owner_id_slug,
+--  add unique `library_alias_u_owner_id_slug` (`owner_id`,`slug`),
+--  add unique `library_alias_u_organization_id_slug` (`organization_id`,`slug`);
 
 insert into evolutions (name, description) values('347.sql', 'add organization_id, constraints to library_alias');
```
leogrim, line 17: We're at `349.sql` already :)

Colin-Lane, line 17: :-1:  I just merged from master like 2 seconds ago!

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -1586,6 +1586,21 @@ class LibraryCommanderImpl @Inject() (
     }
   }
 
+  def updateStarredLibrary(userId: Id[User], libraryId: Id[Library], isStarred: Boolean): Either[LibraryFail, LibraryMembership] = {
+    db.readOnlyMaster { implicit s =>
+      libraryMembershipRepo.getWithLibraryIdAndUserId(libraryId, userId)
+    } match {
+      case None => Left(LibraryFail(NOT_FOUND, "need_to_follow_to_star"))
```
leogrim: Is there a more generic error code for this? Such as `not_following`?

Colin-Lane: I need to check

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryMembershipCommander.scala
```diff
@@ -0,0 +1,77 @@
...
+import play.api.http.Status._
+
+@ImplementedBy(classOf[LibraryMembershipCommanderImpl])
+trait LibraryMembershipCommander {
+  def updateMembership(requestorId: Id[User], request: ModifyLibraryMembershipRequest): Either[LibraryFail, LibraryMembership]
+}
+
+case class ModifyLibraryMembershipRequest(userId: Id[User], libraryId: Id[Library],
+  subscription: Option[Boolean] = None,
+  isStarred: Option[Boolean] = None,
```
leogrim: Can we go for `starred` here to be consistent with `listed`?

Colin-Lane: yes we can.

leogrim:
Nice! @jaredpetker

Colin-Lane:
Evolution needs to be applied before I can merge. Evolution is on library_membership table. I won't merge during the middle of the day, will continue onwards.

kifi-backend/modules/common/conf/evolutions/shoebox/347.sql
```diff
@@ -1,19 +1,19 @@
-# SHOEBOX
+# shoebox
 
-# --- !Ups
+# --- !ups
```
andrewconner: Copy and paste an existing so you get casing right.

kifi-backend/modules/common/conf/evolutions/shoebox/348.sql
```diff
@@ -0,0 +1,9 @@
+# SHOEBOX
```
andrewconner: No need to commit two evolutions in the same PR.

kifi-backend/modules/common/app/com/keepit/model/LibraryMembership.scala
```diff
@@ -26,7 +26,8 @@ case class LibraryMembership(
     lastViewed: Option[DateTime] = None,
     lastEmailSent: Option[DateTime] = None,
     lastJoinedAt: Option[DateTime] = None,
-    subscribedToUpdates: Boolean = false) extends ModelWithState[LibraryMembership] with ModelWithSeqNumber[LibraryMembership] {
+    subscribedToUpdates: Boolean = false,
+    starred: Boolean = false) extends ModelWithState[LibraryMembership] with ModelWithSeqNumber[LibraryMembership] {
```
andrewconner: @eishay has some stories about LinkedIn banning `boolean` database fields, so he may be a bit more eloquent, but we really shouldn't be adding them. The difference between varchar and tinyint, at scale, won't really be noticed by anyone — and if it is, we're in a great situation. However, it's incredibly common to need more than two choices.

This same point was brought up for `subscribedToUpdates` above, and I suppose never got in (... and we've already talked about different kinds of subscriptions, where it's not boolean anymore). Right now, it's no difference between boolean and a varchar in term of engineering effort. Later, it's a giant pain to fix. @stkem

andrewconner:
Merge conflict

kifi-backend/modules/common/conf/evolutions/shoebox/347.sql
Colin-Lane: So weird, everything in this evolution was lowercased. (This one already existed, just fixed a merge conflict with what I was adding).
I'll revert all changes to this evolution.

kifi-backend/modules/common/app/com/keepit/model/LibraryMembership.scala
Colin-Lane: I can update this to use a varchar instead.

kifi-backend/modules/common/app/com/keepit/model/Library.scala
```diff
@@ -153,6 +154,37 @@ object Library extends ModelWithPublicIdCompanion[Library] {
     updatedAt = lib.updatedAt, name = lib.name, description = lib.description, color = lib.color, slug = lib.slug)
 }
 
+abstract class LibraryFilter(val value: String)
+object LibraryFilter {
+  case object OWN extends LibraryFilter("own")
```
leogrim, line 183: Is this `Own` and in `Owned` or as in `Contributing`?

Colin-Lane, line 183: I believe the existing functionality was both? I need to double check

Colin-Lane, line 183: The existing repo methods show it as `membership.access = "owner" or membership.access = "read_write"`

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryRepo.scala
```diff
@@ -41,15 +41,15 @@ trait LibraryRepo extends Repo[Library] with SeqNumberFunction[Library] {
...
 
-  def getOwnerLibrariesForAnonymous(ownerId: Id[User], page: Paginator, ordering: Option[LibraryOrdering])(implicit session: RSession): Seq[Library]
-  def getOwnerLibrariesForOtherUser(userId: Id[User], friendId: Id[User], page: Paginator, ordering: Option[LibraryOrdering])(implicit session: RSession): Seq[Library]
+  def getOwnerLibrariesForAnonymous(ownerId: Id[User], page: Paginator, ordering: List[LibraryOrdering])(implicit session: RSession): Seq[Library]
+  def getOwnerLibrariesForOtherUser(userId: Id[User], friendId: Id[User], page: Paginator, ordering: List[LibraryOrdering])(implicit session: RSession): Seq[Library]
   def getOwnerLibrariesForSelf(userId: Id[User], page: Paginator)(implicit session: RSession): Seq[Library]
-  def getOwnerLibrariesForSelfWithOrdering(userId: Id[User], page: Paginator, ordering: Option[LibraryOrdering])(implicit session: RSession): Seq[Library]
+  def getOwnerLibrariesForSelfWithOrdering(userId: Id[User], page: Paginator, ordering: List[LibraryOrdering])(implicit session: RSession): Seq[Library]
```
leogrim: What about using a `*` parameter here and in similar methods?

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryRepo.scala
```diff
@@ -323,28 +323,31 @@ class LibraryRepoImpl @Inject() (
...
+       """.as[Library].list
+  }
+
+  def getOrderings(ordering: List[LibraryOrdering]) = {
+    if (ordering.isEmpty) "lib.member_count desc, lib.last_kept desc, lib.id desc" else ordering.map {
+      case LibraryOrdering.ALPHABETICAL => "lib.name asc, lib.id desc"
+      case LibraryOrdering.MEMBER_COUNT => "lib.member_count desc"
+      case LibraryOrdering.LAST_KEPT_INTO => "lib.last_kept desc"
+      case LibraryOrdering.STARRED => "CASE lm.starred WHEN \"starred\" THEN 1 ELSE 0 END"
+    }.mkString(",")
```
leogrim: This is assuming your library table is called `lib`, this should probably not be a public method and make that dependency more explicit.
For `starred`, won't that string be compared to `null` automatically?

At some point, the client will probably pass the ordering direction as well, so you may not want to hardcode them.

Overall this looks cool but may be over-engineered, what do you think?

Colin-Lane: I don't particularly like null values in a database. How much of a hassle would it be to update all existing memberships to have "unstarred" as their default?

leogrim: Why don't you like nulls?

kifi-backend/modules/common/app/com/keepit/model/Library.scala
leogrim, line 183: Ok, perhaps it's best to split it into `Owner`, `Contributor` and `Follower` (exclusive sets with each other). And have backward compatibility to deserialize `own` to grab both `Owner` and `Contributor`

leogrim, line 183: Or at least, come up with a better name such as `Mine`, not to confuse it with ownership.

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryRepo.scala
Colin-Lane: They tend to make writing queries a bit harder because you can't cleanly compare against them with `=`, instead you need to use `is Null`

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryRepo.scala
```diff
@@ -334,16 +334,22 @@ class LibraryRepoImpl @Inject() (
...
-      case None =>
-        s"lib.member_count desc, lib.last_kept desc, lib.id desc"
+  def getOwnerLibrariesForUserIdWithOrdering(userId: Id[User], page: Paginator, ordering: Option[LibraryOrdering])(implicit session: RSession): Seq[Library] = {
+    import com.keepit.common.db.slick.StaticQueryFixed.interpolation
+    sql"""select lib.* from library lib inner join library_membership lm on lm.library_id = lib.id and lm.user_id = $userId and (lm.access='owner' or lm.access='read_write')
+         order by #${getOrderBySql(ordering)} limit ${page.itemsToDrop}, ${page.size}
+       """.as[Library].list
+  }
+
+  def getOrderBySql(orderingOpt: Option[LibraryOrdering]) = {
```
andrewconner: Best to pass in the table name

PR #15981, opened 2015-06-27
Cleaned up the input/output for org creation

ryanpbrewster:
Now it takes in a JSON object with a bunch of optional fields (name required)
and returns the standard FullOrganizationInfo

@leogrim @Colin-Lane @DerekBurch @jaredpetker

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
@@ -10,6 +10,8 @@ import com.keepit.common.store.ImageSize
 import com.keepit.model.OrganizationPermission.{ VIEW_ORGANIZATION, EDIT_ORGANIZATION }
 import com.keepit.model._
 
+import scala.util.{ Success, Failure, Try }
+
 @ImplementedBy(classOf[OrganizationCommanderImpl])
 trait OrganizationCommander {
   def get(orgId: Id[Organization]): Organization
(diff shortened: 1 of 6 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
```diff
@@ -50,42 +53,50 @@ class OrganizationCommanderImpl @Inject() (
...
   }
 
+  private def organizationWithModifications(org: Organization, modifications: OrganizationModifications): Organization = {
+    org.withName(modifications.name.getOrElse(org.name))
+      .withDescription(modifications.description.orElse(org.description))
+      .withBasePermissions(modifications.basePermissions.getOrElse(org.basePermissions))
+  }
   def createOrganization(request: OrganizationCreateRequest): Either[OrganizationFail, OrganizationCreateResponse] = {
-    // TODO: if we can find a way to validate that request, maybe we don't have to try/catch
     try {
```
andrewconner: This is pretty unusual in Scala; further, we probably overuse `Try` also. Where's the exception going to be coming out from? Seems the error is always `HANDLE_UNAVAILABLE`, which could be handled directly (and we wouldn't be throwing away other exceptions).

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileOrganizationController.scala
```diff
@@ -28,17 +28,17 @@ class MobileOrganizationController @Inject() (
   // getOrganizationCard
 
   def createOrganization = UserAction.async(parse.tolerantJson) { request =>
```
andrewconner: Why `.async`? Seems all the cases wrap a completed future, right?

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
ryanpbrewster: It comes out of the Handle commander. If it can't find a suitable handle
for an organization it throws an exception.

I could just let it bubble up, but I'd rather just tell the user that we
couldn't find a name for them.

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileOrganizationController.scala
ryanpbrewster: I'm actually not clear on when you should use async.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
andrewconner: I see. Hrm, so usually, I'd recommend that the handle commander return an option or something if we expect a failure. But if we want to leave that behavior, at least just `Try` around the part that may blow up, and leave the rest

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileOrganizationController.scala
andrewconner: Ah, easy. no `async` when the function returns immediately (ie, you have your `Ok` without Futures). When, however, one (or more) of your code branches returns a Future, you need `.async`.

It's just because `Action` in play expects a `Result`. `Action.async` expects a `Future[Result]`.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
leogrim: `HandleCommander` does return an `Option`. I guess this exception (just like the one in `UserCommander`) is a way for us to rollback the session (and it never happened).

ryanpbrewster: Yep, if we can't assign the org a handle I want to rollback the session. Correct me if I'm wrong but that means we actually _must_ throw an exception to break out of the session.

PR #15988, opened 2015-06-29
bug fix

Colin-Lane:
@ryanpbrewster

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/model/Organization.scala
+++ b/kifi-backend/modules/common/app/com/keepit/model/Organization.scala
@@ -34,7 +34,9 @@ case class Organization(
   def withState(newState: State[Organization]): Organization = this.copy(state = newState)
   def withName(newName: String): Organization = this.copy(name = newName)
   def withDescription(newDescription: Option[String]): Organization = this.copy(description = newDescription)
-  def withBasePermissions(newBasePermissions: BasePermissions): Organization = this.copy(basePermissions = newBasePermissions)
+  def withBasePermissions(newBasePermissions: BasePermissions): Organization = {
+    this.copy(basePermissions = basePermissions.copy(permissionsMap = basePermissions.permissionsMap ++ newBasePermissions.permissionsMap))
+  }
 
   def getNonmemberPermissions = basePermissions.forNonmember
   def getRolePermissions(role: OrganizationRole) = basePermissions.forRole(role)
--- a/kifi-backend/modules/common/app/com/keepit/model/Organization.scala
+++ b/kifi-backend/modules/common/app/com/keepit/model/Organization.scala
@@ -35,7 +35,7 @@ case class Organization(
   def withName(newName: String): Organization = this.copy(name = newName)
   def withDescription(newDescription: Option[String]): Organization = this.copy(description = newDescription)
   def withBasePermissions(newBasePermissions: BasePermissions): Organization = {
-    this.copy(basePermissions = basePermissions.copy(permissionsMap = basePermissions.permissionsMap ++ newBasePermissions.permissionsMap))
+    this.copy(basePermissions = new BasePermissions(basePermissions.permissionsMap ++ newBasePermissions.permissionsMap))
   }
 
   def getNonmemberPermissions = basePermissions.forNonmember
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
@@ -76,7 +76,7 @@ class OrganizationCommanderImpl @Inject() (
     lazy val badName = modifications.name.exists(_.isEmpty)
     lazy val badBasePermissions = modifications.basePermissions.exists { bps =>
       // Are there any members that can't even see the organization?
-      OrganizationRole.all exists { role => !(bps.permissionsMap.contains(Some(role)) && bps.forRole(role).contains(VIEW_ORGANIZATION)) }
+      OrganizationRole.all exists { role => !bps.forRole(role).contains(VIEW_ORGANIZATION) }
     }
     !badName && !badBasePermissions
   }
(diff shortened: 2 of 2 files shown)


Colin-Lane:
Validating a request from a user isn't catching the case where the map is missing a key. Do we want to handle this explicitly instead of just throwing a `NoSuchElementException`?

ryanpbrewster:
I think just `OrganizationFail.BAD_PARAMETERS`

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
```diff
@@ -76,7 +76,7 @@ class OrganizationCommanderImpl @Inject() (
     lazy val badName = modifications.name.exists(_.isEmpty)
     lazy val badBasePermissions = modifications.basePermissions.exists { bps =>
       // Are there any members that can't even see the organization?
-      OrganizationRole.all exists { role => !bps.forRole(role).contains(VIEW_ORGANIZATION) }
+      OrganizationRole.all exists { role => !(bps.permissionsMap.contains(Some(role)) && bps.forRole(role).contains(VIEW_ORGANIZATION)) }
```
Colin-Lane: The problem is that this method returns a Boolean, so there is no clean way to specify what went wrong.
I think I should update the method to return `Either[OrganizationFail, Boolean]` or `Option[OrganizationFail]`

ryanpbrewster: I really wanted to do an `Option[OrganizationFail]` when I first wrote this code, but I went with `Boolean` because it seemed clearer. I would be in favor of calling it `whyIsThisRequestBad`.

Maybe we add `isValidRequest(req) = whyIsThisRequestBad(req).isEmpty` around just for convenience.

ryanpbrewster: Also, maybe split this into pieces:
`OrganizationRole.all == bps.permissionsMap.keySet.collect { Some(role) => role } && bps.permissionsMap.isDefinedAt(None)` or something?

ryanpbrewster: Alternately! We could just overwrite the new elements and keep the old ones.

kifi-backend/modules/common/app/com/keepit/model/Organization.scala
```diff
@@ -34,7 +34,9 @@ case class Organization(
   def withState(newState: State[Organization]): Organization = this.copy(state = newState)
   def withName(newName: String): Organization = this.copy(name = newName)
   def withDescription(newDescription: Option[String]): Organization = this.copy(description = newDescription)
-  def withBasePermissions(newBasePermissions: BasePermissions): Organization = this.copy(basePermissions = newBasePermissions)
+  def withBasePermissions(newBasePermissions: BasePermissions): Organization = {
+    this.copy(basePermissions = basePermissions.copy(permissionsMap = basePermissions.permissionsMap ++ newBasePermissions.permissionsMap))
```
ryanpbrewster: I wouldn't do a `basePermissions.copy` here. There is nothing else in `BasePermissions`, so we may as well directly call the constructor.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
```diff
@@ -60,21 +60,30 @@ class OrganizationCommanderImpl @Inject() (
...
   def isValidRequest(request: OrganizationRequest)(implicit session: RSession): Boolean = {
+    getValidationError(request).isEmpty
+  }
+
+  def getValidationError(request: OrganizationRequest)(implicit session: RSession): Option[OrganizationFail] = {
     request match {
       case OrganizationCreateRequest(createrId, initialParameters) =>
-        initialParameters.name.isDefined && areAllValidModifications(initialParameters)
-
+        if (initialParameters.name.isDefined && areAllValidModifications(initialParameters)) None else Some(OrganizationFail.BAD_PARAMETERS)
```
ryanpbrewster: Below you use the convention of `if (bad) Some(fail) else None`. I like that. We should do that here.

It makes sense to go through all the possible ways you could fail before finally yielding `None`.

Colin-Lane: yeah, I may have skipped this case because I didn't need it initially. It's a good idea to cover all our bases now rather than forget later. Fix in progress.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
```diff
@@ -110,18 +119,18 @@ class OrganizationCommanderImpl @Inject() (
...
-          applyNewBasePermissionsToMembers(memberships, org.basePermissions, request.modifications.basePermissions.get)
-        }
+      getValidationError(request) match {
+        case None =>
+          val org = orgRepo.get(request.orgId)
+
+          val modifiedOrg = organizationWithModifications(org, request.modifications)
+          if (request.modifications.basePermissions.nonEmpty) {
+            val memberships = orgMembershipRepo.getAllByOrgId(org.id.get)
+            applyNewBasePermissionsToMembers(memberships, org.basePermissions, request.modifications.basePermissions.get)
```
ryanpbrewster: I think you want to grab the full `BasePermissions` from `modifiedOrg`.

ryanpbrewster:
:+1:

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
```diff
@@ -60,21 +60,30 @@ class OrganizationCommanderImpl @Inject() (
...
   def isValidRequest(request: OrganizationRequest)(implicit session: RSession): Boolean = {
+    getValidationError(request).isEmpty
+  }
+
+  def getValidationError(request: OrganizationRequest)(implicit session: RSession): Option[OrganizationFail] = {
     request match {
       case OrganizationCreateRequest(createrId, initialParameters) =>
-        initialParameters.name.isDefined && areAllValidModifications(initialParameters)
-
+        if (!initialParameters.name.isDefined || !areAllValidModifications(initialParameters)) Some(OrganizationFail.BAD_PARAMETERS) else None
```
yingjieMiao, line 69: could be `initialParameters.name.isEmpty`, this way you wouldn't need the extra `!`

PR #15989, opened 2015-06-29
migrate tag cloud stuff to cortex

yingjieMiao:
@leogrim 

once things start working, I will kill the code in rover.


leogrim:
Great thanks!

PR #15994, opened 2015-06-29
Feature/keep list view

adamjgrant:


kifi-backend/modules/shoebox/angular/src/common/build-css/grid.styl
```diff
@@ -0,0 +1,136 @@
...
+
+column(columns = 12, left = 0, right = 0, constants = true)
+  // Everything is a full width column on mobile by default
+  set-column(12, 0, 0)
+
+  if (constants == true)
+    column-universals()
+
+  // Look for keywords
+  if (columns == "half")
```
andrewconner, line 115: Fancy

kifi-backend/modules/shoebox/angular/src/keeps/keeps.tpl.html
```diff
@@ -38,6 +38,10 @@
   </div>
   <ol class="kf-my-keeps" smart-scroll scroll-distance="scrollDistance" scroll-disabled="isScrollDisabled()" scroll-next="scrollNext()(keeps.length)">
     <div class="kf-keep" ng-hide="keep.unkept && library" ng-class="{'editing': edit.enabled}" ng-repeat="keep in keeps">
+
+      <p ng-if="galleryView">GALLERYVIEW</p>
```
andrewconner: Is this debugging related?

adamjgrant: Yes, just for me to visually see what's going on briefly.

kifi-backend/modules/shoebox/angular/src/keep/keepCard.tpl.html
```diff
@@ -16,7 +16,11 @@
     <div class="kf-keep-title">
       <a class="kf-keep-title-link" ng-href="{{::keep.url}}" target="_blank" ng-click="clickCallback({keep: keep, event: $event})" rel="nofollow" click-action="clickedArticleTitle">{{displayTitle}}</a>
     </div>
-    <div class="kf-keep-description" kf-ellipsis full-text="::keep.summary.description" max-num-lines="::image.maxDescLines || 4"></div>
+    <div class="kf-keep-description" ng-hide="!galleryView && keep.note" kf-ellipsis full-text="keep.summary.description" max-num-lines="image.maxDescLines || defaultDescLines"></div>
+    <div class="kf-keep-note-icon" ng-if="::keep.note" ng-click="editKeepNote($event, keep)">
+      <span class="svg-note-gray"></span>
+    </div>
+    <div class="kf-keep-note-compact" ng-bind-html="::keep.note|noteHtml" ng-if="::keep.note || keep.user.id === me.id"></div>
```
adamjgrant, line 42: Looking for your opinion on this one, @andrewconner and @cvializ 

What I'm trying to do here is make a second instance of these buttons 

screen shot 2015-07-02 at 3 58 43 pm

In the header so it looks like this in compact view:

screen shot 2015-07-02 at 3 59 32 pm

The problem is, the current ones are in the `keepCard` controller and where I want to move it resides in the `keep` itself–the parent controller of `keepCard`.  That introduces a whole bunch of complexities (unless I'm looking at it the wrong way)

So I'm considering using CSS to just absolute position the existing ones so they visually appear in the header.

It's kind of an icky solution, but I don't know if it's more or less icky than the alternative.

Your thoughts?

adamjgrant, line 42: I just realized I'm referencing the wrong piece of code for this discussion but I think maybe my description is clear anyway.

kifi-backend/modules/shoebox/angular/src/keep/keepCard.js
```diff
@@ -19,7 +19,7 @@ angular.module('kifi')
     var descCharsPerMaxWidthLine = 48;
     var maxSizedImageW = 0.4 * (cardInnerW - gutterW);
 
-    function calcImageSize(summary, title) {
```
adamjgrant, line 22: @andrewconner @cvializ It looks like just the "cyclomatic complexity too high" error from JS Hint is due simply to adding another parameter. Can we hide this from jshint? The alternative is to refactor this function.

andrewconner, line 22: Sure.

andrewconner, line 22: That lint error is a bit of a downer anyways, since it usually surfaces at the least fortunate time (adding something minor — bang — not really when it's originally written). It is a good hint that perhaps it's time to refactor, but a bit harsh to break the build over it. Thoughts?

adamjgrant, line 22: It does feel a little icky so it would be good to work on it eventually.

adamjgrant:
@andrewconner @cvializ Truncation of lines is going to need some work but due to the immensity of this PR, it's probably best to pull this in. Afterwards, refactoring the calcImageSize function might be worthwhile.

andrewconner:
Wahoo!

cvializ:
: D

PR #15995, opened 2015-06-29
Org lib migration

ryanpbrewster:
@leogrim @Colin-Lane

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
@@ -608,6 +608,7 @@ class LibraryCommanderImpl @Inject() (
       Left(LibraryFail(FORBIDDEN, "permission_denied"))
     } else {
       val targetMembership = targetMembershipOpt.get
+      val currentSpace = LibrarySpace(userId, organizationId = targetLib.organizationId)
       val newSpace = LibrarySpace(userId, organizationId = modifyReq.orgId.flatMap(_.destination).orElse(targetLib.organizationId))
 
       def validName(newNameOpt: Option[String]): Either[LibraryFail, String] = {
@@ -677,11 +678,9 @@ class LibraryCommanderImpl @Inject() (
         }
 
         val lib = db.readWrite { implicit s =>
-          if (targetLib.slug != newSlug) {
-            // TODO: update alias for organization updates and/or owner updates. (Need to do all updates in one go - only create a single alias)
-            val ownerId = targetLib.ownerId
-            libraryAliasRepo.reclaim(ownerId, newSlug)
-            libraryAliasRepo.alias(ownerId, targetLib.slug, targetLib.id.get)
+          if (newSpace != currentSpace || newSlug != targetLib.slug) {
+            libraryAliasRepo.reclaim(currentSpace, newSlug)
+            libraryAliasRepo.alias(newSpace, targetLib.slug, targetLib.id.get)
           }
           if (targetMembership.listed != newListed) {
             libraryMembershipRepo.save(targetMembership.copy(listed = newListed))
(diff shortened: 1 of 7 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -1403,7 +1418,7 @@ class LibraryCommanderImpl @Inject() (
 
   def getLibraryBySlugOrAlias(ownerId: Id[User], slug: LibrarySlug): Option[(Library, Boolean)] = {
     db.readOnlyMaster { implicit session =>
-      libraryRepo.getBySlugAndUserId(ownerId, slug).map((_, false)) orElse
+      libraryRepo.getBySpaceAndSlug(LibrarySpace.fromUserId(ownerId), slug).map((_, false)) orElse
```
leogrim, line 1424: You probably want to change the commander method as well to take a space, and update the router down the road.

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryRepo.scala
```diff
@@ -164,42 +165,42 @@ class LibraryRepoImpl @Inject() (
...
+      case UserSpace(userId) => getByUserIdAndSlug(userId, slug, excludeStates)
+      case OrganizationSpace(orgId) => getByOrgIdAndSlug(orgId, slug, excludeStates)
     }
   }
 
-  private val getByNameOrSlugCompiled = Compiled { (userId: Column[Id[User]], name: Column[String], slug: Column[LibrarySlug]) =>
-    (for (b <- rows if (b.name === name || b.slug === slug) && b.ownerId === userId) yield b)
+  private def getByUserIdAndNameOrSlug(userId: Id[User], name: String, slug: LibrarySlug, excludeStates: Set[State[Library]])(implicit session: RSession): Option[Library] = {
+    (for (b <- rows if b.ownerId === userId && (b.slug === slug || b.name === name) && !b.state.inSet(excludeStates)) yield b).firstOption
   }
```
leogrim: If you're checking for `b.slug === slug || b.name === name`, then this query could potentially return 2 libraries. So the whole method signature is misleading as it returns an `Option`. By checking both together, you also don't know which one the collision is originating with. Perhaps better to keep these checks separate?

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryView.scala
```diff
@@ -56,7 +60,10 @@ case class LibraryModifyRequest(
   color: Option[LibraryColor] = None,
   listed: Option[Boolean] = None,
   whoCanInvite: Option[LibraryInvitePermissions] = None,
-  subscriptions: Option[Seq[LibrarySubscriptionKey]] = None)
+  subscriptions: Option[Seq[LibrarySubscriptionKey]] = None,
+  // We can move the library from org to user space.
+  // User space is defined as having None for OrganizationId.
```
leogrim: "User Space" being the "Owner's Personal Space"

kifi-backend/modules/shoebox/test/com/keepit/controllers/website/KeepsControllerTest.scala
```diff
@@ -549,7 +549,7 @@ class KeepsControllerTest extends Specification with ShoeboxTestInjector with He
 
           val url1 = urlRepo.save(URLFactory(url = uri1.url, normalizedUriId = uri1.id.get))
           val url2 = urlRepo.save(URLFactory(url = uri2.url, normalizedUriId = uri2.id.get))
-          val mainLib = libraryRepo.getBySlugAndUserId(user1.id.get, LibrarySlug("main"))
+          val mainLib = libraryRepo.getBySpaceAndSlug(LibrarySpace.fromUserId(user1.id.get), LibrarySlug("main"))
```
leogrim, line 552: Implicit conversions exist so that you don't need to do this, you can't just pass in the user id :)

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryRepo.scala
ryanpbrewster: I didn't look into why this method even exists. Something calls it, somewhere. I'll check.

kifi-backend/modules/shoebox/test/com/keepit/controllers/website/KeepsControllerTest.scala
ryanpbrewster, line 552: I actually prefer the explicit conversion. Implicits are nice if the code is littered with method calls that would be broken, but in this case there are only ~5 places where `getBySpaceAndSlug` is called (almost entirely in tests) so I just added them in by hand.

PR #15996, opened 2015-06-29
Add domain-related admin pages

maxov:
_Do not merge_, depends on #15957 for completion

Adds views for finding domains by hostname (under `admin.SliderAdminController.findDomain`) and viewing/modifying an individual domain (under `admin.SliderAdminController.getDomain`).

Depends on #15957 to access the method to save changes to an individual domain.

@camhashemi 
@leogrim


andrewconner:
Wahoo!

PR #15997, opened 2015-06-29
Modify And Delete Org

Colin-Lane:
@leogrim @ryanpbrewster

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
@@ -85,7 +85,7 @@ class OrganizationCommanderImpl @Inject() (
     val badName = modifications.name.exists(_.isEmpty)
     val badBasePermissions = modifications.basePermissions.exists { bps =>
       // Are there any members that can't even see the organization?
-      OrganizationRole.all exists { role => !bps.forRole(role).contains(VIEW_ORGANIZATION) }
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
```diff
@@ -85,7 +85,7 @@ class OrganizationCommanderImpl @Inject() (
     val badName = modifications.name.exists(_.isEmpty)
     val badBasePermissions = modifications.basePermissions.exists { bps =>
       // Are there any members that can't even see the organization?
-      OrganizationRole.all exists { role => !bps.forRole(role).contains(VIEW_ORGANIZATION) }
+      OrganizationRole.all exists { role => !(bps.permissionsMap.contains(Some(role)) && bps.forRole(role).contains(VIEW_ORGANIZATION)) }
```
ryanpbrewster, line 88: No longer necessary. We impose this condition by construction.

Colin-Lane, line 88: Not true, these permissions come from the modification request. The request is not validated until this step.

ryanpbrewster, line 88: Nevermind, this is probably fine.

ryanpbrewster:
:+1:

PR #16000, opened 2015-06-30
Fixes to prod configs

andrewconner:
@leogrim @joshm1


joshm1:
:+1: PR 16,000!

provisioning/roles/kifi.common/files/etc/nginx/conf.d/search.conf
```diff
@@ -9,6 +9,7 @@ server {
   server_name search.kifi.com;
 
   location / {
+    proxy_http_version 1.1;
```
andrewconner, line 12: No clue why this is needed now, but it broke Saturday morning when the machines were deployed. Maybe it was in a global config?

PR #16002, opened 2015-06-30
Feature/svg symbol sprite

adamjgrant:


kifi-backend/modules/shoebox/angular/play-refs/ng/dist
```diff
@@ -1 +0,0 @@
-../../dist
```
andrewconner, line 1: This needs to live. It's a less-than-ideal fix for a server side requirement we have regarding assets. Given a few hours I'm sure we could make it elegant, but it works.

kifi-backend/modules/shoebox/angular/gulpfile.js
```diff
@@ -266,7 +268,13 @@ gulp.task('sprite-classes', function () {
 
 gulp.task('sprite', ['sprite-imports', 'sprite-classes', 'svg-sprite']);
 
-gulp.task('svg-sprite', function() {
+gulp.task('symbol-sprite', function() {
```
adamjgrant: Still more work to do here. There's a setting you can turn on in Sketch to prevent it from exporting metadata. Since a lot of these icons already have metadata in them, minification and showing the unminified SVG in the browser are failing. I might just exclusively use the icons that I created the first go around and I can turn on minification.

kifi-backend/modules/shoebox/marketing
```diff
@@ -1 +1 @@
-Subproject commit f913ec66118bb84a7806fb0a76597eab3e9f3e7a
+Subproject commit f4e498a31b4568a19738c789c619bc357e4ce365
```
andrewconner: Whoops, this changed the marketing site's commit hash. I've noticed people have been doing this accidentally a lot recently. Any ideas why? Didn't seem to happen much before. Is it `git add .`?

kifi-backend/modules/shoebox/marketing
```diff
@@ -1 +1 @@
-Subproject commit f913ec66118bb84a7806fb0a76597eab3e9f3e7a
```
adamjgrant: Not sure what's going on with my git submodules but I'll clean this up before merge.

adamjgrant:
Going to continue on this tomorrow.

andrewconner:
Cool. I suppose this adds a dependency on ruby, which is fine for now, but we may want to rewrite in JS in the future.

kifi-backend/modules/shoebox/marketing
joshm1: this usually happens with `git add .` or `git commit -a` anytime somebody pulls an the repo w/o updating the submodule `git submodule update`

kifi-backend/modules/shoebox/angular/play-refs/ng/dist
andrewconner, line 1: This is a soft link to `kifi-backend/modules/shoebox/angular/dist` in the angular folder (needed in angular dev mode to mirror the real thing, it hits `GET /dist/...`). We also use `GET /dist/...` for server-side assets, but our web server had a change a version back that changed how it kept track of assets. Basically, we needed a nested directory, and if we just included the top level dist, it would also have `src` and even `node_modules`. No good. I'm sure there's a better fix, but this means Play (webserver) is happy and angular dev is happy.

adamjgrant:
I agree. It's a quick win for now since I already have the code, but would be worth revisiting to do in node.

kifi-backend/modules/shoebox/marketing
adamjgrant: I don't know. Why would `git add .` cause this?

kifi-backend/modules/shoebox/angular/play-refs/ng/dist
adamjgrant, line 1: Oh, I didn't do this?

andrewconner:
Yep yep. Taking on technical debt is a-ok when taken intentionally and cumulative cost is low.

kifi-backend/modules/shoebox/angular/play-refs/ng/dist
andrewconner, line 1: Oh weird. I wonder why it deleted. This would probably not work on a Windows machine (because NTFS doesn't support that).

kifi-backend/modules/shoebox/marketing
adamjgrant: Weird. I did a `git submodule update --init --recursive` and a `git pull origin master` in the repo directory. I'll keep poking it.

adamjgrant:
Okay, this should be ready to go @andrewconner 

Before testing you'll need to run `npm install`
Afterwards, running `gulp` should produce a test.html file at `img/symbol-sprites` where you can see the effect. I've added just two SVGs to demonstrate the effect. 

After this is merged, I'd like to use this going forward with new icons starting with those associated with the compact view task. Eventually, it would be nice to get the legacy icons into this workflow as well. I'd be happy to take on that task over time.

andrewconner:
Sounds good. :+1:

adamjgrant:
Niiiice.

screen shot 2015-06-30 at 5 04 05 pm

PR #16013, opened 2015-06-30
Scraper dependencies refactor

maxov:
Moves out scraper dependencies from common, also fixes a few things.

@leogrim


kifi-backend/project/Build.scala
```diff
@@ -157,7 +159,7 @@ object ApplicationBuild extends Build {
   ).settings(
     libraryDependencies ++= scraperDependencies,
     javaOptions in Test += "-Dconfig.resource=application-scraper.conf"
-  ).dependsOn(common % "test->test;compile->compile")
+  ).dependsOn(common % "test->test;compile->compile").dependsOn(rover)
```
maxov: Some scraper classes still depend on SignatureBuilder, so I decided to add a dependency to the rover project

kifi-backend/modules/rover/app/com/keepit/rover/document/utils/SignatureBuilder.scala
```diff
@@ -0,0 +1,41 @@
+package com.keepit.rover.document.utils
+
+import java.io.StringReader
+
+import com.keepit.rover.article.Article
+import com.keepit.rover.article.content.ArticleContent
+import org.apache.lucene.analysis.standard.StandardTokenizer
+import org.apache.lucene.analysis.tokenattributes.CharTermAttribute
+
+class SignatureBuilder(windowSize: Int = 20) extends Signature.Builder(windowSize) {
```
maxov, line 10: This object functions less like a signature builder and more like a signature implementation, I might try to refactor the behavior here a bit

kifi-backend/project/Build.scala
leogrim: Rover already depends on Common, so you can't have it the other way around as well (that would be a circular dependency). So you need to remove the dependency on SignatureBuilder first, then you get to drop the libraries as they're not required anymore. Does this make sense?

kifi-backend/modules/rover/app/com/keepit/rover/document/utils/SignatureBuilder.scala
leogrim, line 10: Usually when a class is deemed a `Builder`, it's because it is a _mutable_ structure that accumulates some state (here it processes and accumulates information we stream through text). A `Signature` is the _immutable_ end result that comes out of the builder. Makes sense?

kifi-backend/project/Build.scala
maxov: I'm not sure how there's a circular dependency, sbt seems fine with it. I added rover as a dependency to the scraper, and from what I see common does not depend on the scraper.

leogrim: Whoups you're right, didn't notice that 80 lines of code separated `commonSettings` from your line. 

What's breaking in Scraper exactly? Can you just kill that code? (after porting over some parts of it if necessary)

kifi-backend/modules/rover/app/com/keepit/rover/document/utils/SignatureBuilder.scala
maxov, line 10: I understand why it's called a builder. So far all usages of the signature builder (except tests) only do one add() call and call build immediately after, so there doesn't seem to be a need to keep state.

kifi-backend/project/Build.scala
maxov: Extractor, DefaultExtractor, and LinkProcessingExtractor seem to use it. I will try to kill it, since those classes do seem isolated

leogrim: These classes can go, but some tests associated with them need to be ported over to `DefaultArticleFetcher` in Rover

kifi-backend/modules/rover/app/com/keepit/rover/document/utils/SignatureBuilder.scala
leogrim, line 10: I guess you're right in our current usage of it, as documents are always fully loaded in memory so that you can process the text in one go and call build. Now it's been built so that we could use it on some larger data, in a way that we can process chunks of it without having to load it all at once. Haven't done that yet though :)

kifi-backend/modules/rover/test/com/keepit/rover/article/fetcher/DefaultArticleFetcherTest.scala
```diff
@@ -0,0 +1,83 @@
...
+
+    "ignore canonical URLs that are the page URL with parameter values URL-encoded twice" in {
+      val url = "https://soundcloud.com/search?q=%D8%AF%DB%8C%D8%A7%D9%84%D9%88%DA%AF"
+      val scraped = fetch("double-encode.txt")
+      val normalization = scraped.content.normalization
+      normalization.canonicalUrl === None
+    }
+
+    "extract keywords" in {
+      //      val extractor = setup("https://cnn.com/url2", "www.cnn.com.health.txt", Some(500))
```
maxov: While the character limit does exist and is by default 100,000, there is no way to fetch an article with a custom limit. Should I remove this specific test?

leogrim: Yep!

kifi-backend/modules/rover/test/com/keepit/rover/article/fetcher/DefaultArticleFetcherTest.scala
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    "ignore canonical URLs that are the page URL HTML-escaped twice" in {
+      val longUrl = "http://www.livejournal.com/gsearch?engine=google&cx=partner-pub-5600223439108080%3A3711723852&cof=FORID%3A10&ie=UTF-8&q=test&sa=Search&siteurl="
+      val scraped = fetch("double-escape.txt")
+      val normalization = scraped.content.normalization
+      normalization.canonicalUrl === None
+    }
+
+    "ignore canonical URLs that are the page URL with parameter values URL-encoded twice" in {
```
maxov: This test fails because `NormalizationCandidateSanitizer.validateCandidateUrl` is no longer being called to sanitize URLs in either JsoupDocument or TikaDocument. However, putting it back is not trivial since it requires the absolute fetched url for some checks, which is not provided. Should I remove this test as well or try to get that behavior back?

leogrim: These checks are not performed in the fetcher anymore, but at a later stage in the pipeline. However we still want to make sure `NormalizationCandidateSanitizerTest` does the right thing. Could you move the tests for each url into a static `NormalizationCandidateSanitizerTest` that just relies on `NormalizationCandidateSanitizer.validate`?

kifi-backend/modules/rover/test/com/keepit/rover/article/fetcher/ArticleFetcherTest.scala
```diff
@@ -0,0 +1,32 @@
...
+
+  val fetcherClass: Class[FetcherType]
+
+  val articleKind: ArticleKind[A]
+  val articleFetchDuration = 10.seconds
+
+  lazy val articleFetcher: FetcherType =
+    withInjector(FileHttpFetcherModule()) { implicit injector =>
+      injector.getInstance(fetcherClass)
+    }
```
leogrim, line 19: This introduces a shared state between your tests, and if you need more injected classes you'll end up using two different injectors in each of them (each call to `withInjector` creates one). You should be able to do this instead:
`def articleFetcher(implicit injector: Injector): FetcherType = injector.getInstance(fetcherClass)`

Then pass `FileHttpFetcherModule()` to `withInjector` in each of your tests. It's not as DRY but makes dependencies and injector creation explicit for each test. 

Then you don't have to extend `RoverTestInjector` and have this class be a mere helper `trait ArticleFetcherTestHelper[A <: Article, FetcherType <: ArticleFetcher[A]] { self: Specification => ... }`

What do you think?

kifi-backend/modules/rover/test/com/keepit/rover/article/fetcher/DefaultArticleFetcherTest.scala
```diff
@@ -0,0 +1,70 @@
...
+      normalization.alternateUrls === Set("http://www.bbc.co.uk/news/technology-25233230", "http://www.bbc.com/news/technology-25233230")
+      normalization.openGraphUrl === Some("http://www.bbc.co.uk/news/technology-25233230")
+    }
+
+    "ignore canonical URLs that are the page URL HTML-escaped twice" in {
+      val longUrl = "http://www.livejournal.com/gsearch?engine=google&cx=partner-pub-5600223439108080%3A3711723852&cof=FORID%3A10&ie=UTF-8&q=test&sa=Search&siteurl="
+      val scraped = fetch("double-escape.txt")
+      val normalization = scraped.content.normalization
+      normalization.canonicalUrl === None
+    }
```
leogrim, line 56: Weird I'm not sure this should be the case, I'm heading for the train, I'll take a closer look when I get in the office :)

PR #16016, opened 2015-07-01
[Do not merge] Skeleton for Proto orgs

ryanpbrewster:
@leogrim take a look. Right now there are no tests or analysis capabilities.

Will probably change "ProtoOrganization" to "OrganizationPrototype" or something along those lines.

@andrewconner @eishay


kifi-backend/modules/shoebox/app/com/keepit/commanders/ProtoOrganizationCommander.scala
```diff
@@ -0,0 +1,53 @@
...
+      case Left(fail) => Left(fail)
+      case Right(createResponse) =>
+        val org = createResponse.newOrg
+        val inviteRequests = invites.collect {
+          case user if user.userId.nonEmpty => OrganizationMemberInvitation(invited = Left(user.userId.get), role = OrganizationRole.MEMBER)
+          case email if email.emailAddress.nonEmpty => OrganizationMemberInvitation(invited = Right(email.emailAddress.get), role = OrganizationRole.MEMBER)
+        }
+        implicit val context = heimdalContextBuilder().build
+        val inviteResult = orgInviteCommander.inviteToOrganization(orgId = org.id.get, inviterId = ownerId, invitees = inviteRequests)
+        Await.result(inviteResult, 10 seconds) match {
```
leogrim: But why?

ryanpbrewster: I wrote four things in response, but none of them were good reasons. We'll await the result in whatever controller calls this method. I'll change this to a map.

leogrim: No need to `Await` ever, in the controller, just use `Action.async` :)

kifi-backend/modules/common/app/com/keepit/model/ProtoOrganizationInvite.scala
```diff
@@ -0,0 +1,40 @@
+package com.keepit.model
+
+import com.keepit.common.db.{ Id, ModelWithState, State, States }
+import com.keepit.common.mail.EmailAddress
+import com.keepit.common.time._
+import org.joda.time.DateTime
+import play.api.libs.functional.syntax._
+import play.api.libs.json._
+
+case class ProtoOrganizationInvite(
```
leogrim: Should we call this a `ProtoOrganizationMembership` instead, so that the concept is self-consistent?

Then they're turned into `OrganizationInvites` (except for the owner I guess, who should just be notified?)

ryanpbrewster: I wanted to allow for email invites. The owner is automatically part of the
org when the create request goes through. The rest of the members get
invited.

I briefly considered having separate memberships and invites. Do you think
we want to instantiate an organization and automatically give people
membership? Currently the only way you can add a member to an organization
(on the front end) is by inviting them and having them accept.

leogrim: Just suggesting a name change, since these do represent memberships to the ProtoOrg, then they translate into invites for the real org. This is great, I like invites, they'll get a notification email and everything :)

PR #16020, opened 2015-07-01
Added a JsObject key-based json formatter

ryanpbrewster:
Example use: deserializing
    `val input = """[ {"name": "Alice", "age": 24},
                        {"name": "Bob", "age": 42} ]"""`

The format you want is `tuple2KeyFormat`, so something like
`val format = tuple2KeyFormat[String, Int]("name", "age")`
`val info: Seq[(String,Int)] = Json.parse(input).as[Seq[(String, Int)]](format)`

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/common/json/package.scala
+++ b/kifi-backend/modules/common/app/com/keepit/common/json/package.scala
@@ -5,6 +5,49 @@ import play.api.data.validation.ValidationError
 
 package object json {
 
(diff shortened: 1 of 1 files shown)


ryanpbrewster:
@leogrim

yingjieMiao:
cool! mind write a test?

ryanpbrewster:
Man would I love to. ((I actually already did, but didn't realize there was
a place to put it in the codebase.))

PR #16022, opened 2015-07-02
Fix centering of the settings gear icon

cvializ:
@andrewconner @ajkochanowicz


adamjgrant:
:+1: Boom goes the dynamite

PR #16023, opened 2015-07-02
Add Airbrake reporting for before and after Angular loads

cvializ:
@andrewconner @ajkochanowicz 

I decided against using the ngHelperAirbrake because it was just a very thin wrapper that didn't provide enough functionality to justify the extra dependency.

It debounces for 5 seconds, but that's open to change.

### DIFF
--- a/kifi-backend/modules/shoebox/angular/dev.html
+++ b/kifi-backend/modules/shoebox/angular/dev.html
@@ -19,6 +19,23 @@
   mixpanel.init("abb7e1226370392c849ec16fadff2584");*/
 </script>
 <script src="/dist/lib.js"></script>
+<!--
+<script>
+  // Uncomment this script to test the Airbrake reporting functionality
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/shoebox/angular/dev.html
```diff
@@ -19,6 +19,23 @@
   mixpanel.init("abb7e1226370392c849ec16fadff2584");*/
 </script>
 <script src="/dist/lib.js"></script>
+<!--
```
andrewconner, line 22: Great. Make sure to add this to `index.html` as well.

andrewconner, line 22: I'm blind.

andrewconner:
Looks good. I'm sure we're about to be surprised with what comes out.

kifi-backend/modules/shoebox/angular/dev.html
cvializ, line 22: ; )

PR #16024, opened 2015-07-02
Fix color of gear icon to match neighboring icons

cvializ:
@andrewconner @ajkochanowicz 

I noticed that the gear icon from the mock was gray, but the rest of the header icons were white, but `:not(:hover) { opacity: .7; }` to make them look gray when not hovered.

So I made the gear svg white and changed some names to reflect that.


adamjgrant:
Nice. I'm hoping in the future, we can just add a `.icon-white` class or something that applies a standard `fill` value.

:+1:

PR #16026, opened 2015-07-02
Turns out airbrake doesn't include a minified source. Quick fix.

andrewconner:
@cvializ I noticed a JS exception in prod, turns out they don't include a `.min.js`, so it was just failing. This is the quick fix, should gzip okay so isn't awful, but we'd prefer to use a minified source. Gulp may need to do this for us, or we manage this lib ourselves.


andrewconner:
Self merging to get fix out.

andrewconner:
Our gulpfile should probably cry a bit when a file doesn't exist, would catch this.

PR #16030, opened 2015-07-02
[do not merge] Proto orgs

ryanpbrewster:
@leogrim 

No analysis capabilities in there yet. Also no tests. And no evolutions. It doesn't touch any other code, though.


kifi-backend/modules/common/conf/evolutions/shoebox/350.sql
```diff
@@ -0,0 +1,35 @@
...
+	id bigint(20) NOT NULL AUTO_INCREMENT,
+	created_at datetime NOT NULL,
+	updated_at datetime NOT NULL,
+	state varchar(20) NOT NULL,
+	name varchar(256) NOT NULL,
+	description text DEFAULT NULL,
+	owner_id bigint(20) NOT NULL,
+
+	PRIMARY KEY(id),
+	INDEX `proto_organization_i_seq` (`seq`),
```
yingjieMiao: field `seq` is not defined.

kifi-backend/modules/common/conf/evolutions/shoebox/350.sql
```diff
@@ -0,0 +1,35 @@
...
+
+CREATE TABLE proto_organization_membership (
+	id bigint(20) NOT NULL AUTO_INCREMENT,
+	created_at datetime NOT NULL,
+	updated_at datetime NOT NULL,
+	state varchar(20) NOT NULL,
+	proto_organization_id bigint(20) NOT NULL,
+	user_id bigint(20) DEFAULT NULL,
+	email_address varchar(512) DEFAULT NULL,
+
```
yingjieMiao, line 26: do you want to have a unique key constraint on `(proto_org_id, user_id)`?

kifi-backend/modules/shoebox/app/com/keepit/commanders/ProtoOrganizationCommander.scala
```diff
@@ -0,0 +1,71 @@
...
+    //orgInviteCommander: OrganizationInviteCommander,
+    implicit val executionContext: ExecutionContext,
+    heimdalContextBuilder: HeimdalContextBuilderFactory) extends ProtoOrganizationCommander with Logging {
+
+  def createProtoOrganization(ownerId: Id[User], name: String): ProtoOrganization = {
+    db.readWrite { implicit session => protoOrgRepo.save(ProtoOrganization(ownerId = ownerId, name = name)) }
+  }
+  def addMembers(protoOrgId: Id[ProtoOrganization], userIds: Set[Id[User]]): Unit = {
+    db.readWrite { implicit session =>
+      val existingMemberIds = protoOrgMembershipRepo.getAllByProtoOrganization(protoOrgId).collect {
```
yingjieMiao, line 34: do we want to retrieve inactive memberships, and reset them to active? This is necessary if we have a unique key constraint on (protoOrgId, userId).

kifi-backend/modules/shoebox/app/com/keepit/commanders/ProtoOrganizationCommander.scala
```diff
@@ -0,0 +1,71 @@
...
+          case email if email.emailAddress.nonEmpty => OrganizationMemberInvitation(invited = Right(email.emailAddress.get), role = OrganizationRole.MEMBER)
+        }
+        /*
+        implicit val context = heimdalContextBuilder().build
+        val inviteResult = orgInviteCommander.inviteToOrganization(orgId = org.id.get, inviterId = ownerId, invitees = inviteRequests)
+        inviteResult map {
+          case Left(fail) => Left(fail)
+          case Right(_) => Right(createResponse)
+        }
+        */
```
yingjieMiao, line 67: comments??

ryanpbrewster, line 67: I am chasing through dependency injection and can't get it to run without
errors.

kifi-backend/modules/common/conf/evolutions/shoebox/350.sql
ryanpbrewster, line 26: That's not a bad idea. I am not a huge fan of the additional complexity it
imposes in the commander, though.

kifi-backend/modules/common/conf/evolutions/shoebox/350.sql
ryanpbrewster: Good catch, I was getting rid of those for now. Once everything works we can go back in and add `seq` for these things where desired.

kifi-backend/modules/common/conf/evolutions/shoebox/350.sql
yingjieMiao, line 26: uniqueness can actually simplify logics. It ensures data integrity, prevent/catch race conditions. e.g. you could have duplicated `(proto_org_id, user_id)` pairs that are all active (likely b/c of race condition). This gives you trouble in counting membership.

ryanpbrewster, line 26: You've convinced me.

PR #16036, opened 2015-07-02
Stop trying to mutate a read-only field, throws an exception

andrewconner:
@cvializ Airbraking found this one :)


cvializ:
That's awesome, I'm interested to see what else might turn up. The PR looks good too, :+1:

PR #16039, opened 2015-07-02
Report HTTP error responses via Airbrake

cvializ:
@andrewconner @ajkochanowicz 

Note: It doesn't check status codes, because Angular _should_ only pass error responses to the responseError interceptor.


andrewconner:
Nice! I did just realize that Forbidden (403) should be ignored. That can happen very easily, like a user not being logged in, invites expiring, etc.

andrewconner:
Merge whenever you're happy.

PR #16040, opened 2015-07-02
org mem indexing

yingjieMiao:
@leogrim 

next steps: augment `KeepIndexable` and `LibraryIndexable` with `orgId`. So that we can update `VisibilityEvaluator`. 

Those are largely orthogonal to this PR, I think.


kifi-backend/modules/search/app/com/keepit/search/index/graph/organization/OrganizationMembershipIndexable.scala
```diff
@@ -0,0 +1,44 @@
...
+import com.keepit.search.index.{ Searcher, Indexable, FieldDecoder }
+import com.keepit.search.util.LongArraySet
+import org.apache.lucene.index.Term
+
+object OrganizationMembershipFields {
+  val userField = "u"
+  val userIdField = "uid"
+  val orgField = "o"
+  val orgIdField = "oid"
+
```
leogrim, line 14: Perhaps we should at least index the user's role? (I mean a value field)

leogrim:
Awesome! We probably want convenience methods defined on `OrganizationMembershipIndexable`, similar to those defined on `LibraryMembershipIndexable` :)

kifi-backend/modules/search/app/com/keepit/search/index/graph/organization/OrganizationMembershipIndexable.scala
yingjieMiao, line 14: from visibility point of view, does role make a difference? For current usage, the field will not be used, right?

yingjieMiao:
convenience methods will be added as they are needed.

kifi-backend/modules/search/app/com/keepit/search/index/graph/organization/OrganizationMembershipIndexable.scala
```diff
@@ -0,0 +1,44 @@
...
+  val decoders: Map[String, FieldDecoder] = Map.empty
+}
+
+object OrganizationMembershipIndexable {
+  import OrganizationMembershipFields._
+
+  def getOrgsByMember(orgMemSearcher: Searcher, memberId: Id[User]): Set[Long] = {
+    LongArraySet.from(orgMemSearcher.findSecondaryIds(new Term(userField, memberId.id.toString), orgIdField).toArray())
+  }
```
yingjieMiao, line 23: one convenient method so far ...

PR #16041, opened 2015-07-02
private keeps/libraries, deleted keeps/libraries filtered from grat email

camhashemi:
@andrewconner also includes the fix for the airbrake you sent me earlier, I would appreciate a review of the batching logic.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/emails/GratificationCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/emails/GratificationCommander.scala
@@ -31,8 +31,6 @@ class GratificationCommander @Inject() (
   private val NUM_WEEKS_BACK = 1
   val UNDER_EXPERIMENT = false
 
-  private val remoteCallQueue = new ReactiveLock(numConcurrent = 5)
-
   private val BATCH_SIZE = 1000
 
   def getLibraryFollowCounts(userId: Id[User]): CountData[Library] = {
@@ -57,15 +55,16 @@ class GratificationCommander @Inject() (
 
     def processBatch(dummyAcc: Unit, batch: Int): Future[Unit] = {
       val userIds = generateUserBatch(batch)
-      val fGratDatas = remoteCallQueue.withLockFuture(heimdal.getEligibleGratDatas(userIds)).map(_.map(augmentData))
-      fGratDatas.map { gratDatas => emailSenderProvider.gratification.sendToUsersWithData(gratDatas); Future.successful(()) }
+      val fGratDatas = heimdal.getEligibleGratDatas(userIds).map(_.map(augmentData))
+      fGratDatas.map { gratDatas => emailSenderProvider.gratification.sendToUsersWithData(gratDatas); () }
     }
 
     if (!UNDER_EXPERIMENT) {
       FutureHelpers.foldLeft(0 to numBatches)(())(processBatch)
     } else {
       val userIds = db.readOnlyReplica { implicit session => userExperimentRepo.getUserIdsByExperiment(ExperimentType.GRATIFICATION_EMAIL) }
-      remoteCallQueue.withLockFuture(heimdal.getEligibleGratDatas(userIds)).map(_.map(augmentData))
+      val fGratDatas = heimdal.getEligibleGratDatas(userIds).map(_.map(augmentData))
+      fGratDatas.map { gratDatas => emailSenderProvider.gratification.sendToUsersWithData(gratDatas); () }
     }
 
   }
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/emails/GratificationCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/emails/GratificationCommander.scala
@@ -3,9 +3,9 @@ package com.keepit.commanders.emails
 import com.google.inject.{ Singleton, Inject }
 import com.keepit.commanders.LocalUserExperimentCommander
 import com.keepit.common.concurrent.ReactiveLock
-import com.keepit.common.db.Id
+import com.keepit.common.db.{ Model, Id }
 import com.keepit.common.db.slick.DBSession.RSession
-import com.keepit.common.db.slick.Database
+import com.keepit.common.db.slick.{ Repo, DbRepo, Database }
 import com.keepit.common.logging.Logging
 import com.keepit.common.mail.{ ElectronicMail, EmailAddress }
 import com.keepit.heimdal.HeimdalServiceClient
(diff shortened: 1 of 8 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/emails/GratificationCommander.scala
```diff
@@ -71,34 +63,46 @@ class GratificationCommander @Inject() (
     val userCount = db.readOnlyReplica { implicit s => userRepo.count }
     val numBatches = userCount / BATCH_SIZE
     (0 to numBatches).foreach { batchNum =>
```
camhashemi: @andrewconner this is the batching logic in question, it coordinates with w/ userRepo.pageIncluding. 

the exception was due to processing an empty list received from generateUserBatch, now we ignore those.

kifi-backend/modules/shoebox/app/com/keepit/commanders/emails/GratificationCommander.scala
```diff
@@ -44,14 +44,6 @@ class GratificationCommander @Inject() (
...
-    val since = currentDateTime.minusWeeks(NUM_WEEKS_BACK)
-    val newConnections = db.readOnlyReplica { implicit s =>
-      userConnectionRepo.getConnectionsSince(userId, since)
-    }
-    newConnections.toSeq
-  }
-
   def getGratData(userId: Id[User]): Future[GratificationData] = heimdal.getGratData(userId).map { augmentData }
 
   def generateUserBatch(batchNum: Int): Seq[Id[User]] = {
```
andrewconner, line 48: This could double send emails. `userRepo.pageIncluding` orders by id desc, so if someone registers while this is going on, the person on the batch border gets double sent. In practice, right now, the window is small, but... (incoming second comment)

kifi-backend/modules/shoebox/app/com/keepit/commanders/emails/GratificationCommander.scala
andrewconner: This appears to immediately load every userId, opening ~90 futures at once. ie, you generate all batches and make requests to `getEligibleGratDatas` at once. This will likely cause problems now, and if not, will pretty soon.

A safer approach is:
1. Generate the first batch
2. Fetch Grat data
3. Send emails
4. Iterate

You can hand-roll this, but `FutureHelpers.whilef` may be an easy ticket.

camhashemi:
@andrewconner updated. I love how FutureHelpers uses recursion in the callback to sequentially process futures w/o blocking... putting that in my bag of tricks.

the next cron is going to turn off the experiment and send me everything as a dress rehearsal. please let me know if you see anything wrong with this.

kifi-backend/modules/shoebox/app/com/keepit/commanders/emails/GratificationCommander.scala
```diff
@@ -44,61 +45,48 @@ class GratificationCommander @Inject() (
...
-      fGratData.onComplete {
-        case Success(gratDatas) =>
-          log.info(s"Grat Data batch retrieval succeeded: batchNum=$batchNum, sending emails")
-          emailSenderProvider.gratification.sendToUsersWithData(gratDatas, sendTo)
-        case Failure(t) => log.error(s"Grat Data batch retrieval failed for batchNum $batchNum. Exception: ${t.getMessage}", t)
-      }
+
+    def processBatch(dummyAcc: Unit, batch: Int): Future[Unit] = {
+      val userIds = generateUserBatch(batch)
+      val fGratDatas = remoteCallQueue.withLockFuture(heimdal.getEligibleGratDatas(userIds)).map(_.map(augmentData))
```
andrewconner: The reactive lock may not be needed anymore, actually

kifi-backend/modules/shoebox/app/com/keepit/commanders/emails/GratificationCommander.scala
```diff
@@ -44,61 +45,48 @@ class GratificationCommander @Inject() (
...
+      fGratDatas.map { gratDatas => emailSenderProvider.gratification.sendToUsersWithData(gratDatas); Future.successful(()) }
```
andrewconner: This will compile, but does `Future.successful(())` do anything? You're already in a Future (`fGratDatas`)?

kifi-backend/modules/shoebox/app/com/keepit/commanders/emails/GratificationCommander.scala
```diff
@@ -44,61 +45,48 @@ class GratificationCommander @Inject() (
...
+    }
+
+    if (!UNDER_EXPERIMENT) {
+      FutureHelpers.foldLeft(0 to numBatches)(())(processBatch)
+    } else {
+      val userIds = db.readOnlyReplica { implicit session => userExperimentRepo.getUserIdsByExperiment(ExperimentType.GRATIFICATION_EMAIL) }
+      remoteCallQueue.withLockFuture(heimdal.getEligibleGratDatas(userIds)).map(_.map(augmentData))
```
andrewconner: Does this do anything? Doesn't call `processBatch`

kifi-backend/modules/sqldb/app/com/keepit/common/db/slick/DbRepo.scala
```diff
@@ -99,13 +100,20 @@ trait DbRepo[M <: Model[M]] extends Repo[M] with FortyTwoGenericTypeMappers with
...
   def page(page: Int = 0, size: Int = 20, excludeStates: Set[State[M]] = Set.empty[State[M]])(implicit session: RSession): Seq[M] = {
-    // todo(Andrew): When Slick 2.1 is released, convert to Compiled query (upgrade necessary for .take & .drop)
+    // todo(Andrew): When Slick 2.2 is released, convert to Compiled query (upgrade necessary for .take & .drop)
     val q = for {
       t <- rows if !t.state.inSet(excludeStates)
     } yield t
     q.sortBy(_.id desc).drop(page * size).take(size).list
   }
 
+  def pageAscendingIds(page: Int = 0, size: Int = 20, excludeStates: Set[State[M]] = Set.empty[State[M]])(implicit session: RSession): Seq[Id[M]] = {
```
andrewconner: I see `page` does this wrong, but the default parameters here do nothing, since you're implementing a trait. The trait implements the method wrappers.

Default params in Scala are implemented by the compiler writing code for you:

``` scala
def add(a: Int = 0, b: Int = 0) = a+b
// compiles to:
def add() = add(0,0)
def add(a: Int) = add(a, 0)
def add(a: Int, b: Int) = a+b
```

andrewconner: Further, not sure `page: Int` should have a default param. Can't think of when I want the oldest 20, and only want that. Make them ask for which batch, explicit > implicit.

andrewconner:
Cool, merge when you're happy with it

PR #16048, opened 2015-07-02
Change the header hover-menus to click-menus.

cvializ:
@andrewconner @ajkochanowicz 

Feedback welcome. Here are some decisions I made:
1. Change the links to buttons. I though it was more semantic, since the click should now only open the dropdown, and the link was duplicated in the list of action items anyway.
2. Clicking a second time on the icon doesn't close the dropdown. We could have it so a second click closes the dropdown, but I know some people double-click _everything_ and that might confuse those users.

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/header/loggedInHeader.tpl.html
+++ b/kifi-backend/modules/shoebox/angular/src/header/loggedInHeader.tpl.html
@@ -31,16 +31,16 @@
     <div class="kf-lih-right kifi-guide-target-3 kifi-guide-2">
       <button class="kf-lih-add-link sprite sprite-add-link" ng-click="addKeeps()">Add a link</button>
       <a href="/tags/manage" class="kf-lih-tags"><span class="svg-tag-white"></span>Manage tags</a>
-      <div class="kf-lih-dropdown-parent" kf-hover-menu>
-        <a href="{{me|profileUrl:'connections'}}" class="kf-lih-friends"><span class="svg-friends-white"></span>Connections</a>
+      <div class="kf-lih-dropdown-parent" kf-click-menu>
+        <button class="kf-lih-friends"><span class="svg-friends-white"></span>Connections</button>
         <menu class="kf-dropdown-menu">
           <a class="kf-dropdown-menu-item" href="{{me|profileUrl:'connections'}}">Your Kifi connections</a>
           <a class="kf-dropdown-menu-item" href="/invite">Invite someone to Kifi</a>
         </menu>
       </div>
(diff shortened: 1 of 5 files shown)


cvializ:
Also: are there other hover-menus we want to change? This PR only includes the header.

andrewconner:
Awesome. How about turning this guy into a click menu as well on library pages?



The current behavior is kind of weird. If you don't have a library image, it opens a 1-element menu (bah!). It should just open the upload dialog immediately. If you _have_ an image, then do the menu. Thoughts?

kifi-backend/modules/shoebox/angular/src/header/loggedInHeader.tpl.html
```diff
@@ -31,16 +31,16 @@
     <div class="kf-lih-right kifi-guide-target-3 kifi-guide-2">
       <button class="kf-lih-add-link sprite sprite-add-link" ng-click="addKeeps()">Add a link</button>
       <a href="/tags/manage" class="kf-lih-tags"><span class="svg-tag-white"></span>Manage tags</a>
-      <div class="kf-lih-dropdown-parent" kf-hover-menu>
-        <a href="{{me|profileUrl:'connections'}}" class="kf-lih-friends"><span class="svg-friends-white"></span>Connections</a>
+      <div class="kf-lih-dropdown-parent" kf-click-menu>
+        <button class="kf-lih-friends"><span class="svg-friends-white"></span>Connections</button>
```
adamjgrant, line 35: Good idea.

adamjgrant:
Nice thinking using a button for semantically indicating a click interaction.

I agree with Andrew, it's a little more legwork, but it did bother me when I saw only one item in the dropdown. But if it's a lot of work, that might be best isolated into a separate task.

cvializ:
Hmm ok, I'll take a look at the one-item-dropdown and decide if it's too big for this task

cvializ:
@andrewconner @ajkochanowicz 

Can you please take another look and let me know what you think before I merge it? I changed the event listener setup so we wouldn't have to deal with synchronizing the click-in/click-out listeners.

cvializ:
Bah I'm feeling good about it

adamjgrant:
Why is the event listener on document?
Le Thu, Jul 2, 2015 à 8:40 PM, Carlos Vializ notifications@github.com a
écrit :

cvializ:
So that I don't need to add a separate event listener to the document to detect when the user clicks away. Debugging was a mess when I was doing that

adamjgrant:
Yeah, you definitely don't want an event listener for each item. You're
smart to put a check for e.target instead.
Le Thu, Jul 2, 2015 à 8:48 PM, Carlos Vializ notifications@github.com a
écrit :

PR #16049, opened 2015-07-02
User checklist backend

andrewconner:
@leogrim This is the backend for a user-facing checklist feature (to get users to experience more of Kifi).


andrewconner:
I want this live so Carlos can use it, but reviews appreciated.

leogrim:
:+1: 
Seems like the randomization of the next tasks to tackle could be complementary to the current user progression, rather than stand on their own - more of a product question I guess.

PR #16050, opened 2015-07-03
Rover url rule

maxov:
Adds url rule capability to the Rover, as well as a bunch of new admin views.
- Add proxy capability to rover
- Add tests for proxy capability in rover
- Remove old HttpProxy
- Remove old UrlPatternRule
- Add proxy admin page for rover
- Add regex tester page for rover
- Add url viewer page for rover

@leogrim


kifi-backend/modules/rover/app/com/keepit/rover/rule/UrlRuleAction.scala
```diff
@@ -0,0 +1,36 @@
...
+import com.keepit.common.db.Id
+import com.keepit.model.HttpProxy
+
+sealed trait UrlRuleAction {
+
+  def &(that: UrlRuleAction): UrlRuleAction = UrlRuleAction.And(List(this, that))
+  
+}
+
+object UrlRuleAction {
```
maxov, line 12: This is an idea for a "pretty" dsl like so, instead of relying on multiple fields within the RoverUrlRule class:

``` scala
UrlRuleAction.UseProxy(proxyId) & UrlRuleAction.NormalizeScheme(normalizationScheme) & UrlRuleAction.StaySensitive(true)
```

Ideally these objects would be persisted directly into the database somehow (I mentioned formatting them as a JSON string), but I'm not sure if it's worth the effort given the low amount of url rules. It would also require a substantial amount of code for the admin view to assemble these actions in a nice way. Same goes for the filters.

As it is RoverUrlRule retains the previous, flat structure, these classes aren't actually used anywhere.

kifi-backend/modules/common/app/com/keepit/rover/model/UrlRule.scala
```diff
@@ -0,0 +1,29 @@
...
+
+import com.keepit.common.db.{ States, State, Id }
+
+import play.api.libs.json._
+import play.api.libs.functional.syntax._
+
+case class UrlRule(
+    id: Option[Id[UrlRule]] = None,
+    state: State[UrlRule] = UrlRuleStates.ACTIVE,
+    pattern: String,
```
leogrim, line 11: Could you add the example column here? It's pretty convenient to force people to test their regex, others to understand it.

kifi-backend/modules/common/app/com/keepit/shoebox/ShoeboxScraperClient.scala
```diff
@@ -21,41 +21,17 @@ import com.keepit.common.cache.TransactionalCaching.Implicits.directCacheAccess
...
-      if (r.json == null) None else r.json.asOpt[HttpProxy]
-    }
-  }
-
-  def getProxyP(url: String): Future[Option[HttpProxy]] = limiter.withLockFuture {
-    call(Shoebox.internal.getProxyP, Json.toJson(url), callTimeouts = longTimeout, routingStrategy = offlinePriority).map { r =>
-      if (r.json == null) None else r.json.asOpt[HttpProxy]
-    }
-  }
 }
```
leogrim: Can we get rid of that class at this point?

kifi-backend/modules/common/conf/evolutions/shoebox/353.sql
```diff
@@ -0,0 +1,18 @@
...
+    created_at datetime NOT NULL,
+    updated_at datetime NOT NULL,
+    pattern varchar(2048) NOT NULL,
+    proxy_id bigint(20) NULL,
+
+    PRIMARY KEY (id),
+    CONSTRAINT rover_url_rule_proxy FOREIGN KEY (proxy_id) REFERENCES rover_http_proxy(id)
+);
+
+insert into evolutions (name, description) values('353.sql', 'create rover_url_rule table');
```
leogrim, line 17: You also want to drop the proxy table and unnecessary columns for UrlPatternRule in Shoebox

kifi-backend/modules/rover/test/com/keepit/rover/article/fetcher/LinkedInProfileArticleFetcherTest.scala
```diff
@@ -0,0 +1,28 @@
...
+        val scraped = fetch("https://www.linkedin.com/in/leogrim", "linkedin_profile.txt")
+        val profile = scraped.content.profile
+
+        profile.id === Some("17558679")
+        profile.title must containKeywords("Grimaldi Software Engineer Kifi")
+        profile.overview must containKeywords("San Francisco Bay Area Stanford")
+        profile.sections must containKeywords("San Francisco Bay Area Scala Engineering Education Technicolor technical answer")
+      }
+
+    }
```
leogrim, line 25: Awesome!

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminRoverController.scala
```diff
@@ -0,0 +1,145 @@
...
+    implicit val executionContext: ExecutionContext) extends AdminUserActions {
+
+  def searchUrl = AdminUserPage { implicit request =>
+    Ok(views.html.admin.roverSearchUrl())
+  }
+
+  def findUrl = AdminUserPage(parse.urlFormEncoded) { implicit request =>
+    val url = request.body.get("url").get.head
+    db.readOnlyReplica { implicit session =>
+      urlRepo.get(url) match {
```
leogrim: You should use `NormalizedURIInterner.getByUri` here, `UrlRepo` is actually going away.

kifi-backend/modules/shoebox/app/com/keepit/model/UrlPatternRuleRepo.scala
```diff
@@ -1,55 +0,0 @@
...
-
-  def getUrlPatternRules()(implicit session: RSession): UrlPatternRules = {
-    urlPatternRulesAllCache.getOrElse(UrlPatternRulesAllKey()) {
-      val result = (for (f <- rows if f.state === UrlPatternRuleStates.ACTIVE) yield f).list
-      val sortedResult = result.sortBy(_.id.get.id)
-      UrlPatternRules(sortedResult)
-    }
-  }
-}
-
```
leogrim, line 55: I'm sorry we got a misunderstanding, some of these are still needed:

`normalization`
`trustedDomain`
`nonSensitive`

PR #16056, opened 2015-07-06
ansible support instance-type specific configuration of jvm attributes

joshm1:
@leogrim @andrewconner 

when we want JVM Xmx/Xms variables to differ by instance type, just create a file under `roles/kifi.common/vars` under the format `<service>/<instance_type>.yml` where instance_type replaces  periods with underscores (e.g., `rover/m3_xlarge.yml`)

if an instance-type specific file doesn't exist, it will look for a file under `vars/<service>.yml`.  If that file doesn't exist, it will use the default value that's in the init.d script.

i included some config files for abook, rover, and search to demonstrate


andrewconner:
Cool, nice

PR #16058, opened 2015-07-06
Proto Organizations

ryanpbrewster:
A proto-Organization is a placeholder organization that holds all the
information necessary to create a real organization (e.g., the owner, name,
and a list of members to invite). There are two db tables for proto orgs,
one for the proto-orgs themselves and another for all the members associated
with them.


ryanpbrewster:
@leogrim, I opted to keep the name `ProtoOrganization` because `OrganizationPrototype` is, in my opinion, too close to actual classes like `OrganzationMembership` and `OrganizationInvite`. I want anything that starts with `Organization` to be associated with real orgs.

leogrim:
It is meant to become a real `Organization` though, so it makes sense to me to keep the name scheme in line with that ecosystem, don't you think?

leogrim:
I don't feel strongly about it actually, `ProtoOrganization` just feels hard to say :)
More importantly, we should discuss tracking continuity from a proto-org to and org, primary ids will be different so we may want to introduce a stable tracking identifier. What do you think?

ryanpbrewster:
What if we assign the proto-org a handle?

kifi-backend/modules/shoebox/app/com/keepit/commanders/ProtoOrganizationCommander.scala
```diff
@@ -0,0 +1,92 @@
...
+  def addMembers(protoOrgId: Id[ProtoOrganization], userIds: Set[Id[User]]): Future[Unit] = SafeFuture {
+    db.readWrite { implicit session =>
+      val existingMemberships = protoOrgMembershipRepo.getAllByProtoOrganization(protoOrgId, states = ProtoOrganizationMembershipStates.all).filter {
+        _.userId.nonEmpty
+      }.toSet
+
+      val inactiveMemberships = existingMemberships.filter(_.state == ProtoOrganizationMembershipStates.INACTIVE)
+      inactiveMemberships.foreach { m => protoOrgMembershipRepo.save(m.withState(ProtoOrganizationMembershipStates.ACTIVE)) }
+
+      val newUserIds = userIds -- existingMemberships.filter(_.userId.isDefined).map(_.userId.get)
```
yingjieMiao: `existingMemberships.flatMap{_.userId}` ?

kifi-backend/modules/shoebox/app/com/keepit/commanders/ProtoOrganizationCommander.scala
```diff
@@ -0,0 +1,92 @@
...
+
+      newUserIds.foreach { uid =>
+        protoOrgMembershipRepo.save(ProtoOrganizationMembership(protoOrgId = protoOrgId, userId = Some(uid)))
+      }
+    }
+  }
+  def removeMembers(protoOrgId: Id[ProtoOrganization], userIds: Set[Id[User]]): Future[Unit] = SafeFuture {
```
yingjieMiao: this is not yet in the trait.

kifi-backend/modules/shoebox/app/com/keepit/commanders/ProtoOrganizationCommander.scala
```diff
@@ -0,0 +1,92 @@
...
+    db.readWrite { implicit session =>
+      val existingMemberships = protoOrgMembershipRepo.getAllByProtoOrganization(protoOrgId, states = ProtoOrganizationMembershipStates.all).filter {
```
yingjieMiao: you can just retrieve active members, instead of `all`.

PR #16059, opened 2015-07-06
Resolving undefined file type error

adamjgrant:


adamjgrant:
@andrewconner @cvializ Gaze upon its glory

kifi-backend/modules/shoebox/angular/src/libraries/libraryHeader.js
```diff
@@ -129,7 +129,7 @@ angular.module('kifi')
 
         scope.onCoverImageFileChosen = function (files) {
           var file = files[0];
-          if (/^image\/(?:jpeg|png|gif)$/.test(file.type)) {
+          if (file && /^image\/(?:jpeg|png|gif)$/.test(file.type)) {
```
adamjgrant, line 132: Shout out to my homie @cvializ

cvializ:
:+1: : D

PR #16060, opened 2015-07-06
visibility logic with org

yingjieMiao:
@leogrim


kifi-backend/modules/search/app/com/keepit/search/engine/VisibilityEvaluator.scala
```diff
@@ -91,6 +100,8 @@ final class KeepVisibilityEvaluator(
           Visibility.FOLLOWER // the keep is in a library I am a member of
         }
       }
+    } else if (orgIds.findIndex(orgIdDocValues.get(docId)) >= 0) { // keep is owned by an org that I am a member of
+      Visibility.NETWORK
     } else if (authorizedLibraryIds.findIndex(libId) >= 0) {
       Visibility.FOLLOWER // the keep is in an authorized library
```
leogrim, line 104: These two `else if` should be swapped, since having an explicit authorization to access a library is treated like a membership (also, `Follower` is stronger than `Network`, as you walk down these conditionals, visibility levels should decrease).

kifi-backend/modules/search/app/com/keepit/search/engine/library/LibraryFromUserScoreVectorSource.scala
```diff
@@ -24,6 +24,7 @@ class LibraryFromUserScoreVectorSource(
     val friendIdsFuture: Future[Set[Long]],
     val restrictedUserIdsFuture: Future[Set[Long]],
     val libraryIdsFuture: Future[(Set[Long], Set[Long], Set[Long], Set[Long])],
+    protected val orgIdsFuture: Future[Set[Long]],
```
leogrim, line 27: Should all of these be protected then?

yingjieMiao, line 27: could be. they are protected elsewhere.

kifi-backend/modules/search/app/com/keepit/search/engine/VisibilityEvaluator.scala
yingjieMiao, line 104: good point.

kifi-backend/modules/search/app/com/keepit/search/index/graph/keep/KeepRecord.scala
```diff
@@ -27,6 +27,7 @@ object KeepRecord {
     out.writeString(record.externalId.id)
     out.writeString(record.note.getOrElse(""))
     out.writeStringSet(record.tags.map(_.tag))
+    out.writeLong(record.orgId.map { _.id }.getOrElse(-1L))
```
leogrim: I don't think that's necessary, since there's already a value field in the document to read it from.

kifi-backend/modules/search/app/com/keepit/search/index/graph/library/LibraryIndexable.scala
```diff
@@ -128,6 +130,10 @@ class LibraryIndexable(library: DetailedLibraryView) extends Indexable[Library,
     doc.add(buildLongValueField(visibilityField, Visibility.toNumericCode(library.visibility)))
     doc.add(buildLongValueField(kindField, Kind.toNumericCode(library.kind)))
 
+    val orgId = library.orgId.getOrElse(Id[Organization](-1))
+    doc.add(buildKeywordField(orgField, orgId.id.toString))
```
leogrim: I don't think this one needs `-1`, it's an indexed field so we do not need to create a long posting list for a fake term. We just need `-1` for the value field below.

yingjieMiao: you are right.

kifi-backend/modules/search/app/com/keepit/search/index/graph/library/LibraryRecord.scala
```diff
@@ -5,22 +5,23 @@ import com.keepit.model._
 import java.io.{ ByteArrayInputStream, ByteArrayOutputStream }
 import org.apache.lucene.store.{ InputStreamDataInput, OutputStreamDataOutput }
 
-case class LibraryRecord(id: Id[Library], name: String, description: Option[String], color: Option[LibraryColor], ownerId: Id[User], slug: LibrarySlug)
+case class LibraryRecord(id: Id[Library], name: String, description: Option[String], color: Option[LibraryColor], ownerId: Id[User], slug: LibrarySlug, orgId: Option[Id[Organization]])
```
leogrim, line 8: Could we have the `orgId` field before `ownerId` for logical consistency? Since this is usually how we reason about it: "get the org or else get the owner"?

leogrim:
Awesome, great job!
We probably want to update the user search with a `UserFromOrganizationsVectorSource` in order to:
- search a user by organization metadata (kind of like we do for `UserFromLibariesVectorSource`)
- load all members of your organizations with visibility `Network` (just like we do for uris)

Then I would change `UserVisibilityEvaluator` to categorize your direct connections as `Owner` rather than `Network`. 

Makes sense?

leogrim:
Actually, looks like `UserFromLibariesVectorSource` may need to be updated to load not just owners but collaborators of your libraries, this is related to collaborative libraries though, unrelated to orgs.

kifi-backend/modules/search/app/com/keepit/search/index/graph/keep/KeepRecord.scala
yingjieMiao: sure. same argument applies to `LibraryRecord`?

leogrim: I see, `LibraryRecord` already has `ownerId` even though it is a value field. I guess we should go for consistency here, either we drop them both for `LibraryRecord` or have them both.

yingjieMiao:
`UserFromOrganizationsVectorSource` seems to be a separate issue and a separate PR.

leogrim:
Yep definitely :)
Just thought of it when taking a look at `UserVisibilityEvaluator` :)

kifi-backend/modules/search/app/com/keepit/search/index/graph/keep/KeepRecord.scala
yingjieMiao: let's drop both of them if they don't provide any value.

yingjieMiao:
@leogrim   updated. please see the last commit.

PR #16063, opened 2015-07-06
fix npe?

yingjieMiao:
@leogrim 

old indexer gives null


leogrim:
Ok, remember to remove this when the new indices are in prod :)

yingjieMiao:
yes

PR #16066, opened 2015-07-06
First look at user checklist implementation

cvializ:
@ajkochanowicz @andrewconner 

Hey this is my first effort at the new user checklist. It isn't ready to merge, but I want to get some feedback about my approach. I'm new to angular so I want to make sure I didn't make any n00b mistakes : P

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/common/services/routeService.js
+++ b/kifi-backend/modules/shoebox/angular/src/common/services/routeService.js
@@ -62,11 +62,15 @@ angular.module('kifi')
         return route('/collections/' + tagId + '/undelete');
       },
 
+      followingLibraries: 'http://support.kifi.com/hc/en-us/articles/202657599-Following-Libraries',
+
       whoToInvite: route('/user/invite/recommended'),
       blockWtiConnection: route('/user/invite/hide'),
       friendRequest: function (id) {
         return env.xhrBase + '/user/' + id + '/friend';
       },
+      socialInvite: route('/invite'),
+      connectTwitter: route('/twitter/request'),
       invite: route('/user/invite'),
       peopleYouMayKnow: function (offset, limit) {
         return route('/user/friends/recommended', {offset: offset, limit: limit});
--- a/kifi-backend/modules/shoebox/angular/src/common/services/routeService.js
+++ b/kifi-backend/modules/shoebox/angular/src/common/services/routeService.js
@@ -69,8 +69,8 @@ angular.module('kifi')
       friendRequest: function (id) {
         return env.xhrBase + '/user/' + id + '/friend';
       },
-      socialInvite: route('/invite'),
-      connectTwitter: route('/twitter/request'),
+      socialInvite: '/invite',
+      connectTwitter: '/twitter/request',
       invite: route('/user/invite'),
       peopleYouMayKnow: function (offset, limit) {
         return route('/user/friends/recommended', {offset: offset, limit: limit});
--- a/kifi-backend/modules/shoebox/angular/src/recos/userChecklist.js
+++ b/kifi-backend/modules/shoebox/angular/src/recos/userChecklist.js
@@ -3,8 +3,8 @@
 angular.module('kifi')
 
 .directive('kfUserChecklist', [
-  '$rootScope', '$window', '$location', '$analytics', 'installService', 'extensionLiaison', 'modalService', 'profileService',
-  function ($rootScope, $window, $location, $analytics, installService, extensionLiaison, modalService, profileService) {
+  '$rootScope', '$window', '$location', '$analytics', 'installService', 'extensionLiaison', 'modalService', 'profileService', 'net',
+  function ($rootScope, $window, $location, $analytics, installService, extensionLiaison, modalService, profileService, net) {
     return {
       restrict: 'A',
       replace: true,
@@ -94,7 +94,13 @@ angular.module('kifi')
           {
             name: 'install_mobile',
             title: 'Get the mobile app on iOS or Android',
-            subtitle: 'Text a link to your phone for Kifi on the go'
+            subtitle: 'Text a link to your phone for Kifi on the go',
+            action: function () {
+              modalService.open({
+                template: 'common/modal/sendMobileAppSMS.tpl.html',
+                scope: scope
+              });
+            }
           },
           {
             name: 'twitter_sync',
@@ -114,6 +120,21 @@ angular.module('kifi')
           }
           return true;
         }
+
+        scope.sms = { phoneNumber: '' };
+
+        scope.triggerSendSMS = function () {
+          net.sendMobileAppSMS({ phoneNumber: scope.sms.phoneNumber })
+            .catch(function () {
+              modalService.open({
+                template: 'common/modal/sendMobileAppSMSError.tpl.html'
+              });
+            });
+
+          // Clear the model so the modal doesn't show the phone number
+          // if the user opens it again later
+          scope.sms.phoneNumber = '';
+        };
 
         scope.$watchCollection(function () {
           return profileService.prefs && profileService.prefs.checklist;
--- a/kifi-backend/modules/shoebox/angular/src/recos/userChecklist.js
+++ b/kifi-backend/modules/shoebox/angular/src/recos/userChecklist.js
@@ -3,8 +3,8 @@
 angular.module('kifi')
 
 .directive('kfUserChecklist', [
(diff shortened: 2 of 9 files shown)


kifi-backend/modules/shoebox/angular/src/recos/userChecklist.js
```diff
@@ -0,0 +1,145 @@
...
+      scope: {},
+      templateUrl: 'recos/userChecklist.tpl.html',
+      link: function (scope, element) {
+        var checklistItems = [
+          {
+            name: 'invite_friends',
+            title: 'Invite three friends to Kifi',
+            subtitle: 'Blah subtitle',
+            action: function () {
+              $location.path('/invite');
```
adamjgrant: Is there a way we could use one of the routing services for this, in case the name of the route changes?

kifi-backend/modules/shoebox/angular/src/recos/userChecklist.js
```diff
@@ -0,0 +1,145 @@
...
+            }
+          },
+          {
+            name: 'keep_pages',
+            title: 'Keep five pages',
+            subtitle: 'Blah subtitle'
```
andrewconner: @JRuff42 Do we have this support page? Or dropping this one?

kifi-backend/modules/shoebox/angular/src/recos/userChecklist.js
```diff
@@ -0,0 +1,145 @@
...
+          },
+          {
+            name: 'follow_libs',
+            title: 'Follow five libraries',
+            subtitle: 'Blah subtitle',
+            action: function () {
+              $window.location.href = 'http://support.kifi.com/hc/en-us/articles/202657599-Following-Libraries';
```
andrewconner: How about a new window?

kifi-backend/modules/shoebox/angular/src/recos/userChecklist.js
```diff
@@ -0,0 +1,145 @@
...
+            }
+          },
+          {
+            name: 'install_ext',
+            title: 'Get the Kifi browser add-on',
+            subtitle: 'The most loved features are in the add-on',
+            action: function () {
+              $location.path('/install');
```
andrewconner: Check out `installService.triggerInstall`. Much friendlier experience (they stay on the page).

kifi-backend/modules/shoebox/angular/src/recos/userChecklist.js
```diff
@@ -0,0 +1,145 @@
...
+                $rootScope.$emit('showGlobalModal', 'importBookmarkFile');
+                $analytics.eventTrack('user_viewed_page', {
+                  'type': '3rdPartyImport'
+                });
+            }
+          },
+          {
+            name: 'install_mobile',
+            title: 'Get the mobile app',
+            subtitle: 'blah blah'
```
andrewconner: For this one, you could do a simple modal allowing them to enter their phone number. We have a backend endpoint that will SMS a link to the app store.

kifi-backend/modules/shoebox/angular/src/recos/userChecklist.js
```diff
@@ -0,0 +1,145 @@
...
+          },
+          {
+            name: 'twitter_sync',
+            title: 'Connect your Twitter account',
+            subtitle: 'blah blah'
```
andrewconner: Go to `https://www.kifi.com/twitter/request`, backend will handle the rest.

kifi-backend/modules/shoebox/angular/src/recos/userChecklist.js
```diff
@@ -0,0 +1,145 @@
...
+          }
+        ];
+
+        profileService.fetchPrefs().then(function (prefs) {
```
andrewconner: Unfortunately this causes another request, and an existing request should already exist (nearly always concurrently, since it happens automatically when the page opens). If you want to stick to the promise approach, you'll want to cache requests. I wrote a pretty simple wrapper called `Clutch` that handles this, see `profileService`'s `meService`. It may be overkill though. Two alternatives:
1. Watch for changes:

``` javascript
scope.$watchGroup(function () {
  return profileService.prefs.checklist;
}, function (checklist) {
  // ...
});
```
1. Listen to a message:

``` javascript
$rootScope.$on('prefsChanged', function () {
  // ...
});
```

Honestly, not sure why the 2nd exists, but I notice that it is firing an event when prefs is updated.

kifi-backend/modules/shoebox/angular/src/recos/userChecklist.tpl.html
```diff
@@ -0,0 +1,10 @@
+<div class="kf-uc">
```
andrewconner: _Usually_ in Angular apps, instead of `element.remove()` above, it's convention to do something like `ng-if="checklist.length > 0"` in the template so it's declarative, vs relying on the directive to side-effect and remove the node.

kifi-backend/modules/shoebox/angular/src/recos/userChecklist.js
```diff
@@ -0,0 +1,145 @@
...
+            }
+          }
+
+          // Don't show the checklist if all items are compelte
+          if (allComplete) {
+            element.remove();
+            return;
+          }
+
+          scope.checklist = checklistItems.map(function (checklistItem) {
```
andrewconner: I might be reading this wrong, but you might want to map the other way (ie, map on `profileChecklist`, and filter out ones you don't know about). This would preserve the order coming from the backend.

andrewconner:
Nice!

kifi-backend/modules/shoebox/angular/src/recos/userChecklist.tpl.html
cvializ: Makes sense. I noticed this creates a new scope so I have to reference $parent.checklist elsewhere in the template, which is weird.

kifi-backend/modules/shoebox/angular/src/recos/userChecklist.js
cvializ: I saw the meService, but it did seem like overkill for this use-case.

For some reason `scope.$watchGroup` or `scope.$watch` have refused to work for me, but `$rootScope.$on` works fine.

kifi-backend/modules/shoebox/angular/src/recos/userChecklist.tpl.html
andrewconner: Do you? Scopes prototypically inherit, so it should just work.

kifi-backend/modules/shoebox/angular/src/recos/userChecklist.js
andrewconner: Strange, we can pair it in a few.

kifi-backend/modules/shoebox/angular/src/recos/userChecklist.js
cvializ: Makes sense, I'll do it that way

kifi-backend/modules/shoebox/angular/src/recos/userChecklist.js
cvializ: Good idea, I went ahead and added this

kifi-backend/modules/shoebox/angular/src/common/services/routeService.js
```diff
@@ -62,11 +62,15 @@ angular.module('kifi')
...
 
+      followingLibraries: 'http://support.kifi.com/hc/en-us/articles/202657599-Following-Libraries',
+
       whoToInvite: route('/user/invite/recommended'),
       blockWtiConnection: route('/user/invite/hide'),
       friendRequest: function (id) {
         return env.xhrBase + '/user/' + id + '/friend';
       },
+      socialInvite: route('/invite'),
+      connectTwitter: route('/twitter/request'),
```
andrewconner: I think `route()` prepends `/site/`. You may want just the raw string

andrewconner: Same for `socialInvite`

cvializ: Got it, I had a feeling it might do that, but I wasn't sure and it was
working in local

adamjgrant:
Great work.

cvializ:
Thank you : D

PR #16070, opened 2015-07-06
Page updates by id

andrewconner:
@jaredpetker wrote most of this. This changes his PR from paging by time to paging by ID. This is better because of ties (same `keptAt` for potentially hundreds of keeps) which happens most commonly during keep imports.

@leogrim


kifi-backend/modules/shoebox/app/com/keepit/model/KeepRepo.scala
```diff
@@ -524,15 +523,21 @@ class KeepRepoImpl @Inject() (
     keeps.list.toSet
   }
 
-  def getRecentKeepsFromFollowedLibraries(userId: Id[User], num: Int)(implicit session: RSession): Seq[Keep] = {
-    import com.keepit.common.db.slick.StaticQueryFixed.interpolation
-    sql"""SELECT #$bookmarkColumnOrder FROM bookmark bm WHERE library_id IN (SELECT library_id FROM library_membership WHERE user_id=$userId AND state='active') AND state='active' AND user_id!=$userId ORDER BY kept_at DESC LIMIT $num;""".as[Keep].list
-  }
-
-  def getRecentKeepsFromFollowedLibraries(userId: Id[User], num: Int, beforeTime: String)(implicit session: RSession): Seq[Keep] = {
+  def getRecentKeepsFromFollowedLibraries(userId: Id[User], num: Int, beforeIdOpt: Option[ExternalId[Keep]], afterIdOpt: Option[ExternalId[Keep]])(implicit session: RSession): Seq[Keep] = {
```
leogrim: Can we start using `limit` instead of `num` or `count` to be consistent with other such methods? Colin actually introduced `Limit` and `Offset` types to also avoid unintended swapping using these parameters.

andrewconner: Sure.

jaredpetker:
+1

PR #16073, opened 2015-07-07
add few companies to the user ip blacklist

eishay:


eishay:
@ryanpbrewster

andrewconner:
Looks good to me

PR #16074, opened 2015-07-07
Quick fix to erroneously showing compact/gallery view buttons

adamjgrant:
@andrewconner @cvializ


adamjgrant:
D'oh. There might be another one of these to fix up some remaining issues but I want to hide these for non-admin users for now.

andrewconner:
Closing because of conflicts with my revert. Feel free to overwrite when you're good.

PR #16076, opened 2015-07-07
Feature/keep list view

adamjgrant:
@andrewconner @cvializ


andrewconner:
Looks good

kifi-backend/modules/shoebox/angular/src/libraries/library.js
```diff
@@ -93,7 +93,7 @@ angular.module('kifi')
         moveToLibrary: true
       }
     };
-    $scope.galleryView = true;
+    $scope.galleryView = !profileService.prefs.use_minimal_keep_card;
```
andrewconner, line 96: :+1:

PR #16079, opened 2015-07-07
created an endpoint to manually test-send scheduled emails

camhashemi:
@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/emails/EmailTemplateProcessor.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/emails/EmailTemplateProcessor.scala
@@ -176,62 +176,66 @@ class EmailTemplateProcessorImpl @Inject() (
   }
 
   private def evalTemplate(text: String, input: DataNeededResult, emailToSend: EmailToSend, emailTipOpt: Option[EmailTip]): String = {
-    tagRegex.replaceAllIn(text, { rMatch =>
-      val tagWrapper = Json.parse(rMatch.group(1)).as[TagWrapper]
-      val tagArgs = tagWrapper.args
+    try {
+      tagRegex.replaceAllIn(text, { rMatch =>
+        val tagWrapper = Json.parse(rMatch.group(1)).as[TagWrapper]
+        val tagArgs = tagWrapper.args
 
-      // only call if Id[User] is expected as the first argument
-      @inline def userId = jsValueAsUserId(tagArgs(0))
-      @inline def user = input.users(userId)
-      @inline def basicUser = BasicUser.fromUser(user)
+        // only call if Id[User] is expected as the first argument
+        @inline def userId = jsValueAsUserId(tagArgs(0))
+        @inline def user = input.users(userId)
+        @inline def basicUser = BasicUser.fromUser(user)
 
-      // only call if Id[Library] is expected as the first argument
-      @inline def libraryId = tagArgs(0).as[Id[Library]]
-      @inline def library: Library = input.libraries(libraryId)
+        // only call if Id[Library] is expected as the first argument
+        @inline def libraryId = tagArgs(0).as[Id[Library]]
+        @inline def library: Library = input.libraries(libraryId)
 
-      @inline def keepId = tagArgs(0).as[Id[Keep]]
-      @inline def keep: Keep = input.keeps(keepId)
+        @inline def keepId = tagArgs(0).as[Id[Keep]]
+        @inline def keep: Keep = input.keeps(keepId)
 
-      tagWrapper.label match {
-        case tags.firstName => basicUser.firstName
-        case tags.lastName => basicUser.lastName
-        case tags.fullName => basicUser.firstName + " " + basicUser.lastName
-        case tags.avatarUrl => toHttpsUrl(input.imageUrls(userId))
-        case tags.profileUrl => config.applicationBaseUrl + "/" + basicUser.username.value
-        case tags.libraryUrl =>
-          config.applicationBaseUrl + libPathCommander.getPath(library)
-        case tags.libraryName => library.name
-        case tags.libraryOwnerFullName =>
-          val libOwner = input.users(library.ownerId)
-          libOwner.fullName
-        case tags.keepName => keep.title.getOrElse("this keep")
-        case tags.keepUrl => keep.url
-        case tags.unsubscribeUrl =>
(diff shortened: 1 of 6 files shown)


kifi-backend/modules/shoebox/app/com/keepit/controllers/internal/EmailTestController.scala
```diff
@@ -93,6 +94,14 @@ class EmailTestController @Inject() (
       getOrElse(Future.successful(BadRequest(s"test sender for $name not found")))
   }
 
+  def testBatchSending(name: String, sendTo: String) = Action.async { request =>
+    name match {
+      case "gratification" => gratificationCommander.batchSendEmails({ id: Id[User] => true }, Some(EmailAddress(sendTo)))
+      case "activity" => emailSenderProvider.activityFeed.apply(Some(EmailAddress(sendTo)))
+    }
+    Future.successful(Ok("sent"))
```
andrewconner, line 102: Doesn't really matter since this is a throwaway endpoint, but the `Future.successful` is only necessary because you used `Action.async` above — it's not actually waiting on any result to return. If you use just `Action`, you can just use `Ok("sent")` directly (though, it'll return that immediately, not when it's sent)

kifi-backend/modules/shoebox/app/com/keepit/commanders/emails/EmailTemplateProcessor.scala
```diff
@@ -176,62 +176,66 @@ class EmailTemplateProcessorImpl @Inject() (
...
+            EmailTrackingParam(
+              subAction = Json.fromJson[String](tagArgs(0)).asOpt,
+              variableComponents = Seq.empty, // todo(josh) this needs to be passed in EmailToSend
+              tip = emailTipOpt,
+              auxiliaryData = emailToSend.auxiliaryData
+            ).encode
+        }
+      })
+    } catch {
+      case ex: IndexOutOfBoundsException => log.error(s"[EmailTemplateProcessor] Index OOB exception. Text: $text"); throw ex
```
andrewconner, line 237: Weird. If this is possible, we should probably refactor any explicit getter that may cause it.

andrewconner:
Looks good, but you can probably skip actually sending emails for this. Not sure the best place to do that or if it would be a massive pain.

kifi-backend/modules/shoebox/app/com/keepit/commanders/emails/EmailTemplateProcessor.scala
camhashemi, line 237: This bug came up when I added keepName and keepUrl tags to the email template, where tagRegex.replaceAllIn is throwing it for a few users (e.g. Eishay) and (mostly) not others. @joshm1 suggested that since replaceAllIn is probably not the cause, I need to see what inputs are triggering it (I added tests for the new tags, I think certain keeps might be doing it).

joshm1, line 237: @camhashemi I suggested it was probably from the `replaceAllIn` call, at least by looking at the stack trace, it looks like the bug originates in `replaceAllIn` and the exception is thrown from a Java stdlib class.  It wasn't abundantly clear what could cause this. I tried running several replaceAllIn just in sbt console and couldn't get the same error, so like I suggested, it's probably low-cost debugging to log `text` and re-throw the exception.  I think we'll see immediately what the cause is when we know what `text` is.  One nitpick though, you can make the change set smaller without having to re-indent the entire method by wrapping the existing curly braces in `try { .. } catch`. I think, if possible, it's best to avoid changes lines because the distorts git diff/blame and makes it harder to trace which change(s) may have broke something. 

``` scala
private def evalTemplate(text: String, input: DataNeededResult, emailToSend: EmailToSend, emailTipOpt: Option[EmailTip]): String = try {
  //existing implementation unchanged
} catch {
  // handle
}
```

joshm1:
If we wanted to skip sending the emails, I'd suggest putting a `dryRun = false` (default) in `EmailToSend` and modifying `EmailTemplateSender` to not send the email if that is true. You'd probably need to add that field to `ProcessedEmailResult` as well and have it derived from `EmailToSend`

kifi-backend/modules/shoebox/app/com/keepit/commanders/emails/EmailTemplateProcessor.scala
camhashemi, line 237: That's an awesome tip, thank you.

camhashemi:
@andrewconner @joshm1 an alternative (and probably easier) option is to not change anything. I can manually test the data pipeline via b01, and manually test the email composition via `EmailTestController`. That way test emails are sent one by one, no load is put on the mail server, and we can test the pipeline by batch as well. It will probably make debugging easier by separating the two issues.

Still want to add the logging though.

kifi-backend/modules/shoebox/app/com/keepit/commanders/emails/EmailTemplateProcessor.scala
yingjieMiao, line 237: `replaceAllIn` may have issues with special characters like `$`

camhashemi, line 237: @yingjieMiao duly noted, but in the logging output I couldn't find a special character outside of <%kf% (tag) %kf%>. I'm hoping it was because I wasn't limiting the volume of tags in the email (should be limited to 7 per category), but not sure why that would be the case.

camhashemi:
@joshm1 @yingjieMiao 

I was able to get a fix by limiting the volume of tags in the email. For example, before we would show the keep names of 75 keeps that Eishay had views on (a bug, I forgot to edit the text email with the html). When I limited volume to the top 7 keeps, no bug. Any idea why?

joshm1:
@camhashemi no idea, that's strange. Did you get the `text` from email(s) that threw this exception? I'd be surprised if only the # of groups matched in the regexp was the sole culprit.

camhashemi:
@joshm1 I didn't see any obvious problems, other than 68 keepIds listed. When I fixed that, I could then process Eishay's email with no problems. It could be that one of those groups that I cut was causing the error. If you're curious, searching for '/IOOB' since 7/8 on Logentries will bring up Eishay's `text` that threw the exception (you have to 'view context').

It could be that there are other users with the same problem waiting for me to find, once I get the cron firing.

PR #16080, opened 2015-07-07
adding more stats to user cluster channel

eishay:


eishay:
@andrewconner @ryanpbrewster

PR #16081, opened 2015-07-07
placing ip tracking clustering reports behind an actor

eishay:
1. don't waste lots of futures and possibly create starvation if there's lots of reporting checks to do
2. avoid race condition (dups etc)


eishay:
@andrewconner @ryanpbrewster

ryanpbrewster:
Nice, I like this change. I've had a lot of trouble with race-like behavior. Hopefully this mitigates some of that.

PR #16087, opened 2015-07-07
Resolving 'Cannot set property 'h' of undefined' error in Airbrake

adamjgrant:
This should also resolve the "undefined 'l' variable" error.


adamjgrant:
@andrewconner @cvializ

andrewconner:
Great

andrewconner:
`undefined` should have a `h`. I've always felt that way.

PR #16088, opened 2015-07-07
Adding sourcemap to JS minification workflow

adamjgrant:
Want to give this a try, @andrewconner @cvializ ?


adamjgrant:
Theoretically, this will automatically start working in Airbrake once deployed.

andrewconner:
Does this expose our pre-uglified source?

andrewconner:
But sure, let's try it.

adamjgrant:
No, just the .map file.

So if you run `gulp assets:local-prod:rev` you'll see kifi.min.js.map and lib.min.js.map in dist

PR #16090, opened 2015-07-07
Org admin page

ryanpbrewster:
@leogrim @andrewconner 

Not all the info is actually displayed on the pages, but it's all there. Still need to do Javascript magic to let admins set things like the org name, add candidate members, invite members, transfer ownership, etc.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/admin/AdminOrganizationController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/admin/AdminOrganizationController.scala
@@ -21,17 +21,7 @@ class AdminOrganizationController @Inject() (
     organizationMembershipCandidateCommander: OrganizationMembershipCandidateCommander,
     organizationInviteCommander: OrganizationInviteCommander) extends AdminUserActions {
 
-  def populateDatabase() = AdminUserPage { implicit request =>
-    val ryan = userCommander.createUser("Ryan", "Brewster", addrOpt = None, state = UserStates.ACTIVE)
-    val users = for (i <- 1 to 20) yield {
-      userCommander.createUser(i.toString, "TestUser", addrOpt = None, state = UserStates.ACTIVE)
-    }
-    val kifiCreateResponse = orgCommander.createOrganization(OrganizationCreateRequest(ryan.id.get, OrganizationModifications(name = Some("Kifi"), description = Some("Knowledge Sharing"))))
-    val bcCreateResponse = orgCommander.createOrganization(OrganizationCreateRequest(ryan.id.get, OrganizationModifications(name = Some("Brewster Corp"), description = Some("Taking over the world!"))))
-    val kifi = kifiCreateResponse.right.get.newOrg
-    organizationMembershipCandidateCommander.addCandidates(kifi.id.get, users.map(_.id.get).toSet)
-    Ok("done")
-  }
+  private val fakeOwnerId = Id[User](97543) // "Fake Owner", a special private Kifi user specifically for this purpose
 
   def organizationsView = AdminUserPage { implicit request =>
     val orgIds = orgCommander.getAllOrganizationIds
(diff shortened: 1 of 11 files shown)


kifi-backend/modules/shoebox/app/com/keepit/admin/AdminUserController.scala
```diff
@@ -9,7 +9,7 @@ import scala.concurrent.{ Await, Future, Promise }
 import scala.concurrent.duration.{ Duration, DurationInt }
 import scala.util.{ Failure, Success, Try }
 
-import com.google.inject.Inject
+import com.google.inject.{ Inject, Singleton }
```
ryanpbrewster, line 12: Are controllers supposed to be Singletons?

kifi-backend/modules/shoebox/conf/shoeboxService.routes
```diff
@@ -675,6 +675,11 @@ GET    /admin/users/mixpanel/delete          @com.keepit.controllers.admin.Admin
 POST    /admin/user/:id/emails/:code/send    @com.keepit.controllers.admin.AdminUserController.sendEmail(id: Id[User], code: String)
 POST    /admin/user/:id/refreshRecos    @com.keepit.controllers.admin.AdminUserController.refreshRecos(id: Id[User])
 
+
+GET     /admin/populatedb             @com.keepit.controllers.admin.AdminOrganizationController.populateDatabase()
```
ryanpbrewster: I am considering making a real route like this. I found it extremely useful for testing the admin HTML pages. Thoughts?

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminOrganizationController.scala
```diff
@@ -0,0 +1,61 @@
...
+    orgMembershipCandidateRepo: OrganizationMembershipCandidateRepo,
+    orgCommander: OrganizationCommander,
+    orgMembershipCommander: OrganizationMembershipCommander,
+    organizationMembershipCandidateCommander: OrganizationMembershipCandidateCommander,
+    organizationInviteCommander: OrganizationInviteCommander) extends AdminUserActions {
+
+  def populateDatabase() = AdminUserPage { implicit request =>
+    val ryan = userCommander.createUser("Ryan", "Brewster", addrOpt = None, state = UserStates.ACTIVE)
+    val users = for (i <- 1 to 20) yield {
+      userCommander.createUser(i.toString, "TestUser", addrOpt = None, state = UserStates.ACTIVE)
```
yingjieMiao: unit test/dev mode is more appropriate for this type of thing. Also, users created this way are not marked as fake users.

leogrim: Agreed, this seems dangerous, you could call this endpoint in production and create a bunch of fake data.

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminUserController.scala
leogrim, line 12: I don't think so, but I don't remember if there's a good reason for them not to be @andrewconner

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
```diff
@@ -35,24 +42,33 @@ class OrganizationCommanderImpl @Inject() (
...
 
-      val avatarPath = organizationAvatarCommander.getBestImage(orgId, ImageSize(200, 200)).map(_.imagePath)
+    val avatarPath = organizationAvatarCommander.getBestImage(orgId, ImageSize(200, 200)).map(_.imagePath)
 
-      // TODO: actually find the number of libraries
-      val numPublicLibs = libraries(LibraryVisibility.PUBLISHED)
-      OrganizationView(orgId = orgId, handle = orgHandle, name = orgName, description = description, avatarPath = avatarPath, members = members,
-        numMembers = memberCount, numLibraries = numPublicLibs)
-    }
+    // TODO: actually find the number of libraries
```
leogrim: Even better, find the number of libraries for each `LibraryVisibility` (hiding secret ones like we do for keeps), this should help us understand organization's main use cases for libraries. I can see Colin already introduced a `LibraryRepo.countLibrariesForOrgByVisibility`

leogrim: Whoops, this is already what you're using and I thought this was `Admin`. Here you're right, you want to personalize this count for the viewer :)

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
```diff
@@ -210,4 +221,39 @@ class OrganizationCommanderImpl @Inject() (
       }
     }
   }
+
+  // For use in the Admin Organization controller. Don't use it elsewhere.
+  def unsafeModifyOrganization(orgId: Id[Organization], modifications: OrganizationModifications): Unit = {
```
leogrim: What about passing in the requester's info and double check for `Admin` experiment?

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
```diff
@@ -210,4 +221,39 @@ class OrganizationCommanderImpl @Inject() (
...
+        val memberships = orgMembershipRepo.getAllByOrgId(org.id.get)
+        applyNewBasePermissionsToMembers(memberships, org.basePermissions, modifiedOrg.basePermissions)
+      }
+    }
+  }
+
+  def getAnalyticsView(orgId: Id[Organization]): AnalyticsOrganizationView = {
+    db.readOnlyReplica { implicit session =>
+      val orgView = getOrganizationViewHelper(orgId)
+      // TODO: get actual numbers
```
leogrim: Note: we have the convention of adding our name to todo's, e.g. `Todo(Léo)`. Increases responsibility for authors and traceability for readers :)

ryanpbrewster: That's a good idea.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
ryanpbrewster: :+1:

PR #16092, opened 2015-07-07
DRY visibility evaluator construction

yingjieMiao:
@leogrim 

reduce repetition. visibility related codes got more centralized as well. 

makes sense?


leogrim:
Love it, thanks!

PR #16099, opened 2015-07-07
Only hiding controls for main and secret libraries

andrewconner:
Persona based libraries (ones created automatically during registration) are technically "system" libraries because the user didn't create them. However, we can let users edit settings and such. This changes what we define as "system".

@ajkochanowicz @cvializ


adamjgrant:
:+1: looks good.

PR #16100, opened 2015-07-07
Sourcemaps will no longer minify

adamjgrant:
.map files should no longer have revision numbers

### DIFF
--- a/kifi-backend/modules/shoebox/angular/gulpfile.js
+++ b/kifi-backend/modules/shoebox/angular/gulpfile.js
@@ -28,7 +28,7 @@ var modRewrite = require('connect-modrewrite');
 var gutil = require('gulp-util');
 var karma = require('karma').server;
 var protractor = require('gulp-protractor').protractor;
-var revall = require('gulp-rev-all');
+var RevAll = require('gulp-rev-all');
 var awspublish = require('gulp-awspublish');
 var parallelize = require('concurrent-transform');
 var through = require('through');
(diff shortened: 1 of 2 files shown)


adamjgrant:
@andrewconner @cvializ

andrewconner:
Cool. Only at concern is that our CDN uploads are set to cache forever. See:

```
    .pipe(parallelize(publisher.publish({'Cache-Control': 'max-age=315360000, no-transform, public'}), 50))
```

which is a great feature. How could we get the updated sourcemaps. Or, could we have them hashed, and dynamically set the sourcemap reference?

adamjgrant:
Will airbrake be looking for this on the CDN?
Le Tue, Jul 7, 2015 à 11:52 PM, Andrew Conner notifications@github.com a
écrit :

adamjgrant:
Nvm. I realized the link is relative. Damn

kifi-backend/modules/shoebox/angular/gulpfile.js
```diff
@@ -443,10 +443,11 @@ gulp.task('test', function (done) {
 });
 
 function compileAssetRevs(opts, dest) {
-  opts = merge({ ignore: [ /^\/favicon.ico$/g, /^sprites\//g ], hashLength: 7 }, opts || {});
+  opts = merge({ dontRenameFile: [ /^\/favicon.ico$/g, /^sprites\//g ], hashLength: 7 }, opts || {});
```
adamjgrant, line 446: Syntax changed in new verison.

adamjgrant:
I think we're good. @andrewconner @cvializ ?

kifi-backend/modules/shoebox/angular/gulpfile.js
cvializ, line 446: Ugh that is the worst when that happens

cvializ:
:+1: looks good to me

PR #16103, opened 2015-07-08
route new connection notification to the user's profile [extension]

camhashemi:
@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/FriendConnectionNotifier.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/FriendConnectionNotifier.scala
@@ -47,7 +47,7 @@ class FriendConnectionNotifier @Inject() (
       title = s"You’re connected with ${respondingUser.firstName} ${respondingUser.lastName} on Kifi!",
       body = s"Enjoy ${respondingUser.firstName}’s keeps in your search results and message ${respondingUser.firstName} directly.",
       linkText = "Invite more friends to kifi",
-      linkUrl = s"https://www.kifi.com/${respondingUser.username}",
+      linkUrl = s"https://www.kifi.com/${respondingUser.username.value}",
       imageUrl = respondingUserImage,
       sticky = false,
       category = category,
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/FriendConnectionNotifier.scala
```diff
@@ -45,9 +45,9 @@ class FriendConnectionNotifier @Inject() (
     val notificationF = elizaServiceClient.sendGlobalNotification( //push needed
       userIds = Set(friendUserId),
       title = s"You’re connected with ${respondingUser.firstName} ${respondingUser.lastName} on Kifi!",
-      body = s"Enjoy ${respondingUser.firstName}’s keeps in your search results and message ${respondingUser.firstName} directly. Find and invite more connections »",
+      body = s"Enjoy ${respondingUser.firstName}’s keeps in your search results and message ${respondingUser.firstName} directly.",
       linkText = "Invite more friends to kifi",
-      linkUrl = "https://www.kifi.com/friends/invite",
+      linkUrl = s"https://www.kifi.com/${respondingUser.username}",
```
andrewconner: @leogrim Is this the best way to do this now?

andrewconner:
Let @leogrim answer that question, and then all good.

kifi-backend/modules/shoebox/app/com/keepit/commanders/FriendConnectionNotifier.scala
camhashemi: I actually have a bug here :-1: need to do `username.value`

leogrim: Yep, you're good.

kifi-backend/modules/shoebox/app/com/keepit/commanders/FriendConnectionNotifier.scala
```diff
@@ -45,9 +45,9 @@ class FriendConnectionNotifier @Inject() (
     val notificationF = elizaServiceClient.sendGlobalNotification( //push needed
       userIds = Set(friendUserId),
       title = s"You’re connected with ${respondingUser.firstName} ${respondingUser.lastName} on Kifi!",
-      body = s"Enjoy ${respondingUser.firstName}’s keeps in your search results and message ${respondingUser.firstName} directly. Find and invite more connections »",
+      body = s"Enjoy ${respondingUser.firstName}’s keeps in your search results and message ${respondingUser.firstName} directly.",
       linkText = "Invite more friends to kifi",
-      linkUrl = "https://www.kifi.com/friends/invite",
+      linkUrl = s"https://www.kifi.com/${respondingUser.username.value}",
```
andrewconner, line 50: re-adding @leogrim. Is this the way, or do we use something with handles?

leogrim, line 50: This is it.

andrewconner:
Looks good then.

PR #16105, opened 2015-07-08
Org Controller cleanup

ryanpbrewster:
Serialization is now done entirely within the OrganizationView (and view-like)
classes. There is pre-emptive support for different versions of serialization.

@leogrim @andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationView.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationView.scala
@@ -1,125 +1,111 @@
 package com.keepit.model
 
-import com.keepit.common.crypto.{ PublicId, PublicIdConfiguration }
-import com.keepit.common.db.{ ExternalId, Id }
-import com.keepit.common.mail.EmailAddress
+import com.keepit.common.crypto.{ PublicIdConfiguration, PublicId }
 import com.keepit.common.store.ImagePath
+import com.keepit.social.BasicUser
 import com.kifi.macros.json
-import play.api.http.Status._
-import play.api.libs.json.Json
-import play.api.mvc.Results.Status
+import play.api.libs.functional.syntax._
+import play.api.libs.json._
 
+// OrganizationView should ONLY contain public information. No internal ids.
+case class OrganizationView(
+  orgId: PublicId[Organization],
+  handle: OrganizationHandle,
+  name: String,
+  description: Option[String],
+  avatarPath: Option[ImagePath],
+  members: Seq[BasicUser],
+  numMembers: Int,
+  numLibraries: Int)
+object OrganizationView {
+  private val defaultWrites: Writes[OrganizationView] = (
+    (__ \ 'id).write[PublicId[Organization]] and
+    (__ \ 'handle).write[OrganizationHandle] and
+    (__ \ 'name).write[String] and
+    (__ \ 'description).writeNullable[String] and
+    (__ \ 'avatarPath).writeNullable[ImagePath] and
+    (__ \ 'members).write[Seq[BasicUser]] and
+    (__ \ 'numMembers).write[Int] and
+    (__ \ 'numLibraries).write[Int]
+  )(unlift(OrganizationView.unapply))
+
+  val website = defaultWrites
+  val mobileV1 = defaultWrites
+}
+
+// OrganizationCard should ONLY contain public information. No internal ids.
+case class OrganizationCard(
+  orgId: PublicId[Organization],
+  handle: OrganizationHandle,
+  name: String,
+  description: Option[String],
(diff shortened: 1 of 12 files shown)


kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationView.scala
```diff
@@ -1,125 +1,111 @@
...
 import com.kifi.macros.json
-import play.api.http.Status._
-import play.api.libs.json.Json
-import play.api.mvc.Results.Status
+import play.api.libs.functional.syntax._
+import play.api.libs.json._
 
-@json case class OrganizationImageInfo(path: ImagePath)
+// OrganizationView should ONLY contain public information. No internal ids.
+case class OrganizationView(
```
andrewconner, line 12: Looking at these two, you think it's worth the complexity overhead to even maintain two classes (View and Card)? Seems like the only difference is members, which is cheap (it's very optimized).

Or rather, if we do like two, why does a card need a description? Or even numLibraries? Just spitballing, but if one is to draw an org and the other is for attribution tagging. Perhaps to seed discussion, here's what other sites do for their minimal group version:

kifi-backend/modules/shoebox/test/com/keepit/commanders/OrganizationMembershipCommanderTest.scala
```diff
@@ -66,8 +63,8 @@ class OrganizationMembershipCommanderTest extends TestKitSupport with Specificat
         val orgMembershipRepo = inject[OrganizationMembershipRepo]
 
         val org = db.readWrite { implicit session =>
-          val org = orgRepo.save(Organization(ownerId = Id[User](1), name = "Luther Corp.", handle = None))
-          orgMembershipRepo.save(org.newMembership(userId = Id[User](1), role = OrganizationRole.OWNER))
+          val owner = UserFactory.user().saved
+          val org = OrganizationFactory.organization().withOwner(owner).withName("Luther Corp.").saved
```
andrewconner, line 67: Wahoo! Test DSLs :+1:

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationView.scala
ryanpbrewster, line 12: Eh...maybe. The thought process that led to multiple "view"-type structures was that for certain use-cases it makes more sense to have a compact view. In the back of my mind was the implicit assumption that "compact view" necessarily meant "more efficient construction", but that's not really the case.

There may be permissions issues, though, where it is reasonable to present a card to someone, but they shouldn't be able to get a full view.

andrewconner, line 12: True. Libraries are like this. We have a full view and a card view. The full view is like your OrganizationView. It's big. Then there's the compact view, "cards". For libraries, we happen to include a decent amount of info in these cards (for design reasons). For orgs, does it make sense? Perhaps an org card could be:

``` scala
case class OrganizationCard(
    id: PublicId[Organization],
    handle: OrganizationHandle,
    name: String,
    avatarPath: ImagePath)
)
```

Thoughts? + @leogrim

leogrim, line 12: Ryan said it well, I'll add to it a bit: we tried to come up with the following convention to avoid the multiplication of models like in libraries:
`ModelView`: for external consumption, display of the model's main full view (i.e. Library's page, Organization's page). That one is the heaviest and may include related content (first page of members & keeps for libraries - first page of members & libraries for organizations)
`ModelCard`: for external consumption, display of one or several models in another context (e.g. attribution, search result). That may typically include some stats.
`IngestableModel`: for internal consumption, e.g. ingestion by `Search`, `Graph`, `Cortex`. Includes internal `Id` and `SequenceNumber` and whatever these other services need.

Coming back to `OrganizationCard`: even if we decide to display a badge, we may want to display a small card with more info on hover (like we do for libraries, people). It might be especially useful in `Search`, whether in the attribution or when returning organizations as search results. In the past, `Search` and `Shoebox` APIs haven't always been in exact sync, mostly because we've been trying to grab as much data from the index as possible (with the initial hope of search servers being independent and replicated in other zones). I've been working on getting them to converge and will make sure we reuse classes in the future, typically these `ModelCard`.

We could introduce a super light level e.g. `OrganizationChip` or `BasicOrganization` that's more like `BasicUser` and doesn't not include anything else than the bare minimum. 

In any case, the idea has been to come up with a convention for these 3-4 levels of details proactively, something that can be reused for other models (`View`, `Card`, `Ingestable`, possibly `Basic`) to avoid polluting the code on the spot with additional classes in the future. 

What do you think?

PR #16106, opened 2015-07-08
Trying again to clean up tags after unkeeping

andrewconner:
@leogrim Earlier fix didn't do it because of notes (I think).

Small performance improvement here as well.


andrewconner:
@leogrim Found the issue. `keepToCollectionRepo.getCollectionsForKeeps` only works on active keeps.

PR #16107, opened 2015-07-08
Fix Safe Search, Tag Search across Kifi

leogrim:
1. This fixes how safe search is enforced. Previously, unsafe content (porn) could surface up if the search hit a specific keep for the page. So porn would show up if kept in your libraries or libraries you follow (expected) but also if a connection kept it (unexpected, a bug I reproduced and that we have luckily not run into). We're now explicitly handling safe / unsafe pages, and this bug is fixed. 
2. The previous implementation prevented us from searching all published keeps - because that could have allowed unsafe content to leak. In particular, this prevented us from looking up all public keeps for a hashtag, Twitter style. This is now fixed, and searching for a tag will find all published keeps for that tag.
3. This also makes it easy to make safe search a setting that you can enable / disable.

What this doesn't fix:
1. Filters. E.g. searching within a library (outstanding bug: search for a tag within a library), within an organization (upcoming) or a user's keeps (not spending time on that last one, though I hope we can spend a little front-end resources on that at some point).
2. This doesn't make keeps from `Main` searchable (I want to).


andrewconner:
Sounds good. I'll assume the code is correct too :)

kifi-backend/modules/search/app/com/keepit/search/engine/uri/UriResultCollector.scala
```diff
@@ -55,15 +55,20 @@ class UriResultCollectorWithBoost(clickBoostsProvider: () => ResultClickBoosts,
...
           queue = networkHits
           actualSharingBoost = 1.0f + sharingBoost - sharingBoost / ctx.degree.toFloat
-        } else {
+        } else if ((visibility & Visibility.SAFE) != 0) {
           queue = othersHits
           actualSharingBoost = 1.0f
+        } else {
+          queue = null
+          actualSharingBoost = 0f
+        }
```
leogrim, line 67: Being very explicit about when safe search kicks in or not.

kifi-backend/modules/search/app/com/keepit/search/engine/VisibilityEvaluator.scala
```diff
@@ -124,7 +127,7 @@ final class KeepVisibilityEvaluator(
         } else if (restrictedUserIds.findIndex(keeperId) >= 0) {
           Visibility.RESTRICTED // explicitly restricted user (e.g. fake user for non-admins)
         } else {
-          Visibility.RESTRICTED // currently not searching published keeps by others
+          Visibility.OTHERS // another published keep
```
leogrim, line 130: And this opens up searching all published keeps (e.g. look up all keeps on the platform for a specific hashtag)

leogrim:
All right, my implementation is wrong as well.

PR #16110, opened 2015-07-08
Do null check to prevent error when image is below 60x40

cvializ:
@andrewconner @ajkochanowicz 

Resolves https://fortytwo.airbrake.io/projects/113130/groups/1459812451957517582/notices/1460205644897234397


cvializ:
I think the error happens when the image is below 60x40, as checked in the condition of the for-loop. This PR prevents the image from appearing at all. Is this the expected behavior? Or should we still show these smaller images in the keep card?

andrewconner:
Nice

adamjgrant:
You got it!

PR #16113, opened 2015-07-08
DRY + Better debug logs

leogrim:


kifi-backend/modules/search/app/com/keepit/search/engine/VisibilityEvaluator.scala
```diff
@@ -96,12 +98,15 @@ final class KeepVisibilityEvaluator(
...
     val userIdDocValues: NumericDocValues,
-    orgIdDocValues: NumericDocValues,
+    val libraryIdDocValues: NumericDocValues,
+    val orgIdDocValues: NumericDocValues,
     visibilityDocValues: NumericDocValues) {
 
   private[this] val published = LibraryFields.Visibility.PUBLISHED
 
-  def apply(docId: Int, libId: Long): Int = {
+  def apply(docId: Int): Int = {
```
yingjieMiao, line 107: haha, I was thinking about this yesterday.

leogrim, line 107: In Library and User searches it may lead us to look up the library id twice, no big deal, these searches aren't performance limiting anyway.

PR #16116, opened 2015-07-08
Organization endpoints for the site

ryanpbrewster:
Almost a direct copy from the Mobile controllers

@andrewconner @leogrim 

Take 2, let's see how it goes.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/OrganizationMembershipController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/OrganizationMembershipController.scala
@@ -1,0 +1,111 @@
+package com.keepit.controllers.website
+
+import com.google.inject.{ Inject, Singleton }
+import com.keepit.commanders.{ UserCommander, OrganizationCommander, OrganizationMembershipCommander }
+import com.keepit.common.controller.{ ShoeboxServiceController, UserActions, UserActionsHelper }
+import com.keepit.common.crypto.{ PublicId, PublicIdConfiguration }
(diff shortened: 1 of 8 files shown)


kifi-backend/modules/shoebox/app/com/keepit/controllers/website/OrganizationMembershipController.scala
```diff
@@ -0,0 +1,111 @@
...
+              case Left(failure) => failure.asErrorResponse
+              case Right(responses) =>
+                val removals = responses.keys.map(r => externalIdMap(r.targetId))
+                Ok(Json.obj("removals" -> removals))
+            }
+        }
+    }
+  }
+
+  def leaveOrganization(pubId: PublicId[Organization]) = UserAction { request =>
```
andrewconner, line 99: `OrganizationAction`?

andrewconner:
Love seeing the tests. Looks good, we'll iterate.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/OrganizationMembershipController.scala
ryanpbrewster, line 99: :sheep: Shucks.

camhashemi:
@ryanpbrewster looks like your IDE re-formatted shoeboxService.routes. I just learned that even if this doesn't have any effect on production, it masks the git blame for those lines. May want to consider reverting that change.

ryanpbrewster:
Damn, nice catch. Github hid that from me.

PR #16118, opened 2015-07-09
Adding svg4everybody and partially reverting for HTTPS issue.

adamjgrant:


adamjgrant:
@andrewconner @cvializ After deploying this, we could resume investigating tomorrow.

PR #16121, opened 2015-07-09
Org pymn: kifi user recommendations and email invite recommendations

camhashemi:
@leogrim @ryanpbrewster


kifi-backend/modules/abook/app/com/keepit/abook/commanders/ABookRecommendationCommander.scala
```diff
@@ -60,6 +60,16 @@ class ABookRecommendationCommander @Inject() (
     futureRecommendations
   }
 
+  def getOrganizationMemberRecommendations(orgId: Id[Organization], offset: Int, limit: Int, bePatient: Boolean = false): Future[Option[Seq[Id[User]]]] = {
+    val start = clock.now()
+    val fRecommendations = generateFutureOrgMemberRecommendations(orgId, bePatient).map(_.map(_.drop(offset).take(limit).map(_._1).toSeq))
```
ryanpbrewster: If this makes sense to you, keep it as is. To me, it's confusing, and some intermediate names would be helpful.

kifi-backend/modules/abook/app/com/keepit/abook/commanders/ABookRecommendationCommander.scala
```diff
@@ -85,6 +95,16 @@ class ABookRecommendationCommander @Inject() (
     futureRecommendations
   }
 
+  def getOrgInviteRecommendations(viewerId: Id[User], orgId: Id[Organization], offset: Int, limit: Int): Future[Seq[OrganizationInviteRecommendation]] = {
```
ryanpbrewster: What is the difference between org member recommendations and org invite recommendations?

camhashemi: The distinction comes from non-org terminology. Friend recommendations ~= member recommendations (i.e. users), and user invite recos ~= org invite recos (i.e. non-users). Probably worth a refactor.

camhashemi: The problem is that "invite" previously meant "invite a user to kifi", whereas in orgs, "invite" applies to both users and non users. How about `getNonUserInviteRecommendationsForOrg`/`getNonUserInviteRecommendationsForUser` and `getUserInviteRecommendationForOrg`/`getFriendRecommendationsForUser`? @leogrim @ryanpbrewster

ryanpbrewster: I like it, but maybe `UserRecosForOrgMember`?

kifi-backend/modules/abook/app/com/keepit/abook/commanders/ABookRecommendationCommander.scala
yingjieMiao: agree. you can still use one liner, but you can replace `_` with meaningful variable names.

kifi-backend/modules/abook/app/com/keepit/abook/commanders/ABookRecommendationCommander.scala
camhashemi: I think the `getWhatforWhom` schema is translatable and understandable for both users and orgs. It's implied that any org invite is to be a member, right?

leogrim:
I would recommend having a separate commander for org and user. 
As in `ABookUserRecommendationCommander` and `ABookOrganizationRecommendationCommander`.

The logics in there are very similar, but do not actually share any code.

@camhashemi I like your naming suggestion, just move away from invite vs connection vs membership semantics, and rely instead on the entity type - just remember how these are generated, a graph: `getRecommendedUsersForUser`, `getRecommendedEmailAccountsForUser`, `getRecommendedUsersForOrganization`, `getRecommendedEmailAccountsForOrganization`.

Then when you bundle users and emails together for org, `getRecommendedMembersForOrganization`. 

I think having separate commanders will ease on the confusion :)

What do you think?

camhashemi:
Very happy to split them up into separate commanders. There is a lot of duplication, so keeping them completely separate will save a lot of headache.

Given that, in practice the call would be `userRecommendationCommander.getRecommendedUsers` or `orgRecommendationCommander.getRecommendedEmailAccounts`, and the `forUsers` or `forOrganization` suffix should be implied? @leogrim I can also just make it explicit...

ryanpbrewster:
If they are separate commanders then it is clear from context. :+1:

kifi-backend/modules/abook/app/com/keepit/abook/commanders/ABookRecommendationCommander.scala
ryanpbrewster: I was looking specifically at `getFriendRecommendationsForUser`. I like `User` more than `Friend`.

leogrim:
Yep that makes sense!

camhashemi:
@leogrim updated.

camhashemi:
@leogrim updated. added fields to `OrganizationMemberRecommendation` and `OrganizationEmailInviteRecommendation`, and changed the evolutions. haven't manually created the tables yet.

kifi-backend/kifi-backend/modules/shoebox/marketing
```diff
@@ -0,0 +1 @@
+Subproject commit 98b8a3de091e072c752d2fbe6b79fe3070673d93
```
camhashemi: this is in error, fixing before merging

kifi-backend/modules/common/app/com/keepit/abook/model/RichSocialConnection.scala
```diff
@@ -57,7 +57,7 @@ object RichSocialConnection {
   )(RichSocialConnection.apply, unlift(RichSocialConnection.unapply))
 }
 
-case class InviteRecommendation(
+case class UserInviteRecommendation(
```
leogrim, line 60: Not sure this renaming is necessary, might be confusing?

kifi-backend/modules/abook/app/com/keepit/abook/commanders/AbookOrganizationRecommendationCommander.scala
```diff
@@ -0,0 +1,202 @@
...
+    oldWTICommander: WTICommander,
+    clock: Clock) extends Logging {
+
+  def hideUserRecommendation(organizationId: Id[Organization], memberId: Id[User], irrelevantMemberId: Id[User]): Unit = {
+    db.readWrite { implicit session =>
+      orgMembershipRecommendationRepo.recordIrrelevantRecommendation(organizationId, memberId, irrelevantMemberId)
+    }
+  }
+
+  def hideNonUserRecommendation(organizationId: Id[Organization], memberId: Id[User], emailAddress: EmailAddress): Unit = {
```
leogrim: Can we say `Email` here to be specific?

kifi-backend/modules/abook/app/com/keepit/abook/commanders/AbookOrganizationRecommendationCommander.scala
```diff
@@ -0,0 +1,202 @@
...
+      userIdAndScoreStream => userIdAndScoreStream.slice(offset, offset + limit).map(_._1).toSeq
+    })
+    fRecommendations.onSuccess {
+      case Some(recommendations) => log.info(s"Computed ${recommendations.length}/${limit} (skipped $offset) member recommendations for organization $orgId in ${clock.now().getMillis - start.getMillis}ms.")
+      case None => log.info(s"Friend recommendations are not available. Returning in ${clock.now().getMillis - start.getMillis}ms.")
+    }
+    fRecommendations
+  }
+
+  def getNonUserRecommendations(orgId: Id[Organization], memberId: Id[User], offset: Int, limit: Int): Future[Seq[OrganizationInviteRecommendation]] = {
```
leogrim: Same here, this is just emails.

kifi-backend/modules/abook/app/com/keepit/abook/commanders/AbookOrganizationRecommendationCommander.scala
```diff
@@ -0,0 +1,202 @@
...
+        fakeUsers <- fFakeUsers
+      } yield {
+        val irrelevantRecommendations = members ++ invitees.flatMap(_.userId) ++ fakeUsers ++ rejectedRecommendations
+        Some(relatedUsers.related.toStream.filter { case (userId, _) => !irrelevantRecommendations.contains(userId) })
+      }
+    }
+  }
+
+  private def generateFutureNonUserRecommendations(orgId: Id[Organization], memberId: Id[User]): Future[Stream[OrganizationInviteRecommendation]] = {
+    val fExistingInvites = shoebox.getOrganizationInviteViews(orgId)
```
leogrim: You're fetching this twice, check out how `AbookUserRecommendationCommander` is sharing these futures
There, the shared stuff is between Email / LinkedIn / Facebook because they're all invites, as opposed to connection requests for Users. But here, you're actually generating invites for Emails and Users. So the right pattern to look at is Email vs LinkedIn vs Facebook, not these vs User. Makes sense?

kifi-backend/modules/abook/app/com/keepit/abook/commanders/AbookOrganizationRecommendationCommander.scala
```diff
@@ -0,0 +1,202 @@
...
+      memberIds <- fMemberIds
+      basicUsers <- shoebox.getBasicUsers(memberIds.toSeq)
+    } yield {
+      val fullNames = basicUsers.values.flatMap(user => Set(user.firstName + " " + user.lastName, user.lastName + " " + user.firstName))
+      fullNames.toSet.map(normalize)
+    }
+  }
+
+  private val diacriticalMarksRegex = "\\p{InCombiningDiacriticalMarks}+".r
+  @inline private def normalize(fullName: String): String = diacriticalMarksRegex.replaceAllIn(Normalizer.normalize(fullName.trim, Normalizer.Form.NFD), "").toLowerCase
```
leogrim: You should pull most of that code into a helper so that you can use it in both commanders.

kifi-backend/modules/common/app/com/keepit/abook/model/RichSocialConnection.scala
camhashemi, line 60: I added `OrganizationInviteRecommendation` with different fields. Specifically, it doesn't allow the option for social profiles, just email, as well as no `lastInvitedAt` field. The thinking was that `OrgInviteRecommendation` could develop over time, so best to keep them separate.

kifi-backend/modules/abook/app/com/keepit/abook/commanders/AbookOrganizationRecommendationCommander.scala
camhashemi: np

kifi-backend/modules/abook/app/com/keepit/abook/commanders/AbookOrganizationRecommendationCommander.scala
camhashemi: :+1:

kifi-backend/modules/abook/app/com/keepit/abook/commanders/AbookOrganizationRecommendationCommander.scala
camhashemi: Are you saying we should share invites between Emails and Users? Because those are being fetched by separate routes, whereas in `AbookUserRecommendationCommander` Facebook and Email recommendations are fetched by the same route. 

Do you want combine `getUserRecommendations` and `getEmailRecommendations` to be fetched by one route the case of orgs?

camhashemi:
@leogrim updated, minus my confusion about sharing futures (see my reply).

camhashemi:
@leogrim consolidated `getUserRecommendations` and `getEmailRecommendations` into `getRecommendations`, and external service results are now shared.

leogrim:
:+1:

PR #16123, opened 2015-07-09
Reverting to SVG Symbols and loading SVG via AJAX

adamjgrant:


kifi-backend/modules/shoebox/angular/src/app.js
```diff
@@ -1,5 +1,7 @@
 'use strict';
 
+$.get('/dist/symbol-sprite.svg', function(data) { document.body.appendChild(data.documentElement) });
```
adamjgrant: Not crazy about the placement of this. It works fine but just feels icky to put jQuery in Angular. Feedback welcome. @andrewconner @cvializ

cvializ:
:+1: no complaints here

PR #16124, opened 2015-07-09
Cv user checklist

cvializ:


cvializ:
@ac @ajkochanowicz

kifi-backend/modules/shoebox/angular/src/recos/recos.styl
```diff
@@ -42,6 +42,9 @@ twoColMinWidth = appSidePad1 + leftColWidth + colGapWidth + rightColWidth + appS
   margin: 0 auto 50px
   max-width: rightColWidth + 20px
 
+  > * + *
```
adamjgrant, line 45: That's quite the selector.

adamjgrant:
:+1:

kifi-backend/modules/shoebox/angular/src/recos/recos.styl
cvializ, line 45: I was hoping you'd notice that. I've found it useful in the past and I wanted to suggest it:
http://alistapart.com/article/axiomatic-css-and-lobotomized-owls

cvializ, line 45: I can totally get rid of it though, but I wanted to discuss it first : P

kifi-backend/modules/shoebox/angular/src/keep/keepCard.styl
```diff
@@ -253,24 +253,25 @@ defaultBorderColorOverlayUsingBefore(borderWidth)  // slightly darkens an image'
...
-    &:first-child
-      margin-left: -(sidePad)
-
-    &:focus
-    &:hover
-      box-shadow: inset 0 0 0 1px #f6f6f6
-
-      &:active
-        background-color: #fcfcfc
+.kf-keep-keep-btn
```
cvializ, line 256: If you're wondering about these changes (and the changes to recoLibraryCard), it's to make the Follow Library and Keep button and look the same regardless of whether they're inside or outside of a keep/library card

PR #16125, opened 2015-07-09
Rewriting mixpanel events

adamjgrant:


adamjgrant:
Just a few quick modifications to mixpanel events @andrewconner @cvializ

cvializ:
simple enough :+1:

PR #16126, opened 2015-07-09
better representation of user info in ip admin pages

eishay:


eishay:
@ryanpbrewster @gratimax

PR #16129, opened 2015-07-09
IntelliJ really wants to align the shoebox routes

ryanpbrewster:
I fixed the issue. `git diff -w` is very useful

@andrewconner @camhashemi @gratimax 

I actually like it better when all the routes are aligned. I think @gratimax is on that side as well. Do we _want_ to unalign everything?

### DIFF
--- a/kifi-backend/modules/shoebox/conf/shoeboxService.routes
+++ b/kifi-backend/modules/shoebox/conf/shoeboxService.routes
@@ -1,982 +1,982 @@
 
 
-GET           /login                                                                 @com.keepit.controllers.core.AuthController.loginPage
-GET           /ss/login                                                              com.keepit.social.providers.LoginPage.login
-GET           /logout                                                                com.keepit.social.providers.LoginPage.logout
+GET     /login                      @com.keepit.controllers.core.AuthController.loginPage
+GET     /ss/login                   com.keepit.social.providers.LoginPage.login
+GET     /logout                     com.keepit.social.providers.LoginPage.logout
 
-GET           /-ss/login                                                             securesocial.controllers.LoginPage.login
-GET           /-logout                                                               securesocial.controllers.LoginPage.logout
+GET     /-ss/login                   securesocial.controllers.LoginPage.login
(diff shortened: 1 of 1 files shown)


camhashemi:
I also like it a lot better when all routes are aligned. The issue is that when you remove and add all those lines, we don't know who originally added them. @joshm1 gave me this tip, not sure what we value here.

ryanpbrewster:
@andrewconner brought up that every time we add a new, longer route, we're going to have a 900+ line diff. IMO this isn't too big a deal, since the conflict is almost always really simple to resolve.

camhashemi:
oops

andrewconner:
> IMO this isn't too big a deal, since the conflict is almost always really simple to resolve

I think that's a big deal, since it makes it very hard to see what the change was and the merge resolution won't be straight forward (picking all upstream, and then the ones I actually changes, means my lines aren't aligned with the rest of the file).

We could have sections aligned, so diffs are smaller. Aligning the whole file has the unfortunate side effect of making the controller references start on column **86**. @ryanpbrewster

ryanpbrewster:
Starting on column 86 is a bummer. I'm going to merge this now, we can have the discussion about stylistic conventions as a separate matter.

joshm1:
pretty source code is nice, and it used to bother me that columns weren't aligned, but it comes at a cost I don't think makes it worth it. change sets are not going to easily differentiate if changes are functional or cosmetic. and even if you indent sections instead of the entire file, when someone wants to create a route that will be the longest route in that section, they'll have to reindent the entire section to get columns realligned.  I've seen the same issue in code when people like to align variables or key/values pairs for map/dict/hash objects.

These types of commits makes it difficult to use git to see when routes were added and who added them.

```
keepit git:(upmaster) ✗ git blame kifi-backend/modules/shoebox/conf/shoeboxService.routes | grep 2015-07-09 | wc -l
     878
```

just my 2 cents :)

ryanpbrewster:
I know that you can use `git diff -w` to get the diff util to ignore whitespace. I wish there were some way to enable git to ignore whitespace when creating changelogs, or at least flag them differently.

PR #16131, opened 2015-07-09
increasing the grat data fetching timeout

camhashemi:
@andrewconner I tested the batching with size 1000 and it took ~12 seconds, the default timeout seems to be 10 sec.


camhashemi:
@andrewconner going to run the cron and send the emails to myself (hopefully) one last time. I'm fairly confident I nabbed the two bugs. Hopefully that's all that's left.

andrewconner:
Looks good

PR #16135, opened 2015-07-09
Added a controller to direct handle requests

ryanpbrewster:
It takes in a handle, determines if it belongs to an org or a user,
then generates the appropriate response and sends it back. It also
includes information about which type it was, as "user" or "org".

@andrewconner how bad of an idea is this?

@ajkochanowicz

### DIFF
--- a/kifi-backend/modules/shoebox/conf/shoeboxService.routes
+++ b/kifi-backend/modules/shoebox/conf/shoeboxService.routes
@@ -525,6 +525,8 @@ POST          /site/organizations/:id/members/modify            @com.keepit.cont
 POST          /site/organizations/:id/members/remove            @com.keepit.controllers.website.OrganizationMembershipController.removeMembers(id: PublicId[Organization])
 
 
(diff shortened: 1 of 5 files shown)


adamjgrant:
The approach sounds great. I'll defer to @andrewconner about the implementation.

kifi-backend/modules/shoebox/conf/shoeboxService.routes
```diff
@@ -525,6 +525,8 @@ POST          /site/organizations/:id/members/modify
 POST          /site/organizations/:id/members/remove                                 @com.keepit.controllers.website.OrganizationMembershipController.removeMembers(id: PublicId[Organization])
 
 
+GET           /site/user-or-org/:handle                                              @com.keepit.controllers.website.UserOrOrganizationController.getByHandle(handle: Handle)
```
adamjgrant, line 528: Glorious

adamjgrant:
Good to merge, @andrewconner ?

PR #16138, opened 2015-07-09
Fix keep card footer in Firefox.

cvializ:
It was not expanding the fit the children's margin. Table-cell bug?

Error:
screen shot 2015-07-09 at 3 34 08 pm

Fixed:
screen shot 2015-07-09 at 3 41 40 pm


cvializ:
@andrewconner @ajkochanowicz

andrewconner:
Thanks

PR #16140, opened 2015-07-10
more orgs/admin discovery

eishay:


eishay:
@ryanpbrewster

kifi-backend/modules/shoebox/app/views/admin/user.scala.html
```diff
@@ -49,10 +49,10 @@
         <th>Total Bookmarks</th><td>@totalBookmarks</td>
       </tr>
       <tr>
-        <th>Organizations</th>
+        <th>Orgs mem/cand</th>
           <td><ul>
-          @for(org <- organizations) { <li><a href="@com.keepit.controllers.admin.routes.AdminOrganizationController.organizationViewById(org)">@org</a></li> }
-          @for(org <- candidateOrganizations) { <li><a href="@com.keepit.controllers.admin.routes.AdminOrganizationController.organizationViewById(org)">@org</a></li> }
+              <li>@adminHelper.orgsDisplay(organizations)</li>
+              <li>@adminHelper.orgsDisplay(candidateOrganizations)</li>
```
ryanpbrewster, line 55: I think it's reasonable to make this into two rows, one for actual orgs and one for candidate orgs.

PR #16142, opened 2015-07-10
More org admin stuff

ryanpbrewster:
@eishay


kifi-backend/.sbtopts
```diff
@@ -6,3 +6,5 @@
 -J-XX:+HeapDumpOnOutOfMemoryError
 -J-XX:+DoEscapeAnalysis
 -J-XX:+UseCodeCacheFlushing
+-J-XX:MaxMetaspaceSize=1g
```
ryanpbrewster, line 9: This issue is actually driving me crazy, so I'm adding it to the repo.

PR #16143, opened 2015-07-10
Jenkins is not a fan of metaspace memory

ryanpbrewster:


andrewconner:
Fixable, we need to move our prod machines to Java 8, and then jenkins to java 8.

PR #16150, opened 2015-07-10
More site endpoints

ryanpbrewster:
`/site/user-or-org/libraries` exists now, @ajkochanowicz, or will as soon as this is deployed.

@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/test/com/keepit/model/UserFactoryHelper.scala
+++ b/kifi-backend/modules/shoebox/test/com/keepit/model/UserFactoryHelper.scala
@@ -10,8 +10,8 @@ object UserFactoryHelper {
 
   implicit class UserPersister(partialUser: PartialUser) {
     def saved(implicit injector: Injector, session: RWSession): User = {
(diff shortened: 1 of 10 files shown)


kifi-backend/modules/shoebox/test/com/keepit/model/UserFactoryHelper.scala
```diff
@@ -9,7 +10,9 @@ object UserFactoryHelper {
 
   implicit class UserPersister(partialUser: PartialUser) {
     def saved(implicit injector: Injector, session: RWSession): User = {
-      val user = injector.getInstance(classOf[UserRepo]).save(partialUser.get.copy(id = None))
+      val userTemplate = injector.getInstance(classOf[UserRepo]).save(partialUser.get.copy(id = None))
+      //val user = injector.getInstance(classOf[HandleCommander]).setUsername(userTemplate, userTemplate.username, overrideProtection = true, overrideValidityCheck = true).get
```
ryanpbrewster, line 14: A bunch of old tests _really_ dislike this line. We used a lot of things like `.withUsername("test")` all over the place, and `test` is not a legal username. Likewise we have tests where 20 members all have the same username.

PR #16151, opened 2015-07-10
WIP keep card simplification, bug fixing

andrewconner:
@ajkochanowicz Not quite ready to merge, but this is the idea. Thoughts?

Main parts:

• Basically everything here is part of the new "keep card". The only thing not included is the box around it, so that it can be contextually styled (and menus added, in the case of recommendations and libraries). 
• Previously, logic about what to actually show was determined by intermixed CSS rules, ng-ifs, etc. It created a pretty fragile structure, where changes could break components. For instance, I broke recommended keep cards (from the homepage) a couple times without realizing it. It also introduced a few bugs. Now, the two views are much more separated. If you need to style one or the other, OR add new elements to one and not the other, it's very simple. Downside is a _little_ bit of duplicated code (about 20 lines total).


kifi-backend/modules/shoebox/angular/src/keep/keepCard.styl
```diff
@@ -276,9 +276,6 @@ defaultBorderColorOverlayUsingBefore(borderWidth)  // slightly darkens an image'
   &.state-kept-private
     color: libraryRed
 
-    @media (min-width: 480px)
```
adamjgrant, line 280: Cool!

kifi-backend/modules/shoebox/angular/src/keep/keepCard.tpl.html
```diff
@@ -1,68 +1,135 @@
...
-  <a class="kf-keep-image" ng-if="::image" ng-class="::{'kf-bottom-clip': image.clipBottom}"
-    ng-style="{'background-image': 'url(\'' + image.url + '\')', width: image.w ? image.w + 'px' : '', 'padding-top': image.h ? image.h + 'px' : ''}"
-    ng-href="{{::keep.url}}" target="_blank" rel="nofollow" tabindex="-1"
-    ng-click="clickCallback({keep: keep, event: $event})" click-action="clickedArticleImage">
-  </a>
-  <div class="kf-keep-content">
-    <div class="kf-keep-meta">
-      <span class="kf-keep-source">{{::keepSource}}</span>
-      <span class="kf-keep-length" ng-if="::keep.summary.wordCount > 0 && !youtubeId">{{::keep.summary.wordCount|timeToRead}} read</span>
+<div class="kf-keep-card">
```
adamjgrant, line 1: `kf-keep-card` as the top-level node makes it way less confusing. Nice.

adamjgrant:
I don't know if this is directly related your approach, but I'm seeing a lot of lag between switching from gallery to compact view. I noticed a similar effect when I dabbled into the dual template approach as well.

adamjgrant:
In theory, it shouldn't matter. It doesn't seem too expensive to toggle `display` on 12 or so `div`s. But it might have something to do with Angular's DOM traversal. When a class is toggled on the `div`, CSS takes care of the mutation which may be why the incumbent approach is faster when toggling.

adamjgrant:
That being said, from a programming/maintenance point-of-view, I do like the approach.

andrewconner:
I think I know why the lag is happening, continuing on this.

andrewconner:
It's fixable.

adamjgrant:
Cool.

andrewconner:
Ug, this is rotting. Will get it in first thing back in the office.

andrewconner:
Merging this, includes some restyling per suggestions from @JRuff42 to make the cards even smaller. It's not complete, largest bug is editing notes — functionally works, UX is awful. But this fixes most of the bugs I found previously.

PR #16152, opened 2015-07-10
reformat admin domain pages

camhashemi:
@leogrim did you want pagination on all domains as well? I vaguely remember talking about it, but couldn't find a record and not sure.


leogrim:
Not sure pagination is very useful, so many of them. Lookups should be enough. Perhaps lookups by regex would be good though.

PR #16159, opened 2015-07-13
Adding older syntax to support IE

adamjgrant:
@andrewconner @cvializ


cvializ:
I looked into this and got a strange error in IE11. It said "Assignment to read-only properties is not allowed in strict mode." Did you try this in IE?

cvializ:
(I had to use `element[0].setAttribute('class',...)` to make it work)

adamjgrant:
_smacks own hand_

adamjgrant:
Getting a proper VM set up. I was being adventurous. Thanks, @cvializ

cvializ:
I merged a fix for this earlier this week from another branch

PR #16161, opened 2015-07-13
Add "Goodies" admin section

maxov:
As suggested by @leogrim. Also moves the regex tester page to the goodies section.


leogrim:
Awesome! We could have @andrewconner internal / public id lookup in here as well
(see https://admin.kifi.com/admin/ids/...)

leogrim:
This can be merged?

andrewconner:
Nice. Yeah feel free to add the `ids` thing.

PR #16164, opened 2015-07-13
Organization Libraries logic

ryanpbrewster:
Added logic for retrieving an organization's libraries.
    Handles both org and library permissions
    Might be kind of slow.

Fixed the way that LibraryCommander was deciding whether a library was visible
Previously it just asked if the user belonged to the organization
This was a problem because private libraries were never private
It now checks to see if the library is ORGANIZATION visible

@andrewconner @leogrim @ajkochanowicz

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
@@ -19,7 +19,7 @@ trait OrganizationCommander {
   def getAllOrganizationIds: Seq[Id[Organization]]
   def getOrganizationView(orgId: Id[Organization]): OrganizationView
   def getOrganizationCards(orgIds: Seq[Id[Organization]]): Map[Id[Organization], OrganizationCard]
-  def getLibraries(orgId: Id[Organization]): Seq[LibraryCardInfo]
+  def getLibrariesVisibleToUser(orgId: Id[Organization], userIdOpt: Option[Id[User]], offset: Offset, limit: Limit): Seq[LibraryInfo]
   def isValidRequest(request: OrganizationRequest)(implicit session: RSession): Boolean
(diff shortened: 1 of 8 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
```diff
@@ -106,7 +107,11 @@ class OrganizationCommanderImpl @Inject() (
       numLibraries = numPublicLibs)
   }
 
-  def getLibraries(orgId: Id[Organization]): Seq[LibraryCardInfo] = ???
+  def getLibrariesVisibleToUser(orgId: Id[Organization], userIdOpt: Option[Id[User]], offset: Offset, limit: Limit): Seq[LibraryInfo] = {
+    val allLibraries = db.readOnlyReplica { implicit session => libraryRepo.getBySpace(LibrarySpace.fromOrganizationId(orgId)) }
+    val visibleLibraries = allLibraries.filter(lib => libraryCommander.canViewLibrary(userIdOpt, lib))
```
ryanpbrewster, line 112: This seems like it will be slow, or open a ton of db sessions. I'm considering adding a `Helper` method in Library Commander so I can reduce the number of sessions. Any better ideas?

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
```diff
@@ -106,7 +107,11 @@ class OrganizationCommanderImpl @Inject() (
...
+    libraryCommander.getLibrarySummaries(visibleLibraries.drop(offset.value.toInt).take(limit.value.toInt).map(_.id.get).toSeq)
```
ryanpbrewster, line 113: Is there a better method around that maps `Set[Id[Library]] -> Set[LibraryCardInfo]` or something like that? All the methods I saw in Library Commander required a bunch of additional arguments.

adamjgrant:
Nice

PR #16173, opened 2015-07-13
Library transfers to/from users/orgs

ryanpbrewster:
This is bad. We should not pass public ids around inside commanders

@leogrim


kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -612,10 +613,30 @@ class LibraryCommanderImpl @Inject() (
       Left(LibraryFail(FORBIDDEN, "permission_denied"))
     } else {
       val targetMembership = targetMembershipOpt.get
-      val currentSpace = LibrarySpace(userId, organizationId = targetLib.organizationId)
-      val newSpace = LibrarySpace(userId, organizationId = modifyReq.orgId.flatMap(_.destination).orElse(targetLib.organizationId))
+      val currentSpace = targetLib.space
+      val newSpaceOpt = modifyReq.orgMove.map { orgMove =>
+        orgMove.orgId match {
+          case None =>
+            println("trying to move lib to owner space")
```
ryanpbrewster: Ooops

PR #16174, opened 2015-07-13
Switching from using BufferedImage to File for image processing

andrewconner:
Rover has experienced memory pressure problems (including crashing for OOM), partially from the use of `BufferedImage`s during image processing. This is Java's native image in-memory format. It is Bad News™ because it means we load images (including large ones) in memory, persist to disk to perform processing actions, serialize back to memory, etc. Image quality and processing speed also suffers.

This should work, tests pass. But I discovered that not all of our image processing has test coverage, so hopefully the compiler is checking any issues. @leogrim @eishay 

In tests, image processing is about 10x faster, and a couple memory leaks have been fixed. Please review to make sure this makes sense. I'm sure the structure of the code can be cleaned up a bit now.

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/commanders/ProcessedImageHelper.scala
+++ b/kifi-backend/modules/common/app/com/keepit/commanders/ProcessedImageHelper.scala
@@ -8,7 +8,7 @@ import java.security.MessageDigest
 import javax.imageio.ImageIO
 
 import com.keepit.common.core.File
-import com.keepit.common.images.Photoshop
+import com.keepit.common.images.{ RawImageInfo, Photoshop }
 import com.keepit.common.net.{ URI, WebService }
 import com.keepit.common.service.RequestConsolidator
 import com.keepit.common.store.{ ImagePath, ImageSize }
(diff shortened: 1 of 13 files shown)


kifi-backend/modules/common/app/com/keepit/commanders/ProcessedImageHelper.scala
```diff
@@ -146,47 +107,44 @@ trait ProcessedImageHelper {
...
-                val key = ImagePath(baseLabel, sourceImage.hash, imageSize, ProcessImageOperation.Original, sourceImage.format)
-                ImageProcessState.ReadyToPersist(key, outFormat, is, image, bytes, ProcessImageOperation.Original)
-            }
-            original match {
-              case Success(o) => Right(resizedSet + o)
-              case Failure(ex) => Left(ImageProcessState.InvalidImage(ex))
+            val original = {
+              val key = ImagePath(baseLabel, sourceImage.hash, imageSize, ProcessImageOperation.Original, sourceImage.format)
+              ImageProcessState.ReadyToPersist(key, outFormat, image, imageInfo, ProcessImageOperation.Original)
             }
```
leogrim, line 123: :+1:

kifi-backend/modules/common/app/com/keepit/common/images/Image4javaWrapper.scala
```diff
@@ -84,34 +88,30 @@ class Image4javaWrapper @Inject() (
...
-    inputFile.deleteOnExit()
-    ImageIO.write(image, "gif", inputFile)
+  def gifToPng(inputFile: File): File = {
 
     val outputFile = TemporaryFile(prefix = "ImageMagicGifToPngImageOut", suffix = ".png").file
     outputFile.deleteOnExit()
 
     val operation = new IMOperation
-    operation.addImage(inputFile.getAbsolutePath)
+    operation.addImage(inputFile.getAbsolutePath + "[0]")
```
andrewconner, line 97: Fun fact, this means "grab the first frame"

kifi-backend/modules/rover/app/com/keepit/rover/image/RoverImageFetcher.scala
```diff
@@ -38,74 +40,65 @@ class RoverImageFetcher @Inject() (
...
 
-                val requiredProcessRequests = {
-                  val existingImageProcessRequests = existingImageInfos.collect {
-                    case scaledImage if scaledImage.kind == ProcessImageOperation.Scale => ScaleImageRequest(scaledImage.imageSize)
-                    case croppedImage if croppedImage.kind == ProcessImageOperation.Crop => CropImageRequest(croppedImage.imageSize)
+                val uploads = (processedImagesReadyToPersist ++ sourceImageReadyToPersist).map { image =>
+                  val is: InputStream = new FileInputStream(image.file)
+                  val put = imageStore.put(image.key, is, image.file.length.toInt, imageFormatToMimeType(image.format)).imap { _ =>
+                    ImageProcessState.UploadedImage(image.key, image.format, image.file, image.imageInfo, image.processOperation)
                   }
```
leogrim, line 71: Check out `AsyncImageHelper.put`

PR #16180, opened 2015-07-13
adding /m/2/user/:id/proile for getting user profile by id instead of…

jaredpetker:
… username


jaredpetker:
@andrewconner

andrewconner:
Why duplicate so much code? /troll

PR #16183, opened 2015-07-14
Better debugging messages when we resize, so if they file goes missing, we know!

andrewconner:
Got an exception in prod about a missing image. Weird, because it doesn't always happen. Adding extra logging, which will helpfully let us know more about the situations it happens in. Files should be deleted by the JVM when it's unloaded, so I'm a bit confused.

`file.length().toInt` most definitely forces the file to exist, so if the line logs, we know it _did_ exist. @leogrim

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/PageCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/PageCommander.scala
@@ -161,7 +161,7 @@ class PageCommander @Inject() (
     val otherLibraryIds = libraries.filterNot(_._2 == userId).map(_._1)
     val memberLibraryIds = libraryMembershipRepo.getWithLibraryIdsAndUserId(otherLibraryIds.toSet, userId).filter(lm => lm._2.isDefined).keys
     val libraryIds = otherLibraryIds.diff(memberLibraryIds.toSeq)
-    val libraryMap = libraryRepo.getLibraries(libraryIds.toSet).filter(_._2.state != LibraryStates.ACTIVE)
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/PageCommander.scala
```diff
@@ -161,7 +161,7 @@ class PageCommander @Inject() (
     val otherLibraryIds = libraries.filterNot(_._2 == userId).map(_._1)
     val memberLibraryIds = libraryMembershipRepo.getWithLibraryIdsAndUserId(otherLibraryIds.toSet, userId).filter(lm => lm._2.isDefined).keys
     val libraryIds = otherLibraryIds.diff(memberLibraryIds.toSeq)
-    val libraryMap = libraryRepo.getLibraries(libraryIds.toSet)
+    val libraryMap = libraryRepo.getLibraries(libraryIds.toSet).filter(_._2.state != LibraryStates.ACTIVE)
```
andrewconner: Any clue why we'd need this, @leogrim? Saw errors where we had been returning inactive libs.

leogrim: Could be two things:
- we don't clean up properly after deleting a library (e.g. membership)
- handling a race condition where the search index thinks a library still exists just after it got deleted (more likely)

andrewconner: Okay, is this fix good?

andrewconner:
Reviewed in person by @leogrim

PR #16190, opened 2015-07-14
db sessions are all about the power of now, they are useless in the f…

eishay:
…uture

For more reading see http://www.amazon.com/The-Power-Now-Spiritual-Enlightenment/dp/1577314808
Caused these exceptions: https://fortytwo.airbrake.io/projects/91268/groups/1330817179855636731/notices/1464558116196074689


eishay:
@ryanpbrewster @gratimax

PR #16193, opened 2015-07-14
fix routing bugs

camhashemi:
@leogrim I think that there was actually a problem with bePatient, it was in some places and not others


leogrim:
Have you removed it from routes in ABook / Shoebox as well?

camhashemi:
bePatient cannot be found in kifi-backend

leogrim:
:+1:

PR #16194, opened 2015-07-14
heimdal cleanup

yingjieMiao:
@andrewconner @leogrim 

admin search experiments are also removed.


yingjieMiao:
Heimdal client was calling three non-existing endpoints.

PR #16200, opened 2015-07-14
Fixed aliasing bug in library space transfers

ryanpbrewster:
Now `LibraryCommander.modifyLibrary` correctly aliases/reclaims aliases for
libraries when it is used to move them between spaces (user or organization).

@leogrim


leogrim:
:+1:

PR #16202, opened 2015-07-14
add new method to get all org recos for admin

camhashemi:
@leogrim

### DIFF
--- a/kifi-backend/modules/abook/app/com/keepit/abook/commanders/AbookOrganizationRecommendationCommander.scala
+++ b/kifi-backend/modules/abook/app/com/keepit/abook/commanders/AbookOrganizationRecommendationCommander.scala
@@ -13,7 +13,7 @@ import com.keepit.common.service.RequestConsolidator
 import com.keepit.common.time._
 import com.keepit.graph.GraphServiceClient
 import com.keepit.graph.model.SociallyRelatedEntitiesForOrg
-import com.keepit.model.{ OrganizationInviteView, User, Organization }
+import com.keepit.model.{ ExperimentType, UserExperiment, OrganizationInviteView, User, Organization }
 import com.keepit.shoebox.ShoeboxServiceClient
 import scala.concurrent.duration._
 import play.api.libs.concurrent.Execution.Implicits.defaultContext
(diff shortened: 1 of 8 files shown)


kifi-backend/modules/abook/app/com/keepit/abook/commanders/AbookOrganizationRecommendationCommander.scala
```diff
@@ -52,24 +52,45 @@ class AbookOrganizationRecommendationCommander @Inject() (
...
     fRecommendations
   }
 
-  def generateFutureRecommendations(orgId: Id[Organization], memberIds: Set[Id[User]]): Future[Stream[OrganizationInviteRecommendation]] = {
+  def getAllRecommendationsForAdmin(orgId: Id[Organization], viewerId: Id[User], offset: Int, limit: Int): Future[Seq[OrganizationInviteRecommendation]] = {
+    val start = clock.now()
+    val fRecommendations = shoebox.getUserExperiments(viewerId).flatMap { experiments =>
+      experiments match {
+        case experiments if experiments.contains(ExperimentType.ADMIN) => generateFutureRecommendations(orgId, viewerId, true).map(_.slice(offset, offset + limit).toSeq)
+        case _ => Future.successful(Seq.empty)
```
leogrim, line 72: You should probably fail with `Not_Authorized` here

kifi-backend/modules/abook/app/com/keepit/abook/commanders/AbookOrganizationRecommendationCommander.scala
```diff
@@ -52,24 +52,45 @@ class AbookOrganizationRecommendationCommander @Inject() (
...
+      }
+    }
+    fRecommendations.onSuccess {
+      case recommendations if recommendations.nonEmpty => log.info(s"Computed ${recommendations.length}/${limit} (skipped $offset) friend recommendations for org $orgId with viewer $viewerId in ${clock.now().getMillis - start.getMillis}ms.")
+      case _ => log.info(s"Org recommendations are not available. Returning in ${clock.now().getMillis - start.getMillis}ms.")
+    }
+    fRecommendations
+  }
+
+  def generateFutureRecommendations(orgId: Id[Organization], viewerId: Id[User], isAdmin: Boolean = false): Future[Stream[OrganizationInviteRecommendation]] = {
```
leogrim, line 82: Be very explicit here, as in `disclosePrivateEmailAddresses` instead of `isAdmin`. This method should be private, and the only method exposed should check for admin rights, otherwise it's very easy to circumvent.

kifi-backend/modules/abook/app/com/keepit/abook/commanders/AbookOrganizationRecommendationCommander.scala
```diff
@@ -122,14 +142,14 @@ class AbookOrganizationRecommendationCommander @Inject() (
...
+  private def generateFutureEmailRecommendations(orgId: Id[Organization], usersToFilterContacts: Set[Id[User]], fRelatedEntities: Future[Option[SociallyRelatedEntitiesForOrg]], fExistingInvites: Future[Set[OrganizationInviteView]]): Future[Stream[OrganizationInviteRecommendation]] = {
     val fNormalizedUsernames = abookRecommendationHelper.getNormalizedUsernames(Right(orgId))
     fRelatedEntities.flatMap { relatedEntities =>
       val relatedEmailAccounts = relatedEntities.map(_.emailAccounts.related) getOrElse Seq.empty
       if (relatedEmailAccounts.isEmpty) Future.successful(Stream.empty)
       else {
         val rejectedEmailInviteRecommendations = db.readOnlyReplica { implicit session => organizationEmailInviteRecommendationRepo.getIrrelevantRecommendations(orgId) }
-        val allContacts = db.readOnlyMaster { implicit session => memberIds.map(contactRepo.getByUserId).reduce(_ ++ _) }
+        val allContacts = db.readOnlyMaster { implicit session => usersToFilterContacts.map(contactRepo.getByUserId).reduce(_ ++ _) }
```
leogrim, line 152: Just drop the filter if `disclosePrivateEmailAddresses`, otherwise use the viewer's address book.
Also, `reduce` isn't safe is `usersToFilterContacts` happens to be empty.

PR #16206, opened 2015-07-14
Better API

andrewconner:
Using `RoverImageStore` instead of `OrganizationAvatarStore` @ryanpbrewster


ryanpbrewster:
Thank god it was not actually necessary to have a separate image store. That made me really sad.

PR #16209, opened 2015-07-14
Kicking ass on Org Routing with Carlos: The Reunion Tour

adamjgrant:
In progress.


kifi-backend/modules/shoebox/angular/img/symbol-sprites/svg.css
```diff
@@ -0,0 +1,17 @@
...
+  Setting these SVG properties to "inherit" lets us override them in the
+  <use> tag using CSS
+
+  "For user agents that support Styling with CSS, the generated ‘g’ element
+  carries along with it the "cascaded" property values on the ‘use’ element
+  which result from the CSS cascade"
+
+  http://www.w3.org/TR/SVG/struct.html#UseElement
+*/
+rect, path, circle, g { fill: inherit; stroke: inherit; stroke-width: inherit; }
```
adamjgrant, line 11: @cvializ Looks like this works already.

screen shot 2015-07-14 at 6 04 13 pm

kifi-backend/modules/shoebox/angular/src/common/services/routeService.js
```diff
@@ -190,6 +190,20 @@ angular.module('kifi')
       },
 
       ////////////////////////////
+      // Org Profile           //
+      ////////////////////////////
+      getOrgProfile: function (handle) {
```
adamjgrant: @cvializ Did I add this? might need to take it out.

kifi-backend/modules/shoebox/angular/img/symbol-sprites/svg.css
cvializ, line 11: Yeah I noticed that after you merged that in. We can cut this out then : D

kifi-backend/modules/shoebox/angular/src/common/services/routeService.js
cvializ: @ajkochanowicz I didn't add it, so you must have : O

PR #16211, opened 2015-07-14
search admin fix

yingjieMiao:
@leogrim

### DIFF
--- a/kifi-backend/modules/shoebox/app/views/admin/indexGrowth.scala.html
+++ b/kifi-backend/modules/shoebox/app/views/admin/indexGrowth.scala.html
@@ -69,19 +69,19 @@
 
     if (chart == "articleIndex") {
       options.title = "Article Index"
-      options.target = ["substr(stats.gauges.search.index.article_v1_*.size,4,5)"]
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/app/views/admin/indexGrowth.scala.html
```diff
@@ -93,7 +93,7 @@
 
     if (chart == "otherIndex") {
       options.title = "Other Indices"
-      options.target = ["substr(stats.gauges.search.index.user.size,4,5)", "substr(stats.gauges.search.index.message.size,4,5)",
+      options.target = ["substr(stats.gauges.search.index.user_v4.size,4,5)", "substr(stats.gauges.search.index.message.size,4,5)",
```
leogrim, line 96: This is going to break next time we increase the version, can we submit generic stats or get the latest version automatically?

yingjieMiao, line 96: planning to add a function to searchClient so that admin knows about latest version.

PR #16213, opened 2015-07-14
Modified the /:name/:slug endpoint to use handles

ryanpbrewster:
Now it is /:handle/:slug and it searches for a handle owner, not a user

@leogrim @andrewconner


kifi-backend/modules/shoebox/app/com/keepit/controllers/website/KifiSiteRouter.scala
```diff
@@ -103,14 +104,18 @@ class KifiSiteRouter @Inject() (
...
-    lookupUsername(username) flatMap {
-      case (user, userRedirectStatusOpt) =>
-        libraryCommander.getLibraryBySlugOrAlias(user.id.get, LibrarySlug(slug)) map {
+  def serveWebAppIfLibraryFound(handle: Handle, slug: String) = WebAppPage { implicit request =>
+    lookupHandle(handle) flatMap {
+      case (handleOwner, userRedirectStatusOpt) =>
+        val libraryOpt = handleOwner match {
+          case Left(org) => libraryCommander.getLibraryBySlugOrAlias(org.id.get, LibrarySlug(slug))
+          case Right(user) => libraryCommander.getLibraryBySlugOrAlias(user.id.get, LibrarySlug(slug))
+        }
```
leogrim, line 143: Shouldn't `LibraryCommander.getLibraryBySlugOrAlias` take a space as well?

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/KifiSiteRouter.scala
```diff
@@ -119,6 +124,21 @@ class KifiSiteRouter @Inject() (
...
 
+  private def lookupHandle(handle: Handle): Option[(Either[Organization, User], Option[Int])] = {
+    val handleOwnerOpt = db.readOnlyMaster { implicit session => handleCommander.getByHandle(handle) }
+    handleOwnerOpt map {
+      case (handleOwner, isPrimary) =>
+        val foundHandle = handleOwner match {
+          case Left(org) => Handle.fromOrganizationHandle(org.getHandle)
+          case Right(user) => Handle.fromUsername(user.username)
+        }
+        if (foundHandle != handle) { // user moved or username normalization
```
leogrim: `// user moved or username normalization` => `// owner moved or handle normalization`

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/KifiSiteRouter.scala
leogrim, line 143: Ahah, actually you did change it, implicit conversion at work here.

ryanpbrewster, line 143: Gotta love those implicits...

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/KifiSiteRouter.scala
ryanpbrewster: :+1:

andrewconner:
@ryanpbrewster We'd like to roll this live. Can you make it so that orgs _only_ route if the user is in an org? Otherwise, NotFound

ryanpbrewster:
You want that, and not the Org experiment?

andrewconner:
To bring meatspace conversations here, route if user is in the org experiment. NotFound otherwise.

ryanpbrewster:
@andrewconner I think this is ready to go.

PR #16215, opened 2015-07-14
Fix Guice exception

leogrim:
@andrewconner


maxov:
I already fixed this in d3d1e856870c1? Not sure which way is the "right" way

leogrim:
This one, since we have a module to gather all these bindings :)

PR #16216, opened 2015-07-14
org-pymn part deux

camhashemi:
@leogrim includes some refactoring, including using Set instead of Seq in some places. really happy with the new solution, it only adds a few lines to the shared logic.

going to test the route manually before adding to admin

### DIFF
--- a/kifi-backend/modules/abook/app/com/keepit/abook/commanders/AbookOrganizationRecommendationCommander.scala
+++ b/kifi-backend/modules/abook/app/com/keepit/abook/commanders/AbookOrganizationRecommendationCommander.scala
@@ -56,7 +56,7 @@ class AbookOrganizationRecommendationCommander @Inject() (
 
   def getRecommendations(orgId: Id[Organization], viewerId: Id[User], offset: Int, limit: Int): Future[Seq[OrganizationInviteRecommendation]] = {
     val start = clock.now()
-    val fRecommendations = generateFutureRecommendations(orgId, viewerId).map {
+    val fRecommendations = generateFutureRecommendations(orgId, viewerId, false).map {
       recoStream => recoStream.slice(offset, offset + limit).toSeq
     }
     fRecommendations.onSuccess {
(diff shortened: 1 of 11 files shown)


andrewconner:
I liked oymn.

camhashemi:
@andrewconner oymn recommends orgs for a user, org-pymn recommends users for an org. oymn is on the way.

kifi-backend/modules/abook/app/com/keepit/abook/commanders/AbookOrganizationRecommendationCommander.scala
```diff
@@ -144,12 +145,19 @@ class AbookOrganizationRecommendationCommander @Inject() (
...
       else {
         val rejectedEmailInviteRecommendations = db.readOnlyReplica { implicit session => organizationEmailInviteRecommendationRepo.getIrrelevantRecommendations(orgId) }
-        val viewersContacts = if (!disclosePrivateEmails) { db.readOnlyMaster { implicit session => contactRepo.getByUserId(viewerId) } } else { Seq.empty }
+        val allContacts = db.readOnlyMaster { implicit session =>
+          if (!disclosePrivateEmails) {
+            contactRepo.getByUserId(viewerId)
+          } else {
+            val relevantEmailAccountIds = relatedEmailAccounts.map { case (id, score) => EmailAccount.fromEmailAccountInfoId(id) }.toSet
+            contactRepo.getByEmailAccountIds(relevantEmailAccountIds)
+          }
```
leogrim, line 139: :+1:

leogrim:
I think you forgot to use `offset` and `limit` in `getRecommendationsForAdmin`, perhaps you can merge these two methods, decide whether to check experiments based on `disclosePrivateEmails` and just pass a different boolean in the controller?

camhashemi:
It should be there? `AbookOrganizationRecommendationCommander[73]`.

But regardless, that's great. Update on the way.

PR #16217, opened 2015-07-14
make sure to call the correct function for the new profile for user id endpoint

jaredpetker:
@andrewconner Third times a charm, I swear it, had to fix the route to call the right function, almost bad things

### DIFF
--- a/kifi-backend/modules/shoebox/conf/shoeboxService.routes
+++ b/kifi-backend/modules/shoebox/conf/shoeboxService.routes
@@ -209,8 +209,8 @@ POST    /m/1/invite                             @com.keepit.controllers.mobile.M
 
 # Profile Page
 GET     /m/1/user/:username/profile             @com.keepit.controllers.mobile.MobileUserProfileController.profile(username)
-GET     /m/2/user/:id/profile                   @com.keepit.controllers.mobile.MobileUserProfileController.profile(id)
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/conf/shoeboxService.routes
```diff
@@ -241,7 +241,7 @@ GET     /m/2/recos/top                          @com.keepit.controllers.mobile.M
 GET     /m/3/recos/top                          @com.keepit.controllers.mobile.MobileRecommendationsController.topRecosV3(recency: Float, uriContext: Option[String] ?= None, libContext: Option[String] ?= None, uctx: Option[String] ?= None, lctx: Option[String] ?= None)
 POST    /m/3/recos/top                          @com.keepit.controllers.mobile.MobileRecommendationsController.topRecosV3Post()
 GET     /m/1/recos/public                       @com.keepit.controllers.mobile.MobileRecommendationsController.topPublicRecos()
-POST    /m/1/recos/trash                        @com.keepit.controllers.mobile.MobileRecommendationsController.trash(id: ExternalId[NormalizedURI])
+POST    /m/1/recos/trash                        @com.kekeepupdatesepit.controllers.mobile.MobileRecommendationsController.trash(id: ExternalId[NormalizedURI])
```
andrewconner: This is probably not right.

PR #16223, opened 2015-07-15
Refactor potential orgs query

maxov:


maxov:
@ryanpbrewster

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationRepo.scala
```diff
@@ -95,21 +76,30 @@ class OrganizationRepoImpl @Inject() (
...
-            group by organization.id;
-      """.as[Organization].list
+              select user_id, organization_id from organization_membership_candidate
+          ) as membership
+          on user.id = membership.user_id
+      ) as membership
+        inner join organization on organization.id = membership.organization_id;
+      """.as[Id[Organization]].list.toSet
+
+    getByIds(orgIds).values.toSeq
```
ryanpbrewster, line 102: Is there any reason this had to be refactored? Couldn't you just have changed `select organization.*` to `select organization.id`?

maxov, line 102: I incorporated the change @eishay wanted (including candidate memberships in this query) as well

ryanpbrewster, line 102: Ah! Right, I recall now.

PR #16228, opened 2015-07-15
Refactor Search Filters for UriSearch

leogrim:
@andrewconner @yingjieMiao 
I think this does the right thing and will fix tag search inside of libraries. It also enables to search within an organization or a user's content (except for Main).

I haven't looked into User Search and Library Search yet (these shouldn't be problematic until we actually have search within an organization).

### DIFF
--- a/kifi-backend/modules/search/app/com/keepit/search/engine/VisibilityEvaluator.scala
+++ b/kifi-backend/modules/search/app/com/keepit/search/engine/VisibilityEvaluator.scala
@@ -1,14 +1,12 @@
 package com.keepit.search.engine
 
 import com.keepit.common.akka.MonitoredAwait
-import com.keepit.search.engine.query.QueryUtil
+import com.keepit.search.SearchFilter
 import com.keepit.search.index.{ WrappedSubReader }
-import com.keepit.search.index.article.ArticleFields
 import com.keepit.search.index.graph.keep.KeepFields
 import com.keepit.search.index.graph.library.LibraryFields
 import com.keepit.search.util.LongArraySet
-import org.apache.lucene.index.{ Term, NumericDocValues }
-import org.apache.lucene.search.DocIdSetIterator
+import org.apache.lucene.index.{ NumericDocValues }
 import scala.concurrent.duration._
 import scala.concurrent.Future
 
@@ -20,6 +18,7 @@ trait VisibilityEvaluator { self: DebugOption =>
   protected val libraryIdsFuture: Future[(Set[Long], Set[Long], Set[Long], Set[Long])]
   protected val orgIdsFuture: Future[Set[Long]]
   protected val monitoredAwait: MonitoredAwait
+  protected val filter: SearchFilter
 
   lazy val myFriendIds = LongArraySet.fromSet(monitoredAwait.result(friendIdsFuture, 5 seconds, s"getting friend ids"))
 
@@ -51,6 +50,7 @@ trait VisibilityEvaluator { self: DebugOption =>
       trustedLibraryIds,
       authorizedLibraryIds,
       orgIds,
+      filter,
       userIdDocValues,
       libraryIdDocValues,
       orgIdDocValues,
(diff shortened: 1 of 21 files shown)


kifi-backend/modules/search/app/com/keepit/search/engine/VisibilityEvaluator.scala
```diff
@@ -97,34 +98,46 @@ final class KeepVisibilityEvaluator(
...
-        Visibility.OWNER // the keep is in my library (I may or may not have kept it)
-      } else {
-        val keeperId = userIdDocValues.get(docId)
-        if (keeperId == userId) {
+  def apply(docId: Int): Int = {
+    val libId = libraryIdDocValues.get(docId)
+    val keeperId = userIdDocValues.get(docId)
+    val orgId = orgIdDocValues.get(docId)
+
+    if (noFilter || isRelevant(libId, keeperId, orgId)) {
```
yingjieMiao, line 127: you already test `noFilter` inside `isRelevant`.

leogrim, line 127: Not the same `isRelevant`

kifi-backend/modules/search/app/com/keepit/search/engine/VisibilityEvaluator.scala
```diff
@@ -147,34 +162,54 @@ final class LibraryVisibilityEvaluator(
...
+    val orgId = orgIdDocValues.get(docId)
+    if (noFilter || isRelevant(libId, ownerId, orgId)) {
+      if (memberLibraryIds.findIndex(libId) >= 0) {
+        if (myOwnLibraryIds.findIndex(libId) >= 0) {
+          Visibility.OWNER // my library
         } else {
-          Visibility.OTHERS // another published library
+          Visibility.MEMBER // a library I am a member of
         }
+      } else if (orgIdDocValues != null && orgIds.findIndex(orgIdDocValues.get(docId)) >= 0) {
```
yingjieMiao, line 194: we can remove `orgIdDocValues != null` now. and `orgId` is already available.

kifi-backend/modules/search/app/com/keepit/search/engine/VisibilityEvaluator.scala
```diff
@@ -97,34 +98,46 @@ final class KeepVisibilityEvaluator(
...
-    } else if (authorizedLibraryIds.findIndex(libId) >= 0) {
-      Visibility.MEMBER // the keep is in an authorized library
-    } else if (orgIdDocValues != null && orgIds.findIndex(orgIdDocValues.get(docId)) >= 0) { // keep is owned by an org that I am a member of
-      Visibility.MEMBER
-    } else {
-      if (visibilityDocValues.get(docId) == published) {
-        val keeperId = userIdDocValues.get(docId)
+      } else if (authorizedLibraryIds.findIndex(libId) >= 0) {
+        Visibility.MEMBER // the keep is in an authorized library
+      } else if (orgIds.findIndex(orgId) >= 0) { // keep is owned by an org that I am a member of
```
yingjieMiao, line 138: looks like `LongArraySet` has method `contains`. (at least the one returned by LongArraySet.fromSorted)

leogrim, line 138: `contains` would lead to boxing of `Long` into an object

kifi-backend/modules/search/app/com/keepit/search/engine/VisibilityEvaluator.scala
leogrim, line 194: Yep, I missed this one, removed it somewhere else.

kifi-backend/modules/search/app/com/keepit/search/engine/VisibilityEvaluator.scala
yingjieMiao, line 138: Interesting. Code paths look alike:

```
override def findIndex(key: Long): Int = mapper(key)
override def contains(key: Long): Boolean = (mapper(key) >= 0)
```

leogrim, line 138: `findIndex` is defined in `LongArraySet` to take a `Long`
`contains` is defined in `Set[T]` to take a `T`

PR #16232, opened 2015-07-15
Org move cleanup

ryanpbrewster:
@leogrim @ajkochanowicz @cvializ 

Cleaned up library transfers to/from orgs/users. The new format is
`{"space": {"user": ext-id}}` for moving to a user space (right now it only allows moving to the library owner's space, by the library owner)
`{"space": {"org": pub-id}}` for moving to an org.

I'm open to changing `"space"` to `"move"` or `"destination"` for the purpose of deserialization.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/model/LibraryRepo.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/model/LibraryRepo.scala
@@ -192,7 +192,7 @@ class LibraryRepoImpl @Inject() (
   }
 
   private def getByUserIdAndSlug(userId: Id[User], slug: LibrarySlug, excludeStates: Set[State[Library]])(implicit session: RSession): Option[Library] = {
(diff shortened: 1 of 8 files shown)


kifi-backend/modules/shoebox/app/com/keepit/model/LibraryRepo.scala
```diff
@@ -192,7 +192,7 @@ class LibraryRepoImpl @Inject() (
   }
 
   private def getByUserIdAndSlug(userId: Id[User], slug: LibrarySlug, excludeStates: Set[State[Library]])(implicit session: RSession): Option[Library] = {
-    (for (b <- rows if b.slug === slug && b.ownerId === userId && !b.state.inSet(excludeStates)) yield b).firstOption
+    (for (b <- rows if b.slug === slug && b.ownerId === userId && b.orgId.isEmpty && !b.state.inSet(excludeStates)) yield b).firstOption
```
ryanpbrewster, line 195: Hilarious bug. :-1:

leogrim:
Nice! :+1:

PR #16233, opened 2015-07-15
remove OrganizationEmailInviteRecommendation

camhashemi:
@leogrim this stems from my early misunderstanding of how OrganizationInvites worked. Since they are mostly the same for Id[User] or EmailAccount, we should just use one repo. That decision and the implementation is up for your review.


camhashemi:
@leogrim I have to merge to get org-pymn out before tomorrow, but would appreciate a review. thanks for your support today.

PR #16238, opened 2015-07-15
org-admin refactor

camhashemi:


kifi-backend/modules/shoebox/app/views/admin/organization.scala.html
```diff
@@ -89,11 +89,11 @@
     </tr>
     @for(reco <- orgStats.memberRecommendations) {
         <tr>
-            <td>@if(reco.identifier.isLeft) { reco.identifier.left.get } else { reco.identifier.right.get }</td>
+            <td>@if(reco.identifier.isLeft) { @{reco.identifier.left.get} } else { @{reco.identifier.right.get} }</td>
```
eishay, line 92: Good job betting the list to the admin.
The list as it is now is not actionable, the user ids are not helping much as they are.
Please have the user name and email in the table and add a link to the admin page of that user (see admin helpers). Many times its clear that the user belongs to the org so the buttons you added are great.

PR #16243, opened 2015-07-15
Admin link from user's public profile

andrewconner:
@cvializ Requested by Eishay.


cvializ:
:+1: looks good to me

PR #16248, opened 2015-07-15
(admin) grab user objects for org member recommendations

camhashemi:
@ryanpbrewster @leogrim

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/UserStatisticsCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/UserStatisticsCommander.scala
@@ -8,6 +8,7 @@ import com.keepit.common.crypto.{ PublicId, PublicIdConfiguration }
 import com.keepit.common.db.Id
 import com.keepit.common.db.slick.DBSession.RSession
 import com.keepit.common.db.slick.Database
+import com.keepit.common.mail.EmailAddress
 import com.keepit.eliza.ElizaServiceClient
 import com.keepit.model._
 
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/UserStatisticsCommander.scala
```diff
@@ -66,7 +67,11 @@ case class OrganizationStatistics(
   members: Set[OrganizationMembership],
   candidates: Set[OrganizationMembershipCandidate],
   membersStatistics: Map[Id[User], MemberStatistics],
-  memberRecommendations: Seq[OrganizationInviteRecommendation])
+  memberRecommendations: Seq[OrganizationMemberRecommendationInfo])
+
+case class OrganizationMemberRecommendationInfo(
+  userOrEmail: Either[User, EmailAddress],
```
leogrim, line 73: What about using `RichContact` here

camhashemi, line 73: The motivation for `OrgMemberRecommendationInfo` is to populate the admin page with a `User` vs `Id[User]`. So `RichContact` wouldn't do that, but do you mean in place of `EmailAddress` only?

camhashemi, line 73: Or do you mean for `AbookOrgRecoCommander.getRecommendations` to return a `RichContact` instead of `EmailAddress`?

leogrim, line 73: Yes I meant instead of `EmailAddress`, check out regular invites

PR #16250, opened 2015-07-15
Removed "lastInvitedAt" from org members

ryanpbrewster:
Invitees still have it. This helps distinguish members from invitees

@cvializ


cvializ:
Hooray! :dancer:

PR #16254, opened 2015-07-15
(Graph) don't wander over "fake org owner" vertex

camhashemi:
@leogrim this should do it


kifi-backend/modules/graph/app/com/keepit/graph/wander/SocialWanderingCommander.scala
```diff
@@ -204,6 +204,7 @@ object SocialWanderlust {
     10015, // Kifi Editorial
     11784, // Kifi Product
     11785, // Kifi Eng
-    16707 // Kifi Twitter
+    16707, // Kifi Twitter
+    97543 // Fake Org Owner
```
leogrim, line 208: :+1:

andrewconner:
That's easy

PR #16262, opened 2015-07-15
Patches issue with notes

adamjgrant:
Not a fix, but a patch so it looks "not broken"

Before

screen shot 2015-07-15 at 7 16 40 pm

After

screen shot 2015-07-15 at 7 18 08 pm


adamjgrant:
@ac @cvializ

cvializ:
:+1: makes sense

PR #16264, opened 2015-07-15
Org request invite

ryanpbrewster:
@leogrim @eishay @andrewconner 

Allow users to "poke" an organization to request an invite. Suggestions for the name are welcome, but I don't want to use "OrganizationMembershipRequest" because that is already used internally to process organization membership add/modify/delete operations.


ryanpbrewster:
Closing this until we decide to revisit it.

PR #16275, opened 2015-07-16
Add hubot to tools

maxov:
@andrewconner 

[Hubot](https://hubot.github.com/) is a fantastic tool, made the actual script really easy


andrewconner:
Looks good.

maxov:
@andrewconner Any way we can deploy changes to hubot automatically? I don't know what we do with the current deploy scripts

andrewconner:
Right now, unfortunately, copy the files to arthur. You can use `scp`. We should have a better way to keep arthur in sync.

PR #16282, opened 2015-07-16
Made org invitations more safe

ryanpbrewster:
We need to stop using `payload.as[T]` since it throws 500s. Much better to
use `payload.validate[T]` and handle the badly formed input directly.

@cvializ @ajkochanowicz 

Now org invites do not need a `type:` field. I figure it out based on whether the field is named `"id"` or `"email"`. Also, it won't throw a 500 on bad requests anymore, it will tell you it couldn't parse the input.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/OrganizationInviteController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/OrganizationInviteController.scala
@@ -29,37 +29,31 @@ class OrganizationInviteController @Inject() (
     implicit val executionContext: ExecutionContext) extends UserActions with OrganizationAccessActions with ShoeboxServiceController {
 
   def inviteUsers(pubId: PublicId[Organization]) = OrganizationUserAction(pubId, OrganizationPermission.INVITE_MEMBERS).async(parse.tolerantJson) { request =>
-    val invites = (request.body \ "invites").as[JsArray].value
-    val msg = (request.body \ "message").asOpt[String].filter(_.nonEmpty)
+    val messageOpt = (request.body \ "message").asOpt[String].filter(_.nonEmpty)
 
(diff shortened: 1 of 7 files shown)


kifi-backend/modules/shoebox/app/com/keepit/controllers/website/OrganizationInviteController.scala
```diff
@@ -29,37 +29,31 @@ class OrganizationInviteController @Inject() (
...
-      val role = (userInvite \ "role").as[OrganizationRole]
-      OrganizationMemberInvitation(Left(userId), role, msg)
-    }
-
-    val emailInfo = emailInvites.map { emailInvite =>
-      val email = (emailInvite \ "email").as[EmailAddress]
-      val role = (emailInvite \ "role").as[OrganizationRole]
-      OrganizationMemberInvitation(Right(email), role, msg)
-    }
+    (request.body \ "invites").validate[Seq[ExternalOrganizationMemberInvitation]] match {
```
ryanpbrewster, line 34: @leogrim how do you feel about this convention? I have companion internal and external objects so that the controller can handle strictly the conversion back and forth between them. The actual business logic is in the models and commanders.

leogrim, line 34: Like it a lot!

andrewconner, line 34: This is great; another way to do it safely is `json.asOpt[MyType]`

ryanpbrewster, line 34: Yep. We talked earlier about possibly sending helpful error messages to the client upon a bad request, earlier, which would be easier with `validate`. Who knows whether it's a good idea or not.

PR #16285, opened 2015-07-16
Fixed a few permissions-based visibility bugs

ryanpbrewster:
Now users can only see libraries that they have permission to see
Same story with organizations

@leogrim @andrewconner @cvializ @ajkochanowicz 

I'd like some feedback on this PR in particular. I did some wonky things with commanders, namely
exposing a lot of internal "helper" methods so that other commanders could use the logic without having to open a ton of DB sessions. I don't feel super confident in it.

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/model/Organization.scala
+++ b/kifi-backend/modules/common/app/com/keepit/model/Organization.scala
@@ -38,6 +38,9 @@ case class Organization(
   def withBasePermissions(newBasePermissions: BasePermissions): Organization = {
     this.copy(basePermissions = new BasePermissions(basePermissions.permissionsMap ++ newBasePermissions.permissionsMap))
   }
+  def hiddenFromNonmembers: Organization = {
+    this.withBasePermissions(BasePermissions(Map(None -> Set())))
+  }
 
   def getNonmemberPermissions = basePermissions.forNonmember
   def getRolePermissions(role: OrganizationRole) = basePermissions.forRole(role)
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
@@ -43,6 +43,7 @@ class OrganizationCommanderImpl @Inject() (
     keepRepo: KeepRepo,
     libraryRepo: LibraryRepo,
     basicUserRepo: BasicUserRepo,
+    libraryMembershipRepo: LibraryMembershipRepo,
     libraryCommander: LibraryCommander,
     airbrake: AirbrakeNotifier,
     implicit val publicIdConfig: PublicIdConfiguration,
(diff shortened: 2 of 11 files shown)


kifi-backend/modules/common/app/com/keepit/model/Organization.scala
```diff
@@ -38,6 +38,9 @@ case class Organization(
   def withBasePermissions(newBasePermissions: BasePermissions): Organization = {
     this.copy(basePermissions = new BasePermissions(basePermissions.permissionsMap ++ newBasePermissions.permissionsMap))
   }
+  def hiddenFromNonmembers: Organization = {
```
ryanpbrewster, line 41: maybe "make secret"? Don't even know if this is really necessary. Mostly for tests.

ryanpbrewster, line 41: Probably better to put this into `OrganizationFactoryHelper` now that I think about it.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
```diff
@@ -93,8 +104,8 @@ class OrganizationCommanderImpl @Inject() (
     val numMembers = orgMembershipRepo.countByOrgId(orgId)
     val avatarPath = organizationAvatarCommander.getBestImage(orgId, ImageSize(200, 200)).map(_.imagePath)
 
-    val librariesByVisibility = libraryRepo.countLibrariesForOrgByVisibility(orgId)
-    val numPublicLibs = librariesByVisibility(LibraryVisibility.PUBLISHED) // TODO: actually find the number of libraries
+    // TODO(ryan): this is bad, surely there is a better way
+    val numLibraries = getLibrariesVisibleToUserHelper(orgId, viewerIdOpt, offset = Offset(0), limit = Limit(Int.MaxValue)).length
```
ryanpbrewster: I end up just grabbing every single library in an organization and filtering out the ones that are hidden, then taking the length. Is there a safer way? For libraries it was very explicit, as part of the `LibraryMembership`, but now "viewing privileges" are implicit and much harder to predict.

leogrim: Can you run a single query relying on a 'Set[Id[Library]]' (the user's `LibraryMembership`s) to get the private ones, include the `organization` visible if and only if the user has a membership and then the public ones?

ryanpbrewster: That should reduce it down to ~3 queries, which is a lot better than `n`. Good idea.

I imagine it should be possible to do a single query with all of that garbage thrown in.
`for(lib <- rows if
(includeOrgLibraries && lib.orgId === orgId && lib.visibility === LibraryVisibility.ORGANIZATION) ||
 (lib.visibility === LibraryVisibility.PUBLISHED) 
|| lib.id.inSet(userLibraryMemberships)) yield lib`

leogrim: You're pulling for an org, `lib.orgId === orgId` should be top level: 
`for(lib <- rows if lib.orgId === orgId && ((includeOrgLibraries && lib.visibility === LibraryVisibility.ORGANIZATION) || (lib.visibility === LibraryVisibility.PUBLISHED) || lib.id.inSet(userLibraryMemberships))) yield lib`

ryanpbrewster: :+1:

I looked around in `LibraryRepo` and there are about a million special-purpose SQL queries. Time to add to the problem.

PR #16286, opened 2015-07-16
Expand prefix search, enable client to disable full text search

leogrim:
This PR @andrewconner @eishay :
1. Makes search filters (library, user, org) available on the mobile API
2. Makes `disableFullTextSearch` available on the site and mobile APIs (to be used in dropdown typeahead)
3. Expand prefix search to support uris (by keep title).

I breaks distributed search APIs so I'll deploy at a quiet time.


andrewconner:
Nice!

PR #16290, opened 2015-07-17
use LibraryCardInfo instead of LibraryInfo for org libraries

camhashemi:
@andrewconner @cvializ @ryanpbrewster


andrewconner:
Looks good!

PR #16300, opened 2015-07-17
Added "path" to Library FullInfos and CardInfos

ryanpbrewster:
@andrewconner 

I don't even want to use git blame to find out who wrote the LibraryController tests, but I hate them. Jesus those are fragile.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/model/LibraryView.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/model/LibraryView.scala
@@ -135,7 +135,6 @@ object LibraryInfo {
   }
 }
 
-@json
 case class LibraryCardInfo(
   id: PublicId[Library],
   name: String,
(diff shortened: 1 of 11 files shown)


kifi-backend/modules/shoebox/app/com/keepit/model/LibraryView.scala
```diff
@@ -214,7 +236,8 @@ case class FullLibraryInfo(
   numFollowers: Int,
   attr: Option[LibrarySourceAttribution] = None,
   whoCanInvite: LibraryInvitePermissions,
-  modifiedAt: DateTime)
+  modifiedAt: DateTime,
+  path: String)
```
andrewconner, line 240: Is this here just for consistency?

ryanpbrewster, line 240: Partly for consistently, and partly because there's one method somewhere that takes a `FullLibraryInfo` and transforms it into a `LibraryCardInfo`, so I needed it there. It seemed like a good idea.

PR #16308, opened 2015-07-18
Org response

camhashemi:
@ryanpbrewster @leogrim @andrewconner OrganizationInfos and MembershipInfos compose OrganizationViews, which are sent to clients. If different clients prefer different serializations of that data, no problem

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
@@ -19,7 +19,7 @@ import scala.util.{ Failure, Success, Try }
 @ImplementedBy(classOf[OrganizationCommanderImpl])
 trait OrganizationCommander {
   def getAllOrganizationIds: Seq[Id[Organization]]
-  def getOrganizationResponse(orgId: Id[Organization], viewerIdOpt: Option[Id[User]]): OrganizationGetResponse
+  def getOrganizationView(orgId: Id[Organization], viewerIdOpt: Option[Id[User]]): OrganizationView
   def getOrganizationCards(orgIds: Seq[Id[Organization]], viewerIdOpt: Option[Id[User]]): Map[Id[Organization], OrganizationCard]
   def getLibrariesVisibleToUser(orgId: Id[Organization], userIdOpt: Option[Id[User]], offset: Offset, limit: Limit): Seq[LibraryCardInfo]
   def isValidRequest(request: OrganizationRequest)(implicit session: RSession): Boolean
@@ -52,11 +52,11 @@ class OrganizationCommanderImpl @Inject() (
   // TODO(ryan): do the smart thing and add a limit/offset
   def getAllOrganizationIds: Seq[Id[Organization]] = db.readOnlyReplica { implicit session => orgRepo.all().map(_.id.get) }
 
-  def getOrganizationResponse(orgId: Id[Organization], viewerIdOpt: Option[Id[User]]): OrganizationGetResponse = {
+  def getOrganizationView(orgId: Id[Organization], viewerIdOpt: Option[Id[User]]): OrganizationView = {
     db.readOnlyReplica { implicit session =>
-      val organizationView = getOrganizationViewHelper(orgId, viewerIdOpt)
-      val orgViewerRelationship = getOrganizationViewerRelationshipHelper(orgId, viewerIdOpt)
-      OrganizationGetResponse(organizationView, orgViewerRelationship)
+      val organizationInfo = getOrganizationInfoHelper(orgId, viewerIdOpt)
+      val orgViewerRelationship = getMembershipInfoHelper(orgId, viewerIdOpt)
+      OrganizationView(organizationInfo, orgViewerRelationship)
     }
   }
   def getOrganizationCards(orgIds: Seq[Id[Organization]], viewerIdOpt: Option[Id[User]]): Map[Id[Organization], OrganizationCard] = {
@@ -64,7 +64,7 @@ class OrganizationCommanderImpl @Inject() (
       orgIds.map { orgId => orgId -> getOrganizationCardHelper(orgId, viewerIdOpt) }.toMap
     }
   }
-  private def getOrganizationViewHelper(orgId: Id[Organization], viewerIdOpt: Option[Id[User]])(implicit session: RSession): OrganizationView = {
+  private def getOrganizationInfoHelper(orgId: Id[Organization], viewerIdOpt: Option[Id[User]])(implicit session: RSession): OrganizationInfo = {
     if (!orgMembershipCommander.getPermissionsHelper(orgId, viewerIdOpt).contains(OrganizationPermission.VIEW_ORGANIZATION)) {
       airbrake.notify(s"Tried to serve up an organization view for org $orgId to viewer $viewerIdOpt, but they do not have permission to view this org")
     }
@@ -84,7 +84,7 @@ class OrganizationCommanderImpl @Inject() (
 
     val numLibraries = countLibrariesVisibleToUserHelper(orgId, viewerIdOpt)
 
-    OrganizationView(
+    OrganizationInfo(
       orgId = Organization.publicId(orgId),
       ownerId = ownerId,
       handle = orgHandle,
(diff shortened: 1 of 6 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
```diff
@@ -99,6 +103,14 @@ class OrganizationCommanderImpl @Inject() (
       numLibraries = numLibraries)
   }
 
+  private def getMembershipInfoHelper(orgId: Id[Organization], viewerIdOpt: Option[Id[User]])(implicit session: RSession): MembershipInfo = {
+    viewerIdOpt.map { userId =>
+      val membershipOpt = orgMembershipRepo.getByOrgIdAndUserId(orgId, userId)
+      val invites = orgInviteRepo.getByOrgIdAndUserId(orgId, userId)
+      MembershipInfo(isInvited = invites.nonEmpty, role = membershipOpt.map(_.role), permissions = membershipOpt.map(_.permissions).getOrElse(orgRepo.get(orgId).basePermissions.forNonmember))
+    }.getOrElse(MembershipInfo(isInvited = false, role = None, permissions = orgRepo.get(orgId).basePermissions.forNonmember))
+  }
```
ryanpbrewster, line 112: You will always need the org here, so just grab it at the beginning.

ryanpbrewster, line 112: Also, the user may be invited even if they aren't a Kifi user. I'm not sure how we want to handle that, but it doesn't feel right to say they're not.

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileOrganizationController.scala
```diff
@@ -50,9 +50,9 @@ class MobileOrganizationController @Inject() (
         orgCommander.modifyOrganization(OrganizationModifyRequest(request.request.userId, request.orgId, modifications)) match {
           case Left(failure) => failure.asErrorResponse
           case Right(response) =>
-            val orgView = orgCommander.getOrganizationView(response.modifiedOrg.id.get, request.request.userIdOpt)
-            implicit val writes = OrganizationView.mobileV1
-            Ok(Json.obj("organization" -> Json.toJson(orgView)))
+            val orgResponse = orgCommander.getOrganizationView(response.modifiedOrg.id.get, request.request.userIdOpt)
```
ryanpbrewster: Call it `OrganizationView` or `OrganizationResponse`, but the val name should match the method name.

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileOrganizationController.scala
```diff
@@ -66,9 +66,9 @@ class MobileOrganizationController @Inject() (
   }
 
   def getOrganization(pubId: PublicId[Organization]) = OrganizationAction(pubId, OrganizationPermission.VIEW_ORGANIZATION) { request =>
-    val orgView = orgCommander.getOrganizationView(request.orgId, request.request.userIdOpt)
+    val orgResponse = orgCommander.getOrganizationView(request.orgId, request.request.userIdOpt)
```
ryanpbrewster: Same thing, either response or view.

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileOrganizationController.scala
```diff
@@ -66,9 +66,9 @@ class MobileOrganizationController @Inject() (
...
     val requesterPermissions = Json.toJson(orgMembershipCommander.getPermissions(request.orgId, request.request.userIdOpt))
-    Ok(Json.obj("organization" -> Json.toJson(orgView)(OrganizationView.mobileV1), "viewer_permissions" -> requesterPermissions))
+    Ok(Json.obj("organization" -> Json.toJson(orgResponse)(OrganizationView.mobileWrites), "viewer_permissions" -> requesterPermissions))
```
ryanpbrewster: I'm fine having `mobileWrites` always be the most "up to date" version, but we're almost certainly going to end up with `v1`, `v2`, etc.

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationInfo.scala
```diff
@@ -30,10 +30,34 @@ object OrganizationView {
...
+  organizationInfo: OrganizationInfo,
+  membershipInfo: MembershipInfo)
+
+object OrganizationView {
+  val websiteWrites: Writes[OrganizationView] = (
+    (__ \ 'organizationInfo).write(OrganizationInfo.defaultWrites) and
+    (__ \ 'membershipInfo).write(MembershipInfo.defaultWrites)
+  )(unlift(OrganizationView.unapply))
+  val mobileWrites = new Writes[OrganizationView] {
+    def writes(o: OrganizationView) = OrganizationInfo.defaultWrites.writes(o.organizationInfo).as[JsObject] ++ MembershipInfo.defaultWrites.writes(o.membershipInfo).as[JsObject]
```
ryanpbrewster, line 58: I think this is not quite what Derek and JP were talking about. They wanted something like `{name: "Kifi", ownerId: 3}` for the `OrganizationView`, and `{name: "Kifi", ownerId: 3, membershipInfo: {isInvited: true, role: "nonmember"}}`. So not quite flat, but flatter. @jaredpetker @DerekBurch correct me if I'm wrong.

I think the reasoning is that this way it can be implemented using inheritance on their end, which is more convenient for code-reuse than containment.

ryanpbrewster:
This looks awesome to me. When I first wrote the code that returned `viewer_permisisons` I was wary that we were heading down a bad path, but this seems very nice. Let's figure out a method for serialization that makes everyone happy (or at least happy-ish) and then push this.

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileOrganizationController.scala
camhashemi: I agree, shouldn't be a problem to cross that bridge when we get there

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
camhashemi, line 112: If `membershipOpt.isDefined`, you don't need the org.

The next thing I'm going to do is walk through the email invite flow. So I'll take care of that case then.

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationInfo.scala
camhashemi, line 58: Whatever's clever, it's only a different serialization for us.

The thing that we are less flexible about is providing a single field outside of its object. E.g. if you want `isInvited`, you're also going to get `permissions` and `role`. That keeps the api consistent at the cost of providing some unnecessary fields sometimes. Separation of concerns should help reduce that cost.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
ryanpbrewster, line 112: Oh, you're right. I got confused because we always give invitees permission to view the organization.

I don't really know what the intended behavior is if a secret organization invites a person via email and that person visits `kifi.com/orghandle` before actually signing up for Kifi. I suppose we ought to just tell them to sign up (at which point we convert their email invite into a user id invite and everything works just fine.)

camhashemi, line 112: Your questions are one step ahead of my answers. Working on this today.

leogrim, line 112: Invitees should be able to see the org, _not_ the organization-visible libraries. They should see a message such as "You've been invited to join this organization. Accept to get full access its content."

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationInfo.scala
leogrim, line 58: What's the benefit of naming fields `organizationInfo` and `membershipInfo` over just `organization` and `membership`? Does the more mouthful version bring any value?

camhashemi, line 58: `organization` is used for the wrapper, so the mouthful has to go here or there. I'm indifferent to where. `membership` works for me.

ryanpbrewster, line 58: Could also change it to `{"view": {"organization": {...}, "membership": {...}}`.

PR #16309, opened 2015-07-20
amplitude server side integration

joshm1:
@andrewconner this is still a WIP but I think it's best to get this out before everything per the spec is final. there are some temporary workarounds that I don't intend to keep around but I'd rather get events flowing into amplitude to get a better eye for what we've got and iterate on that.  for example, there are many events and properties Jen wants killed, i filtered some out in the amplitude client but the real solution is to eliminate them from being sent to heimdal in the first place

### DIFF
--- a/kifi-backend/modules/heimdal/app/com/keepit/heimdal/AmplitudeClient.scala
+++ b/kifi-backend/modules/heimdal/app/com/keepit/heimdal/AmplitudeClient.scala
@@ -1,0 +1,125 @@
+package com.keepit.heimdal
+
+import com.google.common.base.CaseFormat
+import com.keepit.common.akka.SafeFuture
+import com.keepit.common.db.Id
+import com.keepit.model.User
+import play.api.Play.current
+import play.api.libs.concurrent.Execution.Implicits.defaultContext
+import play.api.libs.json._
+import play.api.libs.ws.WS
+
+import scala.concurrent.Future
+
+object AmplitudeClient {
+  val killedProperties = Set("agent", "agentVersion", "client", "clientBuild", "clientVersion", "daysSinceLibraryCreated",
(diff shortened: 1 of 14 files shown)


joshm1:
@leogrim can you take a look? i'd like to get this deployed asap

kifi-backend/modules/heimdal/app/com/keepit/heimdal/AmplitudeClient.scala
```diff
@@ -0,0 +1,125 @@
...
+      filterKeys(key => !AmplitudeClient.killedProperties.contains(key)).
+      partition {
+        // return true if the property is a "user property" (not related to the particular event)
+        case (key, _) =>
+          // ugly - it'd be nicer if a HeimdalContext was aware of which properties are user properties vs event properties,
+          // but that'll require a decent sized refactor which isn't necessary to get this the initial integration out the door
+          (key.startsWith("user") && key != "userAgent") ||
+            key.startsWith("exp_") ||
+            key.startsWith("kcid") ||
+            Seq("gender").contains(key)
```
leogrim, line 76: :+1:

kifi-backend/modules/heimdal/test/com/keepit/test/HeimdalApplication.scala
```diff
@@ -21,7 +21,7 @@ import com.keepit.heimdal.HeimdalServiceTypeModule
 import com.keepit.inject.{ ApplicationInjector, FakeFortyTwoModule }
 import com.keepit.shoebox.FakeShoeboxServiceClientModule
 
-class HeimdalApplication(overridingModules: Module*)(implicit path: File = new File("./modules/shoebox/"))
+class HeimdalApplication(overridingModules: Module*)(implicit path: File = new File("./modules/heimdal/"))
```
leogrim, line 24: whoops

leogrim:
lgtm, let's give it a try!

PR #16313, opened 2015-07-20
Add generic flash handling, show warnings for org creation, change or…

maxov:
…g list order

@eishay 
@ryanpbrewster


andrewconner:
Nice, good primitive for us to use for errors.

PR #16323, opened 2015-07-21
Org domain ownership

maxov:
@camhashemi 
@leogrim 
@ryanpbrewster 

Organizations can now "own" domains.


kifi-backend/modules/common/conf/evolutions/shoebox/359.sql
```diff
@@ -0,0 +1,31 @@
...
+  state varchar(20) NOT NULL,
+  seq bigint(20) not null,
+  organization_id bigint(20) NOT NULL,
+  domain_id bigint(20) NOT NULL,
+
+  PRIMARY KEY (id),
+  CONSTRAINT organization_domain_ownership_organization_id FOREIGN KEY (organization_id) REFERENCES organization(id),
+  CONSTRAINT organization_domain_ownership_domain_id FOREIGN KEY (domain_id) REFERENCES domain(id),
+  CONSTRAINT unique_organization_id_and_domain_id UNIQUE (organization_id, domain_id)
+);
```
ryanpbrewster, line 26: Or even that `domain_id` is `UNIQUE`?

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminOrganizationController.scala
```diff
@@ -224,4 +227,24 @@ class AdminOrganizationController @Inject() (
...
   }
+
+  def addDomainOwnership(orgId: Id[Organization]) = AdminUserAction(parse.tolerantFormUrlEncoded) { implicit request =>
+    val body = request.body
+    val domainName = body.get("domainName").flatMap(_.headOption).get
+    val orgView = com.keepit.controllers.admin.routes.AdminOrganizationController.organizationViewById(orgId)
+    orgDomainOwnershipCommander.addDomainOwnership(orgId, domainName) match {
+      case Left(failure) =>
+        Redirect(orgView).flashing(
+          "error" -> failure.humanString
```
ryanpbrewster, line 238: :grin: I like the naming convention.

kifi-backend/modules/common/conf/evolutions/shoebox/359.sql
maxov, line 26: Yes, a domain can only be owned by an organization once. I don't see how multiple ownerships between the same `(organization_id, domain_id)` pair makes sense.

(btw, this constraint is enforced by the admin, it searches for existing ownerships, active or inactive before creating a new one)

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationDomainOwnershipCommander.scala
```diff
@@ -0,0 +1,81 @@
...
+}
+
+object OrganizationDomainOwnershipCommander {
+
+  sealed trait OwnDomainFailure {
+    def humanString: String
+  }
+
+  case class DomainDidNotExist(domainName: String) extends OwnDomainFailure {
+    override def humanString = s"Domain '$domainName' did not exist!"
```
ryanpbrewster, line 25: "has not been claimed"?

maxov, line 25: Well the data model permits a domain to be owned by many organizations (we will work out the meaning of this later when using the model). So multiple organizations can 'claim' a domain, which I think would give this error a bit of a misleading title. The error that _can_ arise is the one as it is right now, i.e. the domain is not actually registered in our system.

kifi-backend/modules/common/conf/evolutions/shoebox/359.sql
ryanpbrewster, line 26: If someone else writes code in the future that interfaces with the `domain` table, I'd rather their broken code actually break than silently succeed erroneously. I'm a big fan of database constraints.

Does it make sense to go so far as to ensure `domain_id` is `UNIQUE`?

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationDomainOwnershipCommander.scala
ryanpbrewster, line 25: Does that actually make sense? Do we want to allow IBM and IBM-Watson to own `ibm.com`?

maxov, line 25: I talked with @eishay and @leogrim before doing these changes, and we concluded that despite the additional complexity, allowing multiple organizations to own one domain would be worth it and necessary. For hierarchies of organizations, Google and Nest for example (where Nest is a sub-organization of Google), it is especially useful to say that all new accounts from `email@nest.com` should be allowed into both Google and Nest.

kifi-backend/modules/common/conf/evolutions/shoebox/359.sql
maxov, line 26: The way I interpret this constraint is that it enforces that the pair `(organization_id, domain_id)` is unique, not that either `domain_id` or `organization_id` are unique on their own. This means that there can only exist one row in the domain ownership table between the same organization and domain. It changes nothing with code interfacing with the domain or organization tables, except for when records are deleted and we come upon a reference constraint violation. From my tests this is how the constraint behaves, but perhaps that's not how it is supposed to?

I should elaborate that the constraint isn't strictly-speaking "enforced" by the admin. Trying to add an organization as an owner to a domain twice will cause a silent error in the admin (redirecting to the organization page) instead of a 500 exception page because the constraint is violated.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationDomainOwnershipCommander.scala
ryanpbrewster, line 25: That...makes perfect sense.

I still think "has not been claimed" fits here. If anyone has claimed it, it will necessarily exist in the system.

kifi-backend/modules/common/conf/evolutions/shoebox/359.sql
ryanpbrewster, line 26: Correct, regarding the `(a,b)` uniqueness.

The database constraint is very useful, though. It means that if someone is less careful than you have been (say they don't check to see if the pair `(a,b)` already exists, they just add a new one) the db will catch it and fail and we can correct the bug. It also happens to help with race conditions and other hard-to-protect-against errors.

ryanpbrewster, line 26: Re: the discussion below, I think that the way you have it is right. Multiple orgs should be able to own a domain.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationDomainOwnershipCommander.scala
maxov, line 25: I worry that people could confuse "claiming" a domain with "owning" a domain in this case. They could get this error and think, "If this domain has not been claimed, why can't I claim/own it?"

maxov:
@ryanpbrewster @leogrim @camhashemi does this look good to merge?

kifi-backend/modules/common/app/com/keepit/model/OrganizationDomainOwnership.scala
```diff
@@ -0,0 +1,51 @@
...
+}
+
+@json
+case class IngestableOrganizationDomainOwnership(
+  id: Id[OrganizationDomainOwnership],
+  createdAt: DateTime,
+  state: State[OrganizationDomainOwnership],
+  seq: SequenceNumber[OrganizationDomainOwnership],
+  organizationId: Id[Organization],
+  domainId: Id[Domain])
```
camhashemi: newline should end the file

ryanpbrewster:
:+1: lgtm, go for it. Nice work. Maybe fix the EOF newlines first, though.

kifi-backend/modules/common/conf/evolutions/shoebox/359.sql
camhashemi, line 26: On another note: we should have an index on `seq`.

kifi-backend/modules/graph/app/com/keepit/graph/manager/GraphUpdate.scala
```diff
@@ -211,3 +211,14 @@ case object UserIpAddressGraphUpdate extends GraphUpdateKind[UserIpAddressGraphU
...
+case class OrganizationDomainOwnershipGraphUpdate(orgId: Id[Organization], domainId: Id[Domain], state: State[OrganizationDomainOwnership], orgDomainOwnSeq: SequenceNumber[OrganizationDomainOwnership]) extends GraphUpdate {
+  type U = OrganizationDomainOwnershipGraphUpdate
+  def kind: GraphUpdateKind[U] = OrganizationDomainOwnershipGraphUpdate
+  def seq: SequenceNumber[U] = kind.seq(orgDomainOwnSeq.value)
+}
+
+case object OrganizationDomainOwnershipGraphUpdate extends GraphUpdateKind[OrganizationDomainOwnershipGraphUpdate] {
+  val code = "organization_domain_ownership_update"
+  def apply(orgDomainOwn: IngestableOrganizationDomainOwnership): OrganizationDomainOwnershipGraphUpdate = OrganizationDomainOwnershipGraphUpdate(orgDomainOwn.organizationId, orgDomainOwn.domainId, orgDomainOwn.state, orgDomainOwn.seq)
+}
```
camhashemi: newline http://stackoverflow.com/questions/729692/why-should-files-end-with-a-newline

ryanpbrewster: newline

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationDomainOwnershipCommander.scala
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  override def removeDomainOwnership(orgId: Id[Organization], domainId: Id[Domain]): Unit = {
+    db.readWrite { implicit session =>
+      orgDomainOwnershipRepo.getDomainOwnershipBetween(orgId, domainId) foreach { ownership =>
+        orgDomainOwnershipRepo.save(ownership.copy(state = OrganizationDomainOwnershipStates.INACTIVE))
+      }
+    }
+  }
+}
```
ryanpbrewster: newline

kifi-backend/modules/shoebox/test/com/keepit/model/OrganizationDomainOwnershipRepoTest.scala
```diff
@@ -0,0 +1,66 @@
...
+
+        db.readOnlyMaster { implicit s =>
+          orgDomainOwnershipRepo.getOwnershipsForOrganization(org1Id) must containTheSameElementsAs(orgDomainOwnerships.filter(_.organizationId == org1Id))
+        }
+      }
+    }
+
+  }
+
+}
```
camhashemi, line 66: there's also a sanity-checking `GraphManagerTest` for adding new elements to the graph

camhashemi, line 66: newline

kifi-backend/modules/common/app/com/keepit/model/OrganizationDomainOwnership.scala
```diff
@@ -0,0 +1,51 @@
...
+object OrganizationDomainOwnershipStates extends States[OrganizationDomainOwnership]
+
+case class OrganizationDomainOwnership(
+    id: Option[Id[OrganizationDomainOwnership]] = None,
+    createdAt: DateTime = currentDateTime,
+    updatedAt: DateTime = currentDateTime,
+    state: State[OrganizationDomainOwnership] = OrganizationDomainOwnershipStates.ACTIVE,
+    seq: SequenceNumber[OrganizationDomainOwnership] = SequenceNumber.ZERO,
+    organizationId: Id[Organization],
+    domainId: Id[Domain]) extends ModelWithSeqNumber[OrganizationDomainOwnership] with ModelWithState[OrganizationDomainOwnership] {
```
leogrim, line 20: I'd rather keep this table independent from the large `Domain` table, have the actual domain name in there, possibly with a referential integrity constraint based on the domain name.

kifi-backend/modules/common/conf/evolutions/shoebox/359.sql
leogrim, line 26: Actually we explicitly decided with Eishay that a domain should be owned by a single organization, so there should be a uniqueness constraint on `domain_name` (see my comment above about adding it).

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminOrganizationController.scala
leogrim, line 238: We also use `pretty` in many places, as in `toPrettyString`

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationDomainOwnershipCommander.scala
leogrim, line 25: Wait this is not what we concluded though. We said that in the future, we could allow anyone from one organization to join another (say Nest to Google), but that it was better to be explicit about a single org owning a domain.

PR #16325, opened 2015-07-21
decrease ttl for org's user-recos cache

camhashemi:
@eishay @leogrim TTL for org reco's outer repo was 7 days... not sure what the best TTL should be but that's what's holding them up.


kifi-backend/modules/shoebox/app/com/keepit/common/cache/ShoeboxCacheModule.scala
```diff
@@ -345,7 +345,7 @@ case class ShoeboxCacheModule(cachePluginModules: CachePluginModule*) extends Ca
 
   @Provides @Singleton
   def sociallyRelatedEntitiesForOrgCache(stats: CacheStatistics, accessLog: AccessLog, innerRepo: InMemoryCachePlugin, outerRepo: FortyTwoCachePlugin) =
-    new SociallyRelatedEntitiesForOrgCache(stats, accessLog, (innerRepo, 10 minutes), (outerRepo, 7 days))
+    new SociallyRelatedEntitiesForOrgCache(stats, accessLog, (innerRepo, 10 minutes), (outerRepo, 1 hours))
```
leogrim: You may want to leave that one longer than the other one above, otherwise there's no point in refreshing asynchronously.

camhashemi: I'm not sure that it was refreshing though. Because the TTL for the other one was 12 hours but I waited on recos to show up over the weekend and they didn't.

camhashemi: increased to 6 hours

PR #16327, opened 2015-07-21
Change the role "OWNER" to "ADMIN"

ryanpbrewster:
Sorry frontend/mobile

@cvializ @jaredpetker @DerekBurch @ajkochanowicz 

Merging this in 20 minutes or so.

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/model/Organization.scala
+++ b/kifi-backend/modules/common/app/com/keepit/model/Organization.scala
@@ -78,7 +78,7 @@ object Organization extends ModelWithPublicIdCompanion[Organization] {
     BasePermissions(Map(
       None -> Set(OrganizationPermission.VIEW_ORGANIZATION),
 
-      Some(OrganizationRole.OWNER) -> OrganizationPermission.all,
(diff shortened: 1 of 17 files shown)


cvializ:
Will the actual owner still have an `Owner` role, or will we know they're the owner from the org's `ownerId`?

ryanpbrewster:
The latter. They will have an `ADMIN` role.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/OrganizationController.scala
```diff
@@ -45,11 +45,13 @@ class OrganizationController @Inject() (
   }
 
   def modifyOrganization(pubId: PublicId[Organization]) = OrganizationUserAction(pubId, OrganizationPermission.EDIT_ORGANIZATION)(parse.tolerantJson) { request =>
+    println("[RPB] modifyOrg called with body: " + request.body)
```
maxov: I don't think these should be there, they look like debug statements?

ryanpbrewster: Good catch, I was just going through to remove my debug statements.

PR #16328, opened 2015-07-21
Do not send welcome email in dev

maxov:
it gets quite annoying

@camhashemi

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/UserCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/UserCommander.scala
@@ -26,7 +26,6 @@ import com.keepit.search.SearchServiceClient
 import com.keepit.social.{ BasicUser, SocialNetworks, UserIdentity }
 import com.keepit.typeahead.{ KifiUserTypeahead, SocialUserTypeahead, TypeaheadHit }
 import com.kifi.macros.json
-import play.api.Play
 import play.api.libs.json.{ JsObject, JsString, JsSuccess, _ }
 import securesocial.core.{ Identity, Registry, UserService }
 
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/UserCommander.scala
```diff
@@ -389,7 +390,7 @@ class UserCommander @Inject() (
   }
 
   def sendWelcomeEmail(newUser: User, withVerification: Boolean = false, targetEmailOpt: Option[EmailAddress] = None, isPlainEmail: Boolean = true): Future[Unit] = {
-    if (!db.readOnlyMaster { implicit session => userValueRepo.getValue(newUser.id.get, UserValues.welcomeEmailSent) }) {
+    if (!db.readOnlyMaster { implicit session => userValueRepo.getValue(newUser.id.get, UserValues.welcomeEmailSent) } && Play.isProd) {
```
camhashemi: we decided against this, yeah?

ryanpbrewster: The welcome email has a purpose. I set up a filter in my email to automatically archive it and tag it.

maxov: @ryanpbrewster Could you elaborate, what purpose does the email have in development?

ryanpbrewster: It is used to verify that signup worked. This isn't a huge issue anymore, since we don't mess with signup very often, but if we ever decide to change the signup email (or the way that signup works) then it's useful to see that in dev mode.

camhashemi: This is what I use: https://team42.atlassian.net/wiki/display/ENG/Email+Overview. Even in dev mode

andrewconner:
Looks good to me.

PR #16329, opened 2015-07-21
Org lib create

ryanpbrewster:
You can now create libraries directly in organizations (if you have the requisite permissions). To do this, include an (optional) `space: {"org": "<public_org_id>"}` in the add request sent to the controller.


ryanpbrewster:
@jaredpetker @camhashemi @cvializ

PR #16335, opened 2015-07-21
Fixed a site routing issue

ryanpbrewster:
Credit to @leogrim

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/KifiSiteRouter.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/KifiSiteRouter.scala
@@ -130,22 +130,32 @@ class KifiSiteRouter @Inject() (
 
   def serveWebAppIfLibraryFound(handle: Handle, slug: String) = WebAppPage { implicit request =>
     lookupHandle(handle) flatMap {
-      case (handleOwner, userRedirectStatusOpt) =>
+      case (handleOwner, spaceRedirectStatusOpt) =>
+        val handleSpace: LibrarySpace = handleOwner match {
+          case Left(org) => org.id.get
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/app/com/keepit/controllers/website/KifiSiteRouter.scala
```diff
@@ -130,22 +130,32 @@ class KifiSiteRouter @Inject() (
...
             }
-          case Right(user) => libraryCommander.getLibraryBySlugOrAlias(user.id.get, LibrarySlug(slug))
+          case Right(user) =>
+            libraryCommander.getLibraryBySlugOrAlias(handleSpace, LibrarySlug(slug))
         }
         libraryOpt map {
           case (library, isLibraryAlias) =>
-            if (library.slug.value != slug || userRedirectStatusOpt.isDefined) { // library moved
+            val hasUserMovedThisLibrary = isLibraryAlias
+            val wasLibrarySlugNormalized = library.slug.value != slug
```
leogrim, line 152: Note that this is not just normalization, it might be a change of slug entirely if `isLibraryAlias` is true. 
And actually the Boolean seems the other way around regardless

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/KifiSiteRouter.scala
```diff
@@ -130,22 +130,32 @@ class KifiSiteRouter @Inject() (
...
+            val hasUserChangedTheirUsername = spaceRedirectStatusOpt.isDefined
```
leogrim, line 153: This is true is `spaceRedirectStatusOpt` contains a 301, otherwise it might just be a space normalization issue.

PR #16336, opened 2015-07-21
Kifi Site router for libraries

ryanpbrewster:
@leogrim I was more explicit with the checks this time. I think that in the end the logic is entirely unchanged. This is strictly a human-readability PR.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/KifiSiteRouter.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/KifiSiteRouter.scala
@@ -148,14 +148,20 @@ class KifiSiteRouter @Inject() (
         }
         libraryOpt map {
           case (library, isLibraryAlias) =>
-            val hasUserMovedThisLibrary = isLibraryAlias
-            val wasLibrarySlugNormalized = library.slug.value != slug
-            val hasUserChangedTheirUsername = spaceRedirectStatusOpt.isDefined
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/app/com/keepit/controllers/website/KifiSiteRouter.scala
```diff
@@ -148,14 +148,20 @@ class KifiSiteRouter @Inject() (
...
         libraryOpt map {
           case (library, isLibraryAlias) =>
-            val hasUserMovedThisLibrary = isLibraryAlias
-            val wasLibrarySlugNormalized = library.slug.value != slug
-            val hasUserChangedTheirUsername = spaceRedirectStatusOpt.isDefined
-            if (hasUserMovedThisLibrary || wasLibrarySlugNormalized || hasUserChangedTheirUsername) {
+            val libraryHasBeenMoved = isLibraryAlias
+            val handleOwnerChangedTheirHandle = spaceRedirectStatusOpt.contains(MOVED_PERMANENTLY)
+            val wasLibrarySlugNormalized = !libraryHasBeenMoved && library.slug.value != slug
+            val wasHandleNormalized = spaceRedirectStatusOpt.contains(SEE_OTHER)
```
leogrim, line 154: This is all true, though leaks / is implicitly aware of the internal logic of `lookupHandle` :)

ryanpbrewster, line 154: Can you think of a way to make it oblivious to the logic there? Maybe I need to rewrite `lookupHandle`.

PR #16337, opened 2015-07-21
Removing a few no longer needed experiments and related code

stkem:
@leogrim @andrewconner @eishay 
(Not all the ones we deemed to be removed, but it's a start)


kifi-backend/modules/curator/test/com/keepit/curator/FeedDigestEmailSenderTest.scala
```diff
@@ -155,6 +155,9 @@ class FeedDigestEmailSenderTest extends Specification with CuratorTestInjector w
       withDb(modules: _*) { implicit injector =>
         val (shoebox, sender, savedRecoModels, user1, user2, friends) = setupToSend(db)
 
+        shoebox.newKeepsInLibrariesExpectation(user2.id.get) = Seq.empty
+        shoebox.newKeepsInLibrariesExpectation(user1.id.get) = Seq.empty
```
ryanpbrewster, line 159: `===`

stkem, line 159: This is not a test. I'm assigning something to a key in map. :-)

ryanpbrewster, line 159: Oops, my mistake. Not used to mutable data structures.

PR #16338, opened 2015-07-22
Only execute a prefix search for a trailing query term

leogrim:
@stkem @andrewconner 
Basically, if there's something (including a whitespace) after a term, we should consider that the user is done typing that term and shouldn't use it for a prefix search.

Previously: 
- `Léo Grim` would do a prefix search with both `Léo` and `Grim` and `Léon Grimaldi` would match.

Now:
- `Léo Grim` will do a prefix search on `Grim` but not `Léo`, e.g. `Léon Grimaldi` won't match
- `Léo_` (_ representing a trailing whitespace) will not do a prefix search at all

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/search/SearchConfig.scala
+++ b/kifi-backend/modules/common/app/com/keepit/search/SearchConfig.scala
@@ -21,7 +21,8 @@ object SearchConfig {
     Map[String, String](
       // Common Query Parser
       "titleBoost" -> "2.0",
-      "prefixBoost" -> "1.0",
+      "prefixBoost" -> "0.0",
+      "trailingPrefixBoost" -> "0.8",
       "phraseBoost" -> "0.33",
       "siteBoost" -> "1.0",
       "concatBoost" -> "0.8",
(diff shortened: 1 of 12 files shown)


ryanpbrewster:
The way that `IntelliJ` searches for things is the opposite of this, and in a way that I find useful.

Does `stats` match `statistics`? I guess that is a separate issue...

What about "org endpoints" matching "organization endpoints"?

stkem:
code lgtm. What was the motivation here? Doesn't the old way let you find more things. e.g, if you know only part of the first name. Say I'm looking for Tom, and I don't remember his last name, but I know it starts with a D. I search for "Tom D", and as it might turn out he's in the system as "Tomas D...". New way won't find him, right? Contrived example I admit.

leogrim:
Well, sometimes let's say I search for `Stephen Kemmerling`, it can be annoying if `Stephenie Kemmerling` is a match, not a great example either :)

leogrim:
What about simply boosting a trailing term prefix query as opposed to remove non-trailing ones?

stkem:
Yeah, I think that makes sense. If you type "Léo Grim", "Léo Grimaldi" should show up before "Leonardo Grimaldi", since you had one exact match.

PR #16339, opened 2015-07-22
Fixed library joining logic to account for orgs

ryanpbrewster:
@camhashemi @leogrim 

`LibraryCommander.joinLibrary` was one of the functions we were supposed to update to include organization logic. Pretty simple change.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
@@ -1112,11 +1112,14 @@ class LibraryCommanderImpl @Inject() (
       (lib, allInvites, existingActiveMembership)
     }
 
-    if (lib.kind == LibraryKind.SYSTEM_MAIN || lib.kind == LibraryKind.SYSTEM_SECRET) {
+    val isSystemGeneratedLibrary = lib.kind == LibraryKind.SYSTEM_MAIN || lib.kind == LibraryKind.SYSTEM_SECRET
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -1112,11 +1112,14 @@ class LibraryCommanderImpl @Inject() (
       (lib, allInvites, existingActiveMembership)
     }
 
-    if (lib.kind == LibraryKind.SYSTEM_MAIN || lib.kind == LibraryKind.SYSTEM_SECRET) {
+    val isSystemGeneratedLibrary = lib.kind == LibraryKind.SYSTEM_MAIN || lib.kind == LibraryKind.SYSTEM_SECRET
+    val userCanJoinLibraryWithoutInvite = canViewLibrary(Some(userId), lib, authToken) // uses a db session
```
ryanpbrewster, line 1116: Someone double-check my thought process here. "joining" defaults to `READ_ONLY` unless there is explicitly greater permissions. I think we need to add a new `OrganizationPermission` to deal with the case where org members can automatically add themselves as collaborators to org libraries.

andrewconner:
Mergeable?

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
leogrim, line 1116: We probably want the default to be joining org libraries as collaborators if you're part of the org. You're right that we may want to encode that as an explicit permissions though.

ryanpbrewster:
Yes, let's do this.

PR #16342, opened 2015-07-22
batch interning domain names

camhashemi:
@leogrim trying to optimize and resolve this https://fortytwo.airbrake.io/projects/97986/groups/1450329687680885002/notices/1469966177847222475

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/classify/DomainRepo.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/classify/DomainRepo.scala
@@ -63,12 +63,19 @@ class DomainRepoImpl @Inject() (
   }
 
   def internAllByNames(domainNames: Set[String])(implicit session: RWSession): Map[String, Domain] = {
-    domainNames.map { domainName =>
-      get(domainName, None) match {
-        case Some(domain) if domain.state == DomainStates.INACTIVE => domainName -> domain.copy(state = DomainStates.ACTIVE)
-        case Some(domain) => domainName -> domain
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/app/com/keepit/classify/DomainRepo.scala
```diff
@@ -63,12 +63,19 @@ class DomainRepoImpl @Inject() (
...
-    domainNames.map { domainName =>
-      get(domainName, None) match {
-        case Some(domain) if domain.state == DomainStates.INACTIVE => domainName -> domain.copy(state = DomainStates.ACTIVE)
-        case Some(domain) => domainName -> domain
-        case _ => domainName -> save(Domain(hostname = domainName))
+    val existingDomains = getAllByName(domainNames.toSeq).toSet
+
+    val existingDomainByName = existingDomains.map { domain =>
+      domain.state match {
+        case DomainStates.INACTIVE => domain.hostname -> domain.copy(state = DomainStates.ACTIVE)
```
leogrim, line 70: When you intern a model and end up reusing an inactive one, you alway want to make sure you reset all other fields to their default value. Otherwise you might get some unexpected zombie values.
You can either wipe out fields manually, or the safest way is to create a brand new `Domain` object and copy the existing row 'id' into it.

camhashemi, line 70: For example, domain has `isEmailProvider = false` and `autoSensitive = None`. Are you saying that since it's been inactive, we can't assume that the old values still hold?

Should we at least keep `createdAt`?

PR #16347, opened 2015-07-23
use lowercase hostnames in DomainRepo

camhashemi:
@leogrim ContactInterner carries much more complexity than this case. hopefully this does the trick, please let me know if you see anything wrong.

one thing that I noticed in the MySQLIntegrityException is that the failed hostname contained this character "ı"? Not sure how this is handled.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/classify/DomainRepo.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/classify/DomainRepo.scala
@@ -63,7 +63,8 @@ class DomainRepoImpl @Inject() (
   }
 
   def internAllByNames(domainNames: Set[String])(implicit session: RWSession): Map[String, Domain] = {
-    val existingDomains = getAllByName(domainNames.toSeq, None).toSet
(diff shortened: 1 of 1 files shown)


camhashemi:
@leogrim I think I got it. In MySQL, 'ı' == 'i'. In Scala, 'ı' != 'i'. So we ask for 'ı', get back 'i', then try to save 'ı', but MySQL gives a constraint exception since 'i' exists.

kifi-backend/modules/shoebox/app/com/keepit/classify/DomainRepo.scala
```diff
@@ -72,7 +73,7 @@ class DomainRepoImpl @Inject() (
       }
     }.toMap
 
-    val newDomainByName = domainNames.diff(existingDomainByName.keys.toSet).map { domainName =>
+    val newDomainByName = lowerCasedDomainNames.diff(existingDomainByName.keys.toSet).map { domainName =>
```
leogrim, line 76: This is not correct though, you're not interning the lower-cased domain name. You should intern the original domain name, but diff / key by lower-cased name, that's what the EContact stuff does

PR #16349, opened 2015-07-23
Feature/org views

adamjgrant:
Pending...


kifi-backend/modules/shoebox/angular/src/orgProfile/orgProfileMember.js
```diff
@@ -2,35 +2,8 @@
...
-
-    function _toggleControls() {
-      var $scope = this;
-      if ($scope._controlsOpen) {
-        $scope.close();
-      } else {
-        $scope.open();
-      }
-    }
-
```
adamjgrant, line 33: @cvializ I undid a few things here mainly to get it working. The dropdown arrow stopped toggling the menu when I moved it in the DOM. I wasn't sure what the strategy here was for (functions made in duplicate and setting $scope equal to this in each one). Let's talk about it when you're back.

andrewconner:
Wahoo, can't wait to play with it.

adamjgrant:
Not finished, but probably a good point to do a derploy. @andrewconner @leogrim @cvializ @camhashemi

kifi-backend/modules/shoebox/angular/img/symbol-sprites/lib/dropdown-arrow.svg
```diff
@@ -1,7 +1,10 @@
...
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg width="512px" height="512px" viewBox="0 0 512 512" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <defs></defs>
+    <g id="Page-1" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
+        <g id="dropdown-arrow">
+            <g id="Guide" transform="translate(-3.000000, -3.000000)"></g>
+            <path d="M224.278981,361.469297 L245.159913,382.346188 C251.016757,388.201899 260.509092,388.198604 266.361847,382.338616 L287.227863,361.446816 L456.812844,191.652276 C462.666296,185.79159 462.665472,176.300439 456.800433,170.442638 L415.051129,128.744862 C409.190823,122.891788 399.710974,122.907267 393.868815,128.787971 L256,267.566406 L116.132635,127.507602 C110.281062,121.64802 100.794827,121.641407 94.933974,127.503395 L53.2144635,169.23098 C47.358339,175.088238 47.3535959,184.57815 53.2134436,190.436864 L224.278981,361.469297 Z" id="Shape" fill="#909090"></path>
+        </g>
+    </g>
+</svg>
```
adamjgrant, line 10: The only thing I did differently here is to use the icon template on dropbox to keep measurements consistent. You'll notice the arrow is a little fatter as well.

kifi-backend/modules/shoebox/angular/src/orgProfile/orgProfileMember.js
cvializ, line 33: @adam, I'm not sure why it stopped working, but I can explain the strategy. I wanted to create these functions outside of the `link` function to reduce their memory footprint. If inside the `link` function, these functions are instantiated once per member view. When outside the link function, I leverage the `this` binding instead of a closure to enable access to the `$scope`. If it's too confusing this could be a premature optimization, but I thought that some orgs might have a lot of members and that a smaller memory footprint for this view would help.

kifi-backend/modules/shoebox/angular/img/symbol-sprites/lib/dropdown-arrow.svg
cvializ, line 10: Ooh I didn't know we had an icon template. I'll check that out

adamjgrant, line 10: @cvializ Here is where you'll find it:

screen shot 2015-07-24 at 9 51 25 am

There's also a guide layer you can turn on for reference shapes and sizing:

screen shot 2015-07-24 at 9 52 02 am

kifi-backend/modules/shoebox/angular/src/orgProfile/orgProfileMember.js
adamjgrant, line 33: Interesting. Requires a lot of duplication but I can see where the payoff is in terms of optimizations–especially where you have a lot of repeat data. I'm still new to this but I wonder if we could put some logic in `compile` instead of `link`? Gonna do some reading.

PR #16355, opened 2015-07-23
add site to organization model

camhashemi:
@leogrim @ajkochanowicz


maxov:
:+1:

kifi-backend/modules/common/conf/evolutions/shoebox/360.sql
```diff
@@ -0,0 +1,10 @@
+# SHOEBOX
+
+
+# --- !Ups
+
+ALTER TABLE organization ADD site VARCHAR(2048) DEFAULT NULL;
```
camhashemi: is this sizing ok?

andrewconner: Yeah, that's way larger than the maximum domain length.

kifi-backend/modules/common/app/com/keepit/model/Organization.scala
```diff
@@ -27,6 +27,7 @@ case class Organization(
     description: Option[String] = None,
     ownerId: Id[User],
     handle: Option[PrimaryOrganizationHandle],
+    site: Option[String] = None,
```
andrewconner: No need for default param here; use them to simplify an API where the default is nearly always the expected value (e.g., `excludedStates`). In this case, not a big deal for anyone needing to create an org (should only be one place? maybe a small few?) to include it.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
```diff
@@ -185,7 +188,12 @@ class OrganizationCommanderImpl @Inject() (
       // Are there any members that can't even see the organization?
       OrganizationRole.all exists { role => !(bps.permissionsMap.contains(Some(role)) && bps.forRole(role).contains(VIEW_ORGANIZATION)) }
     }
-    !badName && !badBasePermissions
+    val normalizedSiteUrl = modifications.site.map { url =>
+      if (url.startsWith("http://") || url.startsWith("https://")) url
+      else "https://" + url
```
andrewconner, line 193: Implied https? Would be cool if that is alright; it may be, we can check normalized_uri to get an idea of how often HTTPS is there.

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationRepo.scala
```diff
@@ -43,9 +43,10 @@ class OrganizationRepoImpl @Inject() (
     def ownerId = column[Id[User]]("owner_id", O.NotNull)
     def organizationHandle = column[Option[OrganizationHandle]]("handle", O.Nullable)
     def normalizedOrganizationHandle = column[Option[OrganizationHandle]]("normalized_handle", O.Nullable)
+    def site = column[Option[String]]("site", O.Nullable)
     def basePermissions = column[BasePermissions]("base_permissions", O.NotNull)
 
-    def * = (id.?, createdAt, updatedAt, state, seq, name, description, ownerId, organizationHandle, normalizedOrganizationHandle, basePermissions) <> ((Organization.applyFromDbRow _).tupled, Organization.unapplyToDbRow _)
+    def * = (id.?, createdAt, updatedAt, state, seq, name, description, ownerId, organizationHandle, normalizedOrganizationHandle, site, basePermissions) <> ((Organization.applyFromDbRow _).tupled, Organization.unapplyToDbRow _)
```
andrewconner, line 49: Since you added it to the end of the db table (far right side), mind making it that position here? Helps `GetResult`s later down the road.

andrewconner:
Looks good, merge when you're happy.

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationRepo.scala
leogrim, line 49: You can also use `ALTER TABLE organization ADD site VARCHAR(2048) AFTER normalized_handle DEFAULT NULL;` if you do care about positioning :)

leogrim:
:+1:

kifi-backend/modules/common/conf/evolutions/shoebox/360.sql
camhashemi: Looks like the max domain size is 255? so should we just do that to keep it optimal/constrained?

leogrim: This is not a domain though, it's a url.

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationRepo.scala
camhashemi, line 49: I did want to keep it right after `normalized_handle`. Thanks guys.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
camhashemi, line 193: For this case, it doesn't matter. We are just validating the URL and need this prefix for the parser to work.

camhashemi:
@leogrim @andrewconner updated

PR #16358, opened 2015-07-24
More resilient Zookeepering

stkem:
We already have a recovery mechanism for connection loss, but it had some weaknesses, perhaps most notably that it was async and not always called when there was a problem. This should help.

Also removed two unnecessary airbrakes which were responsible for some of the recent pagerduties.

No really all that much closer to understanding why this only recently started acting up, but this should (hopefully) reduce the problems.

This should be deployed carefully (i.e. probably in the morning)

@andrewconner @leogrim

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/common/controller/LoggingFilter.scala
+++ b/kifi-backend/modules/common/app/com/keepit/common/controller/LoggingFilter.scala
@@ -31,8 +31,6 @@ class LoggingFilter() extends EssentialFilter {
         val message = s"Current status of service ${myAmazonInstanceInfo.info.localIp} ${discovery.myStatus.getOrElse("UNKNOWN")}, last changed ${discovery.timeSinceLastStatusChange}ms ago"
         //system is going down, maybe the logger, emails, airbrake are gone already
         println(message)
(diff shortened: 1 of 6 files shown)


andrewconner:
Looks good to me. Hope we also find out what's going on network wise. Has it happened recently, or is it basically randomly continuous? Would love to see the topology and a ping from machines during the issues.

stkem:
Still going to look at the topology if I catch it on time (whoever is on call should probably try as well), but baed on the exceptions and what I've seen in the ZK logs, the view of the world of the affected machines is almost certainly just frozen from what it was before the most recent deploy (of anything).

leogrim:
I have tried looking up the topology when Rover is acting up, but I didn't catch anything interesting so far

andrewconner:
> almost certainly just frozen from what it was before the most recent deploy

Oddly, I once bounced rover and it didn't fix. Then I deployed (new version) and it fixed. May be a red herring, but...

stkem:
Yeah, the underlying issues seem to be something with the network, where it can't connect to ZK (hence the frozen state). This PR should help the process recover on it's own without needing a bounce/deploy (iff the network recovers). I think :-).

PR #16362, opened 2015-07-24
Route library view by handle

cvializ:
@andrewconner @ajkochanowicz


adamjgrant:
I think the only thing we're missing here is resolving `/:org/<library slug>` instead of `/:user/<library slug>` when clicking from org library page. Lemme see if I can get that in there. I'll merge in.

cvializ:
You mean that the library cards on the org page still link to the user library? I noticed that too. Ok, should I wait for you to merge that in before I merge this PR?

adamjgrant:
Yeah, but on second thought, we should go ahead and merge this. It might take me a little while longer to get that figured out.

:+1:

PR #16363, opened 2015-07-24
Feature/org lib route

adamjgrant:
Issue: https://github.com/kifi/keepit/issues/16389

Updating slug in create library modal on orgs page and updating data parsing for new org syntax per #16370


adamjgrant:
Clicking on library will use the handle of the `userOrOrg` type that is showing it.

e.g.
Clicking on Kifi Mission from /kifi will go to /kifi/kifi-mission not /leo/kifi-mission

@cvializ @andrewconner

kifi-backend/modules/shoebox/angular/src/route.js
```diff
@@ -159,13 +159,50 @@ angular.module('kifi')
...
+                throw new Error('could not find library in org with slug ' + slug);
+              }
+            }
+
+            if (type === 'user') {
+              // User library
+              return libraryService.getLibraryByUserSlug($stateParams.handle, $stateParams.librarySlug, $stateParams.authToken);
+            } else {
+              // Org Library
+              return net.userOrOrg($stateParams.handle)
```
andrewconner: Eesh, this may be indicative of an incomplete server API. 3 remote calls to get the library?

cvializ: Yes, when calling `net.getOrgLibraries`, the response returned is missing some properties that `libraryService.getLibraryById` returns, like library.keeps. I'll talk to @camhashemi

adamjgrant: Oh I didn't notice the third one. Empirically, it's also loading slowly.

adamjgrant: So this is because `type` is not defined in the `libraryService` response?

cvializ: No, it's because `keeps` is missing from the libraries returned by `/organizations/:id/libraries`

cvializ: The userOrOrg call is to determine whether the `:handle` is a username or an org name, and then use the correct API for getting the libraries

andrewconner: Sounds good. Loudly complain about broken APIs (this counts).

cvializ:
(note: don't merge this until #16370 is merged)

adamjgrant:
^ Noted.

PR #16366, opened 2015-07-24
add DomainHash to Domain

camhashemi:
@leogrim @gratimax


maxov:
Nice :+1:

kifi-backend/modules/shoebox/app/com/keepit/model/NormalizedURIRepo.scala
```diff
@@ -121,7 +121,7 @@ class NormalizedURIRepoImpl @Inject() (
       case e: Throwable =>
         deleteCache(cleaned)
         throw e
-    }
+    } // @leo, is this the right way to implement the "first persisted wins" behavior? seems unsafe, albeit atomic
```
camhashemi: @leogrim

kifi-backend/modules/common/app/com/keepit/classify/Domain.scala
```diff
@@ -46,6 +53,25 @@ object Domain {
   private val MaxLength = 128
 
   def isValid(s: String): Boolean = DomainRegex.findFirstIn(s).isDefined && s.length <= MaxLength
+
+  def withHash(hostname: String) = Domain(hostname = hostname, hash = DomainHash.hashHostname(hostname))
```
leogrim: This should lowercase, `withHost` may be a better name

kifi-backend/modules/common/app/com/keepit/classify/Domain.scala
```diff
@@ -17,6 +21,7 @@ case class Domain(
     autoSensitive: Option[Boolean] = None,
     manualSensitive: Option[Boolean] = None,
     isEmailProvider: Boolean = false,
+    hash: DomainHash,
```
leogrim: This should probably be optional until the existing data has been fixed

kifi-backend/modules/shoebox/app/com/keepit/admin/UrlController.scala
```diff
@@ -471,4 +473,23 @@ class UrlController @Inject() (
...
   }
+
+  def updateAllDomainHashes() = AdminUserAction { implicit request =>
+    val BATCH_SIZE = 10000
+    val numBatches = db.readOnlyReplica { implicit session => domainRepo.count / BATCH_SIZE }
+
+    def processBatch(dummyAcc: Unit, batch: Int): Future[Unit] = {
+      db.readWriteAsync { implicit session => // we don't really care when this call finishes, but not sure whether to use ...Async or ...Batch for this
+        val domainBatch = domainRepo.pageAscending(batch, BATCH_SIZE)
+        val updatedDomains = domainBatch.map(domain => domain.copy(updatedAt = DateTime.now(), hash = DomainHash.hashHostname(domain.hostname)))
```
leogrim: Let's also lowercase all domains. See my comment above, then just use `withHost`. 
The `updatedAt` field should be taken care of by the repo's `save` method :)

kifi-backend/modules/shoebox/app/com/keepit/classify/DomainRepo.scala
```diff
@@ -11,6 +11,7 @@ trait DomainRepo extends Repo[Domain] {
   def get(domain: String, excludeState: Option[State[Domain]] = Some(DomainStates.INACTIVE))(implicit session: RSession): Option[Domain]
   def getByIds(orgIds: Set[Id[Domain]])(implicit session: RSession): Map[Id[Domain], Domain]
   def getAllByName(domains: Seq[String], excludeState: Option[State[Domain]] = Some(DomainStates.INACTIVE))(implicit session: RSession): Seq[Domain]
+  def getAllByNameUsingHash(domains: Set[String], excludeState: Option[State[Domain]] = Some(DomainStates.INACTIVE))(implicit session: RSession): Set[Domain]
```
leogrim: This should not be exposed outside of the repo. `getAllByName` should rely on the hash internally

kifi-backend/modules/common/app/com/keepit/classify/Domain.scala
```diff
@@ -46,6 +53,25 @@ object Domain {
   private val MaxLength = 128
```
leogrim: This can be increased once the table relies on the hash for constraints / indexing.

kifi-backend/modules/common/app/com/keepit/classify/Domain.scala
```diff
@@ -46,6 +53,25 @@ object Domain {
...
+}
+
+case class DomainHash(hash: String) extends AnyVal {
+  override def toString: String = hash
+  def urlEncoded: String = hash.replaceAllLiterally("+" -> "-", "/" -> "_") // See RFC 3548 http://tools.ietf.org/html/rfc3548#page-6
+}
+
+object DomainHash {
+  def hashHostname(hostname: String): DomainHash = {
+    val binaryHash = MessageDigest.getInstance("MD5").digest(hostname.toLowerCase)
```
leogrim: The `hostname` field should always be lowercased, shouldn't it?

kifi-backend/modules/shoebox/app/com/keepit/model/NormalizedURIRepo.scala
leogrim: Check out `ArticleImageInfoRepo.intern` for a simple example

kifi-backend/modules/shoebox/app/com/keepit/classify/DomainRepo.scala
camhashemi: My goal is to kill getAllByName. This is just for backwards capability before the migration.

leogrim: You should keep 'getAllByName', use the hash internally for lookups

camhashemi: Yes definitely. What I meant was: since hash doesn't exist yet, I'm keeping `getAllByName` as is. After the migration, `getAllByNameUsingHash` is how `getAllByName` will be implemented.

leogrim: :+1:

kifi-backend/modules/common/app/com/keepit/classify/Domain.scala
camhashemi: Not sure what it should be increased to.

leogrim: 256 should be enough: https://en.wikipedia.org/wiki/Domain_Name_System

kifi-backend/modules/common/app/com/keepit/classify/Domain.scala
camhashemi: Maybe implicitly, but it's not a constraint that I see implemented anywhere in the model.

leogrim: No, I'm saying it should be. In the `withHash` / `withHost` method, and probably in a factory method that relies on it :)

kifi-backend/modules/common/app/com/keepit/classify/Domain.scala
```diff
@@ -17,9 +21,11 @@ case class Domain(
     autoSensitive: Option[Boolean] = None,
     manualSensitive: Option[Boolean] = None,
     isEmailProvider: Boolean = false,
+    hash: Option[DomainHash],
     state: State[Domain] = DomainStates.ACTIVE,
     createdAt: DateTime = currentDateTime,
     updatedAt: DateTime = currentDateTime) extends ModelWithState[Domain] {
+  require(this.hostname.toLowerCase == this.hostname, "Domain.hostname must be lowercase")
```
camhashemi: @leogrim only required when we don't use `Domain.withHostname`

kifi-backend/modules/common/app/com/keepit/classify/Domain.scala
```diff
@@ -37,15 +43,35 @@ object Domain {
...
     (__ \ 'updatedAt).format[DateTime]
   )(Domain.apply, unlift(Domain.unapply))
 
   private val DomainRegex = """^(?:[a-z0-9](?:[a-z0-9-]{0,61}[a-z0-9])?\.)+[a-z0-9]+$""".r
-  private val MaxLength = 128
+  private val MaxLength = 256
 
   def isValid(s: String): Boolean = DomainRegex.findFirstIn(s).isDefined && s.length <= MaxLength
+
+  def withHostname(hostname: String): Domain = Domain(hostname = hostname.toLowerCase, hash = Some(DomainHash.hashHostname(hostname.toLowerCase)))
```
camhashemi: @leogrim

camhashemi:
@leogrim updated. @gratimax is waiting on this so I want to get it out there. would appreciate a review.

kifi-backend/modules/common/app/com/keepit/classify/Domain.scala
leogrim: Great! We want to fix the existing data before adding this, otherwise you won't be able to deserialize from the database.

kifi-backend/modules/common/app/com/keepit/classify/Domain.scala
leogrim: What DRYing this into a very explicit `val lowerCasedHostname = hostname.toLowerCase` ?

kifi-backend/modules/shoebox/app/com/keepit/classify/DomainRepo.scala
```diff
@@ -69,18 +75,22 @@ class DomainRepoImpl @Inject() (
...
+    val existingDomains = getAllByName(lowerCasedHostnamesToIntern.toSeq, None).toSet
+
+    val existingHostnames = existingDomains.map(_.hostname)
+
+    val toBeInserted = (lowerCasedHostnamesToIntern -- existingHostnames).map(Domain.withHostname)
 
     val existingDomainByName = existingDomains.map { domain =>
       domain.state match {
-        case DomainStates.INACTIVE => domain.hostname -> save(domain.copy(state = DomainStates.ACTIVE))
+        case DomainStates.INACTIVE => domain.hostname -> save(Domain.withHostname(domain.hostname).copy(id = domain.id))
```
leogrim: :+1:

kifi-backend/modules/common/app/com/keepit/classify/Domain.scala
camhashemi: Okay, will comment out. Also, for the evolution/migration seems like the order is:

1) add `hash` to the table
2) merge + deploy shoebox
3) call migration endpoint

@leogrim edit:
4) `hash` should no longer be an Option/nullable
5) add `hash` as a unique index, then drop `hostname` unique
6) replace `getAllByName` implementation with `getAllByNameUsingHash`

kifi-backend/modules/common/app/com/keepit/classify/Domain.scala
camhashemi: np, I just prefer the one-liner. But explicit is more error-proof.

leogrim: DRY > one-line :)

kifi-backend/modules/shoebox/app/com/keepit/admin/UrlController.scala
```diff
@@ -471,4 +473,23 @@ class UrlController @Inject() (
...
   }
+
+  def updateAllDomainHashes() = AdminUserAction { implicit request =>
+    val BATCH_SIZE = 10000
+    val numBatches = db.readOnlyReplica { implicit session => domainRepo.count / BATCH_SIZE }
+
+    def processBatch(dummyAcc: Unit, batch: Int): Future[Unit] = {
+      db.readWriteAsync { implicit session =>
+        val domainBatch = domainRepo.pageAscending(batch, BATCH_SIZE)
+        val updatedDomains = domainBatch.map(domain => domain.copy(hostname = domain.hostname.toLowerCase, hash = Some(DomainHash.hashHostname(domain.hostname))))
```
leogrim: I think this is wrong, `DomainHash.hashHostname` doesn't lowercase, we should use `withHostname` here.

kifi-backend/modules/common/app/com/keepit/classify/Domain.scala
leogrim: Yep, but check out the migration code first, I think there's an error.

leogrim: 5) `hash` should no longer be an `Option` (in Scala) / nullable (in the database)

kifi-backend/modules/shoebox/app/com/keepit/admin/UrlController.scala
camhashemi: nice catch, thank you.

kifi-backend/modules/shoebox/app/com/keepit/admin/UrlController.scala
```diff
@@ -471,4 +473,23 @@ class UrlController @Inject() (
...
+        updatedDomains.foreach(domainRepo.save)
+        log.info(s"[hashMigration] domains ${domainBatch.head.id.get.id} - ${domainBatch.last.id.get.id} updated")
+        ()
+      }
+    }
+
+    FutureHelpers.foldLeft(1 to numBatches)(())(processBatch)
+
+    Ok
+  }
```
ryanpbrewster, line 494: Why is this using a `foldLeft` instead of a `map`? Do you use the accumulator anywhere?

kifi-backend/modules/common/app/com/keepit/classify/Domain.scala
```diff
@@ -37,15 +43,38 @@ object Domain {
...
     (__ \ 'updatedAt).format[DateTime]
   )(Domain.apply, unlift(Domain.unapply))
 
   private val DomainRegex = """^(?:[a-z0-9](?:[a-z0-9-]{0,61}[a-z0-9])?\.)+[a-z0-9]+$""".r
-  private val MaxLength = 128
+  private val MaxLength = 256
 
   def isValid(s: String): Boolean = DomainRegex.findFirstIn(s).isDefined && s.length <= MaxLength
+
+  def withHostname(hostname: String): Domain = {
```
ryanpbrewster: I don't think we should use the `.withX` style of naming when we're not modifying an existing model. I think `Domain.fromHostname` is clearer.

camhashemi: good note

kifi-backend/modules/common/app/com/keepit/classify/Domain.scala
```diff
@@ -37,15 +43,38 @@ object Domain {
...
     (__ \ 'state).format(State.format[Domain]) and
     (__ \ 'createdAt).format[DateTime] and
     (__ \ 'updatedAt).format[DateTime]
   )(Domain.apply, unlift(Domain.unapply))
 
   private val DomainRegex = """^(?:[a-z0-9](?:[a-z0-9-]{0,61}[a-z0-9])?\.)+[a-z0-9]+$""".r
-  private val MaxLength = 128
+  private val MaxLength = 256
 
   def isValid(s: String): Boolean = DomainRegex.findFirstIn(s).isDefined && s.length <= MaxLength
```
ryanpbrewster: Do you want to use `findFirstIn` here? Rather, do you care if the entire string matches the regex, or only matches a prefix?

kifi-backend/modules/shoebox/app/com/keepit/classify/DomainRepo.scala
```diff
@@ -20,16 +21,16 @@ trait DomainRepo extends Repo[Domain] {
...
     val db: DataBaseComponent,
     val clock: Clock,
-    domainCache: DomainCache) extends DbRepo[Domain] with DomainRepo {
+    domainHashCache: DomainHashCache) extends DbRepo[Domain] with DomainRepo {
   import db.Driver.simple._
 
   //todo(martin) remove this default implementation so we force repos to implement it
   override def invalidateCache(domain: Domain)(implicit session: RSession): Unit = {
-    domainCache.set(DomainKey(domain.hostname), domain)
+    domainHashCache.set(DomainHashKey(DomainHash.hashHostname(domain.hostname)), domain) // refactor this to call domain.hash directly once the table is back-filled
```
ryanpbrewster: I would prepend this comment with a `TODO(cam):` so you can find it more easily.

kifi-backend/modules/shoebox/app/com/keepit/classify/DomainRepo.scala
```diff
@@ -20,16 +21,16 @@ trait DomainRepo extends Repo[Domain] {
...
   }
 
   override def deleteCache(domain: Domain)(implicit session: RSession): Unit = {
-    domainCache.remove(DomainKey(domain.hostname))
+    domainHashCache.remove(DomainHashKey(DomainHash.hashHostname(domain.hostname)))
```
ryanpbrewster: I'd also put a `TODO(cam):` here. Too easy to forget otherwise.

kifi-backend/modules/common/conf/evolutions/shoebox/361.sql
```diff
@@ -0,0 +1,13 @@
+# SHOEBOX
+
+# --- !Ups
+
+ALTER TABLE domain ADD hash VARCHAR(26) DEFAULT NULL;
+
+CREATE UNIQUE INDEX domain_hash_index ON DOMAIN (hash);
```
ryanpbrewster: The naming convention for unique indices I've seen elsewhere is `<table>_u_<field1>_<field2>_...`. So I think this should be `domain_u_hash`.

leogrim: Good catch, let's be careful about existing conventions!

kifi-backend/modules/shoebox/app/com/keepit/admin/UrlController.scala
camhashemi, line 494: We don't use the accumulator, but the behavior of `FutureHelpers.map` is not what we want here. Correct me if I'm wrong, but `.map` fires all calls off at once. `.foldLeft` is going to wait until the first call finishes before calling the next, which is what we want for batching.

kifi-backend/modules/common/app/com/keepit/classify/Domain.scala
camhashemi: not sure where this regex came from and so I have to see what it's doing, will let you know

camhashemi: if you look at the regex, it has an `^` and `$`, so the two behaviors are identical. But `matches` makes more sense.

kifi-backend/modules/shoebox/app/com/keepit/admin/UrlController.scala
ryanpbrewster, line 494: Ah, okay. I think that `sequentialExec` is what you want, then. ((The implementation is exactly what you have here, actually, but I think it is clearer to use the named version. I didn't even realize that you needed them to be sequential.))

camhashemi, line 494: nice, thank you. completely agree.

kifi-backend/modules/shoebox/marketing
```diff
@@ -1 +1 @@
-Subproject commit 1b71438b1a478e2e4d7e78b93c6b5c3e9ea0de4f
+Subproject commit 1bb59b497b330d32816ccb3117a6097d5cc66dd1
```
camhashemi, line 1: this hash is up-to-date with master

camhashemi:
@ryanpbrewster @leogrim thanks for the review. final thumbs up?

kifi-backend/modules/common/app/com/keepit/classify/Domain.scala
```diff
@@ -37,15 +43,38 @@ object Domain {
...
 
   private val DomainRegex = """^(?:[a-z0-9](?:[a-z0-9-]{0,61}[a-z0-9])?\.)+[a-z0-9]+$""".r
-  private val MaxLength = 128
+  private val MaxLength = 256
+
+  def isValid(s: String): Boolean = DomainRegex.pattern.matcher(s).matches && s.length <= MaxLength
+
+  def fromHostname(hostname: String): Domain = {
+    val lowerCasedHostname = hostname.toLowerCase
+    Domain(hostname = lowerCasedHostname, hash = Some(DomainHash.hashHostname(lowerCasedHostname)))
```
leogrim, line 59: You could still use `withHostname` internally here, to DRY it up.

camhashemi, line 59: we refactored `withHostname` to `fromHostname`, so `withHostname` doesn't exist anymore.

leogrim, line 59: ok

ryanpbrewster:
lgtm

kifi-backend/modules/common/app/com/keepit/classify/Domain.scala
```diff
@@ -37,15 +43,38 @@ object Domain {
...
 
-  def isValid(s: String): Boolean = DomainRegex.findFirstIn(s).isDefined && s.length <= MaxLength
+case class DomainHash(hash: String) extends AnyVal {
+  override def toString: String = hash
+  def urlEncoded: String = hash.replaceAllLiterally("+" -> "-", "/" -> "_") // See RFC 3548 http://tools.ietf.org/html/rfc3548#page-6
+}
+
+object DomainHash {
+  def hashHostname(hostname: String): DomainHash = {
+    val binaryHash = MessageDigest.getInstance("MD5").digest(hostname)
```
andrewconner, line 70: MD5 is fairly slow and is a cryptographic hash. Why not a faster hash? MurmurHash is great (we use it elsewhere). (I see now that NormalizedURI uses MD5; probably a mistake then, too)

PR #16371, opened 2015-07-24
fix organization.site bug

camhashemi:
@ajkochanowicz should be fixed now, but good thing it failed because I forgot to add a test. thanks for your patience!


adamjgrant:
thx

PR #16372, opened 2015-07-24
Added api_readme.txt to project

camhashemi:
@andrewconner @leogrim @jaredpetker

### DIFF
--- a/kifi-backend/conf/api_readme.txt
+++ b/kifi-backend/conf/api_readme.txt
@@ -1,0 +1,139 @@
+***API Readme***
+
+Models:
+
+  ExternalId[User] and PublicId[Organization] are both strings that uniquely identify users/orgs respectively. EmailAddress is a string that must look like an email address.
+
+  OrganizationPermission is a string in the set
+    “view_organization"
+    "edit_organization"
+    "invite_members"
+    "modify_members"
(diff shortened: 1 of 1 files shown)


andrewconner:
`kifi-backend/conf` is actually a directory with meaning for Play. Perhaps if you want this create a new one, like `/docs`?

kifi-backend/conf/api_readme.txt
```diff
@@ -0,0 +1,141 @@
...
+    id: PublicId[Organization]
+    ownerId: ExternalId[User]
+    handle: String
+    name: String
+    description: Option[String]
+    avatarPath: Option[String]
+    numMembers: Int
+    numLibraries: Int
+
+  LibraryCardInfo
```
andrewconner, line 118: This, which is basically Scala code, will become out of date without someone or something in charge of maintaining it (like a compiler).

maxov:
I would also recommend using markdown with this, because it would look far nicer. I can convert it if you wish, but the syntax is not hard to learn

andrewconner:
If you want documentation about endpoints, why not just have a test for each endpoint that demonstrates how to use it, and what to expect back? That way it'll be verified on every deploy.

PR #16375, opened 2015-07-24
Refresh the org data when an org invite is accepted.

cvializ:
@andrewconner @ajkochanowicz  I think it's time to have a more centralized location for managing access to org data. Enter: the `orgProfileService`.

I'll be moving all the `net` calls into the service to keep all the knowledge in one place.


andrewconner:
:+1:

adamjgrant:
Glorius!

adamjgrant:
Actually, do you get a 'userProfileService' is not defined error from jslint? @cvializ 
in route.js

cvializ:
Oops my fingers slipped up, I'll fix that

PR #16376, opened 2015-07-24
Update to bootstrap 3.3

maxov:
This PR updates the bootstrap version in the admin to 3.3.5 (we were previously using 2.2.1)

I fixed as much weirdness as I could find, but there could still be some left.


andrewconner:
Ah, we need to stay hip.

maxov:
@eishay look good? I can fix whatever people find as broken in the updated admin

PR #16383, opened 2015-07-25
Remove / deprecate routes

leogrim:
@andrewconner @camhashemi


kifi-backend/modules/shoebox/conf/shoeboxService.routes
```diff
@@ -390,8 +390,8 @@ POST    /site/collections/:id/removeKeeps @com.keepit.controllers.website.KeepsC
 POST    /site/collections/:id/addKeeps @com.keepit.controllers.website.KeepsController.keepToCollection(id: ExternalId[Collection])
 GET     /site/collections/search    @com.keepit.controllers.website.KeepsController.searchUserTags(query: String, limit: Option[Int] = None)
 
+#### Deprecated: use /site/user-or-org/:handle/libraries/:slug
 GET     /site/users/:userStr/libraries/:slug    @com.keepit.controllers.website.LibraryController.getLibraryByPath(userStr:String, slug: String, showPublishedLibraries: Boolean ?= false, is: Option[String] ?= None, authToken: Option[String] ?= None)
```
maxov, line 394: Perhaps the action also deserves a `@deprecated` annotation?

leogrim, line 394: Annotations trigger warnings, they're a bit annoying. This should go away
soon anyway!

PR #16384, opened 2015-07-25
Only show the org invite banner if isInvited is true.

cvializ:
@ajkochanowicz @andrewconner 

This should be the logic regardless of whether authToken is there, correct?


andrewconner:
Yeah looks right. `authToken` may not always be there

adamjgrant:
I was kind of thinking the point of the checklist here https://github.com/kifi/keepit/pull/16363 would be that the changes associated with them would appear in the PR

cvializ:
Oh I see that now, I misunderstood. I thought the checklist was there as a signpost, but that branches and fixes could be anywhere. I can do future changes in that PR

PR #16385, opened 2015-07-25
Remove the organizationInfo wrapper object

cvializ:
@ajkochanowicz is this PR-into-branch flow more what you had in mind?

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/orgProfile/orgProfileCtrl.js
+++ b/kifi-backend/modules/shoebox/angular/src/orgProfile/orgProfileCtrl.js
@@ -5,7 +5,7 @@ angular.module('kifi')
 .controller('OrgProfileCtrl', [
   '$window', '$scope', 'profile',
   function ($window, $scope, profile) {
-    $window.document.title = profile.organizationInfo.name + ' • Kifi';
-    $scope.profile = _.cloneDeep(profile.organizationInfo);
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/shoebox/angular/src/orgProfile/orgProfileCtrl.js
```diff
@@ -5,7 +5,7 @@ angular.module('kifi')
 .controller('OrgProfileCtrl', [
   '$window', '$scope', 'profile',
   function ($window, $scope, profile) {
-    $window.document.title = profile.organizationInfo.name + ' • Kifi';
-    $scope.profile = _.cloneDeep(profile.organizationInfo);
+    $window.document.title = profile.name + ' • Kifi';
+    $scope.profile = _.cloneDeep(profile);
```
andrewconner, line 9: I know you didn't add this, but why clone it? Curious, not critique; for protections against mutations?

cvializ, line 9: You know I was wondering that myself. I think it's necessary/safer when you're manually cacheing. Like with `profileService.me`. If I change the value of `profileService.me` for my one view, it will change for all of them even if it was unexpected. At least that's my intuition

adamjgrant:
:+1:

PR #16386, opened 2015-07-27
Amplitude frontend/angular and backend updates

joshm1:
@andrewconner


kifi-backend/modules/heimdal/app/com/keepit/heimdal/AmplitudeClient.scala
```diff
@@ -16,6 +16,17 @@ object AmplitudeClient {
     "daysSinceUserJoined", "device", "experiments", "extensionVersion", "kcid_6", "kcid_7", "kcid_8",
     "kcid_9", "kcid_10", "kcid_11", "os", "osVersion", "remoteAddress", "serviceInstance", "serviceZone", "userCreatedAt",
     "userId", "userSegment")
+
+  val killedEvents = Set("user_old_slider_sliderShown", "user_expanded_keeper", "user_used_kifi", "user_reco_action",
```
joshm1, line 20: i'd prefer these events were killed at the source and make sure they don't get sent to heimdal, but I'm not sure if the spec was to remove these events only for amplitude or remove them from mixpanel as well. i asked Jen to clarify. she also wants some existing events to be combined/renamed, which I haven't done yet

andrewconner:
Oh whoops, I have been fixing some bugs with tracking. See: https://github.com/kifi/keepit/compare/master...andrewconner:blacklist_images

Not done yet because was trying to get an idea of what Jen actually needed from our analytics. What we're currently doing is _really_ weird.

Yours can go first and then I'll try to resume.

kifi-backend/modules/shoebox/angular/src/common/services/mixpanelService.js
```diff
@@ -70,11 +80,25 @@
...
       } else {
-        mixpanel.track(action, props);
+        trackEventDirectly(action, props);
       }
     }
   }
 
+  function setUser(user) {
+    if ($window.mixpanel) {
+      $window.mixpanel.identify(user);
```
andrewconner, line 90: This is a bug that I fixed in my PR. The user object should never be the identity.

kifi-backend/modules/shoebox/angular/src/common/services/mixpanelService.js
```diff
@@ -70,11 +80,25 @@
...
+    }
+
+    // sanity check shouldn't be necessary, but it's to make sure user.id is
+    // valid and doesn't accidentally get set to a value that's not unique
+    // across users
+    if ($window.amplitude && user && user.id && uuidRegex.test(user.id)) {
+      // does not call amplitude.setUserId because this id is internal only and will not be known here
```
andrewconner, line 97: Is this expected from Jen's end? Why not unify on one ID?

kifi-backend/modules/shoebox/angular/src/common/services/mixpanelService.js
```diff
@@ -113,9 +146,7 @@
 
         trackEventThroughProxy('user_viewed_page', attributes);
       } finally {
-        if (!oldId) {
-          mixpanel.identify(oldId);
-        }
+        if (oldId) { setUser(oldId); }
```
andrewconner, line 149: Does this break expectations since it only does it for mixpanel?

kifi-backend/modules/shoebox/angular/src/common/services/mixpanelService.js
joshm1, line 97: you mean why not use external id for the amplitude user id instead of what we set the "distinct_id" now in mixpanel, which would be something like "user_2" for users, or the IP address for non users, or something really generic if neither of those is found.  currently user IDs in amplitude are just a copy of the mixpanel distinct_id. using a different ID is definitely possible and wouldn't cause problems, it's just not something we discussed. I can bring it up with Jen

kifi-backend/modules/shoebox/angular/src/common/services/mixpanelService.js
joshm1, line 149: I don't think so, or I don't know what you're asking.  I tried to implement the same type of action for amplitude anywhere it was done for mixpanel.  looking at your https://github.com/kifi/keepit/compare/master...andrewconner:blacklist_images#diff-cc148c7745bbc5fed4e75f558c988fc8R122 i see you saw the same bug that I did  (`if (!oldId)`) but your documentation is much better and you seem to understand it more.  The `oldId` (user.id) is set to amplitude's device_id, which is used to link a "user" to a former "non-user" or to a user when we don't know what the userId is (that's likely the case here).  I was certainly a bit confused by this, your explanation in your branch helps me understand what this was for.  do you advise we only set this for

PR #16388, opened 2015-07-27
We changed username to handle in router, so change it in the library controller too.

cvializ:
Should resolve https://fortytwo.airbrake.io/projects/113130/groups/1472022958550462273/notices/1473957619304407660


cvializ:
@andrewconner @ajkochanowicz

PR #16390, opened 2015-07-27
Why do we do this?

andrewconner:
@stkem @eishay @leogrim Removing to help lighten our logs pressure.


stkem:
I think this was for when we airbrake during play shutdown (when the logger might no longer work).

PR #16370, opened 2015-07-24
org api change

camhashemi:
@leo @cvializ @ajkochanowicz @ryanpbrewster this removes the wrapper around `organizationInfo` and `membershipInfo`, now you just get an object with `organization` and `membership` (which were previously suffixed by "info").

PR #16393, opened 2015-07-27
Revert "org api change"

camhashemi:
Reverts kifi/keepit#16370


camhashemi:
reverting until frontend is ready to consume the changes. my fault for not making sure it was okay to merge.

PR #16397, opened 2015-07-27
Added an endpoint for uploading an org avatar

ryanpbrewster:
Also removed a deprecated use of Play implicit execution context

@leogrim @camhashemi @cvializ @ajkochanowicz

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationAvatarCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationAvatarCommander.scala
@@ -148,6 +148,8 @@ class OrganizationAvatarCommanderImpl @Inject() (
 }
 
 object OrganizationAvatarConfiguration {
+  val defaultSize = ScaledImageSize.Medium.idealSize
+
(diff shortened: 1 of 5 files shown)


adamjgrant:
Glorious!

cvializ:
Beautimous. @ryanpbrewster, does this take the same arguments as the user profile image upload?

ryanpbrewster:
@cvializ, I don't think either of those take any arguments (I may be wrong). It needs a request body that is a temporary file, and it returns a `{"uploaded": <path_to_image>}`.

cvializ:
Ok! I'm going to try to reuse the userProfile upload dialogs, so as long as the org and user avatar uploads expect the same request structure we're good to go

andrewconner:
Yeah if my memory serves right, should be the same. You ajax up a binary image, and get back a URL. You're responsible for error handling and some pleasant transition so the user knows that the upload is happening. Errors vary and probably will give you more details than you care; you can probably just simplify and say an error has occurred and they can retry or contact support.

cvializ:
@ryanpbrewster if this is good to merge I'm ready for it :+1:

PR #16398, opened 2015-07-27
Take out LDA, URI-Domain edges from Graph

leogrim:
@eishay @andrewconner as a start

### DIFF
--- a/kifi-backend/modules/graph/app/com/keepit/graph/manager/GraphUpdateFetcher.scala
+++ b/kifi-backend/modules/graph/app/com/keepit/graph/manager/GraphUpdateFetcher.scala
@@ -67,19 +67,7 @@ class GraphUpdateFetcherImpl @Inject() (
         }
       }
 
-      case NormalizedUriGraphUpdate => {
-        shoebox.getIndexableUris(seq.copy(), fetchSize).flatMap { indexableUris =>
-
-          shoebox.internDomainsByDomainNames(indexableUris.map { _.getDomainName }.toSet).imap { domainInfoByName: Map[String, DomainInfo] =>
-            indexableUris.map { uri =>
-              domainInfoByName.get(uri.getDomainName) match {
-                case Some(domain: DomainInfo) if !domain.isEmailProvider => NormalizedUriGraphUpdate.apply(uri, domain.id)
-                case _ => NormalizedUriGraphUpdate(uri, None)
-              }
-            }
-          }
-        }
-      }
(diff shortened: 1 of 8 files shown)


kifi-backend/modules/graph/app/com/keepit/graph/manager/GraphUpdater.scala
```diff
@@ -245,13 +232,13 @@ class GraphUpdaterImpl @Inject() () extends GraphUpdater with Logging {
...
-      writer.removeEdgeIfExists(update.orgId, update.userId, TimestampEdgeReader)
+      writer.removeEdgeIfExists(update.userId, update.orgId, EmptyEdgeReader)
+      writer.removeEdgeIfExists(update.orgId, update.userId, EmptyEdgeReader)
     case OrganizationMembershipStates.ACTIVE =>
       writer.saveVertex(OrganizationData(update.orgId))
       writer.saveVertex(UserData(update.userId))
-      writer.saveEdge(update.userId, update.orgId, TimestampEdgeData(update.createdAt.getMillis))
-      writer.saveEdge(update.orgId, update.userId, TimestampEdgeData(update.createdAt.getMillis))
+      writer.saveEdge(update.userId, update.orgId, EmptyEdgeData)
+      writer.saveEdge(update.orgId, update.userId, EmptyEdgeData)
```
leogrim, line 241: @camhashemi organization memberships shouldn't decay with time

kifi-backend/modules/graph/app/com/keepit/graph/manager/GraphUpdateFetcher.scala
```diff
@@ -67,19 +67,7 @@ class GraphUpdateFetcherImpl @Inject() (
...
-            indexableUris.map { uri =>
-              domainInfoByName.get(uri.getDomainName) match {
-                case Some(domain: DomainInfo) if !domain.isEmailProvider => NormalizedUriGraphUpdate.apply(uri, domain.id)
-                case _ => NormalizedUriGraphUpdate(uri, None)
-              }
-            }
-          }
-        }
-      }
+      case NormalizedUriGraphUpdate => shoebox.getIndexableUris(seq.copy(), fetchSize).imap(_.map(NormalizedUriGraphUpdate.apply))
```
leogrim, line 70: @camhashemi I'm taking this out for now, to alleviate the load on the graph a bit. When we bring it back, we should probably grab the domain ids on Shoebox since we're fetching the uris from there anyway (unlike email addresses that are fetched from ABook).

camhashemi, line 70: Got it.

PR #16399, opened 2015-07-27
Correcting response parsing for org id

adamjgrant:
https://fortytwo.airbrake.io/projects/113130/groups/1474044278289113812/notices/1474044490294403807


adamjgrant:
Fixes issue where libraries would not redirect from org page @ac @camhashemi @cvializ @leogrim

PR #16403, opened 2015-07-27
Added a `/:handle/members` route

ryanpbrewster:
Only routes if `:handle` corresponds to an org

@leogrim @andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/KifiSiteRouter.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/KifiSiteRouter.scala
@@ -90,7 +90,7 @@ class KifiSiteRouter @Inject() (
     } getOrElse notFound(request)
   }
 
-  def serveWebAppIfOrganizationFound(handle: Handle) = WebAppPage { implicit request =>
+  def serveWebAppIfOrganizationFound(handle: OrganizationHandle) = WebAppPage { implicit request =>
     lookupHandle(handle) map {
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/shoebox/app/com/keepit/controllers/website/KifiSiteRouter.scala
```diff
@@ -90,6 +90,28 @@ class KifiSiteRouter @Inject() (
     } getOrElse notFound(request)
   }
 
+  def serveWebAppIfOrganizationFound(handle: Handle) = WebAppPage { implicit request =>
```
leogrim: Should be `OrganizationHandle` here then?

ryanpbrewster: I guess so. Does that actually change the logic at all? We're still going to have to do an implicit conversion back to a `Handle` in order to do a lookup by handle.

ryanpbrewster: My resistance to this is just laziness. We don't have a `PathBindable` implemented for `OrganizationHandle`.

leogrim: Easy to add, at this point we know the type we're expecting. I talked with @andrewconner and we should remove the implicit conversions anyway, as they will make refactoring a pain (my mistake).

Issue #16389, opened 2015-07-27
Orgs MVP


adamjgrant:
**Not using this anymore** Asana ftw
- [x] Create a library should work with org libraries. https://github.com/kifi/keepit/pull/16405
- [ ] [Adam] Modify org libraries. https://github.com/kifi/keepit/pull/16422
- [x] Delete org libraries.
- [x] `isInvited` should be the only thing that determines the "invited" banner showing.
- [x] When accepting an invite, show newly permissioned page.
- [x] Notifications icon floats in middle of header.
- [x] Update code to reflect [endpoint name change](https://github.com/kifi/keepit/pull/16370)
- [ ] Users can move their libraries from their personal space to the org space
- [ ] Users can click somewhere to access their org space (design conversation needed).

Lower priority
- [ ] [Carlos] Upload org image
- [x] Uninvited members of an org should not be able to (appear to) edit title, description. https://github.com/kifi/keepit/pull/16411
- [x] Validations (Org url and successful saving) https://github.com/kifi/keepit/pull/16415
- [x] Superadmin needs to be able to demote members. #16407 
- [ ] Page should redirect to /:handle after deleting a keep

adamjgrant:
btw, the "Maybe' is in reference to issues we're having where jumbled-named libraries appear after deleting a library. I'm told this is an erroneous result of sanitization post deletion and it's being worked on. Afterwards, I can verfiy whether or not modification and deletion is working correctly.

ryanpbrewster:
The jumbled names should be dealt with now. As soon as we deploy to a recent build they should disappear.

adamjgrant:
Yep, sounds like you guys have bigger fish to fry. Godspeed.

adamjgrant:
I have enough of this in Asana now that I'm going to call this issue deprecated. (Makes the sign of the cross)

attn @cvializ @andrewconner @leogrim @ryanpbrewster @camhashemi

andrewconner:
Deprecate away!

andrewconner:
@adamjgrant Closeable?

adamjgrant:
Yup.

PR #16404, opened 2015-07-27
Fix notification icon floats in middle of header.

cvializ:
@ajkochanowicz @andrewconner I accidentally introduced this bug a little while ago with a5e3711. This ought to fix it


cvializ:
Resolves one of #16389

andrewconner:
:+1:

PR #16409, opened 2015-07-27
Organization libraries bug-fix

ryanpbrewster:
Previously it was not checking library state to see if it was ACTIVE


ryanpbrewster:
@ajkochanowicz @andrewconner

adamjgrant:
Terrific

PR #16412, opened 2015-07-27
various fixes (domain and org recos)

camhashemi:
@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/classify/DomainRepo.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/classify/DomainRepo.scala
@@ -80,7 +80,7 @@ class DomainRepoImpl @Inject() (
 
     val existingHostnames = existingDomains.map(_.hostname)
 
-    val toBeInserted = (lowerCasedHostnamesToIntern -- existingHostnames).map(Domain.fromHostname)
+    val toBeInserted = (lowerCasedHostnamesToIntern -- existingHostnames).filter(Domain.isValid).map(Domain.fromHostname)
 
     val existingDomainByName = existingDomains.map { domain =>
       domain.state match {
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/UserStatisticsCommander.scala
```diff
@@ -165,7 +165,7 @@ class UserStatisticsCommander @Inject() (
     Future.sequence(membersStatsFut).imap(_.toMap)
   }
 
-  def organizationStatistics(orgId: Id[Organization], adminId: Id[User], numMemberRecos: Int = 30): Future[OrganizationStatistics] = {
+  def organizationStatistics(orgId: Id[Organization], adminId: Id[User], numMemberRecos: Int = 60): Future[OrganizationStatistics] = {
```
andrewconner, line 168: This doesn't change what you're expecting, since AdminOrganizationController doesn't use the default argument. In general, don't use default arguments when it's a real parameter you intent (or should always) pass in.

kifi-backend/modules/shoebox/app/com/keepit/classify/DomainRepo.scala
```diff
@@ -80,7 +80,7 @@ class DomainRepoImpl @Inject() (
 
     val existingHostnames = existingDomains.map(_.hostname)
 
-    val toBeInserted = (lowerCasedHostnamesToIntern -- existingHostnames).map(Domain.fromHostname)
+    val toBeInserted = (lowerCasedHostnamesToIntern -- existingHostnames).filter(Domain.isValid).map(Domain.fromHostname)
```
andrewconner, line 83: Can you filter out before the Repo call?

PR #16415, opened 2015-07-27
Org Header Validations

adamjgrant:
https://github.com/kifi/keepit/issues/16389


adamjgrant:
Right now validations are sufficiently dumb. You'll only see a generic red error message for blank fields or connection issues and whatever you type in on the frontend will remain there. Unless there is any objection, I think it would be wise to call this good and minimal for now and move on to meatier stuff like getting modifications to work. @andrewconner @cvializ @ryanpbrewster

andrewconner:
@ajkochanowicz :+1:

andrewconner:
I like the phrase "sufficiently dumb"

PR #16416, opened 2015-07-28
Isolated the mediocre parts of the site router

ryanpbrewster:
Will be much easier to clean up when org experiments are deprecated

@andrewconner 

I'm pretty sure this can be cleaned up further, but I tried to push all the unpleasantness into a very small number of isolated functions.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/KifiSiteRouter.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/KifiSiteRouter.scala
@@ -1,20 +1,17 @@
 package com.keepit.controllers.website
 
-import com.keepit.common.cache.TransactionalCaching.Implicits._
-import com.google.inject.{ Provider, Inject, Singleton }
+import com.google.inject.{ Inject, Singleton }
 import com.keepit.commanders._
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/app/com/keepit/controllers/website/KifiSiteRouter.scala
```diff
@@ -192,32 +165,63 @@ class KifiSiteRouter @Inject() (
...
 
-  private def lookupHandle(handle: Handle): Option[(Either[Organization, User], Option[Int])] = {
+  private def lookupUser(handle: Handle) = {
+    // TODO(ryan)[ORG-EXPERIMENT]: when orgs are live, kill the below code and use this commented code
+    /*
+    lookupByHandle(handle) flatMap {
+      case (Left(org), _) => None
+      case (Right(user), redirectStatusOpt) => Some((user, redirectStatusOpt))
+    }
+    */
```
ryanpbrewster, line 175: This is unpleasant because I can't rely on having a `MaybeUserRequest[_]` around like I can for `lookupOrganization`. See `MobileAppFilter`.

andrewconner:
Works for me

PR #16417, opened 2015-07-28
add a try-catch to prevent admin from going down with graph

camhashemi:
@andrewconner @leogrim not sure what you guys think of this, but it's something that @gratimax and Tamila have asked for since graph has been so flaky. this way we report and just return an empty list.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/admin/AdminUserController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/admin/AdminUserController.scala
@@ -39,8 +39,8 @@ import play.api.mvc.{ Action, AnyContent, Result }
 
 import views.html
 import com.keepit.typeahead.{ KifiUserTypeahead, TypeaheadHit, SocialUserTypeahead }
-import com.keepit.common.healthcheck.SystemAdminMailSender
-import com.keepit.abook.model.RichContact
+import com.keepit.common.healthcheck.{ AirbrakeNotifier, SystemAdminMailSender }
+import com.keepit.abook.model.{ OrganizationUserMayKnow, RichContact }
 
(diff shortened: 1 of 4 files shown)


andrewconner:
Down with me, but @leogrim may know better

maxov:
This is great :+1: I also proposed returning an `Either[RecoError, List[Reco]]` for some recommended type `Reco` (like `User` or `Organization`) and a series of types representing errors `RecoError`. This way we could see in the admin that the graph query failed with some sort of error instead of a plain list. An empty list might imply that the graph simply couldn't find and recommendations, while an error message visibly says that there was something wrong with fetching the data.

camhashemi:
ah, @gratimax. that's better.

leogrim:
I'm down, just like the Graph.

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminUserController.scala
```diff
@@ -254,7 +255,11 @@ class AdminUserController @Inject() (
     val userId = user.id.get
     val abookInfoF = abookClient.getABookInfos(userId)
     val econtactCountF = abookClient.getEContactCount(userId)
-    val fOrgRecos = abookClient.getOrganizationRecommendationsForUser(user.id.get, offset = 0, limit = 5)
+    val fOrgRecos = try {
+      abookClient.getOrganizationRecommendationsForUser(user.id.get, offset = 0, limit = 5)
```
maxov, line 259: I just realized, there's a method `fallbackTo` on Future that accomplishes this

camhashemi, line 259: We could use that, but we still want to call `airbrake.notify(ex)`.

camhashemi:
@gratimax I'm going to merge this to get it out there, and we can add error messages later.

PR #16418, opened 2015-07-28
Deploy new graph version

leogrim:
@andrewconner @eishay @camhashemi


camhashemi:
do you have to do this whenever you remove an Vertex/Edge type?

leogrim:
If you're introducing a new type of `GraphUpdate` you're good, but if you're removing or modifying one, then you want to rebuild the graph from scratch for the previous updates to be dropped / reingested. 

We could probably make it possible to reset a single `GraphUpdate` sequence number but we would have to make sure the graph indeed ends up in the same state as if starting entirely from scratch.

PR #16419, opened 2015-07-28
increase number of org-pymn

camhashemi:
@eishay this'll help some. @andrewconner removed the default.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/admin/AdminOrganizationController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/admin/AdminOrganizationController.scala
@@ -108,7 +108,7 @@ class AdminOrganizationController @Inject() (
   }
 
   def organizationViewById(orgId: Id[Organization]) = AdminUserPage.async { implicit request =>
-    val numMemberRecommendations = request.queryString.get("numMemberRecos").map(_.head.toInt).getOrElse(30)
+    val numMemberRecommendations = request.queryString.get("numMemberRecos").map(_.head.toInt).getOrElse(60)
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/app/com/keepit/admin/AdminOrganizationController.scala
```diff
@@ -108,7 +108,7 @@ class AdminOrganizationController @Inject() (
   }
 
   def organizationViewById(orgId: Id[Organization]) = AdminUserPage.async { implicit request =>
-    val numMemberRecommendations = request.queryString.get("numMemberRecos").map(_.head.toInt).getOrElse(30)
+    val numMemberRecommendations = request.queryString.get("numMemberRecos").map(_.head.toInt).getOrElse(60)
```
andrewconner, line 111: You're probably fine for this, but for future Play usage:
1. Passing in parameters to the controller method is better than just checking queryString, because no one looking at the router will be able to tell you use this.
2. `request.queryString.get("numMemberRecos").flapMap(_.headOption.map(_.toInt)).getOrElse(60)`

camhashemi, line 111: will do, thank you

camhashemi, line 111: In the case of (1), could I then use a default parameter `numMemberRecos: Int = 60` instead of the statement `Option[Int].getOrElse(60)`, where both would implement the same behavior? Not sure which makes more sense.

PR #16420, opened 2015-07-28
adding a private "read it later" library to half of the new users who…

eishay:
… would join, adding an experiment for tracking.

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/model/Library.scala
+++ b/kifi-backend/modules/common/app/com/keepit/model/Library.scala
@@ -302,6 +302,7 @@ object LibraryKind {
   case object SYSTEM_MAIN extends LibraryKind("system_main", 0)
   case object SYSTEM_SECRET extends LibraryKind("system_secret", 1)
   case object SYSTEM_PERSONA extends LibraryKind("system_persona", 2)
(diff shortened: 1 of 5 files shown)


eishay:
@andrewconner @leogrim

kifi-backend/modules/common/app/com/keepit/model/Library.scala
```diff
@@ -302,6 +302,7 @@ object LibraryKind {
   case object SYSTEM_MAIN extends LibraryKind("system_main", 0)
   case object SYSTEM_SECRET extends LibraryKind("system_secret", 1)
   case object SYSTEM_PERSONA extends LibraryKind("system_persona", 2)
+  case object SYSTEM_READ_IT_LATER extends LibraryKind("system_read_id_later", 1)
```
andrewconner: It'll come before user created libraries?

eishay: Yes, what do you think?

andrewconner: Probably just let it be mixed in, like persona libraries. They'll see it before it gets buried, and doesn't get annoying.

eishay:
ok

PR #16422, opened 2015-07-28
Showing org slug in library modification

adamjgrant:
https://github.com/kifi/keepit/issues/16389

A key change I'm making here is brining in the whole `profile` into the library controller so we can read all the twiddly bits, not just `profile.type`

Depends on https://github.com/kifi/keepit/pull/16425


adamjgrant:
@cam will merge this into [his PR](https://github.com/kifi/keepit/pull/16425#discussion_r35675701) for the new `getLibraryByHandleAndSlug` endpoint.

PR #16425, opened 2015-07-28
replace getLibraryByPath with getLibraryByHandleAndSlug

camhashemi:
@leogrim @andrewconner @ryanpbrewster. ~~this shouldn't affect clients~~. @ajkochanowicz  @cvializ


kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -1459,21 +1460,21 @@ class LibraryCommanderImpl @Inject() (
...
-            Right(library)
-        }
+  def getLibraryWithHandleAndSlug(handle: Handle, slug: LibrarySlug, viewerId: Option[Id[User]])(implicit context: HeimdalContext): Either[LibraryFail, Library] = {
+    val ownerOpt = db.readOnlyReplica { implicit session => handleCommander.getByHandle(handle) }
+
+    val librarySpaceOpt = ownerOpt match {
+      case None => None // "invalid_handle"
+      case Some((Left(org), _)) => Some(LibrarySpace.fromOrganizationId(org.id.get))
+      case Some((Right(user), _)) => Some(LibrarySpace.fromUserId(user.id.get))
     }
```
ryanpbrewster: `map` is maybe cleaner here.

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileLibraryController.scala
```diff
@@ -138,16 +138,16 @@ class MobileLibraryController @Inject() (
...
 
-  def getLibraryByPathV1(userStr: String, slugStr: String, imageSize: Option[String] = None) = MaybeUserAction.async { request =>
-    getLibraryByPath(request, userStr, slugStr, imageSize, true)
+  def getLibraryByPathV1(handle: Handle, slug: LibrarySlug, imageSize: Option[String] = None) = MaybeUserAction.async { request =>
+    getLibraryByPath(request, handle, slug, imageSize, true)
   }
-  def getLibraryByPathV2(userStr: String, slugStr: String, imageSize: Option[String] = None) = MaybeUserAction.async { request =>
-    getLibraryByPath(request, userStr, slugStr, imageSize, false)
+  def getLibraryByPathV2(handle: Handle, slug: LibrarySlug, imageSize: Option[String] = None) = MaybeUserAction.async { request =>
+    getLibraryByPath(request, handle, slug, imageSize, false)
```
ryanpbrewster, line 145: Oh no, we used to pass around strings? Good change. Thanks for adding the PathBindable.

kifi-backend/modules/shoebox/conf/shoeboxService.routes
```diff
@@ -530,6 +530,8 @@ GET           /site/organizations/:id/libraries                 @com.keepit.cont
 
 
 GET           /site/user-or-org/:handle                   @com.keepit.controllers.website.UserOrOrganizationController.getByHandle(handle: Handle)
+GET           /site/user-or-org/:handle/libraries         @com.keepit.controllers.website.UserOrOrganizationController.getLibrariesByHandle(handle: Handle, page: Int ?= 0, size: Int ?= 12, filter: String ?= "own")
+GET           /site/user-or-org/:handle/libraries/:slug   @com.keepit.controllers.website.LibraryController.getLibraryByHandleAndSlug(handle: Handle, slug: LibrarySlug, authToken: Option[String] ?= None)
```
ryanpbrewster, line 534: I thought we killed these routes for a reason. @leogrim want to weigh in?

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
camhashemi: nice

kifi-backend/modules/shoebox/conf/shoeboxService.routes
leogrim, line 534: @andrewconner want to weigh in?

kifi-backend/modules/shoebox/test/com/keepit/controllers/website/LibraryControllerTest.scala
```diff
@@ -593,7 +599,101 @@ class LibraryControllerTest extends Specification with ShoeboxTestInjector {
...
+                "subscribed":false
+               },
+               "invite": null,
+               "path": "${LibraryPathHelper.formatLibraryPath(basicUser1, Some(org1), lib1.slug)}"
+             },
+             "subscriptions": [],
+             "suggestedSearches": {"terms": [], "weights": []}
+            }
+          """)
+        Json.parse(contentAsString(result1)) must equalTo(expected)
```
ryanpbrewster, line 695: I dislike these kind of tests. They are incredibly fragile, and they aren't focused. If something goes wrong, it's extremely hard to find out what and fix it.

I prefer explicitly testing specific content. So you can test for `contentAsJson(result1) === Json.toJson(libraryCommander.getLibraryCardInfo(lib))` or something, or you can get even more specific (test for the correct id, path, etc.).

kifi-backend/modules/shoebox/conf/shoeboxService.routes
andrewconner, line 534: I don't see a reason we need this, but @ajkochanowicz want to weigh in?

adamjgrant, line 534: Yep, don't need them. @cvializ want to weigh in?

cvializ, line 534: Hmm I think I need them, but maybe @camhashemi can weigh in?

adamjgrant, line 534: If I may weigh in, in one way leaving them _way_ in might weigh it down. @WayneWayan, could you weigh in?

andrewconner, line 534: Based on some weighing from Ryan and Carlos, we decided we need `/site/user-or-org/:handle/libraries/:slug` but don't need `/site/user-or-org/:handle/libraries`.

kifi-backend/modules/shoebox/test/com/keepit/controllers/website/LibraryControllerTest.scala
camhashemi, line 695: these tests are what we talked about in lieu of API documentation. I've used these tests to understand why an API isn't giving me back what I want on the front-end.

ryanpbrewster, line 695: I'm fine with it being in one place (probably whatever method gets called when `/libraries/:id` is hit). But suppose we want to add a field to `LibraryCardInfo` (or whatever is being serialized here). How many tests are going to need to be modified to get that simple change to pass?

ryanpbrewster, line 695: Oh, maybe this _is_ the one place it ought to be tested explicitly?

kifi-backend/modules/shoebox/test/com/keepit/controllers/website/LibraryControllerTest.scala
```diff
@@ -593,7 +599,101 @@ class LibraryControllerTest extends Specification with ShoeboxTestInjector {
...
+        val libraryController = inject[LibraryController]
+
+        val (user1: User, org1: Organization, lib1: Library) = db.readWrite { implicit s =>
+          val user = {
+            val saved = UserFactory.user().withName("Cam", "Hashemi").withCreatedAt(t1).saved
+            handleCommander.setUsername(saved, Username("camhash")).get
+          }
+          val org = {
+            val saved = OrganizationFactory.organization().withName("CamCo").withOwner(user).saved
+            handleCommander.setOrganizationHandle(saved, OrganizationHandle("camco")).get
```
ryanpbrewster: FYI, you can say `OrganizationFactory.organization().withHandle(OrganizationHandle("camco")).saved` and it will do handle the handle.

kifi-backend/modules/shoebox/test/com/keepit/controllers/website/LibraryControllerTest.scala
```diff
@@ -536,18 +536,24 @@ class LibraryControllerTest extends Specification with ShoeboxTestInjector {
...
         status(result1_bad) must equalTo(BAD_REQUEST)
 
-        val testPath2 = com.keepit.controllers.website.routes.LibraryController.getLibraryByPath(extInput, slugInput).url
+        val testPath2 = com.keepit.controllers.website.routes.LibraryController.getLibraryByHandleAndSlug(Handle(extInput), slug).url
         val request2 = FakeRequest("GET", testPath2)
-        val result2 = libraryController.getLibraryByPath(extInput, slugInput, false)(request2)
+        val result2 = libraryController.getLibraryByHandleAndSlug(Handle(extInput), slug)(request2)
         val lib1Updated = db.readOnlyMaster { libraryRepo.get(lib1.id.get)(_) }
-        status(result2) must equalTo(OK)
+        status(result2) must equalTo(400) // sending external ids to this endpoint are deprecated
```
ryanpbrewster: `BAD_REQUEST` instead of `400`

kifi-backend/modules/shoebox/test/com/keepit/controllers/website/LibraryControllerTest.scala
camhashemi, line 695: We should try to make sure our tests aren't repetitive, but the point is that you validate the output all the way to seeing it in JSON, which helps clients or people who work on both front and back-end (like the person who wrote these tests before me)

kifi-backend/modules/shoebox/test/com/keepit/controllers/website/LibraryControllerTest.scala
camhashemi: sweet, thanks

kifi-backend/modules/shoebox/test/com/keepit/controllers/website/LibraryControllerTest.scala
ryanpbrewster, line 695: It's easier for us and easier for front-end/mobile if they only have to figure out how to parse results once. If this is exactly the same format as another endpoint (say `/libraries/:id`, i.e. `LibraryController.getLibraryById`) then I would test to see that they're equivalent.

This way a front-end/mobile engineer can reuse their own code instead of having to either
- manually check to see that the output is exactly the same as some other endpoint (and hope that it's by design and not coincidence)
- write entirely new parsing code

camhashemi, line 695: There is a difference between that endpoint and this one (I'm addressing that problem right now), and even if there wasn't, we build separate tests for each endpoint. I understand what you're getting at, but we haven't structured things that way. If one day we have time to go in restructure our tests, I'm happy to.

Seems like a lot of these were written by full-stack people working to make their JS lives less error-prone, so they write out expected outputs from a JS perspective. They don't have parsing code, the objects are either there or not.

kifi-backend/modules/shoebox/conf/shoeboxService.routes
camhashemi, line 534: Doesn't look like we're using `/site/user-or-org/:handle/libraries` anyways

kifi-backend/modules/shoebox/test/com/keepit/controllers/website/LibraryControllerTest.scala
ryanpbrewster, line 695: Mobile definitely has parsing code.

I don't understand the pushback on this. The test is simpler to write, simpler to read, and less fragile. It will take ~0.1s longer to run, but that seems like an easy tradeoff to me.

If there is an intentional difference then that's a different story.

camhashemi, line 695: Let's sync up in person. I agree that it's simpler to write, I don't think it's simpler to read for clients. It takes longer to refactor, but it's a test that documents the endpoint's output. Removing that for the sake of refactoring isn't a good thing to do imo.

camhashemi:
Per @andrewconner and @ryanpbrewster 's talks, killing `getLibraryByPublicId` might not be a good idea, as the clients only key to get a library would be a `handle` and `slug` combo that can change at any time, it also suffers a performance drop.

Either way, this PR consolidates getting libraries on the backend and removes `showPublishedLibraries`, which doesn't produce desired output. Would like to get this out there.

kifi-backend/modules/curator/app/com/keepit/curator/commanders/AttributionHelper.scala
```diff
@@ -50,7 +50,7 @@ class SeedAttributionHelper @Inject() (
           Future.successful(ret)
         } else {
           val uriIds = uriId2Idx.keys.toSeq
-          search.augment(Some(userId), false, maxKeepersShown = 20, maxLibrariesShown = 15, maxTagsShown = 0, items = uriIds.map(AugmentableItem(_))).map { infos =>
+          search.augment(Some(userId), showPublishedLibraries = false, maxKeepersShown = 20, maxLibrariesShown = 15, maxTagsShown = 0, items = uriIds.map(AugmentableItem(_))).map { infos =>
```
ryanpbrewster, line 53: :+1:

ryanpbrewster:
:+1: lgtm

If/when we end up making this endpoint return the same thing as `/site/libraries/:id` we should simplify the tests, as we discussed in person.

adamjgrant:
For the record, I haven't yet tested the front end on this.

PR #16430, opened 2015-07-28
This shouldn't be called, and is currently very noisy.

andrewconner:
@stkem Commenting out for now just so we don't forget about it. Previously, it was understood that if we're saving a `Collection`, then we have a reason to save it — we must have created, deleted, or changed it. Right now, we're calling `save` to increment the seq num every time someone keeps with that collection. I noticed an import _nailing_ a websocket connection


stkem:
lgtm. Good Catch!

PR #16431, opened 2015-07-28
Factor out pagination for org libs and user libs

cvializ:
@andrewconner @ajkochanowicz forgot to add pagination to the org libraries page


andrewconner:
Nice!

PR #16433, opened 2015-07-28
Domain changes

camhashemi:
@gratimax @leogrim


kifi-backend/modules/common/app/com/keepit/classify/Domain.scala
```diff
@@ -21,11 +22,10 @@ case class Domain(
     autoSensitive: Option[Boolean] = None,
     manualSensitive: Option[Boolean] = None,
     isEmailProvider: Boolean = false,
-    hash: Option[DomainHash],
     state: State[Domain] = DomainStates.ACTIVE,
     createdAt: DateTime = currentDateTime,
     updatedAt: DateTime = currentDateTime) extends ModelWithState[Domain] {
-  //require(this.hostname.toLowerCase == this.hostname, "Domain.hostname must be lowercase")
+  require(hostname == IDN.toASCII(hostname), "Invalid hostname: use Domain.fromHostname(hostname)")
```
leogrim: To be added once all domains have been migrated?

kifi-backend/modules/common/app/com/keepit/classify/Domain.scala
```diff
@@ -2,6 +2,7 @@ package com.keepit.classify
 
 import java.security.MessageDigest
 
+import java.net.IDN
```
leogrim, line 5: So you're not using Google's library in the end? I think Google's lowercased hostnames by default, you may have to do this manually with Java's.

kifi-backend/modules/common/conf/evolutions/shoebox/361.sql
```diff
@@ -2,12 +2,10 @@
 
 # --- !Ups
 
-ALTER TABLE domain ADD hash VARCHAR(26) DEFAULT NULL;
+CREATE INDEX user_ip_addresses_i_seq ON user_ip_addresses(seq);
 
-CREATE UNIQUE INDEX domain_u_hash ON DOMAIN (hash);
+CREATE INDEX organization_membership_i_seq ON organization_membership(seq);
 
-DROP INDEX domain_hostname_index;
```
leogrim, line 9: We want that index back, ideally renamed by our convention though

kifi-backend/modules/common/app/com/keepit/classify/Domain.scala
```diff
@@ -91,16 +70,6 @@ case class DomainKey(hostname: String) extends Key[Domain] {
   def toKey(): String = hostname
 }
```
leogrim, line 92: `DomainKey` should have the same normalization as `hostname`

kifi-backend/modules/shoebox/app/com/keepit/classify/DomainRepo.scala
```diff
@@ -43,18 +39,15 @@ class DomainRepoImpl @Inject() (
...
   def table(tag: Tag) = new DomainTable(tag)
   initTable()
 
   def get(domain: String, excludeState: Option[State[Domain]] = Some(DomainStates.INACTIVE))(implicit session: RSession): Option[Domain] = {
-    domainHashCache.getOrElseOpt(DomainHashKey(DomainHash.hashHostname(domain))) {
-      domainCache.getOrElseOpt(DomainKey(domain)) {
-        (for (d <- rows if d.hostname === domain && d.state =!= excludeState.orNull) yield d).firstOption
-      }
+    domainCache.getOrElseOpt(DomainKey(domain)) {
+      (for (d <- rows if d.hostname === domain && d.state =!= excludeState.orNull) yield d).firstOption
```
leogrim, line 50: `domain` needs to be normalized here, otherwise you're comparing apples to oranges

kifi-backend/modules/shoebox/app/com/keepit/classify/DomainRepo.scala
```diff
@@ -66,12 +59,6 @@ class DomainRepoImpl @Inject() (
   def getAllByName(domains: Seq[String], excludeState: Option[State[Domain]] = Some(DomainStates.INACTIVE))(implicit session: RSession): Seq[Domain] =
     (for (d <- rows if d.hostname.inSet(domains) && d.state =!= excludeState.orNull) yield d).list
```
leogrim, line 60: Same here, `domains` need to be normalized

kifi-backend/modules/common/app/com/keepit/classify/Domain.scala
camhashemi, line 5: Yes, but it doesn't convert to punycode. Would rather convert to lowercase than convert to punycode...

kifi-backend/modules/common/conf/evolutions/shoebox/361.sql
```diff
@@ -2,12 +2,14 @@
...
-ALTER TABLE domain ADD hash VARCHAR(26) DEFAULT NULL;
+CREATE INDEX user_ip_addresses_i_seq ON user_ip_addresses(seq);
 
-CREATE UNIQUE INDEX domain_u_hash ON DOMAIN (hash);
+CREATE INDEX organization_membership_i_seq ON organization_membership(seq);
 
 DROP INDEX domain_hostname_index;
 
-insert into evolutions (name, description) values('361.sql', 'add hash, drop hostname index from `domain`');
+CREATE INDEX domain_u_hostname ON `domain`(hostname);
```
camhashemi: renamed

kifi-backend/modules/common/app/com/keepit/classify/Domain.scala
```diff
@@ -39,68 +39,43 @@ case class Domain(
...
-  def hashHostname(hostname: String): DomainHash = {
-    val binaryHash = MessageDigest.getInstance("MD5").digest(hostname)
-    DomainHash(new String(new Base64().encode(binaryHash), UTF8))
-  }
-
-  implicit val format: Format[DomainHash] = new Format[DomainHash] {
-    def reads(json: JsValue): JsResult[DomainHash] = json.validate[String].map(DomainHash.apply)
-    def writes(o: DomainHash): JsValue = JsString(o.hash)
-  }
+  def isValid(s: String): Boolean = DomainRegex.pattern.matcher(s).matches && Try(IDN.toASCII(s)).isSuccess
```
camhashemi: IDN.toASCII(s) will throw an exception for invalid domain names, so we have to be careful with it

kifi-backend/modules/common/app/com/keepit/classify/Domain.scala
```diff
@@ -39,68 +39,43 @@ case class Domain(
...
     (__ \ 'manualSensitive).formatNullable[Boolean] and
     (__ \ 'isEmailProvider).format[Boolean] and
-    (__ \ 'hash).formatNullable[DomainHash] and
     (__ \ 'state).format(State.format[Domain]) and
     (__ \ 'createdAt).format[DateTime] and
     (__ \ 'updatedAt).format[DateTime]
   )(Domain.apply, unlift(Domain.unapply))
 
   private val DomainRegex = """^(?:[a-z0-9](?:[a-z0-9-]{0,61}[a-z0-9])?\.)+[a-z0-9]+$""".r
-  private val MaxLength = 256
```
camhashemi, line 53: IDN.toASCII checks length

kifi-backend/modules/common/conf/evolutions/shoebox/361.sql
andrewconner: This won't be fast. Besides naming reasons, why drop and recreate? The table will be nuked for this window of time.

kifi-backend/modules/common/app/com/keepit/classify/Domain.scala
```diff
@@ -51,56 +50,33 @@ object Domain {
...
 
 @json
 case class DomainInfo(
   id: Option[Id[Domain]],
-  hostname: String,
+  hostname: NormalizedHostname,
   isEmailProvider: Boolean)
 
+case class NormalizedHostname(value: String) {
+  require(value == IDN.toASCII(value.toLowerCase), "Invalid hostname, use NormalizedHostname.fromHostname")
```
andrewconner: Why not just not use `case`, and write your own `applyOpt`?

kifi-backend/modules/shoebox/test/com/keepit/classify/DomainTest.scala
```diff
@@ -76,22 +76,23 @@ class DomainTest extends Specification with ShoeboxTestInjector {
...
+          val actual2 = domainRepo.get("Google.com").headOption
           actual2.isDefined === true
-          actual2.get.hostname === "google.com"
+          actual2.get.hostname.value === "google.com"
 
-          val notDesired1 = domainRepo.getAllByNameUsingHash(Set("Ğooğle.com")).headOption
-          notDesired1.isDefined === false
+          val actual3 = domainRepo.get("Ğooğle.com").headOption
+          actual3.isDefined === true
+          actual3.get.hostname === NormalizedHostname.fromHostname("Ğooğle.com")
```
andrewconner: Does this pass? Shouldn't it be `xn--oole-zwac.com`?

kifi-backend/modules/common/conf/evolutions/shoebox/361.sql
```diff
@@ -2,12 +2,14 @@
 
 # --- !Ups
```
andrewconner, line 3: I'm sure you realized, but be careful about merging evolutions that you're not intending on applying. This evolution hasn't been safe to apply, but is applied in everyone's dev mode and should be expected to be in production. Multi-step migrations (as you were moving to hash, adding a new index, etc) should be done in multiple steps.

kifi-backend/modules/common/conf/evolutions/shoebox/361.sql
camhashemi: only naming reasons. it's probably fine as is.

kifi-backend/modules/shoebox/test/com/keepit/classify/DomainTest.scala
camhashemi: `NormalizedHostname.fromHostname` does that conversion. I'll add another test to show the raw punycode.

kifi-backend/modules/common/app/com/keepit/classify/Domain.scala
camhashemi: I would like to design this better. So with `applyOpt`, instead of throwing an error return `None`?

leogrim: The assertion should probably be part of a factory method rather than the
class itself, otherwise you may end up not being able to read from the
database if something corrupt gets in there for some reason. Check out
'EmailAddress' for a similar scenario

kifi-backend/modules/common/conf/evolutions/shoebox/361.sql
camhashemi, line 3: that makes sense, didn't think about the multi-step process. thank you.

kifi-backend/modules/common/app/com/keepit/classify/Domain.scala
camhashemi: Looks like `EmailAddress` throws a JsError on reads when something is corrupt. Is that much different from throwing an exception with `require`?

The way I'm thinking about this is that 1) I don't want people to use `NormalizedHostname.apply`, I want them to use `NormalizedHostname.fromHostname`. If something's corrupt in the database (which shouldn't happen if (1) is successful), we should throw an exception and fix the data. Not sure what other considerations there are or how to implement them nicely.

leogrim: It's different because the requirement is not part of the class body, which means you can get around it when deserializing from the db. The assumption there is: validate everything going in or out (i.e. Json serialization, query or path parameters), but trust what's in the db. Even if something bad is in the db, you'll have to get it out before you can fix it. And you don't want to wrap all your repo methods with `try`.

camhashemi:
@andrewconner @leogrim @gratimax looking to merge. important changes are adding validation on the controllers to handle invalid domain names that would cause `NormalizedHostname.fromHostname` to throw an exception.

also, we are using one hostname column in `OrganizationDomainOwnership` which will be typed to `NormalizedHostname`, instead of `String`

kifi-backend/modules/common/app/com/keepit/classify/Domain.scala
```diff
@@ -39,68 +40,57 @@ case class Domain(
...
 
-object DomainHash {
-  def hashHostname(hostname: String): DomainHash = {
-    val binaryHash = MessageDigest.getInstance("MD5").digest(hostname)
-    DomainHash(new String(new Base64().encode(binaryHash), UTF8))
-  }
+object NormalizedHostname {
+  private val DomainRegex = """^(?:[a-z0-9](?:[a-z0-9-]{0,61}[a-z0-9])?\.)+[a-z0-9]+$""".r
+  private val MaxLength = 256
+  def isValid(s: String): Boolean = DomainRegex.pattern.matcher(s).matches && s.length < MaxLength && Try(IDN.toASCII(s)).isSuccess
```
andrewconner, line 73: I don't see this used much, besides in search. How come?

kifi-backend/modules/common/app/com/keepit/classify/Domain.scala
```diff
@@ -39,68 +40,57 @@ case class Domain(
...
+
+  def fromHostname(hostname: String) = NormalizedHostname(IDN.toASCII(hostname.toLowerCase))
```
andrewconner: Probably would be best for this guy not to throw an exception on error. I see that you end up with lots of defensive, error handling code elsewhere to deal with it — and if someone wants to use it and forgets, whomp. See the `reads` below.

kifi-backend/modules/shoebox/test/com/keepit/classify/DomainTest.scala
```diff
@@ -76,22 +76,24 @@ class DomainTest extends Specification with ShoeboxTestInjector {
...
           actual2.isDefined === true
-          actual2.get.hostname === "google.com"
+          actual2.get.hostname.value === "google.com"
 
-          val notDesired1 = domainRepo.getAllByNameUsingHash(Set("Ğooğle.com")).headOption
-          notDesired1.isDefined === false
+          val actual3 = domainRepo.get(NormalizedHostname.fromHostname("Ğooğle.com"))
+          actual3.isDefined === true
+          actual3.get.hostname === NormalizedHostname.fromHostname("Ğooğle.com")
+          actual3.get.hostname.value === "xn--oole-zwac.com"
```
andrewconner, line 96: :+1:

kifi-backend/modules/common/app/com/keepit/classify/Domain.scala
camhashemi, line 73: Previously, this was `Domain.isValid`, but all that it was checking was the hostname. Now that there's a hostname class, I refactored `Domain.isValid` -> `NormalizedHostname.isValid`. I see there's only one use.

kifi-backend/modules/common/app/com/keepit/classify/Domain.scala
camhashemi: What should the on-error behavior be then? It's going to throw an exception in the case that `hostname` doesn't come directly from a valid domain, which is rare but possible. I have to handle that case, even if I use `.recover`.

Would it be better to return an `Option` or `Try`? Still would need the defensive code to handle it, right?

andrewconner: Returning an `Option` wouldn't be defensive, unless you then did a `.get`. You just have to handle both cases.

camhashemi: Okay, np. Thank you.

andrewconner: Either way, this is rotting :/. I do think you should fix this (it's usually unexpected to throw exceptions), but you may want to get this PR in first so you get the burden off.

camhashemi: It's only incrementally more work, since all of the exception handling code converts it into an `Option` to deal with it anyways. Will merge after.

leogrim: Also, aren't you supposed to call 'toLowercase' on the ASCII, not on the original UNICODE domain? Seems like the composition is backwards

kifi-backend/modules/common/conf/evolutions/shoebox/365.sql
```diff
@@ -4,7 +4,7 @@
 
 ALTER TABLE organization_domain_ownership DROP CONSTRAINT unique_organization_id_and_domain_id;
 ALTER TABLE organization_domain_ownership DROP COLUMN domain_id;
-ALTER TABLE organization_domain_ownership ADD domain_hostname VARCHAR(128) NOT NULL;
+ALTER TABLE organization_domain_ownership ADD domain_hostname VARCHAR(256) NOT NULL;
```
leogrim, line 7: Did you mean to change the name of the val from `domain_hostname` to `nomalized_hostname`? (might not be necessary since the type makes it clear)

kifi-backend/modules/common/app/com/keepit/classify/Domain.scala
camhashemi: yes nice catch

kifi-backend/modules/common/conf/evolutions/shoebox/365.sql
camhashemi, line 7: I'd prefer to do that, to match the model and the field type.

leogrim, line 7: I don't feel strongly about this one, but note that this is not a rule, do
you name all your variables exactly like their type? Sometimes it's clearer
if you need to clearly differentiate from some other variables, sometimes a
short descriptive name is better than repeating the type.

E.g. 'UserEmailAddress' just has an 'address' of type 'EmailAddress', it's
shorter and unambiguous.

camhashemi, line 7: No, not all of them, haha. It does help to know from a variable name what to expect from it, but I don't think this case will suffer without prepending `normalized`.

PR #16434, opened 2015-07-28
Replace multi-hop logic with just one api call

cvializ:
@andrewconner @ajkochanowicz


andrewconner:
Net -33 lines, I approve.

cvializ:
:joy: so beautiful

adamjgrant:
Good stuff!

PR #16435, opened 2015-07-29
Show link to organizations near profile widget

adamjgrant:


adamjgrant:
The only thing I'm missing here is the image for the org, but I assume that will come later.
screen shot 2015-07-28 at 5 58 55 pm

adamjgrant:
@cvializ @andrewconner @camhashemi @ryanpbrewster

cvializ:
you can display the org's photo in the template with the `pic` filter: ``

cvializ:
:+1: looks good!

camhashemi:
@ryanpbrewster just added an endpoint for uploading an image, right? We just need to return it with an org then?

adamjgrant:
Looks like the solution there will be simple. I'll take a look tomorrow
morning. Thanks all.

adamjgrant:
I think we're good to go @cvializ @andrewconner @ryanpbrewster @camhashemi

kifi-backend/modules/shoebox/angular/src/common/filter.js
```diff
@@ -32,9 +32,10 @@ angular.module('kifi')
 })
 
 .filter('profileUrl', function () {
-  return function (user, sub) {
-    if (user) {
-      return '/' + user.username + (sub && sub !== 'libraries' ? '/' + sub : '');
+  return function (userOrOrg, sub) {
+    if (userOrOrg) {
+      var handle = 'username' in userOrOrg ? userOrOrg.username : userOrOrg.handle;
+      return '/' + handle + (sub && sub !== 'libraries' ? '/' + sub : '');
```
adamjgrant, line 38: Would be nice to deprecate `username` in favor of `handle` in the future.

cvializ, line 38: Agreed.

On an unrelated note: another syntax you could use that would be more JavaScript-y/terse than a ternary would be

```
var handle = userOrOrg.username || userOrOrg.handle;
```

not that you have to change it, it's more of a stylistic choice so feel free to disregard

kifi-backend/modules/shoebox/angular/src/profile/profileWidget.js
```diff
@@ -19,6 +19,10 @@ angular.module('kifi')
           });
         };
 
+        net.getOrgsForUser(scope.me.id).then(function(result) {
```
andrewconner, line 22: This should be included in the `me` object, it'll always be needed (this route _can_ exist if you want, but should always\* come down for free). @camhashemi

 (*) Perhaps I'm missing a case, but every one I see it will.

cvializ:
:+1:

andrewconner:
Visually, looks great. Only concern is the endpoint (we can merge, but I'd expect that `profileService.me` comes with the orgs, and you stop requesting it directly)

adamjgrant:
I don't see it today. Or are you saying it ought to be included in the future?

screen shot 2015-07-29 at 11 40 21 am

andrewconner:
Yes, it ought to be added (soon, so you don't have to manually wire everything)

adamjgrant:
Gotcha. I'll merge now and will be on the lookout for an endpoint change.

kifi-backend/modules/shoebox/angular/src/common/filter.js
adamjgrant, line 38: I've experienced problems with this approach where JS throws an error "undefined method `username`" simply because it's called in the OR. Otherwise, I definitely prefer this syntax.

adamjgrant, line 38: I'll have to catch it the next time it happens, because as far as I can tell, it _shouldn't_ throw said error

screen shot 2015-07-29 at 11 52 33 am

I might be mixing this up with calling a property on another property which is itself undefined, like

```
a.b.c
```

where `a` does not have a property `b`

cvializ, line 38: yep, that's right. when it gets complicated switching to ternary or `if` are definitely more readable. but to get undefined method with just two you'd have to do `a.b() || a.c()` and have `a.b` or `a.c` be undefined

adamjgrant, line 38: Thanks. Opened a new PR that will include this change. https://github.com/kifi/keepit/pull/16448

adamjgrant:
Opened a second PR for upcoming changes to API and adding @cvializ 's advice https://github.com/kifi/keepit/pull/16448

PR #16437, opened 2015-07-29
fix javascript bug with angular amplitude

joshm1:
@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/common/services/mixpanelService.js
+++ b/kifi-backend/modules/shoebox/angular/src/common/services/mixpanelService.js
@@ -189,12 +189,13 @@
   function getAgentProperties() {
     var amplitude = $window.amplitude;
     var ua = amplitude && amplitude._ua;
+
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/angular/src/common/services/mixpanelService.js
```diff
@@ -189,12 +189,13 @@
   function getAgentProperties() {
     var amplitude = $window.amplitude;
     var ua = amplitude && amplitude._ua;
+
+    // since most of these properties depend on the amplitude SDK, additional
+    // checks are made to ensure we're not referencing undefined/null objects
     return {
-      language: amplitude && amplitude.options.language || null,
-      platform: amplitude && amplitude.options.platform || null,
```
joshm1, line 194: `amplitude.options` is always undefined here. it needs to be defined by our code before we call amplitude.init for it to be useful (which is incomplete for the platform property)

PR #16439, opened 2015-07-29
remove related keeps code from the cortex

eishay:


eishay:
@andrewconner

eishay:
@stkem

stkem:
lgtm. I assume the call to this from shoebox is already gone?

PR #16440, opened 2015-07-29
Sending new users to Eishay instead of Tamila

andrewconner:
@stkem @ajkochanowicz


andrewconner:
Comments still appreciated, but merging to get it out.

stkem:
I don't exactly have Eishay's external id memorized, but rest lgtm :-)

PR #16442, opened 2015-07-29
report the entire stack trace. trying to understand a fail ingestion …

eishay:
…by cortex related to the read it later library


eishay:
@stkem

stkem:
Is the read it later library a new `kind`?

andrewconner:
Aw, if it was a sealed trait, the compiler would tell you the problem. Anyways, looks like there may be a few problems: LibraryIndexable doesn't know about this. It will barf.

Also, system_read_**id**_later

PR #16443, opened 2015-07-29
Removed scaled images from Org Avatar Commander

ryanpbrewster:
This is desired because org avatars should not preserve aspect ratio (they
ought to be square). It also happens to fix a bug where square images were
being scaled and "cropped" to the same size, breaking a DB constraint.

@andrewconner @cvializ

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/commanders/ProcessedImageHelper.scala
+++ b/kifi-backend/modules/common/app/com/keepit/commanders/ProcessedImageHelper.scala
@@ -418,7 +418,9 @@ object ScaledImageSize {
 }
 
 object CroppedImageSize {
-  case object Small extends CroppedImageSize("small", ImageSize(150, 150))
+  case object Tiny extends CroppedImageSize("small", ImageSize(100, 100))
+  case object Small extends CroppedImageSize("medium", ImageSize(150, 150))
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/common/app/com/keepit/commanders/ProcessedImageHelper.scala
```diff
@@ -418,7 +418,9 @@ object ScaledImageSize {
 }
 
 object CroppedImageSize {
-  case object Small extends CroppedImageSize("small", ImageSize(150, 150))
+  case object Tiny extends CroppedImageSize("small", ImageSize(100, 100))
+  case object Small extends CroppedImageSize("medium", ImageSize(150, 150))
```
andrewconner, line 422: This should be renamed, and shouldn't be labeled "small"

kifi-backend/modules/common/app/com/keepit/commanders/ProcessedImageHelper.scala
```diff
@@ -418,7 +418,9 @@ object ScaledImageSize {
...
+  case object Medium extends CroppedImageSize("medium", ImageSize(200, 200))
 
   val allSizes: Seq[CroppedImageSize] = Seq(Small)
```
andrewconner, line 425: This should probably be removed. It's not true.

PR #16444, opened 2015-07-29
Fix undefined reference error.

cvializ:
@andrewconner @ajkochanowicz 

should resolve https://fortytwo.airbrake.io/projects/113130/groups/1474976911655347397/notices/1474977231538136271


adamjgrant:
Nice.

PR #16445, opened 2015-07-29
Consume the org's avatarPath

cvializ:
@ajkochanowicz @andrewconner @ryanpbrewster


adamjgrant:
Looks good, will this change anything in my implementation in https://github.com/kifi/keepit/pull/16435 ?

cvializ:
Nope, in fact it will make it work even better. Before this PR, #16435 will only show a gray silhouette for the avatar

adamjgrant:
:+1:

PR #16446, opened 2015-07-29
suppressing missing error attribute in response

adamjgrant:
Addresses https://fortytwo.airbrake.io/projects/113130/groups/1475437773365093740/notices/1475437779488777579


andrewconner:
:+1:

PR #16447, opened 2015-07-29
order orgMembers by 1) owner, 2) members alphabetically, 3) invitees (DO NOT MERGE)

camhashemi:
@cvializ here you go

@ryanpbrewster

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
@@ -90,10 +90,12 @@ class OrganizationMembershipCommanderImpl @Inject() (
 
   def getMembersAndInvitees(orgId: Id[Organization], limit: Limit, offset: Offset, includeInvitees: Boolean): Seq[MaybeOrganizationMember] = {
     db.readOnlyMaster { implicit session =>
-      val members = organizationMembershipRepo.getByOrgId(orgId, limit, offset)
+      val org = organizationRepo.get(orgId)
+      val (ownerSeq, members) = organizationMembershipRepo.getByOrgId(orgId, limit, offset).partition(_.userId == org.ownerId)
+      val ownerOpt = ownerSeq.headOption
       val invitees = includeInvitees match {
         case true =>
-          val leftOverCount = Limit(Math.max(limit.value - members.length, 0))
+          val leftOverCount = Limit(Math.max(limit.value - members.length + 1, 0))
           val leftOverOffset = if (members.isEmpty) {
             val totalCountForOrg = organizationMembershipRepo.countByOrgId(orgId)
             Offset(offset.value - totalCountForOrg)
@@ -103,7 +105,7 @@ class OrganizationMembershipCommanderImpl @Inject() (
           organizationInviteRepo.getByOrganization(orgId, leftOverCount, leftOverOffset)
         case false => Seq.empty[OrganizationInvite]
       }
-      buildMaybeMembers(members, invitees)
+      buildMaybeMembers(ownerOpt, members, invitees)
     }
   }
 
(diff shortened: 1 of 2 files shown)


cvializ:
hooray!

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
```diff
@@ -155,7 +161,7 @@ class OrganizationMembershipCommanderImpl @Inject() (
       MaybeOrganizationMember(member = Right(contact), role = invitedByAddress.role, lastInvitedAt = Some(invitedByAddress.updatedAt))
     }
 
-    membersInfo ++ invitedByUserIdInfo ++ invitedByEmailAddressInfo
+    ownerInfo.toSeq ++ membersInfo.sortBy(_.member.left.get.firstName) ++ invitedByUserIdInfo ++ invitedByEmailAddressInfo
```
andrewconner, line 164: You can sort by last name too using tuples. Example:

```
scala> Seq(("Andrew", "Conner"), ("Andrew", "Bonner"), ("Andrew", "Lonner"), ("Cam", "Smashemi"), ("Cam", "Hashemi")).sorted
res1: Seq[(String, String)] = List((Andrew,Bonner), (Andrew,Conner), (Andrew,Lonner), (Cam,Hashemi), (Cam,Smashemi))
```

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
```diff
@@ -131,11 +133,15 @@ class OrganizationMembershipCommanderImpl @Inject() (
     }
   }
 
-  private def buildMaybeMembers(members: Seq[OrganizationMembership], invitees: Seq[OrganizationInvite]): Seq[MaybeOrganizationMember] = {
+  private def buildMaybeMembers(owner: Option[OrganizationMembership], members: Seq[OrganizationMembership], invitees: Seq[OrganizationInvite]): Seq[MaybeOrganizationMember] = {
     val invitedUserIds = invitees.filter(_.userId.isDefined)
     val invitedEmailAddresses = invitees.filter(_.emailAddress.isDefined)
     val usersMap = db.readOnlyMaster { implicit session =>
-      basicUserRepo.loadAllActive((members.map(_.userId) ++ invitedUserIds.map(_.userId.get)).toSet)
+      basicUserRepo.loadAllActive((owner.map(owner => Seq(owner.userId)).getOrElse(Seq.empty) ++ members.map(_.userId) ++ invitedUserIds.map(_.userId.get)).toSet)
```
andrewconner, line 140: For the owner, another way to write it with `.toSeq`:

```
scala> Some("Hello!").map(_.head).toSeq
res4: Seq[Char] = List(H)
```

camhashemi:
@cvializ: can't merge yet, test problem. I had to switch to other priorities.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
```diff
@@ -90,10 +90,12 @@ class OrganizationMembershipCommanderImpl @Inject() (
 
   def getMembersAndInvitees(orgId: Id[Organization], limit: Limit, offset: Offset, includeInvitees: Boolean): Seq[MaybeOrganizationMember] = {
     db.readOnlyMaster { implicit session =>
```
ryanpbrewster, line 92: Does this need to be `readOnlyMaster`?

ryanpbrewster:
I think this would be much cleaner if we just assigned an ordering on members/invitees and then sorted by that in the db.

The ordering could be `(isOwner, role, lastname, firstname)` for members and just alphabetical by invite.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
```diff
@@ -90,10 +90,12 @@ class OrganizationMembershipCommanderImpl @Inject() (
...
-      val members = organizationMembershipRepo.getByOrgId(orgId, limit, offset)
+      val org = organizationRepo.get(orgId)
+      val (ownerSeq, members) = organizationMembershipRepo.getByOrgId(orgId, limit, offset).partition(_.userId == org.ownerId)
```
andrewconner, line 94: This will actually yield weird results when they paginate. Each page is sorted, but the pages themselves aren't.

camhashemi, line 94: seems to me like we shouldn't be getting memberships and paginating at the same time. we should get them all, sort, then `slice(offset, offset+limit)`. yeah?

andrewconner, line 94: Sort in the db.

PR #16449, opened 2015-07-29
add OrganizationCard to BasicUserInfo (/site/user/me)

camhashemi:
@andrewconner @ajkochanowicz @cvializ

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/UserController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/UserController.scala
@@ -317,7 +317,7 @@ class UserController @Inject() (
         "numFollowers" -> pimpedUser.numFollowers,
         "experiments" -> experiments.map(_.value),
         "pendingFriendRequests" -> pendingFriendRequests,
-        "orgs" -> pimpedUser.orgs.map(Json.toJson(_)(OrganizationCard.websiteWrites))
+        "orgs" -> pimpedUser.orgs
       )
     Ok(json)
   }
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/UserController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/UserController.scala
@@ -316,7 +316,8 @@ class UserController @Inject() (
         "numConnections" -> pimpedUser.numConnections,
         "numFollowers" -> pimpedUser.numFollowers,
         "experiments" -> experiments.map(_.value),
-        "pendingFriendRequests" -> pendingFriendRequests
+        "pendingFriendRequests" -> pendingFriendRequests,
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/shoebox/app/com/keepit/controllers/website/UserController.scala
```diff
@@ -316,7 +316,8 @@ class UserController @Inject() (
         "numConnections" -> pimpedUser.numConnections,
         "numFollowers" -> pimpedUser.numFollowers,
         "experiments" -> experiments.map(_.value),
-        "pendingFriendRequests" -> pendingFriendRequests
+        "pendingFriendRequests" -> pendingFriendRequests,
+        "orgs" -> pimpedUser.orgs.map(Json.toJson(_)(OrganizationCard.websiteWrites))
```
andrewconner: This part (`Json.toJson(_)(OrganizationCard.websiteWrites)`) shouldn't be necessary. Did the compiler complain without it?

kifi-backend/modules/shoebox/test/com/keepit/controllers/website/UserControllerTest.scala
```diff
@@ -65,7 +67,8 @@ class UserControllerTest extends Specification with ShoeboxTestInjector {
               "numLibraries":0,
               "numConnections":0,
               "numFollowers":0,
-              "pendingFriendRequests":0
+              "pendingFriendRequests":0,
+              "orgs": []
```
andrewconner, line 72: Since it looks like this is the only test affected, how about add an org so we test that they come down right?

andrewconner:
:+1:

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/UserController.scala
camhashemi: unfortunately it did.

@ryanpbrewster I want to give `OrganizationCard` and other client-consumable models an `implicit defaultWrites`. If we need a different serialization for mobile/web, we can explicitly specify. But not using implicits is a pain, especially when the serialization is the same and the model is meant to be consumed.

ryanpbrewster: Change it to an `implicit`. I'm not personally a fan of them, but their convenience is hard to ignore.

kifi-backend/modules/shoebox/test/com/keepit/controllers/website/UserControllerTest.scala
ryanpbrewster, line 72: :+1:

camhashemi:
@adamjgrant here you go, I'll deploy. @andrewconner writing the test as we speak. didn't want to block any longer.

adamjgrant:
Woo hoo!

PR #16451, opened 2015-07-29
Aggressive log cutbacks on Rover

stkem:
Let me know if this is too heavy
@leogrim @andrewconner


stkem:
@eishay

stkem:
Since @leogrim is currently in the air, could one of you @andrewconner or @eishay give this a quick once over if there is anything you know to be essential. We don't have a lot of time left to get our logging under control and we can always revert it when Léo comes back on and sees stuff he needs. (Even with cleanups from yesterday we've already sent ~7GB _so far_ today and it's been a low exception day)

andrewconner:
Looks good to me

PR #16452, opened 2015-07-29
Users can choose space where library is created/modified into

adamjgrant:


andrewconner:
:+1:

kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.js
```diff
@@ -135,7 +149,8 @@ angular.module('kifi')
             return sub.name !== '' && sub.info.url !== '';
           });
           
-          var owner = ('organization' in scope.modalData ? { org: scope.modalData.organization.id } : { user: profileService.me.id });
+          var owner = {};
+          owner['numMembers' in scope.thisSpace ? 'org' : 'user'] = scope.thisSpace.id;
```
adamjgrant: I'm a little iffy about this one–to determine if a selected space is an org, we look at whether or not `numMembers` is a property of the `space`.

It makes sense, as users wouldn't have members, but it's kind of a weird property to rely on.

Currently, `spaces` are just an array consisting of  `profileService.me` plus the orgs in `profileService.me.orgs`

I could add `type: org` to each `org` in `profileService.me.orgs` but I'd rather not add data into it that doesn't actually come from the server.

What do you think, @cvializ ?

cvializ: You're right, it is a little weird to depend on the property being present, but it's weirder to mutate data structures coming from the back end. In cases like this, I like to have identifiers in the code communicate intent. You could add a function to communicate the intent of the property being present

```
function isOrg(space) { return 'numMembers' in scope.thisSpace; }
```

and then also have a variable describing what `'org'` and `'user'` mean

```
var ownerType = (isOrg(scope.thisSpace) ? 'org' : 'user');
```

and then make the assignment

```
owner[ownerType] = scope.thisSpace.id;
```

That should make the weirdness a bit more self-documenting.

adamjgrant: That might be a good compromise. I'll try something like that. Thanks.

adamjgrant:
Okay, I think we're good. @cvializ @andrewconner @andrewconner @camhashemi

cvializ:
:+1: !

PR #16453, opened 2015-07-29
Make OrganizationDomainOwnership reference domain hostname

maxov:
@leogrim 
@camhashemi


kifi-backend/modules/common/app/com/keepit/model/OrganizationDomainOwnership.scala
```diff
@@ -17,13 +17,13 @@ case class OrganizationDomainOwnership(
     state: State[OrganizationDomainOwnership] = OrganizationDomainOwnershipStates.ACTIVE,
     seq: SequenceNumber[OrganizationDomainOwnership] = SequenceNumber.ZERO,
     organizationId: Id[Organization],
-    domainId: Id[Domain]) extends ModelWithSeqNumber[OrganizationDomainOwnership] with ModelWithState[OrganizationDomainOwnership] {
+    domainHostname: String) extends ModelWithSeqNumber[OrganizationDomainOwnership] with ModelWithState[OrganizationDomainOwnership] {
```
leogrim, line 20: Here should have both a `hostname` and `normalizedHostname` of type `NormalizedHostname` to take care of unicode domains. Then `OrganizationDomainOwnership.normalizedHostname` should be 256 ASCII character-long, indexed, and reference `Domain.hostname`. Any lookup / unique logic should use it. 

Meanwhile `OrganizationDomainOwnership.hostname` should be 256 Unicode character-long, and only be used for display to users. Does this make sense? @gratimax @camhashemi @andrewconner

PR #16454, opened 2015-07-29
add a test for `orgs` in `/site/user/me`

camhashemi:
@andrewconner @ryanpbrewster

### DIFF
--- a/kifi-backend/modules/shoebox/test/com/keepit/controllers/website/UserControllerTest.scala
+++ b/kifi-backend/modules/shoebox/test/com/keepit/controllers/website/UserControllerTest.scala
@@ -71,7 +71,7 @@ class UserControllerTest extends Specification with ShoeboxTestInjector {
               "numConnections":0,
               "numFollowers":0,
               "pendingFriendRequests":0,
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/test/com/keepit/controllers/website/UserControllerTest.scala
```diff
@@ -69,7 +71,7 @@ class UserControllerTest extends Specification with ShoeboxTestInjector {
               "numConnections":0,
               "numFollowers":0,
               "pendingFriendRequests":0,
-              "orgs": []
+              "orgs": [{"id":"${Organization.publicId(org.id.get)(inject[PublicIdConfiguration]).id}","ownerId":"${user.externalId}","handle":"${org.getHandle.value}","name":"${org.name}","numMembers":${1},"numLibraries":${0}}]
```
andrewconner: What's the `${1}` for?

camhashemi: To tell future-someone that when they add members to the org, this needs to be a variable. We're not testing the values of `numMembers` and `numLibraries` here, but I could do that.

camhashemi: I see that we just hard-code non-tested values in other places, I could also do that.

andrewconner: Yeah. you're already hard coding values, making Scala call `1.toString` and concat it back into the string may confuse someone.

PR #16468, opened 2015-07-30
Add OrganizationCard to LibraryCardInfo and FullLibraryInfo

camhashemi:
@andrewconner @ryanpbrewster 

lots of refactoring here. importantly, `Organization.getHandle` -> `.handle` and `Organization.handle` -> `.primaryHandle` @leogrim to match the pattern in `User`

also added a new serialization pattern for 22-field case-classes


kifi-backend/modules/common/app/com/keepit/common/Implicits.scala
```diff
@@ -75,6 +76,10 @@ final class TraversableExtensionOps[A](xs: Traversable[A]) {
   def countAll: Map[A, Int] = countBy(identity)
 }
 
+final class JsObjectExtensionOps(x: JsObject) {
+  def nonNullFields: JsObject = JsObject(x.fields.filter { case (_, b) => b != JsNull })
```
ryanpbrewster, line 80: Can we name this `nonNullValues` instead, since it's not actually operating on the fields?

camhashemi, line 80: The naming implies what you're getting back. `nonNullValues` would look like `.values.filter(_.notNull)`, whereas this is `fields.filter(_.notNull)`.

ryanpbrewster, line 80: Ah, my mistake, I was mixing up `field` and `key`.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -298,9 +299,12 @@ class LibraryCommanderImpl @Inject() (
       db.readOnlyMasterAsync { implicit s => basicUserRepo.loadAll(allUsersShown) } //cached
     }
 
-    val orgsByIdF = {
+    val orgCardByIdF = {
       val allOrgsShown = libraries.flatMap { library => library.organizationId }.toSet
-      db.readOnlyMasterAsync { implicit s => orgRepo.getByIds(allOrgsShown) }
+      db.readOnlyMasterAsync { implicit s =>
```
ryanpbrewster: Can this be a `readOnlyReplica`?

camhashemi:
@andrewconner I should have gotten this out yesterday, would appreciate a review.

andrewconner:
Looks good!

kifi-backend/modules/common/app/com/keepit/model/Organization.scala
```diff
@@ -55,14 +55,14 @@ case class Organization(
   def modifiedMembership(membership: OrganizationMembership, newRole: OrganizationRole): OrganizationMembership =
     membership.copy(role = newRole, permissions = getRolePermissions(newRole))
 
-  def getHandle: OrganizationHandle = {
-    handle match {
+  def handle: OrganizationHandle = {
+    primaryHandle match {
```
leogrim, line 59: :+1:

kifi-backend/modules/shoebox/app/com/keepit/model/FullLibraryInfo.scala
```diff
@@ -264,11 +268,38 @@ case class FullLibraryInfo(
   attr: Option[LibrarySourceAttribution] = None,
   whoCanInvite: LibraryInvitePermissions,
   modifiedAt: DateTime,
-  path: String)
+  path: String,
```
leogrim, line 271: Could we type to `LibraryPath`?

PR #16469, opened 2015-07-30
Org avatar commander no longer reuses images

ryanpbrewster:
Every time a user uploads an org avatar, it gets processed and uploaded to S3.
It is unlikely that we will be able to reuse anything from the image store
given how many degrees of freedom a cropped scaled image has.

Also changed all the subclasses of ProcessImageRequest into case classes
so we can actually match on them.

Lastly, added URL parameters to the endpoint that specify the `(x,y)` offset and the `(width, height)` of the crop. Sorry if that makes your life horrible, @cvializ 

@andrewconner @leogrim

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/commanders/ProcessedImageHelper.scala
+++ b/kifi-backend/modules/common/app/com/keepit/commanders/ProcessedImageHelper.scala
@@ -1,14 +1,12 @@
 package com.keepit.commanders
 
-import java.awt.image.BufferedImage
 import java.io._
 import java.math.BigInteger
 import java.net.URLConnection
 import java.security.MessageDigest
-import javax.imageio.ImageIO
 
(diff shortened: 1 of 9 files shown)


ryanpbrewster:
There is no greater feeling in the world than submitting a pull request with more code removed than added.

andrewconner:
Seems reasonable. @cvializ The API change for org avatars is that is all happens in one call (instead of two). Provide the `x`, `y`, `width`, and `height` as url params in the POST upload file request

ryanpbrewster:
@cvializ let me know when I should merge this. I don't want to break the (admittedly fragile) stuff we have kind of working.

cvializ:
(working on other higher-priority tasks for Orgs right now)

cvializ:
Let's do this now, since we've knocked out a lot of the MVP tasks

PR #16477, opened 2015-07-30
Made library slug optional for creating libraries

ryanpbrewster:
We autogenerate it on the backend if it is not provided

@andrewconner @jaredpetker


andrewconner:
Looks good

PR #16479, opened 2015-07-30
More logging killed

andrewconner:
Twitter social graph logs now account for a significant portion of total shoebox logs


andrewconner:
@stkem

kifi-backend/modules/common/app/com/keepit/common/zookeeper/ServiceDiscovery.scala
```diff
@@ -119,7 +119,6 @@ class ServiceDiscoveryImpl(
         }
         return true
       case Some(instance) =>
-        val msg = s"I'm not the leader since my instance is ${myInstance.get} and the leader is $instance"
         if (logMe) logLeader(msg)
```
stkem, line 122: That looks like it won't compile. (`msg` not defined)

PR #16481, opened 2015-07-30
Removing more logs

andrewconner:
@stkem @leogrim This okay?


stkem:
Yes, absolutely! This code is all but dead anyway, with facebook and linkedin not giving us connections any more.

PR #16482, opened 2015-07-30
[definitely don't merge] Tendrilizing Keeps

ryanpbrewster:
@andrewconner @leogrim @gratimax

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/model/KeepToLibrary.scala
+++ b/kifi-backend/modules/common/app/com/keepit/model/KeepToLibrary.scala
@@ -3,6 +3,9 @@ package com.keepit.model
 import com.keepit.common.db._
 import com.keepit.common.time._
 import org.joda.time.DateTime
+import play.api.libs.json.Json
+import play.api.mvc.Results.Status
+import play.api.http.Status._
 
 case class KeepToLibrary(
   id: Option[Id[KeepToLibrary]] = None,
@@ -23,3 +26,38 @@ case class KeepToLibrary(
 }
 
 object KeepToLibraryStates extends States[KeepToLibrary]
+
+sealed abstract class KeepToLibraryFail(val status: Int, val message: String) {
+  def asErrorResponse = Status(status)(Json.obj("error" -> message))
+}
+object KeepToLibraryFail {
+  case object INSUFFICIENT_PERMISSIONS extends KeepToLibraryFail(FORBIDDEN, "insufficient_permissions")
+  case object ALREADY_LINKED extends KeepToLibraryFail(BAD_REQUEST, "link_already_exists")
+  case object NOT_LINKED extends KeepToLibraryFail(BAD_REQUEST, "link_does_not_exist")
+
+  def apply(str: String): KeepToLibraryFail = {
+    str match {
+      case INSUFFICIENT_PERMISSIONS.message => INSUFFICIENT_PERMISSIONS
+      case ALREADY_LINKED.message => ALREADY_LINKED
+      case NOT_LINKED.message => NOT_LINKED
+    }
+  }
+}
+
+sealed abstract class KeepToLibraryRequest {
+  def keepId: Id[Keep]
+  def libraryId: Id[Library]
+  def requesterId: Id[User]
+}
+
+case class KeepToLibraryAttachRequest(
+  keepId: Id[Keep],
+  libraryId: Id[Library],
+  requesterId: Id[User]) extends KeepToLibraryRequest
+case class KeepToLibraryAttachResponse(link: KeepToLibrary)
+
+case class KeepToLibraryDetachRequest(
+  keepId: Id[Keep],
+  libraryId: Id[Library],
+  requesterId: Id[User]) extends KeepToLibraryRequest
+case class KeepToLibraryDetachResponse(dummy: Boolean = false)
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepInterner.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepInterner.scala
@@ -32,6 +32,7 @@ class KeepInterner @Inject() (
   normalizedURIInterner: NormalizedURIInterner,
   keepRepo: KeepRepo,
   libraryRepo: LibraryRepo,
+  keepToLibraryCommander: KeepToLibraryCommander,
   countByLibraryCache: CountByLibraryCache,
   keepToCollectionRepo: KeepToCollectionRepo,
   collectionRepo: CollectionRepo,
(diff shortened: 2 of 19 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepInterner.scala
```diff
@@ -248,6 +249,10 @@ class KeepInterner @Inject() (
           } |> { keep =>
             keepRepo.save(keep)
           }
+        keepToLibraryRepo.getByKeepIdAndLibraryId(savedKeep.id.get, library.id.get) match {
+          case None => keepToLibraryRepo.save(KeepToLibrary(keepId = savedKeep.id.get, libraryId = library.id.get, keeperId = userId))
+          case Some(ktl) =>
```
maxov: A little bit misleading (as if we'll do something with ktl), just `case _ =>` would work

Actually, I think `if (keepToLibraryRepo.getByKeepIdAndLibraryId(savedKeep.id.get, library.id.get).isEmpty) { <logic> }` would be most clear

ryanpbrewster: Haha, good suggestion. I just did exactly that. :)

kifi-backend/modules/common/app/com/keepit/model/KeepToLibrary.scala
```diff
@@ -23,3 +26,38 @@ case class KeepToLibrary(
...
+  keepId: Id[Keep],
+  libraryId: Id[Library],
+  requesterId: Id[User]) extends KeepToLibraryRequest
+case class KeepToLibraryAttachResponse(link: KeepToLibrary)
+
+case class KeepToLibraryDetachRequest(
+  keepId: Id[Keep],
+  libraryId: Id[Library],
+  requesterId: Id[User]) extends KeepToLibraryRequest
+case class KeepToLibraryDetachResponse(dummy: Boolean = false)
```
leogrim, line 63: Should we pick between "attaching" vs "linking" a keep to a library? Or is there a difference I'm missing here? 
`link` sounds a bit generic, what about just `AddKeepToLibraryRequest` and `RemoveKeepFromLibraryRequest`?

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -753,7 +754,9 @@ class LibraryCommanderImpl @Inject() (
         }
 
         // Update visibility of keeps
+        // TODO(ryan): We need to find a new way of describing keep visibility. Denormalizing is no longer possible because it's different for different users
```
leogrim, line 757: This method should go over `KeepToLibrary` as opposed to `Keep`?

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -1211,6 +1215,10 @@ class LibraryCommanderImpl @Inject() (
     }
   }
 
+  // TODO(ryan): is this actually necessary anymore? Check with Léo to see if Search needs it
+  // We may not need to have this concept of "keep ownership" anymore. You can be the author of
+  // a keep, which is its own thing. You can have access to edit a keep, which is also its own
+  // thing.
```
leogrim, line 1220: This method was only used to transfer pre-built libraries to influencers, I'm not sure how much we need it anymore.
Regardless, I think we must retain the `userId` field on a keep, in order to identify the creator / sender. Currently only a keep's creator is allowed to edit its metadata (title, note, image), breaking this assumption would introduce more issues in our current design (who is a keep attributed to when someone other than its creator modifies it?).

kifi-backend/modules/common/app/com/keepit/model/KeepToLibrary.scala
ryanpbrewster, line 63: The verbiage in my head makes sense, as you `attach` two things together, thus generating a `link` between the two. Later you can `detach` those two things, breaking the `link`.

leogrim, line 63: Well you could say that of pretty much any database relationship, it
doesn't mean we always use "attach" and "link". Should we be more business
specific?

ryanpbrewster, line 63: Is your objection specifically to the word `link`?

`attach` and `detach` already take specific requests (`KeepToLibraryAttachRequest` and `KeepToLibraryDetachRequest` respectively). I like being very explicit with the request/response types, but for the method names themselves I'm not as particular.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
ryanpbrewster, line 757: There's a tradeoff. We could just always look up the library to find the library's `visibility`, or we could denormalize the `visibility` onto the `KeepToLibrary` pin. I imagine that the place this is going to be most important is search, so I'll let you dictate this choice.

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
```diff
@@ -365,32 +365,7 @@ class KeepsCommander @Inject() (
...
-    val (mainLib, secretLib) = libraryCommander.getMainAndSecretLibrariesForUser(keep.userId)
-    def getLibFromPrivacy(isPrivate: Boolean) = {
-      if (isPrivate) Some(secretLib.id.get) else Some(mainLib.id.get)
-    }
-    if (isPrivate.isDefined && isPrivate.get != keep.isPrivate) {
-      keepRepo.save(keep.copy(visibility = Keep.isPrivateToVisibility(isPrivate.get), libraryId = getLibFromPrivacy(isPrivate.get)).withTitle(updatedTitle))
-    } else {
-      keepRepo.save(keep.withTitle(updatedTitle))
-    }
-  }
```
ryanpbrewster, line 392: This code was used to migrate from the `public/private` paradigm to the Library-based one (it moves `public -> SYSTEM_MAIN` and `private -> SYSTEM_SECRET`). No longer needed.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
leogrim, line 757: We definitely need to denormalize visibility / orgId / userId from library
onto KeepToLibrary, just like we do today on keep :)

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
leogrim, line 1220: @andrewconner

PR #16485, opened 2015-07-30
Feature/org lib priv

cvializ:
@adamjgrant @andrewconner Org privacy FTW


kifi-backend/modules/shoebox/angular/src/libraries/libraryVisibilitySelector.js
```diff
@@ -0,0 +1,27 @@
...
+        library: '=',
+        space: '='
+      },
+      link: function (scope, element) {
+        scope.spaceIsOrg = function () {
+          return 'numMembers' in scope.space;
+        };
+
+        $timeout(function () {
+          element.find('.library-visibility-checkbox-custom-input').addClass('kf-transitions');
```
adamjgrant, line 22: Cool.

kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.js
```diff
@@ -40,21 +40,42 @@ angular.module('kifi')
...
+              }
+            }
+          }
 
-        scope.setSpace = function(space) {
-          scope.destinationSpace = space;
+          // Prevent non-org lib from having visibility === 'organization'
+          if (!destIsOrg && scope.library.visibility === 'organization') {
+            scope.library.visibility = 'secret';
+          }
```
adamjgrant, line 78: Nice safeguard.

kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.tpl.html
```diff
@@ -15,10 +15,10 @@ <h1 class="dialog-header">
...
           <div class="manage-lib-input-wrapper" ng-if="spaces.length > 1">
             <span class="manage-lib-heading">Space</span>&nbsp;
-            <select class="manage-lib-input" 
-              ng-options="space as (space|name) for space in spaces" 
-              ng-model="thisSpace"
-              ng-change="setSpace(thisSpace)"
+            <select class="manage-lib-input"
+              ng-options="s as (s|name) for s in spaces"
+              ng-model="space.destination"
+              ng-change="onChangeSpace()"
```
adamjgrant, line 21: Do we still need the `ng-change` after implementing dot notation?

cvializ, line 21: Yes, this is an appropriate use of `ng-change`. This time, instead of doing what `ng-model` is supposed to do, it's checking for a change from a user-space to an org-space or vice-versa. Then it modifies the visibility depending on the previous values of the visibility.

adamjgrant:
Cool, :+1:

PR #16486, opened 2015-07-30
one less stack trace printed for rover

stkem:
just stumbled upon it. hope you don't have that one locally yet @andrewconner


kifi-backend/modules/shoebox/marketing
```diff
@@ -1 +1 @@
-Subproject commit 9f815ef0d22143e93c16c2238da403d89803bdc2
```
andrewconner, line 1: Don't merge.

PR #16488, opened 2015-07-30
Alternative design for profile widget

adamjgrant:
screen shot 2015-07-30 at 1 52 46 pm

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/profile/profileWidget.styl
+++ b/kifi-backend/modules/shoebox/angular/src/profile/profileWidget.styl
@@ -121,12 +121,13 @@
 
       tr
         td
-          font-size: 14px
+          font-size: 10px
+          line-height: 16px
 
           &:first-child
-            text-transform: uppercase
             padding-right: 3px
             font-weight: bold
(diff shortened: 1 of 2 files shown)


adamjgrant:
screen shot 2015-07-30 at 1 53 47 pm

adamjgrant:
screen shot 2015-07-30 at 1 56 47 pm

adamjgrant:
Going with the last one. @camhashemi @andrewconner @cvializ @ryanpbrewster

cvializ:
:+1: The last one was my fave

ryanpbrewster:
Can we reduce the `border-radius` of the bounding circle for the avatar? 50% is extreme. I honestly think it should be 0%.

adamjgrant:
@ryanpbrewster Aesthetically, the problem is actually the image being used. Because it has empty space around it, it doesn't appear as a circle. The `border-radius: 50%` is meant to render a perfect circle as it does for the user profile image.

adamjgrant:
Merging for now but we can continue discussions about the display of the org profile image and return to this.

PR #16490, opened 2015-07-30
remove the list publicly checkbox from library page

cvializ:
Remove this feature as per Eishay's and Jen's request. Still sets the value to true by default. Keeps existing libraries as they are.

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/libraryVisibilitySelector.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/libraryVisibilitySelector.js
@@ -13,14 +13,10 @@ angular.module('kifi')
         library: '=',
         space: '='
       },
-      link: function (scope, element) {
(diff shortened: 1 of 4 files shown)


cvializ:
@andrewconner @adamjgrant

kifi-backend/modules/shoebox/angular/src/libraries/libraryVisibilitySelector.js
```diff
@@ -13,14 +13,10 @@ angular.module('kifi')
         library: '=',
         space: '='
       },
-      link: function (scope, element) {
+      link: function (scope) {
         scope.spaceIsOrg = function () {
           return 'numMembers' in scope.space;
         };
-
```
cvializ, line 20: Not sure why the following was here, instead of just adding `.kf-transitions` to the element directly : /

PR #16492, opened 2015-07-30
Backend expects offset/limit, not page/size.

cvializ:
Fixes #43300794771955

(https://app.asana.com/0/5268025683587/43300794771955)


cvializ:
@andrewconner @adamjgrant @ryanpbrewster @camhashemi

camhashemi:
@cvializ @ryanpbrewster the backend can use `page` and `size`. we use it elsewhere, so unless we want to refactor all those as well, let's use `page` for pagination and `offset` for offsets.

PR #16494, opened 2015-07-30
Set up tests for Eliza websocket

maxov:
@ryanpbrewster 
@andrewconner 
@stkem 

Still a WIP...


kifi-backend/modules/eliza/test/com/keepit/eliza/ws/WebSocketTest.scala
```diff
@@ -0,0 +1,69 @@
...
+        resultsPromise.failure(new RuntimeException("Result " + result + " was not expected"))
+        resultsPromise.future
+    }
+  }
+
+  class WebSocketMatcher(m: Matcher[Seq[JsArray]], retries: Int = 0, timeout: FiniteDuration = 60.seconds)(implicit injector: Injector) extends Matcher[Seq[JsArray]] {
+
+    override def apply[N <: Seq[JsArray]](t: Expectable[N]): MatchResult[N] = {
+      m.await(retries, timeout).apply(t.map[Future[Seq[JsArray]]] { value =>
+        feedToSocket(value)
```
andrewconner: The _only_ thing that would make this sweeter is if we could avoid hard coding which websocket controller we're using. Would passing in the class constructor work?

andrewconner:
Max, you are a genius.

ryanpbrewster:
This is really nice work. Let's get some preliminary tests set up (and then see if we can break them!).

stkem:
Nice! I think I remember reading something (but now can't find it any more) about longer blocking operations being discouraged inside of an iteratee. Something to do with Play using a separate very small thread pool for iteratees. Not sure if that is still the case though. In short, `SafeFuture` is certainly not needed, but `Future` _might_ be.

ryanpbrewster:
lgtm, I think we should merge this in.

andrewconner:
This was written for the commit you reverted:

Not sure if you ever figured out a workaround for the `SafeFuture` thing, but Stephen's right that it was intentional: Interatees are evaluated FIFO, so if you have a "get all notifications" call, that actually blocks the iteratee until it resolves. Since we don't care about ordering, we immediately lift into a `Future`.

Since you manually do that for each one, it should be fine.

Only other thing: instead of the Await, could you just end the session when they complete, right?

PR #16495, opened 2015-07-30
orgLibraries: use paginator pattern instead of offset/limit

camhashemi:
@ryanpbrewster @cvializ


kifi-backend/modules/shoebox/angular/src/orgProfile/orgProfileLibrariesCtrl.js
```diff
@@ -11,7 +11,7 @@ angular.module('kifi')
 
     function librarySource(pageNumber, pageSize) {
       return orgProfileService
-        .getOrgLibraries(organization.id, pageNumber * pageSize, pageSize)
+        .getOrgLibraries(organization.id, pageNumber, pageSize)
```
camhashemi: @cvializ doing this on the backend now

ryanpbrewster:
@carlos suggested allowing either `limit/offset` or `page/pageSize` (failing with some error if they are mixed together improperly). I think that's a grand idea.

PR #16496, opened 2015-07-30
Fix rename error and make scope.space optional

cvializ:
Fixes Asana 43300794771961


cvializ:
@andrewconner @adamjgrant

adamjgrant:
Late :+1: :)

PR #16502, opened 2015-07-31
Add tracking to org member page

cvializ:
@andrewconner @adamjgrant

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/orgProfile/orgProfileInviteSearch.js
+++ b/kifi-backend/modules/shoebox/angular/src/orgProfile/orgProfileInviteSearch.js
@@ -42,8 +42,8 @@ angular.module('kifi')
           if (scope.share.message) {
             opts.message = scope.share.message;
           }
-          var cb = scope.modalData.returnAction;
-          return orgProfileService.sendOrgMemberInvite(scope.organization.id, opts).then(cb);
+
+          return orgProfileService.sendOrgMemberInvite(scope.organization.id, opts).then(scope.modalData.returnAction);
         }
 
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/shoebox/angular/src/orgProfile/orgProfileInviteSearch.js
```diff
@@ -42,8 +42,8 @@ angular.module('kifi')
           if (scope.share.message) {
             opts.message = scope.share.message;
           }
-          //return libraryService.shareLibrary(scope.library.id, opts);
-          return orgProfileService.sendOrgMemberInvite(scope.organization.id, opts);
+          var cb = scope.modalData.returnAction;
+          return orgProfileService.sendOrgMemberInvite(scope.organization.id, opts).then(cb);
```
adamjgrant: why assign this to `cb`?

adamjgrant: I mean, as opposed to just putting the function in the `then`. Seems to go against the idea behind promises.

cvializ: Hmm good point, I thought `cb` would communicate my intent more effectively, but looking at it with fresh eyes, `.then(scope.modalData.returnAction)` looks better

andrewconner:
Looks good to me

adamjgrant:
Late to the party but good stuff :+1:

cvializ:
It's ok, there wasn't any :cake: so you didn't miss much

PR #16503, opened 2015-07-31
Removing onboarding + forwards compatibility of library.path

andrewconner:
@adamjgrant @cvializ 

Ryan wants to switch from `library.url` to `library.path`, the JSON field is already there. This PR does two things:

1) Switches to `library.url || library.path` in case it's not everywhere yet.
2) Removes the onboarding for Firefox, where it's currently broken :cry:. Whenever we have some time, one or both of you can play around with the extension and figure this one out.


adamjgrant:
Cool. I haven't actually touched the extension yet, have you @cvializ ?

cvializ:
That's a negative, haven't seen it

PR #16505, opened 2015-07-31
adding primary org id to the user event when reporting to our analyti…

eishay:
…cs services


eishay:
@andrewconner @joshm1

PR #16508, opened 2015-07-31
Refactor Email verification

leogrim:
@andrewconner


kifi-backend/modules/common/conf/evolutions/shoebox/364.sql
```diff
@@ -0,0 +1,9 @@
+# SHOEBOX
+
+# --- !Ups
+
+ALTER TABLE password_reset MODIFY COLUMN sent_to varchar(512) NOT NULL;
+
+insert into evolutions(name, description) values('364.sql', 'modify password_reset.sent_to to varchar(512) NOT NULL');
+
+# --- !Downs
```
leogrim, line 9: Applied already.

kifi-backend/modules/shoebox/app/com/keepit/controllers/core/AuthHelper.scala
```diff
@@ -449,8 +449,8 @@ class AuthHelper @Inject() (
     } yield {
       db.readWrite { implicit s =>
         passwordResetRepo.getByToken(code) match {
-          case Some(pr) if passwordResetRepo.tokenIsNotExpired(pr) =>
-            val email = passwordResetRepo.useResetToken(code, request.headers.get("X-Forwarded-For").getOrElse(request.remoteAddress))
+          case Some(pr) if passwordResetRepo.useResetToken(code, request.headers.get("X-Forwarded-For").getOrElse(request.remoteAddress)) =>
+            emailAddressRepo.getByAddress(pr.sentTo).foreach(userEmailAddressCommander.saveAsVerified(_)) // mark email address as verified
```
leogrim, line 453: @andrewconner I noticed that the result of `useResetToken` was not checked, and that the logic in there also checked for token expiration. So replacing `tokenIsNotExpired` with `useResetToken` in the case conditional seemed to kill both birds. 

Also, I added the email verification part here as opposed to `AuthController.setPasswordPage` (which is when we know the click happened), because this is where the token is actually consumed. What do you think?

andrewconner, line 453: Works for me

andrewconner:
Looks good

PR #16513, opened 2015-07-31
Redirect to the departed library's space page after deletion

cvializ:
Behold, @adamjgrant @andrewconner, a post-delete redirect of utmost propriety


adamjgrant:
Nice!

PR #16514, opened 2015-07-31
add confirmation for library deletion

camhashemi:
@andrewconner is this good enough?

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
@@ -826,12 +826,15 @@ class LibraryCommanderImpl @Inject() (
       }
       libraryAnalytics.deleteLibrary(userId, oldLibrary, context)
       libraryAnalytics.unkeptPages(userId, savedKeeps.keySet.toSeq, oldLibrary, context)
+      searchClient.updateKeepIndex()
       //Note that this is at the end, if there was an error while cleaning other library assets
       //we would want to be able to get back to the library and clean it again
-      db.readWrite { implicit s =>
+      db.readWrite(attempts = 2) { implicit s =>
         libraryRepo.save(oldLibrary.sanitizeForDelete)
       }
-      searchClient.updateKeepIndex()
+      db.readOnlyMaster { implicit s =>
+        require(libraryRepo.get(oldLibrary.id.get).state == LibraryStates.INACTIVE, s"Library ${oldLibrary.id.get} deletion failed")
+      }
       searchClient.updateLibraryIndex()
       None
     }


kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -826,12 +826,15 @@ class LibraryCommanderImpl @Inject() (
...
+      searchClient.updateKeepIndex()
       //Note that this is at the end, if there was an error while cleaning other library assets
       //we would want to be able to get back to the library and clean it again
-      db.readWrite { implicit s =>
+      db.readWrite(attempts = 2) { implicit s =>
         libraryRepo.save(oldLibrary.sanitizeForDelete)
       }
-      searchClient.updateKeepIndex()
+      db.readOnlyMaster { implicit s =>
+        require(libraryRepo.get(oldLibrary.id.get).state == LibraryStates.INACTIVE, s"Library ${oldLibrary.id.get} deletion failed")
```
andrewconner, line 836: Why an assert? Log the actual persisted model a couple lines up. If you think it's not persisting, just log what it is here. require throws an exception, guaranteeing a bad experience.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -826,12 +826,15 @@ class LibraryCommanderImpl @Inject() (
       }
       libraryAnalytics.deleteLibrary(userId, oldLibrary, context)
       libraryAnalytics.unkeptPages(userId, savedKeeps.keySet.toSeq, oldLibrary, context)
+      searchClient.updateKeepIndex()
       //Note that this is at the end, if there was an error while cleaning other library assets
       //we would want to be able to get back to the library and clean it again
-      db.readWrite { implicit s =>
+      db.readWrite(attempts = 2) { implicit s =>
         libraryRepo.save(oldLibrary.sanitizeForDelete)
```
andrewconner, line 833: `tap { l => log.info(s"Deleted lib: $l")}`

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -826,12 +826,15 @@ class LibraryCommanderImpl @Inject() (
       }
       libraryAnalytics.deleteLibrary(userId, oldLibrary, context)
       libraryAnalytics.unkeptPages(userId, savedKeeps.keySet.toSeq, oldLibrary, context)
+      searchClient.updateKeepIndex()
       //Note that this is at the end, if there was an error while cleaning other library assets
       //we would want to be able to get back to the library and clean it again
```
andrewconner, line 831: `log.info("Deleting lib: $oldLibrary")`

PR #16517, opened 2015-07-31
Kill disjoint libs

ryanpbrewster:
@andrewconner @stkem @leogrim

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/model/Keep.scala
+++ b/kifi-backend/modules/common/app/com/keepit/model/Keep.scala
@@ -19,7 +19,6 @@ case class Keep(
     title: Option[String] = None,
     uriId: Id[NormalizedURI],
     isPrimary: Boolean = true, // trick to let us have multiple inactive Keeps while keeping integrity constraints
-    inDisjointLib: Boolean,
     urlId: Id[URL],
     url: String, // denormalized for efficiency
     visibility: LibraryVisibility, // denormalized from this keep’s library
@@ -87,10 +86,10 @@ object Keep {
   }
 
   def applyFromDbRowTuples(firstArguments: KeepFirstArguments, restArguments: KeepRestArguments): Keep = (firstArguments, restArguments) match {
-    case ((id, createdAt, updatedAt, externalId, title, uriId, isPrimary, inDisjointLib, urlId, url),
+    case ((id, createdAt, updatedAt, externalId, title, uriId, isPrimary, urlId, url),
       (isPrivate, userId, state, source, kifiInstallation, seq, libraryId, visibility, keptAt, sourceAttributionId, note, originalKeeperId, organizationId)) =>
       _applyFromDbRow(id, createdAt, updatedAt, externalId, title,
-        uriId = uriId, isPrivate = isPrivate, isPrimary = isPrimary, inDisjointLib = inDisjointLib, urlId = urlId, url = url,
+        uriId = uriId, isPrivate = isPrivate, isPrimary = isPrimary, urlId = urlId, url = url,
         userId = userId, state = state, source = source, kifiInstallation = kifiInstallation,
         seq = seq, libraryId = libraryId, visibility = visibility, keptAt = keptAt,
         sourceAttributionId = sourceAttributionId, note = note, originalKeeperId = originalKeeperId,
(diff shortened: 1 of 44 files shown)


kifi-backend/modules/common/app/com/keepit/model/Keep.scala
```diff
@@ -129,7 +128,6 @@ object Keep {
       (__ \ 'title).readNullable[String] and
       (__ \ 'uriId).read(Id.format[NormalizedURI]) and
       (__ \ 'isPrimary).readNullable[Boolean] and
-      (__ \ 'inDisjointLib).readNullable[Boolean] and
```
ryanpbrewster, line 132: I _think_ this means that internal services will not be affected by this. They will simply assume that all the keep they're reading in are from non-disjoint libraries.

kifi-backend/modules/common/conf/evolutions/shoebox/363.sql
```diff
@@ -0,0 +1,10 @@
+# SHOEBOX
+
+# --- !Ups
+
+ALTER TABLE bookmark ALTER COLUMN in_disjoint_lib tinyint(1) DEFAULT NULL;
+-- DROP INDEX `user_disjoint_primary_uri` ON bookmark;
```
ryanpbrewster: I want to drop this index, but I get evolution errors if I uncomment this line. Any ideas why? (Is it important that it was created as a "KEY" instead of an "INDEX"?)

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepInterner.scala
```diff
@@ -226,12 +222,6 @@ class KeepInterner @Inject() (
         val kNote = note orElse { if (wasInactiveKeep) None else bookmark.note }
         val kTitle = trimmedTitle orElse { if (wasInactiveKeep) None else bookmark.title } orElse uri.title
 
-        if (bookmark.isActive && bookmark.inDisjointLib && bookmark.libraryId.get != library.id.get) {
-          // invalidate if keep URI is active in a system library
-          countByLibraryCache.remove(CountByLibraryKey(bookmark.libraryId.get))
-          countByLibraryCache.remove(CountByLibraryKey(library.id.get))
-        }
```
ryanpbrewster, line 233: Can someone else double-check that this is no longer necessary?

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -1354,32 +1354,22 @@ class LibraryCommanderImpl @Inject() (
...
-                val newKeep = keepRepo.save(existingKeep.copy(userId = userId, libraryId = Some(toLibraryId), visibility = toLibrary.visibility,
-                  inDisjointLib = toLibrary.isDisjoint, source = withSource.getOrElse(k.source), state = KeepStates.ACTIVE))
-                combineTags(k.id.get, existingKeep.id.get)
-                Right(newKeep)
-              } else {
-                combineTags(k.id.get, existingKeep.id.get)
-                Left(LibraryError.AlreadyExistsInDest)
-              }
+              combineTags(k.id.get, existingKeep.id.get)
+              Left(LibraryError.AlreadyExistsInDest)
```
ryanpbrewster, line 1375: It feels weird to me to mutate state (`combineTags`) and _then_ return an error. If we're going to yield an error, shouldn't we avoid any side-effects?

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -1403,29 +1393,22 @@ class LibraryCommanderImpl @Inject() (
...
             case Some(existingKeep) =>
               if (toLibraryId == k.libraryId.get) {
                 Left(LibraryError.AlreadyExistsInDest)
-              } else if (existingKeep.inDisjointLib) {
-                val newKeep = keepRepo.save(existingKeep.copy(libraryId = Some(toLibraryId), visibility = toLibrary.visibility, inDisjointLib = toLibrary.isDisjoint, state = KeepStates.ACTIVE))
-                combineTags(k.id.get, existingKeep.id.get)
-                Right(newKeep)
               } else {
                 keepRepo.save(k.copy(state = KeepStates.INACTIVE))
                 combineTags(k.id.get, existingKeep.id.get)
```
ryanpbrewster, line 1417: Same deal, shouldn't we avoid doing anything if we are going to yield an error?

kifi-backend/modules/shoebox/app/com/keepit/model/KeepRepo.scala
```diff
@@ -201,7 +198,7 @@ class KeepRepoImpl @Inject() (
     keepUriUserCache.getOrElseOpt(KeepUriUserKey(uriId, userId)) {
       val keeps = (for (b <- rows if b.uriId === uriId && b.userId === userId && b.isPrimary === true && b.state === KeepStates.ACTIVE) yield b).list
       if (keeps.length > 1) log.warn(s"[getByUriAndUser] ${keeps.length} keeps found for (uri, user) pair ${(uriId, userId)}")
-      keeps.find(_.inDisjointLib).orElse(keeps.headOption) // order: disjoint, custom
+      keeps.headOption
```
ryanpbrewster, line 201: This method is still called. I didn't check where. I can add functionality to order the keeps such that `SYSTEM_MAIN` and `SYSTEM_PRIVATE` are prioritized, or we can do really funky stuff (like using the `LibraryMembership.priority` field to decide which library the user cares the most about).

My preference would be to just return a `Set[Keep]` instead of an `Option[Keep]`, though.

kifi-backend/modules/common/conf/evolutions/shoebox/363.sql
stkem: INDEX vs KEY should not matter. What is the error you get? Could be that it is used as a foreign key somewhere?

kifi-backend/modules/common/app/com/keepit/model/Keep.scala
stkem, line 132: Yup, the deserialization (`inDisjointLib.exists(b => b)`) treats `Some(false)` and `None` the same. I think you are safe.

stkem:
lgtm overall. I'm not to most qualified for this code though.

kifi-backend/modules/common/conf/evolutions/shoebox/363.sql
ryanpbrewster: I was using the wrong syntax. MySQL is a strange beast.

PR #16522, opened 2015-07-31
Fix gear icon not appearing

cvializ:
@adamjgrant @andrewconner


kifi-backend/modules/shoebox/angular/src/orgProfile/orgProfileLibraries.tpl.html
```diff
@@ -4,6 +4,6 @@
       <span class="kf-upl-create-card-add-icon svg-create-library"></span>
       <div class="kf-upl-create-card-text">Create a Library</div>
     </button>
-    <div kf-user-profile-library-card="::library" ng-repeat="library in libraries"></div>
+    <div kf-user-profile-library-card="::library" ng-repeat="library in libraries" profile-id="{{::userProfile.id}}"></div>
```
adamjgrant, line 7: Where does this get used?

cvializ, line 7: when deciding whether or not to show the gear icon haha

PR #16525, opened 2015-07-31
log library deletion

camhashemi:
@andrewconner thank you

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
@@ -829,11 +829,13 @@ class LibraryCommanderImpl @Inject() (
       searchClient.updateKeepIndex()
       //Note that this is at the end, if there was an error while cleaning other library assets
       //we would want to be able to get back to the library and clean it again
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -829,11 +829,13 @@ class LibraryCommanderImpl @Inject() (
...
       //Note that this is at the end, if there was an error while cleaning other library assets
       //we would want to be able to get back to the library and clean it again
+      log.info(s"Deleting lib: $oldLibrary")
       db.readWrite(attempts = 2) { implicit s =>
         libraryRepo.save(oldLibrary.sanitizeForDelete)
+          .tap { l => log.info(s"Deleted lib: $s") }
       }
       db.readOnlyMaster { implicit s =>
-        require(libraryRepo.get(oldLibrary.id.get).state == LibraryStates.INACTIVE, s"Library ${oldLibrary.id.get} deletion failed")
+        log.info(s"Confirming lib deletion: $libraryRepo.get(oldLibrary.id.get)")
```
ryanpbrewster: Why not check to see if it was deleted and only if it was not do you log it? Maybe a `log.warn`?

PR #16527, opened 2015-07-31
Implementing new rules for showing single owner

adamjgrant:


adamjgrant:
@cvializ @camhashemi @cvializ @ryanpbrewster

adamjgrant:
For keeps only editable by their owner...

screen shot 2015-07-31 at 4 31 21 pm

Profile image shows when not you, outside of a recommendation.

screen shot 2015-07-31 at 4 31 11 pm

Profile image hides within an org, showing org image instead.

screen shot 2015-07-31 at 4 29 49 pm

Profile image hides if you are the owner.

adamjgrant:
Updated to use "with" consistently. It may not always be clear that an image is of an org.

screen shot 2015-07-31 at 4 37 46 pm

cvializ:
simple, I like it

PR #16528, opened 2015-07-31
More logging deaths.

andrewconner:
@eishay @stkem


andrewconner:
Shoebox blew up again because of wait time errors flooding the healthcheck, which just _duplicates_ errors.

PR #16532, opened 2015-08-01
Not killing ourselves when zk blows up

andrewconner:
@stkem @eishay @leogrim


eishay:
The debounce is nice. I assume we're masking the real problem, right? 
Should we restart the zk client (`refreshSession`) when few exact events are coming in a tight loop ?
Should the debounce come only if there is an identical log message in the last sliding time window?

PR #16533, opened 2015-08-02
added amplitude setUserProperties and alias methods; heimdal refactored to enable amplitude tests

joshm1:
also skips recording pingdom-caused events

@andrewconner

### DIFF
--- a/kifi-backend/modules/heimdal/app/com/keepit/dev/HeimdalDevModule.scala
+++ b/kifi-backend/modules/heimdal/app/com/keepit/dev/HeimdalDevModule.scala
@@ -1,11 +1,10 @@
 package com.keepit.dev
 
+import com.keepit.common.cache.{ HashMapMemoryCacheModule, HeimdalCacheModule }
 import com.keepit.common.controller.DevRemoteUserActionsHelperModule
-import com.keepit.heimdal._
+import com.keepit.heimdal.{ DevDelightedModule, HeimdalQueueDevModule, ProdAnalyticsModule, HeimdalModule }
 import com.keepit.helprank.DevReKeepStatsUpdaterModule
 import com.keepit.inject.CommonDevModule
-import com.keepit.common.cache.HashMapMemoryCacheModule
-import com.keepit.common.cache.HeimdalCacheModule
 
 case class HeimdalDevModule() extends HeimdalModule(
   userActionsModule = DevRemoteUserActionsHelperModule(),
--- a/kifi-backend/modules/heimdal/app/com/keepit/dev/HeimdalDevModule.scala
+++ b/kifi-backend/modules/heimdal/app/com/keepit/dev/HeimdalDevModule.scala
@@ -1,7 +1,7 @@
 package com.keepit.dev
 
 import com.keepit.common.controller.DevRemoteUserActionsHelperModule
-import com.keepit.heimdal.{ DevAnalyticsModule, DevDelightedModule, HeimdalQueueDevModule, HeimdalModule }
+import com.keepit.heimdal._
 import com.keepit.helprank.DevReKeepStatsUpdaterModule
 import com.keepit.inject.CommonDevModule
 import com.keepit.common.cache.HashMapMemoryCacheModule
@@ -13,6 +13,7 @@ case class HeimdalDevModule() extends HeimdalModule(
   analyticsModule = DevAnalyticsModule(),
   heimdalQueueModule = HeimdalQueueDevModule(),
   rekeepStatsUpdaterModule = DevReKeepStatsUpdaterModule(),
-  delightedModule = DevDelightedModule()
+  delightedModule = DevDelightedModule(),
+  amplitudeTransportModule = DevAmplitudeTransportModule()
 ) with CommonDevModule {
 }
--- a/kifi-backend/modules/heimdal/app/com/keepit/dev/HeimdalDevModule.scala
+++ b/kifi-backend/modules/heimdal/app/com/keepit/dev/HeimdalDevModule.scala
@@ -10,10 +10,9 @@ import com.keepit.common.cache.HeimdalCacheModule
 case class HeimdalDevModule() extends HeimdalModule(
   userActionsModule = DevRemoteUserActionsHelperModule(),
   cacheModule = HeimdalCacheModule(HashMapMemoryCacheModule()),
-  analyticsModule = DevAnalyticsModule(),
+  analyticsModule = ProdAnalyticsModule(),
   heimdalQueueModule = HeimdalQueueDevModule(),
   rekeepStatsUpdaterModule = DevReKeepStatsUpdaterModule(),
-  delightedModule = DevDelightedModule(),
-  amplitudeTransportModule = DevAmplitudeTransportModule()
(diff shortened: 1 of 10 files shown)


kifi-backend/modules/heimdal/app/com/keepit/dev/HeimdalDevModule.scala
```diff
@@ -10,7 +10,7 @@ import com.keepit.common.cache.HeimdalCacheModule
 case class HeimdalDevModule() extends HeimdalModule(
   userActionsModule = DevRemoteUserActionsHelperModule(),
   cacheModule = HeimdalCacheModule(HashMapMemoryCacheModule()),
-  analyticsModule = DevAnalyticsModule(),
+  analyticsModule = ProdAnalyticsModule(),
```
stkem, line 12: Why?

kifi-backend/modules/heimdal/app/com/keepit/dev/HeimdalDevModule.scala
```diff
@@ -1,7 +1,7 @@
 package com.keepit.dev
 
 import com.keepit.common.controller.DevRemoteUserActionsHelperModule
-import com.keepit.heimdal.{ DevAnalyticsModule, DevDelightedModule, HeimdalQueueDevModule, HeimdalModule }
+import com.keepit.heimdal._
```
stkem: _cringe_ Intellij? :-)

kifi-backend/modules/heimdal/app/com/keepit/heimdal/AmplitudeClient.scala
```diff
@@ -17,9 +18,16 @@ object AmplitudeClient {
     "kcid_9", "kcid_10", "kcid_11", "os", "osVersion", "remoteAddress", "serviceInstance", "serviceZone", "userCreatedAt",
     "userId", "userSegment")
 
-  val killedEvents = Set("user_old_slider_sliderShown", "user_expanded_keeper", "user_used_kifi", "user_reco_action",
-    "anonymous_kept", "user_logged_in", "visitor_expanded_keeper", "visitor_reco_action", "visitor_viewed_notification",
-    "visitor_clicked_notification", "anonymous_visitor_viewed_page", "anonymous_messaged")
+  private val killedEvents = Set("user_old_slider_sliderShown", "user_expanded_keeper", "user_used_kifi", "user_reco_action",
```
stkem, line 21: I have no context, so jus curious, but why are all the anonymous events gone from this?

kifi-backend/modules/heimdal/app/com/keepit/heimdal/PrimaryOrgProvider.scala
```diff
@@ -9,9 +9,13 @@ import com.keepit.common.cache.TransactionalCaching.Implicits.directCacheAccess
 
 import scala.concurrent.Future
 
-class PrimaryOrgProvider @Inject() (
+trait PrimaryOrgProvider {
```
stkem, line 13: What is the point of this thing? Why not use the Service client directly?

kifi-backend/modules/heimdal/app/com/keepit/heimdal/AmplitudeClient.scala
```diff
@@ -30,52 +38,165 @@ object AmplitudeClient {
...
+  def track[E <: HeimdalEvent](event: E)(implicit companion: HeimdalEventCompanion[E]): Future[AmplitudeEventResult] = {
     val eventBuilder = new AmplitudeEventBuilder(event, primaryOrgProvider)
 
-    if (AmplitudeClient.killedEvents.contains(eventBuilder.getEventType())) {
-      Future.successful(Unit)
-    } else new SafeFuture({
-      eventBuilder.buildEventPayload() map { payload =>
-        sendData(payload) map (_ => ())
+    if (AmplitudeClient.skipEventFilters.exists(fn => fn(eventBuilder))) Future.successful(AmplitudeEventSkipped(eventBuilder.getEventType()))
+    else new SafeFuture({
```
stkem, line 59: We haven't always done this, but I've come to the conclusion that iff you are using a `SafeFuture` you should give it a name (second constructor argument). Makes for much better debugging later on.

stkem:
made a few comments. lgtm overall

kifi-backend/modules/heimdal/app/com/keepit/dev/HeimdalDevModule.scala
joshm1, line 12: `DevAnalyticsModule` was replaced with `FakeAnalyticsModule` which I've only included in the test source code.  I moved it to the test code because it depends on `FakeWebService` which only exists for tests

kifi-backend/modules/heimdal/app/com/keepit/dev/HeimdalDevModule.scala
joshm1: yep, sorry, I agree with you on this one. :)

kifi-backend/modules/heimdal/app/com/keepit/heimdal/AmplitudeClient.scala
joshm1, line 21: a few lines below this I added a filter to exclude `all` anonymous events. anonymous events can't be recorded until we decide what device_id to send for them because Amplitude doesn't like device_id=anonymous.  This was done so that if future anonymous events are added they don't have to be manually added to this list.

kifi-backend/modules/heimdal/app/com/keepit/heimdal/AmplitudeClient.scala
joshm1, line 59: :+1:

kifi-backend/modules/heimdal/app/com/keepit/heimdal/PrimaryOrgProvider.scala
joshm1, line 13: I believe @eishay created the original `PrimaryOrgProvider`, I only refactored this to make it possible to mock `PrimaryOrgProvider` in tests.  I wanted to be able to enable real http requests to amplitude in these tests (turned off by default) but still stub the request to shoebox here.

In an earlier implementation, I had issues (details which I've forgotten now) just including `FakeShoeboxServiceClientModule`, which caused me to mock this directly, but some refactoring since then allowed me to just include `FakeShoeboxServiceModule` in the tests instead and drop the provider

joshm1:
thanks for the feedback @stkem, I updated everything you commented on

kifi-backend/modules/heimdal/app/com/keepit/heimdal/AmplitudeClient.scala
stkem, line 21: :+1:

kifi-backend/modules/heimdal/app/com/keepit/dev/HeimdalDevModule.scala
stkem, line 12: :+1:

stkem:
:+1:

PR #16534, opened 2015-08-03
Fix issue in getByArticleInfo

leogrim:
@andrewconner @stkem 
I found _a_ bug but I don't think it's _the_ bug that caused the PagerDuty. There is no `inactive` `ArticleImage` in the database right now (but I can't remember whether the row I deleted to fix the issue was, seems unlikely though).

### DIFF
--- a/kifi-backend/modules/rover/app/com/keepit/rover/model/ArticleInfoHelper.scala
+++ b/kifi-backend/modules/rover/app/com/keepit/rover/model/ArticleInfoHelper.scala
@@ -19,7 +19,7 @@ class ArticleInfoHelper @Inject() (articleInfoRepo: ArticleInfoRepo, articleImag
             if (articleInfo.uriId == uriId) articleInfo else {
               airbrake.notify(s"Found ArticleInfo $kind for url $url with inconsistent uriId: expected $uriId, found ${articleInfo.uriId}.")
               articleInfoRepo.deleteCache(articleInfo)
-              articleImageRepo.getByArticleInfo(articleInfo).foreach(articleImage => articleImageRepo.save(articleImage.copy(uriId = uriId)))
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/rover/app/com/keepit/rover/model/ArticleInfoHelper.scala
```diff
@@ -19,7 +19,7 @@ class ArticleInfoHelper @Inject() (articleInfoRepo: ArticleInfoRepo, articleImag
             if (articleInfo.uriId == uriId) articleInfo else {
               airbrake.notify(s"Found ArticleInfo $kind for url $url with inconsistent uriId: expected $uriId, found ${articleInfo.uriId}.")
               articleInfoRepo.deleteCache(articleInfo)
-              articleImageRepo.getByArticleInfo(articleInfo).foreach(articleImage => articleImageRepo.save(articleImage.copy(uriId = uriId)))
+              articleImageRepo.getByArticleInfo(articleInfo, excludeState = None).foreach(articleImage => articleImageRepo.save(articleImage.copy(uriId = uriId)))
               articleInfoRepo.save(articleInfo.copy(uriId = uriId))
```
leogrim, line 23: @stkem @andrewconner 
Can you see an issue with this code? We broke the foreign key constraint from `ArticleImage` to `ArticleInfo` as we were trying to update `uriId` in both places. This is exactly what would have happened if the `ArticleImage` row was `inactive`, but I don't think that was the case (there isn't any right now).

stkem, line 23: Seems save to me. But as you said doesn't seem to be _the_ problem.

PR #16536, opened 2015-08-03
Put the org in the space selector when adding a lib from an org profile.

cvializ:
@adamjgrant this should do it 8 )

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.js
@@ -80,7 +80,9 @@ angular.module('kifi')
         // to make the angular <select> binding work
         scope.spaces = profileService.me.orgs ? [profileService.me].concat(profileService.me.orgs) : profileService.me;
         scope.space = {};
-        scope.space.current = scope.spaces.filter(function (s) { return s.id === currentSpace.id; }).pop();
+
+        var desiredId = scope.modalData.organization ? scope.modalData.organization.id : currentSpace.id;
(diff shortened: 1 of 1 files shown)


adamjgrant:
Great. Nice and semantic.

kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.js
```diff
@@ -80,7 +80,9 @@ angular.module('kifi')
         // to make the angular <select> binding work
         scope.spaces = profileService.me.orgs ? [profileService.me].concat(profileService.me.orgs) : profileService.me;
         scope.space = {};
-        scope.space.current = scope.spaces.filter(function (s) { return s.id === currentSpace.id; }).pop();
+
+        var desiredId = scope.modalData.organization ? scope.modalData.organization.id : currentSpace.id;
```
adamjgrant, line 84: Actually, without testing this code, it looks like `currentSpace` should be `profileService` no?

adamjgrant, line 84: Semantically, `currentSpace` should always be the space you're looking at.

cvializ, line 84: To close the PR loop: I agree, the naming is confusing, but it's needed to make the objects reference-equal for the `ng-model` and `ng-options` to work together

adamjgrant, line 84: @others: @cvializ and I chatted about this and agreed we needed to do this for Angular to use the same (exact) object to bind the select.

PR #16537, opened 2015-08-03
Remove unnecessary then clause for profile promise

cvializ:
@adamjgrant I saw this and the useless `.then` callback that just returns its argument was bothering me.


adamjgrant:
Nice.

PR #16538, opened 2015-08-03
Default to sorting by library name

ryanpbrewster:
For now. We ought to let the client choose what to sort by, and I have
thoughts about ways to make that pretty. The way it is currently done for
user profile libraries is...pretty hacky. I think there is a nice
clean way to do this.

@leogrim @camhashemi

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/model/LibraryRepo.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/model/LibraryRepo.scala
@@ -455,30 +455,22 @@ class LibraryRepoImpl @Inject() (
   }
 
   // Organization Library Repo methods
-  // TODO(ryan): I hate this method, and it's ugly code-duplication. If someone can fix it I will be so happy.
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/app/com/keepit/model/LibraryRepo.scala
```diff
@@ -455,30 +455,22 @@ class LibraryRepoImpl @Inject() (
...
-      } yield lib
-    } else {
-      for {
-        lib <- rows if lib.state === LibraryStates.ACTIVE && lib.orgId === orgId && (
-          (lib.visibility === (LibraryVisibility.PUBLISHED: LibraryVisibility)) ||
-          lib.id.inSet(viewerLibraryMemberships)
-        )
-      } yield lib
-    }
+  private def visibleOrganizationLibrariesHelper(orgId: Id[Organization], includeOrgVisibleLibraries: Column[Boolean], viewerLibraryMemberships: Set[Id[Library]])(implicit session: RSession) = {
```
camhashemi, line 458: please specify the return type

camhashemi:
:+1: would be great to sort by name, date created, date updated, maybe visibility (private w/ no members, private w/ members, org private, public)

ryanpbrewster:
Agreed, it's sort of low priority for me but would definitely be helpful.

PR #16539, opened 2015-08-03
Why do we log this?

andrewconner:
We were logging several MB of XML every time someone imported contacts.


andrewconner:
@stkem

andrewconner:
I'm sure you'll have no problems with this one.

stkem:
:+1:

PR #16541, opened 2015-08-03
Implementing split dropdown menu only for users who are members of or…

adamjgrant:


adamjgrant:
screen shot 2015-08-03 at 12 34 25 pm
screen shot 2015-08-03 at 12 34 34 pm

adamjgrant:
For users not in orgs, there will be no change. If you are in an org, you'll use the split design where you can quickly jump to your profile or an org.

adamjgrant:
@cvializ @camhashemi @andrewconner @ryanpbrewster

andrewconner:
Works for me

kifi-backend/modules/shoebox/angular/src/common/classes.styl
```diff
@@ -11,6 +11,21 @@
...
+  &.kf-dropdown-menu_split
+    display: flex
+    align-content: stretch
+    padding-top: 0
+    padding-bottom: 0
+
+    .kf-dropdown-menu-nav a
+      &:first-child
+      &:only-child
+        margin-top: 5px
```
cvializ, line 23: I'm a little confused here. Do you need to match the `:only-child` selector? If there's only one element, isn't it both the first and last child anyway?

adamjgrant, line 23: Yes, I thought I was being clever here but I was mixing it up with something else.

Might have been a force of habit from implementations where `last-child` took precedence over `first-child`, but only if they have the same properties.

adamjgrant, line 23: Good catch.

cvializ:
Late to the rating party, but this looks :+1:

PR #16545, opened 2015-08-03
Adding org attribution to library headers.

adamjgrant:


adamjgrant:
@cvializ @andrewconner @ryanpbrewster @camhashemi

cvializ:
Looks good to me!

PR #16548, opened 2015-08-03
adding some backend admin endpoints to orgs

eishay:
@ryanpbrewster @camhashemi

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/admin/AdminOrganizationController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/admin/AdminOrganizationController.scala
@@ -3,13 +3,10 @@ package com.keepit.controllers.admin
 import java.util.concurrent.atomic.AtomicInteger
 
 import com.google.inject.Inject
-import com.keepit.classify.Domain
-import com.keepit.common.core.futureExtensionOps
 import com.keepit.commanders._
 import com.keepit.common.controller._
 import com.keepit.common.crypto.PublicIdConfiguration
 import com.keepit.common.db._
-import com.keepit.common.db.slick.DBSession.RSession
 import com.keepit.common.db.slick.Database
 import com.keepit.model._
 import play.api.libs.json.Json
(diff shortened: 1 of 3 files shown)


eishay:
@ryanpbrewster

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminOrganizationController.scala
```diff
@@ -240,6 +237,19 @@ class AdminOrganizationController @Inject() (
     }
   }
 
+  def setOwner(orgId: Id[Organization]) = AdminUserPage { implicit request =>
+    val newOwnerId = Id[User](request.body.asFormUrlEncoded.get.apply("candidate-id").head.toLong)
+    val org = db.readOnlyReplica { implicit s => orgRepo.get(orgId) }
+    val oldOwnerId = org.ownerId
+    orgCommander.transferOrganization(OrganizationTransferRequest(oldOwnerId, orgId, newOwnerId))
+    val updatedOrg = db.readOnlyReplica { implicit s => orgRepo.get(orgId) }
```
ryanpbrewster, line 245: Should maybe be `readOnlyMaster`.

ryanpbrewster:
:+1: lgtm

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminOrganizationController.scala
```diff
@@ -240,6 +237,19 @@ class AdminOrganizationController @Inject() (
...
+    assume(updatedOrg.ownerId == newOwnerId)
+    if (oldOwnerId == fakeOwnerId) {
+      orgMembershipCommander.removeMembership(OrganizationMembershipRemoveRequest(orgId, requesterId = newOwnerId, targetId = oldOwnerId))
+    }
```
ryanpbrewster, line 249: Can we add a comment here explaining why this code exists? I feel like this is forcing two actions together in a potentially confusing way. I'd rather not have knowledge locked up in our heads when it could be in the codebase.

camhashemi, line 249: @ryanpbrewster naming it `transferOwner` would make it clearer, yeah?

eishay, line 249: oh, sorry, accidentally merged without reviewing the notes, will fix in the next one

PR #16550, opened 2015-08-03
Public org libraries work

andrewconner:
@ryanpbrewster @adamjgrant Side benefit: loading libraries is faster now.


andrewconner:
@cvializ  too!

cvializ:
beautimous

adamjgrant:
Excellent.

PR #16554, opened 2015-08-03
Add instructions for moving libraries on empty org lib page.

cvializ:
@andrewconner, @adamjgrant WIP to have something out there. This also seems to resolve the "Create Library card is missing" bug


adamjgrant:
:+1:

PR #16556, opened 2015-08-04
If user invited to org, add org experiment to user immediately

eishay:
https://app.asana.com/0/42816489233171/44052516696612/f


eishay:
@andrewconner @ryanpbrewster

eishay:
@ryanpbrewster I assume `addMembershipHelper` is called when we're adding an owner, right?

ryanpbrewster:
Yep. It's just addMembership, but it uses a provided DB session.

PR #16557, opened 2015-08-04
refactor

camhashemi:
@gratimax @ryanpbrewster not sure if this is even more readable, albeit more scala-esque

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationDomainOwnershipCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationDomainOwnershipCommander.scala
@@ -71,25 +71,17 @@ class OrganizationDomainOwnershipCommanderImpl @Inject() (
 
   override def addDomainOwnership(orgId: Id[Organization], domainName: String): Either[OwnDomainFailure, OwnDomainSuccess] = {
     db.readWrite { implicit session =>
-      NormalizedHostname.fromHostname(domainName) match {
-        case None => Left(InvalidDomainName(domainName))
-        case Some(normalizedHostname) =>
-          val domainOpt = domainRepo.get(normalizedHostname)
-          domainOpt match {
-            case None => Left(DomainDidNotExist(domainName))
-            case Some(domain) =>
-              val normalizedHostname = domain.hostname
-              orgDomainOwnershipRepo.getOwnershipForDomain(normalizedHostname) match {
-                case Some(own) if own.organizationId != orgId => Left(DomainAlreadyOwned(normalizedHostname.value))
-                case _ =>
-                  val ownership = orgDomainOwnershipRepo.getDomainOwnershipBetween(orgId, normalizedHostname, onlyState = None) match {
-                    case Some(own) if own.state == OrganizationDomainOwnershipStates.ACTIVE => own
-                    case Some(own) => orgDomainOwnershipRepo.save(own.copy(state = OrganizationDomainOwnershipStates.ACTIVE))
-                    case None => orgDomainOwnershipRepo.save(OrganizationDomainOwnership(organizationId = orgId, normalizedHostname = normalizedHostname))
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationDomainOwnershipRepo.scala
```diff
@@ -54,24 +54,16 @@ class OrganizationDomainOwnershipRepoImpl @Inject() (
...
 
-  override def getDomainOwnershipBetween(organization: Id[Organization], domainHostname: NormalizedHostname, onlyState: Option[State[OrganizationDomainOwnership]] = Option(OrganizationDomainOwnershipStates.ACTIVE))(implicit session: RSession): Option[OrganizationDomainOwnership] = {
-    val query =
-      if (onlyState.isDefined) {
-        for { row <- rows if row.organizationId === organization && row.domainHostname === domainHostname && row.state === onlyState } yield row
-      } else {
-        for { row <- rows if row.organizationId === organization && row.domainHostname === domainHostname } yield row
-      }
-    query.firstOption
+  override def getDomainOwnershipBetween(organization: Id[Organization], domainHostname: NormalizedHostname, excludeState: Option[State[OrganizationDomainOwnership]] = Some(OrganizationDomainOwnershipStates.INACTIVE))(implicit session: RSession): Option[OrganizationDomainOwnership] = {
```
ryanpbrewster, line 57: Does this need to be an override? There's no base implementation, right?

maxov, line 57: You can override abstract methods defined in parent traits. This means that if the method in the parent trait is removed or the signature is changed, a compiler error results. I think that this is generally good practice, but others may have opinions on it.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationDomainOwnershipCommander.scala
```diff
@@ -71,25 +71,17 @@ class OrganizationDomainOwnershipCommanderImpl @Inject() (
...
-              }
-          }
+      NormalizedHostname.fromHostname(domainName).fold[Either[OwnDomainFailure, OwnDomainSuccess]](ifEmpty = Left(InvalidDomainName(domainName))) { normalizedHostname =>
+        domainRepo.get(normalizedHostname).fold[Either[OwnDomainFailure, OwnDomainSuccess]](ifEmpty = Left(DomainDidNotExist(domainName))) { domain =>
+          orgDomainOwnershipRepo.getOwnershipForDomain(domain.hostname, excludeState = None).fold[Either[OwnDomainFailure, OwnDomainSuccess]](
+            ifEmpty = Right(OwnDomainSuccess(domain, orgDomainOwnershipRepo.save(orgDomainOwnershipRepo.save(OrganizationDomainOwnership(organizationId = orgId, normalizedHostname = domain.hostname)))))
+          ) {
+              case ownership if ownership.state == OrganizationDomainOwnershipStates.INACTIVE =>
+                Right(OwnDomainSuccess(domain, orgDomainOwnershipRepo.save(ownership.copy(state = OrganizationDomainOwnershipStates.ACTIVE))))
+              case ownership if ownership.organizationId != orgId => Right(OwnDomainSuccess(domain, ownership))
```
maxov: I don't think we mean to return a success here, if the domain is owned by somebody else it means that we cannot claim it.

maxov:
:+1: I think it's more readable, big improvement. Nice!

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationDomainOwnershipCommander.scala
camhashemi: nice catch :8ball:

PR #16562, opened 2015-08-04
Starting to migrate over Keep.libraryId functions to KeepToLibrary

ryanpbrewster:
@andrewconner @leogrim

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepToLibraryCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepToLibraryCommander.scala
@@ -44,8 +44,9 @@ class KeepToLibraryCommanderImpl @Inject() (
     getValidationError(ar) match {
       case Some(fail) => Left(fail)
       case None =>
-        keepToLibraryRepo.getByKeepIdAndLibraryId(ar.keepId, ar.libraryId, excludeStateOpt = None) match {
-          case Some(link) => // existing link guaranteed to be inactive
+        keepToLibraryRepo.getByKeepIdAndLibraryId(ar.keepId, ar.libraryId, excludeStates = Set.empty) match {
+          case Some(link) =>
+            assert(link.isInactive) // existing link guaranteed to be inactive
             val reactivatedLink = keepToLibraryRepo.activate(link.copy(keeperId = ar.requesterId))
             Right(KeepToLibraryAttachResponse(reactivatedLink))
           case None =>
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepToLibraryCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepToLibraryCommander.scala
@@ -4,13 +4,14 @@ import com.google.inject.{ ImplementedBy, Inject, Singleton }
 import com.keepit.common.db.Id
 import com.keepit.common.db.slick.DBSession.{ RSession, RWSession }
 import com.keepit.common.db.slick.Database
+import com.keepit.common.healthcheck.AirbrakeNotifier
 import com.keepit.common.logging.Logging
 import com.keepit.model._
 
 @ImplementedBy(classOf[KeepToLibraryCommanderImpl])
 trait KeepToLibraryCommander {
-  def attach(ar: KeepToLibraryAttachRequest)(implicit session: RWSession): Either[KeepToLibraryFail, KeepToLibraryAttachResponse]
-  def detach(dr: KeepToLibraryDetachRequest)(implicit session: RWSession): Either[KeepToLibraryFail, KeepToLibraryDetachResponse]
+  def addKeepToLibrary(ar: KeepToLibraryAddRequest)(implicit session: RWSession): Either[KeepToLibraryFail, KeepToLibraryAddResponse]
+  def removeKeepFromLibrary(dr: KeepToLibraryRemoveRequest)(implicit session: RWSession): Either[KeepToLibraryFail, KeepToLibraryRemoveResponse]
 }
 
 @Singleton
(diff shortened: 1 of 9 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepToLibraryCommander.scala
```diff
@@ -44,8 +44,9 @@ class KeepToLibraryCommanderImpl @Inject() (
     getValidationError(ar) match {
       case Some(fail) => Left(fail)
       case None =>
-        keepToLibraryRepo.getByKeepIdAndLibraryId(ar.keepId, ar.libraryId, excludeStateOpt = None) match {
-          case Some(link) => // existing link guaranteed to be inactive
+        keepToLibraryRepo.getByKeepIdAndLibraryId(ar.keepId, ar.libraryId, excludeStates = Set.empty) match {
+          case Some(link) =>
+            assert(link.isInactive) // existing link guaranteed to be inactive
```
ryanpbrewster: Not sure if `asserts` are the best way to handle this assumption. I'd like to know if it's false, though.

kifi-backend/modules/shoebox/app/com/keepit/model/KeepRepo.scala
```diff
@@ -41,30 +40,35 @@ trait KeepRepo extends Repo[Keep] with ExternalIdColumnFunction[Keep] with SeqNu
...
   def latestKeptAtByLibraryIds(libraryIds: Set[Id[Library]])(implicit session: RSession): Map[Id[Library], Option[DateTime]]
-  def getKeepsByTimeWindow(uriId: Id[NormalizedURI], url: String, keptAfter: DateTime, keptBefore: DateTime)(implicit session: RSession): Set[Keep]
-  def getRecentKeepsFromFollowedLibraries(userId: Id[User], limit: Int, beforeIdOpt: Option[ExternalId[Keep]], afterIdOpt: Option[ExternalId[Keep]])(implicit session: RSession): Seq[Keep]
-  def getMaxKeepSeqNumForLibraries(libIds: Set[Id[Library]])(implicit session: RSession): Map[Id[Library], SequenceNumber[Keep]]
-  def recentKeepNotes(libId: Id[Library], limit: Int)(implicit session: RSession): Seq[String]
-  def getByLibraryIdAndExcludingVisibility(libId: Id[Library], excludeVisibility: Option[LibraryVisibility], limit: Int)(implicit session: RSession): Seq[Keep]
-  def getKeepSourcesByUser(userId: Id[User])(implicit session: RSession): Seq[KeepSource]
+
+  def deactivate(model: Keep)(implicit session: RWSession): Unit
 }
```
ryanpbrewster, line 73: All of this was just me reordering the methods.

kifi-backend/modules/shoebox/app/com/keepit/model/KeepRepo.scala
```diff
@@ -223,16 +230,20 @@ class KeepRepoImpl @Inject() (
   }
 
   def getByExtIdAndUser(extId: ExternalId[Keep], userId: Id[User])(implicit session: RSession): Option[Keep] = {
-    (for (b <- rows if b.externalId === extId && b.userId === userId) yield b).firstOption
```
ryanpbrewster, line 226: I don't understand these methods at all. An `ExternalId[Keep]` uniquely identifies a keep. Is this just a way of asserting that the keep in question is _also_ owned by the provided user?

kifi-backend/modules/shoebox/app/com/keepit/model/KeepRepo.scala
```diff
@@ -223,16 +230,20 @@ class KeepRepoImpl @Inject() (
...
+    getByExtId(extId) flatMap { keep =>
+      if (keep.userId != userId) {
+        log.info("[KTL] unexpected keep owner: $")
```
ryanpbrewster, line 236: We'll find out soon.

kifi-backend/modules/shoebox/app/com/keepit/model/KeepRepo.scala
```diff
@@ -480,14 +483,15 @@ class KeepRepoImpl @Inject() (
...
-  private val getByExtIdandLibraryIdCompiled = Compiled { (extId: Column[ExternalId[Keep]], libraryId: Column[Id[Library]]) =>
-    (for (b <- rows if b.externalId === extId && b.libraryId === libraryId) yield b)
-  }
   def getByExtIdandLibraryId(extId: ExternalId[Keep], libraryId: Id[Library], excludeSet: Set[State[Keep]])(implicit session: RSession): Option[Keep] = {
-    if (excludeSet.isEmpty) {
-      getByExtIdandLibraryIdCompiled(extId, libraryId).firstOption
-    } else {
-      (for (b <- rows if b.externalId === extId && b.libraryId === libraryId && !b.state.inSet(excludeSet)) yield b).firstOption
+    // TODO(ryan): deprecate ASAP
+    getByExtId(extId, excludeSet).flatMap { keep =>
```
ryanpbrewster, line 489: Same story, is this just a way of asserting that a keep is in a given library? Could it be renamed `getByExtIdIfKeepIsInLibrary`?

kifi-backend/modules/shoebox/test/com/keepit/controllers/website/KeepsControllerTest.scala
```diff
@@ -57,8 +57,9 @@ class KeepsControllerTest extends Specification with ShoeboxTestInjector with He
   def stateForTitle(title: String)(implicit injector: Injector): String = forTitle(title).state.value
 
   def forTitle(title: String)(implicit injector: Injector): Keep = {
+    // hilariously inefficient, but it's only for testing
     inject[Database].readWrite { implicit session =>
-      val keeps = inject[KeepRepo].getByTitle(title)
```
ryanpbrewster, line 61: I killed `KeepRepo.getByTitle` because it was only used in this one spot.

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepToLibraryCommander.scala
andrewconner: If you do `assert`, at least name it. But why do you want this assertion? Making the function idempotent is nice, will help against race conditions and users click twice quickly.

kifi-backend/modules/shoebox/app/com/keepit/model/KeepRepo.scala
andrewconner, line 226: Yes. Otherwise, if I knew your keep's externalid, I could do things with it potentially (it would be up to the consumer of this to ensure; might as well catch it at the source)

andrewconner, line 226: To be clear(er), this is probably an outdated concept with collaborative libraries

kifi-backend/modules/shoebox/app/com/keepit/model/KeepRepo.scala
andrewconner, line 236: As said above, I'm not sure when this is called, but for collaborative libraries that will be unequal.

kifi-backend/modules/shoebox/app/com/keepit/model/KeepRepo.scala
andrewconner, line 489: Yeah. This one is probably more useful, since currently this is true.

kifi-backend/modules/shoebox/test/com/keepit/controllers/website/KeepsControllerTest.scala
andrewconner, line 61: Oh wow, nice catch. That's a strange method.

andrewconner:
Nice job. Few comments, merge when happy :us:

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepToLibraryCommander.scala
ryanpbrewster: :+1: My only concern is that it updates the `keeperId`, and if it's an active KeepToLibrary link then you're basically stealing it from someone.

Right now that doesn't matter. Hopefully in the future it won't matter.

leogrim: Well `KeepToLibrary.keeperId` is really there for tracking purpose isn't it? As in "this is the user who added it to that library". If the keep is removed from the library and added again, it makes sense that the new user would have it. Actually something even clearer that I've been doing lately is creating a brand new model and copying the id of the old one, this way you don't have to worry about "sanitizing" the inactive model and are guaranteed that you get the same end result, whether an inactive (= deleted, really) row was there or not. (recently, see `UserEmailAddress.intern` for an example).

The one user id that shouldn't be touched is `Keep.userId`, the creator / owner of the keep.

What about `addedBy` instead of `keeperId`? "Keeping" is so overloaded in our system that I wouldn't use it here.

leogrim: @ryanpbrewster @andrewconner 

After further reading, I want to further emphasize that thinking of models in terms of successive "activations" and "deactivations" is bound to introduce sneaky errors. Once a model is deactivated, it should be considered gone. If not, we expose ourself to undefined behavior depending on some hypothetical sanitization code. 

Creating a brand new model and only reusing the old id is the best way I can think of to easily avoid that.

On another note, I'm really not a fan of the "linking" vocabulary. I love graphs as much as anyone else, and solid abstractions top to bottom, but once we get to the code, I'd rather use something consistent with the product logic, as in `addKeepToLibrary` and `removeKeepFromLibrary`.

andrewconner: You make good points. One downside of always creating a new model is making constraints a pain (and we end up doing `isPrimary` and such).

> Creating a brand new model and only reusing the old id is the best way I can think of to easily avoid that

Yeah, that would fix it, and I like.

leogrim: I have done that recently in a couple of places, and it makes things pretty clear to me. Usually I end up with a ending case statement `case inactiveModelOpt` that handles both an inactive model or no row at all, and in there a `newModel.copy(id = inactiveModelOpt.flatMap(_.id))`.

andrewconner: Ryan and I discussed on Slack idempotency. He made the point that things like `kept_at` will be updated (good point!). From this, a couple things: `Keep.keptAt` will need to be removed, or at least not used besides the _initial_ keep; ordering inside libraries should use `KeepToLibrary.addedAt` or something (which doesn't exist; we need to make it).

So, this function can't be strictly idempotent, but being able to call it twice without error is a nice property. Léo's point about naming reiterated this: our APIs should be in line with how the product works; it's the easiest to understand, and doesn't let elegance produce impossible situations. If this was `addKeepToLibrary`, then being able to call it twice makes sense and wouldn't be surprising; it ensures a keep is associated with a library.

leogrim: Note that the last part of my rant was acted upon by @ryanpbrewster _before_ I submitted it, the naming was changed on the page reload, impressive :)

ryanpbrewster: Note that editing code while this conversation is updating in the background is hilarious. I was about 15 seconds ahead of y'all, so it was like having a sports commentator.

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepToLibraryCommander.scala
```diff
@@ -19,49 +20,48 @@ class KeepToLibraryCommanderImpl @Inject() (
...
-          case None =>
-            val newLink = keepToLibraryRepo.save(KeepToLibrary(keepId = ar.keepId, libraryId = ar.libraryId, keeperId = ar.requesterId))
-            Right(KeepToLibraryAttachResponse(newLink))
-        }
+  def addKeepToLibrary(ar: KeepToLibraryAddRequest)(implicit session: RWSession): Either[KeepToLibraryFail, KeepToLibraryAddResponse] = {
+    getValidationError(ar).map(Left(_)).getOrElse {
+      keepToLibraryRepo.getByKeepIdAndLibraryId(ar.keepId, ar.libraryId, excludeStates = Set.empty) match {
+        case Some(ktl) =>
+          if (ktl.isActive) {
+            airbrake.notify(s"User ${ar.requesterId} managed to add a keep ${ar.keepId} a library ${ar.libraryId} where it already exists")
```
andrewconner: Why?

ryanpbrewster: I don't expect it to ever go off, but it would mean we have a race condition of some sort.

ryanpbrewster: It also functions as a vaguely more informative assert.

andrewconner: But is it a problem? You could skip the `ALREADY_IN_LIBRARY` thing and just let it through, and consider it normal.

leogrim: Funny story, I'm running into the same issue with `UserCommander.addEmail`, which isn't idempotent either. It just makes things harder, we should always go with `intern` instead of `add` when it's possible. #idempotence4ever

ryanpbrewster: What if user A adds keep K to a library, then user B does the same? Should user B succeed? If so, should `addedBy` reflect user A or B?

leogrim: If you think of it as `interning`, or `getOrElseCreate`, then you shouldn't touch the model if it's already there. And I think it makes sense in this case, that would be the expected product behavior, i.e. full idempotence.

ryanpbrewster: I want calling `addKeepToLibrary` (or `intern`, if that makes more sense) to update the `addedAt` field.

ryanpbrewster: But maybe that isn't the correct behavior, maybe we want it to stay interned. `intern` has much simpler logic, so I'm 100% about that life.

andrewconner: Semantics are probably different than `getOrElseCreate`. In this case (and the other place we have this concept, keeps), `intern` means "get the database in sync with these parameters". Sometimes that means creating, sometimes updating. In Ryan's situation, I'd say that User B succeeds and the db records reflects user B doing it (since it was last).

leogrim: Agree on the semantics part. On the problem at end, I'm not sure, user A first added the keep, `addedAt` should reflect when and how the keep appeared in the library. You could have a `lastAddedAt` field to track whether someone else doubled on that but I'm not sure why you would do that. 

I think of it like adding a participant to a conversation, seems exactly equivalent to me, would you update the `UserThread.{addedAt, addedBy}` field if someone else didn't notice the person was already there and clicked to add him / her again?

andrewconner: For that case, true, because participants are unordered (further, order doesn't really have a meaning). Though, when a user re-keeps, order matters, because they expect it to be at the top. Alternatively, we could handle this case by directing the user to the existing keep or something...

leogrim: But if they rekeep, it means they removed the keep first, so it's a brand new model?

leogrim: And the equivalent for participants would be how multiple libraries are ordered for a single keep. This is different from how keeps are ordered inside of a single library, which would be equivalent to how discussions are ordered from a single participant's view (last active).

PR #16566, opened 2015-08-04
Refactor UserEmailAddress interning / deletion

leogrim:
@andrewconner there's more work to do but we should be in a position to add a unique constraint to `UserEmailAddress` once this goes out. I need to figure out whether I need a hash or not to enforce it though (perhaps I can get away with a prefix index, need to read up on MySQL).

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/UserCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/UserCommander.scala
@@ -211,67 +211,51 @@ class UserCommander @Inject() (
     }
   }
 
-  def addEmail(userId: Id[User], address: EmailAddress, isPrimary: Boolean): Future[Either[String, UserEmailAddress]] = {
+  // todo(Léo): this method isn't resilient to intermediate failures, should be made idempotent and atomic (and confirmation email can be sent async)
+  def addEmail(userId: Id[User], address: EmailAddress, isPrimary: Boolean): Future[Either[String, Unit]] = {
     db.readWrite { implicit session =>
-      if (emailRepo.getByAddressOpt(address).isEmpty) {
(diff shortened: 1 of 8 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/UserCommander.scala
```diff
@@ -211,67 +211,51 @@ class UserCommander @Inject() (
     }
   }
 
-  def addEmail(userId: Id[User], address: EmailAddress, isPrimary: Boolean): Future[Either[String, UserEmailAddress]] = {
+  // todo(Léo): this method isn't resilient to intermediate failures, should be made idempotent and atomic (and confirmation email can be sent async)
```
leogrim, line 214: I need to understand how the front-end behaves when returned with `email already added` before I can clean this up.

kifi-backend/modules/shoebox/app/com/keepit/model/UserEmailAddress.scala
```diff
@@ -29,7 +29,9 @@ case class UserEmailAddress(
       lastVerificationSent = Some(now),
       verificationCode = Some(new BigInteger(128, UserEmailAddress.random).toString(36)))
   }
-  def verified: Boolean = state == UserEmailAddressStates.VERIFIED
+  def clearVerificationCode = copy(lastVerificationSent = None, verificationCode = None)
+  def verificationSent: Boolean = lastVerificationSent.isDefined && verificationCode.isDefined
+  def verified: Boolean = state == UserEmailAddressStates.VERIFIED || (state != UserEmailAddressStates.INACTIVE && verifiedAt.isDefined)
```
leogrim, line 34: Getting ahead of myself here, this will only be used to refactor `VERIFIED` and `UNVERIFIED` states.

andrewconner:
Seems reasonable

PR #16568, opened 2015-08-04
Feature/clean and org

cvializ:
@adamjgrant @andrewconner 

Make 'Organization' the default library visibility when creating a library from an org profile


adamjgrant:
Cool. :+1:

PR #16572, opened 2015-08-04
Do not allow resizing readonly click-to-edit textareas

cvializ:
Minor stylistic/functional thing @adamjgrant


adamjgrant:
I think what we need here is a "read view" where we show plain old HTML content elements when the user is not permissioned to edit. https://app.asana.com/0/42816489233171/44249098003175 I can tackle this one now as I've finished the org spaces selector, or we can chat about it in person. One of the issues I'm anticipating is the profile link being an actual link.

cvializ:
Adam implemented this differently, so I'll close it

adamjgrant:
Thanks, @cvializ I don't know if I explained it well, but I'm concerned about making inputs and textareas behave like other HTML elements. I think we'll be happier in the long run with this approach.

PR #16573, opened 2015-08-04
3.3.70: Workaround for chrome.windows.onFocusChanged bug

andrewconner:
@adamjgrant @cvializ @eishay @leogrim 

This fixes the user reported bug where messages are not marked as read. The issue is a Chrome bug that does not tell us when a window is focused.

I would have preferred to query the Chrome windows and checked focus, but all Chrome APIs are async, and I want to leave `isFocused` as synchronous.


andrewconner:
Merging because it's already deployed.

PR #16578, opened 2015-08-04
More efficient lib permissions checks

andrewconner:
@ryanpbrewster `val` means that it's always evaluated, but it often can be short-circuited if an earlier statement is true.


ryanpbrewster:
I initially had them as `lazy val` but I got a lot of pushback from that.

andrewconner:
Because of accessor performance impact?

PR #16579, opened 2015-08-04
Implementing markup and style for org selector 43038831445882

adamjgrant:


adamjgrant:
This addresses a minor JSHint issue, a missing rule for hiding the checkbox for users not in an org, and for users who are in an org, they will be informed exactly where their library is being created if not in an org.

screen shot 2015-08-04 at 1 26 03 pm

adamjgrant:
@andrewconner @cvializ

cvializ:
wonderful :+1:

PR #16583, opened 2015-08-04
Feature/read only view header

adamjgrant:
Check it @cvializ @andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/common/directives/clickToEdit/clickToEdit.js
+++ b/kifi-backend/modules/shoebox/angular/src/common/directives/clickToEdit/clickToEdit.js
@@ -11,6 +11,7 @@ angular.module('kifi')
         value: '=',
         inputPlaceholder: '=',
         onSave: '=',
+        alink: '=',
         readonly: '='
       },
       replace: true,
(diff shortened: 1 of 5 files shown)


adamjgrant:
Asana: [https://app.asana.com/0/42816489233171/44249098003175](44249098003175)

cvializ:
repetitive, but lightweight. nice

adamjgrant:
Can you suggest some ways to reduce repetition?

adamjgrant:
I added some changes that make the `linkable` syntax more understandable. If you have other ideas, let me know and we can discuss.

cvializ:
If I was more familiar with Angular and could whip this up in no time, I would do something allowing us to inject the desired DOM for the value.

Instead of passing `alink`, we'd pass DOM to the directive with the correct value to be shown when it's not being edited.

```
<span kf-click-to-edit on-save="save" readonly="readonly" input-placeholder="'e.g. http://example.com'" value="profile.site">
  <a href="{{profile.site}}">{{profile.site}}</a>
</span>
```

but I don't know how you'd actually implement it, because I'm lacking in some deeper Angular knowledge

adamjgrant:
Yeah, so something like

```
<[element] class="kf-click-to-edit" ...>
  {{ value }}
</[element]>
```

Where we could specify the DOM element?

adamjgrant:
React might be able to do something like this :)

cvializ:
Haha yeah that's where I'm getting this feeling from, I used React heavily at my last place

PR #16586, opened 2015-08-04
Made HandleCommander ignore inactive users/orgs

ryanpbrewster:
@leogrim 

Check to make sure I'm not breaking any assumptions I'm not aware of.

Ideally, we should just be reclaiming handles when we deactivate the handle owner, but double-checking seems safer.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/admin/AdminUserController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/admin/AdminUserController.scala
@@ -960,8 +960,7 @@ class AdminUserController @Inject() (
     val user = userRepo.get(userId)
 
     userRepo.save(user.withState(UserStates.INACTIVE).copy(primaryEmail = None, primaryUsername = None)) // User
-    // TODO(léo): question from Ryan --> Is this actually correct? reclaimAll does not seem like it actually deactivates handle ownerships
-    handleCommander.reclaimAll(userId, overrideProtection = true, overrideLock = true)
+    handleCommander.reclaimAll(userId)
   }
 
   def deactivateUserEmailAddress(id: Id[UserEmailAddress]) = AdminUserAction { request =>
--- a/kifi-backend/modules/shoebox/app/com/keepit/admin/AdminUserController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/admin/AdminUserController.scala
@@ -960,7 +960,7 @@ class AdminUserController @Inject() (
     val user = userRepo.get(userId)
 
     userRepo.save(user.withState(UserStates.INACTIVE).copy(primaryEmail = None, primaryUsername = None)) // User
-    handleCommander.reclaimAll(userId)
+    handleCommander.reclaimAll(userId, overrideProtection = true, overrideLock = true)
   }
 
   def deactivateUserEmailAddress(id: Id[UserEmailAddress]) = AdminUserAction { request =>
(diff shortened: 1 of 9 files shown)


leogrim:
We shouldn't have active handles for inactive users / orgs in the first place, that means our data is already corrupt in there and our logic broken somewhere else. I'm pretty sure the user deletion cleanup takes care of handles, I see here that you've added something similar for org deletion. Maybe that's all it takes to fix it.

So what about deactivating existing handles owned by inactive orgs (there aren't many I guess) and then add a double-check as a probe to make sure it doesn't happen anymore, i.e. be loud and airbrake if it does. 

Then if we're good, remove the probe, as the code shouldn't suggest that getting bad data is fine and expected here.

Getting an inactive owner in `HandleCommander` is not normal, it hides an underlying bug in our data and can actually affects users, as these zombie active `HandleOwnership` prevent users from claiming the corresponding handles that should be available otherwise.

What do you think?

ryanpbrewster:
Agreed on all points.

ryanpbrewster:
@leogrim I need your input on some of the handle work here, specifically `HandleCommander.deactivateAll`.

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminUserController.scala
```diff
@@ -960,6 +960,7 @@ class AdminUserController @Inject() (
     val user = userRepo.get(userId)
 
     userRepo.save(user.withState(UserStates.INACTIVE).copy(primaryEmail = None, primaryUsername = None)) // User
+    // TODO(léo): question from Ryan --> Is this actually correct? reclaimAll does not seem like it actually deactivates handle ownerships
     handleCommander.reclaimAll(userId, overrideProtection = true, overrideLock = true)
```
leogrim, line 963: True, but they no longer belong to the deactivated user, which is enough and actually preferred. They belong to the system, this way is the handle explicitly locked, we don't release it unexpectedly. 

Scenario: we create an organization for `microsoft`, and we lock that handle and a bunch of others to make sure no one else can ever get `microsoft` (kind of like reserving domain names). Now for some reason, we delete the Microsoft organization. If you deactivate all handles, it means anyone can show up and claim the `microsoft` handle. But if you just reclaim it, it means `microsoft` is now owned by the system, still locked and protected if necessary. If not, another user can claim it just like a regular handle.

The idea is that any handle that is `locked` has been locked explicitly, and should be released explicitly. Because it's usually about protecting some brand (whether it's personal or corporate).

So we shouldn't actually deactivate handles, just reclaim them so that they're no longer associated with the dead entity but any intended protection behavior still exists. Make sense?

kifi-backend/modules/shoebox/app/com/keepit/commanders/HandleCommander.scala
```diff
@@ -76,20 +79,9 @@ class HandleCommanderImpl @Inject() (
...
-    }
-  }
-
-  def reclaimAll(ownerId: HandleOwner, overrideProtection: Boolean = false, overrideLock: Boolean = false)(implicit session: RWSession): Seq[Try[Handle]] = {
-    val handles = handleRepo.getByOwnerId(Some(ownerId)).map(_.handle)
-    handles.map(reclaim(_, overrideProtection = overrideProtection, overrideLock = overrideLock))
+  def reclaimAll(ownerId: HandleOwner)(implicit session: RWSession): Seq[Handle] = {
+    val handleOwnerships = handleRepo.getByOwnerId(Some(ownerId))
+    handleOwnerships.map(own => handleRepo.save(own.copy(ownerId = None)).handle)
   }
```
ryanpbrewster, line 93: @leogrim you're gonna want to take a look at this.

You check for a bunch of things in `getAvailableOwnership` that I'm not sure are relevant when reclaiming handles. There were two issues with the code:
1. It would try and grab the ownership, then give up because the ownership was a `PrimaryOwnership` and thus not available.
2. Even if it could grab the ownership, it never set `ownerId = None` when reclaiming an ownership. It just refreshed the `lastClaimedAt`.

leogrim:
:+1:

PR #16588, opened 2015-08-04
Match backend org upload API. Remove watch variable to close modal.

cvializ:
@andrewconner @adamjgrant

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/common/modal/modal.js
+++ b/kifi-backend/modules/shoebox/angular/src/common/modal/modal.js
@@ -9,7 +9,6 @@ angular.module('kifi')
       restrict: 'A',
       replace: true,
       scope: {
-        forceClose: '=',
         disableScroll: '='
       },
       templateUrl: 'common/modal/modal.tpl.html',
@@ -57,10 +56,8 @@ angular.module('kifi')
         scope.backdropStyle.opacity = attrs.kfOpacity || 0.3;
         scope.backdropStyle.backgroundColor = attrs.kfBackdropColor || 'rgba(0, 40, 90, 1)';
 
-        scope.$watch('forceClose', function (newVal, oldVal) {
-          if (!oldVal && newVal) {
-            scope.close();
-          }
+        scope.$on('forceCloseModal', function () {
+          scope.close();
         });
 
         if (!scope.disableScroll) {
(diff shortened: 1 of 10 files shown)


kifi-backend/modules/shoebox/angular/src/common/modal/modal.js
```diff
@@ -57,10 +56,8 @@ angular.module('kifi')
         scope.backdropStyle.opacity = attrs.kfOpacity || 0.3;
         scope.backdropStyle.backgroundColor = attrs.kfBackdropColor || 'rgba(0, 40, 90, 1)';
 
-        scope.$watch('forceClose', function (newVal, oldVal) {
```
cvializ, line 60: this was gross

kifi-backend/modules/shoebox/angular/src/common/modal/photoUpload/profileImage.js
```diff
@@ -241,32 +243,51 @@ angular.module('kifi')
...
+            picHeight: positioning.imageHeight,
+            cropX: Math.floor(scaling * (maskOffset - positioning.currentLeft)),
+            cropY: Math.floor(scaling * (maskOffset - positioning.currentTop)),
+            cropSize: Math.floor(scaling * maskSize)
+          };
+
+          if (scope.uploadUrl && scope.uploadUrl.indexOf('organization') !== -1) {
+            upload = orgProfileService.uploadOrgAvatar(scope.profile.id, data.cropX, data.cropY, data.picWidth, data.picHeight, scope.files[0]);
+            upload.then(function () {
+              scope.$broadcast('forceCloseModal');
```
andrewconner, line 265: Should we broadcast on the modal's scope?

cvializ, line 265: Isn't it already the modal's scope? `modal.open` takes the current scope as an argument, so it was my understanding that it somehow inherits from the modal-parent's scope

andrewconner, line 265: You're probably right.

adamjgrant:
Looks okay. Sounds like you and @andrewconner are discussing some changes?

PR #16592, opened 2015-08-04
Notifications and notification events

maxov:
Adds notification, notification items, and notification events.

@ryanpbrewster 
@leogrim 
@andrewconner


kifi-backend/modules/eliza/test/com/keepit/eliza/ws/ElizaWebSocketTest.scala
```diff
@@ -130,6 +132,60 @@ class ElizaWebSocketTest extends Specification with ElizaApplicationInjector wit
       }
     }
 
+    "notify of new messages" in {
+      running(new ElizaApplication(modules: _*)) {
```
andrewconner, line 136: Is it required to run an application? Any way we can get away with just the injector? (if you considered this long ago and decided it wasn't feasible, sorry :))

kifi-backend/modules/eliza/app/com/keepit/eliza/model/NotificationEvent.scala
```diff
@@ -0,0 +1,24 @@
...
+  fromUser: Id[User],
+  toUser: Id[User],
+  time: DateTime,
+  data: Option[JsObject]) extends NotificationAction(kind, fromUser, toUser, time, data) with Model[NotificationEvent] {
+
+  override def withId(id: Id[NotificationEvent]): NotificationEvent = copy(id = Some(id))
+
+  override def withUpdateTime(now: DateTime): NotificationEvent = copy(updatedAt = time)
+
+}
```
ryanpbrewster: I think this should extend `ModelWithState[NotificationEvent]` as well.

kifi-backend/modules/eliza/test/com/keepit/eliza/ws/ElizaWebSocketTest.scala
ryanpbrewster, line 136: I think the way he's using the `MockWebSocket` does require the app. Hopefully there won't need to be too many of these tests.

kifi-backend/modules/eliza/app/com/keepit/eliza/model/NotificationEvent.scala
maxov: How would one disable a notification event? It's hooked into a Notification anyway.

kifi-backend/modules/eliza/test/com/keepit/eliza/ws/ElizaWebSocketTest.scala
maxov, line 136: Yep, unfortunately :disappointed:. There's simply too many dependencies in the websocket auth flow to just have the injector.

kifi-backend/modules/common/app/com/keepit/eliza/model/NotificationAction.scala
```diff
@@ -0,0 +1,44 @@
+package com.keepit.eliza.model
+
+import com.keepit.common.db.Id
+import com.keepit.model.{ Library, User }
+import org.joda.time.DateTime
+import play.api.libs.functional.syntax._
+import play.api.libs.json._
+
+class NotificationAction(toUser: Id[User], time: DateTime) {
```
ryanpbrewster: Should NotificationAction be abstract?

kifi-backend/modules/common/app/com/keepit/eliza/model/NotificationAction.scala
```diff
@@ -0,0 +1,44 @@
...
+
+}
+
+case class NewFollower(
+  toUser: Id[User],
+  time: DateTime,
+  followerId: Id[User],
+  library: Library) extends NotificationAction(
+    toUser,
+    time)
```
ryanpbrewster: These `Id` might need to be `ExternalId`

kifi-backend/modules/common/app/com/keepit/eliza/model/NotificationAction.scala
```diff
@@ -0,0 +1,44 @@
...
+
+import com.keepit.common.db.Id
+import com.keepit.model.{ Library, User }
+import org.joda.time.DateTime
+import play.api.libs.functional.syntax._
+import play.api.libs.json._
+
+class NotificationAction(toUser: Id[User], time: DateTime) {
+
+  val kind = implicitly[NotificationActionKind[this.type]]
```
ryanpbrewster: That's pretty cool. :)

kifi-backend/modules/common/app/com/keepit/eliza/model/NotificationActionKind.scala
```diff
@@ -0,0 +1,47 @@
+package com.keepit.eliza.model
+
+import play.api.libs.json._
+
+import scala.annotation.implicitNotFound
+
+@implicitNotFound("No kind object found for action ${N}")
```
ryanpbrewster: What does this do? Throw a compiler error? Warning?

kifi-backend/modules/eliza/app/com/keepit/eliza/model/Notification.scala
```diff
@@ -0,0 +1,20 @@
...
+import com.keepit.common.time._
+import com.keepit.model.User
+import org.joda.time.DateTime
+
+case class Notification(
+    id: Option[Id[Notification]],
+    createdAt: DateTime = currentDateTime,
+    updatedAt: DateTime = currentDateTime,
+    userId: Id[User],
+    lastCheck: DateTime = START_OF_TIME) extends Model[Notification] {
```
ryanpbrewster: Maybe `lastSeen` or `lastChecked`? I like past-tense verbs.

kifi-backend/modules/common/app/com/keepit/eliza/model/NotificationAction.scala
andrewconner: Or a `sealed trait`?

kifi-backend/modules/common/app/com/keepit/eliza/model/NotificationActionKind.scala
```diff
@@ -0,0 +1,47 @@
...
+
+  NotificationActionKind.addKind(this)
+
+}
+
+object NotificationActionKind {
+
+  private var kinds: Map[String, NotificationActionKind[_]] = Map()
+
+  private def addKind(kind: NotificationActionKind[_]) = {
```
andrewconner: Actually not sure this will work. Haven't looked incredibly closely, but since trait bodies are evaluated before the class body, you'll get this situation:

```
trait A {
  val kind: String
  A.addKind(this)
}
object A {
  def addKind(a: A) = println(s"Added: ${a.kind}")
}

class AImpl extends A {
  val kind = "Impl!"
}

scala> new AImpl()
Added: null
res1: AImpl = AImpl@178213b
```

andrewconner: This may do it for you, but I've always had issues with `DelayedInit`: http://www.scala-lang.org/files/archive/nightly/docs/library/index.html#scala.DelayedInit

andrewconner: Alternatively, I've never regretted just explicitly listing every child in the container statically. Downside is verbosity and forgetting to include it, upside is getting some better compiler help about which ones are actually added.

Also downside of dynamic registration: this is lazy, so you have to actually have an instance before it'll be added.

kifi-backend/modules/common/app/com/keepit/eliza/model/NotificationActionKind.scala
```diff
@@ -0,0 +1,47 @@
...
+    kinds = kinds + (kind.name -> kind)
+  }
+
+  def getByName(name: String): Option[NotificationActionKind[_]] = kinds.get(name)
+
+  def kifiLink(path: String): String = s"https://kifi.com/$path"
```
andrewconner: `https://www.kifi.com/$path` otherwise, unnecessary redirect always

kifi-backend/modules/common/app/com/keepit/eliza/model/NotificationActionKind.scala
```diff
@@ -0,0 +1,47 @@
...
+
+  implicit val format = new Format[NotificationActionKind[_]] {
+
+    override def reads(json: JsValue): JsResult[NotificationActionKind[_]] = {
+      val kindName = json.as[String]
```
andrewconner: `(json \ "kind").as[String]` or something?

kifi-backend/modules/eliza/test/com/keepit/eliza/ws/ElizaWebSocketTest.scala
andrewconner, line 136: Womp. Sad.

kifi-backend/modules/common/app/com/keepit/eliza/model/NotificationAction.scala
andrewconner: Well, in here is a `Library`, so weird to mix them. This is an internal class only (by the look of the fields)

kifi-backend/modules/common/app/com/keepit/eliza/model/NotificationAction.scala
```diff
@@ -0,0 +1,44 @@
...
+
+case class NewFollower(
+  toUser: Id[User],
+  time: DateTime,
+  followerId: Id[User],
+  library: Library) extends NotificationAction(
+    toUser,
+    time)
+
+object NewFollower extends NotificationActionKind[NewFollower] {
```
andrewconner: You making notifications typesafe is wonderful.

kifi-backend/modules/common/app/com/keepit/eliza/model/NotificationActionKind.scala
andrewconner: If the implicit isn't found below, it gives a better error message

kifi-backend/modules/common/app/com/keepit/eliza/model/NotificationActionKind.scala
maxov: My plan was for notification actions to be built from json that looks like this:

``` js
{
    "kind": "new_follower",
    "data": ... // arbitrary json data related to notification action
}
```

In this case I'm not sure it makes sense to apply two formats? I was planning to do something like `(json \ "kind").as[NotificationActionKind[_]]` explicitly. If we already use the other pattern, then I am okay changing to that.

kifi-backend/modules/common/app/com/keepit/eliza/model/NotificationActionKind.scala
maxov: That's good to know, I actually didn't know trait bodies were evaluated like this. I'm not keen on using `DelayedInit` since it is deprecated, so I'll explicitly include the list.

kifi-backend/modules/eliza/app/com/keepit/eliza/model/Notification.scala
maxov: Updated in 25bd934bd29

kifi-backend/modules/common/app/com/keepit/eliza/model/NotificationActionKind.scala
andrewconner: Sure, was just commenting on `val kindName = json.as[String]` won't work because it's not a string, right?

kifi-backend/modules/common/app/com/keepit/eliza/model/NotificationAction.scala
maxov: Yeah, I forgot about that. `sealed abstract class` is perfect, it means that you can build new ones with easy syntax:

```
case class NewFollower(
  toUser: Id[User],
  time: DateTime,
  followerId: Id[User],
  library: Library) extends NotificationAction(
    toUser,
    time)
```

kifi-backend/modules/common/app/com/keepit/eliza/model/NotificationAction.scala
maxov: Turns out it doesn't compile :disappointed: need to change it

kifi-backend/modules/common/app/com/keepit/eliza/model/NotificationActionKind.scala
maxov: Fixed in https://github.com/gratimax/keepit/commit/624492cedc2e0213ae8dfdff6ebb42aba3b8d413#diff-bab609abec208d99c880a522b71f5c6fR25

kifi-backend/modules/common/app/com/keepit/eliza/model/NotificationActionKind.scala
maxov: Fixed in https://github.com/gratimax/keepit/commit/624492cedc2e0213ae8dfdff6ebb42aba3b8d413#diff-c082f830dcea7b907e55376dc69bb124R22

kifi-backend/modules/common/app/com/keepit/eliza/model/NotificationAction.scala
maxov: Updated in https://github.com/gratimax/keepit/commit/624492cedc2e0213ae8dfdff6ebb42aba3b8d413#diff-278a5624bb8c8adbce20d11978506081R9

kifi-backend/modules/common/app/com/keepit/eliza/model/NotificationActionKind.scala
maxov: I get what you mean, fixed in 0e49ad15470295a

maxov:
@ryanpbrewster @andrewconner I think it's done, ready for final review.

PR #16593, opened 2015-08-04
Org avatar upload api change

ryanpbrewster:
Org avatar uploads now take 3 `GET` parameters: `(x, y, s)`, defining an `s` by `s` box at position `(x,y)` in the image. That box is first cropped out of the image, then scaled to a couple of useful sizes on the backend (`100x100` and `200x200` currently).

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/commanders/ProcessedImageHelper.scala
+++ b/kifi-backend/modules/common/app/com/keepit/commanders/ProcessedImageHelper.scala
@@ -30,8 +30,9 @@ case class ScaleImageRequest(size: ImageSize) extends ProcessImageRequest {
 case class CenteredCropImageRequest(size: ImageSize) extends ProcessImageRequest {
   val operation = ProcessImageOperation.CenteredCrop
 }
-case class CropScaleImageRequest(offset: ImageOffset, cropSize: ImageSize, size: ImageSize) extends ProcessImageRequest {
+case class CropScaleImageRequest(offset: ImageOffset, cropSize: ImageSize, finalSize: ImageSize) extends ProcessImageRequest {
   val operation = ProcessImageOperation.CropScale
+  def size: ImageSize = finalSize
 }
 
 object ScaleImageRequest {
--- a/kifi-backend/modules/common/app/com/keepit/commanders/ProcessedImageHelper.scala
+++ b/kifi-backend/modules/common/app/com/keepit/commanders/ProcessedImageHelper.scala
@@ -36,7 +36,7 @@ case class CenteredCropImageRequest(size: ImageSize) extends ProcessImageRequest
 case class CropScaleImageRequest(offset: ImageOffset, cropSize: ImageSize, finalSize: ImageSize) extends ProcessImageRequest {
   val operation = ProcessImageOperation.CropScale
   def size: ImageSize = finalSize
-  val pathFragment = s"${cropSize.width}x${cropSize.height}+${offset.x}+${offset.y}->${finalSize.width}x${finalSize.height}" + operation.fileNameSuffix
+  val pathFragment = s"${cropSize.width}x${cropSize.height}+${offset.x}+${offset.y}-${finalSize.width}x${finalSize.height}" + operation.fileNameSuffix
 }
 
 object ScaleImageRequest {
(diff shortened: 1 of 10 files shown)


ryanpbrewster:
@cvializ @andrewconner

cvializ:
Alright, I'll change the front end to pass 3 args

kifi-backend/modules/common/app/com/keepit/commanders/ProcessedImageHelper.scala
```diff
@@ -122,6 +127,10 @@ trait ProcessedImageHelper {
     val resizedImages = sizes.map { processImageRequest =>
       log.info(s"[pih] processing images baseLabel=$baseLabel format=$outFormat to $processImageRequest")
 
+      val isCropScale = processImageRequest match {
+        case _: CropScaleImageRequest => true
+        case _ => false
+      }
```
ryanpbrewster, line 133: Trying to be very safe, so we do not change the way that any of the old images work.

kifi-backend/modules/common/app/com/keepit/commanders/ProcessedImageHelper.scala
```diff
@@ -130,7 +139,11 @@ trait ProcessedImageHelper {
 
       process().map { resizedImage =>
         validateAndGetImageInfo(resizedImage).map { imageInfo =>
-          val key = ImagePath(baseLabel, hash, ImageSize(imageInfo.width, imageInfo.height), processImageRequest.operation, outFormat)
+          val key = if (isCropScale) {
+            ImagePath(baseLabel, hash, processImageRequest, outFormat)
+          } else {
+            ImagePath(baseLabel, hash, ImageSize(imageInfo.width, imageInfo.height), processImageRequest.operation, outFormat)
+          }
```
ryanpbrewster, line 146: Same thing, very cautious.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -888,7 +888,9 @@ class LibraryCommanderImpl @Inject() (
       val userOwnsLibrary = library.ownerId == userId
       val canMoveFromSpace = from match {
         case OrganizationSpace(fromOrg) =>
-          organizationMembershipCommander.getPermissions(fromOrg, Some(userId)).contains(OrganizationPermission.REMOVE_LIBRARIES)
+          library.ownerId == userId
+        // TODO(ryan): when the frontend has UI for this, add it in
+        // organizationMembershipCommander.getPermissions(fromOrg, Some(userId)).contains(OrganizationPermission.REMOVE_LIBRARIES)
```
ryanpbrewster, line 893: This shouldn't be here. Let me fix.

kifi-backend/modules/shoebox/test/com/keepit/controllers/website/OrganizationAvatarControllerTest.scala
```diff
@@ -52,12 +52,12 @@ class OrganizationAvatarControllerTest extends Specification with ShoeboxTestInj
...
           inject[FakeUserActionsHelper].setUser(owner)
-          val route = com.keepit.controllers.website.routes.OrganizationAvatarController.uploadAvatar(pubId, 0, 0, 50, 50)
+          val route = com.keepit.controllers.website.routes.OrganizationAvatarController.uploadAvatar(pubId, 15, 23, 50)
           val request = FakeRequest(route.method, route.url).withBody(fakeFile)
-          val result = inject[OrganizationAvatarController].uploadAvatar(pubId, 0, 0, 50, 50)(request)
+          val result = inject[OrganizationAvatarController].uploadAvatar(pubId, 15, 23, 50)(request)
 
           status(result) === OK
-          Json.parse(contentAsString(result)) === Json.parse("""{"uploaded": "oa/26dbdc56d54dbc94830f7cfc85031481_200x200_cs.png"}""")
+          Json.parse(contentAsString(result)) === Json.parse("""{"uploaded": "oa/26dbdc56d54dbc94830f7cfc85031481_50x50+15+23->200x200_cs.png"}""")
```
andrewconner: Let's keep the characters in the URL as simple as possible, because this will get HTML escaped. And then we have to worry about comparing HTML unescaped paths.

`oa/26dbdc56d54dbc94830f7cfc85031481_50-15-23_200x200_cs.png` or something. (width/height isn't needed to be surfaced, if anyone _did_ doubt that it's a square, they could see the 200x200 later)

ryanpbrewster:
@andrewconner I think this is ready to merge, depending on how hard we want to work to hide the `(width, height)` difference.

PR #16596, opened 2015-08-04
Better moving box?

adamjgrant:


adamjgrant:
https://github.com/kifi/keepit/blob/feature/moving-box/kifi-backend/modules/shoebox/angular/img/symbol-sprites/lib/box-open.svg

adamjgrant:
@cvializ

cvializ:
nice, I was thinking to myself that it needed symmetrical flaps

PR #16603, opened 2015-08-05
mobile contacts search endpoint

camhashemi:
@andrewconner @leogrim I'm pretty sure the alias code is good to kill, it does nothing.


andrewconner:
Sure, fine to remove it. We never finished it.

PR #16607, opened 2015-08-05
Fixing 44387955903552 Long string in org profile header wasn't wrapping

adamjgrant:


adamjgrant:
@andrewconner @cvializ

cvializ:
:+1:

PR #16609, opened 2015-08-05
Long org name wrapping fixes

adamjgrant:


adamjgrant:
@andrewconner @cvializ

andrewconner:
Works for me.

PR #16610, opened 2015-08-05
Fix margin of image, link to /undefined

cvializ:
@adamjgrant @andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/orgProfile/orgProfileMember.tpl.html
+++ b/kifi-backend/modules/shoebox/angular/src/orgProfile/orgProfileMember.tpl.html
@@ -1,7 +1,7 @@
 <li class="kf-org-member">
   <div class="kf-org-member-upper clearfix">
     <div class="kf-org-member-upper-left">
-      <a href="{{member|profileUrl}}" class="kf-org-member-link" kf-track-origin="{{currentPageOrigin}}/member" ng-click="triggerClickedAvatar()">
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/angular/src/orgProfile/orgProfileMember.tpl.html
```diff
@@ -1,7 +1,7 @@
 <li class="kf-org-member">
   <div class="kf-org-member-upper clearfix">
     <div class="kf-org-member-upper-left">
-      <a href="{{member|profileUrl}}" class="kf-org-member-link" kf-track-origin="{{currentPageOrigin}}/member" ng-click="triggerClickedAvatar()">
+      <a href="{{ member.id ? (member|profileUrl) : '#' }}" class="kf-org-member-link" kf-track-origin="{{currentPageOrigin}}/member" ng-click="member.id ? triggerClickedAvatar() : null">
```
adamjgrant, line 4: This is probably because I don't know the context of this problem, but if `member` is not defined, why would this show?

cvializ, line 4: `member` will be defined. `member.id` might not be defined if the invitee is not a Kifi user yet.

adamjgrant, line 4: Gotcha.

adamjgrant:
:+1:

PR #16615, opened 2015-08-05
Alternative design for org selector

adamjgrant:


adamjgrant:
This also handles overflow when orgs > 3

@cvializ @andrewconner 

screen shot 2015-08-05 at 11 07 12 am

cvializ:
The check mark size/placement strikes me as odd, but if you believe in it, :+1:

ryanpbrewster:
Can we just put a happy black/green border around it?

adamjgrant:
@cvializ How about this?

screen shot 2015-08-05 at 11 36 11 am

cvializ:
I think I liked the first one better, not to jerk you back and forth. I think it's the two checkmarks with different meanings so close together

adamjgrant:
I see what you mean. I wonder if it would help if it were indented.

adamjgrant:
@cvializ 

screen shot 2015-08-05 at 11 55 06 am

adamjgrant:
We can keep changing this up but I'm going to merge this in for now so I'm not blocking Dafne.

cvializ:
Alright!

PR #16617, opened 2015-08-05
Remove UserEmailAddressRepo.getAddressOpt

leogrim:
@andrewconner


kifi-backend/modules/shoebox/app/com/keepit/controllers/core/AuthHelper.scala
```diff
@@ -383,60 +378,54 @@ class AuthHelper @Inject() (
...
+    libAuthToken.exists { authToken =>
+      libraryPublicId.exists { publicLibId =>
+        Library.decodePublicId(publicLibId).toOption.exists { libId =>
+          db.readWrite { implicit session =>
+            libraryInviteRepo.getByLibraryIdAndAuthToken(libId, authToken).exists { libraryInvite =>
+              libraryInvite.emailAddress.exists { sentTo =>
+                (sentTo == email) && emailAddressRepo.getByAddressAndUser(userId, email).exists { emailRecord =>
+                  userEmailAddressCommander.saveAsVerified(emailRecord)
+                  true
+                }
```
leogrim, line 396: @andrewconner this is another place I discovered where we verify email addresses. The existing logic might have been wrong though, since we didn't check whether the address the library invite was sent to indeed belongs to the user making the request. Makes sense?

kifi-backend/modules/shoebox/app/com/keepit/controllers/core/AuthHelper.scala
```diff
@@ -383,60 +378,54 @@ class AuthHelper @Inject() (
...
+    (request.body \ "email").asOpt[EmailAddress].map { suppliedEmailAddress =>
       db.readOnlyMaster { implicit session =>
-        getResetEmailAddresses(emailAddrStr)
+        getResetEmailAddresses(suppliedEmailAddress)
       } match {
-        case Some((userId, verifiedEmailAddressOpt)) =>
-          val emailAddresses = Set(EmailAddress(emailAddrStr)) ++ verifiedEmailAddressOpt
+        case Some((userId, resetEmailAddressOpt)) =>
+          val emailAddresses = Set(suppliedEmailAddress) ++ resetEmailAddressOpt
           val emailsF = Future.sequence(emailAddresses.map { email => resetPasswordEmailSender.sendToUser(userId, email) }.toSeq)
```
leogrim, line 420: @andrewconner seems like the intended behavior is to email all of a user's addresses with the reset link, that's right?

kifi-backend/modules/shoebox/app/com/keepit/controllers/core/AuthHelper.scala
```diff
@@ -383,60 +378,54 @@ class AuthHelper @Inject() (
...
-      val emailAddrOpt = emailAddressRepo.getByAddressOpt(email, excludeState = None) // TODO: exclude INACTIVE records
-      emailAddrOpt.map(_.userId) orElse socialRepo.getOpt(SocialId(emailAddrStr), SocialNetworks.FORTYTWO).flatMap(_.userId) map { userId =>
-        emailAddrOpt.filter(_.verified) map { _ =>
-          (userId, None)
-        } getOrElse {
-          // TODO: use user's primary email address once hooked up
-          (userId, emailAddressRepo.getAllByUser(userId).find(_.verified).map(_.address))
-        }
+      emailAddressRepo.getByAddress(email).map(_.userId) map { userId =>
+        (userId, Try(emailAddressRepo.getByUser(userId)).toOption)
```
leogrim, line 408: @andrewconner this now includes unverified email addresses (that now belong to at most one user). The previous behavior did not, which meant that if you had not verified your email address, you could not get a reset password link sent to it, which means it was useless to try and verify email addresses based on that. What's your take here?

kifi-backend/modules/shoebox/app/com/keepit/social/SecureSocialUserServiceImpl.scala
```diff
@@ -63,7 +63,7 @@ class SecureSocialUserPluginImpl @Inject() (
         // find a SUI with the correct email address, we go searching.
         val email = EmailAddress(id.userId)
         db.readOnlyMaster { implicit session =>
-          emailRepo.getByAddressOpt(email).flatMap { emailAddr =>
+          emailRepo.getByAddress(email).flatMap { emailAddr =>
             // todo(andrew): Don't let unverified people log in. For now, we are, but come up with something better.
```
leogrim, line 67: @andrewconner What did you have in mind? What's the issue with unverified people logging in? 
I guess it's kind of related to my questions about reset passwords below.

andrewconner, line 67: Right now, we _should_ let unverified people log in because we've never enforced that users should verify. "soon" we should enforce it, or at least nag users every time they log in.

PR #16619, opened 2015-08-05
Slightly better width handling for org dropdown

adamjgrant:
screen shot 2015-08-05 at 12 29 19 pm

Trying to avoid the above layout when the org names are not long.


adamjgrant:
@andrewconner @cvializ

cvializ:
:+1:

cvializ:
I'm sure there's some ridiculous hack with flexbox that would make this work for all the possible cases. But flexbox is weird, so if this does it I'm happy

PR #16621, opened 2015-08-05
Get rid of UserEmailAddressStates.{VERIFIED / UNVERIFIED}

leogrim:
@andrewconner 
This is to be merged _after_ https://github.com/kifi/keepit/pull/16617


andrewconner:
Looks good. The logic is tricky so it's more that I trust you, but I did read through it :).

leogrim:
I was actually more worried about the other one, I think this one should be fine :)

PR #16622, opened 2015-08-05
Fix org repo methods to only yield active orgs

ryanpbrewster:
We run into major issues if we are dealing with inactive organizations
now, since we free up their handles before deleting them.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationRepo.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationRepo.scala
@@ -1,27 +1,22 @@
 package com.keepit.model
 
-import com.google.inject.{ Provider, ImplementedBy, Inject, Singleton }
+import com.google.inject.{ ImplementedBy, Inject, Provider, Singleton }
 import com.keepit.common.actor.ActorInstance
 import com.keepit.common.db._
-import com.keepit.common.db.slick.DBSession.RSession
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationRepo.scala
```diff
@@ -76,26 +71,27 @@ class OrganizationRepoImpl @Inject() (
...
 
+  // Be aware that this will return inactive organizations, so make sure you want the orgs you ask for
   def getByIds(orgIds: Set[Id[Organization]])(implicit session: RSession): Map[Id[Organization], Organization] = {
     orgCache.bulkGetOrElse(orgIds map orgId2Key) { missingKeys =>
-      val q = { for { row <- rows if row.id.inSet(missingKeys.map { _.id }.toSet) && row.state === OrganizationStates.ACTIVE } yield row }
-      q.list.map { x => (orgId2Key(x.id.get) -> x) }.toMap
-    }.map { case (key, org) => key.id -> org }.toMap
+      val q = { for { row <- rows if row.id.inSet(missingKeys.map { _.id }) } yield row }
+      q.list.map { x => orgId2Key(x.id.get) -> x }.toMap
+    }.map { case (key, org) => key.id -> org }
```
ryanpbrewster, line 79: @eishay I changed this to ignore state. The cache ignores state, so previously this method would have inconsistent behavior depending on whether something was in the cache or not.

PR #16624, opened 2015-08-05
Alternative design for org attribution on library pages

adamjgrant:
screen shot 2015-08-05 at 2 33 02 pm

@cvializ @andrewconner


adamjgrant:
As requested

cvializ:
:+1:

PR #16626, opened 2015-08-05
update based on Ryan and Andrew's comments (naming and typos)

eishay:
@andrewconner @ryanpbrewster


eishay:
will apply this as soon as the code deploys:

``` sql
mysql> update library set kind = "system_read_it_later" where kind = "system_read_id_later";
```

PR #16630, opened 2015-08-06
return sorted memberships and unique invitees

camhashemi:

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
@@ -34,7 +34,7 @@ object MaybeOrganizationMember {
 
 @ImplementedBy(classOf[OrganizationMembershipCommanderImpl])
 trait OrganizationMembershipCommander {
-  def getMembersAndUniqueInvitees(orgId: Id[Organization], limit: Limit, offset: Offset, includeInvitees: Boolean): Seq[MaybeOrganizationMember]
+  def getMembersAndUniqueInvitees(orgId: Id[Organization], offset: Offset, limit: Limit, includeInvitees: Boolean): Seq[MaybeOrganizationMember]
   def getOrganizationsForUser(userId: Id[User], limit: Limit, offset: Offset): Seq[Id[Organization]]
   def getPrimaryOrganizationForUser(userId: Id[User]): Option[Id[Organization]]
   def getAllOrganizationsForUser(userId: Id[User]): Seq[Id[Organization]]
@@ -104,15 +104,24 @@ class OrganizationMembershipCommanderImpl @Inject() (
     }
   }
 
-  def getMembersAndUniqueInvitees(orgId: Id[Organization], limit: Limit, offset: Offset, includeInvitees: Boolean): Seq[MaybeOrganizationMember] = {
+  def getMembersAndUniqueInvitees(orgId: Id[Organization], offset: Offset, limit: Limit, includeInvitees: Boolean): Seq[MaybeOrganizationMember] = {
     db.readOnlyMaster { implicit session =>
-      val members = organizationMembershipRepo.getSortedMembershipsByOrgId(orgId)
+      val members = organizationMembershipRepo.getSortedMembershipsByOrgId(orgId, offset, limit)
       val invitees = includeInvitees match {
         case true =>
-          organizationInviteRepo.getAllByOrganizationAndDecision(orgId, decision = InvitationDecision.PENDING).toSeq // includes resends, needs to be deduped
+          if (members.length < limit.value) {
+            val inviteLimit = Limit(Math.max(limit.value - members.length, 0))
+            val inviteOffset = if (members.isEmpty) {
+              val totalMembers = organizationMembershipRepo.countByOrgId(orgId)
+              Offset(offset.value - totalMembers)
+            } else {
+              Offset(0)
+            }
+            organizationInviteRepo.getByOrganizationAndDecision(orgId, decision = InvitationDecision.PENDING, inviteOffset, inviteLimit)
+          } else Seq.empty[OrganizationInvite]
         case false => Seq.empty[OrganizationInvite]
       }
-      buildMaybeMembers(members, invitees).drop(offset.value.toInt).take(limit.value.toInt)
+      buildMaybeMembers(members, invitees)
     }
   }
 
@@ -141,13 +150,8 @@ class OrganizationMembershipCommanderImpl @Inject() (
   }
 
   private def buildMaybeMembers(members: Seq[OrganizationMembership], invitees: Seq[OrganizationInvite]): Seq[MaybeOrganizationMember] = {
-    import com.keepit.common.time.dateTimeOrdering
-
-    val userInvites = invitees.filter(_.userId.isDefined)
-    val emailInvites = invitees.filter(invite => invite.userId.isEmpty && invite.emailAddress.isDefined)
-
-    val invitedUserIds = userInvites.groupBy(_.userId.get).mapValues(_.maxBy(_.updatedAt)).values
-    val invitedEmailAddresses = emailInvites.groupBy(_.emailAddress.get).mapValues(_.maxBy(_.updatedAt)).values // assumes that if no userId is defined, an email address must be
+    val invitedUserIds = invitees.filter(_.userId.isDefined)
+    val invitedEmailAddresses = invitees.filter(invite => invite.userId.isEmpty && invite.emailAddress.isDefined)
 
     val usersMap = db.readOnlyMaster { implicit session =>
       basicUserRepo.loadAllActive((members.map(_.userId) ++ invitedUserIds.map(_.userId.get)).toSet)
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
@@ -143,7 +143,9 @@ class OrganizationMembershipCommanderImpl @Inject() (
   private def buildMaybeMembers(members: Seq[OrganizationMembership], invitees: Seq[OrganizationInvite]): Seq[MaybeOrganizationMember] = {
     import com.keepit.common.time.dateTimeOrdering
 
-    val (userInvites, emailInvites) = invitees.partition(_.userId.isDefined)
+    val userInvites = invitees.filter(_.userId.isDefined)
+    val emailInvites = invitees.filter(invite => invite.userId.isEmpty && invite.emailAddress.isDefined)
+
     val invitedUserIds = userInvites.groupBy(_.userId.get).mapValues(_.maxBy(_.updatedAt)).values
     val invitedEmailAddresses = emailInvites.groupBy(_.emailAddress.get).mapValues(_.maxBy(_.updatedAt)).values // assumes that if no userId is defined, an email address must be
 
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
@@ -34,7 +34,7 @@ object MaybeOrganizationMember {
 
 @ImplementedBy(classOf[OrganizationMembershipCommanderImpl])
 trait OrganizationMembershipCommander {
-  def getMembersAndInvitees(orgId: Id[Organization], limit: Limit, offset: Offset, includeInvitees: Boolean): Seq[MaybeOrganizationMember]
+  def getMembersAndUniqueInvitees(orgId: Id[Organization], limit: Limit, offset: Offset, includeInvitees: Boolean): Seq[MaybeOrganizationMember]
   def getOrganizationsForUser(userId: Id[User], limit: Limit, offset: Offset): Seq[Id[Organization]]
   def getPrimaryOrganizationForUser(userId: Id[User]): Option[Id[Organization]]
   def getAllOrganizationsForUser(userId: Id[User]): Seq[Id[Organization]]
@@ -104,22 +104,15 @@ class OrganizationMembershipCommanderImpl @Inject() (
     }
   }
 
-  def getMembersAndInvitees(orgId: Id[Organization], limit: Limit, offset: Offset, includeInvitees: Boolean): Seq[MaybeOrganizationMember] = {
+  def getMembersAndUniqueInvitees(orgId: Id[Organization], limit: Limit, offset: Offset, includeInvitees: Boolean): Seq[MaybeOrganizationMember] = {
     db.readOnlyMaster { implicit session =>
-      val members = organizationMembershipRepo.getByOrgId(orgId, limit, offset)
+      val members = organizationMembershipRepo.getSortedMembershipsByOrgId(orgId)
       val invitees = includeInvitees match {
         case true =>
-          val leftOverCount = Limit(Math.max(limit.value - members.length, 0))
-          val leftOverOffset = if (members.isEmpty) {
-            val totalCountForOrg = organizationMembershipRepo.countByOrgId(orgId)
-            Offset(offset.value - totalCountForOrg)
-          } else {
-            Offset(0)
-          }
-          organizationInviteRepo.getByOrganization(orgId, leftOverCount, leftOverOffset)
+          organizationInviteRepo.getAllByOrganizationAndDecision(orgId, decision = InvitationDecision.PENDING).toSeq // includes resends, needs to be deduped
         case false => Seq.empty[OrganizationInvite]
       }
-      buildMaybeMembers(members, invitees)
+      buildMaybeMembers(members, invitees).drop(offset.value.toInt).take(limit.value.toInt)
     }
   }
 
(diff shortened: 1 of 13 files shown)


kifi-backend/modules/shoebox/test/com/keepit/model/OrganizationMembershipRepoTest.scala
```diff
@@ -92,6 +93,43 @@ class OrganizationMembershipRepoTest extends Specification with ShoeboxTestInjec
...
     }
 
+    "get list of members sorted by 1) owner, members, pending invites, 2) first name, last name alphabetically" in {
+      withDb() { implicit injector =>
+        val orgRepo = inject[OrganizationRepo]
+        val orgMemberRepo = inject[OrganizationMembershipRepo]
+
+        val (org, owner, members) = db.readWrite { implicit session =>
+          val owner = user.withName("Zyxwv", "Utsr").saved
+          val members = Random.shuffle(Seq(user.withName("Aaron", "Aaronson").saved, user.withName("Barry", "Barnes").saved, user.withName("Carl", "Carson").saved, user.withName("Carl", "Junior").saved))
```
ryanpbrewster: I should shuffle, then map `_.saved`. Right now your'e saving them in alphabetical order.

kifi-backend/modules/shoebox/test/com/keepit/model/OrganizationMembershipRepoTest.scala
```diff
@@ -92,6 +93,43 @@ class OrganizationMembershipRepoTest extends Specification with ShoeboxTestInjec
...
+          val org = organization().withOwner(owner).withMembers(members).saved
+          (org, owner, members)
+        }
+        val sortedMembers = db.readOnlyMaster { implicit session => orgMemberRepo.getSortedMembershipsByOrgId(org.id.get) }
+
+        sortedMembers.exists(_.userId == owner.id.get) === true
+
+        db.readOnlyMaster { implicit s => orgMemberRepo.getByOrgIdAndUserId(org.id.get, owner.id.get) }.isDefined === true
+
+        val userById = members.+:(owner) groupBy (_.id.get)
```
ryanpbrewster: Odd (maybe clever?) use of `groupBy`. I would do `(members +: owner).map(u => u.id.get -> u)` or something along those lines.

kifi-backend/modules/shoebox/test/com/keepit/model/OrganizationMembershipRepoTest.scala
```diff
@@ -92,6 +93,43 @@ class OrganizationMembershipRepoTest extends Specification with ShoeboxTestInjec
...
+            }
+          }
+
+          sortedMembers.head.userId == owner.id.get && isSortedHelper(sortedMembers.tail)
+        }
+
+        isSorted(sortedMembers, userById) === true
+        isSorted(sortedMembers.reverse, userById) === false
+      }
+    }
```
ryanpbrewster, line 124: You can just brute-force sort them. Use `val expected = all.sortBy(metric)` and then compare to `expected`.

kifi-backend/modules/shoebox/test/com/keepit/model/OrganizationMembershipRepoTest.scala
camhashemi: I guess what I was going for was saving the memberships in non-alphabetical order, you're right.

kifi-backend/modules/shoebox/test/com/keepit/model/OrganizationMembershipRepoTest.scala
```diff
@@ -92,6 +93,43 @@ class OrganizationMembershipRepoTest extends Specification with ShoeboxTestInjec
...
+
+          def isSortedHelper(members: Seq[OrganizationMembership]): Boolean = {
+            members match {
+              case Nil => true
+              case x :: Nil => true
+              case x :: xs => (userById(x.userId).head.firstName < userById(xs.head.userId).head.firstName || userById(x.userId).head.lastName < userById(xs.head.userId).head.lastName) && isSortedHelper(members.tail)
+            }
+          }
+
+          sortedMembers.head.userId == owner.id.get && isSortedHelper(sortedMembers.tail)
```
ryanpbrewster: The `.groupBy` is the reason you have `.head` so many times in here.

kifi-backend/modules/shoebox/test/com/keepit/controllers/website/OrganizationMembershipControllerTest.scala
```diff
@@ -129,7 +130,7 @@ class OrganizationMembershipControllerTest extends Specification with ShoeboxTes
           resultInviteesList.length === n
           (json \ "members" \\ "id").length === 0 // no user ids, only emails
           (json \ "members" \\ "email").length === n
-          (json \ "members" \\ "email").map(v => v.as[EmailAddress]) === invitedEmails.take(n)
+          (json \ "members" \\ "email").map(v => v.as[EmailAddress]).toSet === invitedEmails.take(n).toSet
```
ryanpbrewster: For anything user-facing that comes as a `JsArray`, we should make sure the order is correct. It's weird (as a user) if you look at invites one day and they're in a different order than the next day.

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationMembershipRepo.scala
```diff
@@ -12,11 +12,13 @@ import com.keepit.common.time.Clock
...
 import play.api.libs.json._
 
+import scala.slick.jdbc.{ PositionedResult, GetResult }
+
 @ImplementedBy(classOf[OrganizationMembershipRepoImpl])
 trait OrganizationMembershipRepo extends Repo[OrganizationMembership] with SeqNumberFunction[OrganizationMembership] {
   def getByUserId(userId: Id[User], limit: Limit, offset: Offset, excludeStates: Set[State[OrganizationMembership]] = Set(OrganizationMembershipStates.INACTIVE))(implicit session: RSession): Seq[OrganizationMembership]
   def getAllByUserId(userId: Id[User], excludeStates: Set[State[OrganizationMembership]] = Set(OrganizationMembershipStates.INACTIVE))(implicit session: RSession): Seq[OrganizationMembership]
-  def getByOrgId(orgId: Id[Organization], limit: Limit, offset: Offset, excludeState: Option[State[OrganizationMembership]] = Some(OrganizationMembershipStates.INACTIVE))(implicit session: RSession): Seq[OrganizationMembership]
+  def getSortedMembershipsByOrgId(orgId: Id[Organization], excludeState: Option[State[OrganizationMembership]] = Some(OrganizationMembershipStates.INACTIVE))(implicit session: RSession): Seq[OrganizationMembership]
```
ryanpbrewster: I think this should just be `getByOrgId`. The fact that there is a `limit/offset` included necessarily implies that the memberships are sorted. But I don't feel strongly on this one. Aggressive disambiguation isn't a bad thing.

kifi-backend/modules/shoebox/test/com/keepit/model/OrganizationMembershipRepoTest.scala
camhashemi: either way, the typing is wrong here. even with `groupBy` it should be `groupBy(_.id.get).mapValues(_.head)`. changing it

kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtLibraryController.scala
```diff
@@ -63,7 +63,7 @@ class ExtLibraryController @Inject() (
           color = lib.color,
           visibility = lib.visibility,
           path = libPathCommander.getPath(lib),
-          hasCollaborators = !collabs.isEmpty,
+          hasCollaborators = collabs.nonEmpty,
```
ryanpbrewster, line 66: :+1:

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
```diff
@@ -148,8 +141,12 @@ class OrganizationMembershipCommanderImpl @Inject() (
   }
 
   private def buildMaybeMembers(members: Seq[OrganizationMembership], invitees: Seq[OrganizationInvite]): Seq[MaybeOrganizationMember] = {
-    val invitedUserIds = invitees.filter(_.userId.isDefined)
-    val invitedEmailAddresses = invitees.filter(_.emailAddress.isDefined)
+    import com.keepit.common.time.dateTimeOrdering
+
+    val (userInvites, emailInvites) = invitees.partition(_.userId.isDefined)
```
ryanpbrewster: You actually have to use 2x `filter`s. Remember that it is possible to have no user id _and_ no email address (anonymous invite link).

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
```diff
@@ -148,8 +141,12 @@ class OrganizationMembershipCommanderImpl @Inject() (
...
+    val invitedUserIds = userInvites.groupBy(_.userId.get).mapValues(_.maxBy(_.updatedAt)).values
```
ryanpbrewster: Use the `.map(u => u.id.get -> u)` convention to get a map.

kifi-backend/modules/shoebox/test/com/keepit/model/OrganizationMembershipRepoTest.scala
camhashemi: yup... changing it

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
```diff
@@ -104,22 +104,15 @@ class OrganizationMembershipCommanderImpl @Inject() (
...
         case true =>
-          val leftOverCount = Limit(Math.max(limit.value - members.length, 0))
-          val leftOverOffset = if (members.isEmpty) {
-            val totalCountForOrg = organizationMembershipRepo.countByOrgId(orgId)
-            Offset(offset.value - totalCountForOrg)
-          } else {
-            Offset(0)
-          }
-          organizationInviteRepo.getByOrganization(orgId, leftOverCount, leftOverOffset)
+          organizationInviteRepo.getAllByOrganizationAndDecision(orgId, decision = InvitationDecision.PENDING).toSeq // includes resends, needs to be deduped
```
ryanpbrewster: This feels kind of dangerous to me. Most orgs aren't going to have many members, so it's probably fine, but if we have a big org with 200+ members then:
1. Each hit is going to be expensive
2. There are probably going to be a ton of hits

It's a quadratic expense. This _is_ much simpler, though.

kifi-backend/modules/shoebox/test/com/keepit/model/OrganizationMembershipRepoTest.scala
ryanpbrewster: Per @andrewconner's Slack message earlier today, I'm scared of `.mapValues`.

maxov: Shouldn't be a problem, `_.id.get` has no side-effects

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
camhashemi: got it

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
camhashemi: there can be multiple invites to a user or email, so we need `groupBy` to collect them, then take the most recent value.

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationMembershipRepo.scala
camhashemi: we have `limit/offset` in places where things aren't sorted

ryanpbrewster: That is a bug and we should fix it.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
ryanpbrewster: Ah! Great, ignore me then.

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationMembershipRepo.scala
camhashemi: like `getByUserId`: is the implied sorting by `Id`?

I guess what I mean by the naming is that there is a unique and explicit sorting here, it's different than the others.

ryanpbrewster: That's fair. As I said, I don't feel strongly about this one. That being said, it's good that you killed `getByOrgId`. And we should fix `getByUserId` (or drop the `limit/offset`, it shouldn't be necessary).

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
```diff
@@ -148,8 +141,14 @@ class OrganizationMembershipCommanderImpl @Inject() (
   }
 
   private def buildMaybeMembers(members: Seq[OrganizationMembership], invitees: Seq[OrganizationInvite]): Seq[MaybeOrganizationMember] = {
-    val invitedUserIds = invitees.filter(_.userId.isDefined)
-    val invitedEmailAddresses = invitees.filter(_.emailAddress.isDefined)
+    import com.keepit.common.time.dateTimeOrdering
+
+    val userInvites = invitees.filter(_.userId.isDefined)
+    val emailInvites = invitees.filter(invite => invite.userId.isEmpty && invite.emailAddress.isDefined)
```
ryanpbrewster: Huh. I think it would be a bug if both `userId` and `emailAddress` were defined, but that is a good safeguard. I think I will add it to some of my old code.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
```diff
@@ -148,8 +141,14 @@ class OrganizationMembershipCommanderImpl @Inject() (
...
+
+    val invitedUserIds = userInvites.groupBy(_.userId.get).mapValues(_.maxBy(_.updatedAt)).values
+    val invitedEmailAddresses = emailInvites.groupBy(_.emailAddress.get).mapValues(_.maxBy(_.updatedAt)).values // assumes that if no userId is defined, an email address must be
```
ryanpbrewster: Is this comment still correct? You are explicitly checking for an email address now.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
ryanpbrewster: I like the explicit check for `pending`, nice catch.

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationMembershipRepo.scala
camhashemi: I don't think we need to implement `limit/offset` on everything, it's not going to be used in most cases, just adds overhead. When clients ask for it, it's cheap to implement.

leogrim: Though most likely someone will add yet another method with `limit/offset` where the same one could have been reused. I'd rather have it there and call it with `(0, Int.MAX)`, this makes it clear that we're returning everything in a place where pagination may be desired. What do you think?

camhashemi: I agree, makes sense that we would have it all in one place. Though I'm not so sure about the implicit "sorting". Whatever we paginate needs an ordering for consistency, and it's not always clear what that should be. We can use `updated_at` as a default.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
camhashemi: @ryanpbrewster and I talked in person. If we de-duplicate invites it will be easier to fetch invites from the repo. @leogrim do you think it's necessary to have one row for each invite sent to the same destination? We were thinking to de-dupe them by `userId` or `emailAddress`. Re-sends could just update `updated_at`.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
camhashemi: We have some rows with both `userId` and `emailAddress`. I think @leogrim has said we back-fill, but looking at the table I think that when you send an invite directly to a user's email, we fetch the id and save it in the model.

The intention here is to not send to a user's account AND their email as separate entities.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
camhashemi: good catch, thanks.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
ryanpbrewster: Good point.

camhashemi:
@ryanpbrewster updated. now we sort invites for userIds (by `updated_at`) and email addresses (alphabetically). sorting and slicing is done within the repo, but I still complete the query before de-duping, sorting, and slicing.

leogrim:
It's not ideal to use `updated_at` for business logic, better use an
explicit "last_sent_at"

camhashemi:
@leogrim didn't know you were around. sorry for merging early, I'd appreciate your feedback.

We don't have `last_sent_at` for org invites, but we should add that when we de-dupe invites to a destination.

PR #16634, opened 2015-08-06
Move card icons inline so that orgs can eventually replace the cards

cvializ:
@andrewconner, these changes seem right from a regular web-development perspective. Am I doing anything incorrectly for an extension perspective?

@adamjgrant !


cvializ:
Looks like this:
screen shot 2015-08-05 at 5 37 53 pm

andrewconner:
Looks good! :us: Wahoo!

adamjgrant:
Nice work.

PR #16635, opened 2015-08-06
added_at is now a field in KeepToLibrary

ryanpbrewster:
Need to add a column to the DB table as well. Also coming soon to a theater
near you: changing this field to `NOT NULL`.

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/model/KeepToLibrary.scala
+++ b/kifi-backend/modules/common/app/com/keepit/model/KeepToLibrary.scala
@@ -14,7 +14,15 @@ case class KeepToLibrary(
   state: State[KeepToLibrary] = KeepToLibraryStates.ACTIVE,
   keepId: Id[Keep],
   libraryId: Id[Library],
-  addedBy: Id[User])
+  addedAt: DateTime = currentDateTime,
+  addedBy: Id[User],
+  // A bunch of denormalized fields from Keep
+  uriId: Id[NormalizedURI],
+  isPrimary: Boolean = true,
+  keepOwner: Id[User],
+  // and from Library (which are already denormalized on Keep as well)
+  libraryVisibility: LibraryVisibility,
+  libraryOrganizationId: Option[Id[Organization]])
     extends ModelWithState[KeepToLibrary] {
 
   def withId(id: Id[KeepToLibrary]): KeepToLibrary = this.copy(id = Some(id))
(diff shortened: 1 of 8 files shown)


kifi-backend/modules/common/app/com/keepit/model/KeepToLibrary.scala
```diff
@@ -14,7 +14,15 @@ case class KeepToLibrary(
   state: State[KeepToLibrary] = KeepToLibraryStates.ACTIVE,
   keepId: Id[Keep],
   libraryId: Id[Library],
-  addedBy: Id[User])
+  addedAt: DateTime = currentDateTime,
+  addedBy: Id[User],
+  // A bunch of denormalized fields from Keep
+  uriId: Id[NormalizedURI],
+  isPrimary: Boolean = true,
+  keepOwner: Id[User],
```
leogrim, line 22: Not sure we need this one, what's your use case?

kifi-backend/modules/common/app/com/keepit/model/KeepToLibrary.scala
```diff
@@ -14,7 +14,15 @@ case class KeepToLibrary(
   state: State[KeepToLibrary] = KeepToLibraryStates.ACTIVE,
   keepId: Id[Keep],
   libraryId: Id[Library],
-  addedBy: Id[User])
+  addedAt: DateTime = currentDateTime,
+  addedBy: Id[User],
+  // A bunch of denormalized fields from Keep
+  uriId: Id[NormalizedURI],
```
leogrim, line 20: We should probably have a foreign key constraint on this one.

kifi-backend/modules/common/conf/evolutions/shoebox/370.sql
```diff
@@ -0,0 +1,14 @@
+# SHOEBOX
+
+# --- !Ups
+
+ALTER TABLE keep_to_library ADD COLUMN added_at DATETIME NOT NULL;
+ALTER TABLE keep_to_library ADD COLUMN uri_id BIGINT(20) NOT NULL;
+ALTER TABLE keep_to_library ADD COLUMN is_primary TINYINT(1) NOT NULL;
```
leogrim, line 7: Check out `Keep.isPrimary`, you need to implement this with a nullable `Boolean` (should be `True` or `null`) so that we can have a unique constraint here (until we get rid of it).

ryanpbrewster, line 7: Yup, fixed.

ryanpbrewster:
@leogrim, I'm going to close this and open a new PR.

PR #16636, opened 2015-08-06
Payment System Core Repos

stkem:
@andrewconner @leogrim @eishay (who else?)
Next up tests and then Commander logic (currently in a non compiling state, so not in this PR. Also to keep the diff size more manageable)


kifi-backend/modules/shoebox/app/com/keepit/payments/PaymentMethodRepo.scala
```diff
@@ -0,0 +1,51 @@
...
+    val clock: Clock) extends DbRepo[PaymentMethod] with PaymentMethodRepo {
+
+  import com.keepit.common.db.slick.DBSession._
+  import db.Driver.simple._
+
+  implicit val stripeTokenColumnType = MappedColumnType.base[StripeToken, String](_.token, StripeToken(_))
+
+  //Put this in a common package? Still needs specific tests
+  //ZZZ NOT CORRECT YET!!
+  implicit val trueOrNullColumnType = MappedColumnType.base[TrueOrNull, Boolean](
```
stkem: I spent several hours unsuccessfully trying to make a type mapper that would map `TrueOrNull(true)` to `TRUE` and `TrueOrNull(false)` to `NULL`. I stopped short of implementing my own `JdbcType`. Might be a good use case for our typesafe subscription, @andrewconner ?

leogrim: This is for the same use case as `Keep.primary` isn't it?

leogrim: What about mapping to `Option[Boolean]`?

stkem: Yup, same use case as `Keep.primary`. If we can get it to work we can switch that one out as well.
Mapping to `Option[Boolean]` unfortunately doesn't work either. Slick treats option types specially (i.e. it's not just an understood type like any other) and that treatment hooks in earlier than the type mappers, so when an `Option` comes out of the type mapper, it doesn't know what to do with it.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -902,7 +902,7 @@ class LibraryCommanderImpl @Inject() (
     }
   }
 
-  def getLibrariesUserCanKeepTo(userId: Id[User]): Seq[(Library, LibraryMembership, Set[Id[User]])] = { //ZZZ
+  def getLibrariesUserCanKeepTo(userId: Id[User]): Seq[(Library, LibraryMembership, Set[Id[User]])] = {
```
stkem, line 923: This is unrelated. Snook in here by accident.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaymentMethodRepo.scala
leogrim: I'm currently working on adding `primary` to `UserEmailAddress`, anything I should know so that I can switch to your solution when it's ready?

stkem: Basically just switch out `Option[Boolean]` in your model and table for `TrueOrNull`. Everything else should be taken care of with implicit conversions to and from `Boolean` (i.e. point of use would just use `Boolean` instead of `Option[Boolean]`, which should make that cleaner).

kifi-backend/modules/shoebox/app/com/keepit/payments/PaymentMethod.scala
```diff
@@ -0,0 +1,38 @@
...
+
+case class StripeToken(token: String)
+
+case class PaymentMethod(
+    id: Option[Id[PaymentMethod]] = None,
+    createdAt: DateTime = currentDateTime,
+    updatedAt: DateTime = currentDateTime,
+    state: State[PaymentMethod] = PaymentMethodStates.ACTIVE,
+    accountId: Id[PaidAccount],
+    default: TrueOrNull,
```
leogrim: I'm not a fan of exposing this though, for all purpose this is a `Boolean`. `TrueOrNull` just enables us to have a database constraint, doesn't it? So should we hide this inside of the repo implementation?

stkem: This is the smallest exposure I could come up with. The implicit conversions allow you to just use Booleans at the point of use, i.e. whoever uses the model doesn't really need to know about `TrueOrNull` at all. If you can think of a way to have it just in the repo, without too much Boilerplate, I'm all ears. I'm trying to avoid having to have methods like `applyFromDbRow` and friends just for this.

leogrim: I think @andrewconner mentioned `applyFromDbRow` was actually a good design pattern to have systematically, nope?

stkem: I don't like it very much. Suuuper boilerplaty. You end up effectively spelling out all the fields in you model explicitly _four_ times.

andrewconner: Agreed, but the time to make a repo is tiny compared to maintaining and dealing with eventual migrations. Without applyFromDb, I can't imagine how hard it would be to migrate models like Keep

kifi-backend/modules/shoebox/app/com/keepit/payments/PaymentMethodRepo.scala
andrewconner: I've had a nice type mapper for this, on office now but will add it when I get to a place with internet

kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEvent.scala
```diff
@@ -0,0 +1,203 @@
...
+    case "user_change_credit" => UserChangeCredit()
+    case "user_added" => extras.as[UserAdded]
+    case "user_removed" => extras.as[UserRemoved]
+    case "admin_added" => extras.as[AdminAdded]
+    case "admin_removed" => extras.as[AdminRemoved]
+    case "plan_changed" => extras.as[PlanChanged]
+    case "payment_method_added" => extras.as[PaymentMethodAdded]
+    case "default_payment_method_changed" => extras.as[DefaultPaymentMethodChanged]
+    case "account_contacts_changed" => extras.as[AccountContactsChanged]
+    case _ => throw new Exception(s"Invalid Event Type: $eventType")
```
ryanpbrewster, line 123: I like the strategy we use elsewhere in the codebase:
`case SpecialCredit().eventType => SpecialCredit()`
`case ChargeBack().eventType => ChargeBack()`
but I can see that this has its own merits (for instance, changing the `eventType` string is probably a bad idea because it will lead to issues with existing DB values, and this layout makes that more obvious). I'm not sure where you stand on this.

kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEvent.scala
```diff
@@ -0,0 +1,203 @@
...
+    state: State[AccountEvent] = AccountEventStates.ACTIVE,
+    eventGroup: Long,
+    eventTime: DateTime,
+    accountId: Id[PaidAccount],
+    billingRelated: Boolean,
+    whoDunnit: Option[Id[User]],
+    whoDunnitExtra: JsValue,
+    kifiAdminInvolved: Option[Id[User]],
+    action: AccountEventAction,
+    creditChange: Int, //in cents
```
ryanpbrewster: Do you think it's worth having a type `case class Money(cents: Int)` to make this super explicit everywhere it is used?

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
```diff
@@ -0,0 +1,26 @@
...
+
+import org.joda.time.DateTime
+
+case class PaidPlan(
+    id: Option[Id[PaidPlan]] = None,
+    createdAt: DateTime = currentDateTime,
+    updatedAt: DateTime = currentDateTime,
+    state: State[PaidPlan] = PaidPlanStates.ACTIVE,
+    name: Name[PaidPlan],
+    billingCycle: Int, //in month!
```
ryanpbrewster: Do you think it's worth having a `case class BillingCycle(months: Int)` to make this explicit?

kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEvent.scala
ryanpbrewster, line 123: Actually, I see now that in order to use that you'd have to have the `eventType` be a typed object (like `AccountEventActionType`) and copy all this string matching into there. It does introduce a decent amount of boilerplate, but I am a fan of letting the compiler help catch bugs.

kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEvent.scala
stkem: Yes, it probably is. I'm going to change that.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
stkem: Yes :-) See `Money`.

kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEvent.scala
andrewconner: Even better, it's nice to make a `CurrencyAmount`, which can be constructed using cents and doesn't have a double constructor, but behaves like one.

ryanpbrewster: Oh I like that a lot.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaymentMethodRepo.scala
andrewconner: Here's an idea, I haven't compiled this but it should work and may be a bit more semantic?

``` scala
sealed trait PrimaryOrAuxRecord
object PrimaryOrAuxRecord {
  implicit val primaryOrAuxRecordColumnType = MappedColumnType.base[PrimaryOrAuxRecord, Option[Boolean]](toDb, fromDb)

  def fromDb(b: Option[Boolean]) = {
    b match {
      case Some(true) => PrimaryRecord
      case _ => AuxRecord
    }
  }
  def toDb(p: PrimaryOrAuxRecord) = {
    p match {
      case PrimaryRecord => Some(true)
      case AuxRecord => None
    }
  }
}
object PrimaryRecord extends PrimaryOrAuxRecord
object AuxRecord extends PrimaryOrAuxRecord
```

andrewconner:
Get it in! :+1:

kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEvent.scala
```diff
@@ -0,0 +1,203 @@
...
+  def toDbRow: (String, JsValue)
+}
+
+object AccountEventAction { //There is probably a deeper type hirachy that can be used here...
+
+  trait Payloadless { self: AccountEventAction =>
+    def toDbRow: (String, JsValue) = (eventType, JsNull)
+  }
+
+  case class SpecialCredit() extends AccountEventAction with Payloadless {
```
andrewconner, line 25: `case object`?

kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEvent.scala
andrewconner, line 123: Agreed, the repetition could be avoided.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaymentMethodRepo.scala
stkem: Just had a chance to finally look at this. I tried this already. I'm afraid you can't have a type mapper that maps to an `Option`, it has to be a primitive type that slick already knows about. See previous comment I made to Léo in this exchange.

stkem: I'm giving up on this for now. Spent way too much time trying to get this to work. @andrewconner maybe ask our typesafe contacts (you know, the ones we are paying for) for an idea some time? Not critical at this point, but I would kinda like to know if there is a good way.

kifi-backend/modules/common/conf/evolutions/shoebox/375.sql
```diff
@@ -0,0 +1,86 @@
...
+    UNIQUE INDEX payment_method_u_account_id_default (account_id, default)
+);
+
+
+CREATE TABLE paid_account (
+    id bigint(20) NOT NULL AUTO_INCREMENT,
+    created_at datetime NOT NULL,
+    updated_at datetime NOT NULL,
+    state varchar(20) NOT NULL,
+    org_id bigint(20) NOT NULL,
```
leogrim, line 23: Have we definitely ruled out personal premium plans?

kifi-backend/modules/common/conf/evolutions/shoebox/375.sql
```diff
@@ -0,0 +1,86 @@
...
+    CONSTRAINT paid_account_f_organization FOREIGN KEY (org_id) REFERENCES organization(id)
+);
+
+
+CREATE TABLE paid_plan (
+    id bigint(20) NOT NULL AUTO_INCREMENT,
+    created_at datetime NOT NULL,
+    updated_at datetime NOT NULL,
+    state varchar(20) NOT NULL,
+    name varchar(256) NOT NULL,
```
leogrim, line 39: Is this `name` a key? Would it be better / more convenient to use it for foreign keys that `id`? It would make other models more readable.

kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEvent.scala
```diff
@@ -0,0 +1,205 @@
...
+    Some((e.id, e.createdAt, e.updatedAt, e.state, e.eventGroup, e.eventTime, e.accountId,
+      e.billingRelated, e.whoDunnit, e.whoDunnitExtra, e.kifiAdminInvolved, eventType,
+      extras, e.creditChange, e.paymentMethod, e.paymentCharge, e.memo))
+  }
+}
+
+object AccountEventStates extends States[AccountEvent] {
+  val PENDING = State[AccountEvent]("pending")
+  val PROCESSING = State[AccountEvent]("processing")
+  val FAILED = State[AccountEvent]("failed")
```
leogrim, line 204: Do we need the five states? If you do want custom states (I'm not a fan, but let's say sometimes we do), do you need `ACTIVE` and `INACTIVE`? 

Could we stick to either (`ACTIVE`, `INACTIVE`) or a custom set of states? I think problems arise when both are mixed, and `ACTIVE` and `INACTIVE` are no longer reverse of each other.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidAccountRepo.scala
```diff
@@ -0,0 +1,48 @@
...
+
+  implicit val dollarAmountColumnType = MappedColumnType.base[DollarAmount, Int](_.cents, DollarAmount(_))
+
+  type RepoImpl = PaidAccountTable
+  class PaidAccountTable(tag: Tag) extends RepoTable[PaidAccount](db, tag, "paid_Account") {
+    def orgId = column[Id[Organization]]("org_id", O.NotNull)
+    def planId = column[Id[PaidPlan]]("plan_id", O.NotNull)
+    def credit = column[DollarAmount]("credit", O.NotNull)
+    def userContacts = column[Seq[Id[User]]]("user_contacts", O.NotNull)
+    def emailContacts = column[Seq[EmailAddress]]("email_contacts", O.NotNull)
```
leogrim, line 33: What's the benefit of having email addresses in here in addition to users? 
If we do associate several users to a `PaidAccount`, any reason we're denormalizing them as opposed to having another table?

kifi-backend/modules/shoebox/app/com/keepit/payments/PaymentMethodRepo.scala
```diff
@@ -0,0 +1,44 @@
...
+
+  import com.keepit.common.db.slick.DBSession._
+  import db.Driver.simple._
+
+  implicit val stripeTokenColumnType = MappedColumnType.base[StripeToken, String](_.token, StripeToken(_))
+
+  type RepoImpl = PaymentMethodTable
+  class PaymentMethodTable(tag: Tag) extends RepoTable[PaymentMethod](db, tag, "paid_plan") {
+    def accountId = column[Id[PaidAccount]]("account_id", O.NotNull)
+    def default = column[Boolean]("default", O.Nullable)
```
andrewconner, line 28: `column[Option[Boolean]]("default", O.Nullable)` is preferred here (this actually causes problems if you try to do interpolated queries)

kifi-backend/modules/common/conf/evolutions/shoebox/375.sql
andrewconner, line 23: I think this is fine, we'll add in personal plans later if we want to. Shouldn't be too insane to refactor if we need to. @stkem?

kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEvent.scala
leogrim, line 204: But really, I think we should rather have a separate column tracking this piece of business logic. `State` should just track whether a row is deleted or not.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaymentMethodRepo.scala
andrewconner, line 28: Wasn't this going to use the `PrimaryOrAuxRecord` (or whatever name you decided)?

stkem, line 28: Not reading your email? There was a longer conversation on this pull request (that you were explicitly tagged in). It doesn't work.

kifi-backend/modules/common/conf/evolutions/shoebox/375.sql
stkem, line 23: Yup, that was kind of the idea behind having a separate model for `PaidAccount` instead of having these things be built into orgs. This model is pretty much entirely contained within the payment system (i.e. it does not appear in any outward facing APIs). It would be quite easy to change this to an `Either[Id[Organization], Id[User]]` (at the scala level, it would be two columns - or possibly another table - in the db)

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidAccountRepo.scala
stkem, line 33: The idea is that you can have non user email addresses as billing contacts (slack supports this), so you can, for example, have bills automatically sent to your (outside) accountant.
Also, this is not a denormalization. The information which contacts are associated with which account is not stored anywhere else.

kifi-backend/modules/common/conf/evolutions/shoebox/375.sql
stkem, line 39: Being cleared up on the other PR.

kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEvent.scala
stkem, line 204: I disagree with this. These are two different mental models of what state means. To take the example of `PaidPlan`, if we break it out there would be a `state` column (with only `ACTIVE` and `INACTIVE` and another, let’s call it `kind` for now, although I think it’s really a kind of `state`, which would have possible `CUSTOM`, `NORMAL` and `GRANDFATHERED`. Then upon use, you’d have to check `state==`ACTIVE`and`kind ==…`. That makes it more error prone, not less, imho.

PR #16642, opened 2015-08-06
Fix height of Create Library icon-card

cvializ:
I made a change that made the icon-card square (20px by 20px), but it needs to be library-card dimensions (20px by 28px).


cvializ:
@adamjgrant @andrewconner

adamjgrant:
Cool

PR #16646, opened 2015-08-06
Bugfix/non member owner library

adamjgrant:


kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.js
```diff
@@ -349,6 +304,52 @@ angular.module('kifi')
...
+          }
+          else {
+            // If user unclicks "organization" their destination should be
+            // their current profile.
+            scope.space.destination = scope.me;
+          }
+          onChangeSpace();
+        };
+
+        var onChangeSpace = function () {
```
adamjgrant, line 339: This has been made into its own function instead of `scope.onChangeSpace` as it no longer needs to be called from the view.

kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.js
```diff
@@ -349,6 +304,52 @@ angular.module('kifi')
           scope.library.visibility = 'organization';
         }
```
adamjgrant, line 321: Below: Pushing similar logic into the same code block for easier maintainability

kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.js
```diff
@@ -316,6 +279,7 @@ angular.module('kifi')
             'slug': '',
             'visibility': 'published'
           };
+          scope.library.org = scope.modalData.organization;
```
adamjgrant, line 285: `library.org` immediately gets set to `scope.modalData.organization` if it exists, so we are no longer using two identities of the same thing elsewhere in the code.

kifi-backend/modules/shoebox/angular/src/libraries/manageLibrary.js
```diff
@@ -187,8 +147,12 @@ angular.module('kifi')
...
           });
 
-          var owner = {};
-          owner[ownerType(scope.space.destination)] = scope.space.destination.id;
+          // Create an owner object that declares the type (user/org) for backend.
+          var owner;
+          if (scope.space.current.id !== scope.space.destination.id) {
+            owner = {};
+            owner[ownerType(scope.space.destination)] = scope.space.destination.id;
+          }
```
adamjgrant, line 155: The purpose of this is to make sure `{ space: undefined }` is sent if the user is not moving the library. Previously it would try to move the library to the same space it's already in. Because the user isn't permissioned to move into the library, the backend would deny access.

adamjgrant:
Ready to go @cvializ

cvializ:
everything looks kosher. ship it!

PR #16648, opened 2015-08-06
Close click menu when user clicks child of menu item

cvializ:
@adamjgrant (@andrewconner)


adamjgrant:
Click menus are tricky things, especially on one-page apps

PR #16658, opened 2015-08-06
Revert update

camhashemi:
@ryanpbrewster with the leaping catch at the wall


ryanpbrewster:
It's the only thing left that makes me feel alive.

PR #16663, opened 2015-08-06
Temporary fix to the non-deterministic test

ryanpbrewster:
Just for a bit, so the test stops failing while I improve the logic.


ryanpbrewster:
@camhashemi

PR #16664, opened 2015-08-06
Temporarily added a hack for org experiments

ryanpbrewster:
New users are automatically added to the org experiment if they
have any organization invite for their email address

@eishay


eishay:
lgtm

eishay:
ship it!

PR #16666, opened 2015-08-06
cloning an image of a library when cloning the lib

eishay:


eishay:
@ryanpbrewster

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminLibraryController.scala
```diff
@@ -331,6 +332,8 @@ class AdminLibraryController @Inject() (
         case Some(membership) =>
           libraryMembershipRepo.save(membership.copy(libraryId = lib.id.get, seq = SequenceNumber.ZERO))
       }
+      val image = libraryImageRepoImpl.getActiveForLibraryId(origLib.id.get).head
+      libraryImageRepoImpl.save(image.copy(id = None, libraryId = lib.id.get))
```
ryanpbrewster, line 336: :+1:

PR #16667, opened 2015-08-06
Fixing new lib creation bug introduced by refactoring

adamjgrant:


adamjgrant:
@cvializ

cvializ:
That should do it :+1:

PR #16671, opened 2015-08-07
Make LibraryCommanderTest actually test sort order

ryanpbrewster:
Added a very adversarial tests for sort order.

Also, I'm not sure that sorting by `createdAt` is a good idea. I wish we had
a different metric to sort invites. I personally am a fan of alphabetical.

@camhashemi

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/model/LibraryInviteRepo.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/model/LibraryInviteRepo.scala
@@ -23,7 +23,7 @@ trait LibraryInviteRepo extends Repo[LibraryInvite] with RepoWithDelete[LibraryI
   def getByEmailAddress(email: EmailAddress, excludeStates: Set[State[LibraryInvite]])(implicit session: RSession): Seq[LibraryInvite]
   def getByLibraryIdAndAuthToken(libraryId: Id[Library], authToken: String, includeSet: Set[State[LibraryInvite]] = Set(LibraryInviteStates.ACTIVE))(implicit session: RSession): Seq[LibraryInvite]
   def pageInviteesByLibraryId(libraryId: Id[Library], offset: Int, limit: Int, includeStates: Set[State[LibraryInvite]])(implicit session: RSession): Seq[(Either[Id[User], EmailAddress], Set[LibraryInvite])]
-  def getByLibraryIdAndInviterId(libraryId: Id[Library], inviterId: Id[User], includeStates: Set[State[LibraryInvite]])(implicit session: RSession): Seq[LibraryInvite]
+  def getByLibraryIdAndInviterId(libraryId: Id[Library], inviterId: Id[User], includeStates: Set[State[LibraryInvite]] = Set(LibraryInviteStates.ACTIVE))(implicit session: RSession): Seq[LibraryInvite]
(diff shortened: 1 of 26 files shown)


kifi-backend/modules/shoebox/app/com/keepit/model/LibraryInviteRepo.scala
```diff
@@ -139,9 +139,9 @@ class LibraryInviteRepoImpl @Inject() (
...
     val invitees = {
       val invitesGroupedByInvitee = (for (r <- rows if r.libraryId === libraryId && r.state.inSet(includeStates)) yield r).groupBy(r => (r.userId, r.emailAddress))
-      val inviteesWithLastInvitedAt = invitesGroupedByInvitee.map { case ((userId, emailAddress), invites) => (userId, emailAddress, invites.map(_.createdAt).min) }
-      val sortedInvitees = inviteesWithLastInvitedAt.sortBy { case (userId, emailAddress, firstInvitedAt) => (userId.isDefined, firstInvitedAt) }
-      sortedInvitees.drop(offset).take(limit).map { case (userId, emailAddress, firstInvitedAt) => (userId, emailAddress) }.list
+      val inviteesWithLastInvitedAt = invitesGroupedByInvitee.map { case ((userId, emailAddress), invites) => (userId, emailAddress, invites.map(_.id).min) }
+      val sortedInvitees = inviteesWithLastInvitedAt.sortBy { case (userId, emailAddress, lowestId) => (userId.isDefined, lowestId) }
+      sortedInvitees.drop(offset).take(limit).map { case (userId, emailAddress, lowestId) => (userId, emailAddress) }.list
```
ryanpbrewster, line 144: If someone were gonna object to this PR, it would be here.

kifi-backend/modules/shoebox/test/com/keepit/commanders/LibraryCommanderTest.scala
```diff
@@ -50,6 +52,29 @@ class LibraryCommanderTest extends TestKitSupport with SpecificationLike with Sh
...
+      val followers = UserFactory.users(20).saved
+      val invitedUsers = UserFactory.users(20).saved
+      val invitedEmails = randomEmails(20)
+      val org = OrganizationFactory.organization().withOwner(orgOwner).withMembers(libOwners ++ collaborators).withInvitedUsers(followers).saved
+      for (lo <- libOwners) {
+        LibraryFactory.library().withOwner(lo).withCollaborators(collaborators).withFollowers(followers).withInvitedUsers(invitedUsers).withInvitedEmails(invitedEmails).saved
+        LibraryFactory.library().withOwner(lo).withCollaborators(collaborators).withFollowers(followers).withInvitedUsers(invitedUsers).withInvitedEmails(invitedEmails).withOrganization(org).saved
+      }
+    }
+  }
```
ryanpbrewster, line 76: I think this is a good thing to have around, but I'm not sure it's necessary for most tests. It does slow the tests down by a good amount. I think good practice would be to use this for any repo method that uses an interpolated query.

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryInviteRepo.scala
camhashemi, line 144: I'm just admiring your usage of Slick for the moment.

Wouldn't we want to use `highestId` to get the most recent invite?

ryanpbrewster, line 144: Yup. Right now they're being sorted oldest-first. I'll maintain that for now. I agree that newest-first seems more reasonable.

PR #16672, opened 2015-08-07
Marketing v8, no public release

andrewconner:
@JRuff42 

Btw @eishay, if I'm unavailable, this is how you create a _new_ site version. If you just want to pull in an edit, no Scala changes needed. If you want to change how traffic flows, check line 77 in MarketingSiteRouter. Simple enough.

If they add any _new_ pages, make sure shoeboxSite.router knows about it, or it'll 404.

This is a yolo PR, my branch is way behind master so it's compiling the world and I've been up for 19 hours :)

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/MarketingSiteRouter.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/MarketingSiteRouter.scala
@@ -62,7 +62,8 @@ object MarketingSiteRouter extends AssetsBuilder with Controller with Logging {
   private object Version5 extends LandingVersion { val version = 5 }
   private object Version6 extends LandingVersion { val version = 6 }
   private object Version7 extends LandingVersion { val version = 7 }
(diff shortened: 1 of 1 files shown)


eishay:
Thanks @andrewconner

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/MarketingSiteRouter.scala
```diff
@@ -62,7 +62,8 @@ object MarketingSiteRouter extends AssetsBuilder with Controller with Logging {
   private object Version5 extends LandingVersion { val version = 5 }
   private object Version6 extends LandingVersion { val version = 6 }
   private object Version7 extends LandingVersion { val version = 7 }
-  private val versions = Seq(Version1, Version2, Version3, Version4, Version5, Version6, Version7)
+  private object Version8 extends LandingVersion { val version = 7 }
```
eishay, line 65: fixing `{ val version = 7 }` -> `{ val version = 8 }`

PR #16675, opened 2015-08-07
Clear the library list before when loading a new set of libs

cvializ:
@adamjgrant simple fix, I probably introduced this bug when factoring-out the paginator functionality to share it with the orgProfile page


adamjgrant:
:+1:

PR #16678, opened 2015-08-07
Showing invite member button only to users with admin experiment

adamjgrant:


adamjgrant:
@cvializ

kifi-backend/modules/shoebox/angular/src/orgProfile/orgProfileMemberManage.tpl.html
```diff
@@ -1,4 +1,4 @@
 <ul class="kf-org-member-manage" smart-scroll scroll-distance="'100%'" scroll-disabled="!hasMoreMembers()" scroll-next="fetchMembers()">
-  <!-- <li ng&#45;if="canInvite"><button class="kf&#45;org&#45;member&#45;invite" ng&#45;click="openInviteModal()">Invite new member</button></li> -->
+  <li ng-if="canInvite && me.experiments.indexOf('admin') > -1"><button class="kf-org-member-invite" ng-click="openInviteModal()">Invite new member</button></li>
```
cvializ, line 2: so we're only showing this button to Kifi people, not just anyone with the Org experiment, is that correct?

adamjgrant, line 2: Correct.

cvializ:
:+1: ship it

PR #16680, opened 2015-08-07
admin to filter out removed orgs

eishay:


ryanpbrewster:
We should also change the query so that it only retrieves active orgs.

PR #16683, opened 2015-08-07
Made LibController.addLibrary return a CardInfo

ryanpbrewster:
I don't know why it was returning a FullLibraryInfo, or why in God's name
we have so many different ways of serializing a library. I am not a happy
panda.

@cvializ


kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -1548,6 +1549,11 @@ class LibraryCommanderImpl @Inject() (
     } getOrElse Future.successful(Seq.empty)
   }
 
+  def createLibraryCardInfo(lib: Library, owner: User, viewerOpt: Option[User], withFollowing: Boolean, idealSize: ImageSize): LibraryCardInfo = {
+    db.readOnlyMaster { implicit session =>
+      createLibraryCardInfos(Seq(lib), Map(owner.id.get -> BasicUser.fromUser(owner)), viewerOpt, withFollowing, idealSize).head
+    }
+  }
   def createLibraryCardInfos(libs: Seq[Library], owners: Map[Id[User], BasicUser], viewerOpt: Option[User], withFollowing: Boolean, idealSize: ImageSize)(implicit session: RSession): ParSeq[LibraryCardInfo] = {
```
ryanpbrewster, line 1557: I don't know why this method needs you to hold its hand and provide the owners when it's going to use an open DB session to get a bunch of other stuff anyways. I am sad.

PR #16688, opened 2015-08-07
Adding hover state for profile widget

adamjgrant:
@cvializ

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/profile/profileWidget.styl
+++ b/kifi-backend/modules/shoebox/angular/src/profile/profileWidget.styl
@@ -78,11 +78,10 @@
     margin-top: 15px
     justify-content: center
 
-    tr
+    .stats-row
       &:hover
-        td:last-child
-          .stat
-            text-decoration: underline
+        .stat-title
+          text-decoration: underline
 
     .stat
       width: 33%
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/angular/src/profile/profileWidget.styl
```diff
@@ -78,15 +78,14 @@
...
-    .stat
-      width: 33%
-
+    tr
       &:hover
-        h2
-          text-decoration: underline
+        td:last-child
+          .stat
+            text-decoration: underline
```
cvializ, line 85: I think long-term that giving class names to the `td` elements is the most maintainable way to style these elements, but this works

cvializ:
:+1:

kifi-backend/modules/shoebox/angular/src/profile/profileWidget.styl
adamjgrant, line 85: It's probably a better strategy for consistency of this application. I think outside of it, rarer class names enforces better namespacing and leads to cleaner markup.

adamjgrant, line 85: For the first point, I'll take another look at it and add some classes

PR #16697, opened 2015-08-08
Add a helpful video to the Move a Library card.

cvializ:
@adamjgrant


adamjgrant:
Very helpful @cvializ

PR #16702, opened 2015-08-10
Added a more robust test for org member listing

ryanpbrewster:
@camhashemi

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/model/OrganizationFactory.scala
+++ b/kifi-backend/modules/common/app/com/keepit/model/OrganizationFactory.scala
@@ -18,6 +18,7 @@ object OrganizationFactory {
 
   case class PartialOrganization private[OrganizationFactory] (
       org: Organization,
+      admins: Seq[User] = Seq.empty[User],
       members: Seq[User] = Seq.empty[User],
       invitedUsers: Seq[User] = Seq.empty[User],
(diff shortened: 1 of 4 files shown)


camhashemi:
@ryanpbrewster do these pass? Thanks. There's a bug in the `getMembersAndUniqueInvitees` method, because anonymous invites are considered `active` invites (they are retrieved from the repo and then filtered out in the commander, we should be filtering in the repo).

I was in the middle of adding an `ignoreAnonymous` boolean to the repo method before invites took over, but it needs fixing. Alternatively, we can consider anonymous invites as `accepted` or `inactive` as soon as they're created. That way, we don't have `pending` invites with no invitee.

ryanpbrewster:
Right now they are only testing members. No invite logic yet.

I'm a big fan of these tests, and if you want to add one for invites I'd be thrilled. You can look at `LibraryCommanderTest` for `"get library invites"` for a test (and associated helper methods) that I wrote for library invites.

camhashemi:
@ryanpbrewster still curious to your opinion on the anonymous invite solution

> I was in the middle of adding an ignoreAnonymous boolean to the repo method before invites took over, but it needs fixing. Alternatively, we can consider anonymous invites as accepted or inactive as soon as they're created. That way, we don't have pending invites with no invitee.

ryanpbrewster:
For now, I'd say we should just ignore those when displaying org invites. Maybe we can add a specific query to find only the anonymous invites and the frontend can display those separately.

PR #16705, opened 2015-08-10
[WIP] Org nonuser signup

camhashemi:
@andrewconner @leogrim @ryanpbrewster @stkem 

the goal of this is to quickly build off of the library non-user invite flow, using the auth token to grant view access to a non-user, sign up through a modal, and auto-join them to the org.

I still need to get the front-end logic completed, but I have a couple backend tests confirming that a user can see an org if they have a valid auth token, and when they sign up they are auto-joined into the org.

PR #16708, opened 2015-08-10
Nonuser signup backend

camhashemi:
@ryanpbrewster @leogrim @stkem looking to deploy the back-end first, test the front-end in dev mode (I need the permission-logic from the server to do that), then deploy the front-end.


camhashemi:
this PR is the back-end subset of #16705

PR #16709, opened 2015-08-10
Update move card for design consistency.

cvializ:
| Old | New |
| --- | --- |
| screen shot 2015-08-10 at 11 33 08 am | screen shot 2015-08-10 at 11 32 52 am |

@adamjgrant

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/orgProfile/orgProfileLibraries.styl
+++ b/kifi-backend/modules/shoebox/angular/src/orgProfile/orgProfileLibraries.styl
@@ -1,7 +1,7 @@
 .kf-org-profile-libraries
   margin: 15px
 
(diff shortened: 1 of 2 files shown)


adamjgrant:
Looks good, but the drop shadow is a bit offputting. I think we might have that in the follow buttons, but this is because they overlay with images (whose color mapping we don't know). Could we do it without the shadow?

adamjgrant:
Looks a lot better with the text smaller.

cvializ:
Right that makes sense, I'll take out the drop shadow. Yeah when I saw you make the text smaller, I thought it was a big improvement.

cvializ:
Now with 100% less box shadow.

screen shot 2015-08-10 at 11 43 12 am

PR #16710, opened 2015-08-10
Add a public share link textbox to published libraries' pages.

cvializ:
screen shot 2015-08-10 at 11 40 25 am

@adamjgrant


adamjgrant:
Nice! Anyway we could make the text match the size of the text on the buttons below it?

cvializ:
Good call, I'll update it

adamjgrant:
:+1:

PR #16714, opened 2015-08-10
pass auth token from UserOrOrgController to OrgController

camhashemi:
@ryanpbrewster @andrewconner assumed that the original request from `UserOrOrgController` was passed along implicitly (is this desired behavior?) to `OrganizationController`, but it wasn't.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/OrganizationMembershipController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/OrganizationMembershipController.scala
@@ -26,7 +26,7 @@ class OrganizationMembershipController @Inject() (
     implicit val publicIdConfig: PublicIdConfiguration) extends UserActions with OrganizationAccessActions with ShoeboxServiceController {
 
   // If userIdOpt is provided AND the user can invite members, return invited users as well as members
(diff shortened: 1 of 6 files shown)


kifi-backend/modules/shoebox/app/com/keepit/controllers/website/OrganizationMembershipController.scala
```diff
@@ -26,7 +26,7 @@ class OrganizationMembershipController @Inject() (
     implicit val publicIdConfig: PublicIdConfiguration) extends UserActions with OrganizationAccessActions with ShoeboxServiceController {
 
   // If userIdOpt is provided AND the user can invite members, return invited users as well as members
-  def getMembers(pubId: PublicId[Organization], offset: Int, limit: Int) = OrganizationAction(pubId, OrganizationPermission.VIEW_ORGANIZATION) { request =>
+  def getMembers(pubId: PublicId[Organization], offset: Int, limit: Int) = OrganizationAction(pubId, None, OrganizationPermission.VIEW_ORGANIZATION) { request =>
```
ryanpbrewster, line 29: Why spell out the authToken explicitly here?

ryanpbrewster, line 29: It is because `permissions` has a variable number of arguments?

camhashemi, line 29: did you mean why _not_ spell it out? I should actually, forgot to.

PR #16716, opened 2015-08-10
sort org libraries by lastKept

camhashemi:
@ryanpbrewster per https://app.asana.com/0/42816489233171/44917153045323/f

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/model/LibraryRepo.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/model/LibraryRepo.scala
@@ -472,7 +472,7 @@ class LibraryRepoImpl @Inject() (
   }
 
   def getVisibleOrganizationLibraries(orgId: Id[Organization], includeOrgVisibleLibraries: Boolean, viewerLibraryMemberships: Set[Id[Library]], offset: Offset, limit: Limit)(implicit session: RSession): Seq[Library] = {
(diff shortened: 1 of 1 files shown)


ryanpbrewster:
I really think this is not the correct decision.

camhashemi:
because

ryanpbrewster:
I think that if we're going to sort a list of libraries, it should be to make it as easy as possible to find a specific item. Sorting by most recent keep is neat, because you can naturally see active libraries and hide inactive libraries, but it is not helpful at all if you're trying to find something.

I just chatted with @eishay about it, so he can chime in.

ryanpbrewster:
In the end, Kifi is supposed to enhance productivity. It should be _fast_ and _simple_, even at the expensive of everything else (aesthetics, flexibility, etc).

We've already failed if we have a user scrolling through their libraries, because that's slow. We'd be way better off if they could just find what they're looking for via Search. However, we enter nightmare territory if they're scrolling through their libraries and the libraries are sorted in a way that doesn't help them. I know that on several occasions I have tried looking through my libraries for a specific entity and given up after 30 seconds of frustrating futility. I tend to fall back on scrolling down until libraries stop loading, then using Command+F to text-search for what I want. I found that to be a pretty negative UX.

camhashemi:
You should chat with Jen, as she created the card. We already have search for finding specific libraries, but I don't think it is a big deal either way. Creating a "sort by" drop-down would solve any sorting disagreements we may have and cater to the user's needs.

ryanpbrewster:
Go ahead and merge. We'll figure out the super-involved sorting strategies later. For now this will do. Guess I'd better brush up on my search skills.

PR #16720, opened 2015-08-10
Notification delivery

maxov:
@ryanpbrewster


kifi-backend/modules/common/app/com/keepit/notify/model/NotificationKind.scala
```diff
@@ -14,6 +14,30 @@ trait NotificationKind[N <: NotificationEvent] {
...
+   * of events from the database to find the right group can be expensive. In addition, events like these do not require
+   * advanced grouping behavior and only rely on a few external ids. Therefore, using [[groupIdentifier]] only requires
+   * a simple WHERE sql clause instead of a whole bunch of deserialization.
+   *
+   * It is implied that all notification events with the same group identifier also have the same notification id.
+   *
+   * @param event The event to find the identifier for
+   * @return [[Some]] with the identifier if the identifier exists, [[None]] otherwise
+   */
+  def groupIdentifier(event: N): Option[String] = None
```
maxov, line 31: @ryanpbrewster this chain of reasoning make sense?

kifi-backend/modules/common/app/com/keepit/notify/info/NotificationInfoGenerator.scala
```diff
@@ -0,0 +1,73 @@
...
+    }.flatMap {
+      case (returnsInfo, result) => run(returnsInfo, result.args, result.pickedOne)
+    }
+  }
+
+  private def getArgs[E](args: Args, keyIn: String, pickedOneIn: Boolean): Option[E] = for {
+    picked <- Option(pickedOneIn)
+    if picked
+    key <- Option(keyIn)
+    if key != ""
```
andrewconner: `key.nonEmpty`

kifi-backend/modules/common/app/com/keepit/notify/info/NotificationInfoGenerator.scala
```diff
@@ -0,0 +1,73 @@
...
+    res <- args.get(key)
+  } yield {
+    res.asInstanceOf[E]
```
andrewconner: Ew, I hope this one can be avoided.

kifi-backend/modules/common/test/com/keepit/notify/model/NotificationKindTest.scala
```diff
@@ -11,7 +13,7 @@ class NotificationKindTest extends Specification {
 
     "serialize and deserialize to json with event properly" in {
 
-      val event = DepressedRobotGrumble(Id(1), currentDateTime, "marvin", "life, the universe, and everything")
+      val event = DepressedRobotGrumble(Recipient(Id[User](1)), currentDateTime, "marvin", "life, the universe, and everything")
```
andrewconner, line 16: :laughing:

kifi-backend/modules/common/app/com/keepit/common/path/Path.scala
```diff
@@ -35,6 +37,18 @@ object Path {
...
+  def encode(value: String): EncodedPath = new EncodedPath(value)
+
+  def alreadyEncoded(value: String): EncodedPath = new EncodedPath(URLDecoder.decode(value, UTF8))
+
+  implicit val format: Format[Path] = Format(
+    __.read[String].map(str => alreadyEncoded(str.substring(base.length))),
+    Writes(path => JsString(path.encode.absolute))
+  )
+
+  implicit val encodedFormat: Format[EncodedPath] = Format(
```
stkem, line 49: I'm not sure this does what you want, i.e. make this implicitly available whenever you have a `EncodedPath` instance, since it's in the companion to `Path`. Unless I'm missing something regarding the inheritance?

maxov, line 49: Hmm, I thought that implicits were looked for not only in the companion object of the type in question, but in the companion objects of its supertypes. I should test that to see if it's true, but I haven't actually gotten any compile errors related to this.

kifi-backend/modules/common/app/com/keepit/notify/info/NeedsInfo.scala
```diff
@@ -0,0 +1,88 @@
...
+  case class NeedLibrary(id: Id[Library], arg: String) extends NeedsInfo[Library]
+  case class NeedUserImage(id: Id[User], arg: String) extends NeedsInfo[String]
+
+  def user(id: Id[User], arg: String = ""): NeedsInfo[User] = NeedUser(id, arg)
+  def library(id: Id[Library], arg: String = ""): NeedsInfo[Library] = NeedLibrary(id, arg)
+  def userImage(id: Id[User], arg: String = ""): NeedsInfo[String] = NeedUserImage(id, arg)
+
+}
+
+trait UsingDsl {
```
stkem: I'm not sure a DSL / super high abstraction level is necessarily a good idea for this and perhaps a bit overkill, from a maintainability perspective. It's quite elegant in places, but will be hard to learn and understand fully for the next person who is going to work on this.

kifi-backend/modules/common/app/com/keepit/common/path/Path.scala
maxov, line 49: After some playing around in the repl:

``` scala
scala> trait BlahBlah[A] { def get: Int }
defined trait BlahBlah

scala> :paste
// Entering paste mode (ctrl-D to finish)

class MyObj
class MySubtype extends MyObj

object MyObj {
implicit val blah1 = new BlahBlah[MyObj] { def get = 1 }
implicit val blah2 = new BlahBlah[MySubtype] { def get = 2 }
}

// Exiting paste mode, now interpreting.

defined class MyObj
defined class MySubtype
defined object MyObj

scala> implicitly[BlahBlah[MyObj]]
res2: BlahBlah[MyObj] = MyObj$$anon$1@5a8806ef

scala> implicitly[BlahBlah[MyObj]].get
res3: Int = 1

scala> implicitly[BlahBlah[MySubtype]].get
res4: Int = 2
```

stkem, line 49: Ah nice :-)
Forget I said anything :-)

kifi-backend/modules/common/app/com/keepit/notify/info/NeedsInfo.scala
maxov: I would mostly agree with you, but in this case requests for information need to be abstracted out so that they can be batched (otherwise, the number of service calls to Shoebox could be prohibitively expensive). I couldn't find a way to do that without this kind of DSL, I tried to keep it as simple and easy to use as possible. I'm also going to replace the current HList implementation with a simple object map, something easier to understand.

ryanpbrewster:
This PR is getting pretty huge. Can we merge it (or part of it) before it grows any bigger?

maxov:
@ryanpbrewster Yeah, I plan to merge it today. Need to finish up info generation

kifi-backend/modules/common/app/com/keepit/model/UserExperiment.scala
```diff
@@ -75,13 +75,15 @@ object UserExperimentType {
   val APPLY_RECO_FEEDBACK = UserExperimentType("apply_reco_feedback")
   val SEARCH_LAB = UserExperimentType("search_lab")
   val READ_IT_LATER = UserExperimentType("read_id_later")
+  val NEW_NOTIFS_SYSTEM = UserExperimentType("new_notifs_system")
```
ryanpbrewster, line 78: Is this to indicate that there are new system notifications? If so, maybe `NEW_SYSTEM_NOTIFICATIONS` is a good name.

kifi-backend/modules/common/app/com/keepit/notify/info/NeedsInfo.scala
```diff
@@ -0,0 +1,72 @@
+package com.keepit.notify.info
+
+import com.keepit.common.db.Id
+import com.keepit.model.{ Keep, Library, User }
+import com.keepit.notify.model.{EventArgs, GenEventArgs, NotificationEvent}
+import play.api.libs.functional.{ ~, FunctionalCanBuild }
+
+import scala.concurrent.{ ExecutionContext, Future }
+
+sealed trait NeedsInfo[+A]
```
ryanpbrewster, line 10: All the case classes are singular (NeedUser, NeedLibrary) but the trait is plural (NeedsInfo). Any reason?

kifi-backend/modules/common/app/com/keepit/notify/info/NeedsInfo.scala
```diff
@@ -0,0 +1,72 @@
...
+
+trait PossibleNeeds {
+
+  case class NeedUser(id: Id[User]) extends NeedsInfo[User]
+  case class NeedLibrary(id: Id[Library]) extends NeedsInfo[Library]
+  case class NeedUserImage(id: Id[User]) extends NeedsInfo[String]
```
ryanpbrewster, line 16: ImagePath maybe?

kifi-backend/modules/common/app/com/keepit/notify/info/NeedsInfo.scala
```diff
@@ -0,0 +1,72 @@
...
+  case class NeedKeep(id: Id[Keep]) extends NeedsInfo[Keep]
+  case class NeedLibraryUrl(id: Id[Library]) extends NeedsInfo[String]
+
+  def user(id: Id[User]): NeedsInfo[User] = NeedUser(id)
+  def library(id: Id[Library]): NeedsInfo[Library] = NeedLibrary(id)
+  def userImage(id: Id[User]): NeedsInfo[String] = NeedUserImage(id)
+  def keep(id: Id[Keep]): NeedsInfo[Keep] = NeedKeep(id)
+  def libraryUrl(id: Id[Library]): NeedsInfo[String]
```
ryanpbrewster, line 24: LibraryPath or Path or something?

ryanpbrewster, line 24: Also, why doesn't this have a definition?

kifi-backend/modules/common/app/com/keepit/notify/info/NeedsInfo.scala
```diff
@@ -0,0 +1,72 @@
...
+  object Results {
+    def unapply[E <: NotificationEvent](that: Fetched[E]): Option[EventArgs[E]] = Some(that.results)
+  }
+
+  case class Filled[A](elem) extends NeedsInfo[A]
+
+}
+
+object NeedsInfo extends PossibleNeeds with UsingDsl with ArgsDsl
+
```
ryanpbrewster, line 72: This trait is hard for me to understand. I'm not sure the additional extensibility is worth the cost of this complexity.

kifi-backend/modules/common/app/com/keepit/notify/info/NotificationInfo.scala
```diff
@@ -0,0 +1,27 @@
...
+  implicit val format = (
+    (__ \ "url").format[String] and
+    (__ \ "imageUrl").format[String] and
+    (__ \ "title").format[String] and
+    (__ \ "body").format[String] and
+    (__ \ "linkText").format[String] and
+    (__ \ "extraJson").formatNullable[JsObject]
+  )(NotificationInfo.apply, unlift(NotificationInfo.unapply))
+
+}
```
ryanpbrewster, line 27: One thing I wish we did more was leave comments in models. What are the things we assume about Notifications right now? It looks like we assume that in order to display a Notification we need to have at the very least:
1. A URL that the user gets directed to if they click the notification
2. An image to be displayed
3. A title to be displayed
4. Text to put in the Notification body
5. Link text? I'm surprised by that.

Everything else is extra, and probably depends on the kind of notification.

Is that right? If so, I feel like a comment to that effect would be very useful in helping another engineer get their bearings.

kifi-backend/modules/common/app/com/keepit/notify/model/EventArgs.scala
```diff
@@ -0,0 +1,20 @@
+package com.keepit.notify.model
+
+import com.keepit.notify.info.{NeedsInfo, Args}
+
+case class EventArgs[E <: NotificationEvent](event: NotificationEvent, argsMap: Args = Map()) {
+
+  def args(args: (String, Any)*) = copy(argsMap = this.argsMap ++ args)
```
ryanpbrewster, line 7: I'm always scared by name shadowing. Can we use a variable name that doesn't overlap with the method name?

kifi-backend/modules/common/app/com/keepit/notify/model/Recipient.scala
```diff
@@ -0,0 +1,63 @@
...
+    kind match {
+      case "user" => UserRecipient(Id(rest.toLong))
+      case "email" => EmailAddress.validate(rest).map(EmailRecipient.apply).get
+      case _ => throw new IllegalArgumentException(s"${str} does not have a valid kind")
+    }
+  }
+
+  def apply[T](obj: T)(implicit evidence: T => AsRecipient[T]) = evidence(obj).recipient
+
+  def unapply(recip: Recipient): Option[String] = Some(s"${recip.kind}|${recip.toString}")
```
ryanpbrewster, line 40: Clever.

kifi-backend/modules/common/app/com/keepit/model/UserExperiment.scala
maxov, line 78: It indicates that the user is 'on' the new notifications system.

kifi-backend/modules/common/app/com/keepit/notify/info/NeedsInfo.scala
maxov, line 10: Fixed in 87983a57b

kifi-backend/modules/common/app/com/keepit/notify/info/NeedsInfo.scala
maxov, line 24: Fixed in 87983a57b

kifi-backend/modules/common/app/com/keepit/notify/model/EventArgs.scala
maxov, line 7: Fixed in c498a96d

kifi-backend/modules/common/app/com/keepit/notify/info/NeedsInfo.scala
maxov, line 72: This is the simplest I could get the "information request" system, perhaps seeing [usage](https://github.com/kifi/keepit/blob/27ac6fa95215805a38a9e2a7f506a1afe2952efd/kifi-backend/modules/common/app/com/keepit/notify/model/event/KeepEvent.scala#L41-L65) would help? I'm certainly open to any suggestions, because I don't really like this that much either.

kifi-backend/modules/common/conf/evolutions/shoebox/378.sql
```diff
@@ -0,0 +1,25 @@
+# SHOEBOX
+
+# --- !Ups
+
+alter table notification add COLUMN group_identifier VARCHAR(256);
+
+alter table notification add CONSTRAINT notification_group_identifier unique index(group_identifier);
+
+alter table notification add COLUMN recipient VARCHAR(256) not null DEFAULT('');
```
ryanpbrewster, line 9: Is `''` a valid default recipient? Why even set a default?

kifi-backend/modules/common/app/com/keepit/notify/info/NeedsInfo.scala
maxov, line 16: Fixed in 9d7e659

kifi-backend/modules/common/conf/evolutions/shoebox/378.sql
maxov, line 9: Fixed in 22337ba4f7cbc456

kifi-backend/modules/common/app/com/keepit/notify/info/NotificationInfo.scala
maxov, line 27: Yes, that's exactly it. I added a comment to NotificationInfo in commit 8e6f99b

kifi-backend/modules/eliza/app/com/keepit/eliza/model/NotificationRepo.scala
```diff
@@ -26,11 +28,14 @@ class NotificationRepoImpl @Inject() (
   type RepoImpl = NotificationTable
   class NotificationTable(tag: Tag) extends RepoTable[Notification](db, tag, "notification") {
 
-    def userId = column[Id[User]]("user_id", O.NotNull)
+    def recipient = column[Recipient]("recipient", O.NotNull)
     def lastChecked = column[DateTime]("last_checked", O.NotNull)
     def kind = column[String]("kind", O.NotNull)
+    def groupIdentifier = column[Option[String]]("group_identifier", O.Nullable)
```
ryanpbrewster, line 34: Shouldn't every Notification have a group identifier?

kifi-backend/modules/eliza/app/com/keepit/eliza/model/NotificationRepo.scala
```diff
@@ -26,11 +28,14 @@ class NotificationRepoImpl @Inject() (
...
+    def lastEvent = column[DateTime]("last_event", O.NotNull)
```
ryanpbrewster, line 35: Does this need to be denormalized onto Notification?

kifi-backend/modules/eliza/app/com/keepit/eliza/model/Notification.scala
```diff
@@ -28,9 +28,12 @@ case class Notification(
...
     createdAt: DateTime = currentDateTime,
     updatedAt: DateTime = currentDateTime,
-    userId: Id[User],
+    recipient: Recipient,
     lastChecked: DateTime = START_OF_TIME,
-    kind: NKind) extends Model[Notification] {
+    kind: NKind,
+    groupIdentifier: Option[String] = None,
+    lastEvent: DateTime = currentDateTime,
+    disabled: Boolean = false) extends Model[Notification] {
```
ryanpbrewster, line 36: What does disabled mean?

kifi-backend/modules/eliza/app/com/keepit/eliza/model/NotificationRepo.scala
maxov, line 34: No, only those kinds that need it. Some might be filtered by time, where having a group identifier wouldn't make sense. All notifications of a specific kind either have group identifier or don't have a group identifier.

kifi-backend/modules/eliza/app/com/keepit/eliza/model/Notification.scala
maxov, line 36: It means that the user chose to mark the notification as muted, meaning that it will never be returned as unread to the user.

kifi-backend/modules/eliza/app/com/keepit/eliza/model/NotificationRepo.scala
maxov, line 35: Yes, so that it is very simpl to check which notifications are new. Without it we'd have to check each notification's items' last time to figure out if it is actually unread, either making a whole lot of sql queries or one very, very ugly one. Here, all that is needed is a comparison between lastChecked and lastEvent. This is at a time where performance is especially needed, as the user is scrolling down in their notifications list and expects instant or near-instant loading.

kifi-backend/modules/common/app/com/keepit/notify/info/NeedsInfo.scala
ryanpbrewster, line 72: I think I have an idea, let me write it up.

kifi-backend/modules/common/app/com/keepit/notify/info/NotificationInfo.scala
ryanpbrewster, line 27: Very nice.

kifi-backend/modules/common/app/com/keepit/notify/model/EventArgs.scala
ryanpbrewster, line 7: :+1:

kifi-backend/modules/eliza/app/com/keepit/eliza/model/NotificationRepo.scala
ryanpbrewster, line 34: Okay, makes sense.

kifi-backend/modules/eliza/app/com/keepit/eliza/model/Notification.scala
ryanpbrewster, line 36: Can we explicitly call it something like `muteStatus: MuteStatus` where

```
sealed abstract class MuteStatus(val status: String)
object MuteStatus {
    case class Muted extends MuteStatus("muted")
    case class Unmuted extends MuteStatus("unmuted")
}
```

That more closely follows the conventions elsewhere in our codebase.

maxov, line 36: Should I evolve the column to store a string (status), then? Or perhaps

``` java
sealed abstract class MuteStatus(val muted: Boolean)
object MuteStatus {
    case class Muted extends MuteStatus(true)
    case class Unmuted extends MuteStatus(false)
}
```

Or, it might just be a good idea to add the states field in and add 'muted' as an extra state?

kifi-backend/modules/common/app/com/keepit/notify/info/NeedsInfo.scala
ryanpbrewster, line 72: https://gist.github.com/ryanpbrewster/9d6e79294ac504b56d39

kifi-backend/modules/eliza/app/com/keepit/eliza/model/Notification.scala
ryanpbrewster, line 36: We have been trying to adopt the convention that `state` only encodes two possible values, `ACTIVE` and `INACTIVE`.

We also tend to use explicitly enumerated strings instead of booleans (with a few notable exceptions), hence the preference for `MuteStatus(String)`.

kifi-backend/modules/common/app/com/keepit/notify/info/NeedsInfo.scala
leogrim, line 16: There's `ImagePath` class, is that what you meant @ryanpbrewster ?

ryanpbrewster, line 16: Yeah, but I realized that `ImagePath` needs an implicit `S3ImageConfig` in order to actually produce the URL, so it actually needs to be a `String` by the time it gets to Eliza.

kifi-backend/modules/shoebox/app/com/keepit/notify/info/ShoeboxNotificationInfoSourceImpl.scala
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  override def library(id: Id[Library]): Future[Library] = {
+    val library = db.readOnlyReplica { implicit session =>
+      libraryRepo.get(id)
+    }
+    Future.successful(library)
+  }
+
+}
```
leogrim, line 51: This means you always go to Shoebox to build your notifications, while many of these entities can be retrieved via existing _cached_ endpoints. Are you thinking of changing this for Eliza to directly call these endpoints then patching on its end?

maxov, line 51: Yeah, I actually had an [eliza implementation](https://github.com/kifi/keepit/pull/16720/files#diff-edec1002b5072b0dfd0b3a653f28ee75R12) of NotificationinfoSource, but I'm replacing it with the DB subset system #16913 so the batch endpoints can be cached. The rationale for putting this on shoebox was that it could be more efficient to generate notification info on the shoebox side only when it actually creates the event, just in case Eliza decides to send the event right away, preventing Eliza from having to request the same info again. Eliza would still use the shoebox service client if it wanted to generate info by itself.

kifi-backend/modules/common/app/com/keepit/notify/info/NeedsInfo.scala
leogrim, line 16: You can just inject it.

ryanpbrewster, line 16: :+1: let's do that then, @gratimax. More typesafe and semantically suggestive.

maxov, line 16: Can't really do that because info generation has no concept of dependency injection. I guess it's possible to add a `NeedInfo` for s3 image configuration? Otherwise, perhaps a `ResolvedImagePath` case class that also represents an image path in this way.

ryanpbrewster, line 16: Just send over the `ImagePath`, and when you're actually creating the `NotificationInfo` you can fill it in using `ImagePath.getUrl`. You can do dependency injection there, yeah?

maxov, line 16: No, that's what I mean. There is no way to do dependency injection when generating info because `UsingDbSubset` functions know nothing about it. The only input source they have (and are _designed_ to have) is through the NeedInfo system.

PR #16722, opened 2015-08-10
Removed the KeepState "duplicate"

ryanpbrewster:
We rely strictly on `isPrimary` to determine if a Keep is duplicated or not

@leogrim @stkem


stkem:
Couple of unrelated things in there, but lgtm. Make sure to make the db change first.

PR #16724, opened 2015-08-10
pass auth token through userOrOrg

camhashemi:
@adamjgrant @cvializ this allows non-users to see an org if they have a valid auth token


cvializ:
:+1:

PR #16728, opened 2015-08-10
allow for the ability to push to devices marked as dev, previously we…

jaredpetker:
… were pushing to all devices with the prod certificate, time to obey whether a device is marked as dev or not


jaredpetker:
@eishay @stkem @andrewconner 

We'd like to be able to test push notifications in dev again - it's currently impossible to test ios without it in a development environment, the ios team seems to have never been testing in prod, only in dev. Android just always used the release api key for app boy because this was never implemented fully.

Let me know if anything issues are glaring.

andrewconner:
Nice! Will review.

andrewconner:
Looks reasonable. Let me know if you'd like help testing.

jaredpetker:
We'll know very quickly if it's all going well or not, ok with merging?

andrewconner:
Yep

PR #16731, opened 2015-08-11
Add newly invited members to the list of members.

cvializ:
@adamjgrant @andrewconner Trying to make the invite flow feel more slick


adamjgrant:
Code looks good. What will this do?

cvializ:
When you invite a user via the invite modal, it will add the newly invited user to the list without needing to refresh the page.

andrewconner:
Looks good.

PR #16734, opened 2015-08-11
Fixed site routing

ryanpbrewster:
Swap around SEE_OTHER and MOVED_PERMANENTLY

@andrewconner @leogrim

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/HandleOps.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/HandleOps.scala
@@ -38,7 +38,7 @@ object HandleOps {
   def isValid(handle: String): Boolean = {
     if (handleRegex.matcher(handle).matches) {
       val normalized = normalize(handle)
-      censorList.filter(w => normalized.indexOf(w) >= 0).isEmpty &&
+      !censorList.exists(w => normalized.indexOf(w) >= 0) &&
         !topDomains.contains(normalized) &&
         !commonEnglishWords.contains(normalized) &&
         !topBrands.contains(normalized)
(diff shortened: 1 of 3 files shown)


ryanpbrewster:
Talked to @stkem about this in person today. I'm pretty sure we should be using different HTTP status codes. For simple handle/slug normalization, the resource is _permanently_ at another address, so we should be using `MOVED_PERMANENTLY`. For username/handle changes, the resource _has_ moved permanently, but this change should _NOT_ be cached. Today we ran into an issue where Eishay's browser cached one of our 301s, so when I changed an org's handle his browser refused to detect it; it stuck with the cached redirect it had.

I'm not sure if this is an abuse of HTTP status codes, or intended use.

andrewconner:
The way it is is intentional, but I'm fine with changing it. The reasoning behind using the permanent redirect was that the resource you requested is permanently moved to somewhere else (ie, it was here, it's somewhere else, this change isn't temporary, it's "forever"). Answering the abuse of standard english definitions of "permanent" and "forever", it's obviously not guaranteed to be forever because your domain registration could expire, you completely change your site routing, etc.

Obviously, what matters here is if our current implementation causes problems. If so (like you explained), change away.

leogrim:
The primary intent was to tell search engines to update their index, hence the 301.

ryanpbrewster:
It's a relatively minor issue, and I imagine it won't pop up too often (but I'm sure it will occur again). I'm not familiar with how search engines handle redirects. Should I merge this?

leogrim:
https://support.google.com/webmasters/answer/93633?hl=en

ryanpbrewster:
Seems like a 301 is the "right" choice, but we really don't want users to cache the redirect. Will a 307 (Temporary Redirect) do the thing I want?

ryanpbrewster:
After talking with @andrewconner I think it's probably fine to close this without merging. If it comes up again we'll at least know what the issue was and how to fix it.

PR #16738, opened 2015-08-11
Clean up and use EmailAccount.hash in queries

leogrim:
@andrewconner I won't merge this until ABook is done catching up on hashes and the index is in place.


andrewconner:
:+1:

leogrim:
Done!

PR #16739, opened 2015-08-11
Migration endpoint for User.primaryEmail

leogrim:
@andrewconner makes sense?


kifi-backend/modules/shoebox/app/com/keepit/admin/AdminUserController.scala
```diff
@@ -1072,4 +1072,36 @@ class AdminUserController @Inject() (
...
+        lastFixed = Some {
+          db.readWrite { implicit sesssion =>
+            val users = userRepo.page(0, pageSize)
+            users.foreach { user =>
+              user.primaryEmail.foreach { primaryEmail =>
+                emailRepo.getOwner(primaryEmail) match {
+                  case Some(ownerId) if ownerId != user.id.get => log.error(s"[PrimaryEmailMigration] User.primaryEmail $primaryEmail, of user ${user.id.get} is actually owned by user $ownerId.")
+                  case _ => userEmailAddressCommander.intern(user.id.get, primaryEmail) match {
+                    case Failure(ex) => log.error(s"[PrimaryEmailMigration] Failed to intern User.primaryEmail $primaryEmail of user ${user.id.get}: $ex.")
+                    case Success((emailAddress, _)) => userEmailAddressCommander.setAsPrimaryEmail(emailAddress)
```
leogrim, line 1087: This marks the address as pending-primary if it's not verified.

andrewconner:
Yep looks good to me (sans compiling and running it :))

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminUserController.scala
```diff
@@ -1072,4 +1072,36 @@ class AdminUserController @Inject() (
...
+  def migratePrimaryEmails() = AdminUserPage { implicit request =>
+    SafeFuture {
+      val pageSize = 100
+      var page = 0
+      var total = 0
+      var lastFixed: Option[Int] = None
+      while (!lastFixed.contains(0)) {
+        lastFixed = Some {
+          db.readWrite { implicit sesssion =>
+            val users = userRepo.page(0, pageSize)
```
andrewconner: Super minor, `pageAscending` may be a better choice here.

PR #16740, opened 2015-08-11
Update member list after invites are accepted or canceled. New remove…

cvializ:
… member modal copy

@adamjgrant @andrewconner 

modal code for https://app.asana.com/0/42816489233171/44702170995874


adamjgrant:
Nice, I think we keep running into having to elegantly reload this data. This should be a nice fix for that.

PR #16742, opened 2015-08-11
handle users viewing orgs without the experiment

camhashemi:
@ryanpbrewster

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/shoebox/controllers/OrganizationAccessActions.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/shoebox/controllers/OrganizationAccessActions.scala
@@ -61,13 +61,14 @@ trait OrganizationAccessActions {
                 Future.successful(OrganizationFail.INSUFFICIENT_PERMISSIONS.asErrorResponse)
               }
             case request: NonUserRequest[_] =>
-              log.info(s"[orgAuth] authToken = $authTokenOpt")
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/app/com/keepit/shoebox/controllers/OrganizationAccessActions.scala
```diff
@@ -61,13 +61,14 @@ trait OrganizationAccessActions {
...
               }
             case request: NonUserRequest[_] =>
-              log.info(s"[orgAuth] authToken = $authTokenOpt")
               if (authTokenOpt.exists(authToken => orgInviteCommander.isAuthValid(orgId, authToken))) {
                 val memberPermissions = orgMembershipCommander.getPermissions(orgId, None)
                 block(OrganizationRequest(request, orgId, authTokenOpt, memberPermissions))
               } else {
                 Future.successful(OrganizationFail.INVALID_AUTHTOKEN.asErrorResponse)
               }
+            case _ => Future.successful(NotFound) // we can remove this after we remove the OrgExperiment guard
```
ryanpbrewster, line 70: I don't understand how this is the first time we've seen that match error. The fact that this was missing really concerns me. Nice catch.

camhashemi, line 70: I'm guessing it's the first time because I changed the logic yesterday to handle `NonUserRequest`, and forgot to add back in the wildcard.

PR #16743, opened 2015-08-11
Show message for an org without libraries

cvializ:
@andrewconner @adamjgrant


adamjgrant:
Nice.

cvializ:
@adamjgrant do you think this is expected behavior for someone who can add/move libraries to the org? or should it hide that message since the add cards are there? I'm leaning towards showing it.
screen shot 2015-08-11 at 11 33 55 am

adamjgrant:
I agree. Without an explicit message, users can be lead to believe the data hasn't downloaded yet.

PR #16744, opened 2015-08-11
Show message for org without visible members.

cvializ:
@adamjgrant @andrewconner similar to previous PR. I'm unsure whether orgs ever hide all members from non-members, but I thought it'd be good to practice defensive programming.


andrewconner:
:+1:

PR #16747, opened 2015-08-11
append protocol to organization site

camhashemi:
@ryanpbrewster @andrewconner this is a quick fix, a more robust approach is to change `site : String` to `site : OrganizationSite` with a `Reads[OrganizationSite]` that covers this behavior. or maybe there's middle ground. what do you think?


ryanpbrewster:
We already have a `URL` type. Can't we just use that?

ryanpbrewster:
What issue is this fixing?

EDIT: Oh, I see, people are entering `www.kifi.com` or something and it doesn't have the `http://`. Is that actually a problem?

camhashemi:
`Url` is a DB model, seems heavy for this usage. Organizations should be persisted with a url that includes a protocol.

the clients need the protocol to link somewhere, and it's probably better to have it persisted than to tag it on in the client.

PR #16749, opened 2015-08-11
Macro Annotation that airbrakes when a method is too slow

stkem:
The goal is to have easy performance monitoring and alerts without needed unreliable external tools like syren.
Usage is pretty simple. If you write

``` scala
@AlertingTimer(5 seconds)
def foo(): T { ... }
```

an Airbrake will be triggered (without affecting function call results) if the mean execution time of `foo` ever goes above 5 seconds. The alert will only fire if the function has been executed at least 10 times total so far (for sufficient sample size) and will only fire at most once every 10 minutes (so as not to flood us if things are consistently slow)
Overhead should be minimal, but I wouldn't use it for really small things, i.e. <50ms.

Similarly, if you have a function that returns a future and you want the alert on the completion time of the future, use 

``` scala
@AlertingTimerAsync(5 seconds)
def foo(): Future[T] { ... }
```

Would love some feedback on this @andrewconner @leogrim @ryanpbrewster @gratimax @eishay or whoever else wants to jump in.

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/common/performance/package.scala
+++ b/kifi-backend/modules/common/app/com/keepit/common/performance/package.scala
@@ -192,14 +192,14 @@ package object performance {
             @volatile var lastAlertAt: Option[Long] = None
             @volatile var mean: Long = 0
 
-            def updateAndGetMean(sample: Long): FiniteDuration = synchronized {
+            private def updateAndGetMean(sample: Long): FiniteDuration = synchronized {
               samples = samples + 1
               lastAlertAt = Some(System.nanoTime)
               mean = mean + (sample/samples) - (mean/samples)
               Duration.fromNanos(mean)
             }
 
-            def monitor(t: Long): Unit = {
+            private def monitor(t: Long): Unit = {
               val dur = updateAndGetMean(t)
               if (samples > 10 && dur > limit && !lastAlertAt.exists(_ > System.nanoTime - 600000000000L)) { //alert if the mean is over the limit and the last alert is more than 10 minutes ago
                 airbrake.notify($message)
--- a/kifi-backend/modules/common/app/com/keepit/common/performance/package.scala
+++ b/kifi-backend/modules/common/app/com/keepit/common/performance/package.scala
@@ -5,10 +5,13 @@ import play.api.Logger
 
 import scala.util.Random
 import scala.concurrent.{ Future, ExecutionContext }
+import scala.concurrent.duration.Duration
 
 import scala.reflect.macros._
 import scala.language.experimental.macros
 import scala.annotation.StaticAnnotation
+
+import java.util.UUID
 
 package object performance {
 
@@ -100,8 +103,8 @@ package object performance {
     res
   }
 
-  object statsdMacroInstance extends StatsdMacro("test", false)
-  object statsdAsyncMacroInstance extends StatsdMacro("test", true)
+  object statsdMacroInstance extends StatsdMacro(false)
+  object statsdAsyncMacroInstance extends StatsdMacro(true)
 
   class StatsdTiming(name: String) extends StaticAnnotation {
     def macroTransform(annottees: Any*): Any = macro statsdMacroInstance.impl
@@ -111,7 +114,7 @@ package object performance {
     def macroTransform(annottees: Any*): Any = macro statsdAsyncMacroInstance.impl
   }
 
-  class StatsdMacro(name: String, async: Boolean) {
+  class StatsdMacro(async: Boolean) {
     def impl(c: whitebox.Context)(annottees: c.Expr[Any]*): c.Expr[Any] = {
       import c.universe._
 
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/common/app/com/keepit/common/performance/package.scala
```diff
@@ -143,4 +146,93 @@ package object performance {
...
+
+      val limit = extractAnnotationParameters(c.prefix.tree)
+
+      def modifiedDeclaration(defDecl: DefDef) = {
+        val q"$mods def $defName(...$args): $retType = { ..$body }" = defDecl
+
+        val className = Option(c.enclosingClass).flatMap(c => Option(c.symbol)).map(_.toString).getOrElse("unknown")
+
+        val fullName = s"$className.${defName.toString}"
+        val name = c.universe.TermName(s"_timer_${UUID.randomUUID.toString.replace("-", "_")}")
```
andrewconner, line 177: Won't this play oddly with incremental compilation, because every time the source is parsed, a new class is created?

joshm1:
Very cool!

kifi-backend/modules/common/app/com/keepit/common/performance/package.scala
stkem, line 177: True, but it's only referenced locally, so shouldn't matter. Compiled fine locally for me with incremental. It's additional load on the compiler, thought, that is true.

kifi-backend/modules/common/app/com/keepit/common/performance/package.scala
```diff
@@ -143,4 +146,93 @@ package object performance {
...
+        val message = c.universe.Constant(s"$fullName is taking too long")
+
+        val invokation = if (async) {
+          q"$mods def $defName(...$args): $retType = $name.timeitAsync { ..$body }"
+        } else {
+          q"$mods def $defName(...$args): $retType = $name.timeit { ..$body }"
+        }
+
+        c.Expr(q"""
+          private object $name {
```
andrewconner, line 187: Could you do a single object, and keep references? If you want people to regularly use this, saves on all the generated classes.

stkem, line 187: I thought about that, the problem is that this thing uses airbrake and I don't want to have to inject it here as that will be messy. This way it simply needs to be in scope where the macro is used.
That, plus extra classes are pretty cheap. Every scala closure is class, currently, and that seems to be fine.

maxov:
:+1: I'm not too fluent with macro magic, but this looks pretty great.

ryanpbrewster:
The macro stuff looks awesome (and super magic).

Do we really want to keep track of the mean, though? A rolling mean or a `threshold` and `limit` (where we alert if more than `limit` samples are over the `threshold`) both seem more applicable here.

andrewconner:
Yeah, Ryan's comment is something to think about. An absolute mean basically means the longer a service runs, the harder it is for us to find it.

kifi-backend/modules/common/app/com/keepit/common/performance/package.scala
```diff
@@ -143,4 +146,93 @@ package object performance {
...
+            import scala.concurrent.duration._
+
+            val limit: FiniteDuration = $limit
+            @volatile var samples: Long = 0
+            @volatile var lastAlertAt: Option[Long] = None
+            @volatile var mean: Long = 0
+
+            def updateAndGetMean(sample: Long): FiniteDuration = synchronized {
```
andrewconner: This and one below can be `private`

andrewconner:
Cool!

stkem:
I mostly went with a simple mean to keep it as light as possible. Both a rolling mean and a threshold/limit/count based approach would require storage of an array of values, unless I'm missing something?
(Might still be worth it though, even with extra overhead)

kifi-backend/modules/common/app/com/keepit/common/performance/package.scala
stkem: ah, right. Fixing that.

kifi-backend/modules/common/app/com/keepit/common/performance/package.scala
```diff
@@ -143,4 +146,93 @@ package object performance {
...
+          q"$mods def $defName(...$args): $retType = $name.timeit { ..$body }"
+        }
+
+        c.Expr(q"""
+          private object $name {
+            import scala.concurrent.duration._
+
+            val limit: FiniteDuration = $limit
+            @volatile var samples: Long = 0
+            @volatile var lastAlertAt: Option[Long] = None
```
eishay, line 192: it would be faster, less garbage and well hidden to use `-1L` as a replacement to `None` here instead of `Option`.  Then you won't have to use an additional object with mapping function down there.

ryanpbrewster:
Definitely have to store an array of values.

ryanpbrewster:
Chatted in person, putting something here as a reminder.

One lightweight alternative to a moving average is to have a _resetting_ average. Every `n` method calls, set the average to 0. Alternatively, every `k` minutes (or any other period of time) reset the average.

ryanpbrewster:
Last thing, I promise. Is there a reason you're using the `n += 1; mean += x/n - mean/n` formula instead of a `total += x; mean = total / n` one? I generally prefer the latter, but that may come from my time as a scientist where numerical stability was important.

stkem:
Re mean formula: I honestly didn't think about it all that much. I'm changing it to `total += x; mean = total / n`. Simpler.

PR #16750, opened 2015-08-11
Move text-switching logic out of the template into a directive

cvializ:
also adds special text for canceling invites

@adamjgrant @andrewconner


adamjgrant:
:+1:

PR #16752, opened 2015-08-11
KeepRepo -> KeepToLibrary migration

ryanpbrewster:
KeepRepo shouldn't need to know about libraries anymore.

@leogrim @andrewconner 

I'll go through and tag a few of the things I want feedback on.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/model/KeepRepo.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/model/KeepRepo.scala
@@ -528,14 +528,14 @@ class KeepRepoImpl @Inject() (
 
     (beforeIdOpt.flatMap(getOpt), afterIdOpt.flatMap(getOpt)) match {
       case (Some(before), _) =>
-        sql"""SELECT #$bookmarkColumnOrder FROM bookmark bm WHERE library_id IN (SELECT library_id FROM library_membership WHERE user_id=$userId AND state='active') AND state='active' AND user_id!=$userId AND kept_at <= ${before.keptAt} AND id < ${before.id.get} ORDER BY kept_at DESC LIMIT $limit;""".as[Keep].list
+        sql"""SELECT #$bookmarkColumnOrder FROM bookmark bm WHERE library_id IN (SELECT library_id FROM library_membership WHERE user_id=$userId AND state='active') AND state='active' AND user_id!=$userId AND kept_at <= ${before.keptAt} AND id < ${before.id.get} ORDER BY kept_at DESC, id DESC LIMIT $limit;""".as[Keep].list
       case (None, Some(after)) =>
         // This case is not strictly correct. It's not possible to call after a keep, and get other keeps kept in the same ms.
         // Fortunately, ending in this state where you have a keep id and need ones that happened after (and they happened in the same ms)
         // is nearly impossible. We can't use IDs as tie breakers because old IDs may get updated kept_at fields.
-        sql"""SELECT #$bookmarkColumnOrder FROM bookmark bm WHERE library_id IN (SELECT library_id FROM library_membership WHERE user_id=$userId AND state='active') AND state='active' AND user_id!=$userId AND kept_at > ${after.keptAt} ORDER BY kept_at DESC LIMIT $limit;""".as[Keep].list
+        sql"""SELECT #$bookmarkColumnOrder FROM bookmark bm WHERE library_id IN (SELECT library_id FROM library_membership WHERE user_id=$userId AND state='active') AND state='active' AND user_id!=$userId AND kept_at > ${after.keptAt} ORDER BY kept_at DESC, id DESC LIMIT $limit;""".as[Keep].list
       case (None, None) =>
-        sql"""SELECT #$bookmarkColumnOrder FROM bookmark bm WHERE library_id IN (SELECT library_id FROM library_membership WHERE user_id=$userId AND state='active') AND state='active' AND user_id!=$userId ORDER BY kept_at DESC LIMIT $limit;""".as[Keep].list
+        sql"""SELECT #$bookmarkColumnOrder FROM bookmark bm WHERE library_id IN (SELECT library_id FROM library_membership WHERE user_id=$userId AND state='active') AND state='active' AND user_id!=$userId ORDER BY kept_at DESC, id DESC LIMIT $limit;""".as[Keep].list
     }
   }
 
--- a/kifi-backend/modules/shoebox/app/com/keepit/model/KeepRepo.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/model/KeepRepo.scala
@@ -56,18 +56,18 @@ trait KeepRepo extends Repo[Keep] with ExternalIdColumnFunction[Keep] with SeqNu
   def getCountsByLibrary(libraryIds: Set[Id[Library]])(implicit session: RSession): Map[Id[Library], Int]
   def getByUserIdAndLibraryId(userId: Id[User], libraryId: Id[Library], excludeSet: Set[State[Keep]] = Set(KeepStates.INACTIVE))(implicit session: RSession): Seq[Keep]
   def getByLibraryIds(libraryIds: Set[Id[Library]])(implicit session: RSession): Seq[Keep]
-  // These ones do not yet
+  def recentKeepNotes(libId: Id[Library], limit: Int)(implicit session: RSession): Seq[String]
+  def getByLibraryIdsAndUriIds(libraryIds: Set[Id[Library]], uriIds: Set[Id[NormalizedURI]])(implicit session: RSession): Seq[Keep]
   def getPrimaryByUriAndLibrary(uriId: Id[NormalizedURI], libId: Id[Library])(implicit session: RSession): Option[Keep]
   def getByLibraryIdAndExcludingVisibility(libId: Id[Library], excludeVisibility: Option[LibraryVisibility], limit: Int)(implicit session: RSession): Seq[Keep]
-  def getRecentKeepsFromFollowedLibraries(userId: Id[User], limit: Int, beforeIdOpt: Option[ExternalId[Keep]], afterIdOpt: Option[ExternalId[Keep]])(implicit session: RSession): Seq[Keep]
-  def getMaxKeepSeqNumForLibraries(libIds: Set[Id[Library]])(implicit session: RSession): Map[Id[Library], SequenceNumber[Keep]]
   def getByLibraryWithInconsistentOrgId(libraryId: Id[Library], expectedOrgId: Option[Id[Organization]], limit: Limit)(implicit session: RSession): Set[Id[Keep]]
   def getKeepsFromLibrarySince(since: DateTime, library: Id[Library], max: Int)(implicit session: RSession): Seq[Keep]
+  // These ones do not yet
+  def getRecentKeepsFromFollowedLibraries(userId: Id[User], limit: Int, beforeIdOpt: Option[ExternalId[Keep]], afterIdOpt: Option[ExternalId[Keep]])(implicit session: RSession): Seq[Keep]
+  // I think these ones need to be changed or rethought:
   def librariesWithMostKeepsSince(count: Int, since: DateTime)(implicit session: RSession): Seq[(Id[Library], Int)]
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/shoebox/app/com/keepit/model/KeepRepo.scala
```diff
@@ -499,7 +497,7 @@ class KeepRepoImpl @Inject() (
 
   def librariesWithMostKeepsSince(count: Int, since: DateTime)(implicit session: RSession): Seq[(Id[Library], Int)] = {
     import com.keepit.common.db.slick.StaticQueryFixed.interpolation
-    sql"""select b.library_id, count(*) as cnt from bookmark b, library l where l.id = b.library_id and l.state='active' and l.visibility='published' and b.kept_at > $since group by b.library_id order by count(*) desc limit $count""".as[(Id[Library], Int)].list
+    sql"""select b.library_id, count(*) as cnt from bookmark b, library l where l.id = b.library_id and l.state='active' and l.visibility='published' and b.kept_at > $since group by b.library_id order by count(*) desc, b.library_id asc limit $count""".as[(Id[Library], Int)].list
```
ryanpbrewster, line 500: This method was non-deterministic before. it was undefined how to break ties if two libraries had equal numbers of new keeps. I added a tie-breaker strictly so that my backwards compatibility tests would pass.

kifi-backend/modules/shoebox/app/com/keepit/model/KeepRepo.scala
```diff
@@ -528,14 +526,14 @@ class KeepRepoImpl @Inject() (
...
-        sql"""SELECT #$bookmarkColumnOrder FROM bookmark bm WHERE library_id IN (SELECT library_id FROM library_membership WHERE user_id=$userId AND state='active') AND state='active' AND user_id!=$userId AND kept_at > ${after.keptAt} ORDER BY kept_at DESC LIMIT $limit;""".as[Keep].list
+        sql"""SELECT #$bookmarkColumnOrder FROM bookmark bm WHERE library_id IN (SELECT library_id FROM library_membership WHERE user_id=$userId AND state='active') AND state='active' AND user_id!=$userId AND kept_at > ${after.keptAt} ORDER BY kept_at DESC, id DESC LIMIT $limit;""".as[Keep].list
       case (None, None) =>
-        sql"""SELECT #$bookmarkColumnOrder FROM bookmark bm WHERE library_id IN (SELECT library_id FROM library_membership WHERE user_id=$userId AND state='active') AND state='active' AND user_id!=$userId ORDER BY kept_at DESC LIMIT $limit;""".as[Keep].list
+        sql"""SELECT #$bookmarkColumnOrder FROM bookmark bm WHERE library_id IN (SELECT library_id FROM library_membership WHERE user_id=$userId AND state='active') AND state='active' AND user_id!=$userId ORDER BY kept_at DESC, id DESC LIMIT $limit;""".as[Keep].list
```
ryanpbrewster, line 536: @andrewconner I added a tie-breaker here.

I'll tag you in the version of this that I implemented in `KeepToLibraryRepo`. As we talked about before lunch, I think it's possible to break ties in a consistent manner.

kifi-backend/modules/shoebox/app/com/keepit/model/KeepToLibraryRepo.scala
```diff
@@ -126,4 +145,86 @@ class KeepToLibraryRepoImpl @Inject() (
...
+  /**
+   * ***************************************************
+   * For backwards compatibility with KeepRepo methods *
+   * ***************************************************
+   */
+
+  def getByLibraryIdAndExcludingVisibility(libId: Id[Library], excludeVisibility: Option[LibraryVisibility], limit: Limit)(implicit session: RSession): Seq[KeepToLibrary] = {
+    val q = { for (ktl <- rows if ktl.libraryId === libId && ktl.visibility =!= excludeVisibility.orNull) yield ktl }.take(limit.value)
+    q.list
+  }
```
ryanpbrewster, line 158: This method is used strictly for checking that Keeps are consistent with their Library.

I would _love_ to simply have a DB constraint that ensures that `KTL`s are consistent with the library they're attached to. The reason we don't do this now is (presumably) because it takes time to propagate a library change over to all the keeps it contains, so we can't have that constraint. Is there some clever DB thing we can do to get around this?

kifi-backend/modules/shoebox/app/com/keepit/model/KeepToLibraryRepo.scala
```diff
@@ -126,4 +145,86 @@ class KeepToLibraryRepoImpl @Inject() (
...
+    val recentLibraryKeeps = (beforeIdOpt.map(get), afterIdOpt.map(get)) match {
+      case (None, None) =>
+        allLibraryKeeps
+      case (Some(upper), _) =>
+        allLibraryKeeps.filter(ktl => ktl.addedAt < upper.addedAt || (ktl.addedAt === upper.addedAt && ktl.keepId < upper.keepId))
+      case (None, Some(lower)) =>
+        allLibraryKeeps.filter(ktl => ktl.addedAt > lower.addedAt || (ktl.addedAt === lower.addedAt && ktl.keepId > lower.keepId))
+    }
+    recentLibraryKeeps.sortBy(ktl => (ktl.addedAt desc, ktl.keepId desc)).take(limit.value).list
+  }
```
ryanpbrewster, line 194: @andrewconner this is the version I wrote. You have to provide it with a set of libraries, so we'd need to grab the user's followed libraries ahead of time. I think that should be really fast, though, especially if we can cache it.

kifi-backend/modules/shoebox/app/com/keepit/model/KeepToLibraryRepo.scala
```diff
@@ -126,4 +145,86 @@ class KeepToLibraryRepoImpl @Inject() (
...
+    } else {
+      val idset = libIds.map { _.id }.mkString("(", ",", ")")
+      val q = sql"""select ktl.library_id, max(bm.seq) 
+                    from keep_to_library ktl, bookmark bm
+                    where bm.id = ktl.keep_id and ktl.library_id in #${idset}
+                    group by ktl.library_id"""
+
+      q.as[(Long, Long)].list.map { case (libId, seq) => Id[Library](libId) -> SequenceNumber[Keep](seq) }.toMap
+    }
+  }
```
ryanpbrewster, line 220: @leogrim, I'm not very familiar with how sequence numbers are used. Is this something we want to maintain going forward?

kifi-backend/modules/shoebox/app/com/keepit/model/KeepRepo.scala
```diff
@@ -528,14 +526,14 @@ class KeepRepoImpl @Inject() (
 
     (beforeIdOpt.flatMap(getOpt), afterIdOpt.flatMap(getOpt)) match {
       case (Some(before), _) =>
-        sql"""SELECT #$bookmarkColumnOrder FROM bookmark bm WHERE library_id IN (SELECT library_id FROM library_membership WHERE user_id=$userId AND state='active') AND state='active' AND user_id!=$userId AND kept_at <= ${before.keptAt} AND id < ${before.id.get} ORDER BY kept_at DESC LIMIT $limit;""".as[Keep].list
+        sql"""SELECT #$bookmarkColumnOrder FROM bookmark bm WHERE library_id IN (SELECT library_id FROM library_membership WHERE user_id=$userId AND state='active') AND state='active' AND user_id!=$userId AND kept_at <= ${before.keptAt} AND id < ${before.id.get} ORDER BY kept_at DESC, id DESC LIMIT $limit;""".as[Keep].list
```
andrewconner, line 529: Oo, I think the original was broken, too. `before.id` may be higher or lower than the next keep.

kifi-backend/modules/shoebox/app/com/keepit/model/KeepRepo.scala
andrewconner, line 536: Yeah this one is not a problem, since it's the first set (no before, no after).

kifi-backend/modules/shoebox/app/com/keepit/model/KeepRepo.scala
```diff
@@ -528,14 +526,14 @@ class KeepRepoImpl @Inject() (
...
       case (None, Some(after)) =>
         // This case is not strictly correct. It's not possible to call after a keep, and get other keeps kept in the same ms.
         // Fortunately, ending in this state where you have a keep id and need ones that happened after (and they happened in the same ms)
         // is nearly impossible. We can't use IDs as tie breakers because old IDs may get updated kept_at fields.
-        sql"""SELECT #$bookmarkColumnOrder FROM bookmark bm WHERE library_id IN (SELECT library_id FROM library_membership WHERE user_id=$userId AND state='active') AND state='active' AND user_id!=$userId AND kept_at > ${after.keptAt} ORDER BY kept_at DESC LIMIT $limit;""".as[Keep].list
+        sql"""SELECT #$bookmarkColumnOrder FROM bookmark bm WHERE library_id IN (SELECT library_id FROM library_membership WHERE user_id=$userId AND state='active') AND state='active' AND user_id!=$userId AND kept_at > ${after.keptAt} ORDER BY kept_at DESC, id DESC LIMIT $limit;""".as[Keep].list
```
andrewconner, line 534: I see this one doesn't have the `id` problem (that the above has), and is a weird edge case anyways, so I think this one's fine.

kifi-backend/modules/shoebox/app/com/keepit/model/KeepToLibraryRepo.scala
leogrim, line 220: I'll take another look tomorrow morning (it's almost 2am) but this looks weird, when do we need this?
To drive an integrity plugin maybe?

kifi-backend/modules/shoebox/app/com/keepit/model/KeepToLibraryRepo.scala
leogrim, line 194: Yep, that makes sense, we may also want to grab the user's connections to get keeps from their `Main` discoverable libraries.

andrewconner, line 194: I think it's right. A test that has a lot of these edge-case keeps (with out-of-order `id`s, equal `keep_at`s, etc) and just verifies that they all come out is the gold here.

ryanpbrewster, line 194: Agreed, I'll make an abusive test.

kifi-backend/modules/shoebox/app/com/keepit/model/KeepToLibraryRepo.scala
ryanpbrewster, line 220: Yeah, take a look at the corresponding `KeepRepo` method. That's used in `SuggestedSearchTermScheduler.librariesNeedUpdate`.

kifi-backend/modules/shoebox/app/com/keepit/model/KeepRepo.scala
ryanpbrewster, line 529: Oy, we should fix that. I'm not sure where `latestKeep` is used but it probably isn't doing the right thing. ((Although _most_ Keeps are probably in order.))

kifi-backend/modules/shoebox/app/com/keepit/model/KeepToLibraryRepo.scala
leogrim, line 220: I see, seems reasonable.

kifi-backend/modules/shoebox/test/com/keepit/model/KeepToLibraryRepoTest.scala
```diff
@@ -56,5 +75,223 @@ class KeepToLibraryRepoTest extends Specification with ShoeboxTestInjector {
...
+            val libIds = libs.map(_.id.get).toSet
+            val expected = inject[KeepRepo].getMaxKeepSeqNumForLibraries(libIds)
+            val actual = inject[KeepToLibraryRepo].getMaxKeepSeqNumForLibraries(libIds)
+
+            expected === actual
+          }
+          1 === 1
+        }
+      }
+    }
```
leogrim, line 295: So many tests, it's Christmas already =)

PR #16755, opened 2015-08-12
Dumped some old code that we don't use anymore

ryanpbrewster:
My excuse is that some of it used `Keep.libraryId` which is going to need
to go the way of the dodo anyway.

@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/admin/AdminBookmarkController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/admin/AdminBookmarkController.scala
@@ -107,56 +107,10 @@ class AdminBookmarksController @Inject() (
     }
   }
 
-  //post request with a list of private/public and active/inactive
-  def updateBookmarks() = AdminUserPage { request =>
-    def toBoolean(str: String) = str.trim.toInt == 1
-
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/shoebox/app/com/keepit/admin/AdminBookmarkController.scala
```diff
@@ -153,14 +115,6 @@ class AdminBookmarksController @Inject() (
...
   }
 
-  //this is an admin only task!!!
-  def delete(id: Id[Keep]) = AdminUserPage { request =>
-    db.readWrite { implicit s =>
-      keepRepo.delete(id)
-      Redirect(com.keepit.controllers.admin.routes.AdminBookmarksController.bookmarksView(0))
-    }
-  }
-
```
ryanpbrewster, line 163: I also got rid of this. My impression was that we didn't actually delete things, and this is (hopefully!) functionally equivalent to `deactivate`.

andrewconner, line 163: It's fine to remove; it's original need goes back to a different time.

PR #16762, opened 2015-08-12
Make the invite button look like a yet-to-be-invited user in the list

cvializ:
| Normal State | Hover State |
| --- | --- |
| screen shot 2015-08-12 at 11 25 51 am | screen shot 2015-08-12 at 11 25 41 am |

Please let me know if you feel there are any design issues, or if it looks sloppy, anything off-center, etc. so I can correct it : )

@adamjgrant @andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/orgProfile/orgProfileMemberManage.styl
+++ b/kifi-backend/modules/shoebox/angular/src/orgProfile/orgProfileMemberManage.styl
@@ -12,23 +12,18 @@
     margin-top: 10px
 
 .kf-org-member
+
+.kf-org-member
+  clear: both
+
   &:nth-child(odd)
     background: whitesmoke
     padding: 5px 15px
     margin-left: -15px
     margin-right: -15px
 
-.kf-org-member-invite
-  box-sizing: border-box
-  width: 100%
-  padding: 10px
-  border: 2px dashed lightgray
-  border-radius: 6px
-  background: none
-  margin-bottom: 15px
-  color: textBlackColor
-
 .kf-org-member-email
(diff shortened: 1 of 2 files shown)


adamjgrant:
I like this approach. I think @eishay would be interested in seeing this.

eishay:
I don't like it.
Asana should be the source of truth, did you look at all at 
https://app.asana.com/0/42816489233171/45350105568922
and the link from that at
https://docs.google.com/presentation/d/1tgkNTb9hVQuBs3R8fA0AQJ5TW5gkQT2BmaZMKaqtmKY/edit#slide=id.g643d77856_0_0
?

adamjgrant:
Thanks @eishay. I didn't actually see this task myself. I rarely look at the later tasks (although I should). I'll sync with @cvializ and come up with a solution. This might be a good candidate for a new approach we discussed today, for standardizing components.

cvializ:
@eishay, understood. We hadn't moved to Asana when I began implementing this page, so the initial design was from [a trello card](https://trello.com/c/z2sStv8d/20-org-member-page) containing [this wireframe](https://trello-attachments.s3.amazonaws.com/55885fab43e3e64b50bd3c86/960x1280/5b9750a464ded489d6c7da53caea6ff6/IMG_2301.JPG.jpeg).

I continued to use the dashed border motif because the "Create a Library" card used it, and I wanted to have that continuity where adding items was indicated by a dashed border.

Now I see that we want to more directly emulate our "inspirational" companies' products and I will model my implementation from those. I'm looking forward to developing Kifi standardized styles, @adamjgrant!

PR #16763, opened 2015-08-12
Made convenience classes for dealing with billing

ryanpbrewster:
@stkem I wrote this on the train in this morning.

Currency represents currency. Under the hood it stores things in `cents: Int`.

BillingCycle represents a time period. Under the hood it is using `months: Int`.

BillingRate represents a currency / time. Under the hood it is storing
just the `cost: Currency`, and it has a fixed time period that it uses.
You can construct a BillingRate by dividing a Currency by a BillingCycle,
and it will throw an exception if they are not evenly divisible. I did
that because it is the only way to ensure that
1. We are billing the same amount of money every month
2. We are billing the correct amount of money over the entire billing cycle

If we are willing to relax either of those constraints then we can do other
things (maybe charge a different amount on the last month of a billing cycle).

Now that I'm writing this out it seems more reasonable to only charge once per
billing cycle. I think that might even be what billing cycle actually means...

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/common/util/Currency.scala
+++ b/kifi-backend/modules/common/app/com/keepit/common/util/Currency.scala
@@ -5,11 +5,11 @@ case class IllegalBillingCycleDivision(a: BillingCycle, b: BillingCycle) extends
 
 case class Currency(cents: Int) {
   def +(that: Currency): Currency = Currency(this.cents + that.cents)
-  def -(that: Currency): Currency = Currency(this.cents + that.cents)
+  def -(that: Currency): Currency = Currency(this.cents - that.cents)
   def *(lambda: Int): Currency = Currency(this.cents * lambda)
   def /(lambda: Int): Currency = Currency(this.cents / lambda)
   def /(bc: BillingCycle): BillingRate = BillingRate(this, bc)
-  override def toString = s"$$${cents / 100}.${cents % 100}"
+  override def toString = s"$$${cents / Currency.CENTS_PER_DOLLAR}.${cents % Currency.CENTS_PER_DOLLAR}"
 }
 
 object Currency {
--- a/kifi-backend/modules/common/app/com/keepit/common/util/Currency.scala
+++ b/kifi-backend/modules/common/app/com/keepit/common/util/Currency.scala
@@ -1,0 +1,57 @@
+package com.keepit.common.util
+
+case class UnevenBillingRateException(cost: Currency, cycle: BillingCycle) extends Exception(s"$cost cannot be evenly divided into $cycle")
+case class IllegalBillingCycleDivision(a: BillingCycle, b: BillingCycle) extends Exception(s"$a does not divide evenly by $b")
+
+case class Currency(cents: Int) {
+  def +(that: Currency): Currency = Currency(this.cents + that.cents)
+  def -(that: Currency): Currency = Currency(this.cents + that.cents)
+  def *(lambda: Int): Currency = Currency(this.cents * lambda)
+  def /(lambda: Int): Currency = Currency(this.cents / lambda)
+  def /(bc: BillingCycle): BillingRate = BillingRate(this, bc)
+  override def toString = s"$$${cents / 100}.${cents % 100}"
+}
+
+object Currency {
+  private val CENTS_PER_DOLLAR = 100
+  def cents(n: Int) = Currency(n)
+  def dollars(n: Int) = Currency(CENTS_PER_DOLLAR * n)
+}
+
+case class BillingCycle(months: Int) {
+  def +(that: BillingCycle): BillingCycle = BillingCycle(this.months + that.months)
+  def -(that: BillingCycle): BillingCycle = BillingCycle(this.months - that.months)
+  def *(lambda: Int): BillingCycle = BillingCycle(this.months * lambda)
+  def /(lambda: Int): BillingCycle = BillingCycle(this.months / lambda)
+  def /(that: BillingCycle): Int = {
+    if (this.months % that.months != 0) throw new IllegalBillingCycleDivision(this, that)
+    else this.months / that.months
+  }
+
+  override def toString = {
+    if (months > 1) s"$months months"
+    else s"$months month"
+  }
+}
+object BillingCycle {
+  val MIN_LENGTH = BillingCycle.months(1)
+
+  private val MONTHS_PER_YEAR = 12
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/common/app/com/keepit/common/util/Currency.scala
```diff
@@ -0,0 +1,57 @@
...
+  def dollars(n: Int) = Currency(CENTS_PER_DOLLAR * n)
+}
+
+case class BillingCycle(months: Int) {
+  def +(that: BillingCycle): BillingCycle = BillingCycle(this.months + that.months)
+  def -(that: BillingCycle): BillingCycle = BillingCycle(this.months - that.months)
+  def *(lambda: Int): BillingCycle = BillingCycle(this.months * lambda)
+  def /(lambda: Int): BillingCycle = BillingCycle(this.months / lambda)
+  def /(that: BillingCycle): Int = {
+    if (this.months % that.months != 0) throw new IllegalBillingCycleDivision(this, that)
```
andrewconner, line 27: I'd be happy always rounding down or something.

kifi-backend/modules/common/app/com/keepit/common/util/Currency.scala
```diff
@@ -0,0 +1,57 @@
+package com.keepit.common.util
+
+case class UnevenBillingRateException(cost: Currency, cycle: BillingCycle) extends Exception(s"$cost cannot be evenly divided into $cycle")
+case class IllegalBillingCycleDivision(a: BillingCycle, b: BillingCycle) extends Exception(s"$a does not divide evenly by $b")
+
+case class Currency(cents: Int) {
+  def +(that: Currency): Currency = Currency(this.cents + that.cents)
+  def -(that: Currency): Currency = Currency(this.cents + that.cents)
```
andrewconner: `Currency(this.cents - that.cents)`

kifi-backend/modules/common/app/com/keepit/common/util/Currency.scala
```diff
@@ -0,0 +1,57 @@
...
+  def *(lambda: Int): Currency = Currency(this.cents * lambda)
+  def /(lambda: Int): Currency = Currency(this.cents / lambda)
+  def /(bc: BillingCycle): BillingRate = BillingRate(this, bc)
+  override def toString = s"$$${cents / 100}.${cents % 100}"
+}
+
+object Currency {
+  private val CENTS_PER_DOLLAR = 100
```
andrewconner, line 16: Not sure this is useful since 100 is hard coded a couple lines above anyways.

andrewconner, line 16: Not like this will ever change :)

kifi-backend/modules/common/app/com/keepit/common/util/Currency.scala
ryanpbrewster, line 27: I'd rather have the "always round down" logic explicitly in the business logic than do it implicitly within the `Currency` class, but I don't feel too strongly about it.

kifi-backend/modules/common/app/com/keepit/common/util/Currency.scala
```diff
@@ -0,0 +1,57 @@
+package com.keepit.common.util
+
+case class UnevenBillingRateException(cost: Currency, cycle: BillingCycle) extends Exception(s"$cost cannot be evenly divided into $cycle")
+case class IllegalBillingCycleDivision(a: BillingCycle, b: BillingCycle) extends Exception(s"$a does not divide evenly by $b")
+
+case class Currency(cents: Int) {
```
maxov, line 6: `extends AnyVal`?

maxov, line 6: Also may want an `Integral[Currency]` instance

kifi-backend/modules/common/app/com/keepit/common/util/Currency.scala
stkem, line 27: What does it even mean to divide a `BillingCycle` by another `BillingCycle`? You are effectively dividing a duration by another duration, ending up with a unitless quantity the semantics of which I'm not really clear on. Where would that come up?
Actually, I'm not sure `BillingCycle` needs arithmetic operations at all.

kifi-backend/modules/common/app/com/keepit/common/util/Currency.scala
```diff
@@ -0,0 +1,57 @@
...
+  def +(that: Currency): Currency = Currency(this.cents + that.cents)
+  def -(that: Currency): Currency = Currency(this.cents - that.cents)
+  def *(lambda: Int): Currency = Currency(this.cents * lambda)
+  def /(lambda: Int): Currency = Currency(this.cents / lambda)
+  def /(bc: BillingCycle): BillingRate = BillingRate(this, bc)
+  override def toString = s"$$${cents / Currency.CENTS_PER_DOLLAR}.${cents % Currency.CENTS_PER_DOLLAR}"
+}
+
+object Currency {
+  private val CENTS_PER_DOLLAR = 100
```
stkem, line 16: Overabstraction :-)

kifi-backend/modules/common/app/com/keepit/common/util/Currency.scala
stkem, line 6: +1 on `AnyVal`. Also, more importantly, there is a distinction on how we store amounts of money (which is simply in cents) and how we do math with money. The latter needs to adhere to standard accounting rules in regards to rounding, which regular integer arithmetic does not. Java's `BigDecimal` supports the correct behavior and that is what is commonly used for accounting/financial math. (See the discussion on the spec google doc).
Three options here:
1) `Currency` wraps `BigDecimal` instead of `Int`, with the appropriate arithmetic and we have a type mapper turning that into an `Int` for storage.
2) `Currency` implements proper rounding manually (I don't think that is a great idea)
3) `Currency` is just a dumb wrapper with no methods and all the math is done directly on `BigDecimal` in the business logic and converted upon storage. This is what I originally had in mind, but I'm not particularly attached to it (1 is a good option too)

stkem:
Nice!

andrewconner:
@stkem Do you want to use this? Otherwise let's close it.

stkem:
@andrewconner Not going to use it verbatim, since I have some of this stuff already in the other PR. I might copy/paste some things.

PR #16769, opened 2015-08-12
Add public icon to public org libraries on the extension

cvializ:
screen shot 2015-08-12 at 10 39 55 am

@adamjgrant @andrewconner


andrewconner:
:+1: I see that the library icon looks square up top vs rectangular below. Intentional?

cvializ:
Those are Org avatars. Orgs are rounded rectangles in our design language, so they're rounded rectangles in the extension too.

It could be confusing to have both library cards and org avatars in the same place. If you think that it's a risk, let's discuss a good way to organize these different elements for orgs in the extension with @adamjgrant

andrewconner:
Ohh, I see. I don't think I would have ever guessed that.

cvializ:
Hmm is it bad that you wouldn't have guessed that? Do you feel like it's inconsistent?

adamjgrant:
Yeah, that's the discussion we had around those conventions. To me it's no so much important that anyone would guess the convention is there, but that it shows the org icon in the right dimensions without disturbing the layout of the rest of the extension.

adamjgrant:
Also keep in mind we use silly pictures for our orgs. Real orgs (by and large) will use logos that are much more recognizable as an entity than a philosoraptor.

I think all in all, these icons are very picturesquee.

andrewconner:
Sure, but it's odd that org libraries are shown differently than non-org libraries, since functionally they're the same across the system. Perhaps use the existing library icon (stacks of cards) with an overlay in the corner of the org? Too busy?

adamjgrant:
Yeah, I think we cycled through all the possibilities, honestly. This one was the least offensive. 

When you get into layering libraries and org logos, there's no guarantee of color harmony. If we're relying on those colors to send strong signals, it creates a lot of noise.

adamjgrant:
Which is not to say I'm not open to other possibilities, but this is the best we have so far. Definitely room for improvement...short of redesigning the whole thing. Really, I think the org and personal libraries should just sit in separate areas. Tabs, sections, whatever. But that's a heavier lift.

cvializ:
I'll merge in this "light lift" then

adamjgrant:
:+1:

PR #16772, opened 2015-08-12
Remove all RichSocialConnection (old WTI) related code

leogrim:
@andrewconner @stkem


stkem:
lgtm. Add an evolution to drop the table?

andrewconner:
Wahoo! @leogrim

PR #16773, opened 2015-08-12
Don't let the org Owner abdicate their office.

cvializ:
@adamjgrant @andrewconner the backend already sends an error response if you try, but this will prevent the button from appearing in the first place


adamjgrant:
So this will hide the "demote" button?

cvializ:
Correct, for the owner viewing their own member entry.

adamjgrant:
:+1:

PR #16778, opened 2015-08-13
Starting to add ktl dry-run logic

ryanpbrewster:
The goal is to log anytime we can't replicate current behavior using KTL logic

@leogrim @stkem

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/admin/AdminBookmarkController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/admin/AdminBookmarkController.scala
@@ -146,7 +146,7 @@ class AdminBookmarksController @Inject() (
       }
     }
 
-    val bookmarkTotalCountFuture = keepCommander.getKeepsCountFuture
+    val bookmarkTotalCountFuture = keepCommander.getKeepsCountFuture()
 
     val bookmarkTodayAllCountsFuture = Future {
       timing("load bookmarks counts from today") {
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/shoebox/app/com/keepit/admin/AdminBookmarkController.scala
```diff
@@ -146,7 +146,7 @@ class AdminBookmarksController @Inject() (
       }
     }
 
-    val bookmarkTotalCountFuture = keepCommander.getKeepsCountFuture()
+    val bookmarkTotalCountFuture = keepCommander.getKeepsCountFuture
```
stkem: Omitting `()` can have a really strange effect when the function is defined with `()`, namely that you are simply being returned the function instead of calling it. So when doing this make sure the type system is set up such that it would catch that, because the compiler will otherwise be silent about it, even if just getting the function ends up being a no-op, which should really be a warning.
(Everything might already be totally good here, just wanted to point out the potential pitfall)

stkem:
lgtm

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminBookmarkController.scala
ryanpbrewster: Here I actually removed the parens from the declaration, so I had to remove
from the call. What are your thoughts on
http://stackoverflow.com/questions/7409502/what-is-the-difference-between-def-foo-and-def-foo-in-scala

ryanpbrewster: One of the reasons I'm hesitant is performance. Suppose we have an object `c` with a really expensive parameterless method `c.repm`. If we include parens then it's a visual cue to save the result in a `val` instead of repeatedly invoking the method.

stkem: The rule of thumb I generally try to follow is: If it has side effects or does a significant amount of computation or even I/O use parens. If it's basically just a simple value retrieval you can omit them.

andrewconner:
This is a great way to verify behavior. Awesome @ryanpbrewster

PR #16779, opened 2015-08-13
increase org invite typeahead limit to buffer filtering of org members

camhashemi:
@andrewconner this is a cheap solution, augment the limit knowing the result will be filtered. a more robust one I can think of would be to pass a filter function through to `typeaheadCommander` to apply before taking the limit.


andrewconner:
:+1:

PR #16787, opened 2015-08-13
going around a (maybe slick's bug) that this query will always return a value…

eishay:
…, may be null. Doing a firstOption may return a Some(null) value

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/model/KeepRepo.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/model/KeepRepo.scala
@@ -524,7 +524,8 @@ class KeepRepoImpl @Inject() (
 
   def getDateLastManualKeep(userId: Id[User])(implicit session: RSession): Option[DateTime] = {
     import com.keepit.common.db.slick.StaticQueryFixed.interpolation
-    sql"""select max(kept_at) from bookmark where user_id = $userId and source in ('keeper', 'mobile', 'email', 'site')""".as[DateTime].firstOption
+    //going around a (maybe slick's bug?) that this query will always return a value, may be null. Doing a firstOption may return a Some(null) value
+    Option(sql"""select max(kept_at) from bookmark where user_id = $userId and source in ('keeper', 'mobile', 'email', 'site')""".as[DateTime].first)
   }
 
   def getKeepsByTimeWindow(uriId: Id[NormalizedURI], url: String, keptAfter: DateTime, keptBefore: DateTime)(implicit session: RSession): Set[Keep] = {


kifi-backend/modules/shoebox/app/com/keepit/model/KeepRepo.scala
```diff
@@ -524,7 +524,8 @@ class KeepRepoImpl @Inject() (
 
   def getDateLastManualKeep(userId: Id[User])(implicit session: RSession): Option[DateTime] = {
     import com.keepit.common.db.slick.StaticQueryFixed.interpolation
-    sql"""select max(kept_at) from bookmark where user_id = $userId and source in ('keeper', 'mobile', 'email', 'site')""".as[DateTime].firstOption
+    //going around a (maybe slick's bug?) that this query will always return a value, may be null. Doing a firstOption may return a Some(null) value
+    Option(sql"""select max(kept_at) from bookmark where user_id = $userId and source in ('keeper', 'mobile', 'email', 'site')""".as[DateTime].first)
```
ryanpbrewster, line 528: Can't we do this directly in Slick?

`val sources: Set[KeepSource] = Set(KeepSource.keeper)`
`(for (b <- rows if b.userId === userId && b.source.inSet(sources)) yield b.keptAt).max.run`

kifi-backend/modules/shoebox/app/com/keepit/model/KeepRepo.scala
```diff
@@ -524,7 +524,8 @@ class KeepRepoImpl @Inject() (
 
   def getDateLastManualKeep(userId: Id[User])(implicit session: RSession): Option[DateTime] = {
     import com.keepit.common.db.slick.StaticQueryFixed.interpolation
-    sql"""select max(kept_at) from bookmark where user_id = $userId and source in ('keeper', 'mobile', 'email', 'site')""".as[DateTime].firstOption
+    //going around a (maybe slick's bug?) that this query will always return a value, may be null. Doing a firstOption may return a Some(null) value
```
andrewconner, line 527: Technically, this is correct behavior. You're asking for a `DateTime`, and saying "if there is a row returned, give me `Some(the first one)`; else, give me `None`". Compare these two queries:

```
select max(kept_at) from bookmark where user_id=2; -- returns a row, with NULL
select kept_at from bookmark where user_id=2; -- returns no row
```

Serialize as `Option[DateTime]` to get what you're expecting.

eishay:
I guess we could, didn't make an effort since it seems much more readable in plain sql. makes sense to change if it won't return Some(null)

ryanpbrewster:
If you wanted to be even more Scala-like, I think that
`rows.filter(k => k.userId === userId && k.source.inSet(sources)).map(_.keptAt).max.run` (or something like that) ought to work.

eishay:
fancy :-)

kifi-backend/modules/shoebox/app/com/keepit/model/KeepRepo.scala
eishay, line 527: cool, thanks for clarifying.

PR #16789, opened 2015-08-13
filter out invalid domain names

camhashemi:

weird bug where `ShoeboxServiceClient` sends a set of valid `NormalizedHostname` to `ShoeboxDataPipeline`, but when deserializing finds that one of the hostnames is invalid. This is a workaround to filter hostnames again in Shoebox. Also added logging.


andrewconner:
Logging missing?

camhashemi:
ah, one push away

camhashemi:
updated

PR #16790, opened 2015-08-13
Cut down on the size of the dry-run logs

ryanpbrewster:


kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -156,7 +156,7 @@ class LibraryCommanderImpl @Inject() (
       val oldWay = keepRepo.getByLibrary(libraryId, offset, limit)
       val newWay = ktlRepo.getByLibraryId(libraryId, Offset(offset), Limit(limit)) |> ktlCommander.getKeeps
       if (newWay != oldWay) {
-        log.info(s"[KTL-MATCH] getKeeps: $newWay != $oldWay")
+        log.info(s"[KTL-MATCH] getKeeps($libraryId): ${newWay.map(_id.get)} != ${oldWay.map(_.id.get)}")
```
camhashemi, line 159: should be .map(_.id.get

PR #16791, opened 2015-08-13
Add 'members' to the list of reserved slugs

cvializ:
@camhashemi @andrewconner @adamjgrant 

we don't want users to try to create an org library called "members" that they can't access


andrewconner:
Perfect. :+1: 

Sounds like we should do some backend validation as well.

cvializ:
Definitely, @camhashemi and I had a conversation about it yesterday. I'll leave that side of it up to him and Ryan

PR #16792, opened 2015-08-13
adding a membership now requires the org to have a handle (tests)

camhashemi:
@andrewconner @ryanpbrewster

### DIFF
--- a/kifi-backend/modules/shoebox/test/com/keepit/commanders/OrganizationInviteCommanderTest.scala
+++ b/kifi-backend/modules/shoebox/test/com/keepit/commanders/OrganizationInviteCommanderTest.scala
@@ -13,6 +13,7 @@ import com.keepit.common.social.FakeSocialGraphModule
 import com.keepit.heimdal.{ HeimdalContext, FakeHeimdalServiceClientModule }
 import com.keepit.model.UserFactoryHelper._
 import com.keepit.model.OrganizationFactoryHelper._
+import com.keepit.model.OrganizationFactory
 import com.keepit.model._
 import com.keepit.test.ShoeboxTestInjector
 import org.specs2.mutable.SpecificationLike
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/test/com/keepit/commanders/OrganizationInviteCommanderTest.scala
```diff
@@ -187,24 +187,24 @@ class OrganizationInviteCommanderTest extends TestKitSupport with SpecificationL
...
+      //            UserFactory.user().withId(inviterId).saved
+      //            UserFactory.user().withId(userId).saved
+      //            val org = OrganizationFactory.organization().withOwner(inviterId).withHandle(OrganizationHandle("kifi")).saved
+      //            memberRepo.save(org.newMembership(userId = inviterId, role = OrganizationRole.MEMBER))
+      //            val invite = inviteRepo.save(OrganizationInvite(organizationId = org.id.get, inviterId = inviterId, userId = Some(userId), role = OrganizationRole.ADMIN))
+      //            (org, invite)
+      //          }
+      //          inviteCommander.acceptInvitation(org.id.get, userId, invite.authToken) === Left(OrganizationFail.NO_VALID_INVITATIONS)
+      //        }
+      //      }
```
ryanpbrewster, line 207: Why is this all commented out?

ryanpbrewster, line 207: If it doesn't make sense, we can kill it or skip the test (`skipped("We don't sent invites with roles right now")` at the top of the test).

ryanpbrewster:
Holy cow I'm sorry you had to slog through that.

kifi-backend/modules/shoebox/test/com/keepit/commanders/OrganizationInviteCommanderTest.scala
camhashemi, line 207: I need to check the logic and handle this better, but it was failing and I figured it wasn't important enough of a test to hold up the build. Looking through it right now.

camhashemi, line 207: It was breaking because it was accepting the invitation even though it "shouldn't", maybe we've changed the logic so that it doesn't care about `role` anymore.

PR #16798, opened 2015-08-13
consolidate similar methods and taking Ryan's suggestion for a query

eishay:


eishay:
@ryanpbrewster @andrewconner

PR #16799, opened 2015-08-13
KeepRepo.getByIds should not filter inactive keeps

ryanpbrewster:
@camhashemi @cvializ @adamjgrant 

This should fix the issue. The underlying problem is one I need to look into further.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/model/KeepRepo.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/model/KeepRepo.scala
@@ -183,7 +183,7 @@ class KeepRepoImpl @Inject() (
   }
 
   def getByIds(ids: Set[Id[Keep]])(implicit session: RSession): Map[Id[Keep], Keep] = {
-    val q = for (b <- rows if b.id.inSet(ids) && b.state === KeepStates.ACTIVE) yield b
+    val q = for (b <- rows if b.id.inSet(ids)) yield b
     q.list.map { keep => (keep.id.get, keep) }.toMap
   }
 


kifi-backend/modules/shoebox/app/com/keepit/model/KeepRepo.scala
```diff
@@ -183,7 +183,7 @@ class KeepRepoImpl @Inject() (
   }
 
   def getByIds(ids: Set[Id[Keep]])(implicit session: RSession): Map[Id[Keep], Keep] = {
-    val q = for (b <- rows if b.id.inSet(ids) && b.state === KeepStates.ACTIVE) yield b
+    val q = for (b <- rows if b.id.inSet(ids)) yield b
```
camhashemi, line 186: why not use the `excludeState` pattern?

ryanpbrewster, line 186: In this case, if you pass in an id you should get the model associated with
that id, regardless of what's going on with it. KeepRepo.get also has that
behavior.

On Thu, Aug 13, 2015 at 12:50 PM, Cameron Hashemi notifications@github.com
wrote:

> In kifi-backend/modules/shoebox/app/com/keepit/model/KeepRepo.scala
> https://github.com/kifi/keepit/pull/16799#discussion_r37017464:
> 
> > @@ -183,7 +183,7 @@ class KeepRepoImpl @Inject() (
> >    }
> > 
> >    def getByIds(ids: Set[Id[Keep]])(implicit session: RSession): Map[Id[Keep], Keep] = {
> > -    val q = for (b <- rows if b.id.inSet(ids) && b.state === KeepStates.ACTIVE) yield b
> > -    val q = for (b <- rows if b.id.inSet(ids)) yield b
> 
> why not use the excludeState pattern?
> 
> —
> https://github.com/kifi/keepit/pull/16799/files#r37017464.

## 

Software Engineer

camhashemi, line 186: It just makes it explicit so that if I only want `active` models with those ids, filtering would happen in the db, without having to create a new method. It's cheap to add, you can even set the default to `None`.

PR #16801, opened 2015-08-13
Caught a few more places where Keeps are changed

ryanpbrewster:
KTL needs to keep up with all of these places

@leogrim

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepToLibraryCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepToLibraryCommander.scala
@@ -16,6 +16,8 @@ trait KeepToLibraryCommander {
   // Fun helper methods
   def isKeepInLibrary(keepId: Id[Keep], libraryId: Id[Library])(implicit session: RSession): Boolean
   def getKeeps(ktls: Seq[KeepToLibrary])(implicit session: RSession): Seq[Keep]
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepToLibraryCommander.scala
```diff
@@ -79,4 +81,11 @@ class KeepToLibraryCommanderImpl @Inject() (
     val keepsByIds = keepRepo.getByIds(ktls.map(_.keepId).toSet)
     ktls.map(ktl => keepsByIds(ktl.keepId))
   }
+
+  def changeVisibility(ktl: KeepToLibrary, newVisibility: LibraryVisibility)(implicit session: RWSession): KeepToLibrary = {
+    ktlRepo.save(ktl.withVisibility(newVisibility))
+  }
+  def changeOwner(ktl: KeepToLibrary, newOwnerId: Id[User])(implicit session: RWSession): KeepToLibrary = {
+    ktlRepo.save(ktl.withAddedBy(newOwnerId))
+  }
```
ryanpbrewster, line 90: I'm trying out a new thing where only the `ktlCommander` is allowed to modify the `KeepToLibrary` table. It has been kind of nightmarish to find all the various places that we directly mess with the `Keep` table via `keepRepo.save`, so I'm hoping to save some poor schmuck in the future a little hassle.

Thoughts?

andrewconner:
:+1:

PR #16805, opened 2015-08-13
inactivate a user's org data after user deletion

camhashemi:
@andrewconner @ryanpbrewster led to a bug when I deleted a test user with an org membership


camhashemi:
@ryanpbrewster updated

ryanpbrewster:
Awesome, lgtm

PR #16808, opened 2015-08-13
remove test involving org invite w/ role

camhashemi:
@ryanpbrewster @leogrim removing this test because 1) we don't send invites w/ a role, 2) we don't allow people to send invites above their role


ryanpbrewster:
:+1:

PR #16809, opened 2015-08-13
add org-invited users to the org experiment

camhashemi:
@andrewconner @ryanpbrewster


kifi-backend/modules/shoebox/app/com/keepit/shoebox/controllers/OrganizationAccessActions.scala
```diff
@@ -49,12 +49,9 @@ trait OrganizationAccessActions {
         case Success(orgId) =>
           maybeRequest match {
             case request: UserRequest[_] if request.experiments.contains(UserExperimentType.ORGANIZATION) =>
-              val userIdOpt: Option[Id[User]] = request match {
-                case userRequest: UserRequest[A] => Some(userRequest.userId)
-                case _ => None
-              }
```
ryanpbrewster, line 55: Why was it like this before? Didn't we know that it was a `UserRequest[_]` from one line up?

camhashemi, line 55: that's what I would think

PR #16810, opened 2015-08-13
Retiring older, conflicting kf-button

adamjgrant:


adamjgrant:
@cvializ Discovered that we had an existing `.kf-button` that was only used in a few places. I'm unconflicting it with `.kf-btn`

cvializ:
:+1:

PR #16811, opened 2015-08-13
stop sending emails with reply-to being the sender personal email sin…

eishay:
…ce it hurts our spam rating


eishay:
@andrewconner

PR #16812, opened 2015-08-13
Corrections to container and button styling

adamjgrant:
@cvializ This also corrects some styling that it was accidentally inheriting from the old `.kf-button`

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/common/components/button.styl
+++ b/kifi-backend/modules/shoebox/angular/src/common/components/button.styl
@@ -1,12 +1,19 @@
 .kf-button
   padding: 4px defaultPadding
-  border-radius: defaultBorderRadius/2
-  background: white
-  border: 1px solid textGreyColor
-  color: textGreyColor
+  border-radius: 5px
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/angular/src/app.styl
```diff
@@ -1,4 +1,6 @@
 html
+  font-size: 13px
```
adamjgrant: wtf...

kifi-backend/modules/shoebox/angular/src/common/components/button.styl
```diff
@@ -17,7 +24,7 @@
   .symbol-sprite
     width: 25px
     height: 25px
-    fill: textGreyColor
+    fill: #666
```
cvializ, line 27: not a criticism, just curious: why did you decide to use the hex value instead of a variable?

adamjgrant, line 27: Only because it came from the default styling that existed already. 

I think we should assign this stuff to variables, but they're a bit all over the place right now. I don't think they were used very strictly.

cvializ:
:+1:

PR #16814, opened 2015-08-14
Don't assume String.prototype.trimLeft is available

cvializ:
@adamjgrant @andrewconner fixes a few airbrakes https://fortytwo.airbrake.io/projects/113130/groups/1486367662393960754/notices/1486367662393960753


adamjgrant:
Nice.

PR #16816, opened 2015-08-14
Operation Gatekeeper

ryanpbrewster:
In the middle of editing `copyKeeps` and `moveKeeps`

Need to implement:
  KeepsCommander.createKeep
  KeepsCommander.moveKeep

Next:
  Rename to "KeepCommander" (singular)
  For god's sake stop using this request/response thing for something as
  as simple as KTLCommander

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
@@ -31,7 +31,7 @@ import scala.collection.mutable
 import scala.concurrent.{ ExecutionContext, Future }
 import akka.actor.Scheduler
 import com.keepit.eliza.ElizaServiceClient
-import scala.util.{ Success, Failure }
+import scala.util.{ Try, Success, Failure }
 import com.keepit.common.healthcheck.{ StackTrace, AirbrakeNotifier }
 import com.keepit.common.performance._
 import com.keepit.common.domain.DomainToNameMapper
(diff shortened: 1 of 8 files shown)


andrewconner:
> For god's sake stop using this request/response thing for something as
> as simple as KTLCommander

What does this mean?

ryanpbrewster:
Right now `ktlCommander.internKeepInLibrary` is a one-argument function that takes a `KeepToLibraryInternRequest(keepId, libraryId, userId)`. Instead, it could be a 3-argument function. I've come around to that style of thinking.

andrewconner:
Oh, I see. Yeah a bit odd, haven't seen that.

ryanpbrewster:
One of my goals here is to reduce the number of places where the models are mutated. Right now everybody just plays directly with the DB via `xxxRepo.save` and it's extremely fragile. We've run into a bunch of issues where the DB tables get put into an inconsistent state because we don't defer to smart commander methods wherever possible.

ryanpbrewster:
Yeah, the request/response thing was 100% on me. I made it that way in a textbook case of second-system architecture.

andrewconner:
Alright. One thing to watch for is not making things too cumbersome. `.save` has the nice feature of not caring what you did; obviously, we also know the drawbacks. You don't want to end up having to make `someCommander.updateState()`-type functions for every field in a model.

andrewconner:
For instance, what does `updateNote` give you? `keepRepo.save(keep.withNote(newNote))` is basically what anyone doing it would do.

andrewconner:
`deactivateKeep` I can see, since you may want to do a few things.

ryanpbrewster:
Fair point. But I think we may want to,say, reindex a keep when it's note
is changed.

andrewconner:
That already happens.

andrewconner:
(The whole purpose of using `.save`, which increments sequence numbers)

ryanpbrewster:
@andrewconner @leogrim @camhashemi

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
```diff
@@ -728,6 +764,101 @@ class KeepsCommander @Inject() (
...
+      originalKeeperId = k.originalKeeperId.orElse(Some(userId))
+    )
+
+    currentKeepOpt match {
+      case None =>
+        val persistedKeep = createKeep(newKeep)
+        combineTags(k.id.get, persistedKeep.id.get)
+        Right(persistedKeep)
+
+      case Some(existingKeep) if existingKeep.state == KeepStates.INACTIVE =>
```
camhashemi: maybe use `.isInactive` here or not use `.isInactive` elsewhere for consistency

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
```diff
@@ -728,6 +764,101 @@ class KeepsCommander @Inject() (
...
+  private def combineTags(oldKeepId: Id[Keep], newKeepId: Id[Keep])(implicit s: RWSession) = {
+    val oldSet = keepToCollectionRepo.getCollectionsForKeep(oldKeepId).toSet
+    val existingSet = keepToCollectionRepo.getCollectionsForKeep(newKeepId).toSet
+    val tagsToAdd = oldSet.diff(existingSet)
+    tagsToAdd.map { tagId =>
+      keepToCollectionRepo.getOpt(newKeepId, tagId) match {
+        case None =>
+          keepToCollectionRepo.save(KeepToCollection(keepId = newKeepId, collectionId = tagId))
+        case Some(ktc) if ktc.state == KeepToCollectionStates.INACTIVE =>
+          keepToCollectionRepo.save(ktc.copy(state = KeepToCollectionStates.ACTIVE))
```
camhashemi: just talked with @leogrim about this, we shouldn't "reactivate" anything. instead make a new KTC with the desired fields, only copying `id`

ryanpbrewster: Great catch, I missed that.

On Fri, Aug 14, 2015 at 2:01 PM, Cameron Hashemi notifications@github.com
wrote:

> In
> kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepsCommander.scala
> https://github.com/kifi/keepit/pull/16816#discussion_r37119041:
> 
> > -        combineTags(k.id.get, existingKeep.id.get)
> > -        Left(LibraryError.AlreadyExistsInDest)
> > -    }
> > -  }
> >   +
> > -  // combine tag info on both keeps & saves difference on the new Keep
> > -  private def combineTags(oldKeepId: Id[Keep], newKeepId: Id[Keep])(implicit s: RWSession) = {
> > -    val oldSet = keepToCollectionRepo.getCollectionsForKeep(oldKeepId).toSet
> > -    val existingSet = keepToCollectionRepo.getCollectionsForKeep(newKeepId).toSet
> > -    val tagsToAdd = oldSet.diff(existingSet)
> > -    tagsToAdd.map { tagId =>
> > -      keepToCollectionRepo.getOpt(newKeepId, tagId) match {
> > -        case None =>
> > -          keepToCollectionRepo.save(KeepToCollection(keepId = newKeepId, collectionId = tagId))
> > -        case Some(ktc) if ktc.state == KeepToCollectionStates.INACTIVE =>
> > -          keepToCollectionRepo.save(ktc.copy(state = KeepToCollectionStates.ACTIVE))
> 
> just talked with @leogrim https://github.com/leogrim about this, we
> shouldn't "reactivate" anything. instead make a new KTC with the desired
> fields, only copying id
> 
> —
> https://github.com/kifi/keepit/pull/16816/files#r37119041.

## 

Software Engineer

PR #16818, opened 2015-08-14
Fix incorrect top spacing of persona libraries' titles

cvializ:
| Before | After |
| --- | --- |
| screen shot 2015-08-13 at 5 49 29 pm | screen shot 2015-08-13 at 5 49 42 pm |

@adamjgrant @andrewconner 

The System Library Star/Lock icon was being shown, just without any image inside.


adamjgrant:
Much better!

PR #16819, opened 2015-08-14
adding a 'live orgs' admin page

eishay:
and avoid starting futures in a db session


kifi-backend/modules/shoebox/app/com/keepit/model/LibraryRepo.scala
```diff
@@ -506,6 +507,13 @@ class LibraryRepoImpl @Inject() (
     } yield t).sortBy(_.updatedAt.desc).take(limit).list
   }
 
+  def orgsWithMostLibs()(implicit session: RSession): Seq[(Id[Organization], Int)] = {
+    import StaticQuery.interpolation
```
andrewconner, line 511: Just seeing this. This is a compile warning and has issues. Should be:

```
```

PR #16824, opened 2015-08-14
Hiding quick add when library is empty.

adamjgrant:
https://app.asana.com/0/42964066412484/44137262387851/f

@cvializ @andrewconner


cvializ:
oops clicked merge not comment :fearful: , but :+1:

PR #16828, opened 2015-08-14
Semi-fix for managing personas in dev

maxov:
@andrewconner 
When running the app in dev mode, there are no registered personas, so the persona list comes to the web app completely empty. The way this was handled previously was to not show the contents of persona modal (including the close button), making the frontend inoperable. I made a few quick fixes to the logic in order to make it possible to close the modal when there are no personas. It still looks hacky and is somewhat inconvenient to use the site in dev mode (because the guide is started right after), but at least it is operable. There should be no changes in behavior for environments that have more than one persona (also known as production :stuck_out_tongue: ).

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/persona/managePersona.js
+++ b/kifi-backend/modules/shoebox/angular/src/persona/managePersona.js
@@ -65,7 +65,7 @@ angular.module('kifi')
         };
 
         scope.close = function() {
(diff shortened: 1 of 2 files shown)


ryanpbrewster:
Wow, _that_ was the reason you couldn't use the frontend in dev mode? Uh...thanks for the fix, Max. I'm a little embarrassed that I never looked into it.

andrewconner:
:+1:

PR #16829, opened 2015-08-14
refactor "user accepted your org invitation"

camhashemi:
@ryanpbrewster @leogrim


kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
```diff
@@ -337,9 +337,9 @@ class OrganizationInviteCommanderImpl @Inject() (db: Database,
       elizaClient.sendGlobalNotification( //push sent
         userIds = Set(inviterId),
         title = title,
-        body = s"Click here to see ${invitee.firstName}'s profile.",
-        linkText = s"See ${invitee.firstName}’s profile",
-        linkUrl = s"https://www.kifi.com/${invitee.username.value}",
+        body = s"Click here to view ${org.name}'s libraries.",
```
andrewconner: How about using ’ like below?

PR #16832, opened 2015-08-14
Found a few more missing pieces of ktl tracking

ryanpbrewster:
@leogrim

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepToLibraryCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepToLibraryCommander.scala
@@ -93,8 +93,10 @@ class KeepToLibraryCommanderImpl @Inject() (
   def changeOwner(ktl: KeepToLibrary, newOwnerId: Id[User])(implicit session: RWSession): KeepToLibrary = {
     ktlRepo.save(ktl.withAddedBy(newOwnerId))
   }
+
+  def softRequire(b: Boolean, m: String): Unit = if (!b) airbrake.notify(m)
   def changeUriIdForKeep(keep: Keep, newUriId: Id[NormalizedURI])(implicit session: RWSession): Unit = {
-    require(keep.uriId == newUriId, "URI and Keep don't match.")
+    softRequire(keep.uriId == newUriId, "URI and Keep don't match.") // TODO(ryan): once you're not scared of this anymore, change it to a hard `require`
     ktlRepo.getAllByKeepId(keep.id.get).foreach { ktl =>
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepToLibraryCommander.scala
```diff
@@ -88,4 +93,12 @@ class KeepToLibraryCommanderImpl @Inject() (
   def changeOwner(ktl: KeepToLibrary, newOwnerId: Id[User])(implicit session: RWSession): KeepToLibrary = {
     ktlRepo.save(ktl.withAddedBy(newOwnerId))
   }
+
+  def softRequire(b: Boolean, m: String): Unit = if (!b) airbrake.notify(m)
+  def changeUriIdForKeep(keep: Keep, newUriId: Id[NormalizedURI])(implicit session: RWSession): Unit = {
+    softRequire(keep.uriId == newUriId, "URI and Keep don't match.") // TODO(ryan): once you're not scared of this anymore, change it to a hard `require`
```
leogrim, line 99: Why do you even need the `newUriId` parameter if you're passing the whole keep?

ryanpbrewster, line 99: Right now I am passing in the whole keep strictly as a safety measure (to
allow me to do a `require` check). Eventually I'd like to pass only the
keepId.

On Fri, Aug 14, 2015 at 3:27 PM, Léo Grimaldi notifications@github.com
wrote:

> In
> kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepToLibraryCommander.scala
> https://github.com/kifi/keepit/pull/16832#discussion_r37125341:
> 
> > @@ -88,4 +93,12 @@ class KeepToLibraryCommanderImpl @Inject() (
> >    def changeOwner(ktl: KeepToLibrary, newOwnerId: Id[User])(implicit session: RWSession): KeepToLibrary = {
> >      ktlRepo.save(ktl.withAddedBy(newOwnerId))
> >    }
> > +
> > -  def softRequire(b: Boolean, m: String): Unit = if (!b) airbrake.notify(m)
> > -  def changeUriIdForKeep(keep: Keep, newUriId: Id[NormalizedURI])(implicit session: RWSession): Unit = {
> > -    softRequire(keep.uriId == newUriId, "URI and Keep don't match.") // TODO(ryan): once you're not scared of this anymore, change it to a hard `require`
> 
> Why do you even need the newUriId parameter if you're passing the whole
> keep?
> 
> —
> https://github.com/kifi/keepit/pull/16832/files#r37125341.

## 

Software Engineer

PR #16834, opened 2015-08-14
Payments PlanManagementCommander

stkem:
next up, a whole lotta tests (then payment processing, then admin, …)

### DIFF
--- a/kifi-backend/modules/common/conf/evolutions/shoebox/376.sql
+++ b/kifi-backend/modules/common/conf/evolutions/shoebox/376.sql
@@ -4,7 +4,7 @@
 
 
 alter table account_event
-    add UNIQUE INDEX account_event_u_event_group (event_group);
+    alter column event_group varchar(36);
 
 alter table account_event
     add INDEX account_event_i_account_id_event_time (account_id, event_time);
--- a/kifi-backend/modules/common/conf/evolutions/shoebox/376.sql
+++ b/kifi-backend/modules/common/conf/evolutions/shoebox/376.sql
@@ -1,0 +1,15 @@
+# SHOEBOX
+
+# --- !Ups
+
+
+alter table account_event
+    add UNIQUE INDEX account_event_u_event_group (event_group);
+
+alter table account_event
+    add INDEX account_event_i_account_id_event_time (account_id, event_time);
+
+
+insert into evolutions (name, description) values('376.sql', 'better indexing for account_event');
+
+# --- !Downs
--- a/kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEvent.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEvent.scala
@@ -131,11 +131,11 @@ object AccountEventAction { //There is probably a deeper type hirachy that can b
 
 }
 
-case class EventGroup(id: Long) extends AnyVal
+case class EventGroup(id: String) extends AnyVal
 
 object EventGroup {
   def apply(): EventGroup = {
-    EventGroup(scala.util.Random.nextLong)
+    EventGroup(java.util.UUID.randomUUID.toString)
   }
 }
 
--- a/kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEvent.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEvent.scala
@@ -1,6 +1,7 @@
 package com.keepit.payments
 
 import com.keepit.common.db.{ States, ModelWithState, Id, State }
+import com.keepit.common.crypto.{ ModelWithPublicId, ModelWithPublicIdCompanion }
 import com.keepit.common.time._
 import com.keepit.model.User
 import com.keepit.common.mail.EmailAddress
@@ -10,6 +11,10 @@ import com.kifi.macros.json
 import play.api.libs.json.{ JsValue, JsNull, Json }
 
 import org.joda.time.DateTime
+
+import javax.crypto.spec.IvParameterSpec
+
+case class ActionAttribution(user: Option[Id[User]], admin: Option[Id[User]])
 
 trait AccountEventAction {
   def eventType: String
@@ -92,7 +97,7 @@ object AccountEventAction { //There is probably a deeper type hirachy that can b
   }
 
   @json
-  case class DefaultPaymentMethodChanged(from: Id[PaymentMethod], to: Id[PaymentMethod]) extends AccountEventAction {
+  case class DefaultPaymentMethodChanged(from: Option[Id[PaymentMethod]], to: Id[PaymentMethod]) extends AccountEventAction {
     def eventType: String = "default_payment_method_changed"
     def toDbRow: (String, JsValue) = eventType -> Json.toJson(this)
   }
@@ -128,6 +133,12 @@ object AccountEventAction { //There is probably a deeper type hirachy that can b
 
 case class EventGroup(id: Long) extends AnyVal
 
+object EventGroup {
+  def apply(): EventGroup = {
+    EventGroup(scala.util.Random.nextLong)
+  }
+}
+
 case class AccountEvent(
     id: Option[Id[AccountEvent]] = None,
     createdAt: DateTime = currentDateTime,
@@ -141,16 +152,19 @@ case class AccountEvent(
     whoDunnitExtra: JsValue,
     kifiAdminInvolved: Option[Id[User]],
     action: AccountEventAction,
-    creditChange: DollarAmount, //in cents
+    creditChange: DollarAmount,
     paymentMethod: Option[Id[PaymentMethod]],
-    paymentCharge: Option[DollarAmount], //in cents
-    memo: Option[String]) extends ModelWithState[AccountEvent] {
+    paymentCharge: Option[DollarAmount],
+    memo: Option[String]) extends ModelWithPublicId[AccountEvent] with ModelWithState[AccountEvent] {
 
   def withId(id: Id[AccountEvent]): AccountEvent = this.copy(id = Some(id))
   def withUpdateTime(now: DateTime): AccountEvent = this.copy(updatedAt = now)
 }
 
-object AccountEvent {
+object AccountEvent extends ModelWithPublicIdCompanion[AccountEvent] {
+
+  protected[this] val publicIdPrefix = "ae"
+  protected[this] val publicIdIvSpec = new IvParameterSpec(Array(-72, -49, 51, -61, 42, 43, 123, -61, 64, 122, -121, -55, 117, -51, 12, 21))
 
   def applyFromDbRow(
     id: Option[Id[AccountEvent]],
--- a/kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEvent.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEvent.scala
@@ -210,6 +210,24 @@ object AccountEvent extends ModelWithPublicIdCompanion[AccountEvent] {
       e.billingRelated, e.whoDunnit, e.whoDunnitExtra, e.kifiAdminInvolved, eventType,
       extras, e.creditChange, e.paymentMethod, e.paymentCharge, e.memo))
   }
+
+  def simpleNonBillingEvent(eventTime: DateTime, accountId: Id[PaidAccount], attribution: ActionAttribution, action: AccountEventAction, pending: Boolean = false) = {
+    AccountEvent(
+      state = if (pending) AccountEventStates.PENDING else AccountEventStates.ACTIVE,
+      eventGroup = EventGroup(),
+      eventTime = eventTime,
+      accountId = accountId,
+      billingRelated = false,
+      whoDunnit = attribution.user,
+      whoDunnitExtra = JsNull,
+      kifiAdminInvolved = attribution.admin,
+      action = action,
+      creditChange = DollarAmount(0),
+      paymentMethod = None,
+      paymentCharge = None,
+      memo = None
+    )
+  }
 }
 
 object AccountEventStates extends States[AccountEvent] {
--- a/kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEventRepo.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEventRepo.scala
@@ -20,6 +20,10 @@ trait AccountEventRepo extends Repo[AccountEvent] {
 
   def getByAccountAndState(accountId: Id[PaidAccount], state: State[AccountEvent])(implicit session: RSession): Seq[AccountEvent]
 
+  def getEventsBefore(accountId: Id[PaidAccount], before: DateTime, limit: Int, billingFilter: Option[Boolean])(implicit session: RSession): Seq[AccountEvent]
+
+  def getEvents(accountId: Id[PaidAccount], limit: Int, billingFilter: Option[Boolean])(implicit session: RSession): Seq[AccountEvent]
+
 }
 
 @Singleton
@@ -68,6 +72,22 @@ class AccountEventRepoImpl @Inject() (
 
   def getByAccountAndState(accountId: Id[PaidAccount], state: State[AccountEvent])(implicit session: RSession): Seq[AccountEvent] = {
     (for (row <- rows if row.accountId === accountId && row.state === state) yield row).list
+  }
+
+  def getEventsBefore(accountId: Id[PaidAccount], before: DateTime, limit: Int, billingFilterOpt: Option[Boolean])(implicit session: RSession): Seq[AccountEvent] = {
+    billingFilterOpt.map { billingFilter =>
+      (for (row <- rows if row.accountId === accountId && row.eventTime < before && row.state =!= AccountEventStates.INACTIVE && row.billingRelated === billingFilter) yield row).sortBy(_.eventTime desc).take(limit).list
+    } getOrElse {
+      (for (row <- rows if row.accountId === accountId && row.eventTime < before && row.state =!= AccountEventStates.INACTIVE) yield row).sortBy(_.eventTime desc).take(limit).list
+    }
+  }
+
+  def getEvents(accountId: Id[PaidAccount], limit: Int, billingFilterOpt: Option[Boolean])(implicit session: RSession): Seq[AccountEvent] = {
+    billingFilterOpt.map { billingFilter =>
+      (for (row <- rows if row.accountId === accountId && row.state =!= AccountEventStates.INACTIVE && row.billingRelated === billingFilter) yield row).sortBy(_.eventTime desc).take(limit).list
+    } getOrElse {
+      (for (row <- rows if row.accountId === accountId && row.state =!= AccountEventStates.INACTIVE) yield row).sortBy(_.eventTime desc).take(limit).list
+    }
   }
 
 }
--- a/kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEventRepo.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEventRepo.scala
@@ -35,7 +35,7 @@ class AccountEventRepoImpl @Inject() (
   import db.Driver.simple._
 
   implicit val dollarAmountColumnType = MappedColumnType.base[DollarAmount, Int](_.cents, DollarAmount(_))
-  implicit val eventGroupColumnType = MappedColumnType.base[EventGroup, Long](_.id, EventGroup(_))
+  implicit val eventGroupColumnType = MappedColumnType.base[EventGroup, String](_.id, EventGroup(_))
 
   type RepoImpl = AccountEventTable
   class AccountEventTable(tag: Tag) extends RepoTable[AccountEvent](db, tag, "paid_plan") {
--- a/kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEventRepo.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEventRepo.scala
@@ -20,9 +20,9 @@ trait AccountEventRepo extends Repo[AccountEvent] {
 
   def getByAccountAndState(accountId: Id[PaidAccount], state: State[AccountEvent])(implicit session: RSession): Seq[AccountEvent]
 
-  def getEventsBefore(accountId: Id[PaidAccount], before: DateTime, limit: Int, billingFilter: Option[Boolean])(implicit session: RSession): Seq[AccountEvent]
+  def getEventsBefore(accountId: Id[PaidAccount], before: DateTime, limit: Int, onlyRelatedToBillingOpt: Option[Boolean])(implicit session: RSession): Seq[AccountEvent]
 
-  def getEvents(accountId: Id[PaidAccount], limit: Int, billingFilter: Option[Boolean])(implicit session: RSession): Seq[AccountEvent]
+  def getEvents(accountId: Id[PaidAccount], limit: Int, onlyRelatedToBillingOpt: Option[Boolean])(implicit session: RSession): Seq[AccountEvent]
 
 }
 
@@ -74,17 +74,17 @@ class AccountEventRepoImpl @Inject() (
     (for (row <- rows if row.accountId === accountId && row.state === state) yield row).list
   }
 
-  def getEventsBefore(accountId: Id[PaidAccount], before: DateTime, limit: Int, billingFilterOpt: Option[Boolean])(implicit session: RSession): Seq[AccountEvent] = {
-    billingFilterOpt.map { billingFilter =>
-      (for (row <- rows if row.accountId === accountId && row.eventTime < before && row.state =!= AccountEventStates.INACTIVE && row.billingRelated === billingFilter) yield row).sortBy(_.eventTime desc).take(limit).list
+  def getEventsBefore(accountId: Id[PaidAccount], before: DateTime, limit: Int, onlyRelatedToBillingOpt: Option[Boolean])(implicit session: RSession): Seq[AccountEvent] = {
+    onlyRelatedToBillingOpt.map { onlyRelatedToBilling =>
+      (for (row <- rows if row.accountId === accountId && row.eventTime < before && row.state =!= AccountEventStates.INACTIVE && row.billingRelated === onlyRelatedToBilling) yield row).sortBy(_.eventTime desc).take(limit).list
     } getOrElse {
       (for (row <- rows if row.accountId === accountId && row.eventTime < before && row.state =!= AccountEventStates.INACTIVE) yield row).sortBy(_.eventTime desc).take(limit).list
     }
   }
 
-  def getEvents(accountId: Id[PaidAccount], limit: Int, billingFilterOpt: Option[Boolean])(implicit session: RSession): Seq[AccountEvent] = {
-    billingFilterOpt.map { billingFilter =>
-      (for (row <- rows if row.accountId === accountId && row.state =!= AccountEventStates.INACTIVE && row.billingRelated === billingFilter) yield row).sortBy(_.eventTime desc).take(limit).list
+  def getEvents(accountId: Id[PaidAccount], limit: Int, onlyRelatedToBillingOpt: Option[Boolean])(implicit session: RSession): Seq[AccountEvent] = {
+    onlyRelatedToBillingOpt.map { onlyRelatedToBilling =>
+      (for (row <- rows if row.accountId === accountId && row.state =!= AccountEventStates.INACTIVE && row.billingRelated === onlyRelatedToBilling) yield row).sortBy(_.eventTime desc).take(limit).list
     } getOrElse {
       (for (row <- rows if row.accountId === accountId && row.state =!= AccountEventStates.INACTIVE) yield row).sortBy(_.eventTime desc).take(limit).list
     }
--- a/kifi-backend/modules/shoebox/app/com/keepit/payments/PaidAccountRepo.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/payments/PaidAccountRepo.scala
@@ -12,6 +12,8 @@ import com.google.inject.{ ImplementedBy, Inject, Singleton }
 @ImplementedBy(classOf[PaidAccountRepoImpl])
 trait PaidAccountRepo extends Repo[PaidAccount] {
   def getByOrgId(orgId: Id[Organization], excludeStates: Set[State[PaidAccount]] = Set(PaidAccountStates.INACTIVE))(implicit session: RSession): Option[PaidAccount]
+  def getAccountId(orgId: Id[Organization], excludeStates: Set[State[PaidAccount]] = Set(PaidAccountStates.INACTIVE))(implicit session: RSession): Id[PaidAccount]
+  def getActiveByPlan(planId: Id[PaidPlan])(implicit session: RSession): Seq[PaidAccount]
 }
 
 @Singleton
@@ -43,6 +45,14 @@ class PaidAccountRepoImpl @Inject() (
 
   def getByOrgId(orgId: Id[Organization], excludeStates: Set[State[PaidAccount]] = Set(PaidAccountStates.INACTIVE))(implicit session: RSession): Option[PaidAccount] = {
     (for (row <- rows if row.orgId === orgId && !row.state.inSet(excludeStates)) yield row).firstOption
+  }
+
+  def getAccountId(orgId: Id[Organization], excludeStates: Set[State[PaidAccount]] = Set(PaidAccountStates.INACTIVE))(implicit session: RSession): Id[PaidAccount] = {
+    (for (row <- rows if row.orgId === orgId && !row.state.inSet(excludeStates)) yield row.id).first
+  }
+
+  def getActiveByPlan(planId: Id[PaidPlan])(implicit session: RSession): Seq[PaidAccount] = {
+    (for (row <- rows if row.planId === planId && row.state === PaidAccountStates.ACTIVE) yield row).list
   }
 
 }
--- a/kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
@@ -22,6 +22,7 @@ case class PaidPlan(
 
   def withId(id: Id[PaidPlan]): PaidPlan = this.copy(id = Some(id))
   def withUpdateTime(now: DateTime): PaidPlan = this.copy(updatedAt = now)
+  def withState(state: State[PaidPlan]): PaidPlan = this.copy(state = state)
 }
 
 object PaidPlan extends ModelWithPublicIdCompanion[PaidPlan] {
--- a/kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
@@ -1,10 +1,13 @@
 package com.keepit.payments
 
 import com.keepit.common.db.{ States, ModelWithState, Id, State }
+import com.keepit.common.crypto.{ ModelWithPublicId, ModelWithPublicIdCompanion }
 import com.keepit.common.time._
 import com.keepit.model.Name
 
 import org.joda.time.DateTime
+
+import javax.crypto.spec.IvParameterSpec
 
 case class BillingCycle(month: Int) extends AnyVal
 
@@ -15,10 +18,15 @@ case class PaidPlan(
     state: State[PaidPlan] = PaidPlanStates.ACTIVE,
     name: Name[PaidPlan],
     billingCycle: BillingCycle,
-    pricePerCyclePerUser: DollarAmount) extends ModelWithState[PaidPlan] {
+    pricePerCyclePerUser: DollarAmount) extends ModelWithPublicId[PaidPlan] with ModelWithState[PaidPlan] {
 
   def withId(id: Id[PaidPlan]): PaidPlan = this.copy(id = Some(id))
   def withUpdateTime(now: DateTime): PaidPlan = this.copy(updatedAt = now)
+}
+
+object PaidPlan extends ModelWithPublicIdCompanion[PaidPlan] {
+  protected[this] val publicIdPrefix = "pp"
+  protected[this] val publicIdIvSpec = new IvParameterSpec(Array(-72, -49, 51, -61, 42, 43, 123, -61, 64, 122, -121, -55, 117, -51, 12, 21))
 }
 
 object PaidPlanStates extends States[PaidPlan] {
--- a/kifi-backend/modules/shoebox/app/com/keepit/payments/PaymentMethod.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/payments/PaymentMethod.scala
@@ -1,12 +1,14 @@
 package com.keepit.payments
 
 import com.keepit.common.db.{ States, ModelWithState, Id, State }
-
+import com.keepit.common.crypto.{ ModelWithPublicId, ModelWithPublicIdCompanion }
 import com.keepit.common.time._
 
 import scala.slick.lifted.MappedTo
 
 import org.joda.time.DateTime
+
+import javax.crypto.spec.IvParameterSpec
 
 case class StripeToken(token: String)
 
@@ -17,13 +19,17 @@ case class PaymentMethod(
     state: State[PaymentMethod] = PaymentMethodStates.ACTIVE,
     accountId: Id[PaidAccount],
     default: Boolean,
-    stripeToken: StripeToken) extends ModelWithState[PaymentMethod] {
+    stripeToken: StripeToken) extends ModelWithPublicId[PaymentMethod] with ModelWithState[PaymentMethod] {
 
   def withId(id: Id[PaymentMethod]): PaymentMethod = this.copy(id = Some(id))
   def withUpdateTime(now: DateTime): PaymentMethod = this.copy(updatedAt = now)
 }
 
-object PaymentMethod {
+object PaymentMethod extends ModelWithPublicIdCompanion[PaymentMethod] {
+
+  protected[this] val publicIdPrefix = "pm"
+  protected[this] val publicIdIvSpec = new IvParameterSpec(Array(-72, -49, 51, -61, 42, 43, 123, -61, 64, 122, -121, -55, 117, -51, 12, 21))
+
   def applyFromDbRow(id: Option[Id[PaymentMethod]], createdAt: DateTime, updatedAt: DateTime, state: State[PaymentMethod], accountId: Id[PaidAccount], default: Option[Boolean], stripeToken: StripeToken) = {
     PaymentMethod(id, createdAt, updatedAt, state, accountId, default.exists(b => b), stripeToken)
   }
--- a/kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
@@ -48,7 +48,7 @@ trait PlanManagementCommander {
   def addPaymentMethod(orgId: Id[Organization], stripeToken: StripeToken, attribution: ActionAttribution): Unit
   def changeDefaultPaymentMethod(orgId: Id[Organization], newDefault: Id[PaymentMethod], attribution: ActionAttribution): Unit
 
-  def getAccountEvents(orgId: Id[Organization], before: Option[DateTime], max: Int, billingFilter: Option[Boolean]): Seq[AccountEvent]
+  def getAccountEvents(orgId: Id[Organization], before: Option[DateTime], max: Int, onlyRelatedToBilling: Option[Boolean]): Seq[AccountEvent]
 }
 
 class PlanManagementCommanderImpl @Inject() (
@@ -405,12 +405,12 @@ class PlanManagementCommanderImpl @Inject() (
     ))
   }
 
-  def getAccountEvents(orgId: Id[Organization], beforeOpt: Option[DateTime], limit: Int, billingFilter: Option[Boolean]): Seq[AccountEvent] = db.readOnlyMaster { implicit session =>
+  def getAccountEvents(orgId: Id[Organization], beforeOpt: Option[DateTime], limit: Int, onlyRelatedToBilling: Option[Boolean]): Seq[AccountEvent] = db.readOnlyMaster { implicit session =>
     val accountId = orgId2AccountId(orgId)
     beforeOpt.map { before =>
-      accountEventRepo.getEventsBefore(accountId, before, limit, billingFilter)
+      accountEventRepo.getEventsBefore(accountId, before, limit, onlyRelatedToBilling)
     } getOrElse {
-      accountEventRepo.getEvents(accountId, limit, billingFilter)
+      accountEventRepo.getEvents(accountId, limit, onlyRelatedToBilling)
     }
   }
 
--- a/kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
@@ -1,0 +1,418 @@
+package com.keepit.payments
+
+import com.keepit.common.logging.Logging
+import com.keepit.common.db.slick.Database
+import com.keepit.common.db.slick.DBSession.RSession
+import com.keepit.common.db.Id
+import com.keepit.common.time._
+import com.keepit.model.{ Organization, User, Name }
+import com.keepit.common.mail.EmailAddress
+
+import scala.concurrent.ExecutionContext
+
+import play.api.libs.json.JsNull
+
+import com.google.inject.{ ImplementedBy, Inject }
+
+import org.joda.time.DateTime
+
+class UnauthorizedChange(msg: String) extends Exception(msg)
+class InvalidChange(msg: String) extends Exception
+
+@ImplementedBy(classOf[PlanManagementCommanderImpl])
+trait PlanManagementCommander {
+  def createPaidAccountForOrganization(orgId: Id[Organization], planId: Id[PaidPlan]): Unit
+
+  def registerNewUser(orgId: Id[Organization], userId: Id[User], attribution: ActionAttribution): Unit
+  def registerRemovedUser(orgId: Id[Organization], userId: Id[User], attribution: ActionAttribution): Unit
+  def registerNewAdmin(orgId: Id[Organization], userId: Id[User], attribution: ActionAttribution): Unit
+  def registerRemovedAdmin(orgId: Id[Organization], userId: Id[User], attribution: ActionAttribution): Unit
+
+  def removeUserAccountContact(orgId: Id[Organization], userId: Id[User], attribution: ActionAttribution): Unit
+  def removeEmailAccountContact(orgId: Id[Organization], emailAddress: EmailAddress, attribution: ActionAttribution): Unit
+  def addUserAccountContact(orgId: Id[Organization], userId: Id[User], attribution: ActionAttribution): Unit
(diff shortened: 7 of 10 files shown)


kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
```diff
@@ -0,0 +1,418 @@
...
+
+  private def orgId2AccountId(orgId: Id[Organization])(implicit session: RSession): Id[PaidAccount] = {
+    paidAccountRepo.getAccountId(orgId)
+  }
+
+  def createPaidAccountForOrganization(orgId: Id[Organization], planId: Id[PaidPlan]): Unit = {
+    db.readWrite { implicit session =>
+      val plan = paidPlanRepo.get(planId)
+      if (plan.state != PaidPlanStates.ACTIVE) {
+        throw new InvalidChange("plan_not_active")
```
andrewconner, line 72: Why?

kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEvent.scala
```diff
@@ -92,7 +97,7 @@ object AccountEventAction { //There is probably a deeper type hirachy that can b
   }
 
   @json
-  case class DefaultPaymentMethodChanged(from: Id[PaymentMethod], to: Id[PaymentMethod]) extends AccountEventAction {
+  case class DefaultPaymentMethodChanged(from: Option[Id[PaymentMethod]], to: Id[PaymentMethod]) extends AccountEventAction {
```
ryanpbrewster, line 100: Is this so that you can set the default payment method the first time?

kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
```diff
@@ -0,0 +1,418 @@
...
+  paidAccountRepo: PaidAccountRepo,
+  paidPlanRepo: PaidPlanRepo,
+  implicit val defaultContext: ExecutionContext)
+    extends PlanManagementCommander with Logging {
+
+  private def orgId2AccountId(orgId: Id[Organization])(implicit session: RSession): Id[PaidAccount] = {
+    paidAccountRepo.getAccountId(orgId)
+  }
+
+  def createPaidAccountForOrganization(orgId: Id[Organization], planId: Id[PaidPlan]): Unit = {
```
andrewconner, line 68: `: PaidAccount`?

andrewconner, line 68: Or, if failure is normal, some sort of `Either[PaymentFailure, PaidAccount]`

kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
```diff
@@ -0,0 +1,418 @@
...
+        ))
+      }
+    }
+  }
+
+  def registerNewUser(orgId: Id[Organization], userId: Id[User], attribution: ActionAttribution): Unit = db.readWrite { implicit session =>
+    accountEventRepo.save(AccountEvent(
+      state = AccountEventStates.PENDING,
+      eventGroup = EventGroup(),
+      eventTime = currentDateTime,
```
andrewconner: For all of these, injected `clock.now`?

kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
```diff
@@ -0,0 +1,418 @@
...
+      memo = None
+    ))
+  }
+
+  def removeEmailAccountContact(orgId: Id[Organization], emailAddress: EmailAddress, attribution: ActionAttribution): Unit = db.readWrite { implicit session =>
+    val updatedAccount = paidAccountRepo.getByOrgId(orgId) match {
+      case Some(account) => {
+        account.copy(emailContacts = account.emailContacts.filter(_ != emailAddress))
+      }
+      case None => throw new InvalidChange("account_does_not_exists")
```
andrewconner, line 147: Same as before (I won't write it on each), why throw an exception?

ryanpbrewster:
One thing we should definitely do is have a different sort of `state` to indicate inactive payment methods that is distinct from the normal `state` (which is used as a standin for actually deleting something from the db).

kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
```diff
@@ -0,0 +1,418 @@
...
+
+  def addEmailAccountContact(orgId: Id[Organization], emailAddress: EmailAddress, attribution: ActionAttribution): Unit = db.readWrite { implicit session =>
+    val updatedAccount = paidAccountRepo.getByOrgId(orgId) match {
+      case Some(account) => {
+        account.copy(emailContacts = account.emailContacts.filter(_ != emailAddress) :+ emailAddress)
+      }
+      case None => throw new InvalidChange("account_does_not_exists")
+    }
+    paidAccountRepo.save(updatedAccount)
+    accountEventRepo.save(AccountEvent(
```
andrewconner: These seem very repetitive, and mostly the same. Any chance this could be cleaned up / centralized a bit?

kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEvent.scala
```diff
@@ -128,6 +133,12 @@ object AccountEventAction { //There is probably a deeper type hirachy that can b
 
 case class EventGroup(id: Long) extends AnyVal
 
+object EventGroup {
+  def apply(): EventGroup = {
+    EventGroup(scala.util.Random.nextLong)
```
ryanpbrewster: I think `nextLong()` should have empty parens, because it side-effects.

Also, do these need to be actually unique? The birthday problem is going to surface if we're using random numbers hoping they're all unique.

kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
```diff
@@ -0,0 +1,418 @@
...
+      kifiAdminInvolved = adminWho,
+      action = AccountEventAction.SpecialCredit(),
+      creditChange = amount,
+      paymentMethod = None,
+      paymentCharge = None,
+      memo = memo
+    ))
+  }
+
+  def getCurrentCredit(orgId: Id[Organization]): DollarAmount = db.readOnlyMaster { implicit session =>
```
andrewconner, line 213: This is basically just an accessor for `account.credit`; needed?

kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
```diff
@@ -0,0 +1,418 @@
...
+    val account = paidAccountRepo.getByOrgId(orgId) match {
+      case Some(account) => account
+      case None => throw new Exception("account_does_not_exists")
+    }
+    account.credit
+  }
+
+  def createNewPlan(name: Name[PaidPlan], billingCycle: BillingCycle, price: DollarAmount, custom: Boolean = false): PaidPlan = db.readWrite { implicit session =>
+    paidPlanRepo.save(PaidPlan(
+      state = if (custom) PaidPlanStates.CUSTOM else PaidPlanStates.ACTIVE,
```
andrewconner, line 223: @leogrim has been pushing for using less states in `state`, since something like `CUSTOM` isn't really about the row.

kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
```diff
@@ -0,0 +1,418 @@
...
+      name = name,
+      billingCycle = billingCycle,
+      pricePerCyclePerUser = price
+    ))
+  }
+
+  def grandfatherPlan(id: Id[PaidPlan]): Unit = db.readWrite { implicit session =>
+    val plan = paidPlanRepo.get(id)
+    if (plan.state == PaidPlanStates.ACTIVE) {
+      paidPlanRepo.save(plan.copy(state = PaidPlanStates.GRANDFATHERED))
```
andrewconner: What does this mean?

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
```diff
@@ -15,12 +18,17 @@ case class PaidPlan(
...
-    pricePerCyclePerUser: DollarAmount) extends ModelWithState[PaidPlan] {
+    pricePerCyclePerUser: DollarAmount) extends ModelWithPublicId[PaidPlan] with ModelWithState[PaidPlan] {
 
   def withId(id: Id[PaidPlan]): PaidPlan = this.copy(id = Some(id))
   def withUpdateTime(now: DateTime): PaidPlan = this.copy(updatedAt = now)
 }
 
+object PaidPlan extends ModelWithPublicIdCompanion[PaidPlan] {
+  protected[this] val publicIdPrefix = "pp"
+  protected[this] val publicIdIvSpec = new IvParameterSpec(Array(-72, -49, 51, -61, 42, 43, 123, -61, 64, 122, -121, -55, 117, -51, 12, 21))
```
andrewconner, line 30: This is `Library`'s `IvSpec`

kifi-backend/modules/shoebox/app/com/keepit/payments/PaymentMethod.scala
```diff
@@ -17,13 +19,17 @@ case class PaymentMethod(
...
 
   def withId(id: Id[PaymentMethod]): PaymentMethod = this.copy(id = Some(id))
   def withUpdateTime(now: DateTime): PaymentMethod = this.copy(updatedAt = now)
 }
 
-object PaymentMethod {
+object PaymentMethod extends ModelWithPublicIdCompanion[PaymentMethod] {
+
+  protected[this] val publicIdPrefix = "pm"
+  protected[this] val publicIdIvSpec = new IvParameterSpec(Array(-72, -49, 51, -61, 42, 43, 123, -61, 64, 122, -121, -55, 117, -51, 12, 21))
```
andrewconner, line 31: Same, and below same. :) Generate a new one

kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEvent.scala
ryanpbrewster: https://en.wikipedia.org/wiki/Birthday_attack shows a 0.0001% chance of at least one collision by the time we have 6M accounts, which is admittedly pretty low.

But...consider the expectation value of how bad a collision is. IMO that's about the worst possible case, and totally worth engineering around.

ryanpbrewster: And that's assuming that the Scala random number generator is perfectly random.

kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEventRepo.scala
```diff
@@ -70,4 +74,20 @@ class AccountEventRepoImpl @Inject() (
     (for (row <- rows if row.accountId === accountId && row.state === state) yield row).list
   }
 
+  def getEventsBefore(accountId: Id[PaidAccount], before: DateTime, limit: Int, billingFilterOpt: Option[Boolean])(implicit session: RSession): Seq[AccountEvent] = {
```
ryanpbrewster: Can we call `billingFilterOpt` something more descriptive, like `onlyRelatedToBillingOpt`?

kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEventRepo.scala
```diff
@@ -70,4 +74,20 @@ class AccountEventRepoImpl @Inject() (
...
+    billingFilterOpt.map { billingFilter =>
+      (for (row <- rows if row.accountId === accountId && row.eventTime < before && row.state =!= AccountEventStates.INACTIVE && row.billingRelated === billingFilter) yield row).sortBy(_.eventTime desc).take(limit).list
+    } getOrElse {
+      (for (row <- rows if row.accountId === accountId && row.eventTime < before && row.state =!= AccountEventStates.INACTIVE) yield row).sortBy(_.eventTime desc).take(limit).list
+    }
```
ryanpbrewster, line 82: I actually think a `match` makes more sense here. It'd be 3 lines instead of 5, and it's easier to read at a glance.

kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEventRepo.scala
```diff
@@ -70,4 +74,20 @@ class AccountEventRepoImpl @Inject() (
...
+  }
+
+  def getEvents(accountId: Id[PaidAccount], limit: Int, billingFilterOpt: Option[Boolean])(implicit session: RSession): Seq[AccountEvent] = {
+    billingFilterOpt.map { billingFilter =>
+      (for (row <- rows if row.accountId === accountId && row.state =!= AccountEventStates.INACTIVE && row.billingRelated === billingFilter) yield row).sortBy(_.eventTime desc).take(limit).list
+    } getOrElse {
+      (for (row <- rows if row.accountId === accountId && row.state =!= AccountEventStates.INACTIVE) yield row).sortBy(_.eventTime desc).take(limit).list
+    }
```
ryanpbrewster, line 90: Same thing, I think `match` is prettier.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
```diff
@@ -15,12 +18,17 @@ case class PaidPlan(
     state: State[PaidPlan] = PaidPlanStates.ACTIVE,
     name: Name[PaidPlan],
     billingCycle: BillingCycle,
-    pricePerCyclePerUser: DollarAmount) extends ModelWithState[PaidPlan] {
+    pricePerCyclePerUser: DollarAmount) extends ModelWithPublicId[PaidPlan] with ModelWithState[PaidPlan] {
```
ryanpbrewster, line 21: This seems like it's hard-coding our billing strategy. Are we for sure going to charge per cycle per user?

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
```diff
@@ -15,12 +18,17 @@ case class PaidPlan(
...
 }
 
+object PaidPlan extends ModelWithPublicIdCompanion[PaidPlan] {
+  protected[this] val publicIdPrefix = "pp"
+  protected[this] val publicIdIvSpec = new IvParameterSpec(Array(-72, -49, 51, -61, 42, 43, 123, -61, 64, 122, -121, -55, 117, -51, 12, 21))
+}
+
 object PaidPlanStates extends States[PaidPlan] {
   val GRANDFATHERED = State[PaidPlan]("grandfathered")
   val CUSTOM = State[PaidPlan]("custom")
```
ryanpbrewster, line 35: Can we have a different field instead of overloading `state`? I like being sure that `state != ACTIVE` means the model is dead. These seem orthogonal to that concern. `category` or `kind` or something?

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
```diff
@@ -1,11 +1,14 @@
...
 import com.keepit.common.time._
 import com.keepit.model.Name
 
 import org.joda.time.DateTime
 
+import javax.crypto.spec.IvParameterSpec
+
 case class BillingCycle(month: Int) extends AnyVal
 
 case class PaidPlan(
```
ryanpbrewster, line 14: `PaymentPlan`, maybe?

kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
ryanpbrewster, line 68: I've been talking to @eishay a bit, and I'm starting to come around to `Try[A]` instead of `Either[F, A]` for a thing that can succeed or fail. It feels more semantically explicit.

kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEvent.scala
stkem, line 100: Yup

kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEvent.scala
stkem: These do need to be unique indeed, but only within one account, so the space is much smaller (i.e. unlikely to get to 6M or more). Also, a collision would not be catastrophic, it would simply yield to events being processed in one batch that don't actually go together. Not good, but not the end of the world.
Nevertheless, you are right that it's not worth it. I'll change this to 128bit, which will make a collision about as likely as you being hit by a piece of Apollo 13 twice, right now.

kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEventRepo.scala
stkem: That does sound better. Changing.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
stkem, line 14: That ship has kinda sailed. This name is all over the place including in the db.
Also, to me, a `PaymentPlan` is s schedule by which you make what is effectively a single payment, e.g. you buy a car and pay it off on a Payment Plan. Not a indefinitely recurring thing.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
stkem, line 21: Yup, that has been decisively decided :-)

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
stkem, line 35: That assumption (`state != ACTIVE` means the model is dead) is already not true in the codebase. We use state like this in many places, most notably perhaps `NormalizedURI`.
In this particular case, it' s actually degrees of dead, if you will. `GRANDFATHERED` being a state where to model is dead to the world except for accounts still on that plan.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaymentMethod.scala
stkem, line 31: Why does it matter? If we leak a key we don't have to replace them all?

kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
stkem, line 68: I'm with @ryanpbrewster on this one, generally. However, here failure is not normal (this should be called once during org creation, and possibly by some backfilling tool and that's it) and `PaidAccount` is intended as a private model of sorts that should not be exposed outside the payment system. That why all the APIs deal with OrgIds.

kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
stkem, line 72: You can't put an organization a plan that has been deactivated.

kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
stkem: Ah yes, good idea. Changing.

kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
stkem, line 147: The account not existing at any point where this might be called is a serious error. What else do you suggest?

kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
stkem, line 213: Yes, `PaidAccount` is intended as sort of private model to the payment system an should not be exposed as much as possible. That's why all the APIs are dealing in `Id[Organization]`.

kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
stkem: Yeah, I'll make a shorter constructor for the simple case.

kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
stkem: One step to the grave. The plan has been retired, i.e. is not available to new organizations, but we are currently allowing orgs that are already on this plan to stay on it. (See the spec document)

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
stkem, line 35: The assumption that holds id `state == INACTIVE` `=>` `Model Dead`

kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
ryanpbrewster: Can we use `.withState` to make it easier to search for places where the state is being changed?

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
ryanpbrewster, line 35: So consider what the queries look like if you want to find all active plans. If you say `where state = "active"` that is not correct, which may be surprising. I know that we have this situation elsewhere in our codebase, but that is not a reason to propagate it.

This thing we do where we mark things as inactive instead of deleting them is an externally imposed restriction. In a perfect world we wouldn't even have that column, we'd just delete the row. Tacking extra information into that column feels like it's abusing the notion of state.

ryanpbrewster, line 35: And yes, you could query for `state != "inactive"`, but `GRANDFATHERED` and `CUSTOM` are orthogonal considerations (well, not quite orthogonal as you pointed out, but certainly not aligned).

kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEventRepo.scala
```diff
@@ -70,4 +74,20 @@ class AccountEventRepoImpl @Inject() (
     (for (row <- rows if row.accountId === accountId && row.state === state) yield row).list
   }
 
+  def getEventsBefore(accountId: Id[PaidAccount], before: DateTime, limit: Int, onlyRelatedToBillingOpt: Option[Boolean])(implicit session: RSession): Seq[AccountEvent] = {
+    onlyRelatedToBillingOpt.map { onlyRelatedToBilling =>
+      (for (row <- rows if row.accountId === accountId && row.eventTime < before && row.state =!= AccountEventStates.INACTIVE && row.billingRelated === onlyRelatedToBilling) yield row).sortBy(_.eventTime desc).take(limit).list
```
andrewconner, line 79: May be good to have a secondary sorting; this will cause heinsentests if you test it well.

kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEventRepo.scala
```diff
@@ -70,4 +74,20 @@ class AccountEventRepoImpl @Inject() (
...
+    } getOrElse {
+      (for (row <- rows if row.accountId === accountId && row.eventTime < before && row.state =!= AccountEventStates.INACTIVE) yield row).sortBy(_.eventTime desc).take(limit).list
+    }
+  }
+
+  def getEvents(accountId: Id[PaidAccount], limit: Int, onlyRelatedToBillingOpt: Option[Boolean])(implicit session: RSession): Seq[AccountEvent] = {
+    onlyRelatedToBillingOpt.map { onlyRelatedToBilling =>
+      (for (row <- rows if row.accountId === accountId && row.state =!= AccountEventStates.INACTIVE && row.billingRelated === onlyRelatedToBilling) yield row).sortBy(_.eventTime desc).take(limit).list
```
andrewconner, line 87: Same. time, id is a good way to break the tie.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidAccountRepo.scala
```diff
@@ -45,4 +47,12 @@ class PaidAccountRepoImpl @Inject() (
     (for (row <- rows if row.orgId === orgId && !row.state.inSet(excludeStates)) yield row).firstOption
   }
 
+  def getAccountId(orgId: Id[Organization], excludeStates: Set[State[PaidAccount]] = Set(PaidAccountStates.INACTIVE))(implicit session: RSession): Id[PaidAccount] = {
```
andrewconner, line 50: Will all organizations have paid accounts? (ie, should this be `Option`al)?

kifi-backend/modules/shoebox/app/com/keepit/payments/PaymentMethod.scala
andrewconner, line 31: It's unrelated to the key. There's no reason to have library id 1 be the same as payment method id 1.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
andrewconner, line 35: @leogrim may want to hop in here. At least to bring some context, NormalizedURI has moved away from using tons of states to track status, and Léo has worked to move away from this in an attempt to protect against bugs coming from "give me all active rows" checking for the wrong states.

kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
andrewconner, line 72: Yes, I understand the error. Why `throw`?

kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
stkem: Sure, good point. Changing.

kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
andrewconner, line 68: I probably wasn't verbose enough. Returning _something_ is better than `Unit` (which is _really_ Unit or exception right now). Make error handling explicit, Scala is great at that.

Regarding Try or Either, sure, fine with me. Only problem is that Try basically means "we don't deal with specific errors". I wouldn't be opposed to bringing in a right projected Either for general use either. Thoughts on that?

kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
```diff
@@ -0,0 +1,419 @@
...
+
+import com.google.inject.{ ImplementedBy, Inject }
+
+import org.joda.time.DateTime
+
+class UnauthorizedChange(msg: String) extends Exception(msg)
+class InvalidChange(msg: String) extends Exception
+
+@ImplementedBy(classOf[PlanManagementCommanderImpl])
+trait PlanManagementCommander {
```
andrewconner, line 23: Bit concerned about how nearly everything returns a `Unit`. Why not return something useful?

kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
andrewconner: Why even have this method?

kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
andrewconner, line 147: Return an error; right now you're returning Unit or throwing an exception.

andrewconner, line 147: I wouldn't consider this a serious error either. People double click submit buttons all the time.

kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
```diff
@@ -0,0 +1,419 @@
...
+      memo = None
+    ))
+  }
+
+  def addEmailAccountContact(orgId: Id[Organization], emailAddress: EmailAddress, attribution: ActionAttribution): Unit = db.readWrite { implicit session =>
+    val updatedAccount = paidAccountRepo.getByOrgId(orgId) match {
+      case Some(account) => {
+        account.copy(emailContacts = account.emailContacts.filter(_ != emailAddress) :+ emailAddress)
+      }
+      case None => throw new InvalidChange("account_does_not_exists")
```
andrewconner, line 179: This check seems everywhere, why not compose functions? (You may have to change all the `Unit`s though)

kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEvent.scala
```diff
@@ -141,16 +152,19 @@ case class AccountEvent(
...
-    memo: Option[String]) extends ModelWithState[AccountEvent] {
+    paymentCharge: Option[DollarAmount],
+    memo: Option[String]) extends ModelWithPublicId[AccountEvent] with ModelWithState[AccountEvent] {
 
   def withId(id: Id[AccountEvent]): AccountEvent = this.copy(id = Some(id))
   def withUpdateTime(now: DateTime): AccountEvent = this.copy(updatedAt = now)
 }
 
-object AccountEvent {
+object AccountEvent extends ModelWithPublicIdCompanion[AccountEvent] {
```
andrewconner, line 164: You think these things even need public ids?

kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
andrewconner, line 68: One that I like quite a bit: https://github.com/non/cats/blob/master/core/src/main/scala/cats/data/Xor.scala

maxov, line 68: We already have a transitive dependency on scalaz, who also provide a right-biased Either: https://github.com/scalaz/scalaz/blob/series/7.1.x/core/src/main/scala/scalaz/Either.scala

ryanpbrewster, line 68: One that @leogrim showed me last Friday: http://rapture.io/talks/exceptions/sanfrancisco.html

I am super conflicted about best practices regarding error handling. I like `Either[F, A]` for the explicit type-safety; I like `Try[A]` for the semantic clarity and the easy ability to convert things into `Option` or throw the underlying exception. I do wish that there were a way to say "this is a `Try[A]`, but it exclusively catches exceptions of type `F`", and everything else gets thrown.

kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
ryanpbrewster, line 179: I like the idea of a repo method `getOrFail(orgId): PaidAccount` that just throws this exception inside of it.

kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
leogrim, line 223: Indeed, I saw some comments from @ryanpbrewster in that direction as well.

kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
```diff
@@ -0,0 +1,323 @@
...
+  def getAccountContacts(orgId: Id[Organization]): (Seq[Id[User]], Seq[EmailAddress]) = db.readOnlyMaster { implicit session =>
+    paidAccountRepo.getByOrgId(orgId) match {
+      case Some(account) => {
+        (account.userContacts, account.emailContacts)
+      }
+      case None => throw new Exception("account_does_not_exists")
+    }
+  }
+
+  def grantSpecialCredit(orgId: Id[Organization], amount: DollarAmount, adminWho: Option[Id[User]], memo: Option[String]): Unit = db.readWrite { implicit session =>
```
leogrim, line 199: `adminWho` => `grantedBy`?

kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEvent.scala
```diff
@@ -141,16 +152,19 @@ case class AccountEvent(
     whoDunnitExtra: JsValue,
```
leogrim, line 152: What's `whoDunnitExtra` for?

kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
ryanpbrewster, line 199: I like having `admin` explicitly in there. Maybe `grantedByAdmin: Option[Id[User]]`?

kifi-backend/modules/common/conf/evolutions/shoebox/376.sql
```diff
@@ -0,0 +1,15 @@
+# SHOEBOX
+
+# --- !Ups
+
+
+alter table account_event
+    alter column event_group varchar(36);
```
andrewconner, line 7: I replied to the commit, instead of the PR, but... no unique index?

kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
```diff
@@ -0,0 +1,323 @@
...
+
+  def getCurrentCredit(orgId: Id[Organization]): DollarAmount = db.readOnlyMaster { implicit session =>
+    val account = paidAccountRepo.getByOrgId(orgId) match {
+      case Some(account) => account
+      case None => throw new Exception("account_does_not_exists")
+    }
+    account.credit
+  }
+
+  def createNewPlan(name: Name[PaidPlan], billingCycle: BillingCycle, price: DollarAmount, custom: Boolean = false): PaidPlan = db.readWrite { implicit session =>
```
andrewconner, line 221: Default argument here won't do much, since anyone using it injects the trait.

kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
```diff
@@ -0,0 +1,323 @@
...
+  def removeUserAccountContact(orgId: Id[Organization], userId: Id[User], attribution: ActionAttribution): Unit
+  def removeEmailAccountContact(orgId: Id[Organization], emailAddress: EmailAddress, attribution: ActionAttribution): Unit
+  def addUserAccountContact(orgId: Id[Organization], userId: Id[User], attribution: ActionAttribution): Unit
+  def addEmailAccountContact(orgId: Id[Organization], emailAddress: EmailAddress, attribution: ActionAttribution): Unit
+  def getAccountContacts(orgId: Id[Organization]): (Seq[Id[User]], Seq[EmailAddress])
+
+  def grantSpecialCredit(orgId: Id[Organization], amount: DollarAmount, adminWho: Option[Id[User]], memo: Option[String]): Unit
+  def getCurrentCredit(orgId: Id[Organization]): DollarAmount
+
+  def createNewPlan(name: Name[PaidPlan], billingCycle: BillingCycle, price: DollarAmount, custom: Boolean = false): PaidPlan
```
leogrim, line 40: Related to my comment on the repo PR, it may be easier / more explicit to use `name` as the key to manage plans.

kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
leogrim, line 72: Except `PaidPlanState` can also be `GRANDFATHERED` or `CUSTOM` so is checking for `ACTIVE` the right thing, or should we be checking for `not INACTIVE`? 

See where I'm going here?

kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
```diff
@@ -0,0 +1,323 @@
...
+    }
+  }
+
+  def deactivatePlan(id: Id[PaidPlan]): Unit = db.readWrite { implicit session =>
+    val plan = paidPlanRepo.get(id)
+    val accounts = paidAccountRepo.getActiveByPlan(id)
+    if (accounts.isEmpty) {
+      paidPlanRepo.save(plan.withState(PaidPlanStates.INACTIVE))
+    } else {
+      throw new InvalidChange("plan_in_use")
```
andrewconner, line 245: To take some of our Slack conversation here, let's imagine the caller of this (and yes, we can be sloppy with admin controllers, so let's carry this argument to a consumer-facing method as well):

``` scala
def deactivatePlan(id: Id[PaidPlan]) = AdminAction { request =>
  try {
    planManagementCommander.deactivatePlan(id)
    Ok
  } catch {
    case ex: Throwable => NotFound
  }
}
```

or, you can imagine the same with `Try`. Or:

``` scala
def deactivatePlan(id: Id[PaidPlan]) = AdminAction { request =>
  planManagementCommander.deactivatePlan(id).map(_ => Ok).getOrElse(NotFound)
}
```

or, match on `Success` and `Failure`.

My point is that error handling happens somewhere, and by throwing exceptions, the caller doesn't realize they need to based on the signature (and if that point is lost, it's the pattern we use elsewhere in the system, so it is surprising behavior). If we wanted pretty signatures based on normal use-case, we'd use `null` instead of `Option`.

kifi-backend/modules/common/conf/evolutions/shoebox/376.sql
stkem, line 7: I replied to you comment on the PR on why no index. And certainly not a unique index, as multiple events will be in the same group, hence "group" :-)

kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
```diff
@@ -0,0 +1,323 @@
...
+  def createPaidAccountForOrganization(orgId: Id[Organization], planId: Id[PaidPlan]): Unit = {
+    db.readWrite { implicit session =>
+      val plan = paidPlanRepo.get(planId)
+      if (plan.state != PaidPlanStates.ACTIVE) {
+        throw new InvalidChange("plan_not_active")
+      }
+      paidAccountRepo.getByOrgId(orgId, Set()) match {
+        case Some(pa) if pa.state == PaidAccountStates.ACTIVE => throw new InvalidChange("account_exists")
+        case Some(pa) => paidAccountRepo.save(pa.copy(state = PaidAccountStates.ACTIVE, planId = planId))
+        case None => paidAccountRepo.save(PaidAccount(
```
leogrim, line 77: Recent pattern for interning: we should match on `inactivePlanOpt`, create a new object in both cases and copy the inactive plan's id if it exists. This way there's no difference in creating a new object whether there was an inactive on before or not (data from an `INACTIVE` row should be considered gone and should not be trusted).

Then I think this is incorrect with the two other "active-like" states `CUSTOM` and `GRANDFATHERED`.

kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
andrewconner, line 40: I thought this is the human friendly name, ie "Small Team". It's marketing? Right?

kifi-backend/modules/common/conf/evolutions/shoebox/376.sql
andrewconner, line 7: Sorry, unique was a typo. Okay.

kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
andrewconner, line 68: @ryanpbrewster `Either` or cat's `Xor` (which has a really nice syntax) does exactly what you want (and scalaz's right biased either fixes using `.right` everywhere).

Either way, we have many possible tools that aren't `Unit` (and `Exception` if you catch it).

kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
leogrim, line 40: Oh ok, makes sense

leogrim, line 40: Never mind, didn't realize it was only human-friend / marketing.

leogrim, line 40: @stkem is it actually? Or is it a code / key?

In other places, `Name[T]` is used for type-safety when using a name String as a key for something, I guess that's why I thought that.

kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEvent.scala
stkem, line 152: Any additional information about the person making the change, like IP address and User Agent and such. Per request from Eishay. Will be plumbed through when the controllers are coming. Very free form and really only intended for forensics.

kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEvent.scala
stkem, line 164: This one actually probably not. I originally had pagination by `id`, switched it to time based and left the pid. I'll take it out.

kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEventRepo.scala
stkem, line 79: Yup, adding that.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidAccountRepo.scala
stkem, line 50: I think all organization should have a paid account, yes. Even if many will be on a free plan, it simplifies the mental model, i.e. org<->account one-2-one.

kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
stkem, line 40: Yup, it's mainly intended for Human consumption.

kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
stkem, line 68: `PaidAccount` is intended as an internal model to the payment system, i.e. should not be exposed to anyone else, for easier refactoring later (e.g. if we have paid accounts for users). What do you think this should be returning then? `Try[Unit]`?

kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
stkem, line 72: No, checking for `ACTIVE` is correct here. You need special privileges to be on the other two kinds of plans. 
I see where you are going, in that you need to of course check the right things. Putting it in a different field just means you have to check the right thing somewhere else. It's a different mental model, but prone to the same errors.

kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
stkem, line 77: Not sure what `inactivePlanOpt` is here.
This is a `PaidAccount`, which doesn't have any states other than `ACTIVE` and `INACTIVE`.
`PaidAccount` has a uniqueness constraint on the org so you can't just make a new one.

kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
stkem, line 221: It's also in the trait. I like to have the signatures match exactly when possible for clarity.

kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
andrewconner, line 77: He's proposing overwriting the old one by creating a new object and using the old ID. That way no existing fields can survive.

kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
andrewconner, line 68: That's better. Initially, it seems a bit weird to me that the caller of this isn't trusted to see what it created; I could be wrong though. But yeah, explicit control flow is better imho (and you make the caller check the right things!)

kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
andrewconner, line 72: How will admins do it, then? Don't we want to use the same methods?

kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
andrewconner, line 40: Typed `Name` probably isn't the class to use, then (Léo was explaining to me that it's for typed strings like system values where it type changes — in this case, it's always talking about a marketing name)

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidAccountRepo.scala
andrewconner, line 50: Alright, I suppose we'll need to add to the todo list a checker for this. Otherwise we randomly blow up because something didn't get in sync (like now, since orgs don't have PaidAccounts)

kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEvent.scala
andrewconner, line 164: Oh interesting. Paging by id, when the events are close in time, is actually preferred (since it gets hard to page when there's a bunch of rows with the same time). Up to you, though, I'm sure you have a good intuition about this.

kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEvent.scala
andrewconner, line 152: I wonder if it shouldn't be "free form", and instead follow some loosely structured case class. Seems like IP and user agent should always go in?

kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
andrewconner, line 221: Is it more clear, though? I've seen bugs where someone updated one and not the other. If you update this one, you will never see the expected change (since everyone injects the trait). If you update the other and not this, there's misleading "documentation".

andrewconner, line 221: Also, since this method is pretty critical to the system, I'd personally err away from default params; make someone be super explicit.

stkem:
A lot of reviewing energy went into this, which is awesome. We don't do this through checking often enough and it's particularly critical here since this will be dealing with users (or should I say customers) money.
Two things that have surfaced here, the resolution of which should have benefits beyond this PR, are a bit more fundamental disagreements:
- The use of `state`, specifically wether it is a rich field that can encode multiple states (as it has been used in a few other places in the past, sometime with errors, and is still in use) or a more cleaner approach where state is just `ACTIVE` and `INACTIVE` and is basically just used as a form of soft delete, with other, richer information moved to a non standard field on the model.
  I personally find the rich state approach makes or easier use in places (see some other comments on this and the previous PR) while the other one can be a bit more error prone in some situation because you have to check two things instead of one.
  Nevertheless, imho, the code quality/maintainability difference is marginal at best and _iff_ there is reasonable consensus to go with `ACTIVE`/`INACTIVE` only `state` I'll be happy to refactor this. In that case though we should aggressively go after existing uses of rich state, remove them, and change the typing of `State` which is currently at odds with this pattern.
- Function vs Imperative patterns and particularly checked vs. unchecked exception (i.e. should the errors a function can produce be part of it's type or not, google it for some interesting flame wars). People much smarter than all of us have been having this debate for the better part of 30 years without a clear winner, probably because there isn't one, and we have plenty of examples of both our codebase (I wonder if there is a written position somewhere of why scala went with unchecked for regular exceptions, vs. Java checked ones). There is pros and cons for either. This one just needs a decision. It sounds like we have somewhat of a consensus on checked exceptions with the more functional pattern. I don't always agree, but I'm happy to roll with that.

kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
stkem, line 72: This is intended as something that get's called during signup/org creation where only regular (i.e. `ACTIVE` as it stands now) plans are available for the user to choose. The is another method to change the plan of an org which let's admins pick any (not inactive) plan.
Also, unrelated, but since we are on this line: Regarding the exception, this is a "this really should never happen" scenario, which is where I think "real" exceptions still have a place even with a more function checked exceptions pattern (there are other places in this PR where they don;t have a place, admittedly) since there is really nothing the caller can do about the error and it is arguably actually an "internal sever error".

kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEvent.scala
stkem, line 164: To be honest, I had not entirely made up my mind, which is why the public id is still in there. I'll do some testing.

kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
stkem, line 77: Ah, I see. That makes sense. `Plan` vs `Account` threw me off.

stkem:
Also, this is kind of fun. Haven't had a good debate around software patterns in a while.

andrewconner:
Unfortunately, as you alluded, Scala doesn't have checked exceptions, so this isn't really an option for us — the caller can't reasonably know what will happen. Sometimes, this feature is helpful when there's no good way to break flow, but we don't (or if we do, shouldn't?) have a pattern of using error codes in exceptions hoping the receiver catches it and serializes appropriately. 

We actually do have a good amount of consensus regarding staying away from `throw`, _especially especially_ in APIs. There's only 57 `throw`s in Commanders: a few shouldn't be there, a large chunk are in files including names such as "abook" and "auth", and some are explicitly _for_ the control flow features.

ryanpbrewster:
My take on the "rich `state`" is that it doesn't make sense to use words like "active" and "inactive" (which _feel_ like boolean values) in conjunction with a bunch of other values. `state` is best used as a boolean (encoded by a `varchar` strictly for readability). We never want to do anything with "inactive" models, so most repo methods will default to excluding them.

It also makes a lot of sense to have another field encoding the type of a thing. For instance, if we kill off a "custom" `PaidPlan`, it is an inactive custom plan. It doesn't stop being custom just because it's dead, and it can be useful to see that information in the db (for diagnostics). Likewise, most of our code should be dealing exclusively with "active" models, so you grab the active models and then branch based on the `kind` or `type` or whatever you want to call that other field.

stkem:
Closing this and doing some refactoring. :-)

leogrim:
@ryanpbrewster very well summed up the design / engineering reasons for keeping `state` binary. I'll add a couple of things:
- Team-wise, having `state` always define a soft delete, with no business logic piggy backing on it, makes it easier for other people to work on unknown code. We can build a mental model of what `state` is in our system once and for all. In all situations I've encountered, the logic actually ended up clearer separating concerns.
- As a practical win, if we use `ACTIVE` and `INACTIVE` everywhere, we can actually use these in referential integrity constraints.

PR #16844, opened 2015-08-15
change how user library analytics are sent; adds support for sending these analytics to amplitude

joshm1:
mixpanel has a way to increment numeric properties by a certain delta, amplitude does not. so to
pass these properties to amplitude we need to fetch the final values of them and call "setUserProperties" instead

this replaces the main (but not the only) consumer of the heimdal incrementUserProperties endpoint to call an actor
instead that will fetch certain properties for the user and call heimdal.setUserProperties

it's certainly noted that this increases load on the database, but I don't think it's enough to be of concern

the motivation behind using an actor here is that these are "fire & forget" functions, but because this results
more queries hitting the database (or cache), the actor gives us additional flexibility to limit concurrency,
debouce/consolidate db or API calls, scale, <insert other akka benefits here>, etc

@andrewconner @stkem

### DIFF
--- a/kifi-backend/modules/shoebox/angular/bower.json
+++ b/kifi-backend/modules/shoebox/angular/bower.json
@@ -30,7 +30,7 @@
     "fuse.js": "1.1.0",
     "jquery": "2.1.3",
     "jquery-mousewheel": "3.1.12",
-    "lodash": "2.4.1",
+    "lodash": "3.10.1",
     "normalize-css": "2.1.3",
     "underscore.string": "2.3.3",
     "airbrake-js-client": "~0.4.3",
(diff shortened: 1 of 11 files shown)


kifi-backend/modules/shoebox/angular/bower.json
```diff
@@ -30,7 +30,7 @@
     "fuse.js": "1.1.0",
     "jquery": "2.1.3",
     "jquery-mousewheel": "3.1.12",
-    "lodash": "2.4.1",
+    "lodash": "3.10.1",
```
joshm1, line 33: want to point this out because it's a major version change. i reviewed the [Changelog](https://github.com/lodash/lodash/wiki/Changelog#v300) and cross-referenced compatibility warnings with our code. didn't find anything

andrewconner, line 33: Thanks. @adamjgrant @cvializ

andrewconner:
Seems reasonable. Thanks

stkem:
:+1:

PR #16848, opened 2015-08-17
Removing icons from buttons on member manage page.

adamjgrant:


adamjgrant:
@cvializ @andrewconner

cvializ:
:ok_hand:

PR #16849, opened 2015-08-17
Hiding follow button after unfollowing private library

adamjgrant:


adamjgrant:
@cvializ @andrewconner

cvializ:
:+1:

PR #16851, opened 2015-08-17
Arrange page elements better on mobile screens

cvializ:
@adamjgrant @andrewconner 

| Before | After |
| --- | --- |
| screen shot 2015-08-17 at 10 37 44 am | screen shot 2015-08-17 at 10 38 25 am |
| screen shot 2015-08-17 at 10 37 50 am | screen shot 2015-08-17 at 10 38 31 am |


adamjgrant:
Nitpicking, but could we add just a liiiitle margin around the modal, like 5-10px on the left and right?

Otherwise, looks great!

cvializ:
Nitpicking is welcome! I'll get that in

PR #16852, opened 2015-08-17
Org nonmember view

camhashemi:
@ryanpbrewster @leogrim add org-creators to the org experiment

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
@@ -41,6 +41,7 @@ class OrganizationCommanderImpl @Inject() (
     orgMembershipCommander: OrganizationMembershipCommander,
     organizationMembershipCandidateRepo: OrganizationMembershipCandidateRepo,
     orgInviteRepo: OrganizationInviteRepo,
+    userExperimentRepo: UserExperimentRepo,
     organizationAvatarCommander: OrganizationAvatarCommander,
     userRepo: UserRepo,
     keepRepo: KeepRepo,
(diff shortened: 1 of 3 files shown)


ryanpbrewster:
I don't understand why this is necessary. If they don't have the experiment how would they create a library? Is this strictly via the admin tool?

kifi-backend/modules/shoebox/marketing
```diff
@@ -1 +1 @@
-Subproject commit b5a3231951e305f9aa904506eeba3cb72d959a25
+Subproject commit d3dc4581840dbc8234e5ea2cd6da376fd04d1cd7
```
andrewconner, line 1: Intentional?

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
```diff
@@ -215,6 +216,7 @@ class OrganizationCommanderImpl @Inject() (
               throw new Exception(OrganizationFail.HANDLE_UNAVAILABLE.message)
             }
             orgMembershipRepo.save(org.newMembership(userId = request.requesterId, role = OrganizationRole.ADMIN))
+            userExperimentRepo.save(UserExperiment(userId = request.requesterId, experimentType = UserExperimentType.ORGANIZATION))
```
andrewconner, line 219: This will cause a constraint violation if they're already in, right?

camhashemi:
@ryanpbrewster yes, the admin. even though we add the new owner to the exp using `transferOwner`, in dev mode you have to create an org directly. it just seems better to add the owner to it.

anyways, this should be all be irrelevant once the exp is removed.

kifi-backend/modules/shoebox/marketing
camhashemi, line 1: more like seemingly unavoidable, it's the current hash

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
ryanpbrewster, line 219: If the experiment is active it will. I think we should add a method to `UserExperimentCommander`, something like `ensureUserHasExperiment`, that does the correct thing.

camhashemi, line 219: I didn't know that, :+1: Ryan. will update

PR #16855, opened 2015-08-17
Add OrganizationInviteInfo to org responses

camhashemi:
This adds information regarding the viewer's invite, similar to libraries. 

@ryanpbrewster @andrewconner @leogrim

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
@@ -40,7 +40,7 @@ trait OrganizationInviteCommander {
   def getInviteByOrganizationIdAndAuthToken(orgId: Id[Organization], authToken: String): Option[OrganizationInvite]
   def suggestMembers(userId: Id[User], orgId: Id[Organization], query: Option[String], limit: Int): Future[Seq[MaybeOrganizationMember]]
   def isAuthValid(orgId: Id[Organization], authToken: String): Boolean
-  def getViewerInviteInfo(orgId: Id[Organization], userIdOrAuth: Either[Id[User], String]): Option[OrganizationInviteInfo]
+  def getViewerInviteInfo(orgId: Id[Organization], viewerIdOpt: Option[Id[User]], authTokenOpt: Option[String]): Option[OrganizationInviteInfo]
 }
 
 @Singleton
@@ -469,14 +469,16 @@ class OrganizationInviteCommanderImpl @Inject() (db: Database,
     db.readOnlyReplica { implicit session => organizationInviteRepo.getByOrgIdAndAuthToken(orgId, authToken) }.isDefined
   }
 
-  def getViewerInviteInfo(orgId: Id[Organization], userIdOrAuth: Either[Id[User], String]): Option[OrganizationInviteInfo] = {
-    userIdOrAuth.fold(
-      { userId => getLastSentByOrganizationIdAndInviteeId(orgId, userId) },
-      { authToken => getInviteByOrganizationIdAndAuthToken(orgId, authToken) }
-    ).map {
-        invite: OrganizationInvite =>
-          val inviter = db.readOnlyReplica { implicit session => basicUserRepo.load(invite.inviterId) }
-          OrganizationInviteInfo.fromInvite(invite, inviter)
-      }
+  def getViewerInviteInfo(orgId: Id[Organization], viewerIdOpt: Option[Id[User]], authTokenOpt: Option[String]): Option[OrganizationInviteInfo] = {
+    val inviteOpt = (viewerIdOpt, authTokenOpt) match {
+      case (Some(viewerId), None) => getLastSentByOrganizationIdAndInviteeId(orgId, viewerId)
+      case (None, Some(authToken)) => getInviteByOrganizationIdAndAuthToken(orgId, authToken)
+      case (Some(viewerId), Some(authToken)) => getLastSentByOrganizationIdAndInviteeId(orgId, viewerId).orElse(getInviteByOrganizationIdAndAuthToken(orgId, authToken))
+      case _ => None
+    }
+    inviteOpt.map { invite =>
+      val inviter = db.readOnlyReplica { implicit session => basicUserRepo.load(invite.inviterId) }
+      OrganizationInviteInfo.fromInvite(invite, inviter)
+    }
   }
 }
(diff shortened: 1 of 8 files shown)


kifi-backend/modules/shoebox/app/com/keepit/controllers/website/OrganizationController.scala
```diff
@@ -88,7 +88,12 @@ class OrganizationController @Inject() (
   }
 
   def getOrganization(pubId: PublicId[Organization], authTokenOpt: Option[String] = None) = OrganizationAction(pubId, authTokenOpt, OrganizationPermission.VIEW_ORGANIZATION) { request =>
-    val organizationView = orgCommander.getOrganizationView(request.orgId, request.request.userIdOpt)
+    val viewerInfoOpt: Option[Either[Id[User], String]] = (request.request.userIdOpt, authTokenOpt) match {
+      case (Some(userId), _) => Some(Left(userId))
+      case (_, Some(authToken)) => Some(Right(authToken)) // non-user with valid invite
+      case (None, None) => None // non-user without a valid invite
+    }
```
camhashemi: a questionable pattern begins here. we need either their `userId` or `authToken`, but not both. The alternative seems to be passing through two options to every proceeding function, but then there's a match block at each decision.

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationInfo.scala
```diff
@@ -35,26 +36,40 @@ object OrganizationInfo {
   )(unlift(OrganizationInfo.unapply))
 }
 
-case class MembershipInfo(
-  isInvited: Boolean,
+case class OrganizationMembershipInfo(
+  isInvited: Boolean, // deprecated, kill and use invite.isDefined || invite != null instead
+  invite: Option[OrganizationInviteInfo],
```
camhashemi, line 41: preserving the current behavior in order to debug/develop the front-end code separately. need to inform clients that the API is changing (namely mobile, I'll work on changing front-end).

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
```diff
@@ -460,4 +468,15 @@ class OrganizationInviteCommanderImpl @Inject() (db: Database,
...
+
+  def getViewerInviteInfo(orgId: Id[Organization], userIdOrAuth: Either[Id[User], String]): Option[OrganizationInviteInfo] = {
+    userIdOrAuth.fold(
+      { userId => getLastSentByOrganizationIdAndInviteeId(orgId, userId) },
+      { authToken => getInviteByOrganizationIdAndAuthToken(orgId, authToken) }
+    ).map {
+        invite: OrganizationInvite =>
+          val inviter = db.readOnlyReplica { implicit session => basicUserRepo.load(invite.inviterId) }
+          OrganizationInviteInfo.fromInvite(invite, inviter)
+      }
```
andrewconner: Spacing seems a bit weird here?

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/OrganizationController.scala
andrewconner: That's fine. Do we have a type for this? May help alleviate confusion anyone has when they see it. `String` won't really help, and since the variable's name isn't related either, a type may be the way to go.

andrewconner:
Cool, looks good to me. :+1:

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
camhashemi: right, think I fixed this but didn't push. will do.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/OrganizationController.scala
camhashemi: Do you mean to type `Option[Either[A,B]]`? We could call it `Neither` (left, right, or none) :)

I looked through `common.common` and couldn't find anything, not sure where it would be.

andrewconner: Sorry, I meant typing `authToken`, I totally wasn't clear.

After some more thought: What we did with libraries is just have two parameters, people pass in what they have, and it decides the access. ie, it's not a huge deal to pass in both. In fact, you probably want to pass in both, because what if someone's logged in as a user, but their email was invited (thus, they have a valid authToken)?

camhashemi: I think the way public orgs are currently spec'd, there would be no difference w/ or w/o an auth token if you're logged in. But for private orgs (in the future), you could be logged in but unable to see the org without an auth token. Got it, thanks for putting some thought into that.

camhashemi:
@andrewconner updated.

camhashemi:
Typing AuthToken is a pretty heavy refactor, so saving it for later.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
```diff
@@ -460,4 +468,17 @@ class OrganizationInviteCommanderImpl @Inject() (db: Database,
...
     db.readOnlyReplica { implicit session => organizationInviteRepo.getByOrgIdAndAuthToken(orgId, authToken) }.isDefined
   }
+
+  def getViewerInviteInfo(orgId: Id[Organization], viewerIdOpt: Option[Id[User]], authTokenOpt: Option[String]): Option[OrganizationInviteInfo] = {
+    val inviteOpt = (viewerIdOpt, authTokenOpt) match {
+      case (Some(viewerId), None) => getLastSentByOrganizationIdAndInviteeId(orgId, viewerId)
+      case (None, Some(authToken)) => getInviteByOrganizationIdAndAuthToken(orgId, authToken)
+      case (Some(viewerId), Some(authToken)) => getLastSentByOrganizationIdAndInviteeId(orgId, viewerId).orElse(getInviteByOrganizationIdAndAuthToken(orgId, authToken))
+      case _ => None
+    }
```
ryanpbrewster, line 478: Can we do

```
val userInviteOpt = viewerIdOpt.flatMap(getLastSentByOrganizationIdAndInviteeId(orgId, _))
val authTokenInviteOpt = authTokenOpt.flatMap(getLastSentByOrganizationIdAndAuthToken(orgId, _))
val inviteOpt = userInviteOpt.orElse(authTokenInviteOpt)
```

camhashemi, line 478: nice, thanks!

PR #16856, opened 2015-08-17
Rearrange and style membership actions as buttons instead of as links.

cvializ:
@adamjgrant

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/orgProfile/orgProfileMember.js
+++ b/kifi-backend/modules/shoebox/angular/src/orgProfile/orgProfileMember.js
@@ -20,9 +20,9 @@ angular.module('kifi')
       );
     }
 
-    function _resentInvite() {
+    function _getResentInvite() {
       var $scope = this;
-      return $scope._resentInvite;
+      return $scope.model.resentInvite;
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/angular/src/orgProfile/orgProfileMember.js
```diff
@@ -93,8 +93,7 @@ angular.module('kifi')
       var $scope = this;
       $scope.$emit('inviteMember', $scope.member, function (promise) {
         promise.then(function () {
-          // TODO: WHY $scope.$parent??? Is it because ng-if creates a new scope?
-          $scope.$parent._resentInvite = true;
+          $scope.model.resentInvite = true;
```
adamjgrant, line 96: Nice.

adamjgrant:
:+1:

kifi-backend/modules/shoebox/angular/src/orgProfile/orgProfileMember.js
cvializ, line 96: : P Angular's prototypal inheritance gotchas

PR #16857, opened 2015-08-17
add "saveIfNotExists" behavior to UserExperimentRepo

camhashemi:
no changes to logic, just adding this for future use

@ryanpbrewster @andrewconner


kifi-backend/modules/shoebox/app/com/keepit/model/UserExperimentRepo.scala
```diff
@@ -94,4 +95,10 @@ class UserExperimentRepoImpl @Inject() (
         (UserExperimentType(name), count)
     }
   }
+
+  def ensureUserHasExperiment(userId: Id[User], experimentType: UserExperimentType)(implicit session: RWSession): Unit = {
```
andrewconner, line 99: Repos should return something.

andrewconner:
Check out `UserExperimentCommander.addExperimentForUser`. Does what you want, and doesn't break dynamic experiments.

camhashemi:
I don't see that method there. Do you mean `addDynamicExperiments`?

andrewconner:
Sorry, it's in shoebox's implementation. `LocalUserExperimentCommander`

camhashemi:
nice, wish it was used more! thank you.

PR #16860, opened 2015-08-18
KeepToLibrary sets addedAt to match Keep.keptAt

ryanpbrewster:
This is so that the keep ordering within a library matches existing code. When
we are ready to break from current behavior we should set it back to use
clock.now or something.

@leogrim

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepToLibraryCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepToLibraryCommander.scala
@@ -42,6 +42,7 @@ class KeepToLibraryCommanderImpl @Inject() (
           keepId = keep.id.get,
           libraryId = library.id.get,
           addedBy = addedBy,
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepToLibraryCommander.scala
```diff
@@ -42,6 +42,7 @@ class KeepToLibraryCommanderImpl @Inject() (
           keepId = keep.id.get,
           libraryId = library.id.get,
           addedBy = addedBy,
+          addedAt = keep.keptAt, // TODO(ryan): take this out once we're ready to have keeps in multiple libraries
```
leogrim, line 45: No this should be `clock.now()` today. 

The sorting logic though should look at `Keep.keptAt`

leogrim, line 45: (even though the _existing data_ should be grandfathered with `addedAt = keep.keptAt`

PR #16865, opened 2015-08-18
Lib modify fix

ryanpbrewster:
@leogrim

### DIFF
--- a/kifi-backend/modules/shoebox/test/com/keepit/commanders/LibraryCommanderTest.scala
+++ b/kifi-backend/modules/shoebox/test/com/keepit/commanders/LibraryCommanderTest.scala
@@ -533,7 +533,42 @@ class LibraryCommanderTest extends TestKitSupport with SpecificationLike with Sh
       }
       "correctly change all the denormalized library fields on keeps/ktls" in {
         withDb(modules: _*) { implicit injector =>
-          
+          val libraryCommander = inject[LibraryCommander]
+          val (user, org, lib1) = db.readWrite { implicit session =>
+            val user = UserFactory.user().saved
+            val org = OrganizationFactory.organization().withOwner(user).saved
+            val lib = LibraryFactory.library().withOwner(user).published().saved
+            KeepFactory.keeps(100).map(_.withLibrary(lib).withUser(user)).saved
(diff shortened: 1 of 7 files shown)


kifi-backend/modules/shoebox/test/com/keepit/commanders/LibraryCommanderTest.scala
```diff
@@ -531,6 +531,46 @@ class LibraryCommanderTest extends TestKitSupport with SpecificationLike with Sh
...
+          val (user, org, lib1) = db.readWrite { implicit session =>
+            val user = UserFactory.user().saved
+            val org = OrganizationFactory.organization().withOwner(user).saved
+            val lib = LibraryFactory.library().withOwner(user).published().saved
+            KeepFactory.keeps(100).map(_.withLibrary(lib).withUser(user)).saved
+            (user, org, lib)
+          }
+
+          val lib2 = libraryCommander.modifyLibrary(lib1.id.get, user.id.get, LibraryModifyRequest(visibility = Some(LibraryVisibility.SECRET))).right.get
+          Thread.sleep(1000)
```
ryanpbrewster: I grossed myself out by writing this, but I don't know of another way. The modifications are done in a `Future` that is completely internal to `LibraryCommander`.

andrewconner: Why not return the `Future`? Otherwise, we actually have a way around this. Make sure the commander is using an injected `ExecutionContext` (instead of an imported one), and there's a testing method that flushes it (so basically waits until all `Futures` finish). Not on a computer with the code, but you may find it searching for `flush` or something in tests.

andrewconner: And... thanks for feeling bad. Means you have great intuition :)

andrewconner: https://github.com/kifi/keepit/blob/master/kifi-backend/modules/common/app/com/keepit/common/concurrent/WatchableExecutionContext.scala `.drain()`

PR #16875, opened 2015-08-18
Restoring quick keeper and removing duplicate 'add to library' intera…

adamjgrant:
…ction


adamjgrant:
@andrewconner @cvializ

cvializ:
:+1:

PR #16876, opened 2015-08-18
Ktl join order 3

ryanpbrewster:
@leogrim @stkem 

Please, please, please do not make me do a R5 (revert revert revert revert revert).

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepToLibraryCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepToLibraryCommander.scala
@@ -6,6 +6,7 @@ import com.keepit.common.db.slick.DBSession.{ RSession, RWSession }
 import com.keepit.common.db.slick.Database
 import com.keepit.common.healthcheck.AirbrakeNotifier
 import com.keepit.common.logging.Logging
+import com.keepit.common.time._
 import com.keepit.model._
 
 import scala.util.{ Success, Failure, Try }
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/shoebox/app/com/keepit/model/KeepToLibraryRepo.scala
```diff
@@ -47,6 +49,34 @@ class KeepToLibraryRepoImpl @Inject() (
...
+      seq = r.<<[SequenceNumber[Keep]],
+      libraryId = r.<<[Option[Id[Library]]],
+      visibility = r.<<[LibraryVisibility],
+      keptAt = r.<<[DateTime],
+      sourceAttributionId = r.<<[Option[Id[KeepSourceAttribution]]],
+      note = r.<<[Option[String]],
+      originalKeeperId = r.<<[Option[Id[User]]],
+      organizationId = r.<<[Option[Id[Organization]]]
+    )
+  }
```
ryanpbrewster, line 78: Can we put these somewhere centralized so we can reuse them?

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepToLibraryCommander.scala
```diff
@@ -44,7 +46,7 @@ class KeepToLibraryCommanderImpl @Inject() (
           keepId = keep.id.get,
           libraryId = library.id.get,
           addedBy = addedBy,
-          addedAt = keep.keptAt, // TODO(ryan): take this out once we're ready to have keeps in multiple libraries
+          addedAt = clock.now,
```
stkem, line 49: Shouldn't this still be something similar to `keptAt` since we want to preserve true timestamps for things like imports? (Really only applies to your the first keep of something)

ryanpbrewster, line 49: This is why we order keeps by `keptAt`. However, `ktl.addedAt` should reflect the true time at which a keep was added to a library.

PR #16877, opened 2015-08-18
Let the box expand with its content.

cvializ:
@andrewconner @adamjgrant 

Fixes http://acon.s3.amazonaws.com/GdhGF.png


adamjgrant:
Good catch :+1:

cvializ:
Thanks again, @andrewconner

PR #16878, opened 2015-08-18
No longer mirror every UserConnection with a FortyTwo SocialUserConnection

leogrim:
@andrewconner


andrewconner:
murica_3

leogrim:
Oh man, Clinton was awesome.

PR #16880, opened 2015-08-18
I refuse to R5

ryanpbrewster:
The issue was with `k.*` in the hand-written SQL query. Those things are super
unreliable and I don't like them. Should've gone with the sketchy Slick thing.

@leogrim

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/model/KeepToLibraryRepo.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/model/KeepToLibraryRepo.scala
@@ -21,7 +21,7 @@ trait KeepToLibraryRepo extends Repo[KeepToLibrary] {
   def getAllByLibraryId(libraryId: Id[Library], excludeStates: Set[State[KeepToLibrary]] = Set(KeepToLibraryStates.INACTIVE))(implicit session: RSession): Seq[KeepToLibrary]
   def getAllByLibraryIds(libraryIds: Set[Id[Library]], excludeStates: Set[State[KeepToLibrary]] = Set(KeepToLibraryStates.INACTIVE))(implicit session: RSession): Map[Id[Library], Seq[KeepToLibrary]]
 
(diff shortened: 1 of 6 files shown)


kifi-backend/modules/shoebox/app/com/keepit/model/KeepToLibraryRepo.scala
```diff
@@ -135,15 +107,15 @@ class KeepToLibraryRepoImpl @Inject() (
     getByLibraryIdsHelper(libraryIds, excludeStates).list.groupBy(_.libraryId)
   }
 
-  def getByLibraryIdSorted(libraryId: Id[Library], offset: Offset, limit: Limit)(implicit session: RSession): Seq[Keep] = {
+  def getByLibraryIdSorted(libraryId: Id[Library], offset: Offset, limit: Limit)(implicit session: RSession): Seq[Id[Keep]] = {
     // N.B.: Slick is NOT GOOD at joins (as of 2015-08-18). This query is hand-written for a reason.
     import com.keepit.common.db.slick.StaticQueryFixed.interpolation
-    val q = sql"""select k.* from
```
andrewconner, line 141: Yeah you can't do `*` here if your case class field order is different than the db field order.

kifi-backend/modules/shoebox/app/com/keepit/model/KeepToLibraryRepo.scala
```diff
@@ -135,15 +107,15 @@ class KeepToLibraryRepoImpl @Inject() (
     getByLibraryIdsHelper(libraryIds, excludeStates).list.groupBy(_.libraryId)
   }
 
-  def getByLibraryIdSorted(libraryId: Id[Library], offset: Offset, limit: Limit)(implicit session: RSession): Seq[Keep] = {
+  def getByLibraryIdSorted(libraryId: Id[Library], offset: Offset, limit: Limit)(implicit session: RSession): Seq[Id[Keep]] = {
```
andrewconner, line 110: Previously, we've put joining repo methods in the repo of the _return_ type (so this would be KeepRepo)

PR #16883, opened 2015-08-18
Remove unused authentication code

leogrim:
@andrewconner more code to be removed once I make sure this doesn't break anything in Prod.
(I have looked at the past month logs, didn't see any sign of the code I've removed, so I'm confident we'll be ok)


andrewconner:
Wahoo! If you need help QAing let me know.

leogrim:
I have no clue what the token stuff does, but an in-memory map and it not showing up in the logs let me think it's nothing useful...

The `findByEmail` stuff is just wrong.

PR #16885, opened 2015-08-19
kill the org experiment

camhashemi:
@ryanpbrewster @leogrim @andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/OrganizationController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/OrganizationController.scala
@@ -29,20 +29,17 @@ class OrganizationController @Inject() (
 
   def createOrganization = UserAction(parse.tolerantJson) { request =>
     implicit val context = heimdalContextBuilder.withRequestInfoAndSource(request, KeepSource.site).build
-    if (!request.experiments.contains(UserExperimentType.ORGANIZATION)) BadRequest(Json.obj("error" -> "insufficient_permissions"))
-    else {
-      request.body.validate[OrganizationInitialValues](OrganizationInitialValues.website) match {
-        case _: JsError => OrganizationFail.BAD_PARAMETERS.asErrorResponse
-        case JsSuccess(initialValues, _) =>
-          val createRequest = OrganizationCreateRequest(requesterId = request.userId, initialValues)
-          orgCommander.createOrganization(createRequest) match {
-            case Left(failure) =>
-              failure.asErrorResponse
-            case Right(response) =>
-              val organizationView = orgCommander.getOrganizationView(response.newOrg.id.get, request.userIdOpt, authTokenOpt = None)
-              Ok(Json.toJson(organizationView))
-          }
-      }
+    request.body.validate[OrganizationInitialValues](OrganizationInitialValues.website) match {
+      case _: JsError => OrganizationFail.BAD_PARAMETERS.asErrorResponse
+      case JsSuccess(initialValues, _) =>
+        val createRequest = OrganizationCreateRequest(requesterId = request.userId, initialValues)
+        orgCommander.createOrganization(createRequest) match {
+          case Left(failure) =>
+            failure.asErrorResponse
+          case Right(response) =>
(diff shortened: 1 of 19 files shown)


andrewconner:
Wahoo! nice. Will this affect clients in any way, or they don't use it?

camhashemi:
On web, the only experiment I see being used is `ADMIN`, which makes sense. @jaredpetker are you using the `organization` experiment for anything?

kifi-backend/modules/shoebox/app/com/keepit/shoebox/controllers/OrganizationAccessActions.scala
```diff
@@ -37,37 +39,18 @@ trait OrganizationAccessActions {
...
-            case _ => Future.successful(NotFound) // we can remove this after we remove the OrgExperiment guard
+          val memberPermissions = orgMembershipCommander.getPermissions(orgId, maybeRequest.userIdOpt)
+          val requiredPermissionsSet = requiredPermissions.toSet + OrganizationPermission.VIEW_ORGANIZATION
+          if (requiredPermissionsSet.subsetOf(memberPermissions)) {
+            block(OrganizationRequest(maybeRequest, orgId, authTokenOpt, memberPermissions))
+          } else {
+            Future.successful(OrganizationFail.INSUFFICIENT_PERMISSIONS.asErrorResponse)
           }
-        case _ =>
-          Future.successful(OrganizationFail.INVALID_PUBLIC_ID.asErrorResponse)
```
camhashemi, line 70: I made these two functions almost identical so that I can DRY them up, but I'm having trouble maintaining their signatures in the process.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/OrganizationController.scala
```diff
@@ -96,15 +93,11 @@ class OrganizationController @Inject() (
...
 
-  // TODO(ryan): when organizations are no longer hidden behind an experiment, change this to a MaybeUserAction
-  def getOrganizationsForUser(extId: ExternalId[User]) = UserAction { request =>
-    if (!request.experiments.contains(UserExperimentType.ORGANIZATION)) BadRequest(Json.obj("error" -> "insufficient_permissions"))
-    else {
-      val user = userCommander.getByExternalIds(Seq(extId)).values.head
-      val visibleOrgs = orgMembershipCommander.getVisibleOrganizationsForUser(user.id.get, viewerIdOpt = request.userIdOpt)
-      val orgCards = orgCommander.getOrganizationCards(visibleOrgs, request.userIdOpt).values.toSeq
+  def getOrganizationsForUser(extId: ExternalId[User]) = MaybeUserAction { request =>
+    val user = userCommander.getByExternalIds(Seq(extId)).values.head
```
leogrim, line 97: A better way to write this is: `Seq(user) = userCommander.getByExternalIds(Seq(extId))`

(pattern matching works with `val` assigments)

kifi-backend/modules/shoebox/app/com/keepit/shoebox/controllers/OrganizationAccessActions.scala
leogrim, line 70: If you want to implement `OrganizationUserAction` using `OrganizationAction`, I think you need `OrganizationUserRequest` to extend `OrganizationRequest` and `OrganizationAction` take a block `PartialFunction[OrganizationRequest, Result]` instead of a `OrganizationRequest => Result`. 

My mind is not fresh enough to decide whether having a `PartialFunction` in there is fine / worth it though, cc @andrewconner @ryanpbrewster

leogrim:
Nice job!

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/OrganizationController.scala
camhashemi, line 97: very nice, very scala

leogrim, line 97: Note that it's actually stronger than `.head` because it enforces that the `Seq` has _one and only one_ element, while `.head` just enforces that there's _at least one_.

PR #16895, opened 2015-08-19
Implementing loading spinner as directive

adamjgrant:
Here's my current approach:

The loading spinner is a directive that will hide itself when a truthy variable is passed into the `hide` attribute

```
<div class="kf-some-piece-of-the-app">
  <div kf-loading-spinner hide="loaded"></div>
</div>
```

I thought it would be best to put the toggling logic of something like `loaded` in a service file, such as `orgProfileService`

```
// OrgProfileService.js

getOrgLibraries: function(orgId, page, size) {
  return net.getOrgLibraries(orgId, {...})
    .then(function(response) {
      $rootScope.$emit('loaded');
      return response.data;
    })
}

// OrgProfileLibrariesCtrl.js

$rootScope.$on('loaded', function() {
  $scope.loaded = true;
});
```

It puts the responsibility of toggling in the service but it means if there are two things on the page (remember, this is `$rootScope`) that are loaded, the first one to `$emit` "loaded" will cause all spinners to disappear.

So I yanked that out and kept it simple.  Now the service file isn't changed and each controller that uses a loading spinner is single-responsibility.

```
// OrgProfileLibrariesCtrl.js

orgProfileService.getOrgLibraries(...)
.then(function(libData) {
  $scope.loaded = true;
  return libData.libraries
})
```

What do you think @cvializ @andrewconner ?


kifi-backend/modules/shoebox/angular/src/orgProfile/orgProfileLibraries.tpl.html
```diff
@@ -4,6 +4,7 @@
       <span class="kf-upl-create-card-add-icon svg-create-library"></span>
       <span class="kf-upl-create-card-text">Create a Library</span>
     </button>
+    <div kf-loading-spinner hide="loaded"></div>
```
cvializ, line 7: why not just put `ng-hide` on the `kf-loading-spinner` itself?

adamjgrant, line 7: Mostly for philosophical reasons in keeping it single-responsibility. I was thinking I'd let the directive decide what the true/false state would do. If we were to implement some other logic to the toggling, we'd have to change the implementation on the instances rather than the directive.

But idk, maybe it's YAGNI.

adamjgrant, line 7: An example might be (not saying that we do this) that when loaded is false, we execute an exit animation and delay an actual ng-hide event until 500ms later.

cvializ, line 7: I agree with you philosophically. Angular has a way of handling animations with `ng-hide` https://docs.angularjs.org/api/ng/directive/ngHide#a-note-about-animations-with-nghide- but it could be a headache to work with, so I'm on the fence

PR #16897, opened 2015-08-19
First draft of KeepToUser

ryanpbrewster:
@leogrim @andrewconner @camhashemi @gratimax

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/model/KeepToUser.scala
+++ b/kifi-backend/modules/common/app/com/keepit/model/KeepToUser.scala
@@ -1,0 +1,50 @@
+package com.keepit.model
+
+import com.keepit.common.db._
+import com.keepit.common.time._
+import org.joda.time.DateTime
+
+case class KeepToUser(
+  id: Option[Id[KeepToUser]] = None,
+  createdAt: DateTime = currentDateTime,
+  updatedAt: DateTime = currentDateTime,
+  state: State[KeepToUser] = KeepToUserStates.ACTIVE,
+  keepId: Id[Keep],
+  userId: Id[User],
+  addedAt: DateTime = currentDateTime,
+  addedBy: Id[User],
+  // A denormalized field from Keep
+  uriId: Id[NormalizedURI])
+    extends ModelWithState[KeepToUser] {
+
+  def withId(id: Id[KeepToUser]): KeepToUser = this.copy(id = Some(id))
+  def withUpdateTime(now: DateTime): KeepToUser = this.copy(updatedAt = now)
+  def withState(newState: State[KeepToUser]): KeepToUser = this.copy(state = newState)
+  def withUriId(newUriId: Id[NormalizedURI]) = this.copy(uriId = newUriId)
+
+  def isActive = state == KeepToUserStates.ACTIVE
+  def isInactive = state == KeepToUserStates.INACTIVE
+
+  def sanitizeForDelete = this.withState(KeepToUserStates.INACTIVE)
+}
+
+object KeepToUser {
+  def applyFromDbRow(id: Option[Id[KeepToUser]], createdAt: DateTime, updatedAt: DateTime, state: State[KeepToUser],
+    keepId: Id[Keep], userId: Id[User], addedAt: DateTime, addedBy: Id[User],
+    uriId: Id[NormalizedURI]): KeepToUser = {
+    KeepToUser(
+      id, createdAt, updatedAt, state,
+      keepId, userId, addedAt, addedBy,
+      uriId)
+  }
+
+  def unapplyToDbRow(ktu: KeepToUser) = {
+    Some(
+      (ktu.id, ktu.createdAt, ktu.updatedAt, ktu.state,
+        ktu.keepId, ktu.userId, ktu.addedAt, ktu.addedBy,
+        ktu.uriId)
+    )
+  }
+}
+
+object KeepToUserStates extends States[KeepToUser]
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepToUserCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepToUserCommander.scala
@@ -1,0 +1,75 @@
+package com.keepit.commanders
+
+import com.google.inject.{ ImplementedBy, Inject, Singleton }
+import com.keepit.common.db.Id
+import com.keepit.common.db.slick.DBSession.{ RSession, RWSession }
+import com.keepit.common.db.slick.Database
+import com.keepit.common.logging.Logging
+import com.keepit.common.time._
+import com.keepit.model._
+
+import scala.util.control.NoStackTrace
+import scala.util.{ Failure, Success, Try }
+
+sealed abstract class KeepToUserFail(val message: String) extends Exception(message) with NoStackTrace
+object KeepToUserFail {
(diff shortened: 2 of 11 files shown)


kifi-backend/modules/shoebox/app/com/keepit/model/KeepToLibraryRepo.scala
```diff
@@ -71,40 +69,34 @@ class KeepToLibraryRepoImpl @Inject() (
   def table(tag: Tag) = new KeepToLibraryTable(tag)
   initTable()
 
-  // ought to use a cache
-  def getByIds(ids: Set[Id[KeepToLibrary]])(implicit session: RSession): Map[Id[KeepToLibrary], KeepToLibrary] = {
-    val q = for (row <- rows if row.id.inSet(ids)) yield row
-    q.list.map(ktl => ktl.id.get -> ktl).toMap
-  }
```
ryanpbrewster, line 78: This method was a bad idea when I wrote it. We should not be dealing with KTLs as first-class entities. They should only be grabbed in reference to keeps/libraries.

kifi-backend/modules/shoebox/app/com/keepit/model/KeepToLibraryRepo.scala
```diff
@@ -11,21 +11,19 @@ import org.joda.time.DateTime
...
 
-  def countByKeepId(keepId: Id[Keep], excludeStates: Set[State[KeepToLibrary]] = Set(KeepToLibraryStates.INACTIVE))(implicit session: RSession): Int
-  def getAllByKeepId(keepId: Id[Keep], excludeStates: Set[State[KeepToLibrary]] = Set(KeepToLibraryStates.INACTIVE))(implicit session: RSession): Seq[KeepToLibrary]
-
-  def getCountByLibraryId(libraryId: Id[Library], excludeStates: Set[State[KeepToLibrary]] = Set(KeepToLibraryStates.INACTIVE))(implicit session: RSession): Int
-  def getCountsByLibraryIds(libraryIds: Set[Id[Library]], excludeStates: Set[State[KeepToLibrary]] = Set(KeepToLibraryStates.INACTIVE))(implicit session: RSession): Map[Id[Library], Int]
-  def getAllByLibraryId(libraryId: Id[Library], excludeStates: Set[State[KeepToLibrary]] = Set(KeepToLibraryStates.INACTIVE))(implicit session: RSession): Seq[KeepToLibrary]
-  def getAllByLibraryIds(libraryIds: Set[Id[Library]], excludeStates: Set[State[KeepToLibrary]] = Set(KeepToLibraryStates.INACTIVE))(implicit session: RSession): Map[Id[Library], Seq[KeepToLibrary]]
```
ryanpbrewster, line 22: Since we are committing to `state \in {active, inactive}`, we can use `Option` instead of `Set`.

kifi-backend/modules/shoebox/test/com/keepit/model/KeepFactoryHelper.scala
```diff
@@ -58,6 +55,14 @@ object KeepFactoryHelper {
...
       )
       injector.getInstance(classOf[KeepToLibraryRepo]).save(ktl)
+      val ktu = KeepToUser(
+        keepId = finalKeep.id.get,
+        userId = finalKeep.userId,
+        addedAt = finalKeep.keptAt,
+        addedBy = finalKeep.userId,
+        uriId = finalKeep.uriId
+      )
+      injector.getInstance(classOf[KeepToUserRepo]).save(ktu)
```
ryanpbrewster, line 65: So I'd rather do something like
`injector.getInstance(classOf[KeepsCommander]).persistKeep(keep)`, which would automatically take care of this linking stuff (and it would mean that once we remove the requirement that a keep always exists in a library I wouldn't have to revisit this logic). However, I got a circular dependency when I tried it. Any ideas?

kifi-backend/modules/common/app/com/keepit/model/KeepToUser.scala
```diff
@@ -0,0 +1,69 @@
...
+  }
+}
+
+object KeepToUserStates extends States[KeepToUser]
+
+sealed abstract class KeepToUserFail(val status: Int, val message: String) extends Exception(message) with NoStackTrace {
+  def asErrorResponse = Status(status)(Json.obj("error" -> message))
+}
+object KeepToUserFail {
+  case object NOT_CONNECTED_TO_USER extends KeepToUserFail(BAD_REQUEST, "keep_not_connected_to_user")
```
maxov: This seems a little scary... when would we reach this kind of inconsistent state?

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepToUserCommander.scala
```diff
@@ -0,0 +1,69 @@
...
+import com.keepit.common.db.slick.Database
+import com.keepit.common.logging.Logging
+import com.keepit.common.time._
+import com.keepit.model._
+
+import scala.util.{ Failure, Success, Try }
+
+@ImplementedBy(classOf[KeepToUserCommanderImpl])
+trait KeepToUserCommander {
+  def internKeepInUser(keep: Keep, userId: Id[User], addedBy: Id[User])(implicit session: RWSession): KeepToUser
```
maxov, line 21: I don't usually see sessions being passed in implicitly to commanders, most I've seen make the session themselves

kifi-backend/modules/common/app/com/keepit/model/KeepToUser.scala
maxov: Ah, I read further down. It's not really a failure on the KeepToUser model though, it's a failure when trying to remove it. I'd say define the failure in the commander, because otherwise it can be quite confusing.

kifi-backend/modules/shoebox/test/com/keepit/model/KeepFactoryHelper.scala
maxov, line 65: What's the stacktrace?

ryanpbrewster, line 65: About 2k lines long. I can show you if you'd like.

kifi-backend/modules/common/app/com/keepit/model/KeepToUser.scala
ryanpbrewster: That's a pretty good idea. I got into the habit of defining the failures inside the model because in the case of Organizations there are a bunch of commanders that all share a set of failures. Here, though, it makes more sense to have the failures close to the place where they happen.

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepToUserCommander.scala
ryanpbrewster, line 21: Yeah, I still haven't figured out a great convention for this, but these are fairly light-weight methods and we actually want all of these things to happen within the same session.

For other examples of this happening, see `KeepToLibraryCommander` and `HandleCommander`.

kifi-backend/modules/shoebox/test/com/keepit/model/KeepFactoryHelper.scala
maxov, line 65: Okay, I think I feel emotionally ready to see the 2k-long stacktrace :stuck_out_tongue:

PR #16898, opened 2015-08-19
track orgId in messages when all participants share orgs

camhashemi:
@leogrim @ryanpbrewster 
involves a little more heavy lifting than I'd like, but this is going to be removed once you can message an entire org or at least when keepscussions is a thing

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/common/routes/Router.scala
+++ b/kifi-backend/modules/common/app/com/keepit/common/routes/Router.scala
@@ -163,7 +163,7 @@ object Shoebox extends Service {
     def getOrganizationInviteViews(orgId: Id[Organization]) = ServiceRoute(GET, "/internal/shoebox/database/getOrganizationInviteViews", Param("orgId", orgId))
     def hasOrganizationMembership(orgId: Id[Organization], userId: Id[User]) = ServiceRoute(GET, "/internal/shoebox/database/hasOrganizationMembership", Param("orgId", orgId), Param("userId", userId))
     def getIngestableOrganizationDomainOwnerships(seqNum: SequenceNumber[OrganizationDomainOwnership], fetchSize: Int) = ServiceRoute(GET, "/internal/shoebox/database/getIngestableOrganizationDomainOwnerships", Param("seqNum", seqNum), Param("fetchSize", fetchSize))
-    def getOrganizationsForUsers(userIds: String) = ServiceRoute(GET, "/internal/shoebox/database/getOrganizationsForUsers", Param("userIds", userIds))
+    def getOrganizationsForUsers() = ServiceRoute(POST, "/internal/shoebox/database/getOrganizationsForUsers")
   }
 }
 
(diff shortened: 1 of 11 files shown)


kifi-backend/modules/common/app/com/keepit/common/routes/Router.scala
```diff
@@ -163,6 +163,7 @@ object Shoebox extends Service {
     def getOrganizationInviteViews(orgId: Id[Organization]) = ServiceRoute(GET, "/internal/shoebox/database/getOrganizationInviteViews", Param("orgId", orgId))
     def hasOrganizationMembership(orgId: Id[Organization], userId: Id[User]) = ServiceRoute(GET, "/internal/shoebox/database/hasOrganizationMembership", Param("orgId", orgId), Param("userId", userId))
     def getIngestableOrganizationDomainOwnerships(seqNum: SequenceNumber[OrganizationDomainOwnership], fetchSize: Int) = ServiceRoute(GET, "/internal/shoebox/database/getIngestableOrganizationDomainOwnerships", Param("seqNum", seqNum), Param("fetchSize", fetchSize))
+    def getOrganizationsForUsers(userIds: String) = ServiceRoute(GET, "/internal/shoebox/database/getOrganizationsForUsers", Param("userIds", userIds))
```
ryanpbrewster: Does this have to be a `String`?

kifi-backend/modules/common/app/com/keepit/shoebox/ShoeboxServiceClient.scala
```diff
@@ -800,4 +801,10 @@ class ShoeboxServiceClientImpl @Inject() (
       }
     }
   }
+
+  def getOrganizationsForUsers(userIds: Set[Id[User]]): Future[Map[Id[User], Set[Id[Organization]]]] = {
+    redundantDBConnectionCheck(userIds)
+    val query = userIds.mkString(",")
```
ryanpbrewster: I see that this is also how we implement `getUsers`. Do you think it's possible to use a `Seq[Id[User]]` instead?

kifi-backend/modules/common/app/com/keepit/common/routes/Router.scala
camhashemi: seems like we don't use implicit conversions in `ParamValue` for Collections. @leogrim any reason why this is?

camhashemi: I guess you'd have to enforce that `T` implements toString in `Seq[T]`.

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationMembershipRepo.scala
```diff
@@ -116,6 +117,10 @@ class OrganizationMembershipRepoImpl @Inject() (
     (for (row <- rows if row.id.inSet(membershipIds) && row.state =!= excludeState.orNull) yield (row.id, row)).list.toMap
   }
 
+  def getAllByUserIds(userIds: Set[Id[User]], excludeState: Option[State[OrganizationMembership]] = Some(OrganizationMembershipStates.INACTIVE))(implicit session: RSession): Map[Id[User], Set[OrganizationMembership]] = {
+    (for (row <- rows if row.userId.inSet(userIds) && row.state =!= excludeState.orNull) yield (row.userId, row)).list.groupBy(_._1).mapValues(userIdAndRow => userIdAndRow.map(_._2).toSet)
```
ryanpbrewster: Nice! I'm not sure if there's a cleaner way to do this. Let me know what you think.

```
rows.filter(row => row.userId.inSet(userIds) && row.state =!= excludeState.orNull).list.groupBy(_.userId).mapValues(_.toSet)
```

camhashemi: Did you get that to compile? I like it if it works.

add: I couldn't get it to compile

ryanpbrewster: I had no issues. It's important to remember the `.list` comes _before_ the `.groupBy`.

camhashemi: got it, thanks

camhashemi: `rows.filter` is nice

kifi-backend/modules/common/app/com/keepit/common/routes/Router.scala
camhashemi: nvm everything: `ParamValue(value: Option[String])`, so it'd be converted there anyways.

camhashemi: talked with Leo, it's better to use a post request in case the query string exceeds the limit for GET requests.

PR #16901, opened 2015-08-19
Update keep data properties for multi-lib keeps.

cvializ:
@adamjgrant @andrewconner resolves https://app.asana.com/0/44702170995796/45685995400471

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/keeps/keeps.js
+++ b/kifi-backend/modules/shoebox/angular/src/keeps/keeps.js
@@ -167,7 +167,7 @@ angular.module('kifi')
           var editor = keepEl.find('.kf-knf-editor');
           if (!editor.length) {
             var noteEl = keepEl.find('.kf-keep-note');
-            $injector.get('keepNoteForm').init(noteEl, keep.note, keep.libraryId, keep.id, function update(noteText) {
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/angular/src/keeps/keeps.js
```diff
@@ -167,7 +167,7 @@ angular.module('kifi')
           var editor = keepEl.find('.kf-knf-editor');
           if (!editor.length) {
             var noteEl = keepEl.find('.kf-keep-note');
-            $injector.get('keepNoteForm').init(noteEl, keep.note, keep.libraryId, keep.id, function update(noteText) {
+            $injector.get('keepNoteForm').init(noteEl, keep.note, keep.keeps[0].libraryId, keep.keeps[0].id, function update(noteText) {
```
cvializ, line 170: I think this change is needed because of changes by @ryanpbrewster to allow multi-library keeps? Should this change be on the front end here, or was it a back end error that caused the new structure to leak where it's not supposed to?

adamjgrant, line 170: I'll ping @ryanpbrewster

cvializ:
Talked with @leogrim and this code won't do the right thing. I'll try another angle

PR #16910, opened 2015-08-20
Hooking up PaidAccount creation/deletion to Organization creation/deletion

stkem:
- minor cleanup

@ryanpbrewster @andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
@@ -15,6 +15,7 @@ import com.keepit.heimdal.HeimdalContext
 import com.keepit.model.OrganizationPermission.{ EDIT_ORGANIZATION, VIEW_ORGANIZATION }
 import com.keepit.model._
 import com.keepit.social.BasicUser
+import com.keepit.payments.{ PlanManagementCommander, PaidPlan }
 
 import scala.util.{ Failure, Success, Try }
 
@@ -53,7 +54,8 @@ class OrganizationCommanderImpl @Inject() (
     orgExperimentRepo: OrganizationExperimentRepo,
     organizationAnalytics: OrganizationAnalytics,
     implicit val publicIdConfig: PublicIdConfiguration,
-    handleCommander: HandleCommander) extends OrganizationCommander with Logging {
+    handleCommander: HandleCommander,
+    planManagementCommander: PlanManagementCommander) extends OrganizationCommander with Logging {
 
   def getOrganizationView(orgId: Id[Organization], viewerIdOpt: Option[Id[User]], authTokenOpt: Option[String]): OrganizationView = {
     db.readOnlyReplica { implicit session =>
@@ -216,6 +218,7 @@ class OrganizationCommanderImpl @Inject() (
               throw new Exception(OrganizationFail.HANDLE_UNAVAILABLE.message)
             }
             orgMembershipRepo.save(org.newMembership(userId = request.requesterId, role = OrganizationRole.ADMIN))
+            planManagementCommander.createAndInitializePaidAccountForOrganization(org.id.get, PaidPlan.DEFAULT, request.requesterId, session).get
             organizationAnalytics.trackOrganizationEvent(org, userRepo.get(request.requesterId), request)
             Right(OrganizationCreateResponse(request, org))
         }
@@ -277,6 +280,7 @@ class OrganizationCommanderImpl @Inject() (
 
           orgRepo.save(org.sanitizeForDelete)
           handleCommander.reclaimAll(org.id.get, overrideProtection = true, overrideLock = true)
+          planManagementCommander.deactivatePaidAccountForOrganziation(org.id.get, session)
           organizationAnalytics.trackOrganizationEvent(org, userRepo.get(request.requesterId), request)
           Right(OrganizationDeleteResponse(request))
         case Some(orgFail) => Left(orgFail)
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
@@ -218,7 +218,7 @@ class OrganizationCommanderImpl @Inject() (
               throw new Exception(OrganizationFail.HANDLE_UNAVAILABLE.message)
             }
             orgMembershipRepo.save(org.newMembership(userId = request.requesterId, role = OrganizationRole.ADMIN))
-            planManagementCommander.createAndInitializePaidAccountForOrganization(org.id.get, PaidPlan.DEFAULT, request.requesterId, session).get
+            planManagementCommander.createAndInitializePaidAccountForOrganization(org.id.get, PaidPlan.DEFAULT, request.requesterId, session) //this should get a .get when thing sare solidified
             organizationAnalytics.trackOrganizationEvent(org, userRepo.get(request.requesterId), request)
             Right(OrganizationCreateResponse(request, org))
         }
--- a/kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEventRepo.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEventRepo.scala
@@ -3,7 +3,7 @@ package com.keepit.payments
 import com.keepit.common.db.slick.{ Repo, DbRepo, DataBaseComponent }
 import com.keepit.common.db.slick.DBSession.{ RWSession, RSession }
 import com.keepit.common.db.{ Id, State }
-import com.keepit.common.time.Clock
+import com.keepit.common.time._
 import com.keepit.model.User
 
 import com.google.inject.{ ImplementedBy, Inject, Singleton }
@@ -24,7 +24,7 @@ trait AccountEventRepo extends Repo[AccountEvent] {
 
   def getEvents(accountId: Id[PaidAccount], limit: Int, onlyRelatedToBillingOpt: Option[Boolean])(implicit session: RSession): Seq[AccountEvent]
 
-  def inactivateAll(accountId: Id[PaidAccount])(implicit session: RWSession): Int
+  def deactivateAll(accountId: Id[PaidAccount])(implicit session: RWSession): Int
 
 }
 
@@ -96,8 +96,8 @@ class AccountEventRepoImpl @Inject() (
     relevantEvents.sortBy(r => (r.eventTime desc, r.id)).take(limit).list
   }
 
-  def inactivateAll(accountId: Id[PaidAccount])(implicit session: RWSession): Int = {
-    (for (row <- rows if row.accountId === accountId) yield row.state).update(AccountEventStates.INACTIVE)
+  def deactivateAll(accountId: Id[PaidAccount])(implicit session: RWSession): Int = {
+    (for (row <- rows if row.accountId === accountId) yield (row.state, row.updatedAt)).update((AccountEventStates.INACTIVE, clock.now))
   }
 
 }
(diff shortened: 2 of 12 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
```diff
@@ -216,6 +218,7 @@ class OrganizationCommanderImpl @Inject() (
               throw new Exception(OrganizationFail.HANDLE_UNAVAILABLE.message)
             }
             orgMembershipRepo.save(org.newMembership(userId = request.requesterId, role = OrganizationRole.ADMIN))
+            planManagementCommander.createAndInitializePaidAccountForOrganization(org.id.get, PaidPlan.DEFAULT, request.requesterId, session).get
```
ryanpbrewster: That is a mouthful. I love it. Super explicit.

kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEventRepo.scala
```diff
@@ -94,4 +96,8 @@ class AccountEventRepoImpl @Inject() (
     relevantEvents.sortBy(r => (r.eventTime desc, r.id)).take(limit).list
   }
 
+  def inactivateAll(accountId: Id[PaidAccount])(implicit session: RWSession): Int = {
+    (for (row <- rows if row.accountId === accountId) yield row.state).update(AccountEventStates.INACTIVE)
```
andrewconner: How come?

kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEventRepo.scala
```diff
@@ -24,6 +24,8 @@ trait AccountEventRepo extends Repo[AccountEvent] {
 
   def getEvents(accountId: Id[PaidAccount], limit: Int, onlyRelatedToBillingOpt: Option[Boolean])(implicit session: RSession): Seq[AccountEvent]
 
+  def inactivateAll(accountId: Id[PaidAccount])(implicit session: RWSession): Int
```
ryanpbrewster: deactivate?

kifi-backend/modules/shoebox/test/com/keepit/commanders/OrganizationCommanderTest.scala
```diff
@@ -160,6 +170,9 @@ class OrganizationCommanderTest extends TestKitSupport with SpecificationLike wi
         val orgRepo = inject[OrganizationRepo]
         val orgCommander = inject[OrganizationCommander]
         val orgMembershipRepo = inject[OrganizationMembershipRepo]
+
+        inject[PlanManagementCommander].createNewPlan(Name[PaidPlan]("Test"), BillingCycle(1), DollarAmount(0))
```
andrewconner, line 174: How about a testing factory? See `UserFactory` (and `PartialUser`). Will make your tests a lot more readable.

kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEventRepo.scala
andrewconner: Update `updatedAt` as well. And perhaps invalidate the cache?

kifi-backend/modules/shoebox/app/com/keepit/payments/PaymentMethodRepo.scala
```diff
@@ -24,7 +24,7 @@ class PaymentMethodRepoImpl @Inject() (
   implicit val stripeTokenColumnType = MappedColumnType.base[StripeToken, String](_.token, StripeToken(_))
 
   type RepoImpl = PaymentMethodTable
-  class PaymentMethodTable(tag: Tag) extends RepoTable[PaymentMethod](db, tag, "paid_plan") {
+  class PaymentMethodTable(tag: Tag) extends RepoTable[PaymentMethod](db, tag, "payment_method") {
```
ryanpbrewster, line 27: O.o

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
```diff
@@ -30,6 +30,8 @@ object PaidPlan extends ModelWithPublicIdCompanion[PaidPlan] {
   protected[this] val publicIdPrefix = "pp"
   protected[this] val publicIdIvSpec = new IvParameterSpec(Array(-81, 48, 82, -97, 110, 73, -46, -55, 43, 73, -107, -90, 89, 21, 116, -101))
 
+  val DEFAULT = Id[PaidPlan](1L)
```
andrewconner, line 33: Eesh, this is tricky. I assume this is the free plan. Will this cause problems in tests? And depend on data being in the db?

kifi-backend/modules/shoebox/app/com/keepit/payments/PaymentMethodRepo.scala
stkem, line 27: indeed :-/

kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEventRepo.scala
stkem: sure

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
stkem, line 33: It's for testing. It'll presumably go away once we have client logic to pick a plan.

kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEventRepo.scala
stkem: Ah yes, good point about updatedAt. There should be no caching used for any of the payments models. The amount of data being moved and the frequency of access is small enough that performance gains form caching would be negligible and the consequences of anything being out of sync/stale could be disastrous.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
andrewconner, line 33: The problems it may cause are no row existing, though. If you did create it, you could just pass it in and not need to hardcode it.

kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEventRepo.scala
andrewconner: Why does this method exist?

ryanpbrewster: Actually, yeah, why do we mark those events as inactive? That seems unexpected to me. I guess we could...maybe it's safer if they're dead.

kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEventRepo.scala
```diff
@@ -94,4 +96,8 @@ class AccountEventRepoImpl @Inject() (
     relevantEvents.sortBy(r => (r.eventTime desc, r.id)).take(limit).list
   }
 
+  def deactivateAll(accountId: Id[PaidAccount])(implicit session: RWSession): Int = {
+    (for (row <- rows if row.accountId === accountId) yield (row.state, row.updatedAt)).update((AccountEventStates.INACTIVE, clock.now))
+  }
```
ryanpbrewster, line 101: Sorry to bring this up again, but is there a reason you don't want to just do a `eventRepo.getAllEvents(accountId).foreach(eventRepo.deactivate)` where `deactivate` does the appropriate thing? Efficiency isn't that big of a concern because account deletion doesn't happen all that often and there shouldn't be too many AccountEvents.

stkem, line 101: It's not about efficiency so much as about robustness. `AccountEvent` is really verbose and even a medium sized organization could easily have hundreds after a relatively short amount of time. Loading them all and then writing them back individually would make for a really long session. Since we don't have select for update that makes it particularly unsafe in a distributed environment.

ryanpbrewster, line 101: Hrm. I ask because I've been using the other convention in a number of places and I'm wondering if that's going to come back and bite us.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
stkem, line 33: It's actually fairly easy to do in tests, since the plan is actually created right there on the spot. However, it has since been decided that users will not pick a plan during org creation but rather be automatically on the free plan, so that needs to be encoded somewhere.
An alternative I can think of is to always grab the first `ACTIVE`, `NORMAL` plan form the db with a zero price. In any case, it's always going to depend on data being in the db (even if the user where to pick a plan). Got any other ideas?

ryanpbrewster, line 33: One solution is to have the plan be optional, and if it's `None` that implies the default plan. That breaks a bit of the existing logic, but it seems much more robust to me.

andrewconner, line 33: What do you mean by “ since the plan is actually created right there on the spot”?

kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEventRepo.scala
stkem, line 101: If it's only a few models it's non issue. If it's going into the hundreds it _might_ become a problem, especially if there is a regularly scheduled process that looks at these rows and modifies them (as will be the case here with payment processing)

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
stkem, line 33: @ryanpbrewster yeah, I've considered that. But the problem of where to get the `Id` of the default plan remains.
@andrewconner I just meant that the test set up their own mock db environment at the start so you know exactly what's in the db.

kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEventRepo.scala
andrewconner, line 101: Is this normal, though? Might be missing something. Seems just for org deletion.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
andrewconner, line 33: What if the free plan is in the evolutions? (A side note: we may have many free plans)

kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEventRepo.scala
stkem, line 101: Yeah, this here is just for org deletion, but if we do the inactivation model by model and the payment processing happens to run while this is going on we could end up in a pretty bad state.

ryanpbrewster, line 101: So we could do what we do for library deletion and keeps: kill the account, the kill the events in batches, and make sure that any scheduled processes lookup the account to make sure it's active before they modify anything.

ryanpbrewster, line 101: That sounds like a lot more work, but if we're actually worried about robustness then it's probably at the upper end of that spectrum. Regardless, thanks for showing me something new, that is pretty cool.

andrewconner, line 101: > inactivation model by model and the payment processing happens to run while this is going on we could end up in a pretty bad state

We shouldn't. The first thing deleting an org (or similar) should do is disable the account. Then keep going cleaning up.

If we needed bulletproof assurances here, it's typically accomplished using a table that controls batches (where we can put a lock on an org first, and then do what we need to)

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
stkem, line 33: Going to make this super defensive to start with, i.e. allowing org creation to proceed even if this fails (will airbrake though)

PR #16915, opened 2015-08-20
[WIP] Adding org private icon to library page

adamjgrant:
Waiting for Jen's approval


andrewconner:
Mind moving to do branches on your fork? Would keep this repo a bit cleaner, unless you need it in here.

PR #16916, opened 2015-08-20
Removing settings text

adamjgrant:
Per https://app.asana.com/0/42964066412484/46106592883758


adamjgrant:
@cvializ @andrewconner

cvializ:
:+1:

PR #16918, opened 2015-08-20
Add view/click tracking for orgProfile.members + .libraries

camhashemi:
@andrewconner @adamjgrant @cvializ

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/orgProfile/orgProfileHeader.js
+++ b/kifi-backend/modules/shoebox/angular/src/orgProfile/orgProfileHeader.js
@@ -65,10 +65,19 @@ angular.module('kifi')
           });
       };
 
-      scope.onOrgProfileImageClick = function (event) {
+      scope.onClickTrack = function (event, action) {
+        if (event.which === 1) {
+          if (action === 'clickedAddCoverImage') {
+            angular.element('.kf-oph-pic-file').click();
+          }
+          orgProfileService.trackEvent('user_clicked_page', scope.profile, { action: action });
+        }
+      };
+
+      scope.onOrgProfileImageClick = function (event) { // doesn't look like we're using this anywhere, but if so, should be merged with onClickTrack
         if (event.which === 1) {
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/shoebox/angular/src/orgProfile/orgProfileHeader.js
```diff
@@ -65,10 +65,19 @@ angular.module('kifi')
...
+      scope.onClickTrack = function (event, action) {
+        if (event.which === 1) {
+          if (action === 'clickedAddCoverImage') {
+            angular.element('.kf-oph-pic-file').click();
+          }
+          orgProfileService.trackEvent('user_clicked_page', scope.profile, { action: action });
+        }
+      };
+
+      scope.onOrgProfileImageClick = function (event) { // doesn't look like we're using this anywhere, but if so, should be merged with onClickTrack
```
camhashemi, line 77: see comment.

cvializ, line 77: I think you're right, it isn't used anywhere atm. I think it was going to be used once we had implemented the Org Profile Logo, but it was forgotten.

camhashemi, line 77: so killing it is probably best?

cvializ, line 77: yep, I think so, as long as you've added the tracking functionality with onClickTrack

camhashemi, line 77: right. the only difference when we decide to track "org profile image click" is to pass the action as a parameter, rather than hard-coded inside it.

cvializ, line 77: works for me

cvializ:
:+1:

PR #16921, opened 2015-08-20
[DO NOT MERGE] Keep merging

ryanpbrewster:
not applicable to current logic, but a preview of what will happen in the future

@leogrim

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/model/Keep.scala
+++ b/kifi-backend/modules/common/app/com/keepit/model/Keep.scala
@@ -36,7 +36,7 @@ case class Keep(
     organizationId: Option[Id[Organization]] = None,
     entitiesHash: Option[EntitiesHash] = None) extends ModelWithExternalId[Keep] with ModelWithState[Keep] with ModelWithSeqNumber[Keep] {
 
-  def sanitizeForDelete: Keep = copy(title = None, state = KeepStates.INACTIVE, entitiesHash = Some(KeepConnectionEntities.empty.hash)) // good idea?
+  def sanitizeForDelete: Keep = copy(title = None, state = KeepStates.INACTIVE, isPrimary = false, entitiesHash = Some(KeepConnectionEntities.empty.hash)) // good idea?
 
   def clean(): Keep = copy(title = title.map(_.trimAndRemoveLineBreaks()))
 
--- a/kifi-backend/modules/common/app/com/keepit/model/Keep.scala
+++ b/kifi-backend/modules/common/app/com/keepit/model/Keep.scala
@@ -11,6 +11,8 @@ import play.api.libs.functional.syntax._
 import play.api.libs.json._
 import com.keepit.common.crypto.PublicId
 
+import scala.util.hashing.MurmurHash3
+
 case class Keep(
     id: Option[Id[Keep]] = None,
     createdAt: DateTime = currentDateTime,
@@ -31,9 +33,10 @@ case class Keep(
     sourceAttributionId: Option[Id[KeepSourceAttribution]] = None,
     note: Option[String] = None,
     originalKeeperId: Option[Id[User]] = None,
-    organizationId: Option[Id[Organization]] = None) extends ModelWithExternalId[Keep] with ModelWithState[Keep] with ModelWithSeqNumber[Keep] {
+    organizationId: Option[Id[Organization]] = None,
+    entitiesHash: Option[EntitiesHash]) extends ModelWithExternalId[Keep] with ModelWithState[Keep] with ModelWithSeqNumber[Keep] {
 
-  def sanitizeForDelete: Keep = copy(title = None, state = KeepStates.INACTIVE)
+  def sanitizeForDelete: Keep = copy(title = None, state = KeepStates.INACTIVE, entitiesHash = Some(KeepConnectionEntities.empty.hash)) // good idea?
 
   def clean(): Keep = copy(title = title.map(_.trimAndRemoveLineBreaks()))
 
@@ -68,10 +71,14 @@ case class Keep(
     organizationId = lib.organizationId
   )
 
+  def withEntities(libraries: Set[Id[Library]], users: Set[Id[User]]): Keep = this.copy(entitiesHash = Some(KeepConnectionEntities(libraries, users).hash))
+
   def isActive: Boolean = state == KeepStates.ACTIVE && isPrimary // isPrimary will be removed shortly
   def isInactive: Boolean = state == KeepStates.INACTIVE
 
-  def canBeMergedInto(other: Keep): Boolean = true
+  def canBeMergedInto(other: Keep): Boolean = {
+    entitiesHash == other.entitiesHash && keptAt >= other.keptAt
+  }
 }
 
 object Keep {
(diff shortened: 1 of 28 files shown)


ryanpbrewster:
@andrewconner mind taking a look?

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepCommander.scala
```diff
@@ -774,6 +777,37 @@ class KeepsCommanderImpl @Inject() (
...
     keepRepo.save(keep.withOwner(newOwnerId))
   }
+
+  private def getKeepsByLibrariesAndUsers(libIds: Set[Id[Library]], userIds: Set[Id[User]], uriId: Id[NormalizedURI])(implicit session: RSession): Set[Id[Keep]] = {
+    val seedCandidates = if (libIds.nonEmpty) {
+      ktlRepo.getAllByLibraryAndUri(libIds.head, uriId).map(_.keepId)
+    } else if (userIds.nonEmpty) {
+      ktuRepo.getAllByUserAndUri(userIds.head, uriId).map(_.keepId)
+    } else {
+      throw new Exception("keep is not connected to any users or in any libraries")
```
leogrim: Seems a bit late for this exception, maybe just return an empty `Set` or have a requirement earlier?

ryanpbrewster: Yeah, this was from when I had this method take in a `Keep` and find its libIds and userIds directly. This exception should be in the caller.

kifi-backend/modules/common/app/com/keepit/model/Keep.scala
```diff
@@ -71,6 +71,8 @@ case class Keep(
 
   def isActive: Boolean = state == KeepStates.ACTIVE && isPrimary // isPrimary will be removed shortly
   def isInactive: Boolean = state == KeepStates.INACTIVE
+
+  def canBeMergedInto(other: Keep): Boolean = true
```
ryanpbrewster: This is to preserve current logic, where we aggressively merge things with the same URI. Going forward this should check things like `note`, `title`, and `messageThreadId.isDefined`.

kifi-backend/modules/shoebox/app/com/keepit/integrity/UriIntegrityPlugin.scala
```diff
@@ -67,85 +66,9 @@ class UriIntegrityActor @Inject() (
...
-            case None =>
-              keepToCollectionRepo.save(ktc.copy(keepId = newBm.id.get))
-          }
-        }
-        collections
-      case _ =>
-        Set.empty[Id[Collection]]
-    }
-
-    collectionsToUpdate.foreach(collectionRepo.collectionChanged(_, inactivateIfEmpty = true))
```
ryanpbrewster, line 148: I removed this because I didn't want to deal with it. It relies on the fact that we dedup keeps two at a time. At some point I'll need to sit down and figure out how to adapt this code.

kifi-backend/modules/shoebox/app/com/keepit/model/KeepRepo.scala
```diff
@@ -540,7 +536,6 @@ class KeepRepoImpl @Inject() (
 
   def getRecentKeepsFromFollowedLibraries(userId: Id[User], limit: Int, beforeIdOpt: Option[ExternalId[Keep]], afterIdOpt: Option[ExternalId[Keep]])(implicit session: RSession): Seq[Keep] = {
     import com.keepit.common.db.slick.StaticQueryFixed.interpolation
-    import scala.collection.JavaConversions._
```
ryanpbrewster, line 546: Nooooooo, IntelliJ, stop doing this to me.

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepCommander.scala
```diff
@@ -774,6 +777,37 @@ class KeepsCommanderImpl @Inject() (
...
+    }
+
+    val libCandidates = libIds.foldLeft(seedCandidates) { case (cands, libId) => cands intersect ktlRepo.getAllByLibraryAndUri(libId, uriId).map(_.keepId) }
+    val userAndLibCandidates = userIds.foldLeft(libCandidates) { case (cands, userId) => cands intersect ktuRepo.getAllByUserAndUri(userId, uriId).map(_.keepId) }
```
leogrim: I would expect `seedCandidates` to be pretty small, might be better and clearer to check them one by one rather than trying to reduce the `Set`.

ryanpbrewster: Hrm. Something like

```
seedCandidates.filter { keepId =>
    ktlRepo.getAllByKeepId(keepId).map(_.libraryId).toSet == libIds &&    ktuRepo.getAllByKeepId(keepId).map(_.userId).toSet == userIds
}
```

I like that, it's pretty clear (and probably more efficient).

leogrim: Also, doesn't this end up testing inclusion?

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepCommander.scala
```diff
@@ -774,6 +777,37 @@ class KeepsCommanderImpl @Inject() (
...
+    val libIds = ktlRepo.getAllByKeepId(keep.id.get).map(_.libraryId).toSet
+    val userIds = ktuRepo.getAllByKeepId(keep.id.get).map(_.userId).toSet
+    val mergeCandidates = (getKeepsByLibrariesAndUsers(libIds, userIds, newUri.id.get) |> keepRepo.getByIds).values
+
+    val newKeep = keepRepo.save(uriHelpers.improveKeepSafely(newUri, keep.withNormUriId(newUri.id.get)))
+    ktlCommander.syncKeep(newKeep)
+    ktuCommander.syncKeep(newKeep)
+
+    if (mergeCandidates.exists(keep.canBeMergedInto)) {
+      deactivateKeep(keep)
```
leogrim: This should deactivate the newest keep?

ryanpbrewster: New logic implies that a keep can only be "merged into" another keep if it is newer. So only newer keeps will ever get deactivated. See `Keep.canBeMergedInto`

kifi-backend/modules/common/app/com/keepit/model/Keep.scala
```diff
@@ -68,8 +71,14 @@ case class Keep(
     organizationId = lib.organizationId
   )
 
+  def withEntities(libraries: Set[Id[Library]], users: Set[Id[User]]): Keep = this.copy(entitiesHash = Some(KeepConnectionEntities(libraries, users).hash))
+
   def isActive: Boolean = state == KeepStates.ACTIVE && isPrimary // isPrimary will be removed shortly
   def isInactive: Boolean = state == KeepStates.INACTIVE
+
+  def canBeMergedInto(other: Keep): Boolean = {
+    entitiesHash == other.entitiesHash && keptAt >= other.keptAt
```
ryanpbrewster: Now that I think about it, this should be strict `>`. It will protect against stupid bugs where we try and merge a keep with itself.

PR #16922, opened 2015-08-20
Fix modal position after focus

cvializ:
Resolves https://app.asana.com/0/42816489233171/46149815816769.

@adamjgrant @andrewconner Not sure if this is the best way to fix it, but it works without changing too much


andrewconner:
Oh that's weird.

cvializ:
Too weird, or "okay weird"?

andrewconner:
It's weird it's needed. Is it animating in or something?

cvializ:
It sets a flag to be true, and then uses an `ng-show="flag"`, and I think when the focus calls in the middle of that, it scrolls the page which causes the problem. I think

andrewconner:
Ah

cvializ:
I think you'll feel somewhat better about it now

PR #16924, opened 2015-08-20
Murdered Keep.kifiInstallation

ryanpbrewster:
@andrewconner @stkem

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/model/Keep.scala
+++ b/kifi-backend/modules/common/app/com/keepit/model/Keep.scala
@@ -25,7 +25,6 @@ case class Keep(
     userId: Id[User],
     state: State[Keep] = KeepStates.ACTIVE,
     source: KeepSource,
-    kifiInstallation: Option[ExternalId[KifiInstallation]] = None,
     seq: SequenceNumber[Keep] = SequenceNumber.ZERO,
     libraryId: Option[Id[Library]],
     keptAt: DateTime = currentDateTime,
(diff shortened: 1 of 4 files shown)


stkem:
Missing evolution to actually drop the column (the `GetResult` will break if it's still there, and you _really_ don't want that one to break in prod). Otherwise lgtm.

ryanpbrewster:
Is that true? I think the only time that would break is if someone had a `select bm.* from bookmark` style query. If they're using the (admittedly horrifying) `#$bookmarkColumnOrder` convention they should be fine.

The other way would break (i.e., if I dropped the column before modifying the Slick table definition) because Slick would go looking for a column that doesn't exist, but IIRC this one is safe.

stkem:
You are right, it's fine with `#$bookmarkColumnOrder`, nonetheless leaving the column is just inviting (possibly production only) errors when someone else uses `bm.*` (even if they shouldn't). Plus it's dead data.

ryanpbrewster:
Yep, definitely going to drop the column (it will lock the table for a bit, though, so I'll do that tonight around midnight). Probably would've forgotten to make an evolution for it, so good catch.

ryanpbrewster:
Also, note that if you use `bm.*` your code _will break_ (in production only, as you pointed out). I ran into that problem just a little earlier this week and broke the site. The problem is that the `bookmarkColumnOrder` does not actually match the db column order (things like `state` and `kept_at` are in the wrong places).

PR #16935, opened 2015-08-21
Kifi goes even more global

andrewconner:
Launching Kifi in several new regions: Ascension Island, Bhutan, Cocos Islands, French Guiana, British Indian Ocean Territory, Saint Lucia, Myanmar (formerly known as Burma), Papua New Guinea, Pitcairn Islands, Romania, Serbia, Russia, Rwanda, and Tunisia

@adamjgrant @eishay

### DIFF
--- a/packrat/build.properties
+++ b/packrat/build.properties
@@ -1,1 +1,1 @@
-version=3.3.73
+version=3.3.74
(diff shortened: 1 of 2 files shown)


andrewconner:
http://www.npr.org/sections/thetwo-way/2011/12/02/143049567/why-burma-why-myanmar-why-both

PR #16936, opened 2015-08-21
Org pymn refactor

camhashemi:
@leogrim @ryanpbrewster 

deployment logistics:
the only consumer of this is the org admin page, so for now i just commented it out in order to deploy shoebox first with the new (and unused) routing, then deploy abook to serve that route. the next deploy will uncomment the admin call in.

I can also do it the accurate way, keeping the old route available in abook, etc.


kifi-backend/modules/abook/app/com/keepit/abook/commanders/AbookOrganizationRecommendationCommander.scala
```diff
@@ -171,14 +166,14 @@ class AbookOrganizationRecommendationCommander @Inject() (
 
     @inline def isRelevant(emailAccountId: Id[EmailAccountInfo]): Boolean = {
       relevantEmailAccounts.contains(emailAccountId) &&
-        !rejectedRecommendations.contains(emailAccountId)
+        !rejectedRecommendations.contains(emailAccountId) &&
+        !relevantEmailInvites.contains(emailAccountId)
```
camhashemi: filtering out people who are already invited, just like we're doing for users (in this class)

I think we should make invited users available in the non-empty typeahead, when people are actually looking for them. but for the empty-query, it should be as many new people as we can provide.

kifi-backend/modules/shoebox/app/com/keepit/commanders/UserStatisticsCommander.scala
```diff
@@ -200,7 +200,8 @@ class UserStatisticsCommander @Inject() (
     }
 
     val fMemberRecommendations = try {
-      abook.getRecommendationsForOrg(orgId, adminId, disclosePrivateEmails = true, 0, numMemberRecos + members.size + candidates.size)
+      // abook.getRecommendationsForOrg(orgId, viewerIdOpt = None, 0, numMemberRecos + members.size + candidates.size) // muted to deploy new abook route
```
camhashemi, line 203: mute the consumer until the new route is ready

camhashemi:
@leogrim @eishay the refactor we talked about last week

kifi-backend/modules/shoebox/marketing
```diff
@@ -1 +1 @@
-Subproject commit 37fd67e9fd2be9869fe5498a8e945c83873cd988
+Subproject commit d3dc4581840dbc8234e5ea2cd6da376fd04d1cd7
```
leogrim: watch out

leogrim:
:+1: 
Just remove the subproject commit :)

camhashemi:
thanks! got it

PR #16937, opened 2015-08-21
Add an entitiesHash to Keep

ryanpbrewster:
The entitiesHash encodes the set of (libraries, users) that are connected
to a Keep. This will allow us to quickly find Keeps that have exactly
the same set of connected entities.

I had a really obnoxious bug where I was serializing the Keep into Json
incorrectly, so the Keep cache was getting garbage data (but the DB was fine).
When keepRepo.get was called it was automatically looking in the cache and
pulling out the garbage data. Tough to track down.

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/model/Keep.scala
+++ b/kifi-backend/modules/common/app/com/keepit/model/Keep.scala
@@ -34,9 +34,9 @@ case class Keep(
     note: Option[String] = None,
     originalKeeperId: Option[Id[User]] = None,
     organizationId: Option[Id[Organization]] = None,
-    entitiesHash: Option[EntitiesHash] = None) extends ModelWithExternalId[Keep] with ModelWithState[Keep] with ModelWithSeqNumber[Keep] {
+    connectionsHash: Option[KeepConnectionsHash] = None) extends ModelWithExternalId[Keep] with ModelWithState[Keep] with ModelWithSeqNumber[Keep] {
 
-  def sanitizeForDelete: Keep = copy(title = None, state = KeepStates.INACTIVE, isPrimary = false, entitiesHash = Some(KeepConnectionEntities.empty.hash)) // good idea?
+  def sanitizeForDelete: Keep = copy(title = None, state = KeepStates.INACTIVE, isPrimary = false, connectionsHash = Some(KeepConnections.empty.hash)) // good idea?
 
   def clean(): Keep = copy(title = title.map(_.trimAndRemoveLineBreaks()))
 
@@ -71,7 +71,7 @@ case class Keep(
     organizationId = lib.organizationId
   )
 
-  def withEntities(libraries: Set[Id[Library]], users: Set[Id[User]]): Keep = this.copy(entitiesHash = Some(KeepConnectionEntities(libraries, users).hash))
+  def withConnections(libraries: Set[Id[Library]], users: Set[Id[User]]): Keep = this.copy(connectionsHash = Some(KeepConnections(libraries, users).hash))
 
   def isActive: Boolean = state == KeepStates.ACTIVE && isPrimary // isPrimary will be removed shortly
   def isInactive: Boolean = state == KeepStates.INACTIVE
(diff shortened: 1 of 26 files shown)


ryanpbrewster:
@leogrim @andrewconner @stkem

kifi-backend/modules/common/app/com/keepit/model/Keep.scala
```diff
@@ -180,12 +191,33 @@ object Keep {
         "sourceAttributionId" -> k.sourceAttributionId,
         "note" -> k.note,
         "originalKeeperId" -> k.originalKeeperId.orElse(Some(k.userId)),
-        "organizationId" -> k.organizationId
+        "organizationId" -> k.organizationId,
+        "entitiesHash" -> k.entitiesHash
```
ryanpbrewster: @andrewconner I forgot this line

kifi-backend/modules/common/app/com/keepit/model/Keep.scala
```diff
@@ -31,9 +33,10 @@ case class Keep(
     sourceAttributionId: Option[Id[KeepSourceAttribution]] = None,
     note: Option[String] = None,
     originalKeeperId: Option[Id[User]] = None,
-    organizationId: Option[Id[Organization]] = None) extends ModelWithExternalId[Keep] with ModelWithState[Keep] with ModelWithSeqNumber[Keep] {
+    organizationId: Option[Id[Organization]] = None,
+    entitiesHash: Option[EntitiesHash] = None) extends ModelWithExternalId[Keep] with ModelWithState[Keep] with ModelWithSeqNumber[Keep] {
 
-  def sanitizeForDelete: Keep = copy(title = None, state = KeepStates.INACTIVE)
+  def sanitizeForDelete: Keep = copy(title = None, state = KeepStates.INACTIVE, isPrimary = false, entitiesHash = Some(KeepConnectionEntities.empty.hash)) // good idea?
```
ryanpbrewster: @leogrim this is new and relates to things we've talked about in person. Dead keeps should have `isPrimary = false`. As soon as we drop the constraint on 1 URI / library I'm going to kill `isPrimary` entirely.

kifi-backend/modules/common/app/com/keepit/model/Keep.scala
```diff
@@ -31,9 +33,10 @@ case class Keep(
     sourceAttributionId: Option[Id[KeepSourceAttribution]] = None,
     note: Option[String] = None,
     originalKeeperId: Option[Id[User]] = None,
-    organizationId: Option[Id[Organization]] = None) extends ModelWithExternalId[Keep] with ModelWithState[Keep] with ModelWithSeqNumber[Keep] {
+    organizationId: Option[Id[Organization]] = None,
+    entitiesHash: Option[EntitiesHash] = None) extends ModelWithExternalId[Keep] with ModelWithState[Keep] with ModelWithSeqNumber[Keep] {
```
stkem: `Option` is temporary until the data is backfilled, I assume?

ryanpbrewster: Correct.

kifi-backend/modules/common/app/com/keepit/model/Keep.scala
```diff
@@ -68,8 +71,14 @@ case class Keep(
     organizationId = lib.organizationId
   )
 
+  def withEntities(libraries: Set[Id[Library]], users: Set[Id[User]]): Keep = this.copy(entitiesHash = Some(KeepConnectionEntities(libraries, users).hash))
+
   def isActive: Boolean = state == KeepStates.ACTIVE && isPrimary // isPrimary will be removed shortly
   def isInactive: Boolean = state == KeepStates.INACTIVE
+
+  def canBeMergedInto(other: Keep): Boolean = {
+    entitiesHash == other.entitiesHash && keptAt > other.keptAt
```
stkem: `None == None` might be a problem here. As the db stands right now with this, everything can be merged into the very first keep. Probably not intended?

ryanpbrewster: Hrm. So this method is probably named kind of poorly.

In KeepCommander, before any "merging" is done, I pre-check that the actual `KeepConnectionEntities` match up, which necessarily means that the `entitiesHash` must match. I double-check it here, just to be sure, but really this method is supposed to be "given that these two keeps have exactly the same entities connected to them, and that they have the same URI, should we merge them?"

Right now, the answer is always "yes", and we always merge the older keep into the newer keep (by "merge" I mean we deactivate the old keep). Going forward, the answer should be "as long as the metadata on the keeps is compatible (i.e., the same title, same note, and no discussion).

ryanpbrewster: I'll probably rename this method to `hasStrictlyLessMetadata` or something along those lines. That will make it clear that it is insufficient to check only that before merging.

stkem: :+1:

PR #16939, opened 2015-08-21
[WIP] Activity Feed

adamjgrant:


kifi-backend/modules/shoebox/angular/src/common/micrologger.js
```diff
@@ -0,0 +1,63 @@
...
+
+  micrologger.Spec.prototype.respond = function(args) {
+    for (var _i = 0, len = this.tests.length; _i < len; _i++) {
+      this.tests[_i].respond(args[_i]);
+    }
+  };
+
+  return micrologger;
+}]);
+
```
adamjgrant, line 62: This is a browser monitoring add on [I've been working on](http://adamjgrant.github.io/MicroLogger/) That should give us more detailed reports in Airbrake 

Instead of `Client received HTTP 400`, we'd see something like `[FAIL] get members for org, Client received HTTP 400`

kifi-backend/modules/shoebox/angular/dev.html
```diff
@@ -14,6 +14,15 @@
...
 </div>
+
+<script>
+  // Uncomment this to see micrologger reports in the console.
+  window.ml = {
+    pass: function(fn) { fn(); },
+    fail: function(fn) { fn(); }
+  }
+</script>
+
```
adamjgrant, line 25: Without this, passing tests from micrologger do not get logged to the console. I'd imagine that would be an annoying thing to have in prod.

kifi-backend/modules/shoebox/angular/src/feed/feedCtrl.js
```diff
@@ -0,0 +1,21 @@
...
+    $scope.feed = [];
+
+    (function() {
+      ml.specs.getsFeed = new ml.Spec([
+        new ml.Assert('Feed retrieved in 3 seconds or less', 3000),
+        new ml.Expect('Feed returns a list', function(data) { return data.length; })
+      ]);
+      $scope.feed = [];
+      net.getFeed(10).then(function(response) {
+        ml.specs.getsFeed.respond([undefined, response.data.keeps]);
```
adamjgrant: ...so it can be `respond`ed to in one line.

kifi-backend/modules/shoebox/angular/src/common/micrologger.js
cvializ, line 62: I think I'll need a walkthrough on the syntax/usage, but it looks useful!

andrewconner:
My understanding is we're not killing recos for normal users just yet, just moving people in orgs to the new activity feed. This should probably just be gated with an experiment.

cvializ:
noted, I'll add the recos code back

PR #16940, opened 2015-08-21
Made an Admin library controller method safer

ryanpbrewster:
It now uses LibraryCommander methods instead of touching the repo directly

@eishay


kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -863,6 +860,23 @@ class LibraryCommanderImpl @Inject() (
...
+  def unsafeTransferLibrary(libId: Id[Library], newOwner: Id[User]): Library = {
+    db.readWrite { implicit s =>
+      val owner = userRepo.get(newOwner)
+      assert(owner.state == UserStates.ACTIVE)
+
+      val lib = libraryRepo.getNoCache(libId)
+
+      libraryMembershipRepo.getWithLibraryIdAndUserId(libId, lib.ownerId).foreach { oldOwnerMembership =>
+        libraryMembershipRepo.save(oldOwnerMembership.withState(LibraryMembershipStates.INACTIVE))
+      }
```
ryanpbrewster, line 871: The old owner's membership needs to be removed. Otherwise you'll have a library with multiple owners.

PR #16943, opened 2015-08-21
More notification info generation

maxov:
@leogrim 
@andrewconner 
@ryanpbrewster 

Stopped using the generic DbView and switched to using a less generic BatchedNotificationInfos.


kifi-backend/modules/common/app/com/keepit/notify/info/DbView.scala
```diff
@@ -0,0 +1,101 @@
...
+import com.keepit.common.db.Id
+import com.keepit.common.store.ImagePath
+import com.keepit.model._
+
+/**
+ * Represents a view, or subset, of the shoebox database, mostly from the perspective of Eliza.
+ */
+trait DbView {
+
+  def lookup(kind: String, id: Id[_]): Any
```
ryanpbrewster: Does `kind` have to be a `String`? Can we use a `sealed abstract class` so that the compiler will tell us when we forgot something?

kifi-backend/modules/common/app/com/keepit/notify/info/DbView.scala
```diff
@@ -0,0 +1,101 @@
...
+case class DbViewRequest[M <: HasId[M], R](kind: String, id: Id[M]) {
+
+  /**
+   * Actually look up the request in a db subset. Because the db subset is untyped,
+   * a final cast is done.
+   *
+   * @param subset The subset to look up in
+   * @return The looked up result
+   */
+  def lookup(subset: DbView): R = subset.lookup(kind, id).asInstanceOf[R]
```
ryanpbrewster: This worries me. Can we have a typed map (say a `Map[InfoType, Info[InfoType]]` or something)?

kifi-backend/modules/common/app/com/keepit/notify/info/DbView.scala
maxov: I realized that the keys themselves can be used as kinds, so I'm going to take out all strings. Fixed in 1f21882

kifi-backend/modules/common/app/com/keepit/notify/info/DbView.scala
maxov: I made MapDbView the only implementation of DbView and added strong types to its add and lookup methods. The methods are pretty simple and work generically, so I think that's good enough. See https://github.com/gratimax/keepit/blob/1f2188258caf69aefae099733a0add7a16bf609d/kifi-backend/modules/common/app/com/keepit/notify/info/DbView.scala#L7-L34

kifi-backend/modules/common/app/com/keepit/notify/NotificationEventSender.scala
```diff
@@ -0,0 +1,44 @@
...
+  def send[N <: NotificationEvent](event: N): Unit = send(ExistingDbView()(event))
+
+  def process[N <: NotificationEvent](usingView: UsingDbView[NotificationInfo], existing: Option[ExistingDbView[N]] = None): Future[NotificationInfo] = {
+    val info = getView(usingView.requests, existing)
+    info.map { view =>
+      usingView.fn(view)
+    }
+  }
+
+  def getView(requests: Seq[DbViewRequest[M, T] forSome { type M <: HasId[M]; type T}], existing: Option[ExistingDbView[_]]): Future[DbView] = {
```
ryanpbrewster: What's the difference between using `forSome` and just having `Seq[DbViewRequest[M <: HasId[M], T]]`?

maxov: That would require parameterizing getView on M and T, which is not what is actually there. It's not a sequence of DbViewRequests with the same M and T type, it's a sequence of DbViewRequests with different M and T types that are still related in some way. Using wildcards `Seq[DbViewRequest[_, _]]` also implies the wrong thing, that it is a sequence of DbViewRequests of the same type that have some unknown M and T. This is the only way I could get the compiler to be happy (even though intellij doesn't like it...)

ryanpbrewster: Interesting. I spent about 30 minutes fiddling with this exact problem and couldn't find a solution. That's neat.

kifi-backend/modules/common/app/com/keepit/notify/info/DbView.scala
```diff
@@ -0,0 +1,130 @@
+package com.keepit.notify.info
+
+import com.keepit.common.db.Id
+import com.keepit.common.store.ImagePath
+import com.keepit.model._
+
+/**
+ * Represents a view, or subset, of the shoebox database, mostly from the perspective of Eliza.
```
andrewconner: I'm jumping in the middle here, but could this go in Eliza's package? Or are you hoping for general use?

kifi-backend/modules/common/app/com/keepit/notify/info/DbView.scala
```diff
@@ -0,0 +1,130 @@
...
+    new DbView(keyMap + (key -> assocObjMap))
+  }
+
+  def update[M <: HasId[M]](existing: ExistingDbViewModel[M]): DbView = {
+    val model = existing.model
+    update(existing.key, model.id.get, model)
+  }
+
+  def lookup[M <: HasId[M], R](key: DbViewKey[M, R], id: Id[M]): R =
+    keyMap(key)(id).asInstanceOf[R]
```
andrewconner: These all creep me out a bit. Is this worth worrying about (could this be wrong)? I have several toy examples of horrible things you can do with `asInstanceOf`.

kifi-backend/modules/common/app/com/keepit/notify/info/DbView.scala
maxov: Not really, this is used a bit on the shoebox side as well before sending notifications.

kifi-backend/modules/common/app/com/keepit/notify/info/DbView.scala
maxov: I don't like these either (and there's a couple more), but I couldn't figure out a generic alternative without being really painful. I tried to keep cases here as isolated and as documented as possible. In this specific case, there's a guarantee enforced by the type of the 'update' method that every value in the nested map has the same `R` type as its key, as per the class javadoc.

maxov:
@leogrim 
@andrewconner 
@ryanpbrewster

Ready for review before merge. There should be no functional changes for any user, even with those with the experiment. The next PR or two should set that up.

maxov:
@leogrim 
@andrewconner 
@ryanpbrewster 

Fixed all tests (and guice dependency loops...). Ready ready for review.

maxov:
@leogrim 
@andrewconner 
@ryanpbrewster 

I would like to merge today... please review

andrewconner:
Will do, thanks.

kifi-backend/modules/common/app/com/keepit/shoebox/ShoeboxServiceClient.scala
```diff
@@ -133,6 +133,14 @@ trait ShoeboxServiceClient extends ServiceClient {
   def getIngestableOrganizationDomainOwnerships(seqNum: SequenceNumber[OrganizationDomainOwnership], fetchSize: Int): Future[Seq[IngestableOrganizationDomainOwnership]]
   def getPrimaryOrg(userId: Id[User]): Future[Option[Id[Organization]]]
   def getOrganizationsForUsers(userIds: Set[Id[User]]): Future[Map[Id[User], Set[Id[Organization]]]]
+  def getLibraries(libraryIds: Seq[Id[Library]]): Future[Map[Id[Library], Library]]
+  def getUserImages(userIds: Seq[Id[User]]): Future[Map[Id[User], String]]
+  def getKeeps(keepIds: Seq[Id[Keep]]): Future[Map[Id[Keep], Keep]]
+  def getLibraryUrls(libraryIds: Seq[Id[Library]]): Future[Map[Id[Library], String]]
+  def getLibraryInfos(libraryIds: Seq[Id[Library]]): Future[Map[Id[Library], LibraryNotificationInfo]]
```
leogrim: I'm not sure we need to introduce yet another `Library` derived view

kifi-backend/modules/common/app/com/keepit/shoebox/ShoeboxServiceClient.scala
```diff
@@ -133,6 +133,14 @@ trait ShoeboxServiceClient extends ServiceClient {
...
+  def getLibraryOwners(libraryIds: Seq[Id[Library]]): Future[Map[Id[Library], User]]
```
leogrim: Basically these three methods and `getLibraries` can probably be replaced by one, can't they?

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
```diff
@@ -70,75 +70,75 @@ class OrganizationCommanderImpl @Inject() (
     }
   }
 
-  private def getOrganizationInfoHelper(orgId: Id[Organization], viewerIdOpt: Option[Id[User]])(implicit session: RSession): OrganizationInfo = {
+  def getOrganizationCardHelper(orgId: Id[Organization], viewerIdOpt: Option[Id[User]])(implicit session: RSession): OrganizationCard = {
```
leogrim: What's the difference between this one and the one in `OrganizationCardCommander`?

kifi-backend/modules/shoebox/app/com/keepit/commanders/UserCommander.scala
```diff
@@ -373,7 +344,7 @@ class UserCommander @Inject() (
             val toNotify = contacts.diff(alreadyConnectedUsers) - newUserId
 
             log.info("sending new user contact notifications to: " + toNotify)
-            val emailsF = toNotify.map { userId => emailSender.contactJoined(userId, newUserId) }
+            val emailsF = toNotify.map { userId => contactJoinedEmailSender.get.apply(userId, newUserId) }
```
leogrim, line 346: :+1:

kifi-backend/modules/shoebox/app/com/keepit/commanders/UserCommander.scala
```diff
@@ -416,7 +387,7 @@ class UserCommander @Inject() (
 
   def sendWelcomeEmail(newUser: User, withVerification: Boolean = false, targetEmailOpt: Option[EmailAddress] = None, isPlainEmail: Boolean = true): Future[Unit] = {
     if (!db.readOnlyMaster { implicit session => userValueRepo.getValue(newUser.id.get, UserValues.welcomeEmailSent) }) {
-      val emailF = emailSender.welcome(newUser.id.get, targetEmailOpt, isPlainEmail)
+      val emailF = welcomeEmailSender.get.apply(newUser.id.get, targetEmailOpt, isPlainEmail)
```
leogrim, line 389: :+1:

kifi-backend/modules/common/app/com/keepit/shoebox/ShoeboxServiceClient.scala
```diff
@@ -810,6 +818,46 @@ class ShoeboxServiceClientImpl @Inject() (
...
+  def getOrganizations(orgIds: Seq[Id[Organization]]): Future[Map[Id[Organization], Organization]] = {
+    val payload = Json.toJson(orgIds)
+    call(Shoebox.internal.getOrganizations(), payload).map { _.json.as[Map[Id[Organization], Organization]] }
+  }
+
+  def getOrganizationInfos(orgIds: Seq[Id[Organization]]): Future[Map[Id[Organization], OrganizationNotificationInfo]] = {
+    val payload = Json.toJson(orgIds)
+    call(Shoebox.internal.getOrganizationInfos(), payload).map { _.json.as[Map[Id[Organization], OrganizationNotificationInfo]] }
+  }
+
```
leogrim, line 851: None of these are cached, I'm worried about these new routes as opposed to using existing / well cached ones.

kifi-backend/modules/common/app/com/keepit/eliza/ElizaServiceClient.scala
```diff
@@ -75,7 +76,7 @@ trait ElizaServiceClient extends ServiceClient {
 
   def unsendNotification(messageHandle: Id[MessageHandle]): Unit
 
-  def sendNotificationEvent(event: NotificationEvent): Unit
+  def sendNotificationEvent(event: NotificationEventSender.EventWithInfo): Unit
```
ryanpbrewster: Maybe `sendNotificationEvent(event: NotificationEvent, notificationInfo: Option[NotificationInfo] = None)` or something to make it clear that it's just a notification event with some optional optimization.

kifi-backend/modules/common/app/com/keepit/notify/info/DbView.scala
```diff
@@ -0,0 +1,130 @@
...
+
+  def contained(view: DbView): Boolean = view.contains(key, id)
+
+}
+
+/**
+ * Represents a wrapper of a function that requests a whole bunch of items from a potential db view, then
+ * constructs a value.
+ */
+case class UsingDbView[A](requests: DbViewRequest[M, T] forSome { type M <: HasId[M]; type T }*)(val fn: DbView => A)
```
ryanpbrewster: Good comment. This type signature is really something.

kifi-backend/modules/common/app/com/keepit/shoebox/ShoeboxServiceClient.scala
ryanpbrewster: I understand why things like `getLibraryOwners` is separate from `getLibraries` (even though technically `getLibraries` will also give you the library owners), but I don't understand why `getLibraryInfos` is a function. Why does `Shoebox` even know how to generate a `NotificationInfo`?

kifi-backend/modules/shoebox/app/com/keepit/notify/info/ShoeboxDbViewRequestHandlerImpl.scala
```diff
@@ -0,0 +1,94 @@
...
+        Future.successful(db.readOnlyReplica { implicit session =>
+          orgRepo.getByIds(requests.map(_.id).toSet)
+        }.map {
+          case (id, org) =>
+            val orgImageOpt = organizationAvatarCommander.getBestImageByOrgId(org.id.get, ProcessedImageSize.Medium.idealSize)
+            (id, OrganizationNotificationInfoBuilder.fromOrganization(org, orgImageOpt))
+        })
+      }
+
+}
```
ryanpbrewster: Why are all these things `Future` if we're computing them on the spot?

kifi-backend/modules/shoebox/marketing
```diff
@@ -1 +1 @@
-Subproject commit 37fd67e9fd2be9869fe5498a8e945c83873cd988
+Subproject commit d3dc4581840dbc8234e5ea2cd6da376fd04d1cd7
```
ryanpbrewster: Necessary?

ryanpbrewster:
This is a monster PR, and I'm still not sure I understand all of it. I'm a little concerned about the separation of concerns between Shoebox and Eliza. Right now it feels like Shoebox knows a little bit too much, and it's leaking abstractions.

That being said, this is really incredible work. Awesome job. I _really_ like the interface to `DbView` and `UsingDbView`. Those look super clean to use.

kifi-backend/modules/common/app/com/keepit/eliza/ElizaServiceClient.scala
```diff
@@ -184,7 +185,7 @@ class ElizaServiceClientImpl @Inject() (
     call(Eliza.internal.unsendNotification(messageHandle), attempts = 6, callTimeouts = longTimeout)
   }
 
-  def sendNotificationEvent(event: NotificationEvent): Unit = {
+  def sendNotificationEvent(event: NotificationEventSender.EventWithInfo): Unit = {
```
stkem: For remote service calls I much prefer to always return a `Future` (could well be `Future[Unit]`) so that the caller can know when the call completed (for example in order to sequentialize calls)

kifi-backend/modules/common/app/com/keepit/notify/NotificationEventSender.scala
```diff
@@ -0,0 +1,48 @@
+package com.keepit.notify
+
+import com.google.inject.{Inject, Singleton}
+import com.keepit.common.db.Id
+import com.keepit.eliza.ElizaServiceClient
+import com.keepit.notify.NotificationEventSender.EventWithInfo
+import com.keepit.notify.info._
```
stkem: I'm generally not a fan of `._` imports. Makes it harder to read the code because it's not immediately clear where things are coming from.
Sometimes hard to avoid, especially if you are after a bunch of implicits (another thing that _can_ make code hard to read/reason about, while of course still having very good uses), but as a rule try to minimize it.

kifi-backend/modules/common/app/com/keepit/notify/NotificationEventSender.scala
```diff
@@ -0,0 +1,48 @@
...
+import scala.concurrent.{ExecutionContext, Future}
+
+@Singleton
+class NotificationEventSender @Inject() (
+  infoProcessing: NotificationInfoProcessing,
+  elizaServiceClient: ElizaServiceClient,
+  implicit val ec: ExecutionContext
+) {
+
+  def send[N <: NotificationEvent](wrappedEvent: ExistingDbView[N]): Unit = {
```
stkem: Same thing as with the Service client (Really applies to anything that triggers an asynchronous computation). This should return a `Future[Unit]` s the caller can tell when things are finished (or failed).

kifi-backend/modules/common/app/com/keepit/notify/NotificationEventSender.scala
```diff
@@ -0,0 +1,48 @@
...
+    val event = wrappedEvent.result
+    val infoFut = infoProcessing.process(event.kind.info(Set(event)), Some(wrappedEvent))
+    infoFut.map { info =>
```
stkem: Re my comment right above, this should be `flatMap`

kifi-backend/modules/common/app/com/keepit/notify/NotificationEventSender.scala
ryanpbrewster: I'm not sure I agree. With IntelliJ you can always immediately jump to the definition of a method/class you're using.

I think that by default IntelliJ will import modules explicitly, up to 5, and then it will use a wildcard. That seems pretty reasonable to me.

stkem: No everyone uses IntelliJ

kifi-backend/modules/common/app/com/keepit/notify/info/DbView.scala
stkem: I'm with @andrewconner on this one. While a single use might be fairly safe, the fact that you need it at all is usually an indicator that there is some flaw in the type architecture thats going to come back to bite you later (along the lines of "if you already need casts for the stuff you've thought of, it might very well be that the stuff you haven't though of - i.e. future features - end up being entirely impossible.)
Unfortunately I don't know the high level architecture here well enough to make a recommendation, but I would strongly advocate taking another look at this. (I'm happy to help, but someone is going to have to walk me through the bigger picture first)

kifi-backend/modules/common/app/com/keepit/notify/info/NotificationInfoProcessing.scala
```diff
@@ -0,0 +1,44 @@
...
+    info.map { view =>
+      usingView.fn(view)
+    }
+  }
+
+  def getView(requests: Seq[ExDbViewRequest]): Future[DbView] = {
+    val grouped = requests.groupBy(_.key)
+    // We know that after the groupBy, the type of requests is the same because they have the same key.
+    // This cast also makes the compiler know
+    grouped.toList.asInstanceOf[List[(DbViewKey[M, R], Seq[DbViewRequest[M, R]]) forSome { type M <: HasId[M]; type R}]]
```
andrewconner: How does this work since it's not being returned? How would this side-effect? Should the next line foldLeft on this instead?

kifi-backend/modules/common/app/com/keepit/shoebox/ShoeboxServiceClient.scala
stkem, line 851: Dito. A lot of very similar endpoints.

kifi-backend/modules/common/app/com/keepit/notify/info/NotificationInfoProcessing.scala
maxov: Hmm... I was sure that the compiler was erroring here for some reason, this cast was supposed to stop that. It might have been because I changed the signature of the grouping method. Otherwise, this does pretty much absolutely nothing. I'll look into it, and hopefully remove it

kifi-backend/modules/common/app/com/keepit/notify/info/DbViewRequestHandler.scala
```diff
@@ -0,0 +1,37 @@
...
+trait DbViewRequestHandler {
+
+  val handlers: DbViewRequestHandlers
+
+  def apply[M <: HasId[M], R](key: DbViewKey[M, R], requests: Seq[DbViewRequest[M, R]]): Future[Map[Id[M], R]] =
+    handlers.apply(key, requests)
+
+}
+
+class DbViewRequestHandlers(private val handlers: Map[ExDbViewKey, Seq[ExDbViewRequest] => Future[Map[Id[Any], Any]]]) {
```
stkem: I'm not sure I'm I get the advantage of `DbViewRequestHandlers`. 
1) It abstracts away that you are effectively using a model that is stored somewhere else, which is arguably not a good things if it's too far removed because generally remote data should be accesses sparingly as as purposely as possibly (i.e. as opposed to generic remote data access).
2) More importantly, if I'm not mistaken, this is prone to runtime failure. If someone wants, say, a Library View and the handler for that is not registered it will fail at runtime. There should at the least be method for each of the different types supplied, not a Map with fields (which effectively behaves like a method table). (The fact that I'm not sure about the runtime failure after looking at this for 30 minutes is probably also not a good thing, although I don't know for whom :-/)

kifi-backend/modules/common/app/com/keepit/notify/info/ModelNotificationInfos.scala
```diff
@@ -1,26 +1,42 @@
...
 
-// todo could be more generic
 @json
 case class LibraryImageInfo(path: ImagePath, x: Int, y: Int)
 
 @json
+case class UserNotificationInfo(
+  user: BasicUser,
+  path: Path,
+  imageUrl: String)
```
ryanpbrewster: I'm looking at `BasicUser`, and it already has `pictureName` and `username`. Do we need to have a new class for this?

kifi-backend/modules/common/app/com/keepit/notify/info/ModelNotificationInfos.scala
```diff
@@ -1,26 +1,42 @@
...
 @json
 case class OrganizationNotificationInfo(
   id: PublicId[Organization],
   name: String,
+  abbreviatedName: String,
   handle: Option[PrimaryOrganizationHandle],
-  image: Option[ImagePath])
+  image: Option[ImagePath],
+  path: Path,
+  owner: UserNotificationInfo)
```
leogrim: We already have classes very similar to all of these actually, can we not introduce new ones?
It's better to get a well-cached object with a few extra-fields you may not need than creating new classes for every use-case which prevent us from mutualizing caching, makes sense?

maxov: I'm only introducing `KeepNotificationInfo` and `UserNotificationInfo`. The previous models, `OrganizationNotificationInfo` and `LibraryNotificationInfo` already existed. I will try to take a look and see if there are any models that already do what is needed, since that would be preferable

maxov:
@leogrim 
@andrewconner 
@ryanpbrewster 
@stkem 

Ready, for hopefully the last review?

kifi-backend/modules/common/app/com/keepit/common/routes/Router.scala
```diff
@@ -165,6 +165,10 @@ object Shoebox extends Service {
     def hasOrganizationMembership(orgId: Id[Organization], userId: Id[User]) = ServiceRoute(GET, "/internal/shoebox/database/hasOrganizationMembership", Param("orgId", orgId), Param("userId", userId))
     def getIngestableOrganizationDomainOwnerships(seqNum: SequenceNumber[OrganizationDomainOwnership], fetchSize: Int) = ServiceRoute(GET, "/internal/shoebox/database/getIngestableOrganizationDomainOwnerships", Param("seqNum", seqNum), Param("fetchSize", fetchSize))
     def getOrganizationsForUsers() = ServiceRoute(POST, "/internal/shoebox/database/getOrganizationsForUsers")
+    def getUserInfos() = ServiceRoute(POST, "/internal/shoebox/database/getUserInfos")
+    def getKeepInfos() = ServiceRoute(POST, "/internal/shoebox/database/getKeepInfos")
+    def getLibraryInfos() = ServiceRoute(POST, "/internal/shoebox/database/getLibraryInfos")
+    def getOrganizationInfos() = ServiceRoute(POST, "/internal/shoebox/database/getOrganizationInfos")
```
ryanpbrewster: Are these ones used by Search? If not, we should try to get some overlap.

maxov: Whoops, forgot to take these out. They don't actually point anywhere anymore

kifi-backend/modules/common/app/com/keepit/eliza/ElizaServiceClient.scala
```diff
@@ -184,9 +184,9 @@ class ElizaServiceClientImpl @Inject() (
     call(Eliza.internal.unsendNotification(messageHandle), attempts = 6, callTimeouts = longTimeout)
   }
 
-  def sendNotificationEvent(event: NotificationEvent): Unit = {
+  def sendNotificationEvent(event: NotificationEvent): Future[Unit] = {
     val payload = Json.toJson(event)
-    call(Eliza.internal.sendNotificationEvent(), payload)
+    call(Eliza.internal.sendNotificationEvent(), payload).map(_ => ())
```
ryanpbrewster: Kind of an odd piece of code. I see why it's necessary.

Let's use `imap` (a custom map we wrote for `Future` that reuses the existing execution context instead of using a new thread).

kifi-backend/modules/eliza/app/com/keepit/notify/info/NotificationInfoGenerator.scala
```diff
@@ -0,0 +1,103 @@
...
+class NotificationInfoGenerator @Inject() (
+    shoeboxServiceClient: ShoeboxServiceClient,
+    implicit val ec: ExecutionContext) {
+
+  private def genericKind(notif: Notification) = {
+    // the notification and items don't know that they have the same kind
+    notif.kind.asInstanceOf[NotificationKind[NotificationEvent]]
+  }
+
+  def generateInfo(notifs: Seq[(Notification, Set[NotificationItem])]): Future[Seq[(Notification, Set[NotificationItem], NotificationInfo)]] = {
```
ryanpbrewster: Is this a `Map` (i.e., are there duplicate keys?).

kifi-backend/modules/eliza/app/com/keepit/notify/info/NotificationInfoGenerator.scala
```diff
@@ -0,0 +1,103 @@
...
+    }.toSet
+
+    val libsF = shoeboxServiceClient.getBasicLibraryDetails(libRequests, ProcessedImageSize.Small.idealSize, None)
+    val orgsF = shoeboxServiceClient.getBasicOrganizationsByIds(orgRequests)
+    val keepsF = shoeboxServiceClient.getBasicKeepsByIds(keepRequests)
+
+    val externalIdsF = for {
+      orgs <- orgsF
+      keeps <- keepsF
+    } yield (orgs.values.map(_.ownerId) ++ keeps.values.map(_.ownerId)).toSeq.distinct
```
ryanpbrewster, line 58: `.toSet`?

kifi-backend/modules/eliza/app/com/keepit/notify/info/NotificationInfoGenerator.scala
```diff
@@ -0,0 +1,103 @@
...
+      infoRequest.requests.collect {
+        case r: NotificationInfoRequest.RequestUser => r.id
+      }
+    }.toSet
+
+    val libsF = shoeboxServiceClient.getBasicLibraryDetails(libRequests, ProcessedImageSize.Small.idealSize, None)
+    val orgsF = shoeboxServiceClient.getBasicOrganizationsByIds(orgRequests)
+    val keepsF = shoeboxServiceClient.getBasicKeepsByIds(keepRequests)
+
+    val externalIdsF = for {
```
ryanpbrewster: These are `userIds`, right? `externalIds` is ambiguous.

kifi-backend/modules/eliza/app/com/keepit/notify/info/NotificationInfoGenerator.scala
maxov: Yeah, I guess that makes more sense than what I have.

kifi-backend/modules/eliza/app/com/keepit/notify/info/NotificationInfoGenerator.scala
ryanpbrewster: I see you get the internal ids below. Maybe `externalUserIds`?

maxov: yeah, they are external user Ids. Will rename

kifi-backend/modules/eliza/app/com/keepit/notify/info/NotificationInfoGenerator.scala
maxov, line 58: To service client methods that takes external user Ids takes a seq and returns a seq, so order is still important. I don't want to convert to a set and then back

kifi-backend/modules/shoebox/app/com/keepit/controllers/internal/ShoeboxController.scala
```diff
@@ -533,6 +540,43 @@ class ShoeboxController @Inject() (
...
+    val orgInfos = db.readOnlyReplica { implicit session =>
+      orgRepo.getByIds(orgIds.toSet)
+    }.map {
+      case (id, org) =>
+        val orgImageOpt = organizationAvatarCommander.getBestImageByOrgId(org.id.get, ProcessedImageSize.Medium.idealSize)
+        (id, NotificationInfoModel.organization(org, orgImageOpt))
+    }
+    Ok(Json.toJson(orgInfos))
+  }
+
```
ryanpbrewster: Do these still need to exist?

maxov: Wow, I was sure I removed them. Yeah, they are unecessary

kifi-backend/modules/common/app/com/keepit/notify/model/event/ConnectionEvent.scala
```diff
@@ -22,22 +22,21 @@ trait NewConnectionInviteImpl extends NonGroupingNotificationKind[NewConnectionI
...
   )(NewConnectionInvite.apply, unlift(NewConnectionInvite.unapply))
 
-  override def info(event: NewConnectionInvite): UsingDbSubset[NotificationInfo] = {
-    import NeedInfo._
-    UsingDbSubset(Seq(
-      user(event.inviterId), userImageUrl(event.inviterId)
-    )) { subset =>
-      val inviter = user(event.inviterId).lookup(subset)
-      val inviterImage = userImageUrl(event.inviterId).lookup(subset)
+  override def info(event: NewConnectionInvite): RequestingNotificationInfos[NotificationInfo] = {
```
leogrim: Do we need `override` here?

maxov: No. Is it better style to leave it out? From what I know, having `override` means that not overriding the method leaves compile errors.

kifi-backend/modules/eliza/app/com/keepit/notify/info/NotificationInfoGenerator.scala
```diff
@@ -0,0 +1,102 @@
...
+    } yield {
+      notifs map {
+        case (notif, items) =>
+          val infoRequest = notifInfoRequests(notif)
+          (notif, (items, infoRequest.fn(batchedInfos)))
+      }
+    }
+  }
+
+}
```
leogrim, line 97: :+1:

kifi-backend/modules/eliza/app/com/keepit/notify/info/NotificationInfoGenerator.scala
```diff
@@ -0,0 +1,102 @@
...
+
+import scala.concurrent.{ ExecutionContext, Future }
+
+class NotificationInfoGenerator @Inject() (
+    shoeboxServiceClient: ShoeboxServiceClient,
+    implicit val ec: ExecutionContext) {
+
+  private def genericKind(notif: Notification): NotificationKind[NotificationEvent] = {
+    // the notification and items don't know that they have the same kind
+    notif.kind.asInstanceOf[NotificationKind[NotificationEvent]]
```
stkem:
``````
notif.kind match {
  case x: NotificationKind[NotificationEvent] => x
  case _ => throw Exception("This should be impossible")
}
```?
``````

leogrim: Or maybe use a sequence of typed wrappers instead of map?

leogrim:
lgtm, can we move some of this code out of Common?

maxov:
@leogrim Done. The diff moved up by a lot, but that's because a whole lot of code was just moved.

PR #16944, opened 2015-08-21
add searchUsersByName to search

camhashemi:
@leogrim @ryanpbrewster

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/common/routes/Router.scala
+++ b/kifi-backend/modules/common/app/com/keepit/common/routes/Router.scala
@@ -16,6 +16,7 @@ import com.keepit.cortex.core.{ StatModel, ModelVersion }
 import com.keepit.cortex.models.lda.DenseLDA
 import com.keepit.common.mail.EmailAddress
 import com.keepit.abook.model.{ IngestableContact, EmailAccountInfo }
+import com.keepit.typeahead.UserPrefixSearchRequest
 import org.joda.time.DateTime
 import com.keepit.common.time._
 import com.keepit.social.SocialNetworkType
@@ -183,6 +184,7 @@ object Search extends Service {
     def getLibraryDocument() = ServiceRoute(POST, s"/internal/search/index/library/document")
     def searchKeeps(userId: Id[User], query: String) = ServiceRoute(POST, "/internal/search/search/keeps", Param("userId", userId), Param("query", query))
     def searchUsers() = ServiceRoute(POST, "/internal/search/search/users")
+    def searchUsersByName() = ServiceRoute(POST, "/internal/search/search/userByName")
     def userTypeahead() = ServiceRoute(POST, "/internal/search/search/userTypeahead")
     def explainUriResult(query: String, userId: Id[User], uriId: Id[NormalizedURI], libraryId: Option[Id[Library]], lang: Option[String], debug: Option[String], disablePrefixSearch: Boolean, disableFullTextSearch: Boolean) =
       ServiceRoute(GET, "/internal/search/search/uri/explain", Param("query", query), Param("userId", userId), Param("uriId", uriId), Param("libraryId", libraryId), Param("lang", lang), Param("debug", debug), Param("disablePrefixSearch", disablePrefixSearch), Param("disableFullTextSearch", disableFullTextSearch))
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/search/app/com/keepit/search/controllers/search/SearchController.scala
```diff
@@ -94,6 +95,16 @@ class SearchController @Inject() (
     Ok(Json.toJson(res))
   }
 
+  def searchUsersByName() = Action.async(parse.tolerantJson) { request =>
+    val searchRequest = request.body.as[UserPrefixSearchRequest]
+    val filter = SearchFilter(proximity = Some(ProximityScope.all), user = None, library = None, organization = None)
+    val contextFut = Future.successful(SearchContext(None, orderBy = SearchRanking.relevancy, filter = filter, disablePrefixSearch = false, disableFullTextSearch = true))
+    userSearchCommander.searchUsers(userId = searchRequest.userId, acceptLangs = Seq.empty, experiments = Set.empty, query = searchRequest.query, contextFuture = contextFut, maxHits = searchRequest.maxHits, explain = None).map { userSearchResult =>
```
camhashemi: @leogrim I'm inferring what these parameters do. My intention is to do a system-wide query of users by name.

kifi-backend/modules/common/app/com/keepit/typeahead/Typeahead.scala
```diff
@@ -154,3 +156,7 @@ object TypeaheadHit {
 
 case class TypeaheadHit[I](score: Int, name: String, ordinal: Int, info: I)
 
+@json
+case class UserPrefixSearchRequest(userId: Id[User], query: String, maxHits: Int)
+@json
+case class UserSearchHit(userId: Id[User], score: Float)
```
camhashemi: Created a new class because `TypeaheadHit` wants a name that we don't have. Figured it was better to create a straightforward class than hack `TypeaheadHit` where `name = ""` and `ordinal = 0`.

kifi-backend/modules/search/app/com/keepit/search/controllers/search/SearchController.scala
camhashemi: when we need it, we can extend this to take different `UserSearchRequest` subclasses and encode the search parameters into each request

kifi-backend/modules/common/app/com/keepit/common/routes/Router.scala
```diff
@@ -183,6 +184,7 @@ object Search extends Service {
     def getLibraryDocument() = ServiceRoute(POST, s"/internal/search/index/library/document")
     def searchKeeps(userId: Id[User], query: String) = ServiceRoute(POST, "/internal/search/search/keeps", Param("userId", userId), Param("query", query))
     def searchUsers() = ServiceRoute(POST, "/internal/search/search/users")
+    def searchUsersByName() = ServiceRoute(POST, "/internal/search/search/userByName")
```
leogrim: Can we go for `/internal/search/search/users/name` ?

camhashemi: yup

kifi-backend/modules/common/app/com/keepit/typeahead/Typeahead.scala
leogrim: You're right, could we call it `UserPrefixSearchHit` to be specific until we decided to actually extend these? There's also `UserShardHit` in Search and I'm trying to keep it clear for someone new to jump in :)

kifi-backend/modules/search/app/com/keepit/search/controllers/search/SearchController.scala
```diff
@@ -94,6 +95,16 @@ class SearchController @Inject() (
     Ok(Json.toJson(res))
   }
 
+  def searchUsersByName() = Action.async(parse.tolerantJson) { request =>
+    val searchRequest = request.body.as[UserPrefixSearchRequest]
+    val filter = SearchFilter(proximity = Some(ProximityScope.all), user = None, library = None, organization = None)
```
leogrim: I think you can leave proximity out

kifi-backend/modules/common/app/com/keepit/typeahead/Typeahead.scala
camhashemi: np

kifi-backend/modules/search/app/com/keepit/search/controllers/search/SearchController.scala
leogrim: You need to get the user's experiments here, and `acceptLangs` even though that may matter less here.
Check out one of the search controllers to see how to do that easily

camhashemi:
@leogrim cool to merge?

PR #16948, opened 2015-08-24
Crazy new keep-library-user set URI merge logic

ryanpbrewster:
Keeps are deactivated based on their connectionsHash and the amount of metadata
they have. For now, in order to comply with URI-to-library constraints,
we ALWAYS deactivate older keeps.

@andrewconner @leogrim


andrewconner:
I think your new logic is good about merging. The rest looks good as well.

PR #16950, opened 2015-08-24
org-pymn: don't filter users/emails who have already been invited

camhashemi:
@leogrim @ryanpbrewster talked with Jen, we want people to be able to re-send invites to those with invites pending or declined.

showing an un-clickable "invited" on the "invite" button is something we did for libraries, now we'll only do that immediately after an invite has been sent (completely front-ended). If people complain about spamming, we can be more restrictive.


ryanpbrewster:
The code lgtm.

I think the product decision makes sense, although we should think about letting a user "block" invites from another user/org.

camhashemi:
not a bad idea, but probably not worth doing preemptively. we'll see if people actually get spammed regularly, until then it can be solved via support.

PR #16955, opened 2015-08-24
Added new KTL logic to KeepDecorator

ryanpbrewster:
This makes it so that the PageCommander can access multilib keeps (which in
turn makes them available to the extension).

@andrewconner @leogrim

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepDecorator.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepDecorator.scala
@@ -1,22 +1,31 @@
 package com.keepit.commanders
 
-import com.keepit.common.crypto.{ PublicIdConfiguration }
-import com.google.inject.{ Provider, Inject }
+import com.google.inject.{ ImplementedBy, Inject, Provider, Singleton }
+import com.keepit.common.core._
+import com.keepit.common.crypto.PublicIdConfiguration
 import com.keepit.common.db.Id
 import com.keepit.common.db.slick.Database
 import com.keepit.common.domain.DomainToNameMapper
 import com.keepit.common.healthcheck.AirbrakeNotifier
+import com.keepit.common.logging.Logging
 import com.keepit.common.social.BasicUserRepo
-import com.keepit.common.store.{ S3ImageConfig, ImageSize }
+import com.keepit.common.store.{ ImageSize, S3ImageConfig }
 import com.keepit.model._
 import com.keepit.rover.RoverServiceClient
 import com.keepit.search.SearchServiceClient
-import com.keepit.search.augmentation.{ LimitedAugmentationInfo, AugmentableItem }
+import com.keepit.search.augmentation.{ AugmentableItem, LimitedAugmentationInfo }
 
 import scala.concurrent.{ ExecutionContext, Future }
-import com.keepit.common.core._
 
-class KeepDecorator @Inject() (
+@ImplementedBy(classOf[KeepDecoratorImpl])
+trait KeepDecorator {
+  def decorateKeepsIntoKeepInfos(perspectiveUserIdOpt: Option[Id[User]], showPublishedLibraries: Boolean, keepsSeq: Seq[Keep], idealImageSize: ImageSize, withKeepTime: Boolean): Future[Seq[KeepInfo]]
+  def filterLibraries(infos: Seq[LimitedAugmentationInfo]): Seq[LimitedAugmentationInfo]
+  def getBasicKeeps(userId: Id[User], uriIds: Set[Id[NormalizedURI]], useMultilibLogic: Boolean = false): Map[Id[NormalizedURI], Set[BasicKeep]]
+}
+
+@Singleton
+class KeepDecoratorImpl @Inject() (
     db: Database,
     basicUserRepo: BasicUserRepo,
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepDecorator.scala
```diff
@@ -192,11 +204,19 @@ class KeepDecorator @Inject() (
...
-      val allKeeps = keepRepo.getByLibraryIdsAndUriIds(writeableLibs, uriIds)
-      allKeeps
+      val oldWay = keepRepo.getByLibraryIdsAndUriIds(writeableLibs, uriIds).toSet
+      val newWay = {
+        val direct = ktuRepo.getByUserIdAndUriIds(userId, uriIds).map(_.keepId)
+        val indirectViaLibraries = ktlRepo.getVisibileFirstOrderImplicitKeeps(uriIds, writeableLibs).map(_.keepId)
+        (direct ++ indirectViaLibraries) |> keepRepo.getByIds |> (_.values.toSet)
+      }
+      if (newWay.map(_.id.get) != oldWay.map(_.id.get)) {
+        log.error(s"[KTL-MATCH] getBasicKeeps($userId, $uriIds): ${newWay.map(_.id.get)} != ${oldWay.map(_.id.get)}")
```
leogrim, line 217: Maybe even airbrake for now?

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepDecorator.scala
```diff
@@ -192,11 +204,19 @@ class KeepDecorator @Inject() (
...
-  def getBasicKeeps(userId: Id[User], uriIds: Set[Id[NormalizedURI]]): Map[Id[NormalizedURI], Set[BasicKeep]] = {
+  def getBasicKeeps(userId: Id[User], uriIds: Set[Id[NormalizedURI]], useMultilibLogic: Boolean = false): Map[Id[NormalizedURI], Set[BasicKeep]] = {
     val allKeeps = db.readOnlyReplica { implicit session =>
       val writeableLibs = libraryMembershipRepo.getLibrariesWithWriteAccess(userId)
-      val allKeeps = keepRepo.getByLibraryIdsAndUriIds(writeableLibs, uriIds)
-      allKeeps
+      val oldWay = keepRepo.getByLibraryIdsAndUriIds(writeableLibs, uriIds).toSet
+      val newWay = {
+        val direct = ktuRepo.getByUserIdAndUriIds(userId, uriIds).map(_.keepId)
+        val indirectViaLibraries = ktlRepo.getVisibileFirstOrderImplicitKeeps(uriIds, writeableLibs).map(_.keepId)
```
leogrim, line 213: Typo in `getVisibileFirstOrderImplicitKeeps`. 
Is there a reason not to have a single super-join or `union` query to grab all keeps by `uriId` for a given user?
I guess one advantage is that you can have the direct ones at the top of the list here.

kifi-backend/modules/shoebox/app/com/keepit/model/KeepToLibraryRepo.scala
```diff
@@ -127,17 +127,12 @@ class KeepToLibraryRepoImpl @Inject() (
     getByUserIdAndLibraryIdHelper(userId, libraryId, excludeStateOpt).list
   }
 
-  def getVisibileFirstOrderImplicitKeeps(userId: Id[User], uriId: Id[NormalizedURI])(implicit session: RSession): Set[Id[Keep]] = {
+  def getVisibileFirstOrderImplicitKeeps(uriIds: Set[Id[NormalizedURI]], libraryIds: Set[Id[Library]])(implicit session: RSession): Set[KeepToLibrary] = {
```
leogrim, line 130: Why not just `getKeepsByUriIdsAndLibraryIds`?
Or `getKeepsByUrisInLibraries`?

(this method is viewer independent, so speaking of degree here doesn't make much sense, what do you think?)

ryanpbrewster, line 130: Those are both better names. The naming is a holdover from when I was doing
a monster join on org memberships as well.

On Mon, Aug 24, 2015 at 6:34 PM, Léo Grimaldi notifications@github.com
wrote:

> In
> kifi-backend/modules/shoebox/app/com/keepit/model/KeepToLibraryRepo.scala
> https://github.com/kifi/keepit/pull/16955#discussion_r37824778:
> 
> > @@ -127,17 +127,12 @@ class KeepToLibraryRepoImpl @Inject() (
> >      getByUserIdAndLibraryIdHelper(userId, libraryId, excludeStateOpt).list
> >    }
> > -  def getVisibileFirstOrderImplicitKeeps(userId: Id[User], uriId: Id[NormalizedURI])(implicit session: RSession): Set[Id[Keep]] = {
> > -  def getVisibileFirstOrderImplicitKeeps(uriIds: Set[Id[NormalizedURI]], libraryIds: Set[Id[Library]])(implicit session: RSession): Set[KeepToLibrary] = {
> 
> Why not just getKeepsByUriIdsAndLibraryIds?
> Or getKeepsByUrisInLibraries?
> 
> —
> https://github.com/kifi/keepit/pull/16955/files#r37824778.

## 

Software Engineer

PR #16958, opened 2015-08-25
OrgInviteCommander uses graph recos + prefixSearch of all users

camhashemi:
@leogrim @ryanpbrewster this depends on #16944 

all of the unique logic should be in `OrganizationInviteCommander.suggestMembers`


leogrim:
It looks like this one also has the `disclosePrivateEmails` stuff, is this intentional?

camhashemi:
Once I merge the other PRs and pull from master it'll be cleaner. I just needed those other ones to make this work. Not sure if there's a git way to making that process nicer.

camhashemi:
@leogrim @andrewconner @gratimax updated. the submodule hash is up to date.

kifi-backend/modules/shoebox/marketing
```diff
@@ -1 +1 @@
-Subproject commit d3dc4581840dbc8234e5ea2cd6da376fd04d1cd7
+Subproject commit 37fd67e9fd2be9869fe5498a8e945c83873cd988
```
ryanpbrewster, line 1: Necessary?

camhashemi, line 1: It's not reflective of a change in the submodule (same hash). This one is tough to remove (behind a few branch merges), but I'm going to work on keeping this out of my PRs.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
```diff
@@ -399,57 +401,40 @@ class OrganizationInviteCommanderImpl @Inject() (db: Database,
...
-        emailAddress -> (access, lastInvitedAt)
+        abookClient.getRecommendationsForOrg(orgId, Some(userId), offset = 0, limit = limit).map { orgInviteRecos =>
+          val userRecos = orgInviteRecos.collect { case reco if reco.identifier.isLeft => reco.identifier.left.get }
+          val emailRecos = orgInviteRecos.collect { case reco if reco.identifier.isRight => RichContact(email = reco.identifier.right.get, name = reco.name) }
+          (userRecos, emailRecos)
+        }
+      case Some(validQuery) =>
+        val memberCount = db.readOnlyMaster { implicit session => organizationMembershipRepo.countByOrgId(orgId) }
+        for {
+          users <- searchClient.searchUsersByName(userId, validQuery, limit + memberCount)
```
andrewconner: How about starting these futures outside the for-comprehension? They'll run in parallel

andrewconner:
Cool, :+1:

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
camhashemi: Just learned that there's a difference, thank you.

andrewconner: To help remember, a `for` comprehension is just syntactic sugar (`flatMap` + `map`); it can be rewritten mechanically:

```
for { a <- A; b <- B} yield a+b
// same as:
A.flatMap(a => B.map(b => a+b)) 
```

In that second form, you can see that if `B` happened to be a function call, it would happen _after_ `A` is finished resolving. To fix this, just call it outside:

```
val AF = A() // this returns a future that may take time
val BF = B() // call immediately!
for { a <- AF; b <- BF } yield a + b
```

PR #16959, opened 2015-08-25
Dropping Keep.urlId and URLMigrations

ryanpbrewster:
Keeps are going on a crash diet. Prepare yourselves.

@leogrim @andrewconner @stkem


andrewconner:
This is great. :+1:

kifi-backend/modules/shoebox/app/com/keepit/integrity/UriIntegrityPlugin.scala
```diff
@@ -133,29 +128,6 @@ class UriIntegrityActor @Inject() (
...
-  }
-
-  private def handleURLMigrationNoBookmarks(url: URL, newUriId: Id[NormalizedURI])(implicit session: RWSession): Unit = {
-    log.info(s"migrating url ${url.id} to new uri: $newUriId")
-
-    val oldUriId = url.normalizedUriId
-    urlRepo.save(url.withNormUriId(newUriId).withHistory(URLHistory(clock.now, oldUriId, URLHistoryCause.MIGRATED)))
-    updateNewUri(oldUriId, newUriId)
-  }
-
```
ryanpbrewster, line 158: Is this important at all? @andrewconner

PR #16960, opened 2015-08-25
Remove circular dependency and nested sessions

leogrim:
@camhashemi @andrewconner


kifi-backend/modules/shoebox/marketing
```diff
@@ -1 +1 @@
-Subproject commit 37fd67e9fd2be9869fe5498a8e945c83873cd988
+Subproject commit d3dc4581840dbc8234e5ea2cd6da376fd04d1cd7
```
camhashemi, line 1: this is going from the most recent hash to an older one

leogrim, line 1: ah crap

PR #16975, opened 2015-08-25
Sadism

ryanpbrewster:
@leogrim @andrewconner 

I added an extremely abusive test for URI integrity plugin. It takes ~8s to run. I'm keeping it in because I'm really tired of running into issues like this.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepCommander.scala
@@ -53,7 +53,7 @@ object BulkKeepSelection {
     (__ \ 'keeps).formatNullable[Seq[ExternalId[Keep]]] and
     (__ \ 'tag).formatNullable[ExternalId[Collection]] and
     (__ \ 'exclude).formatNullable[Seq[ExternalId[Keep]]]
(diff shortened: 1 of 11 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepCommander.scala
```diff
@@ -784,15 +785,46 @@ class KeepCommanderImpl @Inject() (
...
+    val connections = KeepConnections(
+      libraries = ktlRepo.getAllByKeepId(keepId).map(_.libraryId).toSet,
+      users = ktuRepo.getAllByKeepId(keepId).map(_.userId).toSet
+    )
+    connections == targetConnections
+  }
+  private def getKeepsByUriAndConnections(uriId: Id[NormalizedURI], targetConnections: KeepConnections)(implicit session: RSession): Set[Keep] = {
+    // keepRepo.getByUriAndConnectionsHash(uriId, targetConnections.hash).filter(k => isKeepConnectedTo(k.id.get, targetConnections))
+    require(targetConnections.libraries.size == 1, s"keep is not in exactly 1 library: ${targetConnections.libraries}!")
+    keepRepo.getPrimaryByUriAndLibrary(uriId, targetConnections.libraries.head).toSet
```
ryanpbrewster, line 798: This is here to preserve backwards compatibility until we drop the URI/library constraint.

PR #16977, opened 2015-08-25
Use the default picture if one does not exist for a push notif

maxov:
@andrewconner 
@camhashemi 
Kind of scary that this hasn't been caught and we've basically failed 12000 push notification sends...


camhashemi:
I'd like to put that `getOrElse(default)` behavior into a method in the `User` model. Something like `getPictureNameOrDefault`

PR #16979, opened 2015-08-25
return emails in typeahead even if they are attached to a user

camhashemi:
@leogrim @andrewconner 

If you type "camhashemi@gmail.com" (associated with my user), you should get that email in the typeahead, not nothing. Even if we want to filter out emails associated with users, we should instead provide that user in the typeahead (more complicated, possibly better). Or better yet, if you send a message to an email associated with a user, send the user the notif, not just the email (could do that too).


andrewconner:
Hm, interesting. How much of a pain would it be to instead return the user when you type the right email? @leogrim

leogrim:
Returning it shouldn't be too hard, we could index email addresses in Search or / and in the `User` typeahead. I'm not sure about the "instead" part, i.e. not showing the email address at all if it belongs to a user. Is this what we'd want though?

camhashemi:
While it would be nice to show the user, not showing the email address could be a point of confusion (the typeahead query doesn't explicitly match the results).

If we decide to show both, it would be another reason to send a message to a user when his/her email is a recipient, not just the email.

andrewconner:
> it would be another reason to send a message to a user when his/her email is a recipient, not just the email.

:+1: That's good.

PR #16982, opened 2015-08-25
Eliza notification integration

maxov:
Hopefully the last or next-to-last notification pr. Depends on #16943, actually integrates the new notification system within eliza (for those with the user experiment).


kifi-backend/modules/common/app/com/keepit/notify/info/DbView.scala
```diff
@@ -0,0 +1,130 @@
...
+    new DbView(keyMap + (key -> assocObjMap))
+  }
+
+  def update[M <: HasId[M]](existing: ExistingDbViewModel[M]): DbView = {
+    val model = existing.model
+    update(existing.key, model.id.get, model)
+  }
+
+  def lookup[M <: HasId[M], R](key: DbViewKey[M, R], id: Id[M]): R =
+    keyMap(key)(id).asInstanceOf[R]
```
stkem: :-(

maxov:
@leogrim 
@stkem 
@andrewconner 
@ryanpbrewster 

This isn't conceptually finished (about half of required methods are implemented), but I would like to merge it soon before it gets too big. I'll split this PR up into a bunch of smaller ones that have more encapsulated changes. Thoughts?

kifi-backend/modules/common/conf/evolutions/shoebox/390.sql
```diff
@@ -0,0 +1,21 @@
...
+
+# --- !Ups
+
+ALTER TABLE notification ADD COLUMN external_id varchar(36) DEFAULT NULL;
+
+ALTER TABLE notification_item ADD COLUMN external_id varchar(36) DEFAULT NULL;
+
+ALTER TABLE notification MODIFY COLUMN last_checked datetime NULL DEFAULT NULL;
+
+ALTER TABLE notification_item ADD COLUMN event_time datetime NOT NULL;
```
ryanpbrewster, line 11: :+1:

kifi-backend/modules/common/conf/evolutions/shoebox/390.sql
```diff
@@ -0,0 +1,21 @@
+# ELIZA
+
+# --- !Ups
+
+ALTER TABLE notification ADD COLUMN external_id varchar(36) DEFAULT NULL;
+
+ALTER TABLE notification_item ADD COLUMN external_id varchar(36) DEFAULT NULL;
+
+ALTER TABLE notification MODIFY COLUMN last_checked datetime NULL DEFAULT NULL;
```
ryanpbrewster: Shouldn't this just be `datetime DEFAULT NULL`? What's the extra `NULL` for?

kifi-backend/modules/eliza/app/com/keepit/eliza/commanders/NotificationMessagingCommander.scala
```diff
@@ -0,0 +1,132 @@
...
+  def notificationItemByExternalId(notifItemId: ExternalId[NotificationItem]): Option[NotificationItem] = {
+    db.readOnlyMaster { implicit session =>
+      notificationItemRepo.getOpt(notifItemId)
+    }
+  }
+
+  private def messageThreadIdForNotif(notif: Notification): ExternalId[MessageThread] = db.readOnlyMaster { implicit session =>
+    val userThread = userThreadRepo.getByNotificationId(notif.id.get)
+    // messaging analytics expects a message thread id, for now
+    userThread.fold(notif.externalId.asInstanceOf[ExternalId[MessageThread]]) { userThread =>
```
stkem: :-( Why is this cast needed?

kifi-backend/modules/eliza/app/com/keepit/eliza/commanders/NotificationMessagingCommander.scala
```diff
@@ -0,0 +1,132 @@
...
+  def setNotificationsUnreadBefore(notif: Notification, recipient: Recipient, item: NotificationItem): Unit = {
+    db.readWrite { implicit session =>
+      notificationRepo.setAllReadBefore(recipient, item.eventTime)
+    }
+    val unreadMessages = getUnreadEnabledNotificationsCountForKind(recipient, NewMessage.name)
+    val unreadNotifications = getUnreadEnabledNotificationsCountExceptKind(recipient, NewMessage.name)
+    recipient match {
+      case UserRecipient(user, _) =>
+        webSocketRouter.sendToUser(user, Json.arr("unread_notifications_count", unreadMessages + unreadNotifications, unreadMessages, unreadNotifications))
+        val pushNotif = MessageThreadPushNotification(notif.externalId.asInstanceOf[ExternalId[MessageThread]], unreadMessages + unreadNotifications, None, None)
```
stkem: same here, why do you need a cast?

kifi-backend/modules/eliza/app/com/keepit/notify/delivery/WsNotificationDelivery.scala
```diff
@@ -2,42 +2,40 @@ package com.keepit.notify.delivery
...
-    //          unread = false,
-    //          extra = info.extraJson
-    //        )
-    //      }
-    //      ()
-    //    }
-    Future.successful(())
+  def deliver(recipient: Recipient, notif: Notification, items: Set[NotificationItem]): Future[Unit] = {
+    val events = items.map(_.event)
+    val kind = events.head.kind.asInstanceOf[NotificationKind[NotificationEvent]]
```
stkem: What is the actual compile time type of `NotificationItem.event.kind`?

kifi-backend/modules/common/conf/evolutions/shoebox/390.sql
maxov: Technically you're right, having `DEFAULT NULL` implies a `NULL` column, I just wanted to be more verbose here because I was having problems locally.

kifi-backend/modules/eliza/app/com/keepit/eliza/commanders/NotificationMessagingCommander.scala
maxov: Messaging analytics expects an external id of a message thread for a notification, but most notifications aren't associated with message threads. This sends the notification external ID, faking it as the message thread. It could possibly mess up analytics, but from what I see nothing is inferred from the external ID of the message thread. In the future I'll change the signature to take a notification external ID once the legacy system is deprecated.

kifi-backend/modules/eliza/app/com/keepit/eliza/commanders/NotificationMessagingCommander.scala
maxov: MessageThreadPushNotification also expects an external ID of a MessageThread, but notification external IDs are fully compatible here. Also same as previous, in the future I'll change the signature to take a notification external ID once the legacy system is deprecated.

kifi-backend/modules/eliza/app/com/keepit/notify/delivery/WsNotificationDelivery.scala
maxov: It's existential, `NotificationKind[_ <: NotificationEvent]`. NotificationItem has no type parameter for the event type since it is persisted in the database as a model, and I wasn't very keen on parameterizing it. I guess it's possible to parameterize Notification and NotificationItem, but then the casts would simply happen at a lower level.

kifi-backend/modules/eliza/app/com/keepit/eliza/commanders/NotificationMessagingCommander.scala
stkem: You don't need a cast for that. Just use `ExternalId[MessageThread](notif.externalId.id)` (If it's needed often you could bake it into a method on `ExternalId[T]`). That way, if the type of `notif.externalId` every changes, wether just structurally or not, you are likely to get an error at compile time rather than runtime.
There is really no place for casts in pure scala code.

kifi-backend/modules/eliza/app/com/keepit/eliza/commanders/NotificationMessagingCommander.scala
stkem: See comment on the other case above. You don't need a cast for this.

kifi-backend/modules/eliza/app/com/keepit/eliza/commanders/NotificationMessagingCommander.scala
maxov: You're totally right, that makes much more sense than casting. Yeah, I should really be very careful and not cast unless it's 100% totally needed, like in this case. I will definitely be more careful with that next time, thanks

kifi-backend/modules/eliza/app/com/keepit/notify/delivery/WsNotificationDelivery.scala
stkem: So basically what you are assuming here is that `NotificationKind[T]` is covariant in `T`, so that any `NotificationKind[_ <: NotificationEvent]` is a `NotificationKind[NotificationEvent]`. You should be able to make that explicit in the definition, i.e. `trait NotificationKind[+N <: NotificationEvent]`, then the compiler will know without the cast.
The need for casts in pure scala code almost always is an indicator of a bug in your typing, as in this case with the missing covariance marker. (And it is in fact possible that `NotificationKind[T]` is in fact _not_ covariant in `T`, which the compiler will tell you if you tell it that that is your assumption)

stkem: Ah, I just tried this and `NotificationKind[T]` is indeed **not** covariant! (i.e. the cast produces an non consistent state which could easily cause error when methods are called on the resulting object later on)

maxov: I wanted to make that type parameter covariant, but I had some issues because of the body of NotificationKind:

``` scala
trait NotificationKind[+N <: NotificationEvent] {

  implicit val format: Format[N] // <- compile error here because Format is invariant in N

  def shouldGroupWith(newEvent: N, existingEvents: Set[N]): Boolean // <- compile error here because N is used in a contravariant position
}
```

I could try to stick some additional type parameters in some places(like format) and that would hopefully fix the problem, like 

``` scala
trait NotificationKind[+N <: NotificationEvent] {
  implicit def format[T >: N]: Format[T] 
  // ...
}
```

stkem: If that works, go for it. Rule of thumb: If you need a cast to get around a compiler type error (as opposed to forcing previously serialized data into a particular type, the other big scenario where one might want to use casts), either your code has a bug or the compiler does.

stkem:
:+1:

PR #16986, opened 2015-08-25
account for user.state = incomplete_signup in a few more places

camhashemi:
@andrewconner _might_ be solving the problem in `KifiSiteRouter`. I can add a guard on auto-joining, but I think it's best to throw an airbrake so we can figure out when/why this is happening.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/core/AuthHelper.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/core/AuthHelper.scala
@@ -377,6 +377,7 @@ class AuthHelper @Inject() (
     implicit val context = heimdalContextBuilder.withRequestInfo(request).build
     val (user, emailPassIdentity) = authCommander.finalizeSocialAccount(sfi, identity, inviteExtIdOpt)
     val emailConfirmedBySocialNetwork = identity.email.map(EmailAddress.validate).collect { case Success(validEmail) => validEmail.copy(address = validEmail.address.trim) }.exists(_.equalsIgnoreCase(sfi.email))
+    if (user.state == UserStates.INCOMPLETE_SIGNUP) airbrake.notify(s"[finalizeSocialAccount] calling finishSignup without an active user! ${user.toString}")
     finishSignup(user, sfi.email, emailPassIdentity, emailConfirmedAlready = emailConfirmedBySocialNetwork, libraryPublicId = libraryPublicId, libAuthToken = libAuthToken,
       orgPublicId = orgPublicId, orgAuthToken = orgAuthToken, isFinalizedImmediately = isFinalizedImmediately)
   }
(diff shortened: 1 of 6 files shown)


kifi-backend/modules/shoebox/app/com/keepit/controllers/core/AuthHelper.scala
```diff
@@ -410,6 +411,7 @@ class AuthHelper @Inject() (
     authCommander.finalizeEmailPassAccount(efi, request.userId, request.user.externalId, request.identityOpt, inviteExtIdOpt).map {
       case (user, email, newIdentity) =>
         val verifiedEmail = verifySignupEmail(request.userId, email, libraryPublicId, libAuthToken)
+        if (user.state == UserStates.INCOMPLETE_SIGNUP) airbrake.notify(s"[finalizeEmailPassAccount] calling finishSignup without an active user! ${user.toString}")
         finishSignup(user, email, newIdentity, emailConfirmedAlready = verifiedEmail, libraryPublicId = libraryPublicId, libAuthToken = libAuthToken, orgPublicId, orgAuthToken, isFinalizedImmediately = false)
```
leogrim, line 413: This state might be normal at this point, before `finishSignup` @andrewconner ?

camhashemi, line 413: We had an org membership where the user's state was `incomplete_signup`, which caused bugs, so this is an attempt to figure out how people are being created with such a state. There's a number of them keeping, joining libraries, etc.

camhashemi, line 413: @leogrim you should be right that by this point things would be good. I had assumed that the user was auto-joined into the org, but I see now that he got a user-invitation. So I can remove this log, the issue isn't that we're auto-joining people with incomplete signup, it's that we're letting them do anything at all.

andrewconner, line 413: > this is an attempt to figure out how people are being created with such a state

So sorry! I wasn't clear earlier. I know how it happens (or at least, how it used to happen?), those users never go through this code path, so likely are broken in other ways as well. It happens if a user abandons registration mid-signup using email, and goes straight to kifi.com. Your other fix is solid, it'll do it.

camhashemi, line 413: I just realized that, will take out the extra stuff.

camhashemi:
@andrewconner another thing I added was to use `user.state != ACTIVE` instead of `user.state == INACTIVE`, since there are more non-active states than `INACTIVE`. This could lead to broken experiences for users with non-active states (like the user who was a member of the org but didn't show up). Is that the right way to go about it in general? Or let them slide?

andrewconner:
So it looks like sometime, BLOCKED stopped being honored. Sigh. Yeah you can redirect if `user.state != ACTIVE`, but for blocked/inactive, we probably should delete their cookies (log them out).

camhashemi:
I'll add that logic (assuming that they won't be allowed to log back in), but I was referring to backend processing more than redirect logic.

 We take for granted that usually `state = active || inactive`. With users should we try to use `state != active` instead of `state == inactive`? this could break things for these non-active users.

PR #16987, opened 2015-08-25
basic client apis for paid plans

stkem:
@andrewconner @ryanpbrewster 
(any frontenders that should be tagged in this?)


kifi-backend/modules/common/conf/evolutions/shoebox/375.sql
```diff
@@ -38,7 +38,7 @@ CREATE TABLE paid_plan (
     state varchar(20) NOT NULL,
     name varchar(256) NOT NULL,
     billing_cycle int(11) NOT NULL,
-    price_per_cycle_per_user int(11) NOT NULL,
```
stkem, line 41: typo, correct in the prod db

kifi-backend/modules/common/conf/evolutions/shoebox/376.sql
```diff
@@ -12,7 +12,7 @@ alter table account_event
 alter table account_event
     add column processing_stage VARCHAR(20) NOT NULL;
 
-alter table paid_account
```
stkem, line 15: typo, correct in the prod db

kifi-backend/modules/shoebox/conf/shoeboxService.routes
```diff
@@ -541,6 +541,17 @@ GET           /site/user-or-org/:handle                   @com.keepit.controller
 GET           /site/user-or-org/:handle/libraries         @com.keepit.controllers.website.UserOrOrganizationController.getLibrariesByHandle(handle: Handle, page: Int ?= 0, size: Int ?= 12, filter: String ?= "own")
 GET           /site/user-or-org/:handle/libraries/:slug   @com.keepit.controllers.website.LibraryController.getLibraryByHandleAndSlug(handle: Handle, slug: LibrarySlug, authToken: Option[String] ?= None)
 
+GET           /site/billing/state                         @com.keepit.controllers.website.PaymentsController.getAccountState(pubId: PublicId[Organization])
```
andrewconner: Should we scope this to something like `/site/admin` or something? I bet there will be a lot of similar routes that aren't technically billing, but administrative.

stkem: Yeah, maybe. I'm not super happy with the `billing` term either. Are you suggesting to put these under `/site/admin` or `/site/admin/billing`?

andrewconner: I'd err towards the latter.

stkem: I like that. Changing it.

andrewconner:
More in depth code review incoming, hoping on a bike and my code review skills are not great while biking.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/PaymentsController.scala
```diff
@@ -0,0 +1,130 @@
...
+
+  def setAccountContacts(pubId: PublicId[Organization]) = OrganizationUserAction(pubId, PLAN_MANAGEMENT_PERMISSION)(parse.tolerantJson) { request =>
+    request.body.asOpt[Seq[SimpleAccountContactSettingRequest]].map { contacts =>
+      val attribution = ActionAttribution(user = Some(request.request.userId), admin = request.request.adminUserId)
+      contacts.foreach { contact =>
+        planCommander.updateUserContact(request.orgId, contact.id, contact.enabled, attribution)
+      }
+      Ok(Json.toJson(planCommander.getSimpleContactInfos(request.orgId)))
+    } getOrElse {
+      BadRequest
```
ryanpbrewster: Can we do a `BadRequest(Json.obj("error" -> "could_not_parse"))` or something? It helps frontend pin down exactly what's wrong when they're tying in logic.

Also, I'm not sure how you feel about it but I like airbraking when we get a badly formatted request, so that in case we break a client API in the future we'll know about it.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/PaymentsController.scala
```diff
@@ -0,0 +1,130 @@
...
+    }
+  }
+
+  def getPlanFeatureSettings(pubId: PublicId[Organization]) = OrganizationUserAction(pubId, PLAN_MANAGEMENT_PERMISSION) { request => //ZZZ TODO: This is currently a dummy
+    val dummyFeatureOne = PlanFeature(name = "keeping", displayName = "Users can keep things", editable = false, default = true)
+    val dummyFeatureTwo = PlanFeature(name = "messaging", displayName = "Users can send messages", editable = true, default = true)
+    Ok(Json.toJson((Seq(
```
ryanpbrewster: I see this is a dummy, but when you implement it I prefer `JsObject` to `JsArray` as a top-level return type. Easier to maintain backwards compatibility.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/PaymentsController.scala
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  @json
+  case class SimplePlanFeaureSettingRequest(name: String, enabled: Boolean)
+
+  def setPlanFeatureSettings(pubId: PublicId[Organization]) = OrganizationUserAction(pubId, PLAN_MANAGEMENT_PERMISSION)(parse.tolerantJson) { request => //ZZZ TODO: This is currently a dummy (just does request format validation)
+    request.body.asOpt[Seq[SimplePlanFeaureSettingRequest]].map { settings =>
+      Ok
+    } getOrElse {
+      BadRequest
```
ryanpbrewster: Same as above, maybe a `"could_not_parse"` or something. If you use `request.body.validate` instead of `asOpt` you can even trace down exactly what was wrong with the request payload and airbrake that. It's sort of shorter, too.

```
request.body.validate[Seq[SimplePlanFeatureSettingRequest]] match {
    case JsError(errs) => BadRequest(Json.obj("error" -> "could_not_parse_request"))
    case JsSuccess(settings, _) => f(settings)
}
```

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/PaymentsController.scala
```diff
@@ -0,0 +1,130 @@
...
+        }
+
+      }
+      case Failure(ex) => BadRequest("invalid_plan_id")
+    }
+  }
+
+  def getEvents(pubId: PublicId[Organization], limit: Int) = OrganizationUserAction(pubId, PLAN_MANAGEMENT_PERMISSION) { request =>
+    val infos = planCommander.getAccountEvents(request.orgId, limit, onlyRelatedToBillingFilter = None).map(planCommander.buildSimpleEventInfo)
+    Ok(Json.toJson(infos))
```
ryanpbrewster: `Json.obj` instead?

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/PaymentsController.scala
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  def getEventsBefore(pubId: PublicId[Organization], limit: Int, beforeTime: DateTime, beforePubId: PublicId[AccountEvent]) = OrganizationUserAction(pubId, PLAN_MANAGEMENT_PERMISSION) { request =>
+    AccountEvent.decodePublicId(beforePubId) match {
+      case Success(beforeId) => {
+        val infos = planCommander.getAccountEventsBefore(request.orgId, beforeTime, beforeId, limit, onlyRelatedToBillingFilter = None).map(planCommander.buildSimpleEventInfo)
+        Ok(Json.toJson(infos))
```
ryanpbrewster: and here

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationMembershipRepo.scala
```diff
@@ -193,6 +194,15 @@ class OrganizationMembershipRepoImpl @Inject() (
   def getTeammates(userId: Id[User])(implicit session: RSession): Set[Id[User]] = {
     getTeammatesCompiled(userId).run.toSet
   }
+
+  def getByRole(orgId: Id[Organization], role: OrganizationRole, excludeState: Option[State[OrganizationMembership]] = Some(OrganizationMembershipStates.INACTIVE))(implicit session: RSession): Seq[Id[User]] = {
+    excludeState match {
+      case Some(exclude) => (for { row <- rows if row.organizationId === orgId && row.state =!= exclude && row.role == role } yield row.userId).list
+      case None => (for { row <- rows if row.organizationId === orgId && row.role == role } yield row.userId).list
+    }
```
ryanpbrewster, line 202: Not sure how much effort it's worth, but you can do neat Slick things like
`rows.filter(row => row.organizationId === orgId && row.role === role && row.state =!= excludeState.orNull)`

kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEvent.scala
```diff
@@ -14,6 +14,16 @@ import org.joda.time.DateTime
 
 import javax.crypto.spec.IvParameterSpec
 
+@json
```
ryanpbrewster, line 17: Does this get passed around between internal services? If not, I'm scared of the `@json` macro. It lets people accidentally send internal information to clients.

ryanpbrewster, line 17: Oh, it has no internal information. :+1: you're awesome.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/PaymentsController.scala
andrewconner: :+1:

kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEventRepo.scala
```diff
@@ -80,11 +80,13 @@ class AccountEventRepoImpl @Inject() (
 
   def getEventsBefore(accountId: Id[PaidAccount], before: DateTime, limit: Int, onlyRelatedToBillingOpt: Option[Boolean])(implicit session: RSession): Seq[AccountEvent] = {
     val accountEvents = rows.filter(row => row.accountId === accountId && row.eventTime < before && row.state =!= AccountEventStates.INACTIVE)
+  def getEventsBefore(accountId: Id[PaidAccount], beforeTime: DateTime, beforeId: Id[AccountEvent], limit: Int, onlyRelatedToBillingOpt: Option[Boolean])(implicit session: RSession): Seq[AccountEvent] = {
+    val accountEvents = rows.filter(row => row.accountId === accountId && (row.eventTime < beforeTime || (row.eventTime === beforeTime && row.id < beforeId)) && row.state =!= AccountEventStates.INACTIVE)
     val relevantEvents = onlyRelatedToBillingOpt match {
       case Some(onlyRelatedToBilling) => accountEvents.filter(_.billingRelated === onlyRelatedToBilling)
       case None => accountEvents
     }
```
ryanpbrewster, line 88: Now that I look at this again, can you do the cool `row.billingRelated =!= onlyRelatedToBillingOpt.orNull` trick here? I guess you'd use it sort of like `excludeStateOpt`.

kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
```diff
@@ -338,13 +386,85 @@ class PlanManagementCommanderImpl @Inject() (
...
+      }
     }
+    SimpleAccountEventInfo(
+      id = AccountEvent.publicId(event.id.get),
+      eventTime = event.eventTime,
+      shortName = shortName,
+      whoDunnit = whoDunnit,
+      creditChange = event.creditChange.cents,
+      paymentCharge = event.paymentCharge.map(_.cents).getOrElse(0),
+      memo = event.memo.getOrElse("")
```
ryanpbrewster: The fact that we do this `getOrElse("")` implies to me that `memo` should be a `writeNullable` field.

kifi-backend/modules/shoebox/app/com/keepit/payments/PlansCommander.scala
```diff
@@ -232,6 +253,27 @@ class PlanManagementCommanderImpl @Inject() (
     (account.userContacts, account.emailContacts)
   }
 
+  def getSimpleContactInfos(orgId: Id[Organization]): Seq[SimpleAccountContactInfo] = db.readOnlyMaster { implicit session =>
+    val account = paidAccountRepo.getByOrgId(orgId)
+    val registered: Map[Id[User], Boolean] = account.userContacts.toSet.map { userId: Id[User] => (userId, true) }.toMap
+    val potential: Map[Id[User], Boolean] = orgMembershipRepo.getByRole(orgId, OrganizationRole.ADMIN).toSet.map { userId: Id[User] => (userId, true) }.toMap
+    val all = potential ++ registered
```
ryanpbrewster, line 269: Won't `registered` have overlapping keys with `potential`? The second map in a `++` overwrites the first map when there's a key collision.

ryanpbrewster, line 269: It looks like you are imposing the value `true` for all of the keys in the map. Is there a reason you wrote it like this instead of something like

```
(account.userContacts.toSet ++ orgMembershipRepo.getByRole(orgId, ADMIN)).map { userId =>
    SimpleAccountContactInfo(bus(userId), enabled = true)
}
```

stkem, line 269: Oh, this is actually a typo. It should be `false` for `potential`. The overwriting of fields in the map is intended. (And hopefully makes more sense now)

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationMembershipRepo.scala
stkem, line 202: It does make the code shorter, but you end up running a bunch of queries with `WHERE .... AND state != NULL` in them, which is a bit unnecessary. I don't really feel strongly about this, but slicks queries are bad enough as it is.

stkem:
Made a bunch of updates based on comments and also added simple very verbose backfilling logic + some logging (trying to track down a rather elusive bug where account events are not being persisted on Account creation. Accounts themselves are persisted just fine.

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminPaymentsController.scala
```diff
@@ -0,0 +1,73 @@
...
+      }.filter(_.state == OrganizationStates.ACTIVE)
+      channel.push(s"All ${orgs.size} organizations loaded\n")
+      orgs.foreach { org =>
+        channel.push("----------------------------------------\n")
+        channel.push(s"Processing org ${org.id.get}: ${org.name}\n")
+        db.readWrite { implicit session =>
+          paidAccountRepo.maybeGetByOrgId(org.id.get) match {
+            case Some(_) => {
+              channel.push(s"Paid account already exists. Doing nothing.\n")
+              Thread.sleep(200)
```
andrewconner: how come?

stkem: So as not to hammer the database during the backfill.

andrewconner: Should be fine, you can use readWriteBatch if you want commits along the way. This is inside a db session anyways, probably shouldn't block inside that thread.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/PaymentsController.scala
```diff
@@ -0,0 +1,137 @@
...
+    val orgInviteCommander: OrganizationInviteCommander,
+    val userActionsHelper: UserActionsHelper,
+    planCommander: PlanManagementCommander,
+    implicit val publicIdConfig: PublicIdConfiguration) extends UserActions with OrganizationAccessActions with ShoeboxServiceController {
+
+  private val PLAN_MANAGEMENT_PERMISSION = OrganizationPermission.EDIT_ORGANIZATION
+
+  def getAccountState(pubId: PublicId[Organization]) = OrganizationUserAction(pubId, PLAN_MANAGEMENT_PERMISSION) { request =>
+    Ok(Json.obj(
+      "credit" -> planCommander.getCurrentCredit(request.orgId).cents,
```
andrewconner, line 34: Can we serialize `DollarAmount` directly as a decimal? Frontend won't do any math.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/PaymentsController.scala
```diff
@@ -0,0 +1,137 @@
...
+      ))
+    } getOrElse {
+      Ok(Json.obj())
+    }
+  }
+
+  def setCreditCardToken(pubId: PublicId[Organization]) = OrganizationUserAction(pubId, PLAN_MANAGEMENT_PERMISSION)(parse.tolerantJson) { request =>
+    (request.body \ "token").asOpt[String] match {
+      case Some(token) => { //ZZZ TODO: Need to exchange this one time token for a permanent one with stripe
+        val attribution = ActionAttribution(user = Some(request.request.userId), admin = request.request.adminUserId)
```
andrewconner, line 53: Looks like `ActionAttribution` will often be able to be generated from the request directly? You could pimp `request` perhaps.

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminPaymentsController.scala
stkem: Ah, you are right. I'll move it out of the db session. Throttling this is still prudent though (and harmless).

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/PaymentsController.scala
```diff
@@ -0,0 +1,137 @@
...
+    }
+  }
+
+  def updatePlan(pubId: PublicId[Organization], planPubId: PublicId[PaidPlan]) = OrganizationUserAction(pubId, PLAN_MANAGEMENT_PERMISSION) { request => //ZZZ TODO: depending on what we decide to do with people who downgrade their plan this might have to do a lot more
+    PaidPlan.decodePublicId(planPubId) match {
+      case Success(planId) => {
+        val attribution = ActionAttribution(user = Some(request.request.userId), admin = request.request.adminUserId)
+        planCommander.changePlan(request.orgId, planId, attribution) match {
+          case Success(_) => Ok(Json.toJson(planCommander.currentPlan(request.orgId).asInfo))
+          case Failure(ex) => BadRequest(ex.getMessage)
```
andrewconner: Mind returning JSON?

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/PaymentsController.scala
```diff
@@ -0,0 +1,137 @@
...
+        }
+
+      }
+      case Failure(ex) => BadRequest("invalid_plan_id")
```
andrewconner: Same, stick to returning JSON?

andrewconner:
Few comments, :+1:

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/PaymentsController.scala
stkem, line 34: Not really about the math, but rather to leave the frontend to freedom to display thing however they like (like a different color depending on if it's positive or negative, or dollars being bigger than cents, stuff like that).
And sending it down as a `double`, because plays json serializer has some issues with floating point numbers. They get converted into some internal representation and then into a JSON Number literal, which does weird things. I've run into this with Recos, where a score like `1.33` turned into `1.3300000000000000000001` in the JSON in the db (I'm making up the exact numbers, but it was in that spirit)

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/PaymentsController.scala
stkem, line 53: Looking into that for the next PR on this (considering a small refactor of how ActionAttribution works anyway)

PR #16990, opened 2015-08-25
Use cached thread pool in dev mode

maxov:
Some tasks were not completing for me because of the limit on threads in dev mode. There are couple awaits that happen within nested futures, meaning that both threads in the pool were taken up awaiting for a task that could never be scheduled. This raises the number of threads in the thread pool (to essentially infinity) so that more tasks can execute. The better fix would be to not use await inside a whole bunch of nested futures, but this does not change any code that actually executes in production.

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/common/concurrent/WatchableExecutionContext.scala
+++ b/kifi-backend/modules/common/app/com/keepit/common/concurrent/WatchableExecutionContext.scala
@@ -1,6 +1,6 @@
 package com.keepit.common.concurrent
 
-import java.util.concurrent.{ LinkedBlockingQueue, ThreadPoolExecutor, TimeUnit }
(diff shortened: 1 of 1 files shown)


maxov:
@eishay
@andrewconner

andrewconner:
Looks good to me. Your assessment is right about `Await`. :(

andrewconner:
Where's the biggest offenders for you? I'm about to remove a couple in eliza.

maxov:
@andrewconner 

https://github.com/kifi/keepit/blob/4a76a9d3b4728b07ec3bbe699fde6ff11a8aad4c/kifi-backend/modules/eliza/app/com/keepit/eliza/commanders/MessagingCommander.scala#L213
https://github.com/kifi/keepit/blob/4a76a9d3b4728b07ec3bbe699fde6ff11a8aad4c/kifi-backend/modules/eliza/app/com/keepit/eliza/commanders/MessagingCommander.scala#L327

Those guys are the ones I have come across so far in testing.

PR #16992, opened 2015-08-25
Feature/activity feed

cvializ:
@andrewconner (@adamjgrant) https://app.asana.com/0/42816489233171/44164549287817


cvializ:
screen shot 2015-08-25 at 2 42 21 pm

cvializ:
@andrewconner :8ball: I see a PR review in our future

kifi-backend/modules/shoebox/angular/src/route.js
```diff
@@ -26,6 +26,26 @@ angular.module('kifi')
     $stateProvider
       .state('home', {  // Home page.
         url: '/',
+        controller: ['$state', 'me', function ($state, me) {
+          var showFeed = (me.orgs.length > 0);
```
andrewconner: Could we deploy to ourselves first (check experiments)? Would let us make sure everything's good before switching some new users over.

kifi-backend/modules/shoebox/angular/dev.html
```diff
@@ -14,6 +14,15 @@
 <div class="kf-app ng-cloak" ng-controller="AppCtrl">
   <div ng-if="$root.userLoggedIn != null" ng-include="'layout/body.tpl.html'"></div>
 </div>
+
+<script>
+  // Uncomment this to see micrologger reports in the console.
```
andrewconner, line 18: Should this be commented then?

kifi-backend/modules/shoebox/angular/src/feed/feedService.js
```diff
@@ -0,0 +1,26 @@
...
+
+angular.module('kifi')
+
+.factory('feedService', [
+  '$http', '$rootScope', 'profileService', 'routeService', '$q', '$analytics', 'net', 'ml',
+  function ($http, $rootScope, profileService, routeService, $q, $analytics, net, ml) {
+
+    var api = {
+      getFeed: function (limit, beforeId, afterId) {
+        ml.specs.getsFeed = new ml.Spec([
```
andrewconner, line 11: Interesting. I saw @adamjgrant's commit about this.

kifi-backend/modules/shoebox/angular/dev.html
cvializ, line 18: This is in `dev.html`, I think @adamjgrant intended for the comment to be in `index.html`

andrewconner:
:+1: Looks good! I know soon it'll be for all orgs-users, but would you like to do a smaller set (just us) first?

kifi-backend/modules/shoebox/angular/src/route.js
cvializ: Naturally. Should I create a new experiment, or should I just make it visible to 'fake' and 'admin'?

kifi-backend/modules/shoebox/angular/dev.html
andrewconner, line 18: Oh okay. Do we do anything in index.html?

kifi-backend/modules/shoebox/angular/src/route.js
andrewconner: The latter sounds good for me, unless you want it smaller than just us?

cvializ: Awesome, just us is fine, but I wanted to double check

kifi-backend/modules/shoebox/angular/dev.html
cvializ, line 18: Not yet, I'll add it in, but commented

PR #16995, opened 2015-08-25
don't use .maxBy on an empty collection

camhashemi:
@andrewconner @leogrim didn't know that .maxBy throws an error when the map is empty

### DIFF
--- a/kifi-backend/modules/heimdal/app/com/keepit/commander/HelpRankCommander.scala
+++ b/kifi-backend/modules/heimdal/app/com/keepit/commander/HelpRankCommander.scala
@@ -158,9 +158,10 @@ class HelpRankCommander @Inject() (
     }
   }
 
-  def getUserWithMostClickedKeeps(userIds: Set[Id[User]]): Id[User] = {
+  def getUserWithMostClickedKeeps(userIds: Set[Id[User]]): Option[Id[User]] = {
     db.readOnlyReplica { implicit session =>
-      keepDiscoveryRepo.getDiscoveryCountsByKeeper(userIds).maxBy(_._2)._1
+      val keepClicksByKeeper = keepDiscoveryRepo.getDiscoveryCountsByKeeper(userIds)
+      if (keepClicksByKeeper.nonEmpty) Some(keepClicksByKeeper.maxBy(_._2)._1) else None // users may not have any keep clicks
     }
   }
 
--- a/kifi-backend/modules/heimdal/app/com/keepit/heimdal/PrimaryOrgProvider.scala
+++ b/kifi-backend/modules/heimdal/app/com/keepit/heimdal/PrimaryOrgProvider.scala
@@ -54,7 +54,7 @@ class PrimaryOrgProviderImpl @Inject() (
         "inviteCount" -> ContextDoubleData(shoeboxValues.inviteCount),
         "collabLibCount" -> ContextDoubleData(shoeboxValues.collabLibCount),
         "messageCount" -> ContextDoubleData(messageCount),
-        "overallKeepViews" -> ContextStringData(userWithMostClickedKeeps.toString)
+        "overallKeepViews" -> ContextStringData(userWithMostClickedKeeps.map(_.toString).getOrElse(""))
       )
     }
   }


kifi-backend/modules/heimdal/app/com/keepit/heimdal/PrimaryOrgProvider.scala
```diff
@@ -54,7 +54,7 @@ class PrimaryOrgProviderImpl @Inject() (
         "inviteCount" -> ContextDoubleData(shoeboxValues.inviteCount),
         "collabLibCount" -> ContextDoubleData(shoeboxValues.collabLibCount),
         "messageCount" -> ContextDoubleData(messageCount),
-        "overallKeepViews" -> ContextStringData(userWithMostClickedKeeps.toString)
+        "overallKeepViews" -> ContextStringData(userWithMostClickedKeeps.map(_.toString).getOrElse(""))
```
andrewconner, line 57: Shouldn't this be a number? `0` perhaps?

camhashemi, line 57: This is per https://app.asana.com/0/32728208409723/44164549287874. Since it's new, it won't be trouble to change the name.

camhashemi, line 57: I think the name is as such to be distinguished from `userWithMostClickedKeeps` within an org (we're not fetching that yet). This one is system-wide keep clicks, hence `overallKeepViews`.

andrewconner, line 57: I trust you, still don't follow that this is for / does. Why would we include a blank string?

camhashemi, line 57: To denote the value is undefined, but it seems like Jen prefers omitting it. I'll do a refactor.

leogrim, line 57: Agreed, seems like it should be a `ContextDoubleData`, otherwise you won't be able to use it as a number for segmentation in Mixpanel / Amplitude.

camhashemi, line 57: The code I refactored in `AmplitudeClient` (also in `MixpanelClient.track`) uses `ContextStringData` for `orgId`, so I figured that's how we deal with ids.

leogrim, line 57: This is not an identifier though, it's a count.

andrewconner, line 57: Either way, is sending `""` is the appropriate way to deal with the lack of something?

leogrim, line 57: Ah no it's not a count, it's an id, my bad read this too quickly on my phone.
So a `String` is indeed what you want, but no we shouldn't send an empty `String`, just leave it out undefined.

camhashemi, line 57: Right, going to leave it out. Thanks guys.

kifi-backend/modules/heimdal/app/com/keepit/commander/HelpRankCommander.scala
```diff
@@ -158,9 +158,10 @@ class HelpRankCommander @Inject() (
     }
   }
 
-  def getUserWithMostClickedKeeps(userIds: Set[Id[User]]): Id[User] = {
+  def getUserWithMostClickedKeeps(userIds: Set[Id[User]]): Option[Id[User]] = {
     db.readOnlyReplica { implicit session =>
-      keepDiscoveryRepo.getDiscoveryCountsByKeeper(userIds).maxBy(_._2)._1
+      val keepClicksByKeeper = keepDiscoveryRepo.getDiscoveryCountsByKeeper(userIds)
+      if (keepClicksByKeeper.nonEmpty) Some(keepClicksByKeeper.maxBy(_._2)._1) else None // users may not have any keep clicks
```
ryanpbrewster, line 164: @camhashemi I know it's a bit late, but there is a function called `maxByOpt` that does exactly this. Look in `com.keepit.common.TraversableOnceExtensions`.

camhashemi, line 164: awesome, thanks for that

PR #16996, opened 2015-08-25
it really sucks when we swallow exceptions!

eishay:
@ryanpbrewster @camhashemi 
Hiding the real cause and just printing the message does not tell me anything about what the root problem of the exception is and its frustrating since its hard to find the error message in the thousands of log lines.
If we print an exception we should get the full stack as well since many has some root cause exception (this one was a db integrity problem), once i printed it the solution was fixed in seconds but finding that exception took way too long (exception cause was `org.h2.jdbc.JdbcSQLException: Referential integrity constraint violation: "ORGANIZATION_F_USER: PUBLIC.ORGANIZATION FOREIGN KEY(OWNER_ID) REFERENCES PUBLIC.USER(ID) (97543)"`).
Whenever we have a system exception we should fail fast and get as much context as possible, when its a user error its fine to just report it in a lower logging intensity.
/cc @andrewconner @leogrim @stkem

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
@@ -211,7 +211,8 @@ class OrganizationCommanderImpl @Inject() (
     Try {
       db.readWrite { implicit session =>
         getValidationError(request) match {
-          case Some(fail) => Left(fail)
+          case Some(fail) =>
+            Left(fail)
           case None =>
             val orgSkeleton = Organization(ownerId = request.requesterId, name = request.initialValues.name, primaryHandle = None, description = None, site = None)
             val orgTemplate = organizationWithModifications(orgSkeleton, request.initialValues.asOrganizationModifications)
@@ -225,9 +226,13 @@ class OrganizationCommanderImpl @Inject() (
         }
       }
     } match {
-      case Success(Left(fail)) => Left(fail)
-      case Success(Right(response)) => Right(response)
-      case Failure(ex) => log.error(ex.getMessage); Left(OrganizationFail.HANDLE_UNAVAILABLE)
+      case Success(Left(fail)) =>
+        Left(fail)
+      case Success(Right(response)) =>
+        Right(response)
+      case Failure(ex) =>
+        log.error(s"could not create organization via $request", ex)
+        Left(OrganizationFail.HANDLE_UNAVAILABLE)
     }
   }
 
--- a/kifi-backend/modules/shoebox/app/com/keepit/admin/AdminOrganizationController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/admin/AdminOrganizationController.scala
@@ -11,6 +11,7 @@ import com.keepit.common.db._
 import com.keepit.common.db.slick.Database
 import com.keepit.heimdal.HeimdalContext
 import com.keepit.model._
+import play.api.{ Mode, Play }
 import play.api.libs.json.Json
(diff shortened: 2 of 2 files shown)


ryanpbrewster:
That one is particularly unfortunate. I knew the try/catch would bite us
someday and I guess I was right. The issue is that it is possible for
HandleCommander to fail to assign a handle, and it does so in a way that is
difficult to predict.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
```diff
@@ -225,9 +226,13 @@ class OrganizationCommanderImpl @Inject() (
...
-      case Success(Left(fail)) => Left(fail)
-      case Success(Right(response)) => Right(response)
-      case Failure(ex) => log.error(ex.getMessage); Left(OrganizationFail.HANDLE_UNAVAILABLE)
+      case Success(Left(fail)) =>
+        Left(fail)
+      case Success(Right(response)) =>
+        Right(response)
+      case Failure(ex) =>
+        log.error(s"could not create organization via $request", ex)
+        Left(OrganizationFail.HANDLE_UNAVAILABLE)
```
leogrim, line 235: This should match on the `Exception` before deciding whether `OrganizationFail.HANDLE_UNAVAILABLE` is the right thing to return isn't it? (that was true for the previous code as well)

`HandleCommander` will return an `InvalidHandleException`.

ryanpbrewster, line 235: It should, I agree.

stkem:
Perhaps the main downside of the return-an-error-instead-of-throwing-an-exception pattern: It puts the reporting burden entirely on the caller.

ryanpbrewster:
Throwing an exception just changes that to put the burden of catching on
the caller. For certain things we should not be catching, such as a DB
constraint violation, so in this case it was just bad code (my bad code).

andrewconner:
> It puts the reporting burden entirely on the caller

Not necessarily — just because we report when we don't catch an exception doesn't mean that's how we should report. Commanders can handle it and aggressively log. And for the caller, handling the error becomes obligatory (unless you're just ignoring the return value immediately). In this case, the logging was just insufficient.

stkem:
In both cases the burden of doing _something_ with the failure is on the caller. You either catch the exception/handle the `Failure` case, or you don't. Error type awareness aside, it's just different syntax.
In the latter case, if you don't do anything with the failure case the error becomes silent (I know that's not really the case here, more general point). The the regular exception case it bubbles out and the system get's a chance to deal with it.
While it arguably makes it harder to deal with errors/failures directly because the type of the function doesn't tell you what kind of exceptions to expect, the risk of silent errors is the worse evil in many situations, imho.

andrewconner:
> In both cases the burden of doing something with the failure is on the caller.

I don't think so. If it's an actual error that can't be recovered from, then the Commander should report. The client is responsible for converting to a reasonable client-facing error, yes.

> You either catch the exception/handle the Failure case, or you don't

Correct. We want to make it where it's clear what the right thing to do is; we can't fix bad code. By having good conventions about reporting and handling, we'll end up with less of this.

> the system get's a chance to deal with it... the risk of silent errors is the worse evil in many situations, imho.

Agreed with the latter. The system doesn't deal with it, it just airbrakes and the client most certainly doesn't get enough information to deal with it. All APIs should think about error cases—this should be baked in to our thought-process. If methods were throwing exceptions, that means we catch them and return good errors. If they return an `Option`/`Either`/`Try`/etc, we match on type and return good errors. Being _able_ to ignore errors doesn't mean it's dealt with.

PR #16997, opened 2015-08-26
Moving away from `TemporaryFile` to fix GC races

andrewconner:
@eishay @leogrim @ryanpbrewster @gratimax @camhashemi @stkem 

This is the fix for the random image test that fails. It also fixes cases where we've seen files mysteriously disappear. They disappear whenever a GC happens during processing the images.

The issue is that Play's `TemporaryFile` makes resource management easy (ie, you do nothing and files magically are cleaned up) by deleting the file when the variable gets GCed. To be clear, given a `tempFile`, if _that_ variable goes out of scope, `tempFile.file` may be deleted—even if you still have a reference to `tempFile.file`.

What was happening was the `Action` handed us a `TemporaryFile` for image uploads (and we created one for image fetches by URL), and then we'd do resizing, producing more `TemporaryFiles`. Since what is returned to the client is just a success message, between the last usage of a `TemporaryFile` and the end, a GC could happen, deleting the files.

This happens most often in tests because we have a lot of GC pressure and are checking for file stats after the action is completed. However, it's also happening occasionally in production.

There's certainly better ways for Play to do this, but for now, the changes are:

1) `ProcessImageHelper` deals exclusively in `File`s. This means you need to delete them when you're done, or they'll leak. (on JVM restart, they _are_ cleaned up, but we could fill a disk before that happens)
2) Controllers that receive a `request.body: TemporaryFile` quickly pass on the `request.body.file: File` reference. From then on, file life-cycles are handled explicitly—unfortunately, because of code duplication, every image processing class now also has to manage this. :frowning:
3) Any usage of `TemporaryFile` needs to be very careful; it's not easy to be sure that the `File` can't be deleted. However, it's usually safe to use them in tests.

Big thanks to @jaredpetker for helping me isolate this one.


kifi-backend/modules/rover/app/com/keepit/rover/image/RoverImageFetcher.scala
```diff
@@ -79,8 +78,9 @@ class RoverImageFetcher @Inject() (
                   case Right(uploadedImages) => {
                     try {
                       db.readWrite(attempts = 3) { implicit session =>
-                        uploadedImages.foreach(imageInfoRepo.intern(sourceImage.hash, imageSource, sourceImage.sourceImageUrl, _))
+                        uploadedImages.foreach(u => imageInfoRepo.intern(sourceImage.hash, imageSource, sourceImage.sourceImageUrl, u._1))
                       }
+                      uploadedImages.foreach(_._2.delete())
```
andrewconner, line 83: Notice we now have to be explicit about file deletion.

kifi-backend/modules/shoebox/test/com/keepit/commanders/KeepImageCommanderTest.scala
```diff
@@ -73,13 +73,13 @@ class KeepImageCommanderTest extends Specification with ShoeboxTestInjector with
         val (user, lib, uri, keep1, keep2) = setup()
 
         {
-          val savedF = commander.setKeepImageFromFile(fakeFile1, keep1.id.get, ImageSource.UserUpload)
+          val savedF = commander.setKeepImageFromFile(fakeFile1.file, keep1.id.get, ImageSource.UserUpload)
```
andrewconner: A cool use of `TemporaryFile`. The test doesn't have to worry about cleaning up.

PR #17000, opened 2015-08-26
truncate org names that are longer than 30 chars

camhashemi:
@leogrim @gratimax  https://app.asana.com/0/32728208409723/43146000537263. 30 chars is a bit less than 2 std above the mean, also keeps notifs looking clean.

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/model/Organization.scala
+++ b/kifi-backend/modules/common/app/com/keepit/model/Organization.scala
@@ -46,6 +46,8 @@ case class Organization(
     this.withBasePermissions(BasePermissions(Map(None -> Set())))
   }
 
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/common/app/com/keepit/model/Organization.scala
```diff
@@ -46,6 +46,8 @@ case class Organization(
     this.withBasePermissions(BasePermissions(Map(None -> Set())))
   }
 
+  def truncatedName = if (this.name.length > 30) this.name.take(30) + "..." else this.name
```
maxov: if you import `com.keepit.common.strings._` you can just call `name.abbreviate(30)`

camhashemi: oo nice

PR #17002, opened 2015-08-26
updated ansible startup init scripts; drop external logentries role

joshm1:
@andrewconner  @stkem 

I started this with the goal of updating the init scripts for a new AMI that will ensure spot instances launch automatically like they used to. it's not done because I realized the the ansible playbook is written so that it only copies service-specific config files (shoebox/search/curator/etc init.d, nginx) after it detects that service being launched. the AMIs have traditionally had all of the config fields for all services (and some were manually updated on the servers without updating Git or the AMI).

as I mentioned in slack, we have a couple options:
- I can update ansible to copy all config files to the server like they used to.  on boot, eddie will detect which service to launch and manually launch that one (same as what it does on AMIs 29 and under). what we'd lose, or have to redo, are some of the dynamic configs that are generated when running the ansible playbook regarding what packages to install, which configuration/templates to copy to the server, etc.
- update the init script to actually download the latest ansible configuration (private hosted on s3) and run ansible to provision itself; doing this, the same ansible config will be isomorphic between running it against a cluster of remote machines and running it on localhost

I think the latter solution is nicer. it prevents fragmentation of server config between ansible and shell scripts. the tradeoff is (1) ansible and it's dependencies will need to live on the server (it's mostly python) and (2) it'll take a little bit longer to finish, and requires a job/script for archiving and uploading ansible config to s3 (not hard, just a little extra work)


stkem:
I'm slightly in favor of solution #2 as well. We've had issues in the past with conflicting nginx configs for different services. I'd like to not have to deal with that and similar things any more (i.e. let's keep only the stuff needed for a particular service on that machine).

stkem:
Overall code lgtm

andrewconner:
What's an estimate on (2)?

joshm1:
@andrewconner I think (2) is half day of work

andrewconner:
Do it then

PR #17007, opened 2015-08-26
Feature/activity feed

cvializ:
@andrewconner removes personas (per @eishay's request) and widens the keep column


andrewconner:
Great. @eishay Do we really want it gone yet before the empty state?

cvializ:
Talked with @eishay, we will wait to remove personas until after we have implemented the empty state

PR #17009, opened 2015-08-26
use library that user kept to

camhashemi:
@andrewconner @cvializ this is a front-end fix. we were previously grabbing the first library, which gets updated once you rekeep, and is sometimes not the library that the user kept to.

alternatively, could do this on the backend.

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/feed/feedCtrl.js
+++ b/kifi-backend/modules/shoebox/angular/src/feed/feedCtrl.js
@@ -13,6 +13,13 @@ angular.module('kifi')
       });
     }
 
+    function libraryKeptTo (keep) {
+      var libraryAndUser = keep.libraries.filter(function(libraryAndUser) {
+        return libraryAndUser[0].id === keep.libraryId;
+      })[0] || keep.libraries[0];
+      return libraryAndUser[0];
+    }
+
     var keepLazyLoader = new Paginator(keepSource, 10);
 
     $scope.feed = [];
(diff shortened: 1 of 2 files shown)


andrewconner:
Fine with me. @cvializ you fine mutating the incoming keep? Only concern I see is this means these keeps are different from others (no `.library`)

kifi-backend/modules/shoebox/angular/src/feed/feedCtrl.js
```diff
@@ -28,7 +35,11 @@ angular.module('kifi')
       keepLazyLoader
       .fetch()
       .then(function (keeps) {
-        $scope.feed = keeps;
+        var updatedKeeps = keeps.map(function(keep) {
```
cvializ, line 38: You could use a `.forEach` here instead, and remove the `return keep;` from the body. `.map` will return a new array, but since you're mutating the items in the original array, you'd be better off just manipulating the items and keeping the original array.

cvializ:
Yep I'm fine with mutating the keeps here, since it's pretty clear that there's a mutation. Nicely done, Cam!

andrewconner:
:+1: merge away

camhashemi:
@cvializ @adamjgrant we should keep an eye out if we need this functionality again, in that case I can do it on the backend

kifi-backend/modules/shoebox/angular/src/feed/feedCtrl.js
camhashemi, line 38: right, that makes sense

PR #17015, opened 2015-08-26
add modify library route for mobile

camhashemi:

@jaredpetker @DerekBurch talked about making mobile and web APIs more similar to make it easy to consolidate them eventually. this would do that, as opposed to adding another field to the old endpoint.


ryanpbrewster:
I wonder if there isn't a better way to do this. I think it was @jaredpetker that suggested we have a single strongly versioned controller, and endpoints of the form `/v1/xyz`, `/v2/xyz`, etc, and an alias from `/site/xyz` to the most recent endpoint. I rather like that idea.

camhashemi:
I do too. So much repetition across web and mobile.

camhashemi:
Actually @ryanpbrewster I think if we want to go with this idea, we should do it consciously as a project. Until then, going to try to add as little code (repetition) as possible.

EDIT: NVM. changed my mind. Mobile's current API is slightly different than web's in naming across all fields, would rather have them change now than later.

camhashemi:
@ryanpbrewster added a mobile serialization to the request class, so that we can re-use one class / don't have to manually de-serialize ever

PR #17017, opened 2015-08-26
Mobile add lib

camhashemi:
@andrewconner @leogrim @ryanpbrewster this just enables the endpoint to take an optional `space`


kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileLibraryController.scala
```diff
@@ -65,19 +69,39 @@ class MobileLibraryController @Inject() (
...
+
+    externalAddRequestValidated match {
+      case JsError(errs) =>
+        airbrake.notify(s"Could not json-validate addLibRequest from ${request.userId}: ${request.body}", new JsResultException(errs))
+        BadRequest(Json.obj("error" -> "badly_formatted_request"))
+      case JsSuccess(externalAddRequest, _) =>
+        val libAddRequest = db.readOnlyReplica { implicit session =>
+          val slug = externalAddRequest.slug.getOrElse(LibrarySlug.generateFromName(externalAddRequest.name))
+          val space = externalAddRequest.space map {
+            case ExternalUserSpace(extId) => LibrarySpace.fromUserId(userRepo.getByExternalId(extId).id.get)
```
andrewconner, line 82: I don't have typing to help me, but shouldn't this be `Some(ExternalUserSpace(extId))` or something?

ryanpbrewster, line 82: I think this is right. We want `Option[ExternalLibrarySpace] => Option[LibrarySpace]` and this is a `map` over a `ExternalLibrarySpace => LibrarySpace`, so the types should check out.

andrewconner, line 82: You're right, I'm wrong.

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileLibraryController.scala
```diff
@@ -65,19 +69,39 @@ class MobileLibraryController @Inject() (
...
-    val color = (jsonBody \ "color").asOpt[LibraryColor]
-    val whoCanInvite = (jsonBody \ "whoCanInvite").asOpt[LibraryInvitePermissions]
-    val slug = LibrarySlug.generateFromName(name)
-    val addRequest = LibraryAddRequest(name = name, visibility = visibility, description = description, slug = slug, color = color, whoCanInvite = whoCanInvite)
+    val externalAddRequestValidated = request.body.validate[ExternalLibraryAddRequest]
+
+    externalAddRequestValidated match {
+      case JsError(errs) =>
+        airbrake.notify(s"Could not json-validate addLibRequest from ${request.userId}: ${request.body}", new JsResultException(errs))
+        BadRequest(Json.obj("error" -> "badly_formatted_request"))
```
ryanpbrewster, line 77: Take a look at something cool that @stkem did for PaymentController: https://github.com/kifi/keepit/pull/16987/files#diff-6327a34a1114e86325b9a756392f2a96R78

camhashemi:
@ryanpbrewster same thing here, adding to one serialization object so we can have one point of reference

PR #17018, opened 2015-08-26
Rewrite activity feed query

leogrim:
@eishay @ryanpbrewster


ryanpbrewster:
Not the prettiest query I've ever seen, but surprisingly nonetheless. lgtm

PR #17023, opened 2015-08-26
filter duplicate emails and fake users (org invite typeahead)

camhashemi:
@leogrim

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
@@ -68,6 +68,7 @@ class OrganizationInviteCommanderImpl @Inject() (db: Database,
     typeaheadCommander: TypeaheadCommander,
     organizationAnalytics: OrganizationAnalytics,
     userExperimentRepo: UserExperimentRepo,
+    userCommander: UserCommander,
     implicit val publicIdConfig: PublicIdConfiguration) extends OrganizationInviteCommander with Logging {
 
   private def getValidationError(request: OrganizationInviteRequest)(implicit session: RSession): Option[OrganizationFail] = {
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
```diff
@@ -425,7 +425,13 @@ class OrganizationInviteCommanderImpl @Inject() (db: Database,
       case (users, contacts) =>
         val nonMembers = db.readOnlyMaster { implicit session =>
           val members = organizationMembershipRepo.getByOrgIdAndUserIds(orgId, users.toSet)
-          users.filter(id => !members.exists(_.userId == id))
+          val fakes = userExperimentRepo.haveExperiment(users.toSet, UserExperimentType.FAKE)
```
leogrim: Check out `UserCommander.getAllFakeUsers`, it's cached.

camhashemi: ah, that makes sense that it's cached. I always thought it was fishy to get all of them.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
```diff
@@ -425,7 +425,13 @@ class OrganizationInviteCommanderImpl @Inject() (db: Database,
...
+          users.filter(id => !members.exists(_.userId == id) && !fakes.contains(id))
+        }
+
+        val uniqueEmailsWithName = {
+          val distinctContacts = contacts.groupBy(_.email.address.toLowerCase).map { case (address, contacts) => (address, contacts.find(_.userId.isDefined).getOrElse(contacts.head)) }
+          contacts.map(contact => distinctContacts(contact.email.address))
         }
```
leogrim, line 436: You can probably use `com.keepit.common.CollectionHelpers.dedupBy` here, or actually check for whether contacts have a name or not. There's some logic in `EContact` that does similar contact merging but I don't think it's reusable off the shelf

camhashemi, line 436: `dedupBy` is indifferent to priority, but I should use `contact.name` instead of `userId`

I would prefer that we do this in `EContactTypeahead`, not sure what you think about that.

leogrim:
Well `EContactTypeahead` does the right thing returning `EContact` though.

camhashemi:
But we shouldn't be interested in duplicate emails from it. There should be a unique constraint on `(userId, emailAddress)` in the table, and we would just update the extra fields as we get them. But since that isn't cheap to do, we can just make sure the deduping is done in `abook` instead of `shoebox`.

The behavior your mean might be `TypeaheadCommander[62]`, but it also picks an arbitrary contact, no tie-breaking priorities.

leogrim:
Not because you could have someone in two different address books with
different information, and we want to preserve that.

camhashemi:
@leogrim I see, then in the DB it could be preserved, but for the typeahead we only care about the one with the most useful information (userId, then name). anyways, this behavior should work for now.

leogrim:
Indeed, I think duplicating after the fact like you do is fine. You do want all the info in the typeahead, because you want the various names to be available for query in there.

camhashemi:
Gotcha, since looks like we score by both name and email address. :+1:  thanks

PR #17028, opened 2015-08-27
cleaning bad urls from users before they blow up down the pipe

eishay:
For example: https://fortytwo.airbrake.io/projects/91268/groups/1495973767015845274/notices/1495978433388760490

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
@@ -1,5 +1,6 @@
 package com.keepit.controllers.website
 
+import com.keepit.common.net.URI
 import com.google.inject.Inject
 import com.keepit.commanders.{ RawBookmarkRepresentation, _ }
 import com.keepit.common.akka.SafeFuture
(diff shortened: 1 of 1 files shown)


eishay:
@andrewconner @leogrim

eishay:
waiting on merge for review, not sure its the best place

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryController.scala
```diff
@@ -500,13 +501,18 @@ class LibraryController @Inject() (
...
+        fail
+      }
       val source = KeepSource.site
       implicit val context = heimdalContextBuilder.withRequestInfoAndSource(request, source).build
 
       val existingKeeps = db.readOnlyMaster { implicit s =>
         keepRepo.getByLibrary(libraryId, 0, Int.MaxValue).map(_.externalId).toSet
       }
-      val (keeps, _, failures, _) = keepsCommander.keepMultiple(fromJson, libraryId, request.userId, source, None, false)
+      val (keeps, _, failures, _) = keepsCommander.keepMultiple(cleanBookmarks, libraryId, request.userId, source, None, false)
```
leogrim, line 515: It looks like `KeepCommander.keepMultiple` already handles and returns failures, what's the benefit of special casing out here? I feel like it mixes up concerns and scatters the logic that decides whether a keep should success or not.

andrewconner, line 515: Agreed, we should do it centrally, since everyone needs this (if the existing code isn't working well enough).

PR #17029, opened 2015-08-27
breaking down library commander a bit and few circular dependencies

eishay:


eishay:

ryanpbrewster:
This is a great idea. I'm not so sure about the breakdown here. I like the way we split up orgs into org logic (create/modify/delete) and org memberships (join/leave).

The more I think about it, the more fond I am of the LibraryFetchCommander. It doesn't do any scary modifications, but all the heavy user-facing info logic is in there. Very cool.

eishay:
sure, how would you do the breakdown? 
this is just a start of course, i see other methods that could move out to one of those commanders or new ones.

ryanpbrewster:
How about {LibraryInfo, Library, LibraryMembership, LibraryInvite} commanders.

LibraryCommander gets create/modify/delete library.

LibraryMembershipCommander gets join/leave/follow/subscribe/reorder.

LibraryInviteCommander gets all the invite logic. I am not sure exactly what things go in here.

LibraryInfoCommander (or LibraryFetchCommander) gets getKeeps/createFullInfo/createCard.

Thoughts?

eishay:
I would drop having a generic `LibraryCommander` since its going to be a kitchen sink. You may have a 'kitchen sink' commander then keep it but move the explicit modifiers group of methods to their own commander.
The rest look good

ryanpbrewster:
I like bundling create/modify/delete logic. Maybe LibraryExistenceCommander? I'm struggling with English words that describe what I'm thinking.

eishay:
changes will be:
`LibraryFetchCommander` to `LibraryInfoCommander`
`LibraryModifierCommander` to `LibraryMutatorCommander`

PR #17031, opened 2015-08-27
Eliza bug logging

camhashemi:
@leogrim @ryanpbrewster eliza isn't logging any exceptions for these airbrakes. trying to see if the query needs optimizing or something else is the matter.

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/eliza/ElizaServiceClient.scala
+++ b/kifi-backend/modules/common/app/com/keepit/eliza/ElizaServiceClient.scala
@@ -255,7 +255,7 @@ class ElizaServiceClientImpl @Inject() (
   }
 
   def getTotalMessageCountForGroup(users: Set[Id[User]]): Future[Int] = {
-    call(Eliza.internal.getTotalMessageCountForGroup, body = Json.toJson(users)).map { response =>
+    call(Eliza.internal.getTotalMessageCountForGroup, body = Json.toJson(users), callTimeouts = longTimeout).map { response =>
       response.json.as[Int]
     }
   }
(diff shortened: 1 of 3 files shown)


ryanpbrewster:
Can we use @stkem's cool timing macros?

camhashemi:
That's what I initially went towards, but looks like it needs a minimum number of calls before it notifies airbrake. This is just to log, fix the bug, then remove the log.

ryanpbrewster:
:+1: okay. Works for me.

stkem:
There is two macros, only one of which (the alerting one) needs a minimum number of calls. The other one `StatsdTiming` just sends timing information to stats starting immediately. Much easier to see performance issues that way (as you get a bunch of nice graphs) than by looking at log lines. I suggest putting that on `elizaStatsCommander.getTotalMessageCountForGroup`.
If you need more fine grained control of what is being timed (e.g. you want to time per user) you can use `NamedStatsdTimer` directly (not a macro, search for the class). It's more light weight than logging too. The only case in which you might want to use logging for performance monitoring is if the function in questions only runs _very_ rarely. Not sure if that is the case here.

stkem:
Also, Happy Birthday @camhashemi !! :cake:

camhashemi:
@stkem thank you for clearing that up. I'll use that instead (you just add the annotation with the method name right?). This is being called pretty frequently.

@andrewconner @leogrim @ryanpbrewster I'm trying to discern whether these airbrakes are from issues with heimdal machines or the eliza queries

see: https://fortytwo.airbrake.io/projects/91826/groups/1496414243974227199/notices/1496414296705017139
https://fortytwo.airbrake.io/projects/91826/groups/1484814562412407987/notices/1496414290514224432

The majority of airbrakes are timeout exceptions on calls to both eliza and shoebox, but the initial logging shows that eliza's query is taking less than 20 ms.

camhashemi:
thank you @stkem

PR #17033, opened 2015-08-27
Rewrite activity feed query to use the new Keepscussion repos

leogrim:
@ryanpbrewster @andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/model/KeepRepo.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/model/KeepRepo.scala
@@ -1,7 +1,5 @@
 package com.keepit.model
 
-import java.sql.Timestamp
-
 import com.google.inject.{ ImplementedBy, Inject, Singleton }
 import com.keepit.commanders.{ LibraryMetadataCache, LibraryMetadataKey, WhoKeptMyKeeps }
 import com.keepit.common.db._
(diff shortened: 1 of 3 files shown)


andrewconner:
:+1:

kifi-backend/modules/shoebox/app/com/keepit/model/KeepRepo.scala
```diff
@@ -80,6 +78,8 @@ class KeepRepoImpl @Inject() (
     val clock: Clock,
     libraryMembershipRepo: LibraryMembershipRepo, // implicit dependency on this repo via a plain SQL query getRecentKeeps
     organizationMembershipRepo: OrganizationMembershipRepo, // implicit dependency on this repo via a plain SQL query getRecentKeeps
+    keepToLibraryRepo: KeepToLibraryRepo, // implicit dependency on this repo via a plain SQL query getRecentKeeps
+    keepToUserRepo: KeepToUserRepo, // implicit dependency on this repo via a plain SQL query getRecentKeeps
```
ryanpbrewster, line 82: I like this, nice comments.

kifi-backend/modules/shoebox/app/com/keepit/model/KeepRepo.scala
```diff
@@ -544,46 +544,58 @@ class KeepRepoImpl @Inject() (
...
+    val keepsFromUser = s"""SELECT ktu.keep_id as id, ktu.added_at as first_added_at FROM $ktu_WHERE_THIS_USER AND $added_at_BEFORE"""
+
+    val keepsAndFirstAddedAt = sql"""
+      SELECT k.id as keep_id, min(k.first_added_at) as added_at
+      FROM ((#$keepsFromLibraries) UNION (#$keepsFromOrganizations) UNION (#$keepsFromUser)) as k
+      GROUP BY k.id
+      HAVING #$added_at_AFTER AND #$added_at_BEFORE
+      ORDER BY added_at DESC, keep_id DESC
+      LIMIT $limit
+    """.as[(Id[Keep], DateTime)].list
```
ryanpbrewster, line 594: Do we need this to be `(Id[Keep], DateTime)`? Can't we just `SELECT k.id as keep_id` and call it a day?

ryanpbrewster, line 594: Oh, I see, we're ordering by an aggregate so you want to save it. I don't think that's strictly necessary, though.

Can you use `SELECT k.id as keep_id ... ORDER BY min(k.first_added_at)`?

ryanpbrewster, line 594: All super low priority changes.

PR #17039, opened 2015-08-27
Org deleted libs

ryanpbrewster:
@eishay @andrewconner


kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -646,176 +660,185 @@ class LibraryCommanderImpl @Inject() (
...
-      val newSubKeysOpt = modifyReq.subscriptions
+    val newSpace = modifyReq.space.getOrElse(library.space)
+    val errorOpts = Stream(
+      validateUserWritePermission,
+      validateSpace(modifyReq.space),
+      validateName(modifyReq.name, newSpace),
+      validateSlug(modifyReq.slug, newSpace),
+      validateVisibility(modifyReq.visibility, newSpace)
+    )
+    errorOpts.flatten.headOption
```
ryanpbrewster, line 729: Not sure if this is the clearest way to express "check if any of these actually contains an error, and return the first one, otherwise return `None`".

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
```diff
@@ -277,6 +277,10 @@ class OrganizationCommanderImpl @Inject() (
           val invites = orgInviteRepo.getAllByOrganization(org.id.get)
           invites.foreach(orgInviteRepo.deactivate)
 
+          libraryRepo.getBySpace(org.id.get, excludeStates = Set.empty).foreach { lib =>
+            libraryCommander.unsafeModifyLibrary(lib, LibraryModifyRequest(space = Some(lib.ownerId)))
+          }
```
ryanpbrewster, line 282: This whole PR was just to find a way to write these 2 lines.

PR #17040, opened 2015-08-27
Cleaning up unnecessary Awaits in Eliza messaging

andrewconner:
@stkem @leogrim @gratimax 

Would love to clean this up more (some really strange stuff in here), but moving on. Last remaining await is for BasicUsers, and there are several dependencies on it being synchronous.

### DIFF
--- a/kifi-backend/modules/eliza/app/com/keepit/eliza/controllers/mobile/MobileMessagingController.scala
+++ b/kifi-backend/modules/eliza/app/com/keepit/eliza/controllers/mobile/MobileMessagingController.scala
@@ -104,24 +104,20 @@ class MobileMessagingController @Inject() (
     val validUserRecipients = users.collect { case JsSuccess(validUser, _) => validUser }
     val validEmailRecipients = emailContacts.collect { case JsSuccess(validContact, _) => validContact }
 
-    val url = (o \ "url").asOpt[String]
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/eliza/app/com/keepit/eliza/controllers/mobile/MobileMessagingController.scala
```diff
@@ -104,24 +104,20 @@ class MobileMessagingController @Inject() (
...
-        case (message, threadInfoOpt, messages) =>
-          Ok(Json.obj(
-            "id" -> message.externalId.id,
-            "parentId" -> message.threadExtId.id,
-            "createdAt" -> message.createdAt,
-            "threadInfo" -> threadInfoOpt,
-            "messages" -> messages.reverse))
-      }
-
-    messageSubmitResponse // todo(JP, Eduardo, Léo): return meaningful error about invalid participants
```
maxov, line 124: So I take it this will never be done? :stuck_out_tongue:

maxov:
:+1: nice work, makes the code clearer

kifi-backend/modules/eliza/app/com/keepit/eliza/controllers/mobile/MobileMessagingController.scala
andrewconner, line 124: Since clients don't do anything with meaningful errors, nah. Until they want it.

PR #17041, opened 2015-08-27
mobile endpoints for adding/removing library images

camhashemi:
@andrewconner @ryanpbrewster @leogrim this is a stronger assertion of using the same controller for mobile and web

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
+++ b/kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
@@ -6,7 +6,7 @@ import com.keepit.common.healthcheck.{ AirbrakeNotifier, AirbrakeError }
 import com.keepit.common.logging.Logging
 import com.keepit.common.core._
 import com.keepit.common.net.URI
-import com.keepit.model.{ UserExperimentType, KifiInstallation, User }
+import com.keepit.model.{KeepSource, UserExperimentType, KifiInstallation, User}
 import play.api.Play
 import play.api.libs.iteratee.Iteratee
 import play.api.mvc._
@@ -60,6 +60,12 @@ case class UserRequest[T](request: Request[T], userId: Id[User], adminUserId: Op
   override def identityOpt = identityOpt0.awaitGet
 
   lazy val kifiInstallationId: Option[ExternalId[KifiInstallation]] = helper.getKifiInstallationIdOpt
+  lazy val source = {
+    if (req.path.startsWith("/site/")) KeepSource.site
+    else if (req.path.startsWith("/m/")) KeepSource.mobile
+    else if (req.path.startsWith("/ext/")) KeepSource.keeper
+    else KeepSource.unknown
+  }
 }
 
 // for backward-compatibility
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
```diff
@@ -60,6 +60,12 @@ case class UserRequest[T](request: Request[T], userId: Id[User], adminUserId: Op
   override def identityOpt = identityOpt0.awaitGet
 
   lazy val kifiInstallationId: Option[ExternalId[KifiInstallation]] = helper.getKifiInstallationIdOpt
+  lazy val source = {
+    if (req.path.startsWith("/site/")) KeepSource.site
+    else if (req.path.startsWith("/m/")) KeepSource.mobile
+    else if (req.path.startsWith("/ext/")) KeepSource.keeper
+    else KeepSource.unknown
+  }
```
camhashemi, line 68: the idea behind this field is to `match` on it, then handle de/serialization according to the client (and version).

ryanpbrewster:
:+1: I like it.

kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
andrewconner, line 68: Why? Whoever calls this already knows what it is.

kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
```diff
@@ -60,6 +60,12 @@ case class UserRequest[T](request: Request[T], userId: Id[User], adminUserId: Op
   override def identityOpt = identityOpt0.awaitGet
 
   lazy val kifiInstallationId: Option[ExternalId[KifiInstallation]] = helper.getKifiInstallationIdOpt
+  lazy val source = {
```
andrewconner, line 63: This is also a bit weird, because the type is `KeepSource` (i.e., the source of a Keep). Not really related to requests. Not sure how/why the library image code uses it.

kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
camhashemi, line 68: If we were going to use one controller, this logic would be re-used. Sounds like that's not sticking, so I'll remove this.

kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
camhashemi, line 63: I could create a new class like `UserAgent`.

andrewconner, line 63: The problem is that it means it becomes opaque how APIs work — looking at a controller, it's not really obvious how it works (until you see how a request was routed). `UserAgent` is already something that contains user agent info.

kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
andrewconner, line 68: If we re-use logic, why would we need `source`? We can have unified, versioned APIs that all clients use.

kifi-backend/modules/common/app/com/keepit/common/controller/UserActions.scala
camhashemi, line 63: I didn't know `UserAgent` existed, but that's exactly what I wanted instead of `request.source`

camhashemi, line 63: I thought it as the opposite: with more controllers/methods, they fall out of sync and get harder to understand / keep track of.

My assumption is that the difference between `OrganizationCard` for one client+version and another is more fields or less fields, based on the same canonical model (`Organization`). So get the core model in one way, and just return the subset of fields the clients accept.

PR #17042, opened 2015-08-27
add mobile route for organization libraries

camhashemi:
@andrewconner @leogrim @ryanpbrewster @stkem 

it's much simpler to add endpoints when they're shared :) when something (required) changes, that's the hard part. interested in how you all feel about this.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryImageController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryImageController.scala
@@ -33,6 +33,7 @@ class LibraryImageController @Inject() (
     extends UserActions with LibraryAccessActions with ShoeboxServiceController {
 
   def uploadLibraryImage(pubId: PublicId[Library], imageSize: Option[String] = None, posX: Option[Int] = None, posY: Option[Int] = None) = (UserAction andThen LibraryOwnerAction(pubId)).async(parse.temporaryFile) { request =>
+    // TODO(cam): this handler is used by mobile and needs to be versioned in case required fields are added or removed
     val libraryId = Library.decodePublicId(pubId).get
     val imageRequest = db.readWrite { implicit session =>
       libraryImageRequestRepo.save(LibraryImageRequest(libraryId = libraryId, source = ImageSource.UserUpload))
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/OrganizationController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/OrganizationController.scala
@@ -90,6 +90,7 @@ class OrganizationController @Inject() (
   }
 
   def getOrganizationLibraries(pubId: PublicId[Organization], offset: Int, limit: Int) = OrganizationAction(pubId, authTokenOpt = None, OrganizationPermission.VIEW_ORGANIZATION) { request =>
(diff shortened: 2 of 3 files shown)


kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryImageController.scala
```diff
@@ -33,6 +33,7 @@ class LibraryImageController @Inject() (
     extends UserActions with LibraryAccessActions with ShoeboxServiceController {
 
   def uploadLibraryImage(pubId: PublicId[Library], imageSize: Option[String] = None, posX: Option[Int] = None, posY: Option[Int] = None) = (UserAction andThen LibraryOwnerAction(pubId)).async(parse.temporaryFile) { request =>
+    // TODO(cam): this handler is used by mobile and needs to be versioned in case required fields are added or removed
```
andrewconner, line 36: This isn't needed if you have a mobile controller that calls it.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/OrganizationController.scala
```diff
@@ -90,6 +90,7 @@ class OrganizationController @Inject() (
   }
 
   def getOrganizationLibraries(pubId: PublicId[Organization], offset: Int, limit: Int) = OrganizationAction(pubId, authTokenOpt = None, OrganizationPermission.VIEW_ORGANIZATION) { request =>
+    // TODO(cam): this handler is used by mobile and needs to be versioned in case required fields are added or removed
```
andrewconner, line 93: Seems like this comment saves 3 lines in a mobile controller (just copying this method). If they need a change later, preso, no extra funkiness needed

andrewconner:
Commander logic can _usually_ be shared with no issues, that's good. Controller sharing gets a bit hairier, since there's nothing helping us keep it consistent. For such a smaller controller method, I'd just throw a copy in the mobile controller.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryImageController.scala
camhashemi, line 36: Then we have to maintain a new controller with mostly the same methods. If we have one controller, we can create duplicate methods as we need them, but not before.

ryanpbrewster:
I think you're on to something here, but this is not quite it.

If we had a single controller for each commander, but heavily versioned, and the version was part of the `request`, we could have the same logic but different serializations and it would all kind of work out.

ryanpbrewster:
I'm super wary of contract-by-comment, though. The compiler doesn't tell us when we are dumb in comments.

camhashemi:
That was sort of the idea behind `request.source` in #17041 looking at `request.path` for client info, could be extended to grab the API version.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/LibraryImageController.scala
andrewconner, line 36: The commander is typed, though, and isn't speaking `JsValue => JsValue`. You wouldn't accidentally change formats here.

camhashemi:
But alas, I bow my head to you who have more years and wisdom than I.

andrewconner:
If we want to unify API endpoints, we can (and should get everyone on board), but let's do it by unifying APIs, not increasing implicitness.

camhashemi:
Got it, thanks for taking the time!

PR #17043, opened 2015-08-27
Add empty state for users without feed activity.

cvializ:
@andrewconner (@adamjgrant) https://app.asana.com/0/39814736029951/46295083643999


andrewconner:
Cool. The images look a bit empty at the bottom to me (without the "Follow"). What do you think?

cvializ:
I agree with you, when the images are full-size. But I think it looks fine at smaller sizes, how they are displayed on the page. Agree/disagree?
screen shot 2015-08-27 at 3 22 56 pm

I could add the follow button pretty easily if you think it looks better.

cvializ:
Looks like this with "follow" on it.

screen shot 2015-08-27 at 3 43 12 pm

I do think I'll use this instead, since it looks more "followable" and that's what we want.

PR #17044, opened 2015-08-27
add organization ids to ext/start

camhashemi:
@cvializ @andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtAuthController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtAuthController.scala
@@ -118,7 +118,7 @@ class ExtAuthController @Inject() (
     Ok(Json.obj(
       "user" -> BasicUser.fromUser(request.user),
       "libraryIds" -> Seq(libraries._1.id.get, libraries._2.id.get).map(Library.publicId),
-      "orgIds" -> organizations,
+      "orgs" -> organizations,
       "installationId" -> installation.externalId.id,
       "experiments" -> request.experiments.map(_.value),
       "rules" -> Json.obj("version" -> "hy0e5ijs", "rules" -> Json.obj("url" -> 1, "shown" -> 1)), // ignored as of extension 3.2.11
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtAuthController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtAuthController.scala
@@ -6,7 +6,7 @@ import scala.util.Failure
 
 import com.google.inject.Inject
 import com.keepit.common.akka.SafeFuture
-import com.keepit.commanders.{ UserIpAddressCommander, AuthCommander, LibraryCommander, LocalUserExperimentCommander }
+import com.keepit.commanders._
 import com.keepit.common.controller.FortyTwoCookies.KifiInstallationCookie
 import com.keepit.common.controller._
 import com.keepit.common.crypto.{ PublicIdConfiguration, RatherInsecureDESCrypt }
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtAuthController.scala
```diff
@@ -114,6 +116,7 @@ class ExtAuthController @Inject() (
     Ok(Json.obj(
       "user" -> BasicUser.fromUser(request.user),
       "libraryIds" -> Seq(libraries._1.id.get, libraries._2.id.get).map(Library.publicId),
+      "orgIds" -> organizations.map(Organization.publicId),
```
andrewconner: Interesting, just the IDs? How will the extension display it?

camhashemi: that's what @cvializ asked for, maybe there was some confusion

andrewconner: What about including a small Organization serialiazable object? Something like OrganizationCard, but I wish that one was smaller. Seems that we'd want name, avatar, id, and handle. I talked with @leogrim and @ryanpbrewster about this yesterday, not sure if anything was resolved.

Unless they have something, to unblock you, org card may be the way to go for now

cvializ: Yeah I'll need enough to display the avatar, name, link etc.

camhashemi: as long as what you need is a subset of `OrganizationCard`, `LibraryCardInfo`, etc, we can just add a new serialization for the extension that excludes what you don't need. no problem.

camhashemi: cool, fix on the way

andrewconner: > new serialization for the extension

Don't complicate it just yet, this problem is everywhere but one place that currently uses org cards (the one place is the homepage right side bar)

ryanpbrewster: :arrow_up: I agree with that guy.

camhashemi: I just want to avoid the situation that there is with `Library` objects, with so much repetition and deprecation due to creating a new case class for each change. If we can have some convention for creating new classes that avoids this, that'd be nice.

Keeping one object with various serializations is flexible to adding fields (you include additional fields where they're needed, omit them where they're not), but I realize it's inflexible to refactors.

Maybe I'm pre-empting to early, seems like we're all weary of the same thing enough to prevent it.

camhashemi:
@andrewconner @cvializ updated. returns `OrganizationCard`. I can create a `LiteOrganizationCard` if we decide to go down that route.

kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtAuthController.scala
```diff
@@ -114,6 +118,7 @@ class ExtAuthController @Inject() (
     Ok(Json.obj(
       "user" -> BasicUser.fromUser(request.user),
       "libraryIds" -> Seq(libraries._1.id.get, libraries._2.id.get).map(Library.publicId),
+      "orgIds" -> organizations,
```
andrewconner: `"orgs"` or `"organizations"` perhaps?

leogrim: I think we'll turn `OrganizationCard` into the light one and trim a few fields there. Let's introduce anything new for now.

PR #17047, opened 2015-08-27
basic credit management for the admin

stkem:
@andrewconner @ryanpbrewster


kifi-backend/modules/common/conf/evolutions/shoebox/385.sql
```diff
@@ -0,0 +1,9 @@
+# SHOEBOX
+
+# --- !Ups
+INSERT INTO `paid_plan` (`id`, `created_at`, `updated_at`, `state`, `name`, `billing_cycle`, `price_per_user_per_cycle`, `kind`)
+VALUES
+  (1, '2015-08-19 21:50:48', '2015-08-19 21:50:48', 'active', 'Test', 1, 0, 'normal');
```
andrewconner, line 6: perhaps `now()`?

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminPaymentsController.scala
```diff
@@ -69,4 +72,27 @@ class AdminPaymentsController @Inject() (
...
+  def grantExtraCredit(orgId: Id[Organization]) = AdminUserAction { request =>
+    val amount = request.body.asFormUrlEncoded.get.apply("amount").head.toInt
+    val passphrase = request.body.asFormUrlEncoded.get.apply("passphrase").head.toString
+    val memoRaw = request.body.asFormUrlEncoded.get.apply("memo").head.toString
+    val memo = if (memoRaw == "") None else Some(memoRaw)
+    val dollarAmount = DollarAmount(amount)
+
+    val org = db.readOnlyMaster { implicit session => organizationRepo.get(orgId) }
+
+    val passphraseCorrect: Boolean = org.primaryHandle.exists(handle => handle.normalized.value == passphrase.reverse)
```
andrewconner, line 84: Perhaps this should be the admin's password, and we double check it?

kifi-backend/modules/shoebox/app/views/admin/organization.scala.html
```diff
@@ -294,6 +301,23 @@
       location.reload();
     };
 
+    $("#extra-credit-submit").click(function () {
```
andrewconner, line 393: Why not just do a regular form?

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminPaymentsController.scala
```diff
@@ -21,6 +22,8 @@ class AdminPaymentsController @Inject() (
     planCommander: PlanManagementCommander,
     db: Database) extends AdminUserActions {
 
+  val EXTRA_SPECIAL_ADMINS = Seq[Id[User]](Id[User](1), Id[User](243))
```
ryanpbrewster: `Set(1, 243).map(Id[User](_))`?

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminPaymentsController.scala
```diff
@@ -69,4 +72,27 @@ class AdminPaymentsController @Inject() (
...
+    val passphrase = request.body.asFormUrlEncoded.get.apply("passphrase").head.toString
+    val memoRaw = request.body.asFormUrlEncoded.get.apply("memo").head.toString
+    val memo = if (memoRaw == "") None else Some(memoRaw)
+    val dollarAmount = DollarAmount(amount)
+
+    val org = db.readOnlyMaster { implicit session => organizationRepo.get(orgId) }
+
+    val passphraseCorrect: Boolean = org.primaryHandle.exists(handle => handle.normalized.value == passphrase.reverse)
+
+    if ((amount < 0 || amount > 10000) && !(EXTRA_SPECIAL_ADMINS.contains(request.userId))) {
```
ryanpbrewster, line 86: So this is why I was worried about using a raw `Int` as currency. I'm pretty sure those are cents, so I can do mental math and see that's $100, but I feel like it'd be much clearer if this were `if (amount < Currency.ZERO || amount > Currency.dollars(100))`?

ryanpbrewster, line 86: Also, if you're gonna use `contains` I'd definitely go with a `Set` for `EXTRA_SPECIAL_ADMINS`.

ryanpbrewster, line 86: Maybe a user experiment for credit-granting admins?

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminPaymentsController.scala
```diff
@@ -69,4 +72,27 @@ class AdminPaymentsController @Inject() (
...
+    } else if (amount == 0) {
+      Ok("Umm, 0 credit?")
+    } else if (!passphraseCorrect) {
+      Ok("So sorry, but your passphrase isn't right.")
+    } else {
+      planCommander.grantSpecialCredit(orgId, dollarAmount, Some(request.userId), memo)
+      Ok(s"Sucessfully granted special credit of $dollarAmount to Organization #orgId.")
+    }
+  }
+
```
ryanpbrewster, line 97: Can we move all the validation logic into the commander, and turn this into a simple one-liner? ((For admin endpoints I don't feel as strongly, so I won't be that sad if you decide to keep it here.))

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminPaymentsController.scala
```diff
@@ -69,4 +72,27 @@ class AdminPaymentsController @Inject() (
     Ok.chunked(enum)
   }
 
+  def grantExtraCredit(orgId: Id[Organization]) = AdminUserAction { request =>
+    val amount = request.body.asFormUrlEncoded.get.apply("amount").head.toInt
```
ryanpbrewster, line 76: I strongly prefer JSON to FormUrlEncoded data. IMO it's much easier to write, debug, and hit them.

kifi-backend/modules/common/conf/evolutions/shoebox/385.sql
stkem, line 6: I ran it with `now()` in the prod db and then took that timestamp and put it here, so we always have constant data in dev mode and tests.

kifi-backend/modules/shoebox/app/views/admin/organization.scala.html
stkem, line 393: I was following other examples in this file. Also going to make it easier to add client side validation down the road.

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminPaymentsController.scala
stkem, line 76: I was following the pattern of other submission on the orgs admin page. Also, this endpoint _should_ be hard to hit.

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminPaymentsController.scala
stkem: Sure :-)

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminPaymentsController.scala
stkem, line 97: Some of these restrictions are very specific to this particular admin endpoint and shouldn't be in the commander logic. Also, more generally speaking, request validation is really the controllers job. The commander shouldn't have to worry about getting bad data (enforced as much as possible through the type system) or having to check for authorization.
The part about the extra special admins is arguable, but the rest (can't give 0 credit, checking the passphrase) is squarely controller responsibility.

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminPaymentsController.scala
stkem, line 84: This is really intended more as a way to make it difficult to submit the form so you can't do it accidentally and have to give it some thought and less as an actual access protection.

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminPaymentsController.scala
stkem, line 86: The validation here is on the raw data the client sends up. That is always going to be an `Int` at some stage. Everywhere down the line it's dealt with as a typed `DollarAmount`

stkem, line 86: Nevertheless I see your point. I've got some `BigDecimal` based currency arithmetic stuff coming hoefully soon that we can use here as well to make it clearer.

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminPaymentsController.scala
```diff
@@ -69,4 +72,27 @@ class AdminPaymentsController @Inject() (
...
+
+    if ((amount < 0 || amount > 10000) && !(EXTRA_SPECIAL_ADMINS.contains(request.userId))) {
+      Ok("You are not special enough to deduct credit or grant more than $100.")
+    } else if (amount == 0) {
+      Ok("Umm, 0 credit?")
+    } else if (!passphraseCorrect) {
+      Ok("So sorry, but your passphrase isn't right.")
+    } else {
+      planCommander.grantSpecialCredit(orgId, dollarAmount, Some(request.userId), memo)
+      Ok(s"Sucessfully granted special credit of $dollarAmount to Organization #orgId.")
```
andrewconner, line 94: #orgId should be $orgId

stkem, line 94: Ah, indeed! Fixing that.

PR #17057, opened 2015-08-28
Fix the featured libraries link. Angular wanted to route it.

cvializ:
@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/common/services/routeService.js
+++ b/kifi-backend/modules/shoebox/angular/src/common/services/routeService.js
@@ -234,7 +234,7 @@ angular.module('kifi')
       // Static Pages            //
       /////////////////////////////
       featuredLibraries: function () {
-        return 'https://www.kifi.com/libraries/featured';
+        return '/libraries/featured';
       }
     };
   }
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/angular/src/common/services/routeService.js
```diff
@@ -229,6 +228,13 @@ angular.module('kifi')
       },
       removePersona: function (personaName) {
         return route('/user/personas/' + personaName);
+      },
+
+      /////////////////////////////
+      // Static Pages            //
+      /////////////////////////////
+      featuredLibraries: function () {
+        return 'https://kifi.com/libraries/featured';
```
andrewconner: `www.`, otherwise it'll redirect.

kifi-backend/modules/shoebox/angular/src/feed/feed.tpl.html
```diff
@@ -5,7 +5,7 @@
     </header>
     <div class="kf-keep">
       <div class="kf-feed-empty-keep">
-        <span>Follow top libraries from our <a href="/libraries/featured">Featured Libraries</a></span>
```
andrewconner, line 8: Another way to do it (I think what they recommend) is `target="_self"`, which will cause Angular not to route it.

cvializ, line 8: got it, I'll do that

cvializ:
how's that?

andrewconner:
:+1:

PR #17060, opened 2015-08-28
fetch group shared message count more directly

camhashemi:
@leogrim @gratimax omitted the "group and sort by week" behavior to speed up the query

### DIFF
--- a/kifi-backend/modules/eliza/app/com/keepit/eliza/model/UserThreadRepo.scala
+++ b/kifi-backend/modules/eliza/app/com/keepit/eliza/model/UserThreadRepo.scala
@@ -504,7 +504,7 @@ class UserThreadRepoImpl @Inject() (
     if (users.isEmpty) Seq.empty[GroupThreadStats]
     else {
       val users_list = users.map(_.id).mkString("(", ",", ")")
-      val query = sql"""
+      sql"""
         select thread_id, created_at, count(*) as c from user_thread
           where user_id in #$users_list
           and replyable = 1
@@ -512,9 +512,7 @@ class UserThreadRepoImpl @Inject() (
           group by thread_id
           order by week(created_at)
           desc
-      """
-
-      query.as[(Long, DateTime, Int)].list.map((GroupThreadStats.apply _).tupled)
+      """.as[(Long, DateTime, Int)].list.map((GroupThreadStats.apply _).tupled)
     }
   }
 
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/eliza/app/com/keepit/eliza/model/UserThreadRepo.scala
```diff
@@ -480,6 +482,23 @@ class UserThreadRepoImpl @Inject() (
...
+    else {
+      val users_list = users.map(_.id).mkString(",")
+      val queryStr = """
+        select thread_id, created_at, count(*) as c from user_thread
+          where user_id in (""" + users_list + """)
+          and replyable = 1
+          group by thread_id
+          having count(*) > 1
+      """
+      val query = new SQLInterpolation_WarningsFixed(StringContext(queryStr)).sql.as[(Long, DateTime, Int)]
```
andrewconner, line 497: Why not use the `sql""" ... """` syntax?

maxov, line 497: No way to do the hacky `users_list` interpolation without it, at least at the time I wrote the original query. I'm sure using a method on Query is an alternative

camhashemi, line 497: I just copied what was already there (this query just removes a couple conditions from another), but I like `sql"""..."""` better. Is there a semantic difference?

kifi-backend/modules/eliza/app/com/keepit/eliza/model/UserThreadRepo.scala
```diff
@@ -480,6 +482,23 @@ class UserThreadRepoImpl @Inject() (
     }
   }
 
+  def getAllSharedThreadsForGroup(users: Seq[Id[User]])(implicit session: RSession): Seq[GroupThreadStats] = {
```
maxov: could you call this `getSharedThreadsForGroup`? Because there is a big difference between `getAllThreadsForGroupByWeek` and `def getAllThreadsForGroupByWeek` and having `allShared` in there makes the intention of this method unclear

kifi-backend/modules/eliza/app/com/keepit/eliza/model/UserThreadRepo.scala
camhashemi, line 497: ah, it's the same.

andrewconner, line 497: @gratimax oh I see. I think you can do `#$users_list`

kifi-backend/modules/eliza/app/com/keepit/eliza/model/UserThreadRepo.scala
camhashemi: Yeah, I see what you mean.

camhashemi:
@andrewconner @gratimax updated

maxov:
:+1:

PR #17062, opened 2015-08-28
handle edge cases to reduce service calls

camhashemi:
@andrewconner @ryanpbrewster


ryanpbrewster:
These are definitely optimizations, but I wonder how much they actually influence the load. Do you have any idea what query (and what part of that query) is taking so long?

camhashemi:
Eliza has a query that was doing more work than it needed to (grouping by week, we don't care about that here). I optimized that which will help, but we were frequently fetching the number of messages shared between groups of users where `.size == 1`. Not only will it reduce load, it's just more accurate.

PR #17063, opened 2015-08-28
Add tooltips to the feed title links.

cvializ:
@andrewconner


cvializ:
If they would render off-screen at the top of the page, I added some logic that pushes them downward instead. Right now scrolling does not change whether it renders above or below, but it could if we wanted that.

screen shot 2015-08-28 at 11 21 33 am

andrewconner:
Cool!

PR #17064, opened 2015-08-28
Drained the execution context to let tests finish

ryanpbrewster:
There are some tests that launch Futures that need the db table. If the test
finishes before the Future needs the db table, the table will be dropped (and
the Future will blow up when it finally gets around to reaching for the table).

This just blocks the test from finishing until the Future executes, so no more
scary JdbcSQLExceptions

Thanks, @andrewconner


maxov:
:+1: :+1: :+1: :+1:

PR #17065, opened 2015-08-28
cache sets of user ids (members) by org id

camhashemi:

Let's see how this improves heimdal's day. If we still have trouble, we could think about doing an in-memory cache on the result of `PrimaryOrgProvider.getOrgContextData`, with the TTL of 1 day. This org-stats fetching happens every time an event gets tracked, but we probably don't care that these general org stats are accurate up-to-the-second. Maybe up-to-the-day or 12 hours.

### DIFF
--- a/kifi-backend/modules/abook/app/com/keepit/common/cache/ABookCacheModule.scala
+++ b/kifi-backend/modules/abook/app/com/keepit/common/cache/ABookCacheModule.scala
@@ -181,5 +181,5 @@ case class ABookCacheModule(cachePluginModules: CachePluginModule*) extends Cach
 
   @Provides @Singleton
   def organizationMembersCache(stats: CacheStatistics, accessLog: AccessLog, innerRepo: InMemoryCachePlugin, outerRepo: FortyTwoCachePlugin) =
-    new OrganizationMembersCache(stats, accessLog, (outerRepo, 14 days))
+    new OrganizationMembersCache(stats, accessLog, (outerRepo, 7 days))
 }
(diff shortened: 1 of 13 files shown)


kifi-backend/modules/common/app/com/keepit/model/OrganizationMembership.scala
```diff
@@ -117,3 +121,12 @@ object IngestableOrganizationMembership {
   implicit val format = Json.format[IngestableOrganizationMembership]
 }
 
+case class OrganizationMembersKey(id: Id[Organization]) extends Key[Set[Id[User]]] {
+  override val version = 4
```
ryanpbrewster: version 1 to start with

kifi-backend/modules/common/app/com/keepit/model/OrganizationMembership.scala
```diff
@@ -117,3 +121,12 @@ object IngestableOrganizationMembership {
...
+  val namespace = "user_ids_by_organization"
```
ryanpbrewster: member_ids maybe?

camhashemi: :+1: I didn't want to confuse it with membership ids, but I think you got it

kifi-backend/modules/abook/app/com/keepit/common/cache/ABookCacheModule.scala
```diff
@@ -178,4 +178,8 @@ case class ABookCacheModule(cachePluginModules: CachePluginModule*) extends Cach
   @Provides @Singleton
   def primaryOrgForUserCache(stats: CacheStatistics, accessLog: AccessLog, innerRepo: InMemoryCachePlugin, outerRepo: FortyTwoCachePlugin) =
     new PrimaryOrgForUserCache(stats, accessLog, (innerRepo, 5 minutes), (outerRepo, 14 days))
+
+  @Provides @Singleton
+  def organizationMembersCache(stats: CacheStatistics, accessLog: AccessLog, innerRepo: InMemoryCachePlugin, outerRepo: FortyTwoCachePlugin) =
+    new OrganizationMembersCache(stats, accessLog, (outerRepo, 14 days))
```
ryanpbrewster: Every single other service has a 7 day cache. Don't think it matters, but for uniformity maybe this should be 7 as well?

PR #17069, opened 2015-08-28
don't send org properties with every event

camhashemi:
@eishay @leogrim per https://app.asana.com/0/32728208409723/44164549287773, we should only send org info with `eventProps` where applicable. We send org data with `userProps` every time.

### DIFF
--- a/kifi-backend/modules/heimdal/app/com/keepit/heimdal/AmplitudeClient.scala
+++ b/kifi-backend/modules/heimdal/app/com/keepit/heimdal/AmplitudeClient.scala
@@ -149,7 +149,7 @@ trait AmplitudeRequestBuilder {
       primaryOrgProvider.getPrimaryOrg(userId).flatMap {
         case Some(orgId) =>
           primaryOrgProvider.getOrgContextData(orgId).map { orgProps =>
-            (xformToSnakeCase(userProps ++ orgProps), xformToSnakeCase(eventProps ++ orgProps))
+            (xformToSnakeCase(userProps ++ orgProps), xformToSnakeCase(eventProps))
           }
         case None =>
           Future.successful(xformToSnakeCase(userProps), xformToSnakeCase(eventProps))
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/heimdal/app/com/keepit/heimdal/AmplitudeClient.scala
```diff
@@ -149,7 +149,7 @@ trait AmplitudeRequestBuilder {
       primaryOrgProvider.getPrimaryOrg(userId).flatMap {
         case Some(orgId) =>
           primaryOrgProvider.getOrgContextData(orgId).map { orgProps =>
-            (xformToSnakeCase(userProps ++ orgProps), xformToSnakeCase(eventProps ++ orgProps))
+            (xformToSnakeCase(userProps ++ orgProps), xformToSnakeCase(eventProps))
```
leogrim, line 152: :+1:

kifi-backend/modules/shoebox/angular/src/feed/feed.tpl.html
```diff
@@ -52,6 +52,7 @@ <h4 class="kf-feed-heading">
     </div>
   </section>
   <section class="kf-feed-item" ng-if="!isLoading() && feed.length < 10" >
+  <section class="kf-feed-item" ng-if="!isLoading() && feed.length < 10 && !hasMoreKeeps()" >
```
camhashemi, line 55: @cvializ

cvializ, line 55: nice catch, dude :+1:

leogrim, line 55: Do you need `feed.length < 10` at all?

cvializ, line 55: Yes, otherwise the add-keep box would show up when you have more than ten keeps. Jen only wants it there when you have fewer than ten.

Also @leogrim, the backend doesn't always return as many items as the frontend asks for. Is that expected behavior?

PR #17070, opened 2015-08-28
Added a getBasicKeepsByIds internal endpoint

ryanpbrewster:
Cached for good measure

@camhashemi could you take a look at this? What am I missing?

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepCommander.scala
@@ -59,6 +59,7 @@ object BulkKeepSelection {
 @ImplementedBy(classOf[KeepCommanderImpl])
 trait KeepCommander {
   def idsToKeeps(ids: Seq[Id[Keep]])(implicit session: RSession): Seq[Keep]
+  def getBasicKeeps(ids: Set[Id[Keep]]): Map[Id[Keep], BasicKeep]
   def getKeepsCountFuture(): Future[Int]
   def getKeep(libraryId: Id[Library], keepExtId: ExternalId[Keep], userId: Id[User]): Either[(Int, String), Keep]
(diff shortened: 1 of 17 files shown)


kifi-backend/modules/shoebox/app/com/keepit/controllers/internal/ShoeboxController.scala
```diff
@@ -463,6 +464,14 @@ class ShoeboxController @Inject() (
     Ok(result)
   }
 
+  def getBasicKeepsByIds() = Action(parse.tolerantJson) { request =>
+    val keepIds = request.body.as[Set[Id[Keep]]]
+    val keepDataById = keepCommander.getBasicKeeps(keepIds)
+
+    implicit val tupleWrites = TupleFormat.tuple2Writes[Id[Keep], BasicKeep]
```
maxov, line 471: seems like it would be easier to just do `keepDataById.toMap` instead of this special format

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepCommander.scala
```diff
@@ -134,6 +135,13 @@ class KeepCommanderImpl @Inject() (
     val idToKeepMap = keepRepo.getByIds(ids.toSet)
     ids.map(idToKeepMap)
   }
+  def getBasicKeeps(ids: Set[Id[Keep]]): Map[Id[Keep], BasicKeep] = {
+    db.readOnlyReplica { implicit session =>
+      keepRepo.getByIds(ids).map {
+        case (id, keep) => id -> BasicKeep(keep.externalId, mine = false, removable = false, keep.visibility, Library.publicId(keep.libraryId.get))
```
leogrim, line 141: I don't think you'll be able to cache this one, this is actually a personal view of a keep (is it mine, can I unkeep it), so you can't really cache it globally, it's dependent on the viewer.

ryanpbrewster:
Thanks @camhashemi 

@gratimax, I _think_ this tests properly. I'm gonna head out but I'll be on later. As @leogrim said in Slack I think we should change `BasicKeep` to use `ownerId: ExternalId[User]` instead of `mine: Boolean`. For now I'd say we should add an extra field and just set `mine = false` for this specific use. I need to check which clients use `mine` and where.

maxov:
Superseded by #17072

PR #17074, opened 2015-08-30
add organizationMessageCountCache

camhashemi:
@leogrim @ryanpbrewster @andrewconner 

`eliza.getTotalMessageCountForGroup` is getting called every time we track something and breaking during peak hours. Since we don't need org message counts to be accurate up-to-the-minute, seems like good place for an in-memory cache.


leogrim:
This cache is just supposed to prevent us from going down when there's a burst of activity. An in-memory cache shouldn't be more than a few seconds / minutes. Regarding MemCache, we should either invalidate it properly (maybe asynchronously) or have a shorter duration that gives us the granularity we want, 1 day seems like a long time, what about 1 hour?

camhashemi:
1 hour should work actually, I misunderstood in-memory caching. But eliza is timing out due to load, just trying to reduce that.

If we invalidate the cache every time a message is sent within an org, would we run into trouble in high-frequency situations?

camhashemi:
I think 1 hour should be fine. I can also add proper cache invalidation, but if we invalidate it every time a message is sent within an org, does that get expensive? That's why I thought to only use the TTL.

PR #17075, opened 2015-08-30
Split lib commander

eishay:


eishay:
@ryanpbrewster @leogrim @andrewconner @camhashemi 
I merged it quickly since its very easy to get conflicts with this PR.
This PR is based on comments I already got of other PRs. More splits to come, LibraryCommander file is now about half its prev size, ~900 lines

PR #17076, opened 2015-08-31
scripts to export raw data from mixpanel and import to amplitude

joshm1:
@stkem 

I have a few minor TODOs but for the most part includes separate scripts to download from MP and then import into amplitude. I started with python for the export but realized node was a better fit for the import script because it's almost all I/O.

i've already ran this for July 1-3 which I'll send to @JRuff42  to look over and test.  I'll whip up a script to automate exporting and importing day-by-day to run on arthur after we've given the events I've imported here a good look over.

https://amplitude.com/app/132905/events/summary?start=20150701&end=20150703&i=1&m=totals&vis=line&e=user_searched&e=user_was_notified&e=user_clicked_page&e=user_connected&e=user_viewed_pane

### DIFF
--- a/tools/mixpanel_export/.gitignore
+++ b/tools/mixpanel_export/.gitignore
@@ -1,0 +1,2 @@
+/node_modules/
+*.pyc
--- a/tools/mixpanel_export/README.md
+++ b/tools/mixpanel_export/README.md
@@ -1,0 +1,23 @@
+## Export from Mixpanel
+
+Advise to only download 1 day at a time. An uncompressed file from 1 day can be 200-350+ MB. The script takes 2 arguments that are the start and end dates in YYYY-MM-DD format, inclusive. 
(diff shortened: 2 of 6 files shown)


stkem:
I'm not really familiar with the apis involved here, but the code lgtm. 
Out of curiosity, why/how did we decide to do the data migration ourselves? I was under the impression that amplitude offered to do it for us for a fee (at least they said as much in one of the early calls we had with them)?

joshm1:
Sterling mentioned that to me, but it was a pretty expensive fee and so far, the time I've invested in this part is about 1/4 what amplitude would've cost. I've also included a lot of customizations that helps to have some knowledge about kifi internals.  I'd also be surprised if Amplitude would've done as much of a custom migration as this since we're not just mapping everything 1-to-1.

PR #17078, opened 2015-08-31
Change 'Add Keep' to 'Keep it' for consistency

cvializ:


cvializ:
@andrewconner small one

PR #17081, opened 2015-08-31
Add a shoebox endpoint for basic organizations

maxov:
@andrewconner 
@leogrim 
@ryanpbrewster 

Adds a shoebox endpoint for organization cards, will be cached.

PR #17085, opened 2015-08-31
add organization and library info to keep cards

camhashemi:
@leogrim @ryanpbrewster @andrewconner 

I didn't remove any fields, so the next step is to deploy, test, then refactor the front-end to consume this.


maxov:
Could this wait until after #17081? I'm taking out `OrganizationCard` there, and I don't want more merge conflicts

camhashemi:
np @gratimax

kifi-backend/modules/common/app/com/keepit/model/Library.scala
```diff
@@ -344,8 +344,8 @@ case class BasicLibrary(id: PublicId[Library], name: String, path: String, visib
 }
 
 object BasicLibrary {
-  def apply(library: Library, owner: BasicUser, org: Option[Organization])(implicit publicIdConfig: PublicIdConfiguration): BasicLibrary = {
-    val path = LibraryPathHelper.formatLibraryPath(owner, org.map(_.handle), library.slug)
+  def apply(library: Library, owner: BasicUser, orgHandle: Option[OrganizationHandle])(implicit publicIdConfig: PublicIdConfiguration): BasicLibrary = {
+    val path = LibraryPathHelper.formatLibraryPath(owner, orgHandle, library.slug)
```
leogrim, line 348: It's weird to have a discrepancy here between how the owner and how the organization is treated. Why did you remove the full `Organization`?

leogrim, line 348: Oh I see, ok!

leogrim:
:+1:

kifi-backend/modules/common/app/com/keepit/model/Library.scala
camhashemi, line 348: If you have an `Organization`, you also have an `OrganizationHandle`, but you might only have an `OrganizationHandle` when creating a `BasicLibrary` (as in my case here).

PR #17086, opened 2015-08-31
allow org library creation via the extension

camhashemi:
@ryanpbrewster @leogrim @andrewconner 

(in addition to the feature) consolidates request handling: now mobile, ext, and web handlers use the same code to create libraries.


ryanpbrewster:
lgtm

PR #17087, opened 2015-08-31
Make sure KeepCommander.copyKeeps update Keep.keptAt

leogrim:
@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepCommander.scala
@@ -24,6 +24,7 @@ import com.keepit.normalizer.NormalizedURIInterner
 import com.keepit.search.SearchServiceClient
 import com.keepit.search.augmentation.{ AugmentableItem, ItemAugmentationRequest }
 import com.keepit.typeahead.{ HashtagHit, HashtagTypeahead, TypeaheadHit }
+import org.joda.time.DateTime
 import play.api.http.Status.{ FORBIDDEN, NOT_FOUND }
 import play.api.libs.functional.syntax._
 import play.api.libs.json._
@@ -892,6 +893,8 @@ class KeepCommanderImpl @Inject() (
       userId = userId,
       libraryId = Some(toLibrary.id.get),
       visibility = toLibrary.visibility,
+      organizationId = toLibrary.organizationId,
+      keptAt = currentDateTime,
       source = withSource.getOrElse(k.source),
       originalKeeperId = k.originalKeeperId.orElse(Some(userId))
     )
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -688,64 +688,6 @@ class LibraryCommanderImpl @Inject() (
     }
   }
 
-  // Return is Set of Keep -> error message
-  private def applyToKeeps(userId: Id[User],
```
leogrim, line 692: Removed this, not sure it was a good idea to DRY this out. A `copy` does not modify the original keeps, while a `move` does, probably better to keep them separate.

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepCommander.scala
```diff
@@ -892,24 +893,21 @@ class KeepCommanderImpl @Inject() (
...
-      case Some(existingKeep) if existingKeep.isInactive =>
-        val persistedKeep = persistKeep(newKeep.withId(existingKeep.id.get), Set(userId), Set(toLibrary.id.get))
+      case inactiveKeepOpt =>
+        val persistedKeep = persistKeep(newKeep.copy(id = inactiveKeepOpt.flatMap(_.id)), Set(userId), Set(toLibrary.id.get))
         combineTags(k.id.get, persistedKeep.id.get)
         Right(persistedKeep)
-
-      case Some(existingKeep) =>
-        combineTags(k.id.get, existingKeep.id.get)
-        Left(LibraryError.AlreadyExistsInDest)
```
leogrim, line 912: @ryanpbrewster

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
ryanpbrewster, line 692: Agreed. I did not like that block of code anyways.

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepCommander.scala
ryanpbrewster, line 912: You'd be a great backpacker, Léo. Always leaving things prettier than you found them.

kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepCommander.scala
```diff
@@ -892,24 +893,21 @@ class KeepCommanderImpl @Inject() (
       userId = userId,
       libraryId = Some(toLibrary.id.get),
       visibility = toLibrary.visibility,
+      organizationId = toLibrary.organizationId,
+      keptAt = currentDateTime,
```
ryanpbrewster, line 897: I think @andrewconner normally suggests `clock.now` instead of `currentDateTime`.

PR #17088, opened 2015-08-31
Keep requesting until the response is empty or we have 'limit' items

cvializ:
@andrewconner (@leogrim)

Hopefully this doesn't look too weird and I'm using promises correctly.

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/feed/feedCtrl.js
+++ b/kifi-backend/modules/shoebox/angular/src/feed/feedCtrl.js
@@ -3,15 +3,30 @@
 angular.module('kifi')
 
 .controller('FeedCtrl', [
-  '$rootScope', '$scope', 'feedService', 'Paginator', 'routeService', 'modalService',
-  function($rootScope, $scope, feedService, Paginator, routeService, modalService) {
+  '$rootScope', '$scope', '$q', 'feedService', 'Paginator', 'routeService', 'modalService',
+  function($rootScope, $scope, $q, feedService, Paginator, routeService, modalService) {
     function feedSource(pageNumber, pageSize) {
-      var lastKeep = $scope.feed[(pageNumber * pageSize) - 1] || $scope.feed[$scope.feed.length - 1];
+      var lastKeep = $scope.feed[$scope.feed.length - 1];
 
-      return feedService.getFeed(pageSize, lastKeep && lastKeep.id)
-      .then(function(keepData) {
-        return keepData.keeps;
-      });
+      function tryGetFullPage(limit, streamEnd, deferred, results) {
+        deferred = deferred || $q.defer();
+        results = results || [];
+
+        feedService.getFeed(limit, streamEnd && streamEnd.id).then(function (keepData) {
+          results = results.concat(keepData.keeps);
+
+          if (keepData.keeps.length === 0 || results.length >= limit) {
+            deferred.resolve(results.slice(0, limit));
+          } else { // keepData.length !== 0 && results.length < pageSize
+            tryGetFullPage(limit, results[results.length - 1], deferred, results);
+          }
+        })
+        ['catch'](deferred.reject);
+
+        return deferred.promise;
+      }
+
+      return tryGetFullPage(pageSize, lastKeep);
     }
 
     function libraryKeptTo (keep) {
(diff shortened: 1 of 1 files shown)


cvializ:
@camhashemi, too

kifi-backend/modules/shoebox/angular/src/feed/feedCtrl.js
```diff
@@ -3,15 +3,30 @@
...
+          } else { // keepData.length !== 0 && results.length < pageSize
+            tryGetFullPage(limit, results[results.length - 1], deferred, results);
+          }
+        })
+        ['catch'](deferred.reject);
+
+        return deferred.promise;
+      }
+
+      return tryGetFullPage(pageSize, lastKeep);
```
leogrim, line 29: Why not just display whatever you get and wait for the user to scroll further down to trigger another fetch?

cvializ, line 29: That would be simpler, but Jen spec'd an extra UI element in the feed (the Add Keep box) that shows when users have < 10 entries in the feed. I need to know whether the user has < 10 entries, and I need to get the first page, but I can't know from just one request that there are more than 10 entries, because the `limit` is unreliable.

If we got 5 entries back on the first request, do we show the Add Keep box? The user could have 8 or 20 total entries available. The Add Keep box would show up on the first page, but disappear on the second, which is bad UX.

Do you think this is the right approach given these requirements?

andrewconner, line 29: @leogrim Should the feed endpoint ever give you way less keeps than you ask for? I'd personally be okay just basing the "add keep box" decision on the first batch. Did you notice problems @cvializ?

cvializ, line 29: @camhashemi noticed an issue when his first batch returned 9 elements and it was showing the "Add Keep box", which would disappear suddenly when the user scrolled down to look at it and the next batch autoloaded.

andrewconner, line 29: I see. @leogrim Do we know why that was happening? One way to avoid it is have a larger buffer between the batch size we ask for and the cutoff.

camhashemi, line 29: If we move the box to the top, we can make a decision off the first batch and live with it even if there are more keeps (just not removing it after it's displayed). This isn't to spec, very few people are supposed to see the "add keep box".

As is, the feed endpoint will give you less depending on the number of keeps within a batch that have the same URI (filtering after the limit). We can filter before the limit or bump up the limit number before hand.

andrewconner, line 29: Why can't the backend give how many you asked for regardless? Or at least have a larger batch to account for dupes, and drop them if they're not needed?

cvializ, line 29: @leogrim and I had a Slack conversation earlier:

> Léo Grimaldi [1:59 PM] 
> A limit should be treated as a `limit`, not a guaranteed `size`, and pagination should stop when you get an empty batch, not a partial one
> 
> Léo Grimaldi [2:02 PM]
> It's probably possible to invest time in filtering in one step, but this logic is only going to get more complex and in my experience it's not a good pattern to rely on a limit to decide whether a stream has ended or not
> 
> Léo Grimaldi [2:02 PM]
> better to make one extra call and check for zero

andrewconner, line 29: @leogrim I agree that it shouldn't be guaranteed to be a size, but it complicates things on the frontend when there's no best effort applied (even fetching a couple more internally would fix it almost always).

Either way, fetching more than the cutoff client side may be an easy fix, or keep this. @cvializ

andrewconner, line 29: And true, pagination should stop only when you get an empty one.

cvializ, line 29: This seems like it will be the most consistent solution, despite the extra frontend code. If you think the code seems legible I'll merge it in

camhashemi, line 29: @leogrim @andrewconner would we prefer to make extra calls versus bumping up the limit and dropping any extras? the latter solution seems it would perform better

cvializ:
Closing due to derpy rebase

PR #17089, opened 2015-08-31
simple paid account locking mechanism

stkem:
@andrewconner @ryanpbrewster

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/payments/AccountLockHelper.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/payments/AccountLockHelper.scala
@@ -1,0 +1,31 @@
+package com.keepit.payments
+
+import com.keepit.common.db.slick.Database
+import com.keepit.common.db.Id
+import com.keepit.model.Organization
+import com.keepit.common.db.slick.DBSession.RWSession
+
+import com.google.inject.Inject
+
+class AccountLockHelper @Inject() (db: Database, paidAccountRepo: PaidAccountRepo) {
+
+  def maybeSessionWithAccountLock[T](orgId: Id[Organization])(f: RWSession => T): Option[T] = db.readWrite { implicit session =>
+    val hasLock = try {
+      paidAccountRepo.tryGetAccountLock(orgId)
+    } catch {
+      case ex: org.h2.jdbc.JdbcSQLException if ex.getErrorCode() == 50200 => false //H2 lock aquisition timeout
+      case ex: java.sql.SQLException if ex.getErrorCode() == 1205 => false //MySQL lock aquisition timeout
+    }
+    if (hasLock) {
+      try {
+        Some(f(session))
+      } finally {
+        paidAccountRepo.releaseAccountLock(orgId)
+      }
+    } else {
+      None
+    }
+  }
+
+}
+
--- a/kifi-backend/modules/shoebox/app/com/keepit/payments/PaidAccountRepo.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/payments/PaidAccountRepo.scala
@@ -37,8 +37,8 @@ class PaidAccountRepoImpl @Inject() (
     def credit = column[DollarAmount]("credit", O.NotNull)
     def userContacts = column[Seq[Id[User]]]("user_contacts", O.NotNull)
     def emailContacts = column[Seq[EmailAddress]]("email_contacts", O.NotNull)
-    def lockedForProcessing = column[Boolean]("locked_for_processing", O.Nullable)
     def * = (id.?, createdAt, updatedAt, state, orgId, planId, credit, userContacts, emailContacts, lockedForProcessing) <> ((PaidAccount.apply _).tupled, PaidAccount.unapply _)
+    def lockedForProcessing = column[Boolean]("locked_for_processing", O.NotNull)
   }
 
   def table(tag: Tag) = new PaidAccountTable(tag)
--- a/kifi-backend/modules/shoebox/app/com/keepit/payments/PaidAccountRepo.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/payments/PaidAccountRepo.scala
@@ -16,6 +16,8 @@ trait PaidAccountRepo extends Repo[PaidAccount] {
   def maybeGetByOrgId(orgId: Id[Organization], excludeStates: Set[State[PaidAccount]] = Set(PaidAccountStates.INACTIVE))(implicit session: RSession): Option[PaidAccount]
   def getAccountId(orgId: Id[Organization], excludeStates: Set[State[PaidAccount]] = Set(PaidAccountStates.INACTIVE))(implicit session: RSession): Id[PaidAccount]
   def getActiveByPlan(planId: Id[PaidPlan])(implicit session: RSession): Seq[PaidAccount]
+  def tryGetAccountLock(orgId: Id[Organization])(implicit session: RWSession): Boolean
+  def releaseAccountLock(orgId: Id[Organization])(implicit session: RWSession): Boolean
 }
 
 @Singleton
@@ -35,7 +37,8 @@ class PaidAccountRepoImpl @Inject() (
     def credit = column[DollarAmount]("credit", O.NotNull)
     def userContacts = column[Seq[Id[User]]]("user_contacts", O.NotNull)
     def emailContacts = column[Seq[EmailAddress]]("email_contacts", O.NotNull)
-    def * = (id.?, createdAt, updatedAt, state, orgId, planId, credit, userContacts, emailContacts) <> ((PaidAccount.apply _).tupled, PaidAccount.unapply _)
+    def lockedForProcessing = column[Boolean]("locked_for_processing", O.Nullable)
+    def * = (id.?, createdAt, updatedAt, state, orgId, planId, credit, userContacts, emailContacts, lockedForProcessing) <> ((PaidAccount.apply _).tupled, PaidAccount.unapply _)
   }
 
   def table(tag: Tag) = new PaidAccountTable(tag)
@@ -59,6 +62,14 @@ class PaidAccountRepoImpl @Inject() (
 
   def getActiveByPlan(planId: Id[PaidPlan])(implicit session: RSession): Seq[PaidAccount] = {
     (for (row <- rows if row.planId === planId && row.state === PaidAccountStates.ACTIVE) yield row).list
+  }
+
+  def tryGetAccountLock(orgId: Id[Organization])(implicit session: RWSession): Boolean = {
+    (for (row <- rows if row.orgId === orgId && row.lockedForProcessing =!= true) yield row.lockedForProcessing).update(true) > 0
+  }
+
+  def releaseAccountLock(orgId: Id[Organization])(implicit session: RWSession): Boolean = {
+    (for (row <- rows if row.orgId === orgId && row.lockedForProcessing === true) yield row.lockedForProcessing).update(false) > 0
   }
 
 }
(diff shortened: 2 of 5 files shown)


kifi-backend/modules/shoebox/app/com/keepit/payments/AccountLockHelper.scala
```diff
@@ -0,0 +1,31 @@
...
+import com.google.inject.Inject
+
+class AccountLockHelper @Inject() (db: Database, paidAccountRepo: PaidAccountRepo) {
+
+  def maybeSessionWithAccountLock[T](orgId: Id[Organization])(f: RWSession => T): Option[T] = db.readWrite { implicit session =>
+    val hasLock = try {
+      paidAccountRepo.tryGetAccountLock(orgId)
+    } catch {
+      case ex: org.h2.jdbc.JdbcSQLException if ex.getErrorCode() == 50200 => false //H2 lock aquisition timeout
+      case ex: java.sql.SQLException if ex.getErrorCode() == 1205 => false //MySQL lock aquisition timeout
```
ryanpbrewster, line 17: I personally feel that it's clearer if you have `val H2LockAcquisitionTimeoutCode = 50200` instead of a comment there. No strong opinion, though.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidAccountRepo.scala
```diff
@@ -61,4 +64,12 @@ class PaidAccountRepoImpl @Inject() (
     (for (row <- rows if row.planId === planId && row.state === PaidAccountStates.ACTIVE) yield row).list
   }
 
+  def tryGetAccountLock(orgId: Id[Organization])(implicit session: RWSession): Boolean = {
+    (for (row <- rows if row.orgId === orgId && row.lockedForProcessing =!= true) yield row.lockedForProcessing).update(true) > 0
+  }
+
+  def releaseAccountLock(orgId: Id[Organization])(implicit session: RWSession): Boolean = {
+    (for (row <- rows if row.orgId === orgId && row.lockedForProcessing === true) yield row.lockedForProcessing).update(false) > 0
```
ryanpbrewster, line 72: I'm assuming that the `> 0` is actually checking `== true`. If that is the case, shouldn't this be `== 0` at the end?

Also if that is the case can we just check `== true` or `== 1` instead of `> 0`? That is a semantically confusing comparison.

stkem, line 72: The update statement actually returns that number of rows that were changed (i.e. _not_ a boolean in disguise), so I'm checking here that something was in fact updated (because we are only updating rows that are not currently locked). You are right though, `==1` is probably semantically a bit clearer.

ryanpbrewster, line 72: I'm super relieved to hear that. Since that's the case then I'm fine with it either way.

kifi-backend/modules/shoebox/app/com/keepit/payments/AccountLockHelper.scala
```diff
@@ -0,0 +1,31 @@
...
+import com.keepit.model.Organization
+import com.keepit.common.db.slick.DBSession.RWSession
+
+import com.google.inject.Inject
+
+class AccountLockHelper @Inject() (db: Database, paidAccountRepo: PaidAccountRepo) {
+
+  def maybeSessionWithAccountLock[T](orgId: Id[Organization])(f: RWSession => T): Option[T] = db.readWrite { implicit session =>
+    val hasLock = try {
+      paidAccountRepo.tryGetAccountLock(orgId)
```
andrewconner, line 14: I wish I knew mysql enough to know if this is safe.

kifi-backend/modules/shoebox/app/com/keepit/payments/AccountLockHelper.scala
```diff
@@ -0,0 +1,31 @@
...
+    } catch {
+      case ex: org.h2.jdbc.JdbcSQLException if ex.getErrorCode() == 50200 => false //H2 lock aquisition timeout
+      case ex: java.sql.SQLException if ex.getErrorCode() == 1205 => false //MySQL lock aquisition timeout
+    }
+    if (hasLock) {
+      try {
+        Some(f(session))
+      } finally {
+        paidAccountRepo.releaseAccountLock(orgId)
```
andrewconner, line 23: The `finally` should probably be over the lock acquisition as well, right? Unsure.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidAccountRepo.scala
```diff
@@ -61,4 +64,12 @@ class PaidAccountRepoImpl @Inject() (
     (for (row <- rows if row.planId === planId && row.state === PaidAccountStates.ACTIVE) yield row).list
   }
 
+  def tryGetAccountLock(orgId: Id[Organization])(implicit session: RWSession): Boolean = {
+    (for (row <- rows if row.orgId === orgId && row.lockedForProcessing =!= true) yield row.lockedForProcessing).update(true) > 0
```
andrewconner, line 68: `updated_at`?

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidAccountRepo.scala
andrewconner, line 72: `updated_at`?

kifi-backend/modules/shoebox/app/com/keepit/payments/AccountLockHelper.scala
ryanpbrewster, line 23: I don't think it needs to be. If you have the lock, you're going to hit the `finally`. If you don't have the lock, you shouldn't need to.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidAccountRepo.scala
```diff
@@ -35,7 +37,8 @@ class PaidAccountRepoImpl @Inject() (
     def credit = column[DollarAmount]("credit", O.NotNull)
     def userContacts = column[Seq[Id[User]]]("user_contacts", O.NotNull)
     def emailContacts = column[Seq[EmailAddress]]("email_contacts", O.NotNull)
-    def * = (id.?, createdAt, updatedAt, state, orgId, planId, credit, userContacts, emailContacts) <> ((PaidAccount.apply _).tupled, PaidAccount.unapply _)
+    def lockedForProcessing = column[Boolean]("locked_for_processing", O.Nullable)
```
andrewconner: This will cause problems. type is `Boolean` but it's nullable. Why not make it `NotNull` and set to false?

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidAccountRepo.scala
ryanpbrewster, line 68: I think that anything you do once the row is locked will end up refreshing the `updated_at`, so I'm not sure it's necessary.

I am also not sure it's even a good idea to change `updated_at` if you don't actually update anything (besides the lock).

kifi-backend/modules/shoebox/test/com/keepit/payments/AccountLockHelperTest.scala
```diff
@@ -0,0 +1,52 @@
...
+        val orgCommander = inject[OrganizationCommander]
+
+        inject[PlanManagementCommander].createNewPlan(Name[PaidPlan]("Test"), BillingCycle(1), DollarAmount(0))
+        val user = db.readWrite { implicit session => UserFactory.user().withName("Mr", "Spock").saved }
+        val createRequest = OrganizationCreateRequest(requesterId = user.id.get, OrganizationInitialValues(name = "Kifi"))
+        val createResponse = orgCommander.createOrganization(createRequest)
+        val org = createResponse.right.get.newOrg
+
+        Some(true) === helper.maybeSessionWithAccountLock(org.id.get) { s1 =>
+          None === helper.maybeSessionWithAccountLock(org.id.get) { s2 =>
```
andrewconner, line 35: Should we be testing db-session-in-session? That usually means something incorrect is happening.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidAccountRepo.scala
ryanpbrewster: :+1:

kifi-backend/modules/shoebox/test/com/keepit/payments/AccountLockHelperTest.scala
andrewconner, line 35: (I see that this returns `None` because it is already acquired _in this transaction_, but isn't what we actually want is another thread doing it?)

kifi-backend/modules/shoebox/test/com/keepit/payments/AccountLockHelperTest.scala
```diff
@@ -0,0 +1,52 @@
...
+            true
+          }
+          Await.result(Future {
+            None === helper.maybeSessionWithAccountLock(org.id.get) { s3 =>
+              true
+            }
+          }, Duration.Inf)
+          true
```
andrewconner, line 43: Hm?

kifi-backend/modules/shoebox/app/com/keepit/payments/AccountLockHelper.scala
stkem, line 23: Correct.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidAccountRepo.scala
stkem: Ah, right. Refactor leftover. Thanks for finding.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidAccountRepo.scala
stkem, line 68: I'm with ryan on this. Intentionally not updating updated_at since no actual data is changing.

kifi-backend/modules/shoebox/test/com/keepit/payments/AccountLockHelperTest.scala
stkem, line 43: This is the return of the outermost `maybeSessionWithAccountLock`. Gotta return something.

kifi-backend/modules/shoebox/test/com/keepit/payments/AccountLockHelperTest.scala
stkem, line 35: The two sessions (and thus the two transactions) are actually fully disjoint (note that the inner one uses different names, and no implicits, to the outer session is not captured.). So from the database perspective this is just two transactions that happen to be going on at the same time, but they don't talk to each other, so to speak. This is simply the easiest way to ensure that one transaction is still going on while the other tries to get the lock.
The multi threaded case is right below this.

andrewconner, line 35: Yes, they're disjoint, but break the build when we don't allow session-in-sessions, which I'm working on.

You can ensure two are open at the same time using locks.

andrewconner, line 35: Also, I'm not convinced this works, since the session that takes the lock _isn't committed_. Am I missing something?

stkem, line 35: Yup, you are missing the db row lock. The session that takes the lock starts with an update on the row, which prevents any other transaction from modifying that row at all until the first transaction is committed. This also prevents other threads from force releasing a lock while a transaction is in progress. In essence it doesn't actually matter what the the `locked_for_processing` field contains the way this is set up. If used as intended, i.e. always with `maybeSessionWithAccountLock` other transaction always either see `locked_for_processing=false` or the row is inaccessible entirely.

PR #17090, opened 2015-08-31
Keep integrity

ryanpbrewster:
@stkem @leogrim 

Would y'all mind making sure this integrity plugin looks reasonable? This is the first one I've written.

I want to be sure that it won't bomb the system. I have it configured to process 100 keeps/minute, every minute. I think that will be safe.

This should also effectively back-fill all the keeps that don't have a `librariesHash` or `participantsHash`, since they are in an inconsistent state.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/integrity/DataIntegrity.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/integrity/DataIntegrity.scala
@@ -25,6 +25,7 @@ class DataIntegrityPluginImpl @Inject() (
     scheduleTaskOnOneMachine(actor.system, 7 minutes, EVERY_N_MINUTE minutes, actor.ref, SystemLibraryCheck, getClass.getSimpleName)
     scheduleTaskOnOneMachine(actor.system, 1 minutes, 1 minutes, actor.ref, LibrariesCheck, getClass.getSimpleName)
     scheduleTaskOnOneMachine(actor.system, 10 minutes, 24 hours, actor.ref, PaymentsMembershipCheck, getClass.getSimpleName)
+    scheduleTaskOnOneMachine(actor.system, 1 minutes, 1 minutes, actor.ref, KeepsCheck, getClass.getSimpleName)
   }
 }
 
@@ -32,6 +33,7 @@ private[integrity] case object CleanOrphans
 private[integrity] case object Cron
 private[integrity] case object SequenceNumberCheck
 private[integrity] case object LibrariesCheck
+private[integrity] case object KeepsCheck
 private[integrity] case object SystemLibraryCheck
 private[integrity] case object PaymentsMembershipCheck
 
@@ -40,7 +42,8 @@ private[integrity] class DataIntegrityActor @Inject() (
   orphanCleaner: OrphanCleaner,
   elizaSequenceNumberChecker: ElizaSequenceNumberChecker,
   libraryChecker: LibraryChecker,
-  paymentsChecker: PaymentsIntegrityChecker)
+  paymentsChecker: PaymentsIntegrityChecker,
+  keepChecker: KeepChecker)
     extends FortyTwoActor(airbrake) with Logging {
 
   def receive() = {
@@ -50,6 +53,8 @@ private[integrity] class DataIntegrityActor @Inject() (
       elizaSequenceNumberChecker.check()
     case LibrariesCheck =>
       libraryChecker.check()
+    case KeepsCheck =>
+      keepChecker.check()
     case SystemLibraryCheck =>
       libraryChecker.checkSystemLibraries()
     case PaymentsMembershipCheck =>
--- a/kifi-backend/modules/shoebox/app/com/keepit/integrity/KeepChecker.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/integrity/KeepChecker.scala
@@ -1,0 +1,165 @@
+package com.keepit.integrity
+
+import com.google.inject.Inject
+import com.keepit.commanders.{ KeepCommander, KeepToLibraryCommander, KeepToUserCommander, LibraryInfoCommander }
+import com.keepit.common.core._
+import com.keepit.common.db.slick.DBSession.{ RSession, RWSession }
+import com.keepit.common.db.slick.Database
+import com.keepit.common.db.{ Id, SequenceNumber }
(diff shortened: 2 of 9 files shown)


kifi-backend/modules/shoebox/app/com/keepit/integrity/KeepChecker.scala
```diff
@@ -0,0 +1,167 @@
...
+    ktuCommander: KeepToUserCommander,
+    userRepo: UserRepo,
+    libraryRepo: LibraryRepo,
+    organizationRepo: OrganizationRepo,
+    systemValueRepo: SystemValueRepo) extends Logging {
+
+  private[this] val lock = new AnyRef
+  private val timeSlicer = new TimeSlicer(clock)
+
+  private[integrity] val KEEP_URI_CHANGED = Name[SequenceNumber[Keep]]("keep_integrity_plugin_keep_uri_changed")
```
stkem: Why are all of these private to the package (as opposed to just "regular" private)? They don't seem to be used outside of this class.

kifi-backend/modules/shoebox/app/com/keepit/integrity/KeepChecker.scala
```diff
@@ -0,0 +1,167 @@
...
+import com.keepit.common.core._
+import com.keepit.common.db.slick.DBSession.{ RSession, RWSession }
+import com.keepit.common.db.slick.Database
+import com.keepit.common.db.{ Id, SequenceNumber }
+import com.keepit.common.healthcheck.AirbrakeNotifier
+import com.keepit.common.logging.{ Logging, NamedStatsdTimer }
+import com.keepit.common.time.{ Clock, _ }
+import com.keepit.model._
+
+class KeepChecker @Inject() (
```
stkem, line 17: `@Singleton`? Otherwise your `lock` can't really be relied upon at all times (it'll still work of course if there is only one user of the class and that one is singleton, as is the case here, I think. Introduces some weird implicit contracts though)

kifi-backend/modules/shoebox/app/com/keepit/integrity/KeepChecker.scala
```diff
@@ -0,0 +1,167 @@
...
+  private[integrity] val KEEP_DEACTIVATED = Name[SequenceNumber[Keep]]("keep_integrity_plugin_keep_deactivated")
+  private[integrity] val KEEP_LIBRARY_CHANGED = Name[SequenceNumber[Keep]]("keep_integrity_plugin_keep_library_changed")
+  private[integrity] val KEEP_PARTICIPANT_CHANGED = Name[SequenceNumber[Keep]]("keep_integrity_plugin_keep_participant_changed")
+  private val KEEP_FETCH_SIZE = 100
+
+  private def getLastSeqNum[T](key: Name[SequenceNumber[T]])(implicit session: RSession) = {
+    systemValueRepo.getSequenceNumber(key).getOrElse(SequenceNumber.ZERO[T])
+  }
+
+  def check(): Unit = lock.synchronized {
```
stkem, line 41: Dangerous, given how frequently this runs. If this every takes a little longer you're blocking `DataIntegrityActor`. Maybe use a reactive lock instead, or just wrap it in a future and keep track of running state in the actor or here so you don't get concurrent runs (the former might be slightly easier)

kifi-backend/modules/shoebox/app/com/keepit/integrity/DataIntegrity.scala
```diff
@@ -25,6 +25,7 @@ class DataIntegrityPluginImpl @Inject() (
     scheduleTaskOnOneMachine(actor.system, 7 minutes, EVERY_N_MINUTE minutes, actor.ref, SystemLibraryCheck, getClass.getSimpleName)
     scheduleTaskOnOneMachine(actor.system, 1 minutes, 1 minutes, actor.ref, LibrariesCheck, getClass.getSimpleName)
     scheduleTaskOnOneMachine(actor.system, 10 minutes, 24 hours, actor.ref, PaymentsMembershipCheck, getClass.getSimpleName)
+    scheduleTaskOnOneMachine(actor.system, 1 minutes, 1 minutes, actor.ref, KeepsCheck, getClass.getSimpleName)
```
stkem, line 28: Does it really need to run this often?

leogrim, line 28: It's getting keeps by sequence number though, so the total amount of work is the same, not like if it were checking / scanning the entire table.

kifi-backend/modules/shoebox/app/com/keepit/integrity/KeepChecker.scala
leogrim, line 41: Might be able to use `BatchProcessingActor` for this use case, we can look into it.

leogrim:
I see that the overall approach is to have separate sequence numbers for each specific subcheck. This approach has pros and cons. A pro is that you can reset only one specific check to rerun on the entire table and introduce new ones that can catch up on their own - and you can use this framework for one-time backfilling methods. The con is that in regular circumstances you'll be pulling the same keep several time for each subcheck. The other approach is to have a single integrity sequence number associated with the keep table, so that you pull each changed keep once and check everything for this one keep. Which means that when you introduce a new check, you rerun all checks on the entire table, which may not be a bad thing if you think about it.

What do you think?

ryanpbrewster:
It'd make each check significantly more expensive, although right now I
have it set up to run evey check every time.

I set it up this way because we may at some point find that one field is
much more troublesome than other fields, so we'd run that check more
frequently.

Also, right now 98% of keeps have the wrong hashes. This checker will be
performing a backfill. We may use it in a similar fashion in the future.

kifi-backend/modules/shoebox/app/com/keepit/integrity/KeepChecker.scala
ryanpbrewster: That is a great question. I modeled this off of `LibraryChecker` which had these declarations littered throughout. I'll make them private.

kifi-backend/modules/shoebox/app/com/keepit/integrity/KeepChecker.scala
ryanpbrewster, line 41: We should also revisit `LibraryChecker` since it uses this same approach.

PR #17094, opened 2015-09-01
Added an admin endpoint to backfill permissions

ryanpbrewster:
@andrewconner @stkem

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/admin/AdminOrganizationController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/admin/AdminOrganizationController.scala
@@ -464,19 +464,16 @@ class AdminOrganizationController @Inject() (
     }
   }
 
-  def addPermissionToAllOrganizations() = AdminUserAction.async(parse.tolerantJson) { implicit request =>
+  def addPermissionToAllOrganizations() = AdminUserAction(parse.tolerantJson) { implicit request =>
     implicit val reads = KeyFormat.key2Reads[OrganizationPermission, String]("permission", "confirmation")
     val (newPermission, confirmation) = request.body.as[(OrganizationPermission, String)]
     assert(confirmation == "i swear i know what i am doing", "admin does not know what they are doing")
     val orgIds = db.readOnlyMaster { implicit session => orgRepo.all.filter(_.isActive).map(_.id.get) }
-    Future {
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/common/app/com/keepit/model/Organization.scala
```diff
@@ -39,9 +39,7 @@ case class Organization(
   def withName(newName: String): Organization = this.copy(name = newName)
   def withDescription(newDescription: Option[String]): Organization = this.copy(description = newDescription)
   def withOwner(newOwner: Id[User]): Organization = this.copy(ownerId = newOwner)
-  def withBasePermissions(newBasePermissions: BasePermissions): Organization = {
-    this.copy(basePermissions = new BasePermissions(basePermissions.permissionsMap ++ newBasePermissions.permissionsMap))
-  }
+  def withBasePermissions(newBasePermissions: BasePermissions): Organization = this.copy(basePermissions = newBasePermissions)
```
stkem: why the change in semantics?

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminOrganizationController.scala
```diff
@@ -463,4 +464,19 @@ class AdminOrganizationController @Inject() (
     }
   }
 
+  def addPermissionToAllOrganizations() = AdminUserAction.async(parse.tolerantJson) { implicit request =>
+    implicit val reads = KeyFormat.key2Reads[OrganizationPermission, String]("permission", "confirmation")
+    val (newPermission, confirmation) = request.body.as[(OrganizationPermission, String)]
+    assert(confirmation == "i swear i know what i am doing", "admin does not know what they are doing")
+    val orgIds = db.readOnlyMaster { implicit session => orgRepo.all.filter(_.isActive).map(_.id.get) }
+    Future {
```
stkem: Why the `Future`? It's in the same thread pool that you are already running in.

kifi-backend/modules/common/app/com/keepit/model/Organization.scala
ryanpbrewster: I was concerned that the `BasePermissions` object was being used without
defined permissions for every role. In the another PR (permissions diff) I
address this problem more directly.

PR #17095, opened 2015-09-01
keepInfo changes should be backwards compatible

camhashemi:
@andrewconner @leogrim everything else added defaults to None, so this should be it


leogrim:
Yep that's the issue.

PR #17101, opened 2015-09-01
Add owner id to BasicLibraryDetails

maxov:
@leogrim 
@ryanpbrewster 

Notifications need this in order to do an extra hop to get the user info

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/model/Library.scala
+++ b/kifi-backend/modules/common/app/com/keepit/model/Library.scala
@@ -8,6 +8,7 @@ import com.keepit.common.cache.{ CacheStatistics, FortyTwoCachePlugin, JsonCache
 import com.keepit.common.crypto.{ PublicId, PublicIdConfiguration, ModelWithPublicId, ModelWithPublicIdCompanion }
 import com.keepit.common.db._
 import com.keepit.common.logging.AccessLog
+import com.keepit.common.path.Path
 import com.keepit.common.strings.UTF8
 import com.keepit.common.time._
 import com.keepit.model.view.LibraryMembershipView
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/common/app/com/keepit/model/Library.scala
```diff
@@ -370,8 +370,9 @@ case class BasicLibraryDetails(
   numFollowers: Int,
   numCollaborators: Int,
   keepCount: Int,
-  membership: Option[LibraryMembership] // viewer
-  )
+  membership: Option[LibraryMembership], // viewer
+  ownerId: Id[User],
+  url: String)
```
leogrim: I'm not sure what ends up in this but I don't think we pass the library url around anywhere. If we do we shouldn't, better pass a typed `Path`

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryInfoCommander.scala
```diff
@@ -141,7 +141,8 @@ class LibraryInfoCommanderImpl @Inject() (
         val numCollaborators = counts.readWrite
         val imageOpt = libraryImageCommander.getBestImageForLibrary(libId, idealImageSize).map(libraryImageCommander.getUrl)
         val membership = membershipsByLibraryId.get(libId).flatten
-        libId -> BasicLibraryDetails(lib.name, lib.slug, lib.color, imageOpt, lib.description, numFollowers, numCollaborators, lib.keepCount, membership)
+        val url = libPathCommander.getPathForLibrary(lib)
```
leogrim: Actually `getPathForLibrary` seems deprecated

PR #17108, opened 2015-09-01
include keepInfo.library in tests

camhashemi:
@ryanpbrewster can we kill all of these asap


ryanpbrewster:
What is even happening in this PR?

PR #17109, opened 2015-09-01
Membership related Account Event integrity checker

stkem:
@andrewconner @ryanpbrewster


kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationRepo.scala
```diff
@@ -134,6 +135,11 @@ class OrganizationRepoImpl @Inject() (
     getByIds(orgIds).values.toSeq
   }
 
+  def getIdSubsetByModulus(modulus: Int, partition: Int)(implicit session: RSession): Set[Id[Organization]] = {
+    import com.keepit.common.db.slick.StaticQueryFixed.interpolation
+    sql"""select id from organization where state='active' and MOD(id, $modulus)=$partition""".as[Id[Organization]].list.toSet
```
ryanpbrewster, line 140: Is this a good idea? Doesn't this require a full table scan?

kifi-backend/modules/shoebox/app/com/keepit/payments/IntegrityChecker.scala
```diff
@@ -0,0 +1,128 @@
...
+import com.keepit.common.db.Id
+import com.keepit.common.time._
+import com.keepit.common.logging.Logging
+
+import scala.util.{ Try, Success, Failure }
+
+import com.google.inject.{ Singleton, Inject }
+
+@Singleton
+class PaymentsIntegrityChecker @Inject() (
```
ryanpbrewster, line 21: Should we name this file `PaymentsIntegrityChecker` just so it matches the class name?

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidAccount.scala
```diff
@@ -29,11 +29,13 @@ case class PaidAccount(
...
     userContacts: Seq[Id[User]],
     emailContacts: Seq[EmailAddress],
-    lockedForProcessing: Boolean = false) extends ModelWithState[PaidAccount] {
+    lockedForProcessing: Boolean = false,
+    frozen: Boolean = false) extends ModelWithState[PaidAccount] {
 
   def withId(id: Id[PaidAccount]): PaidAccount = this.copy(id = Some(id))
   def withUpdateTime(now: DateTime): PaidAccount = this.copy(updatedAt = now)
   def withState(state: State[PaidAccount]): PaidAccount = this.copy(state = state)
+  def freeze: PaidAccount = this.copy(frozen = true) //a frozen account will not be charged anything by the payment processor until unfrozen by an admin. Intended for automatically detected data integrity issues.
```
ryanpbrewster, line 38: Maybe `frozen`, to follow the convention of "past tense description".

ryanpbrewster:
Semantically, what does freezing an account do?

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidAccount.scala
stkem, line 38: Can't do, that is already the name of the field on the class. It's a conflicting overload.

stkem:
@ryanpbrewster see the comment on the `freeze` method (you actually commented on that line): It's stops the payment processor from doing anything for that account (i.e. their functionality stays intact, but they won't be charged anything) until an admin unfreezes the account. Mostly to be used when we automatically detect some discrepancies with the data that we wan't a human to look at before continuing lest we over or under charge a customer.

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationRepo.scala
stkem, line 140: Yes, it does indeed require a full table scan, but the organization table is relatively small and will remain so relative to other entities in the system (and thus relative to the power of the database server) and this only runs once a day. For context, the query currently takes ~100-120ms on average.

ryanpbrewster, line 140: And you're actually returning `O(n)` ids anyways, so it's not like an index is going to make that part faster.

In other integrity plugins, we only check entities that have been modified (as indicated by their sequence number). Why the difference here? Just to be absolutely certain?

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidAccount.scala
ryanpbrewster, line 38: Ah, yes, my mistake.

ryanpbrewster:
Have you implemented the logic that actually makes the `frozen` field do anything?

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationRepo.scala
stkem, line 140: Yeah, basically. This is not relaying on any other mechanism to work correctly (like sequence numbers). Plus just checking everything makes the logic quite a bit easier to grok.

kifi-backend/modules/shoebox/app/com/keepit/payments/IntegrityChecker.scala
stkem, line 21: Yes, we should.

stkem:
Nope, not yet. I have parts of the payment processor, but it's not honoring the frozen field yet. The need for this really only occurred to me today.

ryanpbrewster:
:+1:. Mmmk, just wanted to be sure I understand.

kifi-backend/modules/shoebox/test/com/keepit/model/OrganizationFactoryHelper.scala
```diff
@@ -4,10 +4,12 @@ import com.google.inject.Injector
 import com.keepit.commanders.HandleCommander
 import com.keepit.common.db.slick.DBSession.RWSession
 import com.keepit.model.OrganizationFactory.PartialOrganization
+import com.keepit.payments.{ PlanManagementCommander, PaidPlan, BillingCycle, DollarAmount }
 
 object OrganizationFactoryHelper {
   implicit class OrganizationPersister(partialOrganization: PartialOrganization) {
     def saved(implicit injector: Injector, session: RWSession): Organization = {
+      injector.getInstance(classOf[PlanManagementCommander]).createNewPlan(Name[PaidPlan]("Test"), BillingCycle(1), DollarAmount(0))
```
andrewconner, line 12: This is a session-in-session (createNewPlan creates it's own session, and `saved` here takes an implicit one). @stkem

PR #17112, opened 2015-09-01
Set height so email user icon isn't an oval

cvializ:
@andrewconner (@leo) resolves https://app.asana.com/0/42816489233169/46491013594142/f


cvializ:
screen shot 2015-09-01 at 3 36 04 pm

before and after

leogrim:
Yeah!

PR #17119, opened 2015-09-02
Modifying an org's permissions happens via a diff

ryanpbrewster:
Moved permissions into their own file, fixed tests

@leogrim @stkem


stkem:
:+1:

PR #17120, opened 2015-09-02
Feed content tracking

camhashemi:
@andrewconner @cvializ 

adds tracking to keep views in the feed, in libraries, and search. next step is to augment the context on the back-end.


cvializ:
:+1: looks good to me

PR #17122, opened 2015-09-02
Take out shared behavior for user images and add it to Eliza

maxov:
@ryanpbrewster 
@leogrim 
@andrewconner 
@stkem 

This is needed by eliza, right before it sends out notifications in order to get user image urls. There was some behavior in `S3ImageStore` that was only in shoebox when eliza could also use it, so I moved that to common and added a class in Eliza to provide the same service.

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/common/db/ExternalId.scala
+++ b/kifi-backend/modules/common/app/com/keepit/common/db/ExternalId.scala
@@ -46,27 +46,6 @@ object ExternalId {
     }
   }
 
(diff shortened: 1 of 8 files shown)


andrewconner:
:+1:

maxov:
Turns out that the endpoint was unnecessary, the only thing we needed was to move some shared behavior over to Eliza where it could be used instead of making a cross-service call. I replaced the endpoint with the shared image url store.

stkem:
:+1:

PR #17124, opened 2015-09-02
Eliza side of org messaging

andrewconner:
@stkem This should be it on the eliza side. What we're doing is sending down orgs as "users" in the typeahead, and then the clients think it's just a regular user id. They then send it to eliza when sending a message or add participants, who figures out what was meant.

It's relatively short-term, since longer term, you'd "message" a library (which has the people you want to share with). But until then (probably month++ away), this is here for current org people.


andrewconner:
Got a Slack go ahead.

kifi-backend/modules/eliza/app/com/keepit/eliza/commanders/MessagingCommander.scala
```diff
@@ -648,9 +670,36 @@ class MessagingCommander @Inject() (
   def validateEmailContacts(rawNonUsers: Seq[JsValue]): Seq[JsResult[BasicContact]] = rawNonUsers.map(_.validate[JsObject].map {
     case obj if (obj \ "kind").as[String] == "email" => (obj \ "email").as[BasicContact]
   })
-  def validateRecipients(rawRecipients: Seq[JsValue]): (Seq[JsResult[ExternalId[User]]], Seq[JsResult[BasicContact]]) = {
-    val (rawUsers, rawNonUsers) = rawRecipients.partition(_.asOpt[JsString].isDefined)
-    (validateUsers(rawUsers), validateEmailContacts(rawNonUsers))
+
+  def parseRecipients(rawRecipients: Seq[JsValue]): (Seq[ExternalId[User]], Seq[BasicContact], Seq[PublicId[Organization]]) = {
```
stkem, line 674: wow, that is one hard to read method

PR #17126, opened 2015-09-02
Shoebox side of org messaging, part 1, only for me

andrewconner:
@stkem Need to check performance of this, so only for me.

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/common/concurrent/WatchableExecutionContext.scala
+++ b/kifi-backend/modules/common/app/com/keepit/common/concurrent/WatchableExecutionContext.scala
@@ -21,9 +21,9 @@ class WatchableExecutionContext(mode: Mode.Mode) extends ScalaExecutionContext {
   private[this] val lock = new Object()
   @volatile private[this] var addedAnything = false
   @volatile private[this] var counter: Int = 0
-  @volatile private[this] var maxExeutionCount: Int = 0
+  @volatile private[this] var maxExecutionCount: Int = 0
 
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/common/app/com/keepit/common/concurrent/WatchableExecutionContext.scala
```diff
@@ -31,18 +31,18 @@ class WatchableExecutionContext(mode: Mode.Mode) extends ScalaExecutionContext {
...
-      maxExeutionCount += 1
+      maxExecutionCount += 1
       addedAnything = true
       val wrapper = new Runnable {
         override def run(): Unit = try {
           runnable.run()
           lock.synchronized {
+            if (counter <= 1) lock.notifyAll()
             counter -= 1
-            if (counter < 0) throw new Exception(s"Counter should never be less then zero")
```
stkem, line 41: Why remove this? Seems like a good integrity check.

stkem:
:+1:

kifi-backend/modules/common/app/com/keepit/common/concurrent/WatchableExecutionContext.scala
andrewconner, line 41: Was talking with Eishay about this class; it actually was broken before (`kill` returned the wrong thing). Written as it is in this PR, it's not really possible to get off counting (`counter` is handled last in the normal clause, first in the catch), but even if it could, it's for testing only, and you should never throw an exception inside an EC. (I also grepped a lot of logs to be sure, this hasn't happened on jenkins recently)

PR #17127, opened 2015-09-02
user_viewed_content event augmentation

camhashemi:
@leogrim @joshm1 @eishay 

previously, we were augmenting org tracking events in `AmplitudeClient`, but seems like the protocol is to use an `EventAugmentor`  in `EventTrackingController`. So I refactored and added https://app.asana.com/0/32728208409723/44164549287870 for keeps.


kifi-backend/modules/common/app/com/keepit/common/routes/Router.scala
```diff
@@ -168,6 +168,8 @@ object Shoebox extends Service {
     def getOrgTrackingValues(orgId: Id[Organization]) = ServiceRoute(GET, "/internal/shoebox/database/getOrgTrackingValues", Param("orgId", orgId))
     def getBasicKeepsByIds() = ServiceRoute(POST, "/internal/shoebox/database/getBasicKeepsByIds")
     def getBasicOrganizationsByIds() = ServiceRoute(POST, "/internal/shoebox/database/getBasicOrganizations")
+    def getLibraryMembershipView(libraryId: Id[Library], userId: Id[User]) = ServiceRoute(GET, "/internal/shoebox/database/getLibraryMembershipView", Param("libraryId", libraryId), Param("userId", userId))
+    def getOrganizationUserRelationship(orgId: Id[Organization], userId: Id[User]) = ServiceRoute(GET, "/internal/shoebox/database/getOrganizationMembershipView", Param("orgId", orgId), Param("userId", userId))
```
leogrim: Method name doesn't match the route

kifi-backend/modules/heimdal/app/com/keepit/model/UserEventRepos.scala
```diff
@@ -73,6 +79,40 @@ object UserIdAugmentor extends EventAugmentor[UserEvent] {
...
+    }
+    val libPropsFut = libIdOpt match {
+      case Some(libraryId) => eventContextHelper.getLibraryEventValues(libraryId, userId)
+      case None => Future.successful(Seq.empty[(String, ContextData)])
+    }
+
+    for (orgProps <- orgPropsFut; libProps <- libPropsFut) yield orgProps ++ libProps
+  }
+}
+
```
leogrim, line 115: If this is defined for a single event, why use an augmentor instead of adding these fields in when the event is sent?

kifi-backend/modules/common/app/com/keepit/model/OrganizationMembership.scala
```diff
@@ -88,6 +89,15 @@ object IngestableOrganizationMembership {
...
 
+@json
+case class OrganizationUserRelationship(
+  orgId: Id[Organization],
+  userId: Id[User],
+  role: Option[OrganizationRole],
+  permissions: Option[Set[OrganizationPermission]],
+  isInvited: Boolean,
+  isCandidate: Boolean)
+
```
leogrim, line 100: `BasicOrganizationView` ? The "View" part indicating that this is from the perspective of a specific user?

kifi-backend/modules/heimdal/app/com/keepit/model/UserEventRepos.scala
camhashemi, line 115: This is meant to augment client-sent events. They don't know the relationship between an org/library and a user from wherever a keep may be clicked.

kifi-backend/modules/common/app/com/keepit/model/OrganizationMembership.scala
camhashemi, line 100: Seems like `View` usually indicates a subset of a model's fields, whereas this one aggregates three different models (hence the nonspecific name). It also shouldn't be likened to `BasicOrganization`.

PR #17136, opened 2015-09-03
Feature/ext org priv

cvializ:
@andrewconner @adamjgrant 

| Non-org-user | Org-user, selected a personal space | Org-user, selected an org space | Org-user, private |
| --- | --- | --- | --- |
| screen shot 2015-09-03 at 11 14 47 am | screen shot 2015-09-03 at 10 54 09 am | screen shot 2015-09-03 at 10 54 25 am | screen shot 2015-09-03 at 11 21 09 am |


adamjgrant:
Epic!

adamjgrant:
Actually, could we color the text as well?

cvializ:
Totally, the same shade? I couldn't tell if that would seem too cartoony.

adamjgrant:
Yeah, that might...What if we kept the font weight the same but made the color darker? Either way, whatever you decide on is fine.

cvializ:
@adam, I think this looks OK, but I'm still open to just gray+bold text when selected. With the color it's bold enough to still seem informative
screen shot 2015-09-03 at 12 39 07 pm

PR #17137, opened 2015-09-03
send user invites when email invites can be converted

camhashemi:
@leogrim @ryanpbrewster not sure how to write a test for this one, since it depends on tables in abook


ryanpbrewster:
Does it use a cache? If so you can just insert the things into the Shoebox-side cache.

camhashemi:
it uses a mock API, let's see if I can play with it

camhashemi:
`FakeAbookServiceClient.internKifiContacts` is the API we need to hack (such that it returns a `userId = Some(id)`, but there's not a super clean way to do it from what I can see.

PR #17141, opened 2015-09-03
Styling for orgs in list, adding pictures to token.

andrewconner:
@adamjgrant @cvializ 

The change:

### DIFF
--- a/packrat/styles/keeper/compose.less
+++ b/packrat/styles/keeper/compose.less
@@ -382,7 +382,7 @@ span.kifi-compose-tip-alt {
     left: 10px;
     height: 34px;
     width: 34px;
-    border-radius: 50% !important;
+    border-radius: 50%;
     background: 0 0/cover no-repeat;
     background-image: inherit;
     box-shadow: 0 0 2px rgba(0,0,0,.2) inset;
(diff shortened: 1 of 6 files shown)


packrat/styles/keeper/compose.less
```diff
@@ -387,10 +387,31 @@ span.kifi-compose-tip-alt {
...
+      position: absolute;
+      top: 10px;
+      left: 10px;
+      height: 34px;
+      width: 34px;
+      background: 0 0/cover no-repeat;
+      box-shadow: 0 0 2px rgba(0,0,0,.2) inset;
       background-size: 20px auto;
       background-position: 7px 8px;
+      border-radius: 50% !important;
```
cvializ: the `!important`s make me sad, but the extension does some weird things with specificity, so if it needs them I will understand

cvializ:
:+1:

adamjgrant:
What @cvializ said, and :+1:

packrat/styles/keeper/compose.less
andrewconner: Me too. It was there originally, and I needed to duplicate some code here, so I copied them. I'll be not lazy and test it out without and see what breaks.

andrewconner: You were right, I can't see a need. Removed.

cvializ:
now I give :+1: :+1:

PR #17142, opened 2015-09-03
killing curator replica

eishay:


eishay:
@andrewconner @stkem

stkem:
:+1:

PR #17149, opened 2015-09-03
Update curator.j2

eishay:
@joshm1


eishay:
@joshm1 how to I apply this change?

PR #17153, opened 2015-09-03
allow admins to see fake users in org invite typeahead

camhashemi:
@leogrim thanks for the suggestion

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/admin/AdminUserController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/admin/AdminUserController.scala
@@ -919,6 +919,7 @@ class AdminUserController @Inject() (
     val ownedLibraries = libraryRepo.getAllByOwner(userId).map(_.id.get)
     val ownsCollaborativeLibs = libraryMembershipRepo.getCollaboratorsByLibrary(ownedLibraries.toSet).exists { case (_, collaborators) => collaborators.size > 1 }
     assert(!ownsCollaborativeLibs, "cannot deactivate a user if they own a library with collaborators: either delete the library or transfer its ownership")
+    ownedLibraries.foreach(id => libraryCommander.deleteLibrary(id, userId)) // Libraries
 
     // Personal Info
     userSessionRepo.invalidateByUser(userId) // User Session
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
```diff
@@ -423,7 +423,7 @@ class OrganizationInviteCommanderImpl @Inject() (db: Database,
         val nonMembers = db.readOnlyMaster { implicit session =>
           val members = organizationMembershipRepo.getByOrgIdAndUserIds(orgId, users.toSet)
           val fakes = userCommander.get().getAllFakeUsers()
-          users.filter(id => !members.exists(_.userId == id) && !fakes.contains(id))
+          users.filter(id => !members.exists(_.userId == id) && (!fakes.contains(id) || userExperimentRepo.hasExperiment(userId, UserExperimentType.ADMIN)))
```
ryanpbrewster: Let's compute this once and reuse the result. I think this will hit the db like 20 times per typeahead as written.

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminUserController.scala
```diff
@@ -919,6 +919,7 @@ class AdminUserController @Inject() (
     val ownedLibraries = libraryRepo.getAllByOwner(userId).map(_.id.get)
     val ownsCollaborativeLibs = libraryMembershipRepo.getCollaboratorsByLibrary(ownedLibraries.toSet).exists { case (_, collaborators) => collaborators.size > 1 }
     assert(!ownsCollaborativeLibs, "cannot deactivate a user if they own a library with collaborators: either delete the library or transfer its ownership")
+    ownedLibraries.foreach(id => libraryCommander.deleteLibrary(id, userId)) // Libraries
```
ryanpbrewster, line 923: :+1: :+1: :+1: :+1: :+1:

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationInviteCommander.scala
camhashemi: only if the user is not a member and also a fake user

I think you're right though.

ryanpbrewster: Fair. Still, it's an easy change.

camhashemi: better to guarantee only one query than potentially very many

PR #17157, opened 2015-09-04
add an admin endpoint to remove libraries with inactive owners

camhashemi:
@ryanpbrewster @gratimax

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/admin/AdminLibraryController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/admin/AdminLibraryController.scala
@@ -363,4 +363,11 @@ class AdminLibraryController @Inject() (
     }
   }
 
+  def removeLibrariesWithInactiveOwner = AdminUserAction { implicit request =>
+    val libsAndOwners = db.readOnlyMaster { implicit session => libraryRepo.getLibrariesWithInactiveOwnerMemberships() }
+    implicit val context = HeimdalContext.empty
+    libsAndOwners.foreach { case (libId, userId) => libraryCommander.deleteLibrary(libId, userId) }
+    Ok
+  }
+
 }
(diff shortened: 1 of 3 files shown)


camhashemi:
just merging for sake of time, won't call this until I get a thumbs up

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryRepo.scala
```diff
@@ -514,12 +517,18 @@ class LibraryRepoImpl @Inject() (
...
       val inIds = owners.map { _.id }.mkString("(", ",", ")")
-      val q = sql"""select owner_id, count(*) from library  where owner_id in #${inIds} and kind = 'user_created' and state = 'active' and visibility = 'published' group by owner_id"""
+      val q = sql"""select owner_id, count(*) from library where owner_id in #${inIds} and kind = 'user_created' and state = 'active' and visibility = 'published' group by owner_id"""
 
       val cnts = q.as[(Int, Int)].list.map { case (userId, count) => Id[User](userId) -> count }.toMap
       owners.map { user => user -> cnts.getOrElse(user, 0) }.toMap
     }
   }
+
+  def getLibrariesWithInactiveOwnerMemberships()(implicit session: RSession): Seq[(Id[Library], Id[User])] = {
```
ryanpbrewster, line 527: The name is a bit misleading. You're looking for libraries with inactive owners. You're not really looking at their memberships.

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminLibraryController.scala
```diff
@@ -363,4 +363,11 @@ class AdminLibraryController @Inject() (
     }
   }
 
+  def removeLibrariesWithInactiveOwner = AdminUserAction { implicit request =>
+    val libsAndOwners = db.readOnlyMaster { implicit session => libraryRepo.getLibrariesWithInactiveOwnerMemberships() }
+    implicit val context = HeimdalContext.empty
+    libsAndOwners.foreach { case (libId, userId) => libraryCommander.deleteLibrary(libId, userId) }
```
ryanpbrewster, line 369: This seems really dangerous. I'm hesitant to call methods via `userId` if the model referenced by the id is inactive.

ryanpbrewster, line 369: We can maybe have a method `libraryCommander.unsafeDeleteLibrary(libraryId: Id[Library])` that

kifi-backend/modules/shoebox/app/com/keepit/model/LibraryRepo.scala
camhashemi, line 527: Yup

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminLibraryController.scala
camhashemi, line 369: That's a good call. Also going to add asynchronous batching.

PR #17158, opened 2015-09-04
Add teaser tooltip for non-orgs users.

cvializ:
@andrewconner @adamjgrant


cvializ:
screen shot 2015-09-04 at 8 49 30 am

PR #17160, opened 2015-09-04
properly filter invite typeahead

camhashemi:
@ryanpbrewster thanks for being my back-up brain during some downtime


ryanpbrewster:
:+1:

PR #17164, opened 2015-09-04
Delete zombie libraries

camhashemi:
@ryanpbrewster @gratimax

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/admin/AdminLibraryController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/admin/AdminLibraryController.scala
@@ -373,4 +373,9 @@ class AdminLibraryController @Inject() (
     Ok
   }
 
+  def getLibrariesWithInactiveOwner = AdminUserAction { implicit request =>
+    val libIds = db.readOnlyMaster { implicit session => libraryRepo.getLibrariesWithInactiveOwner() }
+    Ok(Json.obj("ids" -> libIds, "count" -> libIds.length))
+  }
+
 }
--- a/kifi-backend/modules/shoebox/app/com/keepit/admin/AdminLibraryController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/admin/AdminLibraryController.scala
@@ -19,6 +19,7 @@ import com.keepit.search.SearchServiceClient
 import org.apache.commons.lang3.RandomStringUtils
 import play.api.libs.concurrent.Execution.Implicits._
 import play.api.libs.json.Json
+import play.api.mvc.Action
 import views.html
 
 import scala.concurrent.Future
@@ -364,7 +365,7 @@ class AdminLibraryController @Inject() (
     }
   }
 
-  def removeLibrariesWithInactiveOwner = AdminUserAction { implicit request =>
+  def removeLibrariesWithInactiveOwner() = Action { request =>
     // delete all libraries with an inactive owner: no exceptions for collaborative or system libraries
 
     val libIds = db.readOnlyMaster { implicit session => libraryRepo.getLibrariesWithInactiveOwner() }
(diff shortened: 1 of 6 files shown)


kifi-backend/modules/shoebox/app/com/keepit/admin/AdminLibraryController.scala
```diff
@@ -364,9 +365,11 @@ class AdminLibraryController @Inject() (
   }
 
   def removeLibrariesWithInactiveOwner = AdminUserAction { implicit request =>
```
maxov: Maybe you want to add an endpoint to get all libraries that we will delete and return them. That way we can check what will happen before it happens.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -397,6 +398,39 @@ class LibraryCommanderImpl @Inject() (
...
+        libraryInviteRepo.save(invite.withState(LibraryInviteStates.INACTIVE))
+      }
+    }
+
+    val deletedKeepsFut = db.readWriteAsync { implicit session =>
+      keepRepo.getByLibrary(libraryId, 0, Int.MaxValue).foreach(keepCommander.deactivateKeep)
+      searchClient.updateKeepIndex()
+    }
+
+    val deletedLibraryFut = db.readWriteAsync { implicit session =>
```
maxov: You may be deleting the library here before anything else gets deleted which will probably cause integrity constraint violations. You want to move the future into the for-comprehension so that it definitely gets executed _after_ all the other deletions.

camhashemi: good catch, thanks

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminLibraryController.scala
camhashemi: :+1: good way to test the query

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminLibraryController.scala
```diff
@@ -364,10 +365,17 @@ class AdminLibraryController @Inject() (
...
+    val libIds = db.readOnlyMaster { implicit session => libraryRepo.getLibrariesWithInactiveOwner() }
+    FutureHelpers.sequentialExec(libIds)(libraryCommander.unsafeAsyncDeleteLibrary)
+
     Ok
   }
 
+  def getLibrariesWithInactiveOwner = AdminUserAction { implicit request =>
+    val libIds = db.readOnlyMaster { implicit session => libraryRepo.getLibrariesWithInactiveOwner() }
+    Ok(Json.obj("ids" -> libIds, "count" -> libIds.length))
+  }
```
camhashemi, line 380: @gratimax

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -397,6 +398,36 @@ class LibraryCommanderImpl @Inject() (
...
+
+    for {
+      deletedMembers <- deletedMembersFut
+      deletedInvites <- deletedInvitesFut
+      deletedKeeps <- deletedKeepsFut
+    } yield {
+      db.readWriteAsync { implicit session =>
+        libraryRepo.save(libraryRepo.get(libraryId).withState(LibraryStates.INACTIVE))
+        searchClient.updateLibraryIndex()
+      }
```
camhashemi, line 427: @gratimax

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminLibraryController.scala
maxov, line 380: :+1: may also want to return the owner ids

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
maxov, line 427: :+1:, I would also make sure that there aren't any dependencies between  deletedMembersFut, deletedInvitesFut, and deletedKeepsFut

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminLibraryController.scala
camhashemi, line 380: Afaik you can't do much with a response like this (we don't have endpoints to access everything you'd want). This is just a sanity check to make sure the query being used matches the query we want. For investigation, we can/should query the DB manually.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
camhashemi, line 427: :+1:

PR #17170, opened 2015-09-04
Refactor BasicUserLikeEntity into a type synonym of Either and associated extractors

maxov:
@andrewconner 
@stkem 
@ryanpbrewster 

BasicUserLikeEntity had a rather strange hierarchy, being the supertype of both BasicUser and BasicNonUser (which had practically nothing to do with each other). This refactors that type into a synonym for Either as well as some convenience extractor methods to make intentions more clear (`BasicUserLikeEntity.user(u)` instead of `Right(u)`). This makes implementing the new notifications system a bit easier, since it is more clear what is happening during a pattern-match. As an additional bonus, it's possible to use all the Either methods on "BasicUserLikeEntities" now, but I did not convert existing code to do that.

### DIFF
--- a/kifi-backend/modules/eliza/app/com/keepit/eliza/model/MessagingIndexCommander.scala
+++ b/kifi-backend/modules/eliza/app/com/keepit/eliza/model/MessagingIndexCommander.scala
@@ -7,6 +7,7 @@ import com.keepit.common.db.slick.Database
 import com.keepit.model.User
 import com.keepit.shoebox.ShoeboxServiceClient
 import com.google.inject.Inject
+import com.keepit.social.BasicUserLikeEntity
 
 import scala.concurrent.Future
 
@@ -35,6 +36,7 @@ class MessagingIndexCommander @Inject() (
     val userParticipants: Seq[Id[User]] = thread.participants.map(_.allUsers).getOrElse(Set[Id[User]]()).toSeq
     val participantBasicUsersFuture = shoebox.getBasicUsers(userParticipants)
     val participantBasicNonUsers = thread.participants.map(_.allNonUsers).getOrElse(Set.empty).map(NonUserParticipant.toBasicNonUser)
+      .map(BasicUserLikeEntity.apply)
 
     val messages: Seq[Message] = db.readOnlyReplica { implicit session =>
       messageRepo.get(threadId, 0)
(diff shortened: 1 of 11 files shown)


kifi-backend/modules/eliza/app/com/keepit/eliza/model/MessagingIndexCommander.scala
```diff
@@ -67,7 +69,7 @@ class MessagingIndexCommander @Inject() (
         mode = FULL,
         id = Id[ThreadContent](threadId.id),
         seq = seq,
-        participants = participantBasicUsers.values.toSeq ++ participantBasicNonUsers,
+        participants = participantBasicUsers.values.toSeq.map(BasicUserLikeEntity.apply) ++ participantBasicNonUsers,
```
stkem, line 72: Not 100% sure, but I don't think you need the `.apply` here (there are a few other places like this in the PR)

stkem:
:+1:

kifi-backend/modules/eliza/app/com/keepit/eliza/model/MessagingIndexCommander.scala
maxov, line 72: From what I can see it is needed, just having `BasicUserLikeEntity` makes Scala treat the object as a value and results in a compile error

maxov, line 72: Though, I can replace it with `BasicUserLikeEntity(_)` if that's more idiomatic

stkem, line 72: Nah, the `.apply` is fine. I'm just wondering why the compiler requires something here. I just tried this in the interpreter

```
case class Test(x: Int)
Seq[Int](1,2).map(Test)
res0: Seq[Test] = List(Test(1), Test(2))
```

which should be producing structurally very similar code and here the plain name works. Any idea what the difference is? (Pure intellectual curiosity at this point)

maxov, line 72: I'm honestly not sure, here's what happens when you define an object:

```
object Double {
  def apply(n: Int): Int = n * 2
}
Seq(1, 2).map(Double)
<console>:15: error: type mismatch;
 found   : Double.type
 required: Int => ?
       Seq(1, 2).map(Double)
```

And a companion object to your example class:

```
case class Test(x: Int)
object Test {}
Seq[Int](1, 2).map(Test)

<console>:15: error: type mismatch;
 found   : Test.type
 required: Int => ?
       Seq[Int](1, 2).map(Test)
                          ^
```

My suspicion is that Scala does not _really_ create a companion object when you define a lone case class, it is smart enough to know that when you refer to it by name you want to use it as a function with `.apply` or `.unapply`. However, when you do define a companion object, it is possible to pass it around as a value, so Scala cannot tell when you mean to use it as a function with `.apply` or as a value.

PR #17172, opened 2015-09-04
Re-using already open db sessions with ThreadLocal (don't merge)

andrewconner:
@gratimax @eishay @stkem @ryanpbrewster @camhashemi 

This automagically reuses existing ro/rw sessions when possible that were already opened. This is more of a bandaid than an actual fix, but should reduce some acute pain.

The approach is to wrap the underlying Slick session to make `.close()` do nothing if you're in an inner session (or for rw's, commit but do not close). The rest is bookkeeping to track open sessions. I am wrapping the underlying so that from our public API's perspective, nothing changes. Your `ROSession` and `RWSession` work as before and have their own lifecycle in respect to caching and such.

Please review. Since it's the weekend coming up, I'll wait to deploy it so I can monitor closely. This has the potential to leak connections and double close them if done incorrectly.

### DIFF
--- a/kifi-backend/modules/sqldb/app/com/keepit/common/db/slick/Database.scala
+++ b/kifi-backend/modules/sqldb/app/com/keepit/common/db/slick/Database.scala
@@ -1,8 +1,9 @@
 package com.keepit.common.db.slick
 
+import com.keepit.common.db.slick.DBSession.RWSession
 import com.keepit.common.healthcheck.AirbrakeNotifier
 import play.api.Mode.Mode
-import play.api.{ Mode, Play }
+import play.api.{ Logger, Mode, Play }
 
 import scala.slick.jdbc.JdbcBackend.{ Database => SlickDatabase }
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/sqldb/app/com/keepit/common/db/slick/Database.scala
```diff
@@ -39,11 +40,59 @@ trait SlickSessionProvider {
 
 @Singleton
 class SlickSessionProviderImpl extends SlickSessionProvider {
+  import com.keepit.common.db.slick.DBSession.ROSession
+
+  private class SessionWrapper(session: Session, onClose: => Unit) extends Session {
+    override def resultSetType = session.resultSetType
```
maxov: Why are only these overrides?

andrewconner: Not sure, copied from Slick's internal wrapper

PR #17176, opened 2015-09-05
UI to invite members with a comma separated list

cvializ:
@adamjgrant (for Tuesday : ) )

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/common/components/link.styl
+++ b/kifi-backend/modules/shoebox/angular/src/common/components/link.styl
@@ -1,0 +1,13 @@
+.kf-link
+  display: inline-block
+  padding: 0
+  border: 0
+  color: accentGreen
+  background: 0
+
+.kf-link:hover,
+.kf-link-underline
+  text-decoration: underline
+
+.kf-link:hover:not(:active)
+  color: lighten(accentGreen, 20%)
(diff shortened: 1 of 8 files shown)


cvializ:
screen shot 2015-09-04 at 5 32 11 pm
screen shot 2015-09-04 at 5 32 27 pm
screen shot 2015-09-04 at 5 32 01 pm
screen shot 2015-09-04 at 5 32 37 pm

cvializ:
I'm unsure of the angle brackets in the "<< Back ..." link. It was in the spec in Asana, but I know those aren't always 1-1 specs. @adamjgrant do you think this is a good treatment for this use case?

andrewconner:
I agree, I'd prefer even something unicode. ←, ◀, ↜, ⇜, ⇦

adamjgrant:
What Andrew said.

and `&raquo; &laquo;` are nice too.

adamjgrant:
This bit looks a little icky, but I know it's in the specs. 

screen shot 2015-09-08 at 10 06 54 am

cvializ:
Do you think the text "Invite individuals..." is even necessary? I think we could remove it. Unless you think the "large team" text would seem like a misleading label for the content below it.

andrewconner:
Agreed

adamjgrant:
@cvializ I think you're on the right track. We could simplify the text here save some space.

adamjgrant:
screen shot 2015-09-08 at 10 06 54 am

cvializ:
Looks good to me, I'll add that in

cvializ:
screen shot 2015-09-08 at 10 56 00 am
screen shot 2015-09-08 at 10 56 16 am

adamjgrant:
Great! :+1:

PR #17178, opened 2015-09-06
consolidate ansible playbooks and updated documentation

joshm1:
@stkem @andrewconner 

I worked on consolidating as much of the run-list between a localhost run vs a remote run. I also updated the docs a good bit.

I'll schedule a video-conference meeting this week to discuss ansible with anyone who's interested :)


stkem:
:+1:

andrewconner:
Thanks!

andrewconner:
@joshm1 I'm going ahead to merge, just so we're all in sync.

I actually noticed that `cortex-spot-2` and `cortex-spot-4` are not actually spots, they're demands. So I want to fix it, but don't know the best way now (I see that you said Ansible doesn't support persistent requests), so will wait for you to lead me :)

PR #17179, opened 2015-09-07
Fixed a few session-in-session issues

ryanpbrewster:
@andrewconner I'm leaving this in your hands. I may have an internet connection tonight (I will be in Paris), but if you wouldn't mind, could you take a look at this and merge it if it's reasonable.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/OrganizationAvatarController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/OrganizationAvatarController.scala
@@ -30,15 +30,10 @@ class OrganizationAvatarController @Inject() (
     val cropRegion = SquareImageCropRegion(ImageOffset(x, y), s)
     val uploadImageF = orgAvatarCommander.persistOrganizationAvatarsFromUserUpload(request.orgId, request.body.file, cropRegion)
     uploadImageF.map {
-      case Left(fail) => InternalServerError(Json.obj("error" -> fail.reason))
-      case Right(hash) =>
-        orgAvatarCommander.getBestImageByOrgId(request.orgId, OrganizationAvatarConfiguration.defaultSize) match {
-          case Some(avatar: OrganizationAvatar) =>
-            Ok(Json.obj("uploaded" -> avatar.imagePath))
-          case None =>
-            airbrake.notify(s"Uploaded an avatar for org ${request.orgId} but could not retrieve it", new Exception())
-            NotFound(Json.obj("error" -> "image_not_found"))
-        }
+      case Left(fail) => throw new Exception(fail.reason)
+      case Right(avatars) =>
+        val avatar = ProcessedImageSize.pickByIdealImageSize(OrganizationAvatarConfiguration.defaultSize, avatars, strictAspectRatio = false)(_.imageSize).get
+        Ok(Json.obj("uploaded" -> avatar.imagePath))
     }
   }
 }
--- a/kifi-backend/modules/shoebox/test/com/keepit/controllers/mobile/MobileOrganizationInviteControllerTest.scala
+++ b/kifi-backend/modules/shoebox/test/com/keepit/controllers/mobile/MobileOrganizationInviteControllerTest.scala
@@ -2,6 +2,7 @@ package com.keepit.controllers.mobile
 
 import com.google.inject.Injector
 import com.keepit.abook.FakeABookServiceClientModule
+import com.keepit.common.concurrent.WatchableExecutionContext
 import com.keepit.common.controller.FakeUserActionsHelper
 import com.keepit.common.crypto.{ PublicId, PublicIdConfiguration }
 import com.keepit.common.db.{ Id }
@@ -50,6 +51,8 @@ class MobileOrganizationInviteControllerTest extends Specification with ShoeboxT
           val request = route.acceptInvitation(publicOrgId, invite.authToken)
           val result = controller.acceptInvitation(publicOrgId, invite.authToken)(request)
 
+          Thread.sleep(1000) // TODO(ryan): Ask Andrew is there's any way to avoid doing this. WatchableExecutionContext.drain() does not seem to work
+
           status(result) must equalTo(NO_CONTENT)
         }
       }
@@ -72,6 +75,9 @@ class MobileOrganizationInviteControllerTest extends Specification with ShoeboxT
           inject[FakeUserActionsHelper].setUser(invitee)
           val request = route.acceptInvitation(publicOrgId, invite.authToken)
           val result = controller.acceptInvitation(publicOrgId, invite.authToken)(request)
+
+          Thread.sleep(1000) // TODO(ryan): Ask Andrew is there's any way to avoid doing this. WatchableExecutionContext.drain() does not seem to work
+
           status(result) must equalTo(NO_CONTENT)
         }
       }
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepDecorator.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/KeepDecorator.scala
@@ -69,14 +69,9 @@ class KeepDecoratorImpl @Inject() (
         }
 
         val libIdToBasicOrg = {
-          val libId2orgId = idToLibrary.mapValues(lib => lib.organizationId).collect { case (id, Some(orgid)) => id -> orgid }
-          val libId2OrgCard = db.readOnlyMaster { implicit s =>
-            libId2orgId.map {
-              case (libId, orgId) =>
-                libId -> organizationCommander.getBasicOrganization(orgId)
-            }
-          }
-          libId2OrgCard
+          val libIdToOrgId = idToLibrary.mapValues(lib => lib.organizationId).collect { case (id, Some(orgId)) => id -> orgId }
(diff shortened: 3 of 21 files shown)


ryanpbrewster:
I'm like 90% sure everything compiles and passes tests.

ryanpbrewster:
@eishay this PR conflicts with your default org avatar PR (https://github.com/kifi/keepit/pull/17180). I'll wait for you to merge yours, then fix this one.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/OrganizationAvatarController.scala
```diff
@@ -30,15 +30,10 @@ class OrganizationAvatarController @Inject() (
...
-      case Left(fail) => InternalServerError(Json.obj("error" -> fail.reason))
-      case Right(hash) =>
-        orgAvatarCommander.getBestImageByOrgId(request.orgId, OrganizationAvatarConfiguration.defaultSize) match {
-          case Some(avatar: OrganizationAvatar) =>
-            Ok(Json.obj("uploaded" -> avatar.imagePath))
-          case None =>
-            airbrake.notify(s"Uploaded an avatar for org ${request.orgId} but could not retrieve it", new Exception())
-            NotFound(Json.obj("error" -> "image_not_found"))
-        }
+      case Left(fail) => throw new Exception(fail.reason)
```
andrewconner, line 33: This seems a bit weird to explicitly contain image errors in the `Either`, and then throw it on the controller level (this will return a response to client that gives no indication of what went wrong—sometimes that's exactly what you want)

andrewconner:
Thanks. We may want to slow down a bit and make sure we're in agreement about the "where do we open sessions" design pattern. I see you made the avatar commander take in a passed session. @stkem do you have any objections?

kifi-backend/modules/shoebox/test/com/keepit/controllers/mobile/MobileOrganizationInviteControllerTest.scala
```diff
@@ -50,6 +51,8 @@ class MobileOrganizationInviteControllerTest extends Specification with ShoeboxT
           val request = route.acceptInvitation(publicOrgId, invite.authToken)
           val result = controller.acceptInvitation(publicOrgId, invite.authToken)(request)
 
+          Thread.sleep(1000) // TODO(ryan): Ask Andrew is there's any way to avoid doing this. WatchableExecutionContext.drain() does not seem to work
```
andrewconner, line 54: What's the issue here?

andrewconner:
Things to talk about:
- What's the rule of where to put the session opening? Previously, @stkem promoted that we open sessions in commanders. With the pattern of commanders-calling-commanders, this has shown to be a growing issue. I'm fine with something fuzzy, as long as there's agreement: "If the typical use case of someone calling this has a session, then accept a implicit passed-in session; if it seems extraordinarily rare to be in the middle of something else while needing this method, don't accept a session". Thoughts?
- Opening sessions in the controller is also problematic:
  - Since the same session is passed all the way down, the session itself can't easily be interacted with—i.e., closing it because we want it committed.
  - Sessions end up being open for a long time—this was actually the motivation for moving to sessions in commanders.
  - Any time a service-to-service call happens, a new session will have to be opened anyways.
- Ryan has started creating two methods for each functionality: `def x` and `def xHelper`, where `Helper` has a implicit session passed in. Some issues I see: a) duplication, where the compiler can't help you if you pick the non-helper one but indeed did have a session, b) lack of preference, so callers call either, making finding usages harder. However, it does let us access the "inner" part of a session when needed. We should be consistent if we do this, though—right now, it's not.

@stkem @ryanpbrewster

andrewconner:
Speaking of, without rewriting everything to use the `DBIO` monad (which is _really_ slick), I'd love an API like this:

```
def a(b: B)(implicit session: ROSession = createROSession)
```

Not sure if that's possible in Scala though.

stkem:
A few thoughts on this:
- I still think that db sessions should only be created by commanders wherever possible, with controllers only dealing with (de) serialization and authentication. That would mean that for some call that requires multiple commander involvement, there would be some commander "in charge", so to speak.
- When a session does need to be passed on to a helper method, I would strongly advocate for making it explicit (i.e. reserving implicit session parameters for repo methods). An implicit parameter that is being passed around through multiple steps quickly makes it really hard to understand what is going on and not knowing where your session is coming form often inadvertently leads to really long ones.
- I have no fundamental objection to helper methods that take in a session (with the above caveats), but I think overall the better pattern is to use multiple sequential session as opposed to passing one around. It's less efficient but makes for cleaner code, imho. Specifically, unless you need real transactionality (which, admittedly is fairly common and which case I think the helper pattern is fine), a commander should first get all of the data is needs from the DB (and possibly write things), close the session, then call other commanders, and stitch the data together.

andrewconner:
> there would be some commander "in charge", so to speak

The issue is that that's a matter of perspective. Thinking about payments, you'd think it should be self-contained and should manage its own resources (so that it can ensure things get committed right, we use the right _type_ of session, etc). However, turns out it's very common to actually already have a session.

> I would strongly advocate for making it explicit

Not sure I feel so strongly. It _is_ explicit, the session doesn't automatically get created or anything, and the method's signature makes it clear that a session is needed.

> a commander should first get all of the data is needs from the DB (and possibly write things), close the session, then call other commanders, and stitch the data together

I think this is what does happen now. Based on looking at quite a few of these The most common issue is usually three hops: `CommanderA` calls `SomethingElse` which calls `CommanderB`. From `CommanderA`'s perspective, everything's great.

ryanpbrewster:
How do y'all feel about introducing another layer between Commander and
Repo that explicitly cannot open sessions but had business logic in it.

Commander methods are then defined in terms of transactionally grouped
operations, and all the actual business logic in defined in this lieutenant
class.

I like the separation of concerns that this brings. Every class has its own
concerns. Controllers handle user requests and ideally only do
deserialization, then defer to commanders. Commanders batch together things
that need to happen inside a single DB session. Lieutenants handle the
business logic of glueing together DB interactions. Repos are just places
to store DB queries.

I'm not sure that actually solves the problem completely, but it would make
it much harder to create session-in-session issues, and it would encourage
us to group computations better.

andrewconner:
I'm not sure this works, because dependencies don't often follow a strict hierarchy (and not everything even needs a database—most LOC don't). Besides adding yet another layer of abstraction,

> Commanders batch together things that need to happen inside a single DB session.

Who groups _these_ together? Other commanders? Then it's not a single session. Controllers? Then controllers are big. It's common to want several sessions, especially when saving things.

> Lieutenants handle the business logic of glueing together DB interactions

It's not always clear where the API you're writing is needed; as before, from the person writing a Payments method, it's pretty clear that this should be stand-alone and be transactional. However, it's sometimes needed inside a larger process.

I suppose to re-iterate a point, one of the most common offenders happens in three classes, where there's a step of indirection.

ryanpbrewster:
It's 4:30am here, so if this doesn't make sense feel free to ignore it.

I'm not sure I agree. In my ideal world, the control flow looks like this:
    1. A user makes a request
    2. The appropriate controller accepts the request (in terms of external values), opens a short db session, translates the external request into internal values, then closes the request.
    3. The controller calls a very small sequence of commander methods (for almost all cases I'd expect this to be a single method). These commander methods are transactional. Every one opens a single db session, does some stuff, then commits it to the db. I can't think of many examples where a user request needs to open more than 1 or 2 individual db sessions.
    4. Each commander method opens the appropriate type of session and starts calling Lieutenant (holy cow I wish I had chosen a better naming convention. The military theme did not work out in my favor) methods that read/write from the db and do stuff.
    5. The Lieutenant methods may make relatively complex queries (i.e., repo methods), or do sort of involved write methods. However, they are (ostensibly) grouped by transactional atomicity, so they at least make sense together.
    6. The commander closes sessions as they complete, then aggregates the data into the results that the controller expects.
    7. The controller receives the commander's result, opens a db session to convert it into external variables (calling only very simple repo methods), then returns it to the user.

There are definitely issues with this.

It may be that we want to group blocks of similar code together in the commander. That's fine. As long as commander methods are always called outside of sessions, they are happy. Similarly, Lieutenant methods must always be called _inside_ of sessions. We can establish a very strong convention here. No commander methods ever get called from inside sessions. No Lieutenant methods ever get called from outside sessions. You could have something like

```
def commanderMethodA() = {
    val (a,b) = db.readOnlyMaster { implicit session =>
        val a = lieutenantA.doA()
        val b = lieutenantA.doB(a)
        (a,b)
    }
    val c = lieutenant.nonDbLogic(a, b)
    val d = commanderMethodB(a, b, c)
    db.readWrite { implicit session => lieutenant.doD(d) }
}
```

and it would be immediately obvious if anything was wrong. Commander methods (even private ones) are assumed to open a db session. Easy to prevent mix-ups. Similarly, within the Lieutenant (good examples of things that would immediately be classified as Lieutenants are `HandleCommander` and `KeepToLibraryCommander`) you are not even allowed to declare a `db: Database`, so we reduce the chances of session-in-session mixups.

# 

Basically, Commanders are in charge of monitoring DB usage. There are methods for transactional logic, and methods that combine independent transactions together for convenience (so controllers can stay small), but they _never_ pass implicit sessions around or call other commander methods within a session.

Lieutenants are in charge of complex logic, and they _never_ call commander methods.

Repos are still just places to combine related queries.

Controllers continue to be lightweight interfaces between user requests and commander responses (doing internal/external conversion as necessary to preserve data sanitation and prevent internal data leaks).

PR #17180, opened 2015-09-08
adding default avarars to orgs

eishay:
DONT MERGE UNTIL ALL ORGS ARE GRANDFATHERED WITH AN AVATAR

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/model/BaseImage.scala
+++ b/kifi-backend/modules/common/app/com/keepit/model/BaseImage.scala
@@ -69,9 +69,7 @@ sealed trait ImageStoreInProgress extends ImageProcessState
 sealed trait ImageProcessDone extends ImageProcessState
 
 sealed trait ImageProcessSuccess extends ImageProcessDone
-sealed abstract class ImageStoreFailure(val reason: String, val cause: Option[Throwable] = None) extends ImageProcessState with ImageProcessDone {
-  def asException: Exception = cause.map(ex => new Exception(reason, ex)).getOrElse(new Exception(reason))
-}
+sealed abstract class ImageStoreFailure(val reason: String, val cause: Option[Throwable] = None) extends ImageProcessState with ImageProcessDone
 
 object ImageProcessState {
   // In-progress
--- a/kifi-backend/modules/common/app/com/keepit/model/BaseImage.scala
+++ b/kifi-backend/modules/common/app/com/keepit/model/BaseImage.scala
@@ -69,7 +69,9 @@ sealed trait ImageStoreInProgress extends ImageProcessState
 sealed trait ImageProcessDone extends ImageProcessState
 
 sealed trait ImageProcessSuccess extends ImageProcessDone
-sealed abstract class ImageStoreFailure(val reason: String, val cause: Option[Throwable] = None) extends ImageProcessState with ImageProcessDone
+sealed abstract class ImageStoreFailure(val reason: String, val cause: Option[Throwable] = None) extends ImageProcessState with ImageProcessDone {
+  def asException: Exception = cause.map(ex => new Exception(reason, ex)).getOrElse(new Exception(reason))
+}
 
 object ImageProcessState {
   // In-progress
--- a/kifi-backend/modules/common/app/com/keepit/model/Organization.scala
+++ b/kifi-backend/modules/common/app/com/keepit/model/Organization.scala
@@ -197,7 +197,7 @@ case class BasicOrganization(
     handle: OrganizationHandle,
     name: String,
     description: Option[String],
-    avatarPath: Option[ImagePath]) {
+    avatarPath: ImagePath) {
 
   def abbreviatedName = this.name.abbreviate(33)
 
@@ -210,7 +210,7 @@ object BasicOrganization {
     (__ \ 'handle).format[OrganizationHandle] and
     (__ \ 'name).format[String] and
     (__ \ 'description).formatNullable[String] and
-    (__ \ 'avatarPath).formatNullable[ImagePath]
+    (__ \ 'avatarPath).format[ImagePath]
   )(BasicOrganization.apply, unlift(BasicOrganization.unapply))
 }
 
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationAvatarCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationAvatarCommander.scala
@@ -19,9 +19,10 @@ import scala.util.{ Failure, Success }
 
 @ImplementedBy(classOf[OrganizationAvatarCommanderImpl])
 trait OrganizationAvatarCommander {
-  def getBestImageByOrgId(orgId: Id[Organization], imageSize: ImageSize): Option[OrganizationAvatar]
-  def getBestImagesByOrgIds(orgIds: Set[Id[Organization]], imageSize: ImageSize): Map[Id[Organization], Option[OrganizationAvatar]]
-  def persistOrganizationAvatarsFromUserUpload(orgId: Id[Organization], imageFile: File, cropRegion: SquareImageCropRegion): Future[Either[ImageStoreFailure, ImageHash]]
+  def getBestImageByOrgId(orgId: Id[Organization], imageSize: ImageSize): OrganizationAvatar
+  def getBestImagesByOrgIds(orgIds: Set[Id[Organization]], imageSize: ImageSize): Map[Id[Organization], OrganizationAvatar]
+  def persistOrganizationAvatarsFromUserUpload(orgId: Id[Organization], imageFile: File, cropRegion: SquareImageCropRegion): Future[ImageHash]
+  def saveNewAvatars(orgId: Id[Organization], imageHash: ImageHash, uploadedImages: Set[ImageProcessState.UploadedImage]): Unit
 }
 
 @Singleton
@@ -33,45 +34,42 @@ class OrganizationAvatarCommanderImpl @Inject() (
     val webService: WebService,
     private implicit val executionContext: ExecutionContext) extends OrganizationAvatarCommander with ProcessedImageHelper with Logging {
 
-  def getBestImageByOrgId(orgId: Id[Organization], imageSize: ImageSize): Option[OrganizationAvatar] = {
+  def getBestImageByOrgId(orgId: Id[Organization], imageSize: ImageSize): OrganizationAvatar = {
     getBestImagesByOrgIds(Set(orgId), imageSize).head._2
   }
-  def getBestImagesByOrgIds(orgIds: Set[Id[Organization]], imageSize: ImageSize): Map[Id[Organization], Option[OrganizationAvatar]] = {
+
+  def getBestImagesByOrgIds(orgIds: Set[Id[Organization]], imageSize: ImageSize): Map[Id[Organization], OrganizationAvatar] = {
     val candidatesById = db.readOnlyReplica { implicit session => orgAvatarRepo.getByOrgIds(orgIds) }
     orgIds.map { orgId =>
-      orgId -> ProcessedImageSize.pickByIdealImageSize(imageSize, candidatesById(orgId), strictAspectRatio = false)(_.imageSize)
+      orgId -> ProcessedImageSize.pickByIdealImageSize(imageSize, candidatesById(orgId), strictAspectRatio = false)(_.imageSize).get
     }.toMap
   }
 
-  def persistOrganizationAvatarsFromUserUpload(orgId: Id[Organization], imageFile: File, cropRegion: SquareImageCropRegion): Future[Either[ImageStoreFailure, ImageHash]] = {
+  def persistOrganizationAvatarsFromUserUpload(orgId: Id[Organization], imageFile: File, cropRegion: SquareImageCropRegion): Future[ImageHash] = {
     fetchAndHashLocalImage(imageFile).flatMap {
-      case Left(storeError) => Future.successful(Left(storeError))
+      case Left(storeError) => throw storeError.asException
       case Right(sourceImage) =>
         validateAndGetImageInfo(sourceImage.file) match {
-          case Failure(validationError) => Future.successful(Left(ImageProcessState.InvalidImage(validationError)))
+          case Failure(validationError) => throw validationError
           case Success(imageInfo) =>
             val uploadedImagesFut = persistOrganizationAvatarsFromSourceImage(orgId, sourceImage, imageInfo, cropRegion)
             uploadedImagesFut.imap {
-              case Left(uploadError) => Left(uploadError)
+              case Left(uploadError) => throw uploadError.asException
               case Right(uploadedImages) =>
-                try {
-                  saveNewAvatars(orgId, sourceImage, uploadedImages)
-                  Right(sourceImage.hash)
-                } catch {
-                  case repoError: Exception =>
-                    log.error(s"Failed to update OrganizationAvatarRepo after processing image from user uploaded file: $repoError")
-                    Left(ImageProcessState.DbPersistFailed(repoError))
-                }
+                val hash = sourceImage.hash
+                saveNewAvatars(orgId, hash, uploadedImages)
+                hash
             }
         }
     }
   }
 
-  def persistOrganizationAvatarsFromSourceImage(orgId: Id[Organization], sourceImage: ImageProcessState.ImageLoadedAndHashed, imageInfo: RawImageInfo, cropRegion: SquareImageCropRegion): Future[Either[ImageStoreFailure, Set[ImageProcessState.UploadedImage]]] = {
+  private def persistOrganizationAvatarsFromSourceImage(orgId: Id[Organization], sourceImage: ImageProcessState.ImageLoadedAndHashed, imageInfo: RawImageInfo, cropRegion: SquareImageCropRegion): Future[Either[ImageStoreFailure, Set[ImageProcessState.UploadedImage]]] = {
     val necessary: Set[ProcessImageRequest] = OrganizationAvatarConfiguration.sizes.map { finalSize => CropScaleImageRequest(offset = cropRegion.offset, cropSize = cropRegion.size, finalSize = finalSize.idealSize) }
     val processedImages = processAndPersistImages(sourceImage.file, imagePathPrefix, sourceImage.hash, sourceImage.format, necessary)(photoshop)
     processedImages match {
-      case Left(processingError) => Future.successful(Left(processingError))
+      case Left(processingError) =>
+        Future.successful(Left(processingError))
       case Right(processedImagesReadyToPersist) =>
         val uploads = processedImagesReadyToPersist.map { image =>
           imageStore.put(image.key, image.file, imageFormatToMimeType(image.format)).imap { _ =>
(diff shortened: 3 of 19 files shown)


eishay:
@andrewconner

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationAvatarCommander.scala
```diff
@@ -33,45 +34,44 @@ class OrganizationAvatarCommanderImpl @Inject() (
...
 
-  def persistOrganizationAvatarsFromUserUpload(orgId: Id[Organization], imageFile: File, cropRegion: SquareImageCropRegion): Future[Either[ImageStoreFailure, ImageHash]] = {
+  def persistOrganizationAvatarsFromUserUpload(orgId: Id[Organization], imageFile: File, cropRegion: SquareImageCropRegion): Future[ImageHash] = {
     fetchAndHashLocalImage(imageFile).flatMap {
-      case Left(storeError) => Future.successful(Left(storeError))
+      case Left(storeError) => throw storeError.asException
       case Right(sourceImage) =>
         validateAndGetImageInfo(sourceImage.file) match {
-          case Failure(validationError) => Future.successful(Left(ImageProcessState.InvalidImage(validationError)))
+          case Failure(validationError) => throw validationError
```
ryanpbrewster: We've talked about this in the past, and I know you like failing loudly. I think that is a good strategy in most cases, but here I think we should consider giving the client a useful error message (`"unsupported_image_type"` maybe?). I think users are easily frustrated by mysterious server errors, but if they get helpful messages they are much more forgiving.

I do agree that we should always airbrake. I want to know when users manage to break something.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationAvatarCommander.scala
```diff
@@ -33,45 +34,44 @@ class OrganizationAvatarCommanderImpl @Inject() (
...
+              case Left(uploadError) => throw uploadError.asException
               case Right(uploadedImages) =>
-                try {
-                  saveNewAvatars(orgId, sourceImage, uploadedImages)
-                  Right(sourceImage.hash)
-                } catch {
-                  case repoError: Exception =>
-                    log.error(s"Failed to update OrganizationAvatarRepo after processing image from user uploaded file: $repoError")
-                    Left(ImageProcessState.DbPersistFailed(repoError))
-                }
```
ryanpbrewster, line 64: :+1: for user-facing code we should not catch these errors.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationAvatarCommander.scala
```diff
@@ -33,45 +34,44 @@ class OrganizationAvatarCommanderImpl @Inject() (
...
+                val hash = sourceImage.hash
+                saveNewAvatars(orgId, hash, uploadedImages)
+                hash
             }
         }
     }
   }
 
-  def persistOrganizationAvatarsFromSourceImage(orgId: Id[Organization], sourceImage: ImageProcessState.ImageLoadedAndHashed, imageInfo: RawImageInfo, cropRegion: SquareImageCropRegion): Future[Either[ImageStoreFailure, Set[ImageProcessState.UploadedImage]]] = {
+  private def persistOrganizationAvatarsFromSourceImage(orgId: Id[Organization], sourceImage: ImageProcessState.ImageLoadedAndHashed, imageInfo: RawImageInfo, cropRegion: SquareImageCropRegion): Future[Either[ImageStoreFailure, Set[ImageProcessState.UploadedImage]]] = {
```
ryanpbrewster, line 95: :+1: good catch

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
```diff
@@ -214,21 +215,26 @@ class OrganizationCommanderImpl @Inject() (
 
   def createOrganization(request: OrganizationCreateRequest)(implicit eventContext: HeimdalContext): Either[OrganizationFail, OrganizationCreateResponse] = {
     Try {
-      db.readWrite { implicit session =>
-        getValidationError(request) match {
-          case Some(fail) =>
-            Left(fail)
-          case None =>
+      db.readWrite { implicit session => getValidationError(request) } match {
```
ryanpbrewster: `readOnlyReplica` instead of `readWrite`.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
```diff
@@ -214,21 +215,26 @@ class OrganizationCommanderImpl @Inject() (
...
+            val org = handleCommander.autoSetOrganizationHandle(orgRepo.save(orgTemplate)) getOrElse (throw OrganizationFail.HANDLE_UNAVAILABLE)
             orgMembershipRepo.save(org.newMembership(userId = request.requesterId, role = OrganizationRole.ADMIN))
             planManagementCommander.createAndInitializePaidAccountForOrganization(org.id.get, PaidPlan.DEFAULT, request.requesterId, session) //this should get a .get when thing sare solidified
             organizationAnalytics.trackOrganizationEvent(org, userRepo.get(request.requesterId), request)
-            Right(OrganizationCreateResponse(request, org))
-        }
+            org
+          }
+          //hard coded detault image at https://djty7jcqog9qu.cloudfront.net/oa/076fccc32247ae67bb75d48879230953_1024x1024-0x0-200x200_cs.jpg
+          val imageHash = ImageHash("076fccc32247ae67bb75d48879230953")
```
ryanpbrewster: Can we store these all as `val`s within the `Organization` companion object instead?

kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtLibraryController.scala
```diff
@@ -69,6 +69,9 @@ class ExtLibraryController @Inject() (
       case (lib, membership, collaboratorsIds) =>
         val owner = basicUserById.getOrElse(lib.ownerId, throw new Exception(s"owner of $lib does not have a membership model"))
         val collabs = (collaboratorsIds - request.userId).map(basicUserById(_)).toSeq
+        val orgAvatar = lib.organizationId.map { orgId =>
+          orgAvatarsById.get(orgId).getOrElse(throw new Exception(s"No avatar for org of lib $lib"))
```
ryanpbrewster: Any reason we don't just use `orgAvatarsById(orgId)` and let it throw a `KeyNotFound` exception?

kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtLibraryController.scala
```diff
@@ -117,7 +120,9 @@ class ExtLibraryController @Inject() (
           case Left(fail) => Status(fail.status)(Json.obj("error" -> fail.message))
           case Right(lib) =>
             val data = db.readOnlyMaster { implicit session =>
-              val orgAvatar = lib.organizationId.flatMap(organizationAvatarCommander.getBestImageByOrgId(_, ExtLibraryController.defaultImageSize).map(_.imagePath))
+              val orgAvatar = lib.organizationId.map { orgId =>
+                organizationAvatarCommander.getBestImageByOrgId(orgId, ExtLibraryController.defaultImageSize).imagePath
+              }
```
ryanpbrewster, line 125: Here you use the val name `orgAvatar` to represent the avatar's image path. Above you use it for the avatar itself. Can we change this to `orgAvatarPath`, or change the code on line 73 to include the `.map(_.imagePath)`?

kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtUserController.scala
```diff
@@ -45,9 +45,9 @@ class ExtUserController @Inject() (
         Json.obj(
           "name" -> (org.name + " Members"),
           "id" -> org.orgId,
-          "pictureName" -> ("../../../../" + org.avatarPath.map(_.path).getOrElse("NONE"): String), // one weird trick
+          "pictureName" -> ("../../../../" + org.avatarPath.path: String), // one weird trick
```
ryanpbrewster, line 48: `// one weird trick` still applicable?

kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtUserController.scala
```diff
@@ -45,9 +45,9 @@ class ExtUserController @Inject() (
...
           "kind" -> "org",
-          "avatarPath" -> (org.avatarPath.map(_.path).getOrElse("NONE"): String),
+          "avatarPath" -> (org.avatarPath.path: String),
```
ryanpbrewster, line 50: Is this type declaration necessary?

kifi-backend/modules/shoebox/test/com/keepit/model/OrganizationFactoryHelper.scala
```diff
@@ -37,6 +40,10 @@ object OrganizationFactoryHelper {
       for (invitedEmail <- partialOrganization.invitedEmails) {
         orgInvRepo.save(OrganizationInvite(organizationId = org.id.get, inviterId = org.ownerId, emailAddress = Some(invitedEmail), role = OrganizationRole.MEMBER))
       }
+      val imageHash = ImageHash("076fccc32247ae67bb75d48879230953")
+      val orgAvatarRepo = injector.getInstance(classOf[OrganizationAvatarRepoImpl])
+      orgAvatarRepo.save(OrganizationAvatar(organizationId = org.id.get, width = 100, height = 100, format = ImageFormat.JPG, kind = ProcessImageOperation.CropScale, imagePath = ImagePath("oa/076fccc32247ae67bb75d48879230953_1024x1024-0x0-100x100_cs.jpg"), source = UserUpload, sourceFileHash = imageHash, sourceImageURL = None))
+      orgAvatarRepo.save(OrganizationAvatar(organizationId = org.id.get, width = 200, height = 200, format = ImageFormat.JPG, kind = ProcessImageOperation.CropScale, imagePath = ImagePath("oa/076fccc32247ae67bb75d48879230953_1024x1024-0x0-200x200_cs.jpg"), source = UserUpload, sourceFileHash = imageHash, sourceImageURL = None))
```
ryanpbrewster: One benefit of having these default avatars stored in the `Organization` companion object is that these become a lot less mysterious.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationAvatarCommander.scala
ryanpbrewster: Let's handle that later. I think for this PR it's fine.

ryanpbrewster:
I like this change a lot. It makes sense from a product perspective, and it ends up making the code a little cleaner. I had a few minor style suggestions, and this lgtm to me overall.

kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtUserController.scala
andrewconner, line 48: Yes. The `../../../../` is the weird bit.

kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtUserController.scala
andrewconner, line 50: It was, not any more

eishay, line 50: no idea, first time i saw it, didn't touch

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationAvatarCommander.scala
andrewconner: Agreed, this one _is_ a user error. Clients should be able to handle it.

eishay: talked with @andrewconner, i'll change this code back to the way it was

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
eishay: thanks

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationAvatarCommander.scala
andrewconner, line 64: Curious where this try/catch came from. Did it fail? In this case, it's perfectly okay to exception out. You're in a Future, so not destroying a stack, and it's a genuine unexpected system error

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationAvatarCommander.scala
```diff
@@ -33,45 +34,44 @@ class OrganizationAvatarCommanderImpl @Inject() (
...
+      case Left(storeError) => throw storeError.asException
       case Right(sourceImage) =>
         validateAndGetImageInfo(sourceImage.file) match {
-          case Failure(validationError) => Future.successful(Left(ImageProcessState.InvalidImage(validationError)))
+          case Failure(validationError) => throw validationError
           case Success(imageInfo) =>
             val uploadedImagesFut = persistOrganizationAvatarsFromSourceImage(orgId, sourceImage, imageInfo, cropRegion)
             uploadedImagesFut.imap {
-              case Left(uploadError) => Left(uploadError)
+              case Left(uploadError) => throw uploadError.asException
```
andrewconner: Why not just change the signature of `persistOrganizationAvatarsFromSourceImage`?

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
```diff
@@ -214,21 +215,26 @@ class OrganizationCommanderImpl @Inject() (
...
+          val uploadedImages = Set(
+            ImageProcessState.UploadedImage(key = ImagePath("oa/076fccc32247ae67bb75d48879230953_1024x1024-0x0-100x100_cs.jpg"), format = ImageFormat.JPG, imageInfo = RawImageInfo(ImageFormat.JPG.value, 100, 100), processOperation = ProcessImageOperation.CropScale),
+            ImageProcessState.UploadedImage(key = ImagePath("oa/076fccc32247ae67bb75d48879230953_1024x1024-0x0-200x200_cs.jpg"), format = ImageFormat.JPG, imageInfo = RawImageInfo(ImageFormat.JPG.value, 200, 200), processOperation = ProcessImageOperation.CropScale))
+          organizationAvatarCommander.saveNewAvatars(org.id.get, imageHash, uploadedImages)
+          Right(OrganizationCreateResponse(request, org))
```
andrewconner, line 231: Re: our conversation: May be nice not to persist anything.

kifi-backend/modules/common/app/com/keepit/model/Organization.scala
```diff
@@ -211,7 +211,7 @@ object BasicOrganization {
     (__ \ 'handle).format[OrganizationHandle] and
     (__ \ 'name).format[String] and
     (__ \ 'description).formatNullable[String] and
-    (__ \ 'avatarPath).formatNullable[ImagePath]
+    (__ \ 'avatarPath).format[ImagePath]
```
andrewconner, line 214: Oo, I wonder why this is a `Format`, vs just a `Writes`

andrewconner:
To summarize Eishay and my real life discussion:
- My initial thought was that org default avatars wouldn't be persisted, but instead would just be returned to clients if there was no avatar (makes it easier to change in the future, because things are so heavily cached).
- Some image processing errors should be returned to clients/callers so they can handle them. (throwing an exception guarantees that they have no idea what happened). For commanders, it may be a client receiving the output, or a system process.
- I'm sad that the once-generic ProcessedImageHelper has been duplicated and the image processing APIs are diverging. :frowning:

kifi-backend/modules/common/app/com/keepit/model/Organization.scala
eishay, line 214: think we should change it all to writes?

kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtLibraryController.scala
eishay: I rather have an explicit exception, know which library had a missing org avatar. It would make debugging much faster in case problems happen

kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtLibraryController.scala
eishay, line 125: sure!

eishay:
@andrewconner @ryanpbrewster updated

kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtUserController.scala
andrewconner, line 50: It's just that `Json.obj` uses implicit conversions to achieve its goal, and Scala doesn't infer the right thing with options when getOrElse is used. It can be removed (and was needed previously)

kifi-backend/modules/common/app/com/keepit/model/Organization.scala
andrewconner, line 214: Yeah, it should never be read. But not sure if it's used in a cache or something; no need to do anything now. Just a warning signal, since we've had issues of someone using a class like this for client APIs.

kifi-backend/modules/common/app/com/keepit/model/BaseImage.scala
```diff
@@ -69,7 +69,9 @@ sealed trait ImageStoreInProgress extends ImageProcessState
 sealed trait ImageProcessDone extends ImageProcessState
 
 sealed trait ImageProcessSuccess extends ImageProcessDone
-sealed abstract class ImageStoreFailure(val reason: String, val cause: Option[Throwable] = None) extends ImageProcessState with ImageProcessDone
+sealed abstract class ImageStoreFailure(val reason: String, val cause: Option[Throwable] = None) extends ImageProcessState with ImageProcessDone {
+  def asException: Exception = cause.map(ex => new Exception(reason, ex)).getOrElse(new Exception(reason))
```
andrewconner: Not used?

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationAvatarCommander.scala
```diff
@@ -19,9 +19,10 @@ import scala.util.{ Failure, Success }
 
 @ImplementedBy(classOf[OrganizationAvatarCommanderImpl])
 trait OrganizationAvatarCommander {
-  def getBestImageByOrgId(orgId: Id[Organization], imageSize: ImageSize): Option[OrganizationAvatar]
-  def getBestImagesByOrgIds(orgIds: Set[Id[Organization]], imageSize: ImageSize): Map[Id[Organization], Option[OrganizationAvatar]]
+  def getBestImageByOrgId(orgId: Id[Organization], imageSize: ImageSize): OrganizationAvatar
+  def getBestImagesByOrgIds(orgIds: Set[Id[Organization]], imageSize: ImageSize): Map[Id[Organization], OrganizationAvatar]
   def persistOrganizationAvatarsFromUserUpload(orgId: Id[Organization], imageFile: File, cropRegion: SquareImageCropRegion): Future[Either[ImageStoreFailure, ImageHash]]
+  def saveNewAvatars(orgId: Id[Organization], imageHash: ImageHash, uploadedImages: Set[ImageProcessState.UploadedImage]): Unit
```
andrewconner: I believe this wasn't in the trait before. Does it need to be?

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationAvatarCommander.scala
```diff
@@ -33,16 +34,41 @@ class OrganizationAvatarCommanderImpl @Inject() (
...
     val candidatesById = db.readOnlyReplica { implicit session => orgAvatarRepo.getByOrgIds(orgIds) }
     orgIds.map { orgId =>
-      orgId -> ProcessedImageSize.pickByIdealImageSize(imageSize, candidatesById(orgId), strictAspectRatio = false)(_.imageSize)
+      val avatarOpt = ProcessedImageSize.pickByIdealImageSize(imageSize, candidatesById.get(orgId).getOrElse(defaultOrgImages(orgId)), strictAspectRatio = false)(_.imageSize)
+      val avatar = avatarOpt.getOrElse(throw new Exception(s"no avatar for org $orgId with image size $imageSize"))
+      orgId -> avatar
     }.toMap
   }
 
+  private def defaultOrgImages(orgId: Id[Organization]): Seq[OrganizationAvatar] = {
```
andrewconner, line 49: :+1:

andrewconner:
Minor comments, thanks!

eishay:
thanks @andrewconner, cleaned it up

PR #17182, opened 2015-09-08
Added a site endpoint for keepable libraries

ryanpbrewster:
@eishay I think this might conflict with your org avatar PR as well. Should be a one-line change (line 284 in `LibraryController`). Sorry.


ryanpbrewster:
@cvializ @adamjgrant I ended up making a separate endpoint for this instead of modifying `/site/libraries`, because the `"membership"` field there was not optional and I wasn't sure if that would break anything.

This new endpoint (`/site/keepableLibraries`) works exactly the same way as `/ext/libraries`. It takes an optional URL parameter (`"includeOrgLibraries"`) which defaults to `false`. If the user is a member of that library already, it will have a `membership` field. If not, it won't (and you'll need to make the user join the library before trying to add a keep).

PR #17190, opened 2015-09-08
Fixing suggested searches query URL

andrewconner:
@cvializ @adamjgrant

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/common/filter.js
+++ b/kifi-backend/modules/shoebox/angular/src/common/filter.js
@@ -40,6 +40,9 @@ angular.module('kifi')
 
 .filter('tagUrl', function () {
   return function (tag) {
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/angular/src/common/filter.js
```diff
@@ -40,6 +40,9 @@ angular.module('kifi')
 
 .filter('tagUrl', function () {
   return function (tag) {
+    if (_.startsWith(tag, '#')) {
```
andrewconner, line 43: This is because tags do not begin with #. "Suggested tags" seems to include it, because it was a twitter hashtag.

cvializ:
:+1: :+1:

adamjgrant:
:+1:

PR #17191, opened 2015-09-08
Serving notifications

maxov:
@stkem 
@andrewconner 
@leogrim 
@ryanpbrewster 

Still a WIP

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/notify/info/StandardNotificationInfo.scala
+++ b/kifi-backend/modules/common/app/com/keepit/notify/info/StandardNotificationInfo.scala
@@ -1,7 +1,8 @@
 package com.keepit.notify.info
 
-import com.keepit.common.db.Id
+import com.keepit.common.db.{ ExternalId, Id }
 import com.keepit.model.{ User, NotificationCategory }
+import com.keepit.notify.model.event.NewMessage
 import com.keepit.social.BasicUser
 import play.api.libs.json.JsObject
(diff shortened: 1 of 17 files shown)


kifi-backend/modules/common/app/com/keepit/notify/info/StandardNotificationInfo.scala
```diff
@@ -1,7 +1,8 @@
 package com.keepit.notify.info
 
-import com.keepit.common.db.Id
+import com.keepit.common.db.{ ExternalId, Id }
```
stkem: The added import of `ExternalId` seems unused in the diff. Am I missing something?

stkem:
code looks good to me on first read through. Do we have test coverage on the modified websocket routes here now?

maxov:
@stkem No, it's possible to add some tests but it would be quite difficult because of all of the moving parts.. I'll definitely try to do that when the logic is more finished. For now I'm just testing it locally, running with a dev server.

kifi-backend/modules/common/app/com/keepit/notify/info/StandardNotificationInfo.scala
maxov: thanks. I removed the bad imports

maxov:
@stkem I added tests, this look good to merge?

stkem:
:+1:

PR #17193, opened 2015-09-08
Adding orgMemberAccess to LibraryCardInfo

andrewconner:
@adamjgrant Here's the field you need.


adamjgrant:
Great! Thanks @andrewconner

PR #17197, opened 2015-09-08
Feature/ext keep collaborate

cvializ:
@adamjgrant @andrewconner.

screen shot 2015-09-08 at 4 33 17 pm

We might want to consider having a separate section for Open Collaboration libraries, since it looks a little weird to have a ton of them down at the end:

screen shot 2015-09-08 at 4 33 00 pm


adamjgrant:
Cool :+1:

PR #17199, opened 2015-09-08
Duplicate name okay

camhashemi:
@andrewconner @gratimax 

Had a bug that wasn't surfaced until now


kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -144,7 +144,9 @@ class LibraryCommanderImpl @Inject() (
                   }
                   lib
                 case Some(lib) =>
-                  val newLib = libraryRepo.save(lib.copy(state = LibraryStates.ACTIVE))
```
camhashemi, line 147: we should never do this on objects that are sanitized for deletion

PR #17202, opened 2015-09-09
Use one-time-binding to show the add keep box

cvializ:
@andrewconner @adamjgrant force reevaluation until the feed is loaded


adamjgrant:
Elegant fix

adamjgrant:
:+1:

PR #17205, opened 2015-09-09
Fixed key-not-found exception for repo method

ryanpbrewster:
@andrewconner @cvializ maybe this is helpful?

I think it is good practice to make sure that any method that takes a `xs: Set[A]` and returns a `Map[A, B]` at least has `_.contains(x)` for all `x <- xs`.


ryanpbrewster:
I personally think it is better to do a `xs.map { x => x -> result.getOrElse(x, y0) }` than `result.withDefaultValue(y0)`, because the former has the property that `ans.keySet === xs`. Maps with default values at least won't throw `KeyNotFoundException`, but they don't have the "correct" key set.

andrewconner:
Thanks, was just about to commit a not-as-good solution! Works for me. Will merge in the morning just in case there's issues.

ryanpbrewster:
:+1: looks like it passes tests. Double-check my logic, if you don't mind, though. I'm not going to be able to hotfix issues if they crop up.

PR #17209, opened 2015-09-09
'user-or-org/:handle' returns invited organization info

camhashemi:
@andrewconner @gratimax @stkem

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileUserProfileController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileUserProfileController.scala
@@ -30,7 +30,7 @@ class MobileUserProfileController @Inject() (
   userValueRepo: UserValueRepo,
   orgMembershipRepo: OrganizationMembershipRepo,
   organizationCommander: OrganizationCommander,
-  orgInviteCommander: OrganizationInviteCommander,
+  orgInviteRepo: OrganizationInviteRepo,
   userCommander: UserCommander,
   userProfileCommander: UserProfileCommander,
   collectionCommander: CollectionCommander,
@@ -61,7 +61,7 @@ class MobileUserProfileController @Inject() (
           val numConnections = userConnectionRepo.getConnectionCount(profile.userId)
           val userBio = userValueRepo.getValueStringOpt(profile.userId, UserValueName.USER_DESCRIPTION)
           val orgMemberships = orgMembershipRepo.getAllByUserId(profile.userId)
-          val pendingOrgs = orgInviteCommander.getInvitesByInviteeAndDecision(profile.userId, InvitationDecision.PENDING).groupBy(_.organizationId).keys.map { orgId =>
+          val pendingOrgs = orgInviteRepo.getByInviteeIdAndDecision(profile.userId, InvitationDecision.PENDING).groupBy(_.organizationId).keys.map { orgId =>
             organizationCommander.getOrganizationInfo(orgId, viewerOpt.flatMap(_.id))
           }
           val orgInfos = orgMemberships.map { orgMembership => organizationCommander.getOrganizationInfo(orgMembership.organizationId, viewerOpt.flatMap(_.id)) }
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileUserProfileController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileUserProfileController.scala
@@ -30,6 +30,7 @@ class MobileUserProfileController @Inject() (
   userValueRepo: UserValueRepo,
   orgMembershipRepo: OrganizationMembershipRepo,
   organizationCommander: OrganizationCommander,
+  orgInviteCommander: OrganizationInviteCommander,
   userCommander: UserCommander,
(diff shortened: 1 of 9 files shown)


kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileUserProfileController.scala
```diff
@@ -56,12 +57,15 @@ class MobileUserProfileController @Inject() (
...
       case Some(profile) =>
         val (numLibraries, numCollabLibraries, numFollowedLibs, numInvitedLibs) = userProfileCommander.countLibraries(profile.userId, viewerOpt.map(_.id.get))
-        val (numConnections, userBiography, orgCards) = db.readOnlyMaster { implicit s =>
+        val (numConnections, userBiography, orgInfos, pendingOrgs) = db.readOnlyMaster { implicit s =>
           val numConnections = userConnectionRepo.getConnectionCount(profile.userId)
           val userBio = userValueRepo.getValueStringOpt(profile.userId, UserValueName.USER_DESCRIPTION)
           val orgMemberships = orgMembershipRepo.getAllByUserId(profile.userId)
-          val orgCards = orgMemberships.map { orgMembership => organizationCommander.getOrganizationInfo(orgMembership.organizationId, viewerOpt.flatMap(_.id)) }
-          (numConnections, userBio, orgCards)
+          val pendingOrgs = orgInviteCommander.getInvitesByInviteeAndDecision(profile.userId, InvitationDecision.PENDING).groupBy(_.organizationId).keys.map { orgId =>
```
andrewconner: db session-in-session. For `getInvitesByInviteeAndDecision` you can just call the repo method here?

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationInfo.scala
```diff
@@ -22,18 +22,18 @@ case class OrganizationInfo(
...
-    (__ \ 'handle).write[OrganizationHandle] and
-    (__ \ 'name).write[String] and
-    (__ \ 'description).writeNullable[String] and
-    (__ \ 'site).writeNullable[String] and
-    (__ \ 'avatarPath).write[ImagePath] and
-    (__ \ 'members).write[Seq[BasicUser]] and
-    (__ \ 'numMembers).write[Int] and
-    (__ \ 'numLibraries).write[Int]
-  )(unlift(OrganizationInfo.unapply))
+  implicit val defaultFormat: Format[OrganizationInfo] = (
```
andrewconner: I see you changed this to be a `Format` because of `UserProfileStats`. Can you change that one, instead, to be a `Writes`?

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileUserProfileController.scala
camhashemi: nice catch, indeed I can.

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationInfo.scala
camhashemi: `UserProfileStats` is a `Writes`. The reason I changed it is for deserialization in tests.

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileUserProfileController.scala
andrewconner: (To avoid the sessions problem, you'll need to move the `organizationCommander.getOrganizationInfo` outside this area)

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationInfo.scala
andrewconner: Oh I see, you're right. Any way to avoid that? Types with available `Reads` are begging to be used for request APIs.

camhashemi: I can create an explicit 'testReads' for tests to use. Anybody tempted to use it for request APIs should be warned off by the name or a comment. How's that sound?

andrewconner: Put the reads in the test itself and you have a deal :).

kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileUserProfileController.scala
camhashemi: seems like the problem is `getInvitesByInviteeAndDecision` (creates a session), `getOrganizationInfo` takes an implicit session

andrewconner: Oh cool, you good then. I was used to commanders not taking sessions. Yeah just move to the orgInviteRepo for the first call and you have it.

camhashemi: It is a bit weird. It could be called `getOrganizationInfoHelper` to inform us that it takes a session. I'm curious if you think the "Helper" pattern is useful.

PR #17212, opened 2015-09-09
Minor fixes, mostly dealing with org collaboration

andrewconner:
@adamjgrant This fixes your issue where org member access didn't come down. We now assume read write everywhere (serialization and deserialization) when appropriate if not set.

@cvializ This excludes secret libraries from the org keep list, and includes invited libraries. We'll have to check performance, but should be okay—I may add a cache to this. I _believe_ this fixes the issue you saw earlier. If not, bug me and we'll keep digging.

@JRuff42 This makes re-keeping from a keep always mean they you're keeping the URL, _not_ copying the keep.

@camhashemi This blocks non-active users from using the extension.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtAuthController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtAuthController.scala
@@ -69,6 +69,10 @@ class ExtAuthController @Inject() (
         })
 
     val (libraries, organizations, installation, urlPatterns, isInstall, isUpdate) = db.readWrite { implicit s =>
+      val user = request.user
+      if (user.state != UserStates.ACTIVE) {
+        throw new RuntimeException(s"User $userId (${user.firstName} ${user.lastName}) is not active, denying access.")
+      }
       val libraries = libraryInfoCommander.getMainAndSecretLibrariesForUser(userId)
       val orgIds = orgMembershipCommander.getAllOrganizationsForUser(userId)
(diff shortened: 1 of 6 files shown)


kifi-backend/modules/shoebox/app/com/keepit/controllers/ext/ExtAuthController.scala
```diff
@@ -69,6 +69,10 @@ class ExtAuthController @Inject() (
         })
 
     val (libraries, organizations, installation, urlPatterns, isInstall, isUpdate) = db.readWrite { implicit s =>
+      val user = request.user
+      if (user.state != UserStates.ACTIVE) {
+        throw new RuntimeException(s"User $userId (${user.firstName} ${user.lastName}) is not active, denying access.")
```
camhashemi: would it be possible to log them out of the extension as well?

andrewconner: Sure, will change. This endpoint speaks JSON so should return JSON, but this is a pretty nuclear option so I don't mind forcing our way. I _think_ it would work

camhashemi: It could return a JSON error that the client acts upon

andrewconner: Actually, handling this a different way. Blowing up here, telling the extension it's not logged in in another place.

adamjgrant:
thanks, @andrewconner

cvializ:
^ seconded :+1:

PR #17213, opened 2015-09-09
Squash an angular error about duplicate entries in a repeater

cvializ:


cvializ:
@adamjgrant small changes

adamjgrant:
Nice optimizations.

PR #17221, opened 2015-09-10
Can we agree that default params for case classes are a bad idea?

andrewconner:
Also, keeping variable names consistent.

@leogrim @camhashemi @adamjgrant


kifi-backend/modules/shoebox/app/com/keepit/model/FullLibraryInfo.scala
```diff
@@ -49,7 +49,7 @@ case class ExternalLibraryCreateRequest(
   whoCanInvite: Option[LibraryInvitePermissions] = None,
   subscriptions: Option[Seq[LibrarySubscriptionKey]] = None,
   space: Option[ExternalLibrarySpace] = None,
-  organizationMemberAccess: Option[LibraryAccess] = None)
+  orgMemberAccess: Option[LibraryAccess] = None)
```
andrewconner, line 52: Default param means: ”I don't care if someone forgets to include this; typical use case is that it's not needed"

PR #17225, opened 2015-09-10
Fixing one-column layout on Safari

adamjgrant:
Restores issue where homepage would show as only one column in Safari.


adamjgrant:
@andrewconner @cvializ

cvializ:
Epic

PR #17226, opened 2015-09-10
add pendingOrgs to /site/user/me

camhashemi:
@andrewconner @gratimax 

killed a couple session-in-sessions and a JSON matching test too


kifi-backend/modules/shoebox/test/com/keepit/controllers/website/UserControllerTest.scala
```diff
@@ -57,43 +57,15 @@ class UserControllerTest extends Specification with ShoeboxTestInjector {
         status(result) must equalTo(OK)
         contentType(result) must beSome("application/json")
 
-        val expected = Json.parse(s"""
```
maxov, line 60: :+1:

PR #17228, opened 2015-09-10
Fix notif time

maxov:
@stkem 
@andrewconner 
@leogrim 
@ryanpbrewster 

As it turns out, MySQL doesn't store milliseconds on dates, while h2 does. Locally this code worked fine since the returned dates all matched, but in MySQL the millisecond part of the date gets chopped off so that some returned dates are rounded, while others are not. This PR makes sure that read notifications and items all come from the datastore so that we are sure they have consistent time.

tl;dr I'm going to kill something when I happen upon another incompatibility between h2 and mysql...


stkem:
Actually, MySQL can store milliseconds on datetimes (as of 5.6), but you have to tell it to explicitly in the row definition

PR #17231, opened 2015-09-10
Turning on org messaging, turning off open collaboration

andrewconner:
@camhashemi @cvializ @eishay


andrewconner:
Getting it out to unbreak the ext

PR #17235, opened 2015-09-10
Adding extra sensitivity to textarea flow.

adamjgrant:
https://app.asana.com/0/42964066412484/49246930872814

@andrewconner @cvializ


cvializ:
I'm intrigued by the `##`. What does that do?

adamjgrant:
It just adds two characters to the element `clickToEdit` is looking at to judge how tall the textarea should be. With the two extra characters, that element will get taller two characters sooner than the textarea.

cvializ:
oh sneaky :+1:

PR #17236, opened 2015-09-10
Fix filtering open collaboration libraries

cvializ:
@andrewconner @adamjgrant It no longer shows you as a collaborator of Open Collaboration libraries when filtering.


adamjgrant:
:+1:

PR #17237, opened 2015-09-10
Adding org selector to library widget

adamjgrant:
https://app.asana.com/0/42964066412484/49246122596873

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/common/directives/orgSelector/orgSelector.js
+++ b/kifi-backend/modules/shoebox/angular/src/common/directives/orgSelector/orgSelector.js
@@ -1,0 +1,56 @@
+'use strict';
+
+angular.module('kifi')
+
+.directive('kfOrgSelector', ['profileService',
+  function(profileService) {
+    return {
+      restrict: 'A',
+      templateUrl: 'common/directives/orgSelector/orgSelector.tpl.html',
+      $scope: { 
+        libraryProps: '='
+      },
+      link: function($scope) {
+        $scope.me = profileService.me;
+
+        $scope.unsetOrg = function() {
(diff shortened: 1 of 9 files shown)


adamjgrant:
screen shot 2015-09-10 at 5 37 53 pm
screen shot 2015-09-10 at 5 37 58 pm

adamjgrant:
@andrewconner @cvializ

cvializ:
The top check mark looks cut off to me

andrewconner:
Nice. The spacing of the bottom privacy section (very airy) and above it in the location section (tight) seem to conflict to me. What do you think?

adamjgrant:
@cvializ You might be right, I'll take a look.

@andrewconner Yeah, maybe we can do something about that. We should also consider the spacing that appears on hover, it's justified a bit more

screen shot 2015-09-10 at 5 42 03 pm

That being said, I think we can tighten it. Stay tuned.

andrewconner:
Thanks for the screenshots!

cvializ:
^ what he said

adamjgrant:
screen shot 2015-09-10 at 5 44 08 pm
screen shot 2015-09-10 at 5 44 18 pm

I thought it might have been due to `overflow: hidden` somewhere, but it's actually not cut off. However, it does appear differently than on the manage library modal.

Anyway, here are tweaks @andrewconner @cvializ 

screen shot 2015-09-10 at 5 52 21 pm

andrewconner:
I like

cvializ:
Oh I think it's the border radius is too large for the size of the
checkmark. Try decreasing those radii

adamjgrant:
(discussed with @cvializ offline)

PR #17239, opened 2015-09-10
[WIP] payments permissions progress

camhashemi:
@stkem @gratimax @andrewconner


kifi-backend/modules/shoebox/app/com/keepit/payments/PaidAccount.scala
```diff
@@ -30,7 +30,8 @@ case class PaidAccount(
     userContacts: Seq[Id[User]],
     emailContacts: Seq[EmailAddress],
     lockedForProcessing: Boolean = false,
-    frozen: Boolean = false) extends ModelWithState[PaidAccount] {
+    frozen: Boolean = false,
+    settingsConfiguration: PlanSettingsConfiguration) extends ModelWithState[PaidAccount] {
```
stkem: Gentle reminder to add an evolution for this new field

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidAccountRepo.scala
```diff
@@ -39,7 +41,8 @@ class PaidAccountRepoImpl @Inject() (
     def emailContacts = column[Seq[EmailAddress]]("email_contacts", O.NotNull)
     def lockedForProcessing = column[Boolean]("locked_for_processing", O.NotNull)
     def frozen = column[Boolean]("frozen", O.NotNull)
-    def * = (id.?, createdAt, updatedAt, state, orgId, planId, credit, userContacts, emailContacts, lockedForProcessing, frozen) <> ((PaidAccount.apply _).tupled, PaidAccount.unapply _)
+    def settingsConfig = column[PlanSettingsConfiguration]("settings_configuration", O.NotNull)
```
stkem: Name doesn't match up with the line below.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
```diff
@@ -11,17 +14,9 @@ import org.joda.time.DateTime
...
-case class PlanFeature(name: String, displayName: String, editable: Boolean, default: Boolean)
+import play.api.libs.json.Format
 
-@json
-case class PlanFeatureSetting(feature: PlanFeature, enabled: Boolean)
-
-object PlanFeatureSetting {
-  def toBasePermissions(settings: Seq[PlanFeatureSetting]): BasePermissions = ??? //ZZZ TODO: When we know what the right permissions actually are.
-}
+case class BillingCycle(month: Int) extends AnyVal
```
stkem, line 21: Did you remove the @json annotation from this, or is the diff just weird?

stkem, line 21: Oops, never mind. Never had one and doesn't need it.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
```diff
@@ -63,3 +59,77 @@ object PaidPlan extends ModelWithPublicIdCompanion[PaidPlan] {
 }
 
 object PaidPlanStates extends States[PaidPlan]
+
+sealed abstract class PlanFeature(val name: String, val editable: Boolean, val default: Boolean)
```
stkem: I know I didn't do this in my previous version either, but perhaps it makes sense here to use `Name[PlanFeature]` instead of `String` directly for `name`.
Also keep in mind that `default` here is going to have to become something a little more generic, to at least accommodate admin-only/everyone/no-one settings (and possibly more). Same goes for `enabled` on `PlanFeatureSetting`.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
```diff
@@ -63,3 +59,77 @@ object PaidPlan extends ModelWithPublicIdCompanion[PaidPlan] {
...
+object PlanFeature {
+  case object ADD_PUBLIC_LIBRARY extends PlanFeature(name = "public_library_creation", editable = true, default = true)
+  case object INVITE_MEMBERS extends PlanFeature(name = "invite_members", editable = true, default = true)
+  case object GROUP_MESSAGING extends PlanFeature(name = "group_messaging", editable = true, default = true)
+  case object MOVE_LIBRARY extends PlanFeature(name = "move_library", editable = true, default = true)
+  case object EXPORT_KEEPS extends PlanFeature(name = "export_keeps", editable = false, default = true)
+  case object VIEW_MEMBERS extends PlanFeature(name = "view_members", editable = false, default = true)
+  case object EDIT_LIBRARY extends PlanFeature(name = "edit_library", editable = true, default = true)
```
stkem: I'm not sure it makes sense to have these as globally static objects, as they may (and probably often will) depend on the plan they belong to. E.g. a lot of the features will have `editable=false` on the free plan and `editable=true` on the pro plan. Maybe just have the names as static constants?

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
```diff
@@ -63,3 +59,77 @@ object PaidPlan extends ModelWithPublicIdCompanion[PaidPlan] {
...
+
+  implicit val format: Format[PlanFeature] = new Format[PlanFeature] {
```
stkem: Is this format used anywhere yet?

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
```diff
@@ -63,3 +59,77 @@ object PaidPlan extends ModelWithPublicIdCompanion[PaidPlan] {
...
+    def reads(json: JsValue): PlanFeature = PlanFeature((json \ "name").as[String])
+    def writes(o: PlanFeature): JsValue = Json.obj("name" -> o.name)
+  }
+
+  def apply(name: String): PlanFeature = {
+    name match {
+      case ADD_PUBLIC_LIBRARY.name => ADD_PUBLIC_LIBRARY
```
stkem: I'm not sure this will work. See my comment above about feature values depending on the plan.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
```diff
@@ -63,3 +59,77 @@ object PaidPlan extends ModelWithPublicIdCompanion[PaidPlan] {
...
+  def toPermission(feature: PlanFeature) = {
+    feature match {
+      case ADD_PUBLIC_LIBRARY => OrganizationPermission.ADD_PUBLIC_LIBRARIES
+      case INVITE_MEMBERS => OrganizationPermission.INVITE_MEMBERS
+      case GROUP_MESSAGING => OrganizationPermission.GROUP_MESSAGING
+      case MOVE_LIBRARY => OrganizationPermission.MOVE_LIBRARY
+      case EXPORT_KEEPS => OrganizationPermission.EXPORT_KEEPS
+      case VIEW_MEMBERS => OrganizationPermission.VIEW_MEMBERS
+      case EDIT_LIBRARY => OrganizationPermission.FORCE_EDIT_LIBRARIES
+      case _ => OrganizationPermission("none") // TODO: handle features which don't map to a permission
```
stkem: Maybe throw an exception here until we have such features, so we can't forget about it.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
```diff
@@ -63,3 +59,77 @@ object PaidPlan extends ModelWithPublicIdCompanion[PaidPlan] {
...
+    }
+  }
+}
+
+sealed abstract class PlanFeatureSetting(enabled: Boolean, value: String)
```
stkem: I'm not sure I understand the meaning of `enabled` and `value` here. Could you explain what represents what?

stkem:
Made some comments. Right direction! :+1:

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
camhashemi: `enabled` means the setting should be enforced and `value` is the toggle of the setting, they don't make sense without a `PlanFeature` attached.

I removed `PlanFeature` from here, thinking that `PlanSettingsConfiguration` would create a mapping between `PlanFeature` and `PlanFeatureSetting`. But I'm starting to favor the way you initially had it, using a `Set[PlanFeatureSetting]` to determine a plan's configuration.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
camhashemi: I misunderstood `editable` and `default` to be static. You're right, we can just use the names.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
camhashemi: It will be used in `PaidPlan`

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
stkem: This may have been my bad in explaining them as static. I wasn't thinking about across plans at the time.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
stkem: Re `enabled` and `value`: The `enabled` field initially was just the setting. I don't think we need a separate field to determine wether a setting should be enforced. All settings should be enforced.
As for `value` that should be some sort of strongly typed entity, not just a string (in relation to the `options` thing we discussed in slack)

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
camhashemi: Something about features needs to be static: `name`, `options`, and `default` are static (and diverse) per feature. `setting` and `editable` vary per plan/account.

PR #17241, opened 2015-09-10
Send notifications and message chats for new system

maxov:
I decided to narrow the scope a bit and take out the stuff dealing with message chats. Now the system only works with notifications, but it is functional and finished. It's ready for us to turn on the experiment and QA it. I also added a backfill plugin that will run over the DB and backfill legacy notifications so we can take out non-replyable message threads and users when it finishes.

### DIFF
--- a/kifi-backend/modules/eliza/app/com/keepit/eliza/commanders/NotificationCommander.scala
+++ b/kifi-backend/modules/eliza/app/com/keepit/eliza/commanders/NotificationCommander.scala
@@ -143,6 +143,14 @@ class NotificationCommander @Inject() (
 
   }
 
+  def getNotifForMessageThread(userId: Id[User], messageThreadId: Id[MessageThread]): Option[Notification] = {
+    db.readOnlyMaster { implicit session =>
+      userThreadRepo.getUserThread(userId, messageThreadId).notificationId.map { id =>
+        notificationRepo.get(id)
+      }
+    }
+  }
+
   /**
    * Gets the message thread that is potentially associated with a notification. This should only return a message thread
    * if the notification kind is of [[com.keepit.notify.model.event.NewMessage]], as that is the only kind associated with
--- a/kifi-backend/modules/eliza/app/com/keepit/eliza/commanders/NotificationCommander.scala
+++ b/kifi-backend/modules/eliza/app/com/keepit/eliza/commanders/NotificationCommander.scala
@@ -68,14 +68,13 @@ class NotificationCommander @Inject() (
           notificationRepo.getByGroupIdentifier(recipient, LegacyNotification, id).fold {
             val threadId = (json \ "thread").as[String]
             val messageThread = messageThreadRepo.get(ExternalId[MessageThread](threadId))
-            if (messageThread.replyable) {
-              backfillMessageThreadForUser(userId, messageThread.id.get)
-            } else {
+            val userThread = userThreadRepo.getUserThread(userId, messageThread.id.get)
+            if (!messageThread.replyable) {
               val notif = notificationRepo.save(Notification(
                 recipient = recipient,
                 kind = LegacyNotification,
                 lastEvent = time,
-                groupIdentifier = Some(id)
+                groupIdentifier = Some(userThread.id.get.id.toString)
               ))
               val item = notificationItemRepo.save(NotificationItem(
                 notificationId = notif.id.get,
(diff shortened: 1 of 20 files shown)


maxov:
@stkem 
@andrewconner 
@leogrim 
@ryanpbrewster

kifi-backend/modules/eliza/app/com/keepit/eliza/commanders/NotificationCommander.scala
```diff
@@ -88,30 +87,30 @@ class NotificationCommander @Inject() (
...
-    val userThread = userThreadRepo.getUserThread(userId, messageThread.id.get)
+    val userThread = userThreadRepo.getUserThread(userId, messageThreadId)
     val lastChecked = Seq(userThread.lastActive, userThread.lastSeen, userThread.notificationLastSeen).collect {
       case Some(time) => time
     }.maxOpt
-    val messages = messageRepo.get(messageThread.id.get, 0)
+    val messages = messageRepo.get(messageThreadId, 0)
     val lastEvent = messages.map(_.createdAt).max
-    val groupIdentifier = messageThread.id.get.toString
+    val groupIdentifier = messageThreadId.toString
```
stkem: Using `toString` for something like this _can_ be dangerous, as it is often understood to be for human consumption and thus someone might change for whatever reason (e.g. they don't like how things look in the logs) and inadvertently brake things in other places that are really hard to catch pre production. I would suggest either having a method on the class that produces a string with clear naming independent of `toString`, or getting the primitive out first and calling it's `toString`, which imho can be considered stable.
(If this is already a primitive (not clear form the diff alone) I take it all back.

kifi-backend/modules/eliza/app/com/keepit/eliza/model/NotificationRepo.scala
```diff
@@ -90,7 +90,7 @@ class NotificationRepoImpl @Inject() (
 
   def getNotificationsWithNewEventsCount(recipient: Recipient)(implicit session: RSession): Int = {
     val unread = for (
-      row <- rows if row.recipient === recipient && row.hasNewEvent
+      row <- rows if row.recipient === recipient && row.hasNewEvent && row.kind =!= "legacy"
```
stkem: I should've probably noted this earlier but missed it: How about typing the `kind` field, or at least having some enum like things with the possible kinds, so you don't have to use string literals in multiple places. I.e. this would become `row.kind =!= NotificationKind.LEGACY` or something like that.

kifi-backend/modules/eliza/app/com/keepit/notify/NotificationBackfiller.scala
```diff
@@ -0,0 +1,66 @@
...
+  override def enabled: Boolean = true
+  override def onStart() {
+    val (initDelay, freq) = if (Play.isDev) (15 seconds, 15 seconds) else (5 minutes, 3 minutes)
+    log.info(s"[onStart] NotificationBackfillerPlugin started with initDelay=$initDelay freq=$freq")
+    scheduleTaskOnOneMachine(actor.system, initDelay, freq, actor.ref, BackfillNotifications, BackfillNotifications.getClass.getName)
+  }
+
+}
+
+class NotificationBackfiller @Inject() (
```
stkem, line 39: You may want to consider throttling this thing is some way for the initial backfill (unless it already is and I'm not seeing it). Could bring Eliza to it's knees otherwise.

stkem:
code lgtm. As before, do we have test coverage on the modified websocket endpoints now? (i.e. to make sure it's not breaking the format, although admittedly, dev mode testing goes a long way here, which I'm sure you are doing)

kifi-backend/modules/eliza/app/com/keepit/notify/NotificationBackfiller.scala
maxov, line 39: It runs a query that operates on 500 old user threads each time. The next time the query runs, it backfills the next 500 threads. I don't know if that's too many at a time, because it does happen sequentially, in different db sessions

kifi-backend/modules/eliza/app/com/keepit/eliza/model/NotificationRepo.scala
maxov: I can replace this with LegacyNotification.name, it is an object after all

kifi-backend/modules/eliza/app/com/keepit/eliza/commanders/NotificationCommander.scala
maxov: You're right, I forgot about this.

kifi-backend/modules/eliza/app/com/keepit/notify/NotificationBackfiller.scala
stkem, line 39: Oh, no, I misunderstood something in the code. You're totally good.

stkem:
:+1:

PR #17243, opened 2015-09-11
Fix corner case from out of sync data.

cvializ:
@adamjgrant

Steps to reproduce:
- A library owner is the sole collaborator on a library
- The owner invites another user to collaborate on the library from the library page.
- The other user accepts on their browser
- The inviter, without refreshing the page, clicks "See all" and opens the collaborator list
- The list of collaborators shows the new collaborator, but the page behind is missing them.
- The inviter removes the new user

Expected outcome:
- The library header still shows the owner

Actual outcome
- The library header no longer shows the owner

It was just an off-by-one sort of problem

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/libraryHeader.js
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/libraryHeader.js
@@ -708,10 +708,14 @@ angular.module('kifi')
             var lib = scope.library;
             if (lib && libraryId === lib.id && lib.membership) {
               if (lib.membership.access === 'read_only') {
-                lib.numFollowers--;
+                if (lib.numFollowers > 0) {
+                  lib.numFollowers--;
+                }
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/libraryHeader.js
```diff
@@ -708,10 +708,14 @@ angular.module('kifi')
...
+                if (lib.numFollowers > 0) {
+                  lib.numFollowers--;
+                }
                 _.remove(lib.followers, {id: profileService.me.id});
               } else if (lib.membership.access === 'read_write') {
-                lib.numCollaborators--;
+                if (lib.numCollaborators > 0) {
+                  lib.numCollaborators--;
+                }
                 _.remove(lib.collaborators, {id: profileService.me.id});
```
adamjgrant, line 719: Nice safeguards.

adamjgrant:
:+1:

PR #17244, opened 2015-09-11
Fix feed in safari and delighted display

cvializ:
@adamjgrant Derek mentioned delighted was showing up funny, which I knew resulted from our row-wrap fix earlier. I went ahead and knocked it out


andrewconner:
:+1:

PR #17248, opened 2015-09-11
Fixing bug where space did not change on modify of library

adamjgrant:


adamjgrant:
@andrewconner @cvializ This should fix for at least the managelibrary modal. Will get this in now and continue testing.

cvializ:
:+1: awesome

PR #17249, opened 2015-09-11
Quick fix to remaining bug with keep card widget

adamjgrant:
Suppresses `undefined` warning for keep to library widget


cvializ:
:+1: nice catch

PR #17254, opened 2015-09-11
Payments model

camhashemi:
@stkem @andrewconner @gratimax a cleaner version of the `PlanFeature` settings and permissions model


kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
```diff
@@ -63,3 +61,107 @@ object PaidPlan extends ModelWithPublicIdCompanion[PaidPlan] {
...
+}
+
+trait PlanFeature {
+  val name: Name[PlanFeature]
+  val editable: Boolean
+  val options: Seq[Setting]
+  val default: Setting
+}
+
+case class PermissionsFeature(name: Name[PlanFeature], editable: Boolean, options: Seq[PermissionSetting], default: PermissionSetting) extends PlanFeature
```
maxov: :+1:

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidAccount.scala
```diff
@@ -30,7 +35,8 @@ case class PaidAccount(
     userContacts: Seq[Id[User]],
     emailContacts: Seq[EmailAddress],
     lockedForProcessing: Boolean = false,
-    frozen: Boolean = false) extends ModelWithState[PaidAccount] {
+    frozen: Boolean = false,
+    settingsConfiguration: Map[Name[PlanFeature], Setting]) extends ModelWithState[PaidAccount] {
```
camhashemi: @stkem we might want to put the whole `PlanFeature` in here, so that we only have to do one lookup to populate a settings page.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
```diff
@@ -63,3 +61,107 @@ object PaidPlan extends ModelWithPublicIdCompanion[PaidPlan] {
...
+  implicit val format = Format(__.read[String].map(Setting(_)), new Writes[Setting] { def writes(o: Setting): JsValue = JsString(o.value) })
+
+  def apply(str: String): Setting = str match {
+    case str if Try(OrganizationRole(str)).isSuccess => PermissionSetting(Some(OrganizationRole(str)))
+    case "none" => PermissionSetting(None)
+    case _ => throw new Exception(s"called Setting.apply with an unsupported value $str")
+  }
+}
+
+case class FeatureSetting(settingsByFeature: Map[Name[PlanFeature], Setting])
```
maxov: The name here might be confusing since this doesn't actually extend Setting. maybe FeatureSettings or FeatureSettingsMap

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidAccountRepo.scala
```diff
@@ -29,6 +30,10 @@ class PaidAccountRepoImpl @Inject() (
   import db.Driver.simple._
 
   implicit val dollarAmountColumnType = MappedColumnType.base[DollarAmount, Int](_.cents, DollarAmount(_))
+  implicit val settingsConfigColumnType = MappedColumnType.base[Map[Name[PlanFeature], Setting], String](
```
maxov: These _might_ belong in FortyTwoGenericTypeMappers, but I'm not sure. Might be too specific to plans

kifi-backend/modules/shoebox/app/com/keepit/payments/PlanManagementCommander.scala
```diff
@@ -490,5 +496,17 @@ class PlanManagementCommanderImpl @Inject() (
     )
   }
 
+  def getPlanFeatureSettings(orgId: Id[Organization]): Map[Name[PlanFeature], FeatureSetting] = {
+    val settingsConfig = db.readOnlyReplica { implicit session => paidAccountRepo.getByOrgId(orgId).settingsConfiguration }
+    Map.empty
```
maxov: This intended?

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidAccountRepo.scala
camhashemi: Maybe, only a few things in there are truly "generic"

kifi-backend/modules/shoebox/app/com/keepit/payments/PlanManagementCommander.scala
camhashemi: the main focus of this PR atm should be on the models. Everything else is mocked / tied off.

kifi-backend/modules/shoebox/test/com/keepit/model/PaidAccountFactoryHelper.scala
```diff
@@ -0,0 +1,16 @@
+package com.keepit.model
+
+import com.google.inject.Injector
+import com.keepit.common.db.slick.DBSession.RWSession
+import com.keepit.model.PaidAccountFactory.PartialPaidAccount
+import com.keepit.payments.{ PaidAccount, PaidAccountRepo }
+
+object PaidAccountFactoryHelper {
+  implicit class PaidAccountPersister(partialPaidAccount: PartialPaidAccount) {
+    def saved(implicit injector: Injector, session: RWSession): PaidAccount = {
```
maxov, line 10: (not really about the PR) I haven't seen this before, but I actually quite like this pattern. I think there must be a way to do it generically so that we can have this for all tests and repos

kifi-backend/modules/shoebox/app/com/keepit/payments/PlanManagementCommander.scala
maxov: then perhaps add a todo of some sort, just so that we know someone will eventually implement it

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
```diff
@@ -63,3 +61,107 @@ object PaidPlan extends ModelWithPublicIdCompanion[PaidPlan] {
...
+
+object FeatureSetting {
+
+  implicit val format = new Format[FeatureSetting] {
+    def writes(o: FeatureSetting): JsValue = JsObject(o.settingsByFeature.map { case (name, setting) => name.name -> JsString(setting.value) }.toSeq)
+    def reads(json: JsValue): JsResult[FeatureSetting] = {
+      val settingsByFeatureTry = Try(json.as[JsObject].fields.map {
```
maxov: better to do `json.validate[JsObject].map _` so that it returns a JsResult and you don't have to do the pattern match below

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
```diff
@@ -63,3 +61,107 @@ object PaidPlan extends ModelWithPublicIdCompanion[PaidPlan] {
...
+        case (name, setting) if PermissionsFeatureNames.ALL.contains(Name[PlanFeature](name)) =>
+          if (setting.as[String] == "none") Name[PlanFeature](name) -> PermissionSetting(role = None)
+          else Name[PlanFeature](name) -> PermissionSetting(Some(OrganizationRole(setting.as[String])))
+        case _ => throw new Exception(s"tried to read invalid json=$json as a FeatureSetting")
```
maxov: IllegalArgumentException?

kifi-backend/modules/shoebox/test/com/keepit/model/PaidAccountFactoryHelper.scala
camhashemi, line 10: it makes writing tests so much easier. Luckily, it's easy to create a `Factory` class. There's usually a way to do things generically, but the amount of work it takes to create and use something incredibly generic is usually not worth it.

maxov:
:100: looks quite nice to me

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
camhashemi: I think this code could be improved, but we're doing the pattern match on the `Try` to know whether or not the serialization was valid. I think having `JsResult` early on would add another pattern match.

kifi-backend/modules/common/app/com/keepit/model/OrganizationPermission.scala
```diff
@@ -7,23 +7,33 @@ sealed abstract class OrganizationPermission(val value: String)
...
   case object EDIT_ORGANIZATION extends OrganizationPermission("edit_organization")
   case object INVITE_MEMBERS extends OrganizationPermission("invite_members")
   case object MODIFY_MEMBERS extends OrganizationPermission("modify_members")
   case object REMOVE_MEMBERS extends OrganizationPermission("remove_members")
   case object ADD_LIBRARIES extends OrganizationPermission("add_libraries")
+  case object ADD_PUBLIC_LIBRARIES extends OrganizationPermission("add_public_libraries")
   case object REMOVE_LIBRARIES extends OrganizationPermission("remove_libraries")
   case object FORCE_EDIT_LIBRARIES extends OrganizationPermission("force_edit_libraries")
+  case object GROUP_MESSAGING extends OrganizationPermission("group_messaging")
+  case object MOVE_LIBRARY extends OrganizationPermission("move_library")
```
ryanpbrewster: `MOVE_LIBRARIES`?

kifi-backend/modules/common/app/com/keepit/model/OrganizationPermission.scala
```diff
@@ -7,23 +7,33 @@ sealed abstract class OrganizationPermission(val value: String)
 
 object OrganizationPermission {
   case object VIEW_ORGANIZATION extends OrganizationPermission("view_organization")
+  case object VIEW_MEMBERS extends OrganizationPermission("view_members")
   case object EDIT_ORGANIZATION extends OrganizationPermission("edit_organization")
   case object INVITE_MEMBERS extends OrganizationPermission("invite_members")
   case object MODIFY_MEMBERS extends OrganizationPermission("modify_members")
   case object REMOVE_MEMBERS extends OrganizationPermission("remove_members")
   case object ADD_LIBRARIES extends OrganizationPermission("add_libraries")
+  case object ADD_PUBLIC_LIBRARIES extends OrganizationPermission("add_public_libraries")
```
ryanpbrewster: `MAKE_LIBRARIES_PUBLIC` or `PUBLISH_LIBRARIES`? To make it clear that also deals with modifying an org visible library.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
```diff
@@ -63,3 +61,107 @@ object PaidPlan extends ModelWithPublicIdCompanion[PaidPlan] {
...
+}
+
+object PermissionsFeatureNames {
+  val ADD_PUBLIC_LIBRARY = Name[PlanFeature]("public_library_creation")
+  val INVITE_MEMBERS = Name[PlanFeature]("invite_members")
+  val GROUP_MESSAGING = Name[PlanFeature]("group_messaging")
+  val MOVE_LIBRARY = Name[PlanFeature]("move_library")
+  val EXPORT_KEEPS = Name[PlanFeature]("export_keeps")
+  val VIEW_MEMBERS = Name[PlanFeature]("view_members")
+  val EDIT_LIBRARY = Name[PlanFeature]("edit_library")
```
ryanpbrewster: Can we just do `OrganizationPermission.FORCE_EDIT_LIBRARIES.value`? That way the compiler can catch most of our spelling errors and we only have to do things right once.

kifi-backend/modules/shoebox/test/com/keepit/payments/PaidFeatureSettingsTest.scala
```diff
@@ -0,0 +1,127 @@
...
+
+        // admin can create public libraries
+        val adminCreateRequest = LibraryCreateRequest(name = "Alphabetter Soup", slug = "alphabetter", visibility = LibraryVisibility.PUBLISHED, space = Some(LibrarySpace(admin.id.get, org.id)))
+        val adminLibResponse = libraryCommander.createLibrary(adminCreateRequest, admin.id.get)
+        adminLibResponse.isRight === true
+
+        // member cannot create public libraries
+        val memberCreateRequest = LibraryCreateRequest(name = "Alphabest Soup", slug = "alphabest", visibility = LibraryVisibility.PUBLISHED, space = Some(LibrarySpace(member.id.get, org.id)))
+        val memberLibResponse = libraryCommander.createLibrary(memberCreateRequest, member.id.get)
+        memberLibResponse.isRight === false
```
ryanpbrewster: `memberLibResponse must beLeft` produces more useful error messages.

kifi-backend/modules/shoebox/test/com/keepit/payments/PaidFeatureSettingsTest.scala
```diff
@@ -0,0 +1,127 @@
...
+    "be configurable" in {
+      withDb(modules: _*) { implicit injector =>
+
+        //
+        // configure PaidFeature set and PaidFeatureSettings
+        //
+
+        val (org, owner, admin, member, paidPlan, paidAccount) = setup()
+
+        val invitees = db.readWrite { implicit sessoin => UserFactory.users(3).saved }
```
ryanpbrewster: `session`

kifi-backend/modules/shoebox/test/com/keepit/payments/PaidFeatureSettingsTest.scala
```diff
@@ -0,0 +1,127 @@
...
+
+        val orgInviteCommander = inject[OrganizationInviteCommander]
+        val ownerInviteRequest = OrganizationInviteSendRequest(org.id.get, owner.id.get, targetEmails = Set.empty, targetUserIds = Set(invitees(0).id.get))
+        val ownerInviteResponse = Await.result(orgInviteCommander.inviteToOrganization(ownerInviteRequest), Duration(5, "seconds"))
```
ryanpbrewster, line 104: I usually just do `Duration.Inf`. No strong feelings on this though.

kifi-backend/modules/common/app/com/keepit/model/OrganizationPermission.scala
stkem: :+1:

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidAccountRepo.scala
stkem: Yeah, `FortyTwoGenericTypeMappers` is already kind of unpleasantly large. I don't really feel strongly about this, but I'd advocate for keeping them here until uses in a bunch of places come up.

kifi-backend/modules/shoebox/test/com/keepit/payments/PaidFeatureSettingsTest.scala
stkem, line 104: I'm actually a fan of using relatively generous but short durations in tests. It's a free, if very fuzzy, performance check. If the awaits start to fail you know you've got a problem somewhere.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
```diff
@@ -63,3 +61,107 @@ object PaidPlan extends ModelWithPublicIdCompanion[PaidPlan] {
...
+  implicit val format = new Format[FeatureSetting] {
+    def writes(o: FeatureSetting): JsValue = JsObject(o.settingsByFeature.map { case (name, setting) => name.name -> JsString(setting.value) }.toSeq)
+    def reads(json: JsValue): JsResult[FeatureSetting] = {
+      val settingsByFeatureTry = Try(json.as[JsObject].fields.map {
+        case (name, setting) if PermissionsFeatureNames.ALL.contains(Name[PlanFeature](name)) =>
+          if (setting.as[String] == "none") Name[PlanFeature](name) -> PermissionSetting(role = None)
+          else Name[PlanFeature](name) -> PermissionSetting(Some(OrganizationRole(setting.as[String])))
+        case _ => throw new IllegalArgumentException(s"tried to read invalid json=$json as a FeatureSetting")
+      }.toMap)
+      settingsByFeatureTry match {
```
stkem: I think it's a little weird to use a `Try` in this manner, i.e. wrap a value and then immediately process it. `Try` is great for passing around something that might be an error (a capability already covered here by `JsResult`) but if you are going to handle all errors immediately anyway, I think a regular `try {} catch {}` block makes things more readable.

stkem:
Very nice! :+1:

kifi-backend/modules/common/app/com/keepit/model/OrganizationPermission.scala
camhashemi: I'm renaming it to `RECLAIM_LIBRARIES`, which is more accurate to the user story: users without this permission can't move any libraries back into their personal space. Does that name make sense?

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
camhashemi: :+1: good idea

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidAccount.scala
```diff
@@ -30,7 +35,8 @@ case class PaidAccount(
     userContacts: Seq[Id[User]],
     emailContacts: Seq[EmailAddress],
     lockedForProcessing: Boolean = false,
-    frozen: Boolean = false) extends ModelWithState[PaidAccount] {
+    frozen: Boolean = false,
+    settingsByFeature: Map[Name[PlanFeature], Setting]) extends ModelWithState[PaidAccount] {
```
camhashemi: @andrewconner this is an org's settings at any given moment

kifi-backend/modules/common/app/com/keepit/model/OrganizationPermission.scala
```diff
@@ -7,23 +7,33 @@ sealed abstract class OrganizationPermission(val value: String)
...
   case object REMOVE_LIBRARIES extends OrganizationPermission("remove_libraries")
   case object FORCE_EDIT_LIBRARIES extends OrganizationPermission("force_edit_libraries")
+  case object GROUP_MESSAGING extends OrganizationPermission("group_messaging")
+  case object MOVE_LIBRARY extends OrganizationPermission("move_library")
+  case object EXPORT_KEEPS extends OrganizationPermission("export_keeps")
```
camhashemi, line 21: @jaredpetker @DerekBurch anything in this list is a permission that will be sent back with a member's permissions

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
```diff
@@ -63,3 +61,107 @@ object PaidPlan extends ModelWithPublicIdCompanion[PaidPlan] {
...
+  implicit val format = Format(__.read[String].map(Setting(_)), new Writes[Setting] { def writes(o: Setting): JsValue = JsString(o.value) })
+
+  def apply(str: String): Setting = str match {
+    case str if Try(OrganizationRole(str)).isSuccess => PermissionSetting(Some(OrganizationRole(str)))
+    case "none" => PermissionSetting(None)
+    case _ => throw new IllegalArgumentException(s"called Setting.apply with an unsupported value $str")
+  }
+}
+
+case class FeatureSetting(settingsByFeature: Map[Name[PlanFeature], Setting])
```
stkem: Maybe `FeatureSettings`, since it's really multiple?

kifi-backend/modules/common/app/com/keepit/model/OrganizationPermission.scala
andrewconner: Hrm, what if it was never in the personal location to begin with? `move_library_away_from_org` too long?

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/PaymentsController.scala
```diff
@@ -85,22 +85,16 @@ class PaymentsController @Inject() (
   }
 
   def getPlanFeatureSettings(pubId: PublicId[Organization]) = OrganizationUserAction(pubId, PLAN_MANAGEMENT_PERMISSION) { request => //ZZZ TODO: This is currently a dummy
```
andrewconner: Does this do anything anymore?

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
```diff
@@ -63,3 +61,107 @@ object PaidPlan extends ModelWithPublicIdCompanion[PaidPlan] {
 }
 
 object PaidPlanStates extends States[PaidPlan]
+
+sealed abstract class Setting(val value: String) {
+  override def toString = this.value
+}
+@json
+case class PermissionSetting(role: Option[OrganizationRole]) extends Setting(role.map(_.value).getOrElse("none"))
```
andrewconner: `RoleSetting` perhaps? This is always wrapping a role, right? (or in the case of no role, it's explicitly talking about the `None` role)

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
```diff
@@ -63,3 +61,107 @@ object PaidPlan extends ModelWithPublicIdCompanion[PaidPlan] {
 }
 
 object PaidPlanStates extends States[PaidPlan]
+
+sealed abstract class Setting(val value: String) {
+  override def toString = this.value
```
andrewconner: I'm wary of this pattern, because it encourages you to call `toString` (often implicitly) to get the contents, when you should do `yourSetting.value`. We've had breakage before when someone changed the `toString`, thinking it was for presentation only.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
```diff
@@ -63,3 +61,107 @@ object PaidPlan extends ModelWithPublicIdCompanion[PaidPlan] {
...
+      settingsByFeatureTry match {
+        case Failure(errs) => JsError(errs.getStackTrace.toString)
+        case Success(settingsByFeature) => JsSuccess(FeatureSetting(settingsByFeature))
+      }
+    }
+  }
+
+  def toBasePermissions(settings: FeatureSetting): BasePermissions = {
+    val namesAndSettings = settings.settingsByFeature.toSet
+    val permissionsByRole = namesAndSettings.collect { case (name, setting: PermissionSetting) => setting.role -> PlanFeature.toPermission(name) }
```
andrewconner: `setting: Setting`?

andrewconner: Oh I see, you're just ignoring all the other ones. Perhaps leave a comment saying so?

camhashemi: `PermissionSetting` is typed as such because it is map-able to values in `BasePermissions`. I can leave a comment.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
camhashemi: Yes, it does. I think the name as is links it to `BasePermissions`, which is how it should be.

kifi-backend/modules/common/app/com/keepit/model/OrganizationPermission.scala
camhashemi, line 21: *sent back or not sent back

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
stkem: On second read, I think this should be a set of roles, as, e.g. the "everyone" choice we have on a bunch of settings would map the appropriate permission to the `admin` and `member` roles.

andrewconner: > I think this should be a set of roles
> So the client sends a set, or the conversion happens before here?
> 
> I think the name as is links it to BasePermissions
> That's a map from role to a set of permissions, right? This is just the role part.

Also, I think it's weird to have a special string for the none case. Seems close to a [Magic String](https://en.wikipedia.org/wiki/Magic_string). If this holds a set of roles, then this class's concerns are away from the external API (which won't be passing in a set? or will?)—strange to keep around a special string (`"none"`) this far

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
```diff
@@ -63,3 +61,107 @@ object PaidPlan extends ModelWithPublicIdCompanion[PaidPlan] {
...
 object PaidPlanStates extends States[PaidPlan]
+
+sealed abstract class Setting(val value: String) {
+  override def toString = this.value
+}
+@json
+case class PermissionSetting(role: Option[OrganizationRole]) extends Setting(role.map(_.value).getOrElse("none"))
+
+object Setting {
+  implicit val format = Format(__.read[String].map(Setting(_)), new Writes[Setting] { def writes(o: Setting): JsValue = JsString(o.value) })
```
andrewconner: Is just having a `Setting` enough to determine the type of setting (i.e., `PermissionSetting`)? Moving forward, only `OrganizationRole` names work (and the special `"none"`). However, once you have another valid `Setting`, you'd either a) scrap this, b) force any "member" to be parsed as a `PermissionSetting`, but other strings may come out otherwise.

Hope that makes sense.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
```diff
@@ -63,3 +61,107 @@ object PaidPlan extends ModelWithPublicIdCompanion[PaidPlan] {
...
+    def writes(o: FeatureSetting): JsValue = JsObject(o.settingsByFeature.map { case (name, setting) => name.name -> JsString(setting.value) }.toSeq)
+    def reads(json: JsValue): JsResult[FeatureSetting] = {
+      val settingsByFeatureTry = Try(json.as[JsObject].fields.map {
+        case (name, setting) if PermissionsFeatureNames.ALL.contains(Name[PlanFeature](name)) =>
+          if (setting.as[String] == "none") Name[PlanFeature](name) -> PermissionSetting(role = None)
+          else Name[PlanFeature](name) -> PermissionSetting(Some(OrganizationRole(setting.as[String])))
+        case _ => throw new IllegalArgumentException(s"tried to read invalid json=$json as a FeatureSetting")
+      }.toMap)
+      settingsByFeatureTry match {
+        case Failure(errs) => JsError(errs.getStackTrace.toString)
```
andrewconner: Kind of strange to stick a stack trace into a `JsError`.

kifi-backend/modules/common/app/com/keepit/model/OrganizationPermission.scala
camhashemi: `RELOCATE_ORG_LIBRARIES` maybe

kifi-backend/modules/shoebox/test/com/keepit/payments/PaidFeatureSettingsTest.scala
camhashemi: :+1:

kifi-backend/modules/common/app/com/keepit/model/OrganizationPermission.scala
camhashemi: `MOVE_ORG_LIBRARIES`. You shouldn't be able to move it anywhere other than the org of interest.

andrewconner: Perfect

kifi-backend/modules/common/app/com/keepit/model/OrganizationPermission.scala
camhashemi: `PUBLISH_LIBRARIES`

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
camhashemi: I see what you mean about `role`, will rename. `none` is used in `BasePermissions` to key non-member permissions (see `PermissionsMap`), and that was my intention here.

andrewconner: Yep, that's fine. The specialness of `"none"` is a bit strange to me (in `BasePermissions`, it's a `None`)

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
camhashemi: It makes sense. `Setting` should have a `kind` field.

andrewconner: Should it? Currently, `Setting` is the choice. The typing could be just as easily tied to the feature, leaving it so the choice doesn't also have to have a back reference.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/PaymentsController.scala
camhashemi: no, update to this pr on the way

kifi-backend/modules/shoebox/app/com/keepit/payments/PlanManagementCommander.scala
```diff
@@ -492,5 +498,17 @@ class PlanManagementCommanderImpl @Inject() (
     )
   }
 
+  def getPlanFeatureSettings(orgId: Id[Organization]): Map[Name[PlanFeature], FeatureSetting] = {
+    val settingsConfig = db.readOnlyReplica { implicit session => paidAccountRepo.getByOrgId(orgId).settingsByFeature }
+    Map.empty
```
stkem: A nice thing to use in Scala if you haven't implemented something is the magic `???` statement (actually a function, I believe). Makes it much easier to find than a dummy return value and if it should make it into prod somehow at least you'll get a nice "Not Implemented" exception instead of possible silent but really strange behavior).
Dito a few lines below.

camhashemi: thank you

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
```diff
@@ -263,21 +264,6 @@ class OrganizationCommanderImpl @Inject() (
       }
     }
   }
-  private def applyNewBasePermissionsToMembers(memberships: Set[OrganizationMembership], oldBasePermissions: BasePermissions, newBasePermissions: BasePermissions)(implicit session: RWSession): Unit = {
```
camhashemi, line 266: Needed to move `applyNewBasePermissionsToMembers` to `PlanManagementCommander` to avoid a circular dependency with `OrganizationMembershipCommander`. Thinking about creating a `PermissionsCommander` to deal with all the permissions logic.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/PaymentsController.scala
```diff
@@ -68,9 +68,6 @@ class PaymentsController @Inject() (
     Ok(Json.toJson(planCommander.getSimpleContactInfos(request.orgId)))
   }
 
-  @json
-  case class SimpleAccountContactSettingRequest(id: ExternalId[User], enabled: Boolean)
```
camhashemi, line 72: moved all payment request/response classes to `PaymentRequest`

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
```diff
@@ -63,3 +62,188 @@ object PaidPlan extends ModelWithPublicIdCompanion[PaidPlan] {
...
+    val permissionsByRole = namesAndSettings.collect { case (name, setting: PermissionSetting) => setting.role -> PlanFeature.toPermission(name) }
+      .groupBy(_._1).map { case (role, rolesAndPermissions) => role -> rolesAndPermissions.map { _._2 } }
+    BasePermissions(permissionsByRole)
+  }
+
+  def toPermissionsByRole(featureSettings: Set[FeatureSetting]): Map[Option[OrganizationRole], Set[OrganizationPermission]] = {
+    val permissionFeatureSettings = featureSettings.collect { case featureSetting if featureSetting.isPermissionFeature => PermissionFeatureSetting(featureSetting.name.name, featureSetting.setting.value) }
+
+    permissionFeatureSettings.flatMap {
+      case PermissionFeatureSetting(name, PermissionSetting(None)) if PermissionsFeatureNames.CASCADING.contains(name) => {
```
camhashemi: apply the current `OrganizationPermission` to all superior `OrganizationRole`s only if specified by `PermissionsFeatureNames.CASCADING`

kifi-backend/modules/shoebox/app/com/keepit/payments/PaymentRequest.scala
```diff
@@ -0,0 +1,33 @@
+package com.keepit.payments
+
+import com.keepit.common.db.ExternalId
+import com.keepit.model.{ User, Name }
+import com.kifi.macros.json
+import play.api.libs.json.{ JsValue, Json, JsObject, Writes }
+
+sealed abstract class PaymentRequest
+
+case class AccountFeatureSettingsView(settingAndEnabledByName: Map[Name[PlanFeature], (Setting, Boolean)])
```
camhashemi: @andrewconner this is response object that would populate the "org feature settings" modal tab

camhashemi:
@stkem @andrewconner @gratimax @ryanpbrewster updated. implements the APIs for getting/setting `PaidAccount.featureSettings` with permissions propagation.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
```diff
@@ -63,3 +62,188 @@ object PaidPlan extends ModelWithPublicIdCompanion[PaidPlan] {
...
+      }
+    }
+
+    def writes(o: PermissionSetting): JsValue = JsString(o.role.map(_.value).getOrElse("none"))
+  }
+
+  def apply(value: String): PermissionSetting = {
+    value match {
+      case value if Set("member", "admin").contains(value) => PermissionSetting(Some(OrganizationRole(value)))
+      case "none" => PermissionSetting(None)
```
camhashemi: @andrewconner `BasePermissions` (really a wrapper around `PermissionsMap`) does use a magic `"none"`. We can change that, but until then I think it'd be easier to keep the differences low.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidAccountRepo.scala
```diff
@@ -39,9 +44,10 @@ class PaidAccountRepoImpl @Inject() (
     def emailContacts = column[Seq[EmailAddress]]("email_contacts", O.NotNull)
     def lockedForProcessing = column[Boolean]("locked_for_processing", O.NotNull)
     def frozen = column[Boolean]("frozen", O.NotNull)
+    def featureSettings = column[Set[FeatureSetting]]("feature_settings", O.NotNull)
     def modifiedSinceLastIntegrityCheck = column[Boolean]("modified_since_last_integrity_check", O.NotNull)
     def activeUsers = column[Int]("active_users", O.NotNull)
-    def * = (id.?, createdAt, updatedAt, state, orgId, planId, credit, userContacts, emailContacts, lockedForProcessing, frozen, modifiedSinceLastIntegrityCheck, activeUsers) <> ((PaidAccount.apply _).tupled, PaidAccount.unapply _)
+    def * = (id.?, createdAt, updatedAt, state, orgId, planId, credit, userContacts, emailContacts, lockedForProcessing, frozen, featureSettings, modifiedSinceLastIntegrityCheck, activeUsers) <> ((PaidAccount.apply _).tupled, PaidAccount.unapply _)
```
andrewconner: Why added to the middle of the case class, but end of the table? Will make interpolated queries tricker.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
```diff
@@ -63,3 +62,188 @@ object PaidPlan extends ModelWithPublicIdCompanion[PaidPlan] {
...
+      case _ => throw new Exception(s"tried calling PermissionSetting.apply with invalid value=$value")
+    }
+  }
+}
+
+trait FeatureSetting {
+  val name: Name[PlanFeature]
+  val setting: Setting
+
+  def isPermissionFeature = PermissionsFeatureNames.ALL.contains(this.name)
```
andrewconner: I see a lot of these `isPermissionFeature`... is this needed? Usually in Scala you just patten match on the type:

```
someFeature match {
  case permissionFeature: PermissionFeature =>
  case _ => ???
}
```

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
```diff
@@ -63,3 +62,188 @@ object PaidPlan extends ModelWithPublicIdCompanion[PaidPlan] {
...
+}
+object FeatureSetting {
+
+  implicit val format = new Format[FeatureSetting] {
+    def reads(json: JsValue): JsResult[FeatureSetting] = {
+      ((json \ "name").as[String], (json \ "setting").as[String]) match {
+        case (name, setting) if PermissionsFeatureNames.ALL contains Name[PlanFeature](name) => JsSuccess(PermissionFeatureSetting(name, setting)) // I don't need Name[PlanFeature] because of type erasure, right?
```
andrewconner: What's the comment about?

camhashemi: I'm wondering if `contains Name(name)` here is equivalent to `contains Name[PlanFeature](name)`

kifi-backend/modules/shoebox/app/com/keepit/payments/PlanManagementCommander.scala
```diff
@@ -492,5 +503,78 @@ class PlanManagementCommanderImpl @Inject() (
...
+  def getAccountFeatureSettings(orgId: Id[Organization]): AccountFeatureSettingsView = {
+    db.readOnlyReplica { implicit session =>
+      val account = paidAccountRepo.getByOrgId(orgId)
+      val plan = paidPlanRepo.get(account.planId)
+      AccountFeatureSettingsView(plan.features, account.featureSettings)
+    }
+  }
+
+  def setAccountFeatureSettings(orgId: Id[Organization], settings: Set[FeatureSetting]): AccountFeatureSettingsView = {
+    val oldAccount = db.readOnlyReplica { implicit session => paidAccountRepo.getByOrgId(orgId) }
```
andrewconner: Wouldn't you want transactionality for this whole thing?

camhashemi: yup, thanks.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
```diff
@@ -63,3 +62,188 @@ object PaidPlan extends ModelWithPublicIdCompanion[PaidPlan] {
...
+  val name: Name[PlanFeature]
+  val setting: Setting
+
+  def isPermissionFeature = PermissionsFeatureNames.ALL.contains(this.name)
+}
+object FeatureSetting {
+
+  implicit val format = new Format[FeatureSetting] {
+    def reads(json: JsValue): JsResult[FeatureSetting] = {
+      ((json \ "name").as[String], (json \ "setting").as[String]) match {
```
andrewconner: You can parse directly as `Name[PlanFeature]` as long as you have a format for that around.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
```diff
@@ -63,3 +62,188 @@ object PaidPlan extends ModelWithPublicIdCompanion[PaidPlan] {
...
+
+  def apply(name: String, setting: String): FeatureSetting = {
+    name match {
+      case name if PermissionsFeatureNames.contains(name) => PermissionFeatureSetting(name, setting)
+      case _ => throw new IllegalArgumentException(s"tried to call FeatureSetting.apply with unsupported name=$name")
+    }
+  }
+  def apply(name: Name[PlanFeature], setting: Setting): FeatureSetting = FeatureSetting(name.name, setting.value)
+
+  def isPermissionFeature(featureSetting: FeatureSetting): Boolean = PermissionsFeatureNames.ALL.contains(featureSetting.name)
```
andrewconner: Same as above, is this needed? The type should be `PermissionFeatureSetting`.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
```diff
@@ -63,3 +62,188 @@ object PaidPlan extends ModelWithPublicIdCompanion[PaidPlan] {
...
+
+  def toSettingByName(featureSettings: Set[FeatureSetting]): Map[Name[PlanFeature], Setting] = featureSettings.map { featureSetting => featureSetting.name -> featureSetting.setting }.toMap
+  def fromSettingByName(settingByName: Map[Name[PlanFeature], Setting]): Set[FeatureSetting] = settingByName.map { case (name, setting) => FeatureSetting(name, setting) }.toSet
+
+  def toBasePermissions(featureSettings: Set[FeatureSetting]): BasePermissions = {
```
andrewconner: Why include here? Only permission settings would map to a `BasePermission`, right?

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
```diff
@@ -63,3 +62,188 @@ object PaidPlan extends ModelWithPublicIdCompanion[PaidPlan] {
...
+    val namesAndSettings = FeatureSetting.toSettingByName(featureSettings).toSet
+
+    // collects values of type PermissionSetting in settings, mapping them to a BasePermissions
+    val permissionsByRole = namesAndSettings.collect { case (name, setting: PermissionSetting) => setting.role -> PlanFeature.toPermission(name) }
+      .groupBy(_._1).map { case (role, rolesAndPermissions) => role -> rolesAndPermissions.map { _._2 } }
+    BasePermissions(permissionsByRole)
+  }
+
+  def toPermissionsByRole(featureSettings: Set[FeatureSetting]): Map[Option[OrganizationRole], Set[OrganizationPermission]] = {
```
andrewconner: Same, this seems specific to permission settings?

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
```diff
@@ -63,3 +62,188 @@ object PaidPlan extends ModelWithPublicIdCompanion[PaidPlan] {
...
+        List(Some(OrganizationRole.MEMBER) -> PlanFeature.toPermission(name), Some(OrganizationRole.ADMIN) -> PlanFeature.toPermission(name))
+      }
+      case PermissionFeatureSetting(name, setting) => List(setting.role -> PlanFeature.toPermission(name))
+    }.groupBy(_._1).map { case (role, rolesAndPermissions) => role -> rolesAndPermissions.map(_._2) }
+  }
+}
+
+case class PermissionFeatureSetting(name: Name[PlanFeature], setting: PermissionSetting) extends FeatureSetting
+object PermissionFeatureSetting {
+  def apply(name: String, setting: String): PermissionFeatureSetting = {
```
andrewconner: Why have an apply for this and for `FeatureSetting`? Either you already know it's a permission feature (in which case, you should have the `PermissionSetting`) or you don't (in which case `FeatureSetting.apply` handles it), right?

andrewconner: Ah, I see you're using this apply from `FeatureSetting`. Feels kind of awkward—like it's always double checking based on the name.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
```diff
@@ -63,3 +62,188 @@ object PaidPlan extends ModelWithPublicIdCompanion[PaidPlan] {
...
+    if (PermissionsFeatureNames.ALL.contains(Name[PlanFeature](name))) {
+      setting match {
+        case none if none.toLowerCase == "none" => PermissionFeatureSetting(Name[PlanFeature](name), PermissionSetting(role = None))
+        case role => PermissionFeatureSetting(Name[PlanFeature](name), PermissionSetting(Some(OrganizationRole(role))))
+      }
+    } else { throw new IllegalArgumentException(s"tried calling PermissionFeatureSetting.apply with unsupported name=$name and setting=$setting") }
+  }
+
+  def format = new Format[PermissionFeatureSetting] {
```
andrewconner: Do you need a format for this one also? To use it, you'll have to pattern match your `FeatureSetting` to pull out the `PermissionFeatureSetting` values, then call this. Doesn't the `FeatureSetting.format` do the same?

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
```diff
@@ -63,3 +62,188 @@ object PaidPlan extends ModelWithPublicIdCompanion[PaidPlan] {
...
+      .groupBy(_._1).map { case (role, rolesAndPermissions) => role -> rolesAndPermissions.map { _._2 } }
+    BasePermissions(permissionsByRole)
+  }
+
+  def toPermissionsByRole(featureSettings: Set[FeatureSetting]): Map[Option[OrganizationRole], Set[OrganizationPermission]] = {
+    val permissionFeatureSettings = featureSettings.collect { case featureSetting if featureSetting.isPermissionFeature => PermissionFeatureSetting(featureSetting.name.name, featureSetting.setting.value) }
+
+    permissionFeatureSettings.flatMap {
+      case PermissionFeatureSetting(name, PermissionSetting(None)) if PermissionsFeatureNames.CASCADING.contains(name) => {
+        List(None -> PlanFeature.toPermission(name), Some(OrganizationRole.MEMBER) -> PlanFeature.toPermission(name), Some(OrganizationRole.ADMIN) -> PlanFeature.toPermission(name))
```
andrewconner: If the feature names were typed to enforce that they're "permission" features, I think you could avoid having to hope that `PlanFeature.toPermission` would work (as-is, the compiler isn't really helping you to ensure there's a valid permission mapping there)

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
ryanpbrewster, line 266: That feels super weird to me.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
maxov: You're right that at runtime (everything is made up and the) types don't matter, but the compiler would still try to infer `Name[PlanFeature](name)`  if you leave it off (Scala doesn't leave anything to runtime short of asInstanceOf, isInstanceOf or pattern matching from what I know)

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/PaymentsController.scala
```diff
@@ -84,23 +81,17 @@ class PaymentsController @Inject() (
...
   }
 
-  @json
-  case class SimplePlanFeaureSettingRequest(name: String, enabled: Boolean)
-
-  def setPlanFeatureSettings(pubId: PublicId[Organization]) = OrganizationUserAction(pubId, PLAN_MANAGEMENT_PERMISSION)(parse.tolerantJson) { request => //ZZZ TODO: This is currently a dummy (just does request format validation)
-    request.body.validate[Seq[SimplePlanFeaureSettingRequest]] match {
-      case JsSuccess(settings, _) => Ok
+  def setAccountFeatureSettings(pubId: PublicId[Organization]) = OrganizationUserAction(pubId, PLAN_MANAGEMENT_PERMISSION)(parse.tolerantJson) { request => //ZZZ TODO: This is currently a dummy (just does request format validation)
+    request.body.validate[SimpleAccountFeatureSettingRequest] match {
```
maxov: use `.fold` defined on JsResult instead of pattern matching

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidAccountRepo.scala
camhashemi: I believe it's added to the middle of both

kifi-backend/modules/shoebox/app/com/keepit/payments/PaymentRequest.scala
```diff
@@ -0,0 +1,49 @@
...
+
+sealed abstract class PaymentRequest
+
+case class AccountFeatureSettingsRequest(featureSettings: Set[FeatureSetting])
+object AccountFeatureSettingsRequest {
+  implicit val reads = new Reads[AccountFeatureSettingsRequest] {
+    def reads(json: JsValue): JsResult[AccountFeatureSettingsRequest] = {
+      json.validate[JsObject].map { obj =>
+        val featureSettings = obj.fields.map {
+          case (key, value) if Feature.get(key).isDefined && Feature.get(key).get.verify(value.as[String]) => FeatureSetting(key, value.as[String])
```
camhashemi, line 20: @andrewconner

camhashemi:
@andrewconner @stkem @ryanpbrewster this is in a better place now. tests that check if permissions can be modified and enforced properly are passing. 

the model is much cleaner, thanks to @andrewconner. still need to add tests for the remaining features, but I'd appreciate a review of the model in `PaidPlan` and the permissions propagation in `PlanManagementCommander`. Everything else will work off that.

andrewconner:
Nice, will go through tomorrow morning. Thanks for your persistence :)

kifi-backend/modules/shoebox/app/com/keepit/payments/PaymentRequest.scala
```diff
@@ -0,0 +1,60 @@
...
+import com.keepit.common.db.ExternalId
+import com.keepit.model.{ User }
+import com.kifi.macros.json
+import play.api.libs.json._
+import play.api.mvc.Results.Status
+import play.api.http.Status._
+
+import scala.util.control.NoStackTrace
+
+sealed abstract class PaymentRequest
```
andrewconner, line 12: What's this guy for?

kifi-backend/modules/shoebox/app/com/keepit/payments/PaymentRequest.scala
andrewconner, line 20: Here's an idea:

```
for {
  (key, JsString(value) <- obj.fields
  feature <- Feature.get(key) if feature.verify(value)
} yield FeatureSetting(key, value)
```

(and you can `getOrElse` for the `JsError` case)

kifi-backend/modules/shoebox/app/com/keepit/payments/PaymentRequest.scala
```diff
@@ -0,0 +1,60 @@
...
+
+case class AccountFeatureSettingsRequest(featureSettings: Set[FeatureSetting])
+object AccountFeatureSettingsRequest {
+  implicit val reads = new Reads[AccountFeatureSettingsRequest] {
+    def reads(json: JsValue): JsResult[AccountFeatureSettingsRequest] = {
+      json.validate[JsObject].map { obj =>
+        val featureSettings = obj.fields.map {
```
andrewconner: Looks like `JsObject`s have `.fieldSet`. Super minor, but I just noticed that it exists.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaymentRequest.scala
camhashemi, line 12: I was going to have all `Request` case classes in this file inherit from it, not sure what gains I'd get from that right away.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaymentRequest.scala
camhashemi: That returns a `scala.collection.Set` (compiler doesn't like that). Not sure if there's a conversion to `scala.collection.immutable.Set` without calling `asInstanceOf` or `toSet` like I already have.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaymentRequest.scala
camhashemi, line 12: Besides opening the door for reusable behavior between request-wrapper classes, having a superclass makes it easier to look up the `PaymentRequest` file with all of the requests in it, versus needing to remember the name of a single subclass. Not sure if that's a strong enough reason for inheritance.

kifi-backend/modules/shoebox/marketing
```diff
@@ -1 +1 @@
-Subproject commit cdb888e7ef4b056a5b48323620e3f6150533bddf
+Subproject commit 7e80bd2b869033d9f311eb5e0cc681558e829113
```
camhashemi, line 1: this is the hash `marketing/master` is on

kifi-backend/modules/shoebox/test/com/keepit/model/PaidPlanFactory.scala
```diff
@@ -0,0 +1,47 @@
...
+    Set( // very fragile way to create a plan feature set to ensure that our testing feature set matches existing features
+      PlanFeature(Feature.get("publish_libraries").get.name, Feature.get("publish_libraries").get.options.find(_ == "member").get, editable = true),
+      PlanFeature(Feature.get("invite_members").get.name, Feature.get("invite_members").get.options.find(_ == "member").get, editable = true),
+      PlanFeature(Feature.get("group_messaging").get.name, Feature.get("group_messaging").get.options.find(_ == "member").get, editable = true),
+      PlanFeature(Feature.get("force_edit_libraries").get.name, Feature.get("force_edit_libraries").get.options.find(_ == "disabled").get, editable = true),
+      PlanFeature(Feature.get("view_members").get.name, Feature.get("view_members").get.options.find(_ == "anyone").get, editable = true),
+      PlanFeature(Feature.get("move_org_libraries").get.name, Feature.get("move_org_libraries").get.options.find(_ == "member").get, editable = true),
+      PlanFeature(Feature.get("create_slack_integration").get.name, Feature.get("create_slack_integration").get.options.find(_ == "disabled").get, editable = true)
+    )
```
camhashemi, line 21: @ryanpbrewster @stkem @leogrim curious if you guys have improvements on this testing pattern. We need a basic feature configuration to play with.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
```diff
@@ -54,12 +53,78 @@ object PaidPlan extends ModelWithPublicIdCompanion[PaidPlan] {
...
+sealed trait Feature {
+  def name: String
+  def options: Seq[String]
+  def verify(setting: String) = options.contains(setting)
+}
+object Feature {
+  def get(name: String): Option[Feature] = ALL.find(_.name == name)
+  val ALL: Set[Feature] = Set(PublishLibraries, InviteMembers, GroupMessaging, EditLibrary, ViewMembers, MoveOrganizationLibraries, CreateSlackIntegration)
+}
+sealed abstract class OrganizationPermissionFeature(val permission: OrganizationPermission) extends Feature {
```
leogrim, line 79: This and what follows looks a bit weird, since roles are ordered should we just assume that if a role can do something, anyone role higher can as well? Looks like this is exactly what these maps are repeating. Also lack of typing?

kifi-backend/modules/shoebox/test/com/keepit/model/PaidPlanFactory.scala
ryanpbrewster, line 21: All of these hardcoded strings make my skin crawl. I think it'd also be much easier to understand without them.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
camhashemi, line 79: Yeah, we can do that. It's nice to have things explicit, but I can't imagine any scenario when members can do things admins can't.

Typing is on the way in a refactor. I'd rather implement the permission enforcement in one go, and then a refactor in one go, rather than mixed up.

kifi-backend/modules/shoebox/test/com/keepit/model/PaidPlanFactory.scala
camhashemi, line 21: It will be.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaidPlan.scala
```diff
@@ -54,12 +53,78 @@ object PaidPlan extends ModelWithPublicIdCompanion[PaidPlan] {
...
+}
+case object MoveOrganizationLibraries extends OrganizationPermissionFeature(OrganizationPermission.MOVE_ORG_LIBRARIES) {
+  val roleOptions = Map("disabled" -> Seq.empty, "admin" -> Seq(Some(OrganizationRole.ADMIN)), "member" -> Seq(Some(OrganizationRole.MEMBER), Some(OrganizationRole.ADMIN)))
+}
+case object CreateSlackIntegration extends OrganizationPermissionFeature(OrganizationPermission.CREATE_SLACK_INTEGRATION) {
+  val roleOptions = Map("disabled" -> Seq.empty, "member" -> Seq(Some(OrganizationRole.MEMBER), Some(OrganizationRole.ADMIN)), "admin" -> Seq(Some(OrganizationRole.ADMIN)))
+}
+case object EditOrganization extends OrganizationPermissionFeature(OrganizationPermission.EDIT_ORGANIZATION) {
+  val roleOptions = Map("member" -> Seq(Some(OrganizationRole.MEMBER), Some(OrganizationRole.ADMIN)), "admin" -> Seq(Some(OrganizationRole.ADMIN)))
+}
```
ryanpbrewster, line 117: Just to document the conversation we had:

I think that `roleOptions` does not need to exist. The mapping from `Setting` to `Set[Option[OrganizationRole]]` should exist within the `Setting` companion object.

```
sealed abstract class Setting(val targets: Set[Option[OrganizationRole]])
object Setting {
    case object ANYONE extends Setting(OrganizationRole.allOpts)
    case object MEMBERS extends Setting(OrganiationRole.all.map(Some(_)))
    case object DISABLED extends Setting(Set.empty)
    case object ADMINS extends Setting(Set(Some(OrganizationRole.ADMIN)))
}
```

ryanpbrewster, line 117: I also need to think more about whether these `Features` are anything more than lightweight wrappers around a specific `OrganizationPermission`. If that is the case, maybe that's fine.

camhashemi, line 117: I would probably call it `OrganizationPermissionSetting`, since we may need to account for different types of settings in the future. Also, `Setting` is a wrapper around keys that clients get and send to the server, first and foremost. The `targets` is for us to know what to do given a `Setting`.

ryanpbrewster, line 117: Ah, right. I'm not smart today. Definitely include the `value: String` in there.

PR #17255, opened 2015-09-11
Sort libraries in a more pleasant way

andrewconner:
Want to go ahead and get this one out. Just orders them by last kept to, by anyone. We'll feel it and see if it should be tweaked? @cvializ


cvializ:
:+1: let's feel it

PR #17257, opened 2015-09-11
Bugfix/org privacy

adamjgrant:
https://app.asana.com/0/42964066412484/49726470967934

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/common/directives/orgSelector/orgSelector.js
+++ b/kifi-backend/modules/shoebox/angular/src/common/directives/orgSelector/orgSelector.js
@@ -19,7 +19,6 @@ angular.module('kifi')
         $scope.unsetOrg = function() {
           $scope.libraryProps.selectedOrgId = undefined;
           $scope.space.destination = $scope.me;
-          onChangeSpace();
         };
 
         $scope.setOrg = function(id) { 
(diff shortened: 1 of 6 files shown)


adamjgrant:
@andrewconner @cvializ

kifi-backend/modules/shoebox/angular/src/common/directives/orgSelector/orgSelector.js
```diff
@@ -29,34 +28,41 @@ angular.module('kifi')
...
-          var currIsOrg = $scope.spaceIsOrg($scope.space.current);
-          var destIsOrg = $scope.spaceIsOrg($scope.space.destination);
+        // Store privacy setting when switching between org and user profile
+        var privacyToken = {
+          org: $scope.library.visibility
+        };
 
-          if (currIsOrg !== destIsOrg) {
-            if (currIsOrg) {
+        $scope.$watch('space.destination', function(newValue, oldValue) {
```
adamjgrant, line 42: Fundamental change: Instead of comparing the space where the library currently exists and where it's going, it checks the value of the org selector against its previous value.

The outcome is that we can store the privacy setting of the library when it was assigned to an org (whether or not the data was persisted), and restore it if the user clicks back to the org.

cvializ, line 42: Makes sense to me

cvializ:
:+1:

PR #17259, opened 2015-09-11
Feature/org page

adamjgrant:


adamjgrant:
@cvializ @andrewconner Org creation page. 
screen shot 2015-09-14 at 10 34 45 am

Layout feels a little weird. I wonder if we should just make it one column.

kifi-backend/modules/shoebox/angular/src/route.js
```diff
@@ -117,6 +117,16 @@ angular.module('kifi')
         controller: 'OrgProfileLibrariesCtrl',
         templateUrl: 'orgProfile/orgProfileLibraries.tpl.html'
       })
+      .state('organizations', {
+        url: '/organizations',
```
cvializ: since the url is user-facing, should we call it `/teams` instead?

kifi-backend/modules/shoebox/angular/src/orgProfile/orgProfileCreateCtrl.js
```diff
@@ -0,0 +1,22 @@
...
+  '$scope', 'orgProfileService', 'net', '$location', 'ml',
+  function($scope, orgProfileService, net, $location, ml) {
+    $scope.orgName = '';
+    $scope.createOrg = function() {
+      ml.specs.createOrg = new ml.Spec([
+        new ml.Expect('Org was assigned a handle', function(handle) { return (handle !== undefined); }),
+        new ml.Assert('Created org was returned in 3 seconds or less', 3000)
+      ]);
+
+      net.createOrg({name: $scope.orgName}).then(function(response) {
```
cvializ: The call to `net.createOrg` might be better placed in the `orgProfileService`, that way the controller doesn't have to know about the ML specs.

cvializ:
It looks OK to me as two columns, but maybe with more rows it'd look a bit weirder.

kifi-backend/modules/shoebox/angular/src/orgProfile/orgProfileCreateCtrl.js
adamjgrant: Good point.

adamjgrant:
Going to add a few more things after conv. with Jen...please hold.

kifi-backend/modules/shoebox/angular/src/route.js
```diff
@@ -117,6 +117,16 @@ angular.module('kifi')
...
+        'abstract': true,
+        template: '<ui-view/>'
+      })
+      .state('organizations.new', {
+        url: '/new',
```
andrewconner, line 126: I don't know `ui-router` well... does this mean the URL is `/new`, or this is `/organizations/new`?

cvializ, line 126: `/organizations/new`. since `organizations.new` is a child of the `organizations` state, its `url` property is appended to its parent's.

andrewconner, line 126: Cool, slick.

adamjgrant:
Ready @andrewconner @cvializ

cvializ:
:+1: killer

PR #17263, opened 2015-09-14
use a mobile-custom serialization for organization views

camhashemi:
@andrewconner @ryanpbrewster @jaredpetker @DerekBurch bringing back the mobile-custom org serialization, also sending permissions with a user's org list

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationInfo.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationInfo.scala
@@ -80,9 +80,14 @@ case class OrganizationView(
   membershipInfo: OrganizationMembershipInfo)
 
 object OrganizationView {
-  implicit val writes: Writes[OrganizationView] = new Writes[OrganizationView] {
+  val defaultWrites: Writes[OrganizationView] = new Writes[OrganizationView] {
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationInfo.scala
```diff
@@ -80,10 +80,15 @@ case class OrganizationView(
   membershipInfo: OrganizationMembershipInfo)
 
 object OrganizationView {
-  implicit val writes: Writes[OrganizationView] = new Writes[OrganizationView] {
+  val defaultWrites: Writes[OrganizationView] = new Writes[OrganizationView] {
     def writes(o: OrganizationView) = Json.obj("organization" -> OrganizationInfo.defaultWrites.writes(o.organizationInfo),
       "membership" -> OrganizationMembershipInfo.defaultWrites.writes(o.membershipInfo))
   }
+
+  val mobileWrites: Writes[OrganizationView] = new Writes[OrganizationView] {
```
andrewconner, line 88: Should this just be the normal `Writes`? Easiest time to get everyone to agree on an API is before it's released.

camhashemi, line 88: @adamjgrant @cvializ should weigh in if they have a preference. I believe @andrewconner  wanted `organization` to be just an organization, not dependent on the viewer's `membership` object. I talked with mobile, they prefer this way.

PR #17266, opened 2015-09-14
Match the styling of dropdown buttons to the dropdown anchors

cvializ:
@adamjgrant


kifi-backend/modules/shoebox/angular/src/common/classes.styl
```diff
@@ -38,6 +38,7 @@
     background: transparent
     box-sizing: border-box
     width: 100%
+    font-smoothing: antialiased
```
adamjgrant, line 41: mad skillz

adamjgrant:
:+1:

PR #17267, opened 2015-09-14
Increasing feed timer

adamjgrant:
Increasing the timeout of this a little. We're getting a lot of airbrake reports. I think the occasional 3 seconds of delay happening now and then is reasonable, especially considering that users can have slow connections.

@andrewconner @cvializ


cvializ:
sounds reasonable :+1:

PR #17268, opened 2015-09-15
Changing db constraint for library slugs to account for orgs.

andrewconner:
This is not strictly correct, since it won't protect against personal libs having the same slug. Since we already check in code for this situation, I'm going with it (right now it's making it so you can't have the same slug in personal and org locations). We _could_ do a `TrueOrNull` thing here, but worth it?

@ryanpbrewster @camhashemi

### DIFF
--- a/kifi-backend/modules/common/conf/evolutions/shoebox/393.sql
+++ b/kifi-backend/modules/common/conf/evolutions/shoebox/393.sql
@@ -1,0 +1,15 @@
+# SHOEBOX
+
+# --- !Ups
+
+-- mysql:
+--    ALTER TABLE library DROP KEY `library_owner_id_slug`;
+--    CREATE UNIQUE INDEX library_owner_id_org_id_slug on library(owner_id, organization_id, slug);
+
+ALTER TABLE library DROP CONSTRAINT IF EXISTS library_owner_id_slug;
+CREATE UNIQUE INDEX library_owner_id_org_id_slug on library(owner_id, organization_id, slug);
+
+insert into evolutions (name, description) values('393.sql', 'Changing constraints for library to account for orgs with lib slugs');
+
+
+# --- !Downs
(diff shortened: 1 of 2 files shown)


camhashemi:
seems like there's only a couple places (create and modify) where this would be a problem. but you might want to add a check to `LibraryCommander.unsafeTransferLibrary`. Also, @ryanpbrewster why does `LibraryCommander.modifyLibrary` end up calling `LibraryCommander.unsafeModifyLibrary`?

ryanpbrewster:
`validateModifyRequest` makes sure that a modification is valid (requester has permission, not putting things into inconsistent states, etc).
`unsafeModifyLibrary` does the business logic of applying a modification to a library.

`modifyLibrary` just calls `validate` to make sure the request is valid, then `unsafeModify` to actually perform it. It's a convenience method.

ryanpbrewster:
I'm also not sure that this constraint is compliant with current behavior. Right now, a user can only own one library with a given slug. This has pros and cons.

The con is that since orgs can hold libraries, we run into strange situations where a user can't create a `"general"` library in their org space if they already own a `"general"` library in another space (personal or another org). That seems confusing and sort of unintuitive. They're distinct spaces, so there isn't any sort of slug collision, but we still forbid it.

On the other hand, there's no way to avoid weirdness. Suppose we apply this evolution and change the code to account for that constraint. I am now allowed to create a `"ryan"` library in my personal space and my org space. What happens if I want to transfer my library back into my personal space? Do we forbid it because of the slug collision? What if another user creates a `"ryan"` library and tries to transfer it to me? It will fail depending on the space it exists in?

andrewconner:
This was reported as a bug, and I think it is one; it doesn't actually protect anything, because if I stop being the owner of a library, we run into the problem that you may have a `"ryan"` library already.

Seems like the issue is just when moving the library around; easily solvable by renaming the slug on move if needed? (this is what most consumer operating systems let you do)

andrewconner:
Merging to be in line with production. Happy to write the auto-renaming slug stuff on library move, if that seems reasonable to you @ryanpbrewster

PR #17269, opened 2015-09-15
replacing mobile recommendation stream with keepstream with

eishay:


eishay:
DONT MERGE YET. 
I'll do it and test it live with old mobile clients.

PR #17271, opened 2015-09-15
Active menu item state for org profile header

adamjgrant:
@andrewconner @cvializ

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/orgProfile/orgProfileHeader.js
+++ b/kifi-backend/modules/shoebox/angular/src/orgProfile/orgProfileHeader.js
@@ -14,7 +14,6 @@ angular.module('kifi')
     },
     templateUrl: 'orgProfile/orgProfileHeader.tpl.html',
     link: function (scope) {
-      scope.activetab = $state.current.activetab;
       scope.state = $state;
 
       var lastSavedInfo = {};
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/shoebox/angular/src/orgProfile/orgProfileHeader.js
```diff
@@ -14,6 +14,8 @@ angular.module('kifi')
     },
     templateUrl: 'orgProfile/orgProfileHeader.tpl.html',
     link: function (scope) {
+      scope.state = $state;
```
cvializ, line 17: something about exposing the full `$state` object to the template rubs me the wrong way

adamjgrant, line 17: I had originally tried just pulling out `$state.current.activetab` but it wouldn't dynamically update. It sort of treated it as a string literal.

adamjgrant, line 17: I'll try just assigning `$state.current`

adamjgrant, line 17: That didn't work. Angular docs suggest you can use `$route` to watch the route, but I couldn't get this to work, `unknown provider $route` error. Got any ideas?

adamjgrant, line 17: I could also manually assign it on `on $stateChangeSuccess`...

cvializ, line 17: eh I'm looking at docs/SO now and I'm starting to think it's a micro-optimization. ship it

PR #17272, opened 2015-09-15
Payment Processing

stkem:
Hopefully easy to follow. I've tried to keep it compact. Tests & Admin pages (force a charge, enter credit card directly, etc.) coming soon.
@andrewconner @camhashemi @ryanpbrewster @eishay @gratimax


kifi-backend/modules/shoebox/app/com/keepit/payments/AccountLockHelper.scala
```diff
@@ -9,18 +9,31 @@ import com.google.inject.Inject
 
 class AccountLockHelper @Inject() (db: Database, paidAccountRepo: PaidAccountRepo) {
 
-  def maybeSessionWithAccountLock[T](orgId: Id[Organization])(f: RWSession => T): Option[T] = db.readWrite { implicit session =>
+  def aquireAccountLockForSession(orgId: Id[Organization], session: RWSession, attempts: Int = 1): Boolean = {
```
camhashemi: `acquire`

kifi-backend/modules/shoebox/app/com/keepit/payments/PaymentProcessingCommander.scala
```diff
@@ -0,0 +1,116 @@
...
+  paidAccountRepo: PaidAccountRepo,
+  paidPlanRepo: PaidPlanRepo,
+  clock: Clock,
+  accountLockHelper: AccountLockHelper,
+  stripeClient: StripeClient,
+  airbrake: AirbrakeNotifier,
+  implicit val defaultContext: ExecutionContext)
+    extends PaymentProcessingCommander with Logging {
+
+  private val MAX_BALANCE = DollarAmount.wholeDollars(-100) //if you are carrying a balance of more than $100 we will charge you immediately
```
camhashemi: seems like it should be `100` vs `-100`

edit: nvm, I see. If you owe us more than `100`, we'll charge you.

stkem: Exactly. Maybe my choice of terms/words is not the best. Open for other suggestions.

kifi-backend/modules/shoebox/app/com/keepit/payments/AccountLockHelper.scala
stkem: Thanks, Fixing.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaymentProcessingCommander.scala
camhashemi: The code makes sense, the comment may be misleading. If it said "if you are carrying a balance of less than -$100, ...", it would be more aligned to the implementation.

stkem: You are right. I'll make that clearer.

maxov: In addition, I think it would be clearer to call this MIN_IMMEDIATE_CHARGE_BALANCE and the other variable MAX_IMMEDIATE_CHARGE_BALANCE, something along those lines. I know what you were going for in terms of calling this MAX_BALANCE, but I think it's more confusing than the alternative.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaymentProcessingCommander.scala
```diff
@@ -0,0 +1,116 @@
...
+              whoDunnitExtra = JsNull,
+              kifiAdminInvolved = None,
+              action = AccountEventAction.PlanBillingCharge(),
+              creditChange = amount,
+              paymentMethod = pm.id,
+              paymentCharge = Some(amount),
+              memo = None,
+              chargeId = Some(chargeId)
+            ))
+            log.info(s"[PPC][${account.orgId}] Processed charge fro amount $amount: $description")
```
andrewconner: "for" :)

andrewconner:
Looks good. Mind including summaries of what's new/changed?

Would love some tests!

stkem:
I've added a bunch of _very_ basic admin functionality to view and change the plan of an org, add or replace their credit card on file and force account processing. The full loop now works in dev mode with stripe test mode (i.e. you'll actually see charges appear in the stripe dashboard).
I'll merge this shortly so people can start playing with it in both dev and prod (which also uses stripe test more for now, so low risk). Scenario unit tests coming in the next PR.

PR #17275, opened 2015-09-15
Only showing upload new signage if editable

adamjgrant:
Showing "upload new" on avatar only for users with edit permissions.


adamjgrant:
@cvializ

cvializ:
:+1:!

PR #17277, opened 2015-09-15
refactor org messaging event properties

camhashemi:
@andrewconner @joshm1 just being more explicit about what `orgId` refers to, to avoid collisions and confusion


andrewconner:
Sure, whatever Jen wants :)

PR #17278, opened 2015-09-15
Update the title when the user visits the stream.

cvializ:
@adamjgrant @andrewconner 

It wasn't updating when moving from an org profile or user profile, so it showed the feed as having a user profile title.


cvializ:
Going to do this in `feed/feedCtrl.js` instead

PR #17279, opened 2015-09-15
Update the title when the user visits the stream.

cvializ:
@adamjgrant @andrewconner


cvializ:
I'm failing at pushing this change

PR #17280, opened 2015-09-15
Bugfix/stream title

cvializ:
@adamjgrant @andrewconner 

Sorry for the spam. This one also moves the new user checklist and goodies out of `src/recos` to banish the recos folder for good.


adamjgrant:
Banish it! :+1:

PR #17292, opened 2015-09-16
propagate org member permissions to library permissions on org libraries

camhashemi:
@andrewconner @ryanpbrewster @gratimax

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/admin/AdminOrganizationController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/admin/AdminOrganizationController.scala
@@ -461,15 +461,15 @@ class AdminOrganizationController @Inject() (
   }
 
   def addPermissionToAllOrganizations() = AdminUserAction(parse.tolerantJson) { implicit request =>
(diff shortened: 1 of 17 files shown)


kifi-backend/modules/shoebox/app/com/keepit/admin/AdminOrganizationController.scala
```diff
@@ -461,15 +461,15 @@ class AdminOrganizationController @Inject() (
   }
 
   def addPermissionToAllOrganizations() = AdminUserAction(parse.tolerantJson) { implicit request =>
-    implicit val reads = KeyFormat.key2Reads[OrganizationPermission, String]("permission", "confirmation")
-    val (newPermission, confirmation) = request.body.as[(OrganizationPermission, String)]
+    implicit val reads = KeyFormat.key2Reads[PermissionsDiff, String]("permissions", "confirmation")
+    val (permissionsDiff, confirmation) = request.body.as[(PermissionsDiff, String)]
```
camhashemi, line 465: @ryanpbrewster in case we ever want to add non-member permissions (now) or remove a permission

PR #17293, opened 2015-09-16
Don't open the invite/create lib modal for users without permission

cvializ:
@andrewconner @adamjgrant opps


cvializ:
@camhashemi

PR #17297, opened 2015-09-16
Adding instructions for micrologger

adamjgrant:
@cvializ @andrewconner


adamjgrant:
Merging, as this is not really a code commit anyway.

PR #17303, opened 2015-09-17
Bugfix/keep to library widget

adamjgrant:


adamjgrant:
@andrewconner @cvializ This should fix it...surprisingly. The "create" pane was visible for a second or two when the calculations were being made. By using `ng-if` instead, its removal is performed in the DOM rather than in CSS. Inadvertently, this fixes the race condition.

andrewconner:
Nice!

cvializ:
Huh, neat

PR #17320, opened 2015-09-17
[WIP] Upgrades to tooltip css rules to allow markup in tooltips

cvializ:
@adamjgrant @andrewconner WIP for upsell tooltip


adamjgrant:
Looking good. I would just double check implementation of the tooltip on privacy indicators for library cards isn't affected by this change.

cvializ:
Merged this will another PR.

PR #17328, opened 2015-09-18
amplitude - modify user viewed page/pane events

joshm1:
@stkem 

satisfies requirements outlined in this sheet that combine `{user,visitor}_viewed_{page,pane}` events with specific rules regarding particular properties

https://docs.google.com/spreadsheets/d/1uojxmtr59XRqHQNmBiNo-l2QrjjLwXdxXuYGE9AszYI/edit#gid=1897077350

### DIFF
--- a/kifi-backend/modules/heimdal/app/com/keepit/heimdal/AmplitudeClient.scala
+++ b/kifi-backend/modules/heimdal/app/com/keepit/heimdal/AmplitudeClient.scala
@@ -26,6 +26,12 @@ object AmplitudeClient {
   val skipEventFilters: Seq[AmplitudeEventBuilder[_] => Boolean] = Seq(
     (eb: AmplitudeEventBuilder[_]) => eb.eventType.startsWith("anonymous_"),
     (eb: AmplitudeEventBuilder[_]) => eb.heimdalContext.get[String]("userAgent").exists(_.startsWith("Pingdom")),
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/heimdal/app/com/keepit/heimdal/AmplitudeClient.scala
```diff
@@ -222,6 +230,14 @@ class AmplitudeEventBuilder[E <: HeimdalEvent](val event: E)(implicit companion:
     })
   }
 
+
+  override def getUserAndEventProperties(): (Map[String, ContextData], Map[String, ContextData]) = {
+    super.getUserAndEventProperties() match {
+      case (userProperties, eventProperties) =>
```
stkem: Nothing wrong with this, but I think it's a bit odd to use a match statement to unpack a tuple. You could just do 
`val (userProperties, eventProperties) = super.getUserAndEventProperties()`
But maybe thats just me.

stkem:
lgtm

kifi-backend/modules/heimdal/app/com/keepit/heimdal/AmplitudeClient.scala
joshm1: you're right. i don't usually do that so i'm not sure what inspired it at the time :)

PR #17331, opened 2015-09-18
counting four possible visibility levels levels of keeps instead of only two and removing the ALL_KEEPS_VIEW experiment

eishay:
There's a question of wether its right to count "org visible" as "public" keeps, but we can have it in a secondary PR since this PR is not changing the current status.


eishay:
@andrewconner

PR #17336, opened 2015-09-18
stop gap assertion error fix

stkem:
this was what was preventing the removal of users. Integrity checker/backfiller needs to run first before this can be reenabled.
@eishay @andrewconner @gratimax @camhashemi


maxov:
:+1:

PR #17339, opened 2015-09-18
Removing keep image from the site

andrewconner:
@cvializ Requested by a user. Seems like a reasonable request.


cvializ:
:+1: seems legit

andrewconner:
I realized the title of this PR is misleading. This adds the ability to remove a keep image from the site.

PR #17344, opened 2015-09-19
add BasicOrganizationView, giving membership info to org objects

camhashemi:
@ryanpbrewster @andrewconner we initially talked about keeping the `membership` independent of the `organization` object, but mobile and @cvializ are asking for this now.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationInfo.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationInfo.scala
@@ -111,18 +111,18 @@ object BasicOrganizationView {
   }
 }
 
-case class FullOrganizationView(
+case class OrganizationView(
   organizationInfo: OrganizationInfo,
   membershipInfo: OrganizationMembershipInfo)
 
-object FullOrganizationView {
-  val defaultWrites: Writes[FullOrganizationView] = new Writes[FullOrganizationView] {
-    def writes(o: FullOrganizationView) = Json.obj("organization" -> OrganizationInfo.defaultWrites.writes(o.organizationInfo),
+object OrganizationView {
+  val defaultWrites: Writes[OrganizationView] = new Writes[OrganizationView] {
+    def writes(o: OrganizationView) = Json.obj("organization" -> OrganizationInfo.defaultWrites.writes(o.organizationInfo),
       "membership" -> OrganizationMembershipInfo.defaultWrites.writes(o.membershipInfo))
   }
 
-  val mobileWrites: Writes[FullOrganizationView] = new Writes[FullOrganizationView] {
-    def writes(o: FullOrganizationView) = OrganizationInfo.defaultWrites.writes(o.organizationInfo).as[JsObject] ++
+  val mobileWrites: Writes[OrganizationView] = new Writes[OrganizationView] {
+    def writes(o: OrganizationView) = OrganizationInfo.defaultWrites.writes(o.organizationInfo).as[JsObject] ++
       Json.obj("membership" -> OrganizationMembershipInfo.defaultWrites.writes(o.membershipInfo).as[JsObject])
   }
 }
--- a/kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationInfo.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationInfo.scala
@@ -9,7 +9,6 @@ import org.joda.time.DateTime
 import play.api.libs.functional.syntax._
 import play.api.libs.json._
 
-// OrganizationView should ONLY contain public information. No internal ids.
 case class OrganizationInfo(
     orgId: PublicId[Organization],
     ownerId: ExternalId[User],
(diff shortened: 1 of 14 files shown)


cvializ:
bump @andrewconner @ryanpbrewster

andrewconner:
Good with me.

camhashemi:
@cvializ do we need web-handling for this? will this break things?

cvializ:
I see that you removed orgMembership from the library object, but nothing was using that property anyway. I think it's good, unless I missed another property being removed

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationInfo.scala
```diff
@@ -88,18 +87,42 @@ object OrganizationInviteInfo {
...
-case class OrganizationView(
+case class BasicOrganizationView(
+  basicOrganization: BasicOrganization,
+  membershipInfo: OrganizationMembershipInfo)
+object BasicOrganizationView {
+  implicit val defaultWrites: Writes[BasicOrganizationView] = new Writes[BasicOrganizationView] {
+    def writes(o: BasicOrganizationView) = Json.toJson(o.basicOrganization).as[JsObject] + ("membership" -> Json.toJson(o.membershipInfo))
+  }
+
+  val testReads = new Reads[BasicOrganizationView] {
```
ryanpbrewster: This is just to make tests easier? Do you think it actually makes them easier? I've thought a bit about whether it actually makes sense to test our serialization by making sure that our deserialization works. At this point I actually prefer to directly test the serialized output against what we expect (field by field, not checking the entire JSON blob).

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationInfo.scala
```diff
@@ -88,18 +87,42 @@ object OrganizationInviteInfo {
...
+      val handle = (json \ "handle").as[OrganizationHandle]
+      val name = (json \ "name").as[String]
+      val description = (json \ "description").as[Option[String]]
+      val avatarPath = (json \ "avatarPath").as[ImagePath]
+      val basicOrg = BasicOrganization(id, ownerId, handle, name, description, avatarPath)
+      val membershipInfo = (json \ "membership").as[OrganizationMembershipInfo](OrganizationMembershipInfo.testReads)
+      JsSuccess(BasicOrganizationView(basicOrg, membershipInfo))
+    }
+  }
+}
```
ryanpbrewster, line 102: I think there may be a clever Play JSON thing that will handle this a bit more cleanly. If I figure it out I'll change this; it's certainly fine for testing (although I still think that deserialization is the wrong way to handle tests).

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationInfo.scala
```diff
@@ -88,18 +87,42 @@ object OrganizationInviteInfo {
...
+
+case class FullOrganizationView(
```
ryanpbrewster: My vote is for `FullOrganizationView` to be named `OrganizationView`

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationInfo.scala
camhashemi: yeah, I do that too, extracting the fields I expect to be there, but this is more robust than that (you can't add untested fields), without the pain of an entire JSON matching.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaymentRequest.scala
```diff
@@ -17,7 +17,7 @@ object AccountFeatureSettingsRequest {
     def reads(json: JsValue): JsResult[AccountFeatureSettingsRequest] = {
       json.validate[JsObject].map { obj =>
         val featureSettings = obj.fieldSet.map {
-          case (key, value) if Feature.get(key).isDefined && Feature.get(key).get.verify(value.as[String]) => FeatureSetting(key, value.as[String])
+          case (key: String, value: JsValue) if Feature.get(key).isDefined && Feature.get(key).get.verify(value.as[String]) => FeatureSetting(key, value.as[String])
```
ryanpbrewster, line 20: Why are we checking to see if `Feature.get(key).isDefined`? I think we should have `FeatureSettings` being a map from `Feature -> Setting` and then we define a `Feature.apply` method that uses a `match` statement. That is the convention we use elsewhere in the codebase.

andrewconner, line 20: Agreed. I think the original PR has a suggestion on how to do it cleanly.

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationInfo.scala
maxov, line 102: I think just `(json).as[BasicOrganization]` and `(json \ "membership").as[OrganizationMembershipInfo]`would work fine, I don't think that play json errors with extra fields

ryanpbrewster, line 102: Correct, and that'd definitely work. I think there's an easy way to make Play serialize a single JsObject into a tuple, though (and therefore into a case class that wraps a tuple). Seems slightly cleaner.

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationInfo.scala
ryanpbrewster: You convinced me. This seems reasonable.

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationInfo.scala
maxov, line 102: There's no Play tuple format that I know of. We define a special tuple format that serializes and deserializes into json arrays.

ryanpbrewster, line 102: You're right, I guess your way is probably the cleanest.

```
val testReads: Reads[BasicOrganizationView] = (
  __.read[BasicOrganization] and
  (__ \ "membership").read[OrganizationMembershipInfo]
)(BasicOrganizationView.apply _)
```

maxov, line 102: :+1:

camhashemi, line 102: nice, that's awesome. thanks!

kifi-backend/modules/shoebox/app/com/keepit/payments/PaymentRequest.scala
camhashemi, line 20: yup. this is all coming soon.

PR #17345, opened 2015-09-19
Amplitude user viewed page migration from mixpanel

joshm1:
@stkem 

migration rules are defined here: https://docs.google.com/spreadsheets/d/1uojxmtr59XRqHQNmBiNo-l2QrjjLwXdxXuYGE9AszYI/edit#gid=1897077350


stkem:
:+1:

PR #17350, opened 2015-09-21
Permission commander

ryanpbrewster:
@camhashemi @leogrim 

@andrewconner this is another example of a "commander" that is entirely composed of methods that take an implicit session. ((Also, this change got rid of a few session-in-session issues.))

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileOrganizationController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileOrganizationController.scala
@@ -5,6 +5,7 @@ import com.keepit.commanders._
 import com.keepit.common.controller._
 import com.keepit.common.crypto.{ PublicId, PublicIdConfiguration }
 import com.keepit.common.db.ExternalId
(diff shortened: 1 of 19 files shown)


kifi-backend/modules/shoebox/app/com/keepit/controllers/mobile/MobileOrganizationController.scala
```diff
@@ -15,12 +16,14 @@ import scala.concurrent.ExecutionContext
...
-    val orgCommander: OrganizationCommander,
-    val orgMembershipCommander: OrganizationMembershipCommander,
-    val orgInviteCommander: OrganizationInviteCommander,
+    orgCommander: OrganizationCommander,
+    orgMembershipCommander: OrganizationMembershipCommander,
+    orgInviteCommander: OrganizationInviteCommander,
     userCommander: UserCommander,
     heimdalContextBuilder: HeimdalContextBuilderFactory,
     val userActionsHelper: UserActionsHelper,
+    val db: Database,
```
ryanpbrewster, line 25: `db` necessary only because `OrganizationAction` needs to open one to check permissions.

PR #17353, opened 2015-09-21
View members permission

camhashemi:
@ryanpbrewster apparently this is the most important place for this to be enforced...

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
@@ -135,7 +135,7 @@ class OrganizationCommanderImpl @Inject() (
     }
     val members = userRepo.getAllUsers(memberIds).values.toSeq
     val membersAsBasicUsers = members.map(BasicUser.fromUser)
-    val memberCount = orgMembershipRepo.countByOrgId(orgId)
+    val memberCount = members.length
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
```diff
@@ -128,10 +129,13 @@ class OrganizationCommanderImpl @Inject() (
     val site = org.site
     val ownerId = userRepo.get(org.ownerId).externalId
 
-    val memberIds = orgMembershipRepo.getSortedMembershipsByOrgId(orgId, Offset(0), Limit(Int.MaxValue)).map(_.userId)
+    val memberIds = {
+      if (viewerPermissions.contains(OrganizationPermission.VIEW_MEMBERS)) orgMembershipRepo.getSortedMembershipsByOrgId(orgId, Offset(0), Limit(Int.MaxValue)).map(_.userId)
```
ryanpbrewster, line 133: Why are we getting `Int.MaxValue` of these memberships? Isn't that super expensive for large orgs?

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
```diff
@@ -128,10 +129,13 @@ class OrganizationCommanderImpl @Inject() (
...
+      else Seq.empty
```
ryanpbrewster, line 134: Can we airbrake here?

PR #17358, opened 2015-09-22
enforce group messaging permissions

camhashemi:
@leogrim @andrewconner @ryanpbrewster 

seems like we don't have request validation on messaging, so I wasn't sure what to do here. Right now just throwing an airbrake, but I can return a 403 + fail message as well.


andrewconner:
Do you mean checking if someone truly can message someone else (ie, is a friend, in an org with the person, etc)?

maxov:
It would probably be easiest to throw an exception when the permission fails https://github.com/kifi/keepit/pull/17358/files#diff-85d03714ddd6e9728953e91c147a8cd7R671 and then catching it when the message action is sent https://github.com/kifi/keepit/blob/master/kifi-backend/modules/eliza/app/com/keepit/eliza/controllers/ext/ExtMessagingController.scala#L65, returning a 403. I know that the extension, at least, shows an 'error' state when sending a new message fails.

camhashemi:
@andrewconner I meant being able to message the entire org at once. I assumed that's what the feature spec is, but I need to ask Jen about messaging other people in your org individually as well.

@gratimax that sounds okay

ryanpbrewster:
This only hits shoebox when the thread is first created, right? lgtm

camhashemi:
@ryanpbrewster right. after the thread is created, it's a normal thread with all the org participants

kifi-backend/modules/eliza/app/com/keepit/eliza/commanders/MessagingCommander.scala
```diff
@@ -677,25 +687,31 @@ class MessagingCommander @Inject() (
...
-      (messageThread, messagesWithBasicUser) <- basicMessageCommander.getThreadMessagesWithBasicUser(userId, thread)
-      threadInfos <- buildThreadInfos(userId, Seq(thread), Some(url))
-    } yield {
-      val actions = userRecipients.map(id => (Left(id), "message")) ++ nonUserRecipients.collect {
-        case NonUserEmailParticipant(address) => (Right(address), "message")
-      }
-      shoebox.addInteractions(userId, actions)
+    val resFut =
+      canSendToOrgs.flatMap { canSend =>
+        if (!canSend) throw new Exception("insufficient_org_permissions")
```
camhashemi, line 696: @ryanpbrewster throw it and catch it. easier than using an `Either` with all of the different maps and complex signatures going on in these handlers.

PR #17359, opened 2015-09-22
send PaidPlan.name instead of PaidPlan.kind to clients

camhashemi:
@stkem @andrewconner thanks for clearing this up


stkem:
:+1:

PR #17366, opened 2015-09-22
Do not have ZooKeeperClient rely on the default execution context

leogrim:
@andrewconner @stkem @eishay

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/common/zookeeper/ZooKeeperClient.scala
+++ b/kifi-backend/modules/common/app/com/keepit/common/zookeeper/ZooKeeperClient.scala
@@ -1,6 +1,5 @@
 package com.keepit.common.zookeeper
 
-import com.google.inject.{ Inject, Singleton, ImplementedBy }
 import com.keepit.common.core.extras
 import com.keepit.common.logging.Logging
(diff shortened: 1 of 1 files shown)


kifi-backend/modules/common/app/com/keepit/common/zookeeper/ZooKeeperClient.scala
```diff
@@ -154,14 +153,14 @@ class ZooKeeperClientImpl(val servers: String, val sessionTimeout: Int,
...
-  zkSession.compareAndSet(null, Await.result(connect(), Duration.Inf))
+  zkSession.compareAndSet(null, connect())
 
   def session[T](f: ZooKeeperSession => T): T = f(zkSession.get)
 
   def close(): Unit = zkSession.get.close()
 }
 
-class ZooKeeperSessionImpl(zkClient: ZooKeeperClientImpl, promise: Promise[Unit]) extends ZooKeeperSession with Logging {
+class ZooKeeperSessionImpl(zkClient: ZooKeeperClientImpl, initialConnection: Promise[Unit]) extends ZooKeeperSession with Logging {
```
leogrim, line 163: mere renaming

PR #17367, opened 2015-09-22
Fix difficult-to-click space an visibility in the extension

cvializ:
@andrewconner https://app.asana.com/0/42816489233169/51693083115961


cvializ:
Merged, because it's a small code change. Any objections are still welcome

PR #17368, opened 2015-09-22
Moved LibraryPermissions into PermissionCommander

ryanpbrewster:
This was like 4x more painful than I thought it would have been.

@camhashemi


kifi-backend/modules/shoebox/app/com/keepit/commanders/PermissionCommander.scala
```diff
@@ -40,5 +41,72 @@ class PermissionCommanderImpl @Inject() (
...
+        val libraryCanBeViewed = lib.visibility match {
+          case LibraryVisibility.PUBLISHED => true
+          case LibraryVisibility.SECRET => false
+          case LibraryVisibility.DISCOVERABLE =>
+            userIdOpt.contains(lib.ownerId)
+          case LibraryVisibility.ORGANIZATION =>
+            userIdOpt.exists(orgMembershipRepo.getByOrgIdAndUserId(lib.organizationId.get, _).isDefined)
+        }
+        val userHasInvite = userIdOpt.exists { userId => libraryInviteRepo.getWithLibraryIdAndUserId(libId, userId).nonEmpty }
+        if (userHasInvite || libraryCanBeViewed) libraryPermissionsByAccess(lib, LibraryAccess.READ_ONLY)
```
camhashemi, line 108: nice :+1:

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryInfoCommander.scala
```diff
@@ -126,6 +127,10 @@ class LibraryInfoCommanderImpl @Inject() (
     libPathCommander.getPathForLibrary(library)
   }
 
+  def createMembershipInfo(mem: LibraryMembership)(implicit session: RSession): LibraryMembershipInfo = {
```
camhashemi, line 130: this should take in an `Option[LibraryMembership]` and return a `LibraryMembershipInfo` with the appropriate fields. Clients need `LibraryMembershipInfo.permissions` whether or not a membership exists.

PR #17369, opened 2015-09-22
Supporting zip files in bookmark uploads

andrewconner:
@JRuff42 @leogrim


andrewconner:
Last commit fixes the guide. @eishay @gratimax  merging

PR #17375, opened 2015-09-23
Feature/settings tab

cvializ:
@andrewconner (@camhashemi ) A big one, I know. Combined Adam's handoff of the settings page with my initial effort with the upsell icons. Then built on top of both


kifi-backend/modules/shoebox/angular/src/common/directives/orgSelector/orgSelector.tpl.html
```diff
@@ -1,16 +1,48 @@
...
     <h1>{{ me|name }}</h1>
+    <!-- show the upsell if the selected space is an org, and not if I'm an org member who has permission to move org libraries. -->
+    <span
+      ng-if="spaceIsOrg(space.destination) && !(space.destination.membership.role === 'member' && space.destination.membership.permissions.indexOf(ORG_PERMISSION.MOVE_ORG_LIBRARIES) !== -1)"
+      kf-upsell
+      position="right"
+      on-click="onClickUpsellRelocate()"
+      on-hover="onHoverUpsellRelocate()"
+    >
+      <span ng-if="space.destination.planName.indexOf('Paid') !== -1">
```
andrewconner: Ooo, I don't think this will work long term. It works because the paid one is called "Test Paid", but I'm sure we'll have plenty that won't.

Raised this with Cam/Ryan, it shouldn't be there. Bad news is I'm about to break this for you. Good news is it won't randomly break later.

cvializ: Yup @camhashemi and I discussed this and temporarily implemented it this way. I wasn't sure of how it would be exposed long term, so I hacked this in.

andrewconner: Wouldn't it be better to check if they have `move_org_libraries` in the settings?

cvializ: No, because `move_org_libraries` is still present even for non-paid orgs. It's just set to a default value, and I need to know whether the plan is paid or not to decide what text to show.

camhashemi: @cvializ this should be `remove_libraries` now (`move_org_libraries` is deprecated)

andrewconner:
Wahoo!

andrewconner:
Broke your "Paid" check (should be "paid" now), and I see Cam's comment about `remove_libraries`. Besides that, merge ho

PR #17384, opened 2015-09-24
Add a GroupIdentifier typeclass

maxov:
@leogrim 

I made GroupIdentifier a typeclass that knows how to serialize/deserialize a value to a string. This way, we can use something like `GroupIdentifier[(Id[User], Id[User])]` without having to write out an implementation.

still a WIP


kifi-backend/modules/common/app/com/keepit/notify/model/GroupIdentifier.scala
```diff
@@ -0,0 +1,57 @@
...
+ * A shortcut to grouping events quickly. If the group identifier function returns Some for a notification kind,
+ * then a new event of that kind will automatically be grouped with the notification with that identifier.
+ *
+ * Typically grouping is more intelligent and requires reading a couple events from the database and deserializing
+ * JSON. For events which can be grouped with other events far earlier, deserializing a whole bunch
+ * of events from the database to find the right group can be expensive. In addition, events like these do not require
+ * advanced grouping behavior and only rely on a few external ids. Therefore, using [[GroupIdentifier]] only requires
+ * a simple WHERE sql clause on the notification table instead of a whole bunch of deserialization.
+ */
+trait GroupIdentifier[A] { self =>
```
leogrim, line 16: Why do you need the self reference here?

kifi-backend/modules/common/app/com/keepit/notify/model/NotificationKind.scala
```diff
@@ -41,14 +25,18 @@ trait NotificationKind[N <: NotificationEvent] {
   def shouldGroupWith(newEvent: N, existingEvents: Set[N]): Boolean
 }
 
+abstract class GroupIdentifierNotificationKind[N <: NotificationEvent, G](implicit val gid: GroupIdentifier[G]) extends NotificationKind[N] {
```
leogrim: `GroupingNotificationKind` for consistency with the other one?

kifi-backend/modules/common/app/com/keepit/notify/model/GroupIdentifier.scala
maxov, line 16: in the implementation of inmap, I reference self within a new instance of GroupIdentifier

leogrim, line 16: Oh right

kifi-backend/modules/eliza/app/com/keepit/eliza/commanders/NotificationCommander.scala
```diff
@@ -214,4 +219,100 @@ class NotificationCommander @Inject() (
     }
   }
 
+  private def shouldGroupWith(event: NotificationEvent, items: Set[NotificationItem]): Boolean = {
```
maxov, line 221: I decided to move these methods from NotificationProcessing to NotificationCommander, didn't make sense with the completeNotification endpoint

leogrim:
@gratimax this looks awesome, could you quickly resolve the merge conflicts and bring it in?

PR #17385, opened 2015-09-24
Removed permissions from admin page

ryanpbrewster:
It's dangerous to modify them right now because of how paid plan features
interact with permissions.

@camhashemi @stkem

### DIFF
--- a/kifi-backend/modules/common/conf/evolutions/shoebox/385.sql
+++ b/kifi-backend/modules/common/conf/evolutions/shoebox/385.sql
@@ -1,9 +1,5 @@
 # SHOEBOX
 
 # --- !Ups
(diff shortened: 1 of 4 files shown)


ryanpbrewster:
@andrewconner I think that this practice of introducing the default paid plan via an evolution is going to continue biting us going forward. Is there a better strategy?

I think that I'd honestly prefer having some special-case logic in `PlanManagementCommander` that checks to see if it exists and if not creates the appropriate thing.

My actually preferred strategy is to have a paid plan be optional, and if it's null then we just fill it in manually in the code. So we have a `planManagementCommander.getPlan(planIdOpt: Option[Id[Plan]]): PaidPlan` that does a `.map(planRepo.get).getOrElse(defaultPlan)`. But I guess at this point that ship has sailed.

PR #17388, opened 2015-09-24
Eliza message look-here segment fixes

andrewconner:
@leogrim 
@ryanpbrewster included is the image file fix


andrewconner:
merging

PR #17398, opened 2015-09-24
[WIP - Don't merge] Makes the user nav tabs look like the org nav tabs for consistency

jaredpetker:
@cvializ @andrewconner Seemed a bit annoying that the tabs look different on profiles vs orgs, maybe this wasn't desired, just dabbling. Seems like the styles should probably be merged at some point if this is desirable.


cvializ:
Good call, I think this is a good idea too

leogrim:
Should we have these tabs back on mobile as well? Have a consistent look for profiles across the service?

jaredpetker:
@cvializ @andrewconner more or less visually done, we should probably check it out, will refactor some styles and stuff if it's something we want, perhaps have to wait for jen to check it out as well

andrewconner:
This is sweet.

jaredpetker:
This is kinda done, except when the screen gets smaller (width-wise) the
tabs stack on top of each other, unsure if we are ok with that behavior for
someone using this on mobile.

andrewconner:
@cvializ Your call re: jared's comment above?

cvializ:
That behavior is OK with me for now

andrewconner:
Onwards!

jaredpetker:
missed this, will merge in the morning when I can test fully as well.

PR #17402, opened 2015-09-24
Org checker

ryanpbrewster:
@leogrim @stkem

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/integrity/OrganizationChecker.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/integrity/OrganizationChecker.scala
@@ -1,0 +1,105 @@
+package com.keepit.integrity
+
+import com.google.inject.{Inject, Singleton}
+import com.keepit.commanders.LibraryCommander
+import com.keepit.common.db.slick.DBSession.RWSession
+import com.keepit.common.db.slick.Database
+import com.keepit.common.db.{Id, SequenceNumber}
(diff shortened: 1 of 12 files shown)


kifi-backend/modules/shoebox/app/com/keepit/integrity/OrganizationChecker.scala
```diff
@@ -0,0 +1,111 @@
...
+      if (zombieInvites.nonEmpty) {
+        airbrake.notify(s"[ORG-STATE-MATCH] Dead org $orgId has zombie invites: ${zombieInvites.map(_.id.get)}")
+        zombieInvites.foreach(orgInviteRepo.deactivate)
+      }
+
+      val zombiePaidAccount = paidAccountRepo.maybeGetByOrgId(org.id.get, excludeStates = Set(PaidAccountStates.INACTIVE))
+      if (zombiePaidAccount.isDefined) {
+        airbrake.notify(s"[ORG-STATE-MATCH] Dead org $orgId has zombie paid account: ${zombiePaidAccount.map(_.id.get)}")
+        planManagementCommander.deactivatePaidAccountForOrganziation(org.id.get, session)
+      }
```
ryanpbrewster, line 83: @stkem do I need to do this, or does the `PaymentsIntegrityChecker` do this better and I'm just getting in the way?

stkem, line 83: You do indeed need to do this.

PR #17410, opened 2015-09-24
Fix lib validation

ryanpbrewster:
@leogrim @cvializ 

I feel like there's a lot of overlap between the `validateCreateRequest` and `validateModifyRequest`, but I couldn't think of a clean way to DRY them. Any suggestions?


ryanpbrewster:
Also, I'd suggest only looking at the middle commit directly. The first commit is just a rename, so it has a huge but uninteresting diff.

PR #17413, opened 2015-09-25
Evernote imports

andrewconner:
@eishay @stkem 

Not promoting it until we QA a bit more (with more than just my notes)

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/model/Keep.scala
+++ b/kifi-backend/modules/common/app/com/keepit/model/Keep.scala
@@ -285,6 +285,8 @@ object KeepSource {
   val kippt = KeepSource("Kippt")
   val pocket = KeepSource("Pocket")
   val instapaper = KeepSource("Instapaper")
+  val evernote = KeepSource("Evernote")
+  val diigo = KeepSource("Diigo")
   val tagImport = KeepSource("tagImport")
   val emailReco = KeepSource("emailReco")
   val userCopied = KeepSource("userCopied")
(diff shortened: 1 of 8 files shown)


eishay:
yey

kifi-backend/modules/common/app/com/keepit/model/Keep.scala
```diff
@@ -292,13 +294,11 @@ object KeepSource {
...
 
-  val valid = Set(keeper, bookmarkImport, site, mobile, email, default, bookmarkFileImport, kippt, pocket, instapaper, emailReco, twitterFileImport)
-
-  val imports = Set(bookmarkImport, kippt, pocket, instapaper, bookmarkFileImport, twitterFileImport)
+  val imports = Set(bookmarkImport, kippt, pocket, instapaper, evernote, diigo, bookmarkFileImport, twitterFileImport)
 
   // Sources that are from users uploading files, bulk actions, inputting URLs, etc.
   // These may be old links
-  val bulk = Set(site, bookmarkImport, kippt, pocket, instapaper, bookmarkFileImport, twitterFileImport, userCopied, unknown)
+  val bulk = imports ++ Set(userCopied, unknown, evernote)
```
stkem: Unless I'm seriously misreading something, `imports` already contains `evernote`. Why add it again?

stkem:
lgtm (no idea what evernotes data format actually looks like though, so I'm taking that part on good faith :-)

kifi-backend/modules/common/app/com/keepit/model/Keep.scala
andrewconner: Mistake :)

PR #17416, opened 2015-09-25
Unit tests for everything that touches credit

stkem:
I’m nearly certain that this covers all code paths that touch credit in any way.

@andrewconner @camhashemi @ryanpbrewster


kifi-backend/modules/shoebox/marketing
```diff
@@ -1 +1 @@
-Subproject commit e0ee78b339c34882042abcaba792056fbe3a99db
+Subproject commit b316722dcaef610ad5240856581929ae22e9e5d4
```
andrewconner: Don't think this should be merged

kifi-backend/modules/shoebox/test/com/keepit/payments/PaymentProcessingTest.scala
```diff
@@ -0,0 +1,362 @@
...
+    val remainingPrice = fullPrice.multiply(new BigDecimal(fraction, MATH_CONTEXT), MATH_CONTEXT).setScale(0, RoundingMode.HALF_DOWN)
+    DollarAmount(remainingPrice.intValueExact)
+  }
+
+  "PlanManagementCommander" should {
+    "correctly adjust credit for added/removed users, special credit and changed plans" in {
+      withDb(modules: _*) { implicit injector =>
+        val commander = inject[PlanManagementCommander]
+        val accountRepo = inject[PaidAccountRepo]
+        val userId = Id[User](1)
```
andrewconner: How about create these records? I suppose that if this works, it means that foreign key constraints aren't checked.

kifi-backend/modules/shoebox/test/com/keepit/payments/PaymentProcessingTest.scala
```diff
@@ -0,0 +1,362 @@
...
+
+          PaidAccountFactory.paidAccount().withPlan(plan.id.get).withOrganization(org.id.get)
+            .withBillingCycleStart(billingCycleStart)
+            .withCredit(initialCredit)
+            .withActiveUsers(3)
+            .saved
+
+        }
+
+        Await.result(commander.processAccount(accountPre), Duration.Inf) should throwA[Exception](message = "missing_default_payment_method")
```
andrewconner, line 356: This exception is appropriately caught in the code that processes every user?

andrewconner:
Nice!

kifi-backend/modules/shoebox/test/com/keepit/payments/PaymentProcessingTest.scala
stkem, line 356: Yes, sort of. The method here (`processAccount`) actually already airbrakes this error and then raises exception. The method one up (i.e. the one that runs through all the accounts) just swallows it silently. What I would like to add there is a summary email of every run (to `billing@kifi.com` or something) with who got charged how much and who failed for what reason. Not a blocker, but nice to have.

kifi-backend/modules/shoebox/marketing
stkem: Ah, yes. This was unintentional. Must have gotten in when I merged in master. Anyone know of an easy way of taking out just this form the PR?

kifi-backend/modules/shoebox/test/com/keepit/payments/PaymentProcessingTest.scala
stkem: The only place the `userId` appears in the code paths tested in this test is in a json blob in `AccountEvent`s, which precludes a foreign key.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
```diff
@@ -256,7 +256,7 @@ class OrganizationCommanderImpl @Inject() (
             val orgTemplate = organizationWithModifications(orgSkeleton, request.initialValues.asOrganizationModifications)
             val org = handleCommander.autoSetOrganizationHandle(orgRepo.save(orgTemplate)) getOrElse (throw OrganizationFail.HANDLE_UNAVAILABLE)
             orgMembershipRepo.save(org.newMembership(userId = request.requesterId, role = OrganizationRole.ADMIN))
-            planManagementCommander.createAndInitializePaidAccountForOrganization(org.id.get, PaidPlan.DEFAULT, request.requesterId, session) //this should get a .get when things are solidified
+            planManagementCommander.createAndInitializePaidAccountForOrganization(org.id.get, PaidPlan.DEFAULT, request.requesterId, session)
```
ryanpbrewster: Is this no longer true? Did you want to put the `.get` in?

kifi-backend/modules/shoebox/marketing
ryanpbrewster: Go into the submodule, checkout `e0ee78b339...`, then back out and commit just that change.

```
cd modules/shoebox/marketing
git checkout e0ee78b339
cd ../../../
git ci -a -m "Fixing marketing submodule"
```

kifi-backend/modules/shoebox/test/com/keepit/payments/PaymentProcessingTest.scala
ryanpbrewster: I'm with Andrew. Even though it probably won't make a difference, it's a bad habit and it may end up confusing new engineers later on. It's almost as quick (and probably less confusing) to just do `val users = db.readWrite { implicit session => UserFactory.users(2).saved }` and then use `users(0).id.get` directly.

kifi-backend/modules/shoebox/test/com/keepit/payments/FakeStripeClient.scala
```diff
@@ -21,11 +23,15 @@ class FakeStripeClientImpl extends StripeClient {
...
   }
 
-  def processCharge(amount: DollarAmount, token: StripeToken, description: String): Future[StripeChargeResult] = Future.successful {
-    val num = chargeCounter.getAndIncrement()
-    val trans = FakeTransaction(s"faketransaction_$num", amount, token, description)
-    transactions(token).append(trans)
-    StripeChargeSuccess(amount, trans.id)
+  def processCharge(amount: DollarAmount, token: StripeToken, description: String): Future[StripeChargeResult] = {
+    if (failingMode) {
+      Future.failed(new Exception("boom"))
```
camhashemi, line 28: error message could be a tad more informative

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
stkem: Oh, odd, merge issue. I have the `.get` locally. I'll make another commit.

kifi-backend/modules/shoebox/marketing
stkem: Sounds like you've done this before? :-)

kifi-backend/modules/shoebox/test/com/keepit/payments/FakeStripeClient.scala
stkem, line 28: Haha, yeah. It's just for a test though.

kifi-backend/modules/shoebox/marketing
ryanpbrewster: :disappointed: a few times.

On the upside, if you `git rebase` instead of `git merge` this is no longer an issue.

kifi-backend/modules/shoebox/test/com/keepit/payments/PaymentProcessingTest.scala
stkem: Fine with me. I'll make the change.

PR #17419, opened 2015-09-25
Factor code out of UserCommander, remove circular dependency, testing

leogrim:
@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/UserEmailAddressCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/UserEmailAddressCommander.scala
@@ -3,7 +3,7 @@ package com.keepit.commanders
 import com.google.inject.{ ImplementedBy, Inject, Singleton }
 import com.keepit.commanders.emails.EmailConfirmationSender
 import com.keepit.common.db.Id
-import com.keepit.common.db.slick.DBSession.{ RSession, RWSession }
+import com.keepit.common.db.slick.DBSession.RWSession
 import com.keepit.common.db.slick.Database
 import com.keepit.common.logging.Logging
 import com.keepit.common.mail.EmailAddress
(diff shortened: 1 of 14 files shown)


maxov:
:+1: :+1: :tada: :balloon:

kifi-backend/modules/shoebox/app/com/keepit/commanders/emails/EmailTemplateProcessor.scala
```diff
@@ -63,7 +64,7 @@ class EmailTemplateProcessorImpl @Inject() (
     libraryRepo: LibraryRepo,
     userRepo: UserRepo,
     keepRepo: KeepRepo,
-    userCommander: Provider[UserCommander],
+    s3ImageStore: S3ImageStore,
```
leogrim, line 67: Victory match point.

kifi-backend/modules/shoebox/app/com/keepit/commanders/UserEmailAddressCommander.scala
```diff
@@ -136,7 +140,7 @@ class UserEmailAddressCommanderImpl @Inject() (db: Database,
 
     if (!force && isLast) Failure(new LastEmailAddressException(emailAddress))
     else if (!force && isLastVerified) Failure(new LastVerifiedEmailAddressException(emailAddress))
-    else if (!force && isPrimaryEmail(emailAddress)) Failure(new PrimaryEmailAddressException(emailAddress))
+    else if (!force && emailAddress.primary) Failure(new PrimaryEmailAddressException(emailAddress))
```
ryanpbrewster, line 143: Can we call this `.isPrimary` instead of `.primary`?

leogrim, line 143: It's called 'primary' in the db for consistency with a field called `verified`, but yeah actually I was about to do that next :)

PR #17420, opened 2015-09-25
Add the 'change team settings' setting.

cvializ:
@andrewconner @camhashemi


kifi-backend/modules/shoebox/angular/src/orgProfile/orgProfileSettingsCtrl.js
```diff
@@ -11,6 +11,19 @@ angular.module('kifi')
 
     $scope.settingsSectionTemplateData = [
       {
+        heading: '',
+        fields: [
+          {
+            title: 'Who can change team settings?',
```
camhashemi, line 17: there's already a tab with the name "settings" (which is very different than this), so probably best to use "edit team profile" or "edit team info". @JRuff42

PR #17423, opened 2015-09-25
Fixed a nasty bug in the permissions checking

ryanpbrewster:
The way that Feature Settings work makes checking permissions more difficult
than it used to be. I want to reiterate that we really need to revisit this
model for org permissions.


camhashemi:
To be fair, this "bug" was caught by tests and never in prod. Not sure what this has to do with Feature Settings, we need to check the permission one way or another.

PR #17426, opened 2015-09-25
add enforcements and tests for group messaging permission

camhashemi:
@andrewconner @ryanpbrewster


andrewconner:
Excellent :+1:

PR #17429, opened 2015-09-25
Add the main property, so JPM runs.

cvializ:
@andrewconner Misleading branch name. I was going to bump the version number but you said you would, so I just added this property instead to make JPM work


cvializ:
Added extra styling code

PR #17432, opened 2015-09-25
Refactor SecureSocial, decouple from SocialUserInfo, deprecate FortyTwo network

leogrim:
@andrewconner @eishay @stkem 
Once this is merged, the `FortyTwo` network will be effectively out, we can remove the corresponding data from the database (i.e. clean up `SocialUserInfo` and `SocialConnection` tables)


andrewconner:
Goodness. Will look at this morning.

kifi-backend/modules/shoebox/app/com/keepit/model/UserCredRepo.scala
```diff
@@ -31,8 +34,34 @@ class UserCredRepoImpl @Inject() (val db: DataBaseComponent, val clock: Clock) e
...
+        val savedCred = save(cred)
+        log.info(s"[internUserPassword] Persisted new $savedCred for user $userId")
+        savedCred
+    }
+  }
+
+  def verifyPassword(userId: Id[User], password: String)(implicit session: RSession): Boolean = {
+    findByUserIdOpt(userId).exists { userCred =>
+      val currentPasswordInfo = PasswordInfo(UserCred.passwordHasher, userCred.credentials)
+      val hasher = Registry.hashers.currentHasher
```
andrewconner: If you want, we could break the dependency on SecureSocial's registry, keep a global static `new BCryptPasswordHasher(Play.current)`.

kifi-backend/modules/shoebox/app/com/keepit/model/UserCredRepo.scala
```diff
@@ -31,8 +34,34 @@ class UserCredRepoImpl @Inject() (val db: DataBaseComponent, val clock: Clock) e
...
+      hasher.matches(currentPasswordInfo, password)
```
andrewconner: I'm sure it used to do this, but this technically shouldn't be in the db session, because—by design—this is slow (can be 50-100ms). Will block a db connection.

kifi-backend/modules/common/app/com/keepit/social/UserIdentity.scala
```diff
@@ -1,29 +1,50 @@
...
+      passwordInfo = passwordInfo
+    )
+    UserIdentity(user.id, socialUser)
+  }
+}
+
+class NewUserIdentity(userId: Option[Id[User]], socialUser: SocialUser) extends MaybeUserIdentity(userId, socialUser)
 
-  def unapply(u: UserIdentity) = Some(u.userId, u.socialUser, u.allowSignup)
+object NewUserIdentity {
```
andrewconner, line 47: This is great

kifi-backend/modules/shoebox/app/com/keepit/commanders/UserCommander.scala
```diff
@@ -338,48 +341,37 @@ class UserCommander @Inject() (
...
-  private def setNewPassword(userId: Id[User], sui: SocialUserInfo, newPassword: String): Identity = {
-    val pwdInfo = Registry.hashers.currentHasher.hash(newPassword)
-    val savedIdentity = UserService.save(UserIdentity(
-      userId = sui.userId,
-      socialUser = sui.credentials.get.copy(passwordInfo = Some(pwdInfo))
-    ))
-    val updatedCred = db.readWrite { implicit session =>
-      userCredRepo.findByUserIdOpt(userId) map { userCred =>
-        userCredRepo.save(userCred.withCredentials(pwdInfo.password))
+  def changePassword(userId: Id[User], newPassword: String, oldPassword: Option[String]): Try[Unit] = Try {
```
andrewconner, line 344: We will want to test this one very well

kifi-backend/modules/shoebox/app/com/keepit/model/UserCredRepo.scala
leogrim: Makes sense, I'll move it out.

kifi-backend/modules/shoebox/app/com/keepit/model/UserCredRepo.scala
```diff
@@ -31,8 +34,34 @@ class UserCredRepoImpl @Inject() (val db: DataBaseComponent, val clock: Clock) e
...
+      case Some(cred) if cred.state == UserCredStates.ACTIVE => {
+        val updatedCred = cred.withCredentials(credentials)
+        if (cred != updatedCred) {
+          val savedCred = save(updatedCred)
+          log.info(s"[internUserPassword] Persisted updated $savedCred for user $userId")
+          savedCred
+        } else cred
+      }
+      case inactiveCredOpt =>
+        val cred = UserCred(id = inactiveCredOpt.flatMap(_.id), userId = userId, credentials = credentials)
```
andrewconner, line 53: Set as active in this case?

kifi-backend/modules/common/app/com/keepit/social/UserIdentity.scala
leogrim, line 47: Could be better but it's a start :)

kifi-backend/modules/shoebox/marketing
```diff
@@ -1 +1 @@
-Subproject commit e0ee78b339c34882042abcaba792056fbe3a99db
+Subproject commit d3dc4581840dbc8234e5ea2cd6da376fd04d1cd7
```
andrewconner, line 1: Shouldn't be here

kifi-backend/modules/shoebox/app/com/keepit/commanders/UserCommander.scala
leogrim, line 344: Much more straightforward then the old one actually, since we only need to check one place.

kifi-backend/modules/shoebox/app/com/keepit/model/UserCredRepo.scala
leogrim, line 53: Policy of treating inactive models as deleted.

andrewconner:
_Seems reasonable_. But this is so hard to test statically. I'll help you QA it when you merge.

Things to test:

• Email user creation
• Change email of user, then register another user with the original email address
• Facebook user creation
• Changing password on site
• Resetting password ("forgot password")

Anything else to add?

kifi-backend/modules/shoebox/app/com/keepit/model/UserCredRepo.scala
andrewconner, line 53: Why would we update this record then?

leogrim, line 53: Because there's a unique constraint on `userId`

andrewconner, line 53: But not make it active?

andrewconner, line 53: This is reusing an existing id, saving new data, but not marking it as active. What case would we want this?

leogrim, line 53: It's a brand new model, `active` by default

andrewconner, line 53: You're right. Thought it was a `.copy`

PR #17436, opened 2015-09-26
Introduce a BlockingFilter

leogrim:
@andrewconner @eishay


kifi-backend/modules/common/app/com/keepit/common/controller/BlockingFilter.scala
```diff
@@ -0,0 +1,18 @@
+package com.keepit.common.controller
+
+import play.api.mvc.{ Results, Result, RequestHeader, Filter }
+
+import scala.concurrent.Future
+
+object BlockingFilter extends Filter {
+  def apply(nextFilter: (RequestHeader) => Future[Result])(requestHeader: RequestHeader): Future[Result] = {
+    val ip = requestHeader.headers.get("X-Forwarded-For").getOrElse(requestHeader.remoteAddress)
+    if (blocked.contains(ip)) Future.successful(Results.Redirect("/logout"))
```
eishay, line 10: maybe do something like this ?

``` scala
if (blocked.contains(ip) && requestHeader.path != "/logout") Future.successful(Results.Redirect("/logout"))
```

leogrim, line 10: @andrewconner already did I believe

PR #17440, opened 2015-09-28
Diigo and Evernote import UI

andrewconner:
@leogrim @eishay

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/common/controller/BlockingFilter.scala
+++ b/kifi-backend/modules/common/app/com/keepit/common/controller/BlockingFilter.scala
@@ -7,7 +7,7 @@ import scala.concurrent.Future
 object BlockingFilter extends Filter {
   def apply(nextFilter: (RequestHeader) => Future[Result])(requestHeader: RequestHeader): Future[Result] = {
     val ip = requestHeader.headers.get("X-Forwarded-For").getOrElse(requestHeader.remoteAddress)
-    if (blocked.contains(ip)) Future.successful(Results.Redirect("/logout"))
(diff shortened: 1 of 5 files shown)


kifi-backend/modules/common/app/com/keepit/common/controller/BlockingFilter.scala
```diff
@@ -7,7 +7,7 @@ import scala.concurrent.Future
 object BlockingFilter extends Filter {
   def apply(nextFilter: (RequestHeader) => Future[Result])(requestHeader: RequestHeader): Future[Result] = {
     val ip = requestHeader.headers.get("X-Forwarded-For").getOrElse(requestHeader.remoteAddress)
-    if (blocked.contains(ip)) Future.successful(Results.Redirect("/logout"))
+    if (blocked.contains(ip) && !requestHeader.path.contains("logout") && requestHeader.session.get(KifiSession.FORTYTWO_USER_ID).isDefined) Future.successful(Results.Redirect("/logout"))
```
leogrim, line 10: :+1:

kifi-backend/modules/shoebox/angular/index.html
```diff
@@ -3,6 +3,7 @@
 <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
 <meta name="viewport" content="width=device-width,initial-scale=1">
 <base href="/">
+<link rel="preconnect" href="https://d1dwdv9wd966qu.cloudfront.net">
```
andrewconner, line 6: @cvializ Trying to see if this makes a difference. Check out the diff for this file. The biggest changes are just pushing analytics further down, and this `preconnect`. Also made cloudfront urls use `https://` instead of `//`, since the site is _only_ delivered over `https`. Seems okay?

cvializ, line 6: yup, looks good :+1:

PR #17441, opened 2015-09-28
Added a bunch of explicit tests for lib permission

ryanpbrewster:
@andrewconner @camhashemi @leogrim

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/model/LibraryMembership.scala
+++ b/kifi-backend/modules/common/app/com/keepit/model/LibraryMembership.scala
@@ -103,6 +103,7 @@ sealed abstract class LibraryPermission(val value: String)
 object LibraryPermission {
   case object VIEW_LIBRARY extends LibraryPermission("view_library")
   case object EDIT_LIBRARY extends LibraryPermission("edit_library")
+  case object PUBLISH_LIBRARY extends LibraryPermission("publish_library")
   case object MOVE_LIBRARY extends LibraryPermission("move_library")
   case object DELETE_LIBRARY extends LibraryPermission("delete_library")
   case object INVITE_FOLLOWERS extends LibraryPermission("invite_followers")
@@ -125,6 +126,7 @@ object LibraryPermission {
     str match {
       case VIEW_LIBRARY.value => VIEW_LIBRARY
       case EDIT_LIBRARY.value => EDIT_LIBRARY
+      case PUBLISH_LIBRARY.value => PUBLISH_LIBRARY
       case MOVE_LIBRARY.value => MOVE_LIBRARY
       case DELETE_LIBRARY.value => DELETE_LIBRARY
       case INVITE_FOLLOWERS.value => INVITE_FOLLOWERS
@@ -137,7 +139,6 @@ object LibraryPermission {
       case REMOVE_OTHER_KEEPS.value => REMOVE_OTHER_KEEPS
       case EXPORT_KEEPS.value => EXPORT_KEEPS
       case CREATE_SLACK_INTEGRATION.value => CREATE_SLACK_INTEGRATION
-      // TODO(ryan): should we have a `case _ => ???` here, and what should ??? be
     }
   }
 }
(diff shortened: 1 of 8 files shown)


kifi-backend/modules/common/app/com/keepit/model/LibraryMembership.scala
```diff
@@ -125,6 +126,7 @@ object LibraryPermission {
     str match {
       case VIEW_LIBRARY.value => VIEW_LIBRARY
       case EDIT_LIBRARY.value => EDIT_LIBRARY
+      case PUBLISH_LIBRARY.value => PUBLISH_LIBRARY
```
camhashemi: not sure that this helps clients. Say that you have a library in your personal space. Having or not having this permission on that library doesn't mean anything, since it depends which org you want to publish in. If you have three orgs, only one of which you can publish to, what does it mean to have `PUBLISH_LIBRARY` in `library.permissions`?

kifi-backend/modules/shoebox/app/com/keepit/commanders/PermissionCommander.scala
```diff
@@ -42,13 +42,38 @@ class PermissionCommanderImpl @Inject() (
...
+  def getLibraryPermissions(libId: Id[Library], userIdOpt: Option[Id[User]])(implicit session: RSession): Set[LibraryPermission] = {
+    val lib = libraryRepo.get(libId)
+    val libMembershipOpt = userIdOpt.flatMap { userId => libraryMembershipRepo.getWithLibraryIdAndUserId(libId, userId) }
+
+    val libAccessOpt = libMembershipOpt match {
+      case Some(libMembership) => Some(libMembership.access)
+      case None =>
+        val viewerHasImplicitAccess = lib.visibility match {
+          case LibraryVisibility.DISCOVERABLE =>
+            userIdOpt.contains(lib.ownerId) // this would be a really bad sign, since they should have a LibraryMembership
```
camhashemi, line 54: we should throw an airbrake here

kifi-backend/modules/common/app/com/keepit/model/LibraryMembership.scala
ryanpbrewster: Hrm, good point. By strict interpretation, it means "you can publish this library in its current space". I'm not sure how to effectively communicate the full range of permissions to clients, though.

leogrim: Maybe this should come down as an org permissions, not a library permission, and checked as such by clients when an org's space is picked?

camhashemi: I believe that's how it's done now. Pretty sure I added `LibraryPermission.PUBLISH_LIBRARY` and after talking to clients, realized they didn't need it as a library permission.

ryanpbrewster: I'm wary of this strategy because it pushes the library permission-checking logic out to the clients. There is already kind of complex interaction between org and library permissions, and I'm sure it's only going to get more complex as time goes on.

ryanpbrewster: And yes, it is already passed down as an org permission. But we also pass down `OrganizationPermission.FORCE_EDIT_LIBRARIES`, and that manifests itself as a `LibraryPermission.EDIT_LIBRARIES` under certain circumstances. I'd rather perform that logic on the backend and send reliable permissions out to the clients.

camhashemi: It "duplicates" permission-checking logic out to clients, which we do for lots of things already. Better to not let a user make an action at all, than to allow them and then return an error (which may or may not be elegantly presented).

Besides, they're going to have to check for `PUBLISH_LIBRARY` on any org that is not the current space anyways, so why not make it consistent for any org? Otherwise, they have two different places to look for permissions (not really, they're going to choose the most consistent one)

PR #17453, opened 2015-09-28
Hide library followers for the default follow libs

ryanpbrewster:
@andrewconner


kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryInfoCommander.scala
```diff
@@ -278,7 +278,7 @@ class LibraryInfoCommanderImpl @Inject() (
         val (collaboratorCount, followerCount, keepCount) = countsByLibraryId(libId)
```
andrewconner, line 280: How about special casing this guy too?

ryanpbrewster, line 280: The counts?

andrewconner, line 280: Yeah, specifically `100911`

andrewconner, line 280: ermmm `followerCount`

PR #17454, opened 2015-09-28
Feature/show org credit

cvializ:
| Before | After |
| --- | --- |
| screen shot 2015-09-28 at 2 40 23 pm | screen shot 2015-09-28 at 2 40 04 pm |


andrewconner:
Wahoo. The after is better.

PR #17456, opened 2015-09-28
Update visibility icons on keep cards and library headers

DerekBurch:
This removes the background colors on library visibility in the library header, renames 'organization private' to 'team' and removes the red circle behind the private lock on library cards.

https://app.asana.com/0/20751731968782/51714360963530

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/libraries/libraryHeader.styl
+++ b/kifi-backend/modules/shoebox/angular/src/libraries/libraryHeader.styl
@@ -387,7 +387,6 @@ followerPicDiam = 2.2em
   background-position: .9em .46em
   background-size: .9em auto
   background-repeat: no-repeat
-  background-color: libraryRed
 
   .kf-system-secret &
     display: none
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/angular/src/libraries/libraryHeader.styl
```diff
@@ -404,14 +403,12 @@ followerPicDiam = 2.2em
     margin-right: 5px
 
 .kf-lh-org-private
-  background: readableYellowColor
 
   &::after
-    content: 'Organization Private'
+    content: 'Team'
```
andrewconner, line 408: If your reaction to this was, "Oh, it's weird to use CSS to choose text". Yes, yes it is.

andrewconner:
-7 lines, a very good commit

kifi-backend/modules/shoebox/angular/src/libraries/libraryHeader.styl
DerekBurch, line 408: Should most strings just be in the template files then? (For when changing things in the future)

andrewconner, line 408: I'd err towards yes... @cvializ probably has a better informed opinion. CSS is just usually the last place I look to change text. It's more readable (for me, at least) just to have it where it happens in the template. But I may be behind the times

PR #17459, opened 2015-09-28
use basic org view instead of basic org for /ext/start

camhashemi:
@cvializ this will return the same org object as you get from `/user/me`, for example


camhashemi:
@andrewconner

andrewconner:
Good with me

PR #17471, opened 2015-09-29
No personas on backend anymore, instead FTUE

andrewconner:
`has_seen_ftue` will be included for new users. Existing users will be `undefined`


andrewconner:
@jaredpetker

PR #17474, opened 2015-09-29
temporarily running payments integrity checker at lot more aggressively

stkem:
to fix some old data
@ryanpbrewster @andrewconner


ryanpbrewster:
Should we maybe make a generic admin endpoint to manually trigger any one of these integrity checks?

stkem:
Yeah, that is probably a good idea. Don't want to have to click this for every org here right not though.

PR #17477, opened 2015-09-29
Add Team color to keeper for things kept into team libraries

DerekBurch:
I believe this handles https://app.asana.com/0/20751731968782/51751162185848. Though I am unsure if my changes to howKept have broken things in places. There also may be other places that team color should show, but I didn't see anything in my quick look.


andrewconner:
Looking through usages of `howKept`. I notice that right now it'll make the address bar icon green for all but private... that's probably the right thing.

andrewconner:
Check out keeper.js:513. Would need to run it to make sure that still works as intended.

andrewconner:
@cvializ fyi

DerekBurch:
Oh that thing on keeper.js:513. It should probably at least remove the 'kifi-organization' style. Though it would be better if we could just clear any of the visibility styles (without listing each one).

DerekBurch:
I'll ask jen about the address bar icon color. I lean towards making it the team color when its kept to an org lib, but she can decide.

cvializ:
:+1: overall, and to the most recent comment

DerekBurch:
I think I fixed all the things mentioned here. I'll merge this in if you guys are fine with the fixes/changes.

packrat/main.js
```diff
@@ -111,11 +111,21 @@ PageData.prototype = {
...
+        'published': 'public',
+        'discoverable': 'public'
+      };
+      var maxVisibilityIndex = -1;
       var mine = keeps.filter(isMine);
-      if (mine.length) {
-        return mine.every(isSecret) ? 'private' : 'public';
-      }
-      return 'other';
+      mine.length && mine.forEach(function (item) {
```
cvializ: good thought, but you don't need the check for `mine.length`, because `mine.forEach` will be a noop if it's empty

cvializ:
:+1: noice

packrat/main.js
DerekBurch: Oh yea good point I was trying to keep the if statement from before, but yea its useless. Will remove it.

PR #17478, opened 2015-09-29
add BasicLibraryDetails.permissions (viewer permissions)

camhashemi:
@leogrim this was the first (hopefully only) thing to fall through the cracks of the `LibraryCardInfo` and `FullLibraryInfo.permissions` refactor. Front-end expects this object to look like other libraries.

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/model/Library.scala
+++ b/kifi-backend/modules/common/app/com/keepit/model/Library.scala
@@ -381,7 +381,8 @@ case class BasicLibraryDetails(
   keepCount: Int,
   membership: Option[LibraryMembershipInfo], // viewer
   ownerId: Id[User],
-  url: Path)
+  url: Path,
+  permissions: Set[LibraryPermission])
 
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/common/app/com/keepit/model/Library.scala
```diff
@@ -381,7 +381,8 @@ case class BasicLibraryDetails(
   keepCount: Int,
   membership: Option[LibraryMembershipInfo], // viewer
   ownerId: Id[User],
-  url: Path)
+  url: Path,
+  permissions: Set[LibraryPermission])
```
camhashemi, line 385: a comment above tells me to ask search before changing @leogrim

andrewconner, line 385: I imagine there is search index or cache invalidation to worry about.

leogrim, line 385: No problem here, you're _adding_ a field so just make sure to deploy Shoebox first.

PR #17479, opened 2015-09-29
fifty bucks for each new org

stkem:
@ryanpbrewster @andrewconner


andrewconner:
:money_with_wings: :moneybag: :dollar:

PR #17482, opened 2015-09-29
Split settings out. Add billing and export. Will refactor more.

cvializ:
@andrewconner (@camhashemi )

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/teamSettings/teamSettings.js
+++ b/kifi-backend/modules/shoebox/angular/src/teamSettings/teamSettings.js
@@ -133,7 +133,7 @@ angular.module('kifi')
       var items = Array.prototype.slice.apply(arguments);
       var options = [ // This is what the <select>s will read from
         { label: 'Admins only', value: ORG_SETTING_VALUE.ADMIN },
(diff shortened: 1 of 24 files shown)


andrewconner:
:100: :100: :100: :100: :100: :100: :100: :100:

kifi-backend/modules/shoebox/angular/src/teamSettings/teamSettings.js
```diff
@@ -0,0 +1,175 @@
...
+
+    $scope.getHeadingAnchor = function (heading) {
+      return heading.toLowerCase().replace(' ', '-');
+    };
+
+    function getOptions() {
+      var items = Array.prototype.slice.apply(arguments);
+      var options = [ // This is what the <select>s will read from
+        { label: 'Admins only', value: ORG_SETTING_VALUE.ADMIN },
+        { label: 'Members only', value: ORG_SETTING_VALUE.MEMBER },
```
camhashemi: 'Members only' seems to deny 'admins', maybe 'All members'... not sure

cvializ: I like it!

PR #17484, opened 2015-09-30
add account activity to the admin page

camhashemi:
@stkem @eishay


kifi-backend/modules/shoebox/conf/shoeboxService.routes
```diff
@@ -559,7 +559,7 @@ POST          /site/admin/billing/contacts                      @com.keepit.cont
 GET           /site/admin/billing/featureSettings               @com.keepit.controllers.website.PaymentsController.getAccountFeatureSettings(pubId: PublicId[Organization])
 POST          /site/admin/billing/featureSettings               @com.keepit.controllers.website.PaymentsController.setAccountFeatureSettings(pubId: PublicId[Organization])
 POST          /site/admin/billing/plan                          @com.keepit.controllers.website.PaymentsController.updatePlan(pubId: PublicId[Organization], planPubId: PublicId[com.keepit.payments.PaidPlan])
-GET           /site/admin/billing/events                        @com.keepit.controllers.website.PaymentsController.getEvents(pubId: PublicId[Organization], limit: Int)
+GET           /site/admin/billing/events                        @com.keepit.controllers.website.PaymentsController.getEvents(pubId: PublicId[Organization], offset: Int, limit: Int)
```
stkem: What is `offset` for? If it is pagination, that is not a reliable method to do that, since new events may come in as you paginate. In fact that (i.e. pagination) is what the endpoint right below this is for (`eventsBefore`).

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminPaymentsController.scala
```diff
@@ -2,12 +2,14 @@ package com.keepit.controllers.admin
 
 import com.keepit.common.controller.{ AdminUserActions, UserActionsHelper }
 import com.keepit.common.db.slick.Database
-import com.keepit.model.{ OrganizationRepo, OrganizationStates, Organization, User }
+import com.keepit.model._
```
stkem: :-(

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminPaymentsController.scala
```diff
@@ -132,4 +135,47 @@ class AdminPaymentsController @Inject() (
...
+      s"Account Activity for ${org.name}",
+      { page: Int => com.keepit.controllers.admin.routes.AdminPaymentsController.getAccountActivity(orgId, page) }.andThen(asPlayHtml),
+      page,
+      events.length,
+      PAGE_SIZE))
+  }
+
 }
+
+case class AdminAccountEventView(
```
stkem, line 178: I think it might be a good idea to include the (real, not public) id of the event here, so it's easy to find in the database when there is a problem.

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminPaymentsController.scala
```diff
@@ -132,4 +135,47 @@ class AdminPaymentsController @Inject() (
...
+      memo = accountEvent.memo
+    )
+  }
+
+  def asPlayHtml(obj: Any) = HtmlFormat.raw(obj.toString)
+
+  def getAccountActivity(orgId: Id[Organization], page: Int) = AdminUserAction { implicit request =>
+    val PAGE_SIZE = 50
+    val events = planCommander.getAccountEvents(orgId, page * PAGE_SIZE, (page + 1) * PAGE_SIZE, onlyRelatedToBillingFilter = None)
+      .map { createAdminAccountEventView }
```
stkem: Nothing wrong with this as such, but when just passing in a function to `map`, you'd more idiomatically write it as `.map(createAdminAccountEventView)` (curly braces being _typically_ reserved for blocks)

kifi-backend/modules/shoebox/conf/shoeboxService.routes
camhashemi: got it, thanks

camhashemi: yes, trying to paginate here

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminPaymentsController.scala
camhashemi, line 178: great. if there's anything else you might find useful, please let me know.

camhashemi:
@stkem updated

kifi-backend/modules/shoebox/app/views/admin/accountActivity.scala.html
```diff
@@ -0,0 +1,32 @@
...
+                <tr>
+                    <td>@{event.action.eventType}</td>
+                    <td>@adminHelper.dateTimeDisplay(event.eventTime)</td>
+                    <td>@{event.billingRelated}</td>
+                    <td>@event.whoDunnit.map(adminHelper.userDisplay(_)).getOrElse("none")</td>
+                    <td>@event.adminInvolved.map(adminHelper.userDisplay(_)).getOrElse("none")</td>
+                    <td>@{event.creditChange.cents * 100}</td>
+                    <td>@{event.paymentCharge.map(_.cents).getOrElse(0) * 100}</td>
+                    <td>@{event.memo.getOrElse("")}</td>
+                </tr>
```
stkem, line 27: Would be good to actually display the id :-)

camhashemi, line 27: crap! thanks

PR #17487, opened 2015-09-30
send accepted org invite notif to inviter

camhashemi:
@leogrim @ryanpbrewster seems like this was removed in a refactor, not sure if it was intentional or not.

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/UserInteractionCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/UserInteractionCommander.scala
@@ -18,8 +18,8 @@ class UserInteractionCommander @Inject() (
     }.as[List[JsObject]]
 
     val newJson = recipients collect {
-      case (UserRecipient(id), interaction) => Json.obj("user" -> id, "action" -> interaction.value)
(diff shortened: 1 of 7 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/UserInteractionCommander.scala
```diff
@@ -81,5 +81,5 @@ object UserInteraction {
 }
 
 sealed trait InteractionRecipient
-case class UserRecipient(id: Id[User]) extends InteractionRecipient
-case class EmailRecipient(address: EmailAddress) extends InteractionRecipient
+case class UserInteractionRecipient(id: Id[User]) extends InteractionRecipient
+case class EmailInteractionRecipient(address: EmailAddress) extends InteractionRecipient
```
camhashemi, line 85: these have naming collisions with subclasses of `keepit.notify.model.Recipient`

leogrim, line 85: :+1:

leogrim:
lgtm!

ryanpbrewster:
:+1: merge away

PR #17488, opened 2015-09-30
[Option 1] Stop denormalizing permissions from an organization's feature settings

ryanpbrewster:
@leogrim @camhashemi @andrewconner 

This is an overhaul of the permissions architecture. The current architecture is complex:

```
OrganizationMember.permissions: Set[OrganizationPermission]
Organization.basePermissions: Map[Option[OrganizationRole], Set[OrganizationPermission]]
PaidAccount.featureSettings: Set[FeatureSetting]
```

were all "sources of truth" that had to be kept in sync. It allowed for some cool behavior (specific organization members could be individually granted permissions), but overall it seems fragile.

The new architecture is relatively straightforward.

```
OrganizationConfiguration.settings: OrganizationSettings
```

OrganizationSettings can be thought of as a `Map[Feature, FeatureSetting[Feature]]`, but it's typesafe and you can't set a feature to a setting that doesn't make sense. That is the single source of truth, and all permissions are derived from the `OrganizationSettings`.


kifi-backend/modules/common/app/com/keepit/model/OrganizationSettings.scala
```diff
@@ -0,0 +1,189 @@
...
+    __.read[String].map(Feature(_)),
+    Writes { x => JsString(x.value) }
+  )
+
+  case object PublishLibraries extends Feature {
+    type F = PublishLibraries.type
+    val value = OrganizationPermission.PUBLISH_LIBRARIES.value
+    def permissions = Set(OrganizationPermission.PUBLISH_LIBRARIES)
+
+    case object DISABLED extends FeatureSetting[F]("disabled") { val rolesAffected = OrganizationRole.NOONE }
```
andrewconner: I like pronouncing "noone" as if it rhymes with "moon".

kifi-backend/modules/common/app/com/keepit/model/OrganizationSettings.scala
```diff
@@ -0,0 +1,189 @@
...
+    case object MEMBERS extends FeatureSetting[F]("members") { val rolesAffected = OrganizationRole.MEMBERS_UP }
+    val settings: Set[FeatureSetting[F]] = Set(DISABLED, ADMINS, MEMBERS)
+  }
+
+  case object InviteMembers extends Feature {
+    type F = InviteMembers.type
+    val value = OrganizationPermission.INVITE_MEMBERS.value
+    def permissions = Set(OrganizationPermission.INVITE_MEMBERS)
+
+    case object DISABLED extends FeatureSetting[F]("disabled") { val rolesAffected = OrganizationRole.NOONE }
```
andrewconner: Could `rolesAffected` just be inside the constructor?

kifi-backend/modules/common/app/com/keepit/model/OrganizationSettings.scala
```diff
@@ -0,0 +1,189 @@
...
+  def value: String
+  def settings: Set[FeatureSetting[F]]
+  def permissions: Set[OrganizationPermission]
+
+  def toSetting(x: String): Option[FeatureSetting[F]] = settings.find(_.value == x)
+  implicit def settingFormat: Format[FeatureSetting[F]] = Format(__.read[String].map(toSetting(_).get), Writes { x => JsString(x.value) } )
+}
+
+sealed abstract class FeatureSetting[F <: Feature](val value: String) {
+  def rolesAffected: Set[Option[OrganizationRole]]
```
andrewconner: Seems like previously, settings (generically) did not have to concern themselves with roles. Now, all settings have to define roles they affect (empty set included). What's the advantage?

ryanpbrewster: It avoids type hell. I'm not thrilled with it, but it's much cleaner and more maintainable than the alternative IMO.

kifi-backend/modules/common/app/com/keepit/model/OrganizationSettings.scala
ryanpbrewster: :+1: good suggestion. I was actually in the middle of doing that. :)

kifi-backend/modules/common/app/com/keepit/model/OrganizationSettings.scala
andrewconner: I'm not sure it is; settings that have no connection to permissions will have to carry along this. Having to avoid type hell may be a sign that it's overly complicated.

kifi-backend/modules/common/app/com/keepit/model/OrganizationSettings.scala
```diff
@@ -0,0 +1,187 @@
...
+    }
+  )
+}
+
+sealed trait Feature { self =>
+  type F >: self.type <: Feature
+  def instance: F = self
+  def value: String
+  def settings: Set[FeatureSetting[F]]
+  def permissions: Set[OrganizationPermission]
```
andrewconner, line 45: This is the bit that seems out of place.

kifi-backend/modules/common/app/com/keepit/model/OrganizationSettings.scala
andrewconner: The alternative was keeping what we had, or building on it. It used types less, but was simple and avoided type hell.

I'm a big fan of unifying permissions and settings, the rest is great and is better than before. That's separate from this, though.

camhashemi: Seems like we're all wary about biasing `Feature` and `FeatureSetting` towards `OrganizationPermission` and `OrganizationRole`. Can we make subclasses that include those fields, but leave `Feature` and `FeatureSetting` generic?

ryanpbrewster: There are two options (within this framework) that I see:
    1. Every `Feature` defines `permissions` which are just the set of permissions that it grants, defaulting to `Set.empty`. Every `FeatureSetting` defines the `roles` that it affects, probably defaulting to `Set.empty`. For features that are unconnected to permissions, the defaults make sense. This is a little ugly because it's weird to put `Set.empty` in places where it seems entirely unnecessary to specify permissions/roles.
    2. Try and subclass `Feature` into `OrgPermissionFeature`, which defines `permissions`, and subclass `FeatureSetting[F <: Feature]` into `OrgPermissionFeatureSetting[F <: OrgPermissionFeature]` which defines `roles`. I tried this initially. It was difficult to get the types to work out, and was more verbose. I'm happy to take another look into it if you think that's a non-negligibly better solution.

It's also possible that this is the entirely incorrect approach. In practice I've found it to be much more pleasant to work with than our current system. It caught a few of my silly errors at compile time.

camhashemi: I don't think mixing up settings and features was a huge problem in the previous system. By having static objects for each setting, you can just to a lookup like your `toSetting`. The huge problem was the interdependencies between permissions and settings, which you've cleaned up well.

If we bias all `Feature` and `FeatureSetting` to roles and permissions (or have to carry around unused fields/methods), I don't think heavily typing is "more maintainable" going forward. We'll likely add more fields, functions, etc to `Feature` and `FeatureSetting` regarding permissions and permissions propagation, which would all be irrelevant to certain features.

PR #17490, opened 2015-09-30
Argh, comparing different types with `==` should REALLY be a compiler warning

stkem:
I think this might have been what caused the integrity checker to malfunction. We will see shortly.

@ryanpbrewster @camhashemi


kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEventRepo.scala
```diff
@@ -102,7 +102,7 @@ class AccountEventRepoImpl @Inject() (
   }
 
   def getMemebershipEventsInOrder(accountId: Id[PaidAccount])(implicit session: RSession): Seq[AccountEvent] = {
```
camhashemi, line 104: `getMembership`

camhashemi:
nice find!

stkem:
well, took me the better part of a day to find :-(

PR #17491, opened 2015-09-30
Don't break stylus reload on syntax errors and update to new gulp-stylus

DerekBurch:
The stylus errors will still be output in the console.


cvializ:
:clap: :toilet:

andrewconner:
:100:

PR #17493, opened 2015-09-30
[Option 2] Less typesafe, more extensible

ryanpbrewster:
@andrewconner @camhashemi @leogrim


kifi-backend/modules/common/app/com/keepit/model/OrganizationSettings.scala
```diff
@@ -0,0 +1,136 @@
...
+
+
+sealed trait Feature[K <: FeatureKind] { self =>
+  val value: String
+  def settings: Set[FeatureSetting[K]]
+
+  private def toSetting(x: String): Option[FeatureSetting[K]] = settings.find(_.value == x)
+  def settingReads: Reads[FeatureSetting[K]] = __.read[String].map(x => toSetting(x).get)
+
+  def makePair(js: JsValue): FeatureSettingPair[K] = FeatureSettingPair(this, js.as(settingReads))
```
ryanpbrewster, line 50: This is my least favorite part of this PR

kifi-backend/modules/common/app/com/keepit/model/OrganizationSettings.scala
```diff
@@ -0,0 +1,128 @@
...
+    case InviteMembers.value => InviteMembers
+    case MessageOrganization.value => MessageOrganization
+    case ForceEditLibraries.value => ForceEditLibraries
+    case ViewOrganization.value => ViewOrganization
+    case ViewMembers.value => ViewMembers
+    case RemoveLibraries.value => RemoveLibraries
+    case CreateSlackIntegration.value => CreateSlackIntegration
+    case EditOrganization.value => EditOrganization
+    case ExportKeeps.value => ExportKeeps
+  }
```
camhashemi, line 81: `case _ => ViewOrganization` might save our ass at some point, since so many things depend on this serialization to work.

kifi-backend/modules/common/app/com/keepit/model/OrganizationSettings.scala
```diff
@@ -0,0 +1,128 @@
...
+
+import play.api.libs.json._
+
+case class OrganizationSettings(kvs: Set[FeatureSettingPair[_ <: FeatureKind]]) {
+  def setAll(newKvs: Set[FeatureSettingPair[_ <: FeatureKind]]): OrganizationSettings = {
+    val newFeatures = newKvs.map(_.feature)
+    this.copy(kvs = kvs.filter(kv => !newFeatures.contains(kv.feature)) ++ newKvs)
+  }
+
+  def set[K <: FeatureKind](newKvs: (Feature[K], FeatureSetting[K])*): OrganizationSettings = setAll(newKvs.map(kv => FeatureSettingPair(kv._1, kv._2)).toSet)
```
leogrim: I would say `with` rather than `set` since this is making a copy, not mutating anything (just like with do on models).

kifi-backend/modules/common/app/com/keepit/model/OrganizationSettings.scala
leogrim, line 50: Could call it `reads` and return a `JsResult`, maybe even have `Feature` extends the `Reads` trait?

ryanpbrewster, line 50: I have a `Feature[_]` and I need to create a `FeatureSettingPair[_]`. So somehow I need to go from `(Feature[_], JsValue) => FeatureSettingPair[_]`. I couldn't figure out a way to serialize the `FeatureSetting` and construct the `FeatureSettingPair` in two steps.

kifi-backend/modules/common/app/com/keepit/model/OrganizationSettings.scala
ryanpbrewster: :+1:

kifi-backend/modules/common/app/com/keepit/model/OrganizationSettings.scala
leogrim, line 50: Sure, just saying we can use the regular `Reads` API to make this more familiar

camhashemi, line 50: where you use this above, you start with a `JsValue` for`Feature` as well, couldn't you make this a part of `FeatureSettingPair` and have it construct both of them from two `JsValue`?

kifi-backend/modules/common/app/com/keepit/model/OrganizationSettings.scala
```diff
@@ -0,0 +1,128 @@
...
+  def settings: Set[FeatureSetting[K]]
+
+  private def toSetting(x: String): Option[FeatureSetting[K]] = settings.find(_.value == x)
+  def settingReads: Reads[FeatureSetting[K]] = __.read[String].map(x => toSetting(x).get)
+
+  def makePair(js: JsValue): FeatureSettingPair[K] = FeatureSettingPair(this, js.as(settingReads))
+}
+sealed abstract class PermissionFeature(val permission: OrganizationPermission) extends Feature[FeatureKind.Permission] {
+  val value = permission.value
+}
```
leogrim, line 54: I don't know if we should tie the feature's name to a single permission, we could have features that impact several permissions at once, or a different permission for each role.

kifi-backend/modules/common/app/com/keepit/model/OrganizationSettings.scala
```diff
@@ -0,0 +1,128 @@
...
+  case object CreateSlackIntegration extends PermissionFeature(OrganizationPermission.CREATE_SLACK_INTEGRATION) {
+    val settings: Set[FeatureSetting[FeatureKind.Permission]] = Set(DISABLED, ADMINS, MEMBERS)
+  }
+
+  case object EditOrganization extends PermissionFeature(OrganizationPermission.EDIT_ORGANIZATION) {
+    val settings: Set[FeatureSetting[FeatureKind.Permission]] = Set(DISABLED, ADMINS, MEMBERS)
+  }
+
+  case object ExportKeeps extends PermissionFeature(OrganizationPermission.EXPORT_KEEPS) {
+    val settings: Set[FeatureSetting[FeatureKind.Permission]] = Set(DISABLED, ADMINS, MEMBERS)
```
camhashemi, line 126: just got word from @JRuff42 that we're removing `MEMBERS` for this feature

PR #17496, opened 2015-09-30
deactivate all org invites when the inviter leaves the org

camhashemi:
@leogrim @ryanpbrewster


ryanpbrewster:
This is a product question. Are we sure that it makes sense to cancel a member's invites when they leave?

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationInviteRepo.scala
```diff
@@ -188,6 +189,10 @@ class OrganizationInviteRepoImpl @Inject() (val db: DataBaseComponent, val clock
     getByOrgIdAndUserIdCompiled(organizationId, userId, state).list
   }
 
+  def getByOrgIdAndInviterId(organizationId: Id[Organization], inviterId: Id[User], excludeState: Option[State[OrganizationInvite]] = Some(OrganizationInviteStates.INACTIVE))(implicit s: RSession): Seq[OrganizationInvite] = {
+    rows.filter(row => row.organizationId === organizationId && row.inviterId === inviterId && row.state =!= excludeState.orNull).run
```
ryanpbrewster, line 193: `.list`? I'm surprised that `.run` works here.

ryanpbrewster:
lgtm!

camhashemi:
@ryanpbrewster actually, this is going to change. We want to keep invites around such that we can get more people in orgs / Kifi. The question is how to do that in an elegant way.

camhashemi:
Jen (and I) wants us to keep these invites active, but `OrganizationMembershipCommander.addMembership` (justifiably) asserts that the inviter must be in the org, with appropriate permissions. If we throw away that assertion, there would be few validation checks left on adding an org membership.

One solution is to transfer the `OrganizationInvite.inviterId` to the current org owner. This ensures that all of these invites will always be acceptable. We may have to mute "invite accepted" notifications and/or inform the org owner in this case.

Curious as to your thoughts @leogrim

PR #17499, opened 2015-09-30
Added some general deep-link routing

ryanpbrewster:
@andrewconner @jaredpetker 

This breaks at least one test, so I'm gonna have to go back and make sure things work properly. Does this seem like it's on the right track, though?

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/PathCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/PathCommander.scala
@@ -13,7 +13,11 @@ class PathCommander @Inject() (
     basicUserRepo: BasicUserRepo,
     userRepo: UserRepo) {
 
-  def pathForLibrary(lib: Library): Path = Path(getPathForLibrary(lib))
+  // TODO(ryan): I feel bad for "fixing" this problem like this, but a bunch of existing
+  // code directly calls getPathForLibrary, which calls LibraryPathHelper, and that class
+  // does The Wrong Thing (it explicitly puts the "/" at the beginning of the link), while
+  // a `Path` assumes that there is no leading slash.
+  def pathForLibrary(lib: Library): Path = Path(getPathForLibrary(lib).tail)
 
   def pathForUser(user: User): Path = Path(user.username.value)
 
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/KifiSiteRouter.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/KifiSiteRouter.scala
@@ -76,7 +76,10 @@ class KifiSiteRouter @Inject() (
 
   def generalRedirect(data: String) = MaybeUserAction { implicit request =>
     val linkOpt = deepLinkRouter.generateRedirectUrl(Json.parse(data).as[JsObject])
-    linkOpt.map(path => Redirect(path.relative)).getOrElse(NotFound)
+    linkOpt.map(path => Redirect("/" + path.relative)).getOrElse {
+      airbrake.notify(s"Could not figure out how to redirect deep-link: ${data}")
+      Redirect("/get")
+    }
   }
 
   def handleInvitePage(friend: Option[String]) = WebAppPage { implicit request =>
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/KifiSiteRouter.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/KifiSiteRouter.scala
@@ -5,6 +5,7 @@ import com.keepit.commanders._
 import com.keepit.common.cache.TransactionalCaching.Implicits._
 import com.keepit.common.controller._
 import com.keepit.common.core._
+import com.keepit.common.crypto.{ PublicIdConfiguration, PublicId }
 import com.keepit.common.db.{ Id, ExternalId }
 import com.keepit.common.db.slick.Database
 import com.keepit.common.healthcheck.AirbrakeNotifier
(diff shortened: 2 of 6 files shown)


kifi-backend/modules/shoebox/app/com/keepit/controllers/website/KifiSiteRouter.scala
```diff
@@ -69,6 +74,11 @@ class KifiSiteRouter @Inject() (
     }
   }
 
+  def generalRedirect(data: String) = MaybeUserAction { implicit request =>
+    val linkOpt = deepLinkRouter.generateRedirectUrl(Json.parse(data).as[JsObject])
+    linkOpt.map(path => Redirect(path.relative)).getOrElse(NotFound)
```
andrewconner: In this case, log it (perhaps airbrake initially?), and forward to `"/get"`. That's a redirect that will go to the app store if mobile, and to the site if not. For this, it's the best we can do.

kifi-backend/modules/shoebox/test/com/keepit/controllers/website/DeepLinkRouterTest.scala
```diff
@@ -0,0 +1,95 @@
...
+    FakeKeepImportsModule(),
+    FakeCryptoModule()
+  )
+
+  "DeepLinkRouter" should {
+    implicit val context = HeimdalContext.empty
+    implicit def publicIdConfig(implicit injector: Injector) = inject[PublicIdConfiguration]
+    implicit def deepLinkRouter(implicit injector: Injector) = inject[DeepLinkRouter]
+    implicit class PathWrapper(pathOpt: Option[Path]) {
+      def ===(that: Option[Path]) = pathOpt.map(_.absolute) must equalTo(that.map(_.absolute))
```
andrewconner, line 44: Typeclasses! Wahoo, awesome.

kifi-backend/modules/shoebox/test/com/keepit/controllers/website/DeepLinkRouterTest.scala
```diff
@@ -0,0 +1,95 @@
...
+          val user = db.readWrite { implicit session =>
+            UserFactory.user().saved
+          }
+          val deepLink = Json.obj("t" -> "nf", "uid" -> user.externalId)
+          val link = Path(s"${user.username.value}")
+          deepLinkRouter.generateRedirectUrl(deepLink) === Some(link)
+        }
+      }
+      "for a new message" in {
+        skipped("need to figure out how to route a message-related deep-link")
```
andrewconner, line 85: Yep, for now, not needed since the site can't do it.

kifi-backend/modules/shoebox/test/com/keepit/controllers/website/DeepLinkRouterTest.scala
ryanpbrewster, line 44: It's a sad day in a programmer's life when their willpower finally breaks and they sit down to start writing `implicit class` wrappers.

ryanpbrewster, line 44: Or a happy day, it's hard to me to say. :)

andrewconner, line 44: I guess the regular `===` didn't work because `value` is private? I know there's a way to tell the regular `===` how to compare then. Something something `beEqualTo`

ryanpbrewster, line 44: `Path` is not a case class, so it doesn't have an intelligent `==` that `===` can use. I think it was checking hashes or something?

At least that's what I assumed. It turns out that we use `Path` slightly inconsistently, so maybe that was the whole issue.

kifi-backend/modules/shoebox/app/com/keepit/commanders/PathCommander.scala
```diff
@@ -13,7 +13,11 @@ class PathCommander @Inject() (
     basicUserRepo: BasicUserRepo,
     userRepo: UserRepo) {
 
-  def pathForLibrary(lib: Library): Path = Path(getPathForLibrary(lib))
+  // TODO(ryan): I feel bad for "fixing" this problem like this, but a bunch of existing
```
andrewconner, line 16: Does this break anything?

kifi-backend/modules/shoebox/app/com/keepit/commanders/PathCommander.scala
```diff
@@ -13,7 +13,11 @@ class PathCommander @Inject() (
...
+  // code directly calls getPathForLibrary, which calls LibraryPathHelper, and that class
+  // does The Wrong Thing (it explicitly puts the "/" at the beginning of the link), while
```
andrewconner, line 18: here's a unicode symbol for you to use: ™

kifi-backend/modules/shoebox/test/com/keepit/controllers/website/DeepLinkRouterTest.scala
ryanpbrewster, line 44: Nah, it's definitely doing _something_, but not the right thing.

kifi-backend/modules/shoebox/app/com/keepit/commanders/PathCommander.scala
ryanpbrewster, line 16: This is my workaround to avoid breaking anything. I think we should go in and globally fix the way that `Path` and `String` are used wrt paths.

PR #17501, opened 2015-10-01
Do not schedule calls to Embedly, rely on content changes detected by…

leogrim:
… our own scrapers 

@andrewconner


andrewconner:
:+1:

PR #17504, opened 2015-10-01
Start using gulp-svgstore instead of custom ruby script for symbol sprites

jaredpetker:
@cvializ @andrewconner 

This moves over to a more standard way to build these symbol svgs using already existing node packages / gulp tasks and move off of our custom ruby script.

Also lays some groundwork with some thoughts I had on being able to color svgs more fully. Currently you can only change the fill color of an svg, this adds the ability to be able to specify values for keys in each svg, such as

`<path stroke="color1" ....`

Which can then be used as attributes on the svg symbol sprites:

`<svg  kf-symbol-sprite    icon="some-icon" color1="#FF000" color2=....></svg>`

This actually works but you have to edit the symbol-sprites.svg manually until it's automated

Just the first simple way that came to mind, these values can potentially be inserted depending on layer names in the svg in part of the svg compilation steps, open to other thoughts if this is something that sounds interesting.


andrewconner:
Nice. @cvializ you were there for Adam's ruby script. How does this look?

cvializ:
This looks pretty good! I am 100% behind moving away from the custom Ruby script, so this is awesome.

One downside of the color implementation, though, is we want to be able to specify the colors in CSS instead of in JavaScript. If we set the colors through the directive, it takes away the ability to style the SVG with the stylesheet. On hover for example, we want to be able to use `.my-svg-symbol-sprite:hover { fill: red; stroke: green; }` instead of adding event listeners to check for that change.

http://codepen.io/FWeinb/post/quick-tip-svg-use-style-two-colors

^ This is the direction I wanted to take to make two-tone colors possible, using the special CSS color `currentColor` in our SVG file. Maybe you could investigate making this method possible with our symbol-sprites?

jaredpetker:
Yeaa, that's an issue, which is why I haven't fully invested into it yet. In the very least this moves to using more standard node / gulp build tools and can pull out the stuff from the directive.

I'm with you on the styling, again, the reason I haven't doubled down into it. I saw the stuff with currentColor, it would be really nice to be able to alter multiple colors, i.e. more than 2, but perhaps that is just impossible to do with css at the moment. Can think about it some more.

jaredpetker:
@cvializ I'm removing the edits to the symbol-sprite directive for now into a separate pull request that can be thought about, this one now just has improvements to use nodejs + gulp plugins instead of the ruby script

ryanpbrewster:
Is this PR mergeable?

jaredpetker:
It was once upon a time, I'll have to test it again to make sure it's all
kosher still, probably is, can't guarantee it.

ryanpbrewster:
How would you feel about closing this PR and revisiting it later?

leogrim:
@DerekBurch @andrewconner @cvializ can we make up the fate of this? I have no idea what it's about ;)

andrewconner:
Carlos's call.

cvializ:
I'll check it out and determine its fate

ryanpbrewster:
https://media.giphy.com/media/iJxHzcuNcCJXi/giphy.gif

PR #17505, opened 2015-10-01
[Option 3] No typesafety at all

ryanpbrewster:
@leogrim @camhashemi @andrewconner


kifi-backend/modules/common/app/com/keepit/model/OrganizationSettings.scala
```diff
@@ -0,0 +1,156 @@
...
+}
+
+sealed trait Feature {
+  val value: String
+  def settings: Set[FeatureSetting]
+
+  private def toSetting(x: String): Option[FeatureSetting] = settings.find(_.value == x)
+  def settingReads: Reads[FeatureSetting] = __.read[String].map(x => toSetting(x).get)
+}
+sealed trait ExtraPermissions {
```
camhashemi: every time I see this name I have to ask "what does that mean again"? A better name could be "FeaturePermissions"

ryanpbrewster: It's a trait, so it gets mixed into specific features. Something like
`case object MyFavoriteFeature extends Feature with ExtraPermissions`

camhashemi: sure, but still not sure what the trait specifies

i.e. when does a `Feature extends ExtraPermissions` and when does it not

leogrim:
lgtm, let's run with this

kifi-backend/modules/common/app/com/keepit/model/OrganizationSettings.scala
```diff
@@ -0,0 +1,157 @@
...
+  def withSettings(newKvs: (Feature, FeatureSetting)*): OrganizationSettings = setAll(newKvs.toMap)
+  def extraPermissionsFor(roleOpt: Option[OrganizationRole]): Set[OrganizationPermission] = kvs.collect {
+    case (feature: FeatureWithPermissions, setting: FeatureSetting) => feature.extraPermissionsFor(roleOpt, setting)
+  }.toSet.flatten
+
+  def diff(that: OrganizationSettings): Set[Feature] = {
+    this.kvs.collect {
+      case (feature, setting) if that.kvs(feature) != setting => feature
+    }.toSet
+  }
```
camhashemi, line 18: noice

ryanpbrewster:
Okay team, I'm gonna merge this in about an hour. I'm hoping everything will go smoothly, but I have a healthy amount of skepticism.

andrewconner:
Looks good to me. :+1: merge away

kifi-backend/modules/common/conf/evolutions/shoebox/398.sql
```diff
@@ -0,0 +1,28 @@
...
+ALTER TABLE paid_plan ADD COLUMN editable_features text NOT NULL;
+ALTER TABLE paid_plan ADD COLUMN default_settings text NOT NULL;
+
+CREATE TABLE organization_configuration (
+    id bigint(20) NOT NULL AUTO_INCREMENT,
+    created_at datetime NOT NULL,
+    updated_at datetime NOT NULL,
+    state varchar(20) NOT NULL,
+    organization_id bigint(20) NOT NULL,
+    settings text NOT NULL,
```
camhashemi, line 19: could be nice at some point to have a column for each feature, then we can just add/drop/edit a column whenever we want.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaymentRequest.scala
```diff
@@ -1,7 +1,7 @@
 package com.keepit.payments
 
-import com.keepit.common.db.ExternalId
-import com.keepit.model.{ Name, User }
+import com.keepit.common.db.{ Id, ExternalId }
+import com.keepit.model._
```
camhashemi, line 4: :(

camhashemi:
:+1: lgtm.

PR #17509, opened 2015-10-01
Payment Charge Email Notification (not the final copy)

stkem:
@ryanpbrewster @camhashemi @andrewconner @leogrim


kifi-backend/modules/shoebox/app/com/keepit/payments/StripeClient.scala
```diff
@@ -87,4 +89,8 @@ class StripeClientImpl(mode: Mode, implicit val ec: ExecutionContext) extends St
     StripeToken(customer.getId())
   }
 
+  def getLastFourDigitsOfCard(token: StripeToken): Future[String] = lock.withLock {
+    Customer.retrieve(token.token).getSources().getData().get(0).asInstanceOf[Card].getLast4
```
stkem, line 93: Did I mention that the Stripe Java client is terrible? :-/

kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEventRepo.scala
```diff
@@ -117,7 +117,7 @@ class AccountEventRepoImpl @Inject() (
   }
 
   def getMembershipEventsInOrder(accountId: Id[PaidAccount])(implicit session: RSession): Seq[AccountEvent] = {
-    (for (row <- rows if row.accountId === accountId && (row.eventType == "user_added" || row.eventType == "user_removed")) yield row).sortBy(r => (r.eventTime asc, r.id asc)).list
+    (for (row <- rows if row.accountId === accountId && (row.eventType === "user_added" || row.eventType === "user_removed")) yield row).sortBy(r => (r.eventTime asc, r.id asc)).list
```
stkem, line 120: I fixed this in a previous PR, but it got undone in someones bad merge.

kifi-backend/modules/shoebox/app/com/keepit/payments/StripeClient.scala
andrewconner, line 93: Idiomatic Java :smile: 

Does it ever return `null`? May want to wrap things in options.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaymentProcessingCommander.scala
```diff
@@ -46,6 +50,43 @@ class PaymentProcessingCommanderImpl @Inject() (
...
+        Try(emailRepo.getByUser(userId)).toOption
+      }.flatten
+      (userContacts, orgRepo.get(account.orgId))
+    }
+    val emails = (account.emailContacts ++ userContacts).toSet.toSeq
+
+    lastFourFuture.map { lastFour =>
+      val subject = "We've charged you card for your Kfi Organization ${org.name}"
+      val htmlBody = """|You card on file ending in $lastFour has been charged $amount (ref. $chargeId).
+      |For more details please consult your account history at <a href="https://www.kifi.com/kifi/settings">https://www.kifi.com/kifi/settings<a>.
```
andrewconner: Hard coded org handle?

kifi-backend/modules/shoebox/app/com/keepit/payments/PaymentProcessingCommander.scala
```diff
@@ -46,6 +50,43 @@ class PaymentProcessingCommanderImpl @Inject() (
...
+      |
+      |Thanks,
```
andrewconner: Not html? Will be on all one line, right?

kifi-backend/modules/common/app/com/keepit/common/mail/EmailAddress.scala
```diff
@@ -97,9 +97,10 @@ object SystemEmailAddress {
   val SALES = EmailAddress("sales@kifi.com")
   val ASHLEY = EmailAddress("ashley@kifi.com") //keep it around for twitter communication
   val CONGRATS = EmailAddress("congrats@kifi.com")
+  val BILLING = EmailAddress("BILLING@kifi.com")
```
ryanpbrewster: Why all caps?

stkem: Umm, good question. My brain probably had caps lock on from typing the variable name. I'll change it.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaymentProcessingCommander.scala
camhashemi: since you have the anchor, could be more pleasant on the eyes to omit the protocol from the hypertext.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaymentProcessingCommander.scala
stkem: Oops. Fixing.

kifi-backend/modules/shoebox/app/com/keepit/payments/StripeClient.scala
stkem, line 93: I have no idea if it ever returns null, but if it does a serious contract is violated, i.e. if we have a token there should always be at least one card at the other end. We can't really recover in any meaningful way if it doesn't.

kifi-backend/modules/shoebox/app/com/keepit/payments/PaymentProcessingCommander.scala
stkem: Line breaks should be preserved. Will find out.

kifi-backend/modules/shoebox/app/com/keepit/payments/StripeClient.scala
andrewconner, line 93: Sure, but we can do better than NPE, like airbraking what the token was or something. As-is, if it returns a null, it's a time bomb that doesn't immediately go off, and will go off somewhere else.

PR #17510, opened 2015-10-01
Introduce UserIdentityCache

leogrim:
@andrewconner 
The invalidation has to happen in several places. It's not actually a consequence of the new code, just that we used to have stale data in `SocialUserInfo`. We could use the user's external id instead of the email address as the `IdentityId` identifier, would make the invalidation easier in `UserRepo` and `UserCredRepo` (not having to look up all the user's email addresses) but it would still have to happen there as well.


andrewconner:
Awesome

PR #17512, opened 2015-10-01
Actually do the mobile redirect

ryanpbrewster:
@jaredpetker @DerekBurch does this look right for a mobile app redirect?


kifi-backend/modules/shoebox/app/views/mobile/mobileAppRedirect.scala.html
```diff
@@ -35,5 +35,7 @@
   } else {
```
ryanpbrewster, line 35: Specifically this android thing

PR #17513, opened 2015-10-01
org invites are valid even after the inviter leaves or loses permissions

camhashemi:
@ryanpbrewster @leogrim 

talked with Jen, we want these invites to be acceptable. we don't care if permissions or memberships change between sending and accepting invites: any member can cancel invites they don't want active, so we don't need to manage it for them.


ryanpbrewster:
For the record, I'm really uncomfortable with this change. I don't think it makes sense to let people accept invites from non-members. I understand the product reasoning for wanting this, but I don't think that this situation is going to come up often, and it's not worth the added complexity.

leogrim:
Jen made a call on this one: https://app.asana.com/0/32728208409723/53668530804866
Org members are notified when someone new joins, if they don't like it they can always remove that person. So let's go with this :)

ryanpbrewster:
Yeah, it makes sense and I understand the reasoning. The code lgtm, nice work @camhashemi

kifi-backend/modules/shoebox/test/com/keepit/controllers/mobile/MobileOrganizationInviteControllerTest.scala
```diff
@@ -54,71 +54,6 @@ class MobileOrganizationInviteControllerTest extends Specification with ShoeboxT
...
-            val invite = inject[OrganizationInviteRepo].save(OrganizationInvite(organizationId = org.id.get, inviterId = inviter.id.get, userId = invitee.id, role = OrganizationRole.MEMBER))
-            (invitee, invite)
-          }
-
-          inject[FakeUserActionsHelper].setUser(invitee)
-          val request = route.acceptInvitation(publicOrgId, Some(invite.authToken))
-          val result = controller.acceptInvitation(publicOrgId, Some(invite.authToken))(request)
-          result === OrganizationFail.NO_VALID_INVITATIONS
-        }
-      }
```
ryanpbrewster, line 98: Can we keep these two tests around (and modify the second one so that it ensures that the invite is accepted)?

kifi-backend/modules/shoebox/test/com/keepit/controllers/website/OrganizationInviteControllerTest.scala
```diff
@@ -74,27 +74,6 @@ class OrganizationInviteControllerTest extends Specification with ShoeboxTestInj
...
-            val invite = inject[OrganizationInviteRepo].save(OrganizationInvite(organizationId = org.id.get, inviterId = inviter.id.get, userId = invitee.id, role = OrganizationRole.MEMBER))
-            (invitee, invite)
-          }
-
-          inject[FakeUserActionsHelper].setUser(invitee)
-          val request = route.acceptInvitation(publicOrgId, Some(invite.authToken))
-          val result = controller.acceptInvitation(publicOrgId, Some(invite.authToken))(request)
-          result === OrganizationFail.NO_VALID_INVITATIONS
-        }
-      }
```
ryanpbrewster, line 96: Makes sense to kill this :+1:

kifi-backend/modules/shoebox/test/com/keepit/controllers/website/OrganizationInviteControllerTest.scala
```diff
@@ -306,16 +286,35 @@ class OrganizationInviteControllerTest extends Specification with ShoeboxTestInj
...
+        withDb(controllerTestModules: _*) { implicit injector =>
+          val (org, owner, inviter, cannotInvite, notAMember, notAMember2) = setupInviters()
+          implicit val config = inject[PublicIdConfiguration]
+          val publicId = Organization.publicId(org.id.get)
+
+          inject[FakeUserActionsHelper].setUser(notAMember)
+          val jsonInput = Json.parse(
+            s"""{ "invites": [
+             |{ "id":  "${notAMember2.externalId.id}"}
+                             |]}""".stripMargin)
```
ryanpbrewster: wacky formatting

kifi-backend/modules/shoebox/test/com/keepit/controllers/mobile/MobileOrganizationInviteControllerTest.scala
camhashemi, line 98: :+1: good call

PR #17515, opened 2015-10-01
Sub source to kept events

andrewconner:
@DerekBurch Alright, this tiny bit should do it.

Basically, on any sane keeping endpoint, if you throw in a `subsource` url query string, it'll make its way to the event. If you don't, it won't. Same on any platform. Works?


andrewconner:
:blue_car:

DerekBurch:
🎆

andrewconner:
That's beautiful.

PR #17518, opened 2015-10-01
user profile and lib links -> deeplinks

ryanpbrewster:
Added a profileLink() function that deeplinks to a user's profile, and a libraryLink() that deeplinks
to a library's page

It should be used anywhere that we want a link (aka. href=blah). For URLs,
which are surfaced to users, we should stick to the URL version

@andrewconner @jaredpetker @DerekBurch

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/DeepLinkRouter.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/controllers/website/DeepLinkRouter.scala
@@ -20,8 +20,6 @@ class DeepLinkRouter @Inject() (
 
   def generateRedirectUrl(data: JsObject): Option[Path] = {
     (data \ "t").as[String] match {
-      // TODO(ryan): I think all of these should be added
-      // START_LIST
       case "vh" => Some(Path("")) // view home
       case "vf" => Some(Path("friends")) // view friends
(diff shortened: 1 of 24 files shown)


kifi-backend/modules/shoebox/app/com/keepit/controllers/website/DeepLinkRouter.scala
```diff
@@ -20,6 +20,17 @@ class DeepLinkRouter @Inject() (
...
+      // TODO(ryan): I think all of these should be added
+      // START_LIST
+      case "h" => Some(Path("")) // home
+      case "f" => Some(Path("friends")) // friends
+      case "il" => Some(Path("/me/libraries/invited")) // invited libraries
+      case "o" =>
+        val orgIdOpt = (data \ "oid").asOpt[PublicId[Organization]].flatMap(pubId => Organization.decodePublicId(pubId).toOption)
+        val orgOpt = orgIdOpt.map { orgId => db.readOnlyReplica { implicit session => orgRepo.get(orgId) } }
+        orgOpt.map(org => pathCommander.pathForOrganization(org))
+      // END_LIST
```
ryanpbrewster: @andrewconner @jaredpetker @DerekBurch 
how do you all feel about these additions to the push notification specs?

PR #17521, opened 2015-10-01
Org invite push notif

camhashemi:
@ryanpbrewster @leogrim

### DIFF
--- a/kifi-backend/modules/common/app/com/keepit/common/routes/Router.scala
+++ b/kifi-backend/modules/common/app/com/keepit/common/routes/Router.scala
@@ -233,6 +233,7 @@ object Eliza extends Service {
     def sendGeneralPushNotification() = ServiceRoute(POST, "/internal/eliza/sendGeneralPushNotification")
     def sendLibraryPushNotification() = ServiceRoute(POST, "/internal/eliza/sendLibraryPushNotification")
     def sendUserPushNotification() = ServiceRoute(POST, "/internal/eliza/sendUserPushNotification")
+    def sendOrgPushNotification() = ServiceRoute(POST, "/internal/eliza/sendOrgPushNotification")
     def sendToUserNoBroadcast() = ServiceRoute(POST, "/internal/eliza/sendToUserNoBroadcast")
     def sendToUser() = ServiceRoute(POST, "/internal/eliza/sendToUser")
     def sendToAllUsers() = ServiceRoute(POST, "/internal/eliza/sendToAllUsers")
--- a/kifi-backend/modules/common/app/com/keepit/eliza/ElizaServiceClient.scala
+++ b/kifi-backend/modules/common/app/com/keepit/eliza/ElizaServiceClient.scala
@@ -13,11 +13,13 @@ import com.keepit.common.zookeeper.ServiceCluster
 import com.keepit.notify.model.event.NotificationEvent
 import com.keepit.search.index.message.ThreadContent
 import com.keepit.common.cache.TransactionalCaching.Implicits.directCacheAccess
+import com.kifi.macros.json
 
 import scala.collection.mutable
 import scala.concurrent.{ ExecutionContext, Future }
 
 import play.api.libs.json._
+import play.api.libs.functional.syntax._
 
 import com.google.inject.Inject
 import com.google.inject.util.Providers
@@ -30,10 +32,12 @@ import com.keepit.common.core._
 sealed case class LibraryPushNotificationCategory(name: String)
 sealed case class UserPushNotificationCategory(name: String)
 sealed case class SimplePushNotificationCategory(name: String)
+sealed case class OrgPushNotificationCategory(name: String)
 
 object SimplePushNotificationCategory {
   val PersonaUpdate = SimplePushNotificationCategory("PersonaUpdate")
   val HailMerryUpdate = SimplePushNotificationCategory("HailMerryUpdate")
+  val OrganizationInvitation = SimplePushNotificationCategory("OrganizationInvitation")
 }
 
 object UserPushNotificationCategory {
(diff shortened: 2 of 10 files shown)


kifi-backend/modules/common/app/com/keepit/eliza/ElizaServiceClient.scala
```diff
@@ -60,6 +74,22 @@ object PushNotificationExperiment {
   implicit val format = Json.format[PushNotificationExperiment]
 }
 
+case class OrgPushNotificationRequest( // pretty bare right now, add org-specific fields as needed
+  userId: Id[User],
+  message: String,
+  pushNotificationExperiment: PushNotificationExperiment,
+  category: OrgPushNotificationCategory,
+  force: Boolean = false)
```
camhashemi, line 82: We'll probably want to add to this (e.g. `org.externalId`), but since we don't need them yet, I won't try to predict exactly what.

kifi-backend/modules/eliza/app/com/keepit/realtime/AppBoy.scala
```diff
@@ -93,7 +93,7 @@ class AppBoy @Inject() (
         val pushType = lupn.category match {
           case LibraryPushNotificationCategory.LibraryChanged => "lr"
           case LibraryPushNotificationCategory.LibraryInvitation => "li"
-          case _ => "lr"
+          case _ => airbrake.notify(s"unsupported library push notification category ${lupn.category.name}"); "lr"
```
ryanpbrewster: Nice :+1:

kifi-backend/modules/eliza/app/com/keepit/realtime/AppBoy.scala
```diff
@@ -107,9 +107,16 @@ class AppBoy @Inject() (
           case UserPushNotificationCategory.UserConnectionRequest => "fr"
           case UserPushNotificationCategory.ContactJoined => "us"
           case UserPushNotificationCategory.NewLibraryFollower => "nf"
-          case _ => "us"
+          case UserPushNotificationCategory.NewOrganizationMember => "om"
+          case _ => airbrake.notify(s"unsupported user push notification category ${upn.category.name}"); "us"
```
ryanpbrewster: I see you're putting this strategy to good use. I'm not sure, though, maybe it's better to fail than to send this type of notification. `"us"` specifically requires a field `"uid"` to be defined, so it may break things on clients.

ryanpbrewster: If these push notifications are done in a `SafeFuture` then we can throw exceptions safely and nothing will break.

ryanpbrewster:
Looks good to me, nice work. This was long overdue and I'm glad we finally tackled it.

kifi-backend/modules/eliza/app/com/keepit/realtime/AppBoy.scala
camhashemi: I hate that we're returning an arbitrary event (what's the desired behavior, to tell the user a contact joined when no such thing happened?). It makes no sense, I just have this fear of removing things that may have hidden effects. Happy to remove this and fail.

ryanpbrewster: :+1: I agree, it's just scary to have things that may be failing silently. The airbraking is a good call. I still think it'd be okay to throw an exception. Think about how much safer it would feel to modify this code if you saw that we were throwing exceptions. If you see that, there's basically nothing you can do that would be _worse_.

camhashemi: I actually wouldn't modify it, it'd be perfect. I have no idea why we'd allow an unsupported push notification to go through, unless we were implicitly supporting them (which is why I'm skeptical of modifying it)

kifi-backend/modules/eliza/app/com/keepit/realtime/AppBoy.scala
```diff
@@ -107,9 +107,16 @@ class AppBoy @Inject() (
...
           case UserPushNotificationCategory.ContactJoined => "us"
           case UserPushNotificationCategory.NewLibraryFollower => "nf"
-          case _ => "us"
+          case UserPushNotificationCategory.NewOrganizationMember => "om"
+          case _ => throw new Exception(s"unsupported user push notification category ${upn.category.name}")
         }
         json.as[JsObject] ++ Json.obj("t" -> pushType, "uid" -> upn.userExtId, "un" -> upn.username.value, "purl" -> upn.pictureUrl)
+      case opn: OrganizationPushNotification =>
+        val pushType = opn.category match {
+          case OrgPushNotificationCategory.OrganizationInvitation => "oi"
```
camhashemi, line 116: @DerekBurch @jaredpetker new notification type, should route to the invites pane

kifi-backend/modules/eliza/app/com/keepit/realtime/AppBoy.scala
```diff
@@ -107,9 +107,16 @@ class AppBoy @Inject() (
           case UserPushNotificationCategory.UserConnectionRequest => "fr"
           case UserPushNotificationCategory.ContactJoined => "us"
           case UserPushNotificationCategory.NewLibraryFollower => "nf"
-          case _ => "us"
+          case UserPushNotificationCategory.NewOrganizationMember => "om"
```
camhashemi, line 110: @jaredpetker @DerekBurch new notification type, should route to the new member's page (you get a `uid` with this one)

kifi-backend/modules/common/app/com/keepit/common/routes/Router.scala
```diff
@@ -233,6 +233,7 @@ object Eliza extends Service {
     def sendGeneralPushNotification() = ServiceRoute(POST, "/internal/eliza/sendGeneralPushNotification")
     def sendLibraryPushNotification() = ServiceRoute(POST, "/internal/eliza/sendLibraryPushNotification")
     def sendUserPushNotification() = ServiceRoute(POST, "/internal/eliza/sendUserPushNotification")
+    def sendOrgPushNotification() = ServiceRoute(POST, "/internal/eliza/sendOrgPushNotification")
```
andrewconner, line 236: This route does not exist, has been broken. @camhashemi

kifi-backend/modules/common/app/com/keepit/eliza/ElizaServiceClient.scala
```diff
@@ -52,6 +56,16 @@ object LibraryPushNotificationCategory {
   val LibraryInvitation = LibraryPushNotificationCategory("LibraryInvitation")
 }
 
+object OrgPushNotificationCategory {
+  val OrganizationInvitation = OrgPushNotificationCategory("OrganizationInvitation")
+  implicit val format = new Format[OrgPushNotificationCategory] {
+    def reads(json: JsValue) = json.as[String] match {
+      case "OrganizationInvitation" => JsSuccess(OrgPushNotificationCategory("OrganizationInvitation"))
```
andrewconner, line 63: `OrganizationInvitation.name`

PR #17523, opened 2015-10-02
Feature/settings

cvializ:
@andrewconner ( @camhashemi ) remove the Kifi-admin experiment guard on team-settings. The incomplete settings tabs (like Plan Settings, billing, etc) are still Kifi-admin only. Add the Angular 404 for non-admins trying to access the team settings page.

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/teamSettings/teamSettings.js
+++ b/kifi-backend/modules/shoebox/angular/src/teamSettings/teamSettings.js
@@ -3,10 +3,10 @@
 angular.module('kifi')
 
 .controller('TeamSettingsCtrl', [
-  '$window', '$scope', 'orgProfileService', 'profileService', 'billingService',
-  'messageTicker', 'ORG_SETTING_VALUE',
-  function ($window, $scope, orgProfileService, profileService, billingService,
-            messageTicker, ORG_SETTING_VALUE) {
+  '$window', '$rootScope', '$scope', 'orgProfileService', 'profileService',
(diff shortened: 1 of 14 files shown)


kifi-backend/modules/shoebox/angular/src/teamSettings/teamSettings.js
```diff
@@ -166,7 +174,7 @@ angular.module('kifi')
       return message;
     }
 
-    if ($scope.viewer.membership.role === 'admin') {
+    if ($scope.viewer && $scope.viewer.membership && $scope.viewer.membership.role === 'admin') {
```
camhashemi: nice catch. is `scope.viewer` ever not there?

camhashemi:
:+1: I believe the 404 was the biggest hole in the whole thing, so we should be ready.

camhashemi:
If we really wanted to be safe, we would deploy the new 404, test, then release the experiment. But it'd be rare to have `kifi.com/kifi/settings` without being a part of the org.

kifi-backend/modules/shoebox/angular/src/teamSettings/teamSettings.js
cvializ: yeah for some reason it wasn't there when I wasn't a member of the org

camhashemi: that's bad, but good to know! I'll test it out

PR #17524, opened 2015-10-02
Revert "org invites are valid even after the inviter leaves or loses …

ryanpbrewster:
Reverts kifi/keepit#17513


ryanpbrewster:
@camhashemi

PR #17527, opened 2015-10-02
propagate create_slack_integration setting to library permissions

camhashemi:
@ryanpbrewster @leogrim

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/PermissionCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/PermissionCommander.scala
@@ -119,7 +119,9 @@ class PermissionCommanderImpl @Inject() (
   def combineOrganizationAndLibraryPermissions(lib: Library, libPermissions: Set[LibraryPermission], orgPermissions: Set[OrganizationPermission]): Set[LibraryPermission] = {
     val addedPermissions = Map[Boolean, Set[LibraryPermission]](
       (lib.canBeModified && libPermissions.contains(LibraryPermission.VIEW_LIBRARY) && orgPermissions.contains(OrganizationPermission.FORCE_EDIT_LIBRARIES)) ->
-        Set(LibraryPermission.EDIT_LIBRARY, LibraryPermission.MOVE_LIBRARY)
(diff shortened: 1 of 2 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/PermissionCommander.scala
```diff
@@ -119,7 +119,9 @@ class PermissionCommanderImpl @Inject() (
   def combineOrganizationAndLibraryPermissions(lib: Library, libPermissions: Set[LibraryPermission], orgPermissions: Set[OrganizationPermission]): Set[LibraryPermission] = {
     val addedPermissions = Map[Boolean, Set[LibraryPermission]](
       (lib.canBeModified && libPermissions.contains(LibraryPermission.VIEW_LIBRARY) && orgPermissions.contains(OrganizationPermission.FORCE_EDIT_LIBRARIES)) ->
-        Set(LibraryPermission.EDIT_LIBRARY, LibraryPermission.MOVE_LIBRARY)
+        Set(LibraryPermission.EDIT_LIBRARY, LibraryPermission.MOVE_LIBRARY, LibraryPermission.DELETE_LIBRARY),
```
ryanpbrewster, line 122: Are we sure that they can actually delete the library?

ryanpbrewster:
lgtm

Nice test, too. Good call.

kifi-backend/modules/shoebox/app/com/keepit/commanders/PermissionCommander.scala
camhashemi, line 122: asked @JRuff42 , she says the `edit library` feature lets them to whatever an owner can

ryanpbrewster, line 122: `LibraryCommander.deleteLibrary` actually checks `lib.ownerId == requesterId`, so we should either change that or change this.

camhashemi, line 122: I'll change it there. Trying to keep our logic to spec to reduce confusion between us and product.

PR #17530, opened 2015-10-02
disabled and admin should not be valid values for view_member

cvializ:
@camhashemi @ryanpbrewster Jen says these were the spec. Not sure if you wanted to adjust or not, but here's a heads up


camhashemi:
definitely. I'll go in and sync everything up later.

PR #17533, opened 2015-10-03
Always compile

andrewconner:


cvializ:
always be compiling
always be closing

PR #17535, opened 2015-10-03
Feature/extension settings

cvializ:
@andrewconner


cvializ:
@JRuff42 @andrewconner 
screen shot 2015-10-05 at 11 14 54 am
screen shot 2015-10-05 at 11 14 43 am

cvializ:
Got rid of the hover color when the option is disabled

JRuff42:
Nice!  Ever so slight wording enhancements:

On #1 - Change disable to "Not applicable"

On #2 "Kyfy has disable creation of publicly visible libraries."  (Just to
be clear that the company may still have some public ones).

PR #17540, opened 2015-10-05
add routes for /settings

camhashemi:
@leogrim @cvializ


kifi-backend/modules/shoebox/conf/shoeboxService.routes
```diff
@@ -465,9 +465,6 @@ GET     /site/libraries/:ids/images         @com.keepit.controllers.website.Libr
 GET     /site/libraries/recos/top           @com.keepit.controllers.website.LibraryRecommendationsController.topLibRecos()
 POST    /site/libraries/recos/feedback      @com.keepit.controllers.website.LibraryRecommendationsController.updateLibraryRecommendationFeedback(id: PublicId[Library])
 
-GET     /site/recos/public          @com.keepit.controllers.website.RecommendationsController.topPublicRecos()
-POST    /site/recos/feedback        @com.keepit.controllers.website.RecommendationsController.updateUriRecommendationFeedback(id: ExternalId[NormalizedURI])
```
camhashemi, line 469: @cvializ found these in `routes.js`, but didn't find any usages. should be okay

cvializ, line 469: :+1: yep we banished recos from the frontend code

leogrim:
:+1:

PR #17541, opened 2015-10-05
Sliiiight background colour for transparent keep images, better (!?) aspect ratio

andrewconner:
@cvializ


cvializ:
:panda_face: :clock12:

PR #17542, opened 2015-10-05
add elizaServiceClient.flush()

camhashemi:
@andrewconner


andrewconner:
Word

kifi-backend/modules/common/app/com/keepit/eliza/ElizaServiceClient.scala
```diff
@@ -190,6 +186,11 @@ class ElizaServiceClientImpl @Inject() (
     broadcast(Eliza.internal.sendToAllUsers, data)
   }
 
+  def flush(userId: Id[User]): Future[Unit] = {
```
andrewconner, line 189: Thanks for returning `Future[Unit]`

andrewconner:
Merging because I want this

kifi-backend/modules/common/app/com/keepit/eliza/ElizaServiceClient.scala
camhashemi, line 189: That's just to let us know when the future completes right?

PR #17545, opened 2015-10-05
add orgConfig to OrganizationInfo (for /user/me)

camhashemi:
@ryanpbrewster @andrewconner 

eliminates the need to call `GET /featureSettings` for each org in `/user/me`

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryInfoCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryInfoCommander.scala
@@ -228,7 +228,7 @@ class LibraryInfoCommanderImpl @Inject() (
       val allOrgsShown = libraries.flatMap { library => library.organizationId }.toSet
       db.readOnlyReplicaAsync { implicit s =>
         val orgMap = orgRepo.getByIds(allOrgsShown)
-        organizationCommander.getBasicOrganizationViews(orgMap.keys.toSet, viewerUserIdOpt, authTokenOpt = None)
+        orgMap.keys.map(orgId => orgId -> organizationCommander.getBasicOrganizationViewHelper(orgId, viewerUserIdOpt, authTokenOpt = None)).toMap
       }
     }
 
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
@@ -30,6 +30,8 @@ trait OrganizationCommander {
   def getOrganizationInfo(orgId: Id[Organization], viewerIdOpt: Option[Id[User]])(implicit session: RSession): OrganizationInfo
   def getOrganizationInfos(orgIds: Set[Id[Organization]], viewerIdOpt: Option[Id[User]]): Map[Id[Organization], OrganizationInfo]
   def getAccountFeatureSettings(orgId: Id[Organization]): OrganizationSettingsResponse
+  def getExternalOrgConfiguration(orgId: Id[Organization]): ExternalOrganizationConfiguration
+  def getExternalOrgConfigurationHelper(orgId: Id[Organization])(implicit session: RSession): ExternalOrganizationConfiguration
   def getBasicOrganizations(orgIds: Set[Id[Organization]]): Map[Id[Organization], BasicOrganization]
   def getBasicOrganization(orgId: Id[Organization])(implicit session: RSession): BasicOrganization
   def getOrganizationLibrariesVisibleToUser(orgId: Id[Organization], userIdOpt: Option[Id[User]], offset: Offset, limit: Limit): Seq[LibraryCardInfo]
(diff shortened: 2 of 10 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
```diff
@@ -184,7 +197,8 @@ class OrganizationCommanderImpl @Inject() (
       avatarPath = avatarPath,
       members = membersAsBasicUsers,
       numMembers = memberCount,
-      numLibraries = numLibraries)
+      numLibraries = numLibraries,
+      orgConfig = orgConfig)
```
leogrim: No need for a prefix when there's no ambiguity, let's just go for `config`?

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
```diff
@@ -147,6 +149,17 @@ class OrganizationCommanderImpl @Inject() (
     }
   }
 
+  def getExternalOrgConfiguration(orgId: Id[Organization]): ExternalOrganizationConfiguration = {
+    db.readOnlyReplica(implicit session => getExternalOrgConfigurationHelper(orgId))
+  }
```
ryanpbrewster, line 155: I've learned to be wary of these methods. Is there anywhere that we actually use this method? It opens a db session and it's startlingly easy to get session-in-session issues. IMO, most places are better off opening their own db session if they need the helper method below.

ryanpbrewster, line 155: Ah, I see that it is. :+1:

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
```diff
@@ -147,6 +149,17 @@ class OrganizationCommanderImpl @Inject() (
...
+
+  def getExternalOrgConfigurationHelper(orgId: Id[Organization])(implicit session: RSession): ExternalOrganizationConfiguration = {
+    val config = orgConfigRepo.getByOrgId(orgId)
+    val plan = paidPlanRepo.get(paidAccountRepo.getByOrgId(orgId).planId)
```
ryanpbrewster: Apparently we are not supposed to be touching `paidAccountRepo`. `PlanManagementCommander.currentPlan` does the right thing here. I need to go back and fix some of the other code I wrote as well.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
ryanpbrewster: :+1: I'm with Léo

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
camhashemi: ah right, to keep it private to the `payments` package. nice catch.

kifi-backend/modules/shoebox/app/com/keepit/model/OrganizationInfo.scala
```diff
@@ -34,21 +35,9 @@ object OrganizationInfo {
...
-    (__ \ 'ownerId).read[ExternalId[User]] and
-    (__ \ 'handle).read[OrganizationHandle] and
-    (__ \ 'name).read[String] and
-    (__ \ 'description).readNullable[String] and
-    (__ \ 'site).readNullable[String] and
-    (__ \ 'avatarPath).read[ImagePath] and
-    (__ \ 'members).read[Seq[BasicUser]] and
-    (__ \ 'numMembers).read[Int] and
-    (__ \ 'numLibraries).read[Int]
-  )(OrganizationInfo.apply _)
```
ryanpbrewster, line 51: :+1: I think it's fine to manually test fields

ryanpbrewster:
lgtm

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationCommander.scala
camhashemi, line 155: I see what you mean. Depends if we want to break the "controllers shouldn't open sessions" pattern.

You see that it is what? Not sure what the :+1: is for

ryanpbrewster, line 155: It is used elsewhere. I think it's fine as-is.

ryanpbrewster, line 155: As an aside, I think the "controllers shouldn't open sessions" rule is lower priority than the "sessions should not be opened inside of other sessions" rule. The controllers rule was meant to prevent controllers from opening a single db session that would last for thousands of ms (basically deferring that to multiple smaller sessions inside the commander). In this case, the db session would explicitly only be used for that one request.

camhashemi, line 155: I just created it and used it elsewhere, so I can remove it. But I agree that it's fine. With a combination of uncommenting the session-in-session detector in `Database.enteringSession` every once in a while, and having a `Helper` method available for when you already have a session, commander methods that create sessions shouldn't be a problem.

kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryInfoCommander.scala
```diff
@@ -228,7 +228,7 @@ class LibraryInfoCommanderImpl @Inject() (
       val allOrgsShown = libraries.flatMap { library => library.organizationId }.toSet
       db.readOnlyReplicaAsync { implicit s =>
         val orgMap = orgRepo.getByIds(allOrgsShown)
-        organizationCommander.getBasicOrganizationViews(orgMap.keys.toSet, viewerUserIdOpt, authTokenOpt = None)
+        orgMap.keys.map(orgId => orgId -> organizationCommander.getBasicOrganizationViewHelper(orgId, viewerUserIdOpt, authTokenOpt = None)).toMap
```
ryanpbrewster, line 231: Why is this happening? This seems super inefficient.

camhashemi, line 231: session-in-session. we can move the same logic to `organizationCommander` if we want

ryanpbrewster, line 231: We should probably have a method like `getBasicOrganizationViewsHelper` that does the batch processing.

ryanpbrewster, line 231: Oh, I see, we have never had that behavior. I'll put it in my queue.

PR #17560, opened 2015-10-05
Org invite validation

camhashemi:
@ryanpbrewster let's try this again

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
@@ -192,7 +192,7 @@ class OrganizationMembershipCommanderImpl @Inject() (
     val targetOpt = organizationMembershipRepo.getByOrgIdAndUserId(request.orgId, request.targetId)
 
     (requesterOpt, request) match {
-      case (_, OrganizationMembershipAddRequest(_, _, _, newRole)) =>
+      case (_, _: OrganizationMembershipAddRequest) =>
         if (targetOpt.isDefined) Some(OrganizationFail.ALREADY_A_MEMBER)
         else None
 
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
@@ -191,7 +191,6 @@ class OrganizationMembershipCommanderImpl @Inject() (
     val requesterPermissions = permissionCommander.getOrganizationPermissions(request.orgId, Some(request.requesterId))
     val targetOpt = organizationMembershipRepo.getByOrgIdAndUserId(request.orgId, request.targetId)
 
-<<<<<<< HEAD
     requesterOpt match {
       case None => Some(OrganizationFail.NOT_A_MEMBER)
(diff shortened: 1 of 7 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
```diff
@@ -191,29 +191,24 @@ class OrganizationMembershipCommanderImpl @Inject() (
...
-
-          case OrganizationMembershipRemoveRequest(_, _, _) =>
-            val requesterIsOwner = (requester.userId == org.ownerId) && !targetOpt.exists(_.userId == org.ownerId)
-            val requesterRemovingSelf = targetOpt.exists(t => t.userId == requester.userId && t.userId != org.ownerId)
-            val requesterOutranksTarget = targetOpt.exists(_.role < requester.role) && requesterPermissions.contains(REMOVE_MEMBERS)
-            if (!(requesterIsOwner || requesterRemovingSelf || requesterOutranksTarget)) Some(OrganizationFail.INSUFFICIENT_PERMISSIONS)
-            else None
-        }
+    (requesterOpt, request) match {
+      case (_, OrganizationMembershipAddRequest(_, _, _, newRole)) =>
```
ryanpbrewster: this can probably be `OrganizationMembershipAddRequest(_, _, _, _)`

ryanpbrewster: or even `case (_, _: OrganizationMembershipAddRequest) =>`

kifi-backend/modules/shoebox/test/com/keepit/controllers/mobile/MobileOrganizationInviteControllerTest.scala
```diff
@@ -92,7 +91,28 @@ class MobileOrganizationInviteControllerTest extends Specification with ShoeboxT
...
+            val org = OrganizationFactory.organization().withOwner(owner).withWeakMembers().saved
+
+            val invite = inject[OrganizationInviteRepo].save(OrganizationInvite(organizationId = org.id.get, inviterId = nonMember.id.get, userId = invitee.id, role = OrganizationRole.MEMBER))
+            (invitee, invite)
+          }
+
+          inject[FakeUserActionsHelper].setUser(invitee)
+          val request = route.acceptInvitation(publicOrgId, Some(invite.authToken))
+          val result = controller.acceptInvitation(publicOrgId, Some(invite.authToken))(request)
+          status(result) must equalTo(NO_CONTENT)
```
ryanpbrewster, line 115: I prefer to test that the db ends up in the correct state in addition to checking that the response is correct.

ryanpbrewster:
lgtm. Scary, but good.

kifi-backend/modules/shoebox/app/com/keepit/commanders/OrganizationMembershipCommander.scala
camhashemi: ah, right

kifi-backend/modules/shoebox/test/com/keepit/controllers/mobile/MobileOrganizationInviteControllerTest.scala
camhashemi, line 115: :+1:

PR #17564, opened 2015-10-06
tracking fixes

camhashemi:


kifi-backend/modules/shoebox/angular/src/orgProfile/orgProfileMemberManage.js
```diff
@@ -259,8 +259,10 @@ angular.module('kifi')
                 });
               }
             });
-
-            memberPageAnalytics({ action: 'clickedInvite', orgInvitees: invitees });
+            var inviteeIds = invitees.map(function(invitee) {
+                return invitee.id;
+             });
+            memberPageAnalytics({ action: 'clickedInvite', orgInvitees: inviteeIds });
```
camhashemi, line 265: @cvializ this should fix https://fortytwo.airbrake.io/projects/91268/groups/1522038547434457718/notices/1524900661782138606. The tracking deserializer expects values to be either a primitive or an array of primitives. In this case invitees are objects with one field, so a quick fix.

kifi-backend/modules/eliza/app/com/keepit/eliza/commanders/MessagingCommander.scala
```diff
@@ -654,11 +653,12 @@ class MessagingCommander @Inject() (
 
     val orgIds = validOrgRecipients.map(o => Organization.decodePublicId(o)).filter(_.isSuccess).map(_.get)
 
-    val canSendToOrgs = shoebox.getUserPermissionsByOrgId(orgIds.toSet, userId).map { permissionsByOrgId =>
-      !permissionsByOrgId.exists {
+    val cantSendToOrgs = shoebox.getUserPermissionsByOrgId(orgIds.toSet, userId).map { permissionsByOrgId =>
+      permissionsByOrgId.exists {
         case (orgId, permissions) =>
-          !permissions.contains(OrganizationPermission.GROUP_MESSAGING)
-            .tap { _ => airbrake.notify(s"user $userId was able to send to org $orgId without permissions!") }
```
camhashemi, line 661: @ryanpbrewster should've done `.tap { if (_) airbrake.notify }`... fail

ryanpbrewster, line 661: I'm not sure I see the use for `.tap` here. Shouldn't it just be

```
if (!permissions.contains(OrganizationPermission.GROUP_MESSAGING) {
    airbrake.notify(...)
}
```

ryanpbrewster, line 661: But yeah, the number of times I mess up when using `_` is absurd. I almost always use `(x => x)` now, if for no other reason than because IntelliJ will do type inference for me.

camhashemi, line 661: I wanted to create the boolean once, act upon it, and return it. All in once swoop.

PR #17568, opened 2015-10-06
Feature/settings

cvializ:
@andrewconner ( @camhashemi this removes the multiple requests to featureSettings)


andrewconner:
Wahoo :+1:

PR #17569, opened 2015-10-06
Fixes bad suggested libraries from appearing

andrewconner:
@leogrim

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/RelatedLibraryCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/RelatedLibraryCommander.scala
@@ -15,6 +15,7 @@ import play.api.libs.concurrent.Execution.Implicits._
 import scala.concurrent.duration._
 import scala.collection.mutable
 import com.keepit.common.cache.TransactionalCaching.Implicits.directCacheAccess
+import com.keepit.common.core._
 
 import scala.concurrent.Future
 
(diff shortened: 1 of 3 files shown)


kifi-backend/modules/shoebox/app/com/keepit/commanders/RelatedLibraryCommander.scala
```diff
@@ -49,45 +49,46 @@ class RelatedLibraryCommanderImpl @Inject() (
...
-    val userLibs: Set[Id[Library]] = userIdOpt match {
-      case Some(userId) => db.readOnlyReplica { implicit s => libMemRepo.getWithUserId(userId) }.map { _.libraryId }.toSet
-      case None => Set()
+    val library = db.readOnlyReplica { implicit session =>
+      libraryRepo.get(libId)
     }
 
-    def isUnwantedLibrary(relatedLib: RelatedLibrary): Boolean = {
-      val lib = relatedLib.library
+    if (library.state == LibraryStates.INACTIVE || library.visibility == LibraryVisibility.SECRET || library.visibility == LibraryVisibility.ORGANIZATION || library.keepCount == 0) {
```
andrewconner, line 57: Check these rules. Sound good?

kifi-backend/modules/shoebox/app/com/keepit/commanders/RelatedLibraryCommander.scala
```diff
@@ -148,6 +140,13 @@ class RelatedLibraryCommanderImpl @Inject() (
     }.map { _.libs }
   }
 
+  private lazy val fakeUsers = userCommander.getAllFakeUsers()
```
andrewconner, line 143: Caches for the app duration. I think this is reasonable because new fake accounts likely won't be recommended for a while, right?

leogrim, line 143: :+1:

kifi-backend/modules/shoebox/app/com/keepit/commanders/RelatedLibraryCommander.scala
leogrim, line 57: :+1:

ryanpbrewster, line 57: Can we replace `library.state == LibraryStates.INACTIVE` with `library.isInactive`?

PR #17570, opened 2015-10-06
sort account activit in reverse chron (admin)

camhashemi:
@stkem @ryanpbrewster


kifi-backend/modules/shoebox/app/com/keepit/admin/AdminPaymentsController.scala
```diff
@@ -166,7 +166,7 @@ class AdminPaymentsController @Inject() (
     val PAGE_SIZE = 50
     val (allEvents, org) = db.readOnlyMaster { implicit s =>
       val account = paidAccountRepo.getByOrgId(orgId)
-      val allEvents = accountEventRepo.getByAccountAndState(account.id.get, AccountEventStates.ACTIVE)
+      val allEvents = accountEventRepo.getByAccountAndState(account.id.get, AccountEventStates.ACTIVE).sortBy(-1 * _.eventTime.getMillis)
```
stkem, line 169: Why not do this in the db query?

camhashemi, line 169: :+1: will do

PR #17573, opened 2015-10-06
Added a cache from (orgId, userId) => permissions

ryanpbrewster:
Kind of tricky to invalidate the cache for only one particular org. When an
org's settings change, we need to invalidate `(orgId, *)`. I implemented a
two-cache strategy.
  The first cache is `Id[Organization] => Int`
  The second cache is `(Id[Organization], Int, Id[User]) => Set[OrganizationPermission]`

To go from `(Id[Organization], Id[User]) => Set[OrganizationPermission]`, first
hit cache #1 to get the `Int` for that orgId, then use that "namespace key" to hit
cache #2.

To invalidate an entire org, just drop the orgId key from cache #1

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/PermissionCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/PermissionCommander.scala
@@ -7,6 +7,7 @@ import com.keepit.common.db.slick.DBSession.RSession
 import com.keepit.common.db.slick.Database
 import com.keepit.common.healthcheck.AirbrakeNotifier
 import com.keepit.common.logging.{ AccessLog, Logging }
+import com.keepit.common.performance.StatsdTiming
 import com.keepit.model._
 import play.api.libs.functional.syntax._
 import play.api.libs.json._
@@ -39,6 +40,7 @@ class PermissionCommanderImpl @Inject() (
     airbrake: AirbrakeNotifier,
     implicit val executionContext: ExecutionContext) extends PermissionCommander with Logging {
 
+  @StatsdTiming("PermissionCommander.getOrganizationPermissions")
   def getOrganizationPermissions(orgId: Id[Organization], userIdOpt: Option[Id[User]])(implicit session: RSession): Set[OrganizationPermission] = {
     /*
     PSA regarding these caches:
@@ -70,6 +72,7 @@ class PermissionCommanderImpl @Inject() (
     }
   }
 
+  @StatsdTiming("PermissionCommander.getLibraryPermissions")
   def getLibraryPermissions(libId: Id[Library], userIdOpt: Option[Id[User]])(implicit session: RSession): Set[LibraryPermission] = {
     val lib = libraryRepo.get(libId)
     val libMembershipOpt = userIdOpt.flatMap { userId => libraryMembershipRepo.getWithLibraryIdAndUserId(libId, userId) }
--- a/kifi-backend/modules/shoebox/app/com/keepit/commanders/PermissionCommander.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/commanders/PermissionCommander.scala
@@ -1,7 +1,7 @@
 package com.keepit.commanders
 
 import com.google.inject.{ ImplementedBy, Inject, Singleton }
-import com.keepit.common.cache.{ JsonCacheImpl, FortyTwoCachePlugin, CacheStatistics, Key }
+import com.keepit.common.cache._
 import com.keepit.common.db.Id
 import com.keepit.common.db.slick.DBSession.RSession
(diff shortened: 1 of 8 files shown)


ryanpbrewster:
@camhashemi @andrewconner @eishay

kifi-backend/modules/shoebox/app/com/keepit/commanders/PermissionCommander.scala
```diff
@@ -1,14 +1,19 @@
 package com.keepit.commanders
 
 import com.google.inject.{ ImplementedBy, Inject, Singleton }
+import com.keepit.common.cache.{ JsonCacheImpl, FortyTwoCachePlugin, CacheStatistics, Key }
 import com.keepit.common.db.Id
 import com.keepit.common.db.slick.DBSession.RSession
 import com.keepit.common.db.slick.Database
 import com.keepit.common.healthcheck.AirbrakeNotifier
-import com.keepit.common.logging.Logging
+import com.keepit.common.logging.{ AccessLog, Logging }
```
ryanpbrewster, line 9: don't know why this is here, maybe unneeded?

kifi-backend/modules/shoebox/app/com/keepit/commanders/PermissionCommander.scala
```diff
@@ -33,9 +40,10 @@ class PermissionCommanderImpl @Inject() (
     implicit val executionContext: ExecutionContext) extends PermissionCommander with Logging {
 
   def getOrganizationPermissions(orgId: Id[Organization], userIdOpt: Option[Id[User]])(implicit session: RSession): Set[OrganizationPermission] = {
-    // TODO(ryan): this needs to look in orgPermissionsCache by (orgId, userIdOpt) to see if the permissions are cached
-    // If not, it should compute them directly and add them to the cache
-    computeOrganizationPermissions(orgId, userIdOpt)
+    val orgPermissionsNamespace = orgPermissionsNamespaceCache.getOrElse(OrganizationPermissionsNamespaceKey(orgId))(OrganizationPermissionsNamespace(orgId, Random.nextInt()))
+    orgPermissionsCache.getOrElse(OrganizationPermissionsKey(orgPermissionsNamespace, userIdOpt)) {
+      computeOrganizationPermissions(orgId, userIdOpt)
+    }
```
ryanpbrewster, line 57: I considered adding a comment here for clarity. Thoughts?

andrewconner:
@leogrim

kifi-backend/modules/shoebox/app/com/keepit/commanders/PermissionCommander.scala
camhashemi, line 9: used by the cache

kifi-backend/modules/shoebox/app/com/keepit/commanders/PermissionCommander.scala
```diff
@@ -145,5 +153,28 @@ class PermissionCommanderImpl @Inject() (
       OrganizationPermission.MANAGE_PLAN
     )
   }
+}
+
+case class OrganizationPermissionsNamespace(orgId: Id[Organization], value: Int)
```
camhashemi: adding the PR description here would be helpful to somebody seeing this for the first time

kifi-backend/modules/shoebox/app/com/keepit/commanders/PermissionCommander.scala
```diff
@@ -145,5 +153,28 @@ class PermissionCommanderImpl @Inject() (
...
+    (__ \ 'value).format[Int]
+  )(OrganizationPermissionsNamespace.apply, unlift(OrganizationPermissionsNamespace.unapply))
+}
+case class OrganizationPermissionsNamespaceKey(orgId: Id[Organization]) extends Key[OrganizationPermissionsNamespace] {
+  override val version = 1
+  val namespace = "org_permissions_namespace"
+  def toKey(): String = orgId.id.toString
+}
+class OrganizationPermissionsNamespaceCache(stats: CacheStatistics, accessLog: AccessLog, innermostPluginSettings: (FortyTwoCachePlugin, Duration), innerToOuterPluginSettings: (FortyTwoCachePlugin, Duration)*)
+  extends JsonCacheImpl[OrganizationPermissionsNamespaceKey, OrganizationPermissionsNamespace](stats, accessLog, innermostPluginSettings, innerToOuterPluginSettings: _*)
```
camhashemi: could we just store the `value` as the value and prepend the `orgId` outside of the cache hit? that way we can use a `PrimitiveCacheImpl[OrganizationPermissionsNameSpaceKey, Int]`, which should have better performance.

camhashemi:
looks great! we went from having denormalized permissions in three db columns to having one (well, two) caches for everything.

PR #17577, opened 2015-10-06
Added deep links to password reset emails

ryanpbrewster:
Made the DeepLinkRouter a little safer, and it now produces deep links (as well
as routing them). Only the Password Reset emails use this functionality, but
eventually all of the deep links should be produced there.

@leogrim @jaredpetker


ryanpbrewster:
I don't think we really want to merge this PR.

PR #17581, opened 2015-10-06
add memos to add/change card + charge account events

camhashemi:
@stkem @ryanpbrewster per https://app.asana.com/0/32728208409723/53377771713323

### DIFF
--- a/kifi-backend/modules/shoebox/app/com/keepit/admin/AdminPaymentsController.scala
+++ b/kifi-backend/modules/shoebox/app/com/keepit/admin/AdminPaymentsController.scala
@@ -137,12 +137,11 @@ class AdminPaymentsController @Inject() (
       data("cvc").head.trim,
       data("cardholderName").head.trim
     )
-    stripeClient.getPermanentToken(cardDetails, s"Manually Entered through admin ${request.userId} for org $orgId").flatMap { token =>
-      stripeClient.getLastFourDigitsOfCard(token).map { lastFour =>
-        val pm = planCommander.addPaymentMethod(orgId, token, ActionAttribution(user = None, admin = Some(request.userId)), lastFour)
-        val event = planCommander.changeDefaultPaymentMethod(orgId, pm.id.get, ActionAttribution(user = None, admin = Some(request.userId)), lastFour)
-        Ok(event.toString)
-      }
+    stripeClient.getPermanentToken(cardDetails, s"Manually Entered through admin ${request.userId} for org $orgId").map { token =>
+      val lastFour = cardDetails.number.takeRight(4)
+      val pm = planCommander.addPaymentMethod(orgId, token, ActionAttribution(user = None, admin = Some(request.userId)), lastFour)
+      val event = planCommander.changeDefaultPaymentMethod(orgId, pm.id.get, ActionAttribution(user = None, admin = Some(request.userId)), lastFour)
+      Ok(event.toString)
     }
   }
 
@@ -168,7 +167,7 @@ class AdminPaymentsController @Inject() (
     val PAGE_SIZE = 50
     val (allEvents, org) = db.readOnlyMaster { implicit s =>
       val account = paidAccountRepo.getByOrgId(orgId)
-      val allEvents = accountEventRepo.getByAccountAndState(account.id.get, AccountEventStates.ACTIVE).sortBy(-1 * _.eventTime.getMillis)
+      val allEvents = accountEventRepo.getByAccountAndState(account.id.get, AccountEventStates.ACTIVE)
       val org = organizationRepo.get(orgId)
       (allEvents, org)
     }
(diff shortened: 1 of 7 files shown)


kifi-backend/modules/shoebox/app/com/keepit/admin/AdminPaymentsController.scala
```diff
@@ -137,10 +137,12 @@ class AdminPaymentsController @Inject() (
       data("cvc").head.trim,
       data("cardholderName").head.trim
     )
-    stripeClient.getPermanentToken(cardDetails, s"Manually Entered through admin ${request.userId} for org $orgId").map { token =>
-      val pm = planCommander.addPaymentMethod(orgId, token, ActionAttribution(user = None, admin = Some(request.userId)))
-      val event = planCommander.changeDefaultPaymentMethod(orgId, pm.id.get, ActionAttribution(user = None, admin = Some(request.userId)))
-      Ok(event.toString)
+    stripeClient.getPermanentToken(cardDetails, s"Manually Entered through admin ${request.userId} for org $orgId").flatMap { token =>
+      stripeClient.getLastFourDigitsOfCard(token).map { lastFour =>
```
stkem: Why make that extra call? You have the whole credit card number in `cardDetails`.

stkem:
High level, I'm not sure the memo is the right place for this. If this is a permanent (and uniform) part of this type of event it should go into the action.

camhashemi:
That's fine, it just seemed like memo was a good place for it. "Welcome to Kifi" is the only other memo, also uniform. What do you intend it to be for?

kifi-backend/modules/shoebox/app/com/keepit/admin/AdminPaymentsController.scala
camhashemi: :+1: didn't see that there

stkem:
"Welcome to Kifi" is not uniform. It's not on all `special_credit` events, just on the ones coming with signup. The memo really is intended for extra information for a specific event or a group of events, not every event of a particular type. In your case the last four digits would be on every `DefaultPaymentMethodChanged` event, e.g.

camhashemi:
ah, got it. seems like it should go in `eventTypeExtras` then?

stkem:
Yup, at the db level. It should be a field on the `action` in scala land.

kifi-backend/modules/shoebox/app/com/keepit/controllers/website/PaymentsController.scala
```diff
@@ -56,11 +56,13 @@ class PaymentsController @Inject() (
   def setCreditCardToken(pubId: PublicId[Organization]) = OrganizationUserAction(pubId, OrganizationPermission.MANAGE_PLAN).async(parse.tolerantJson) { request =>
     (request.body \ "token").asOpt[String] match {
       case Some(token) => {
-        stripeClient.getPermanentToken(token, s"Card for Org ${request.orgId} added by user ${request.request.userId} with admin ${request.request.adminUserId}").map { realToken =>
-          val attribution = ActionAttribution(user = Some(request.request.userId), admin = request.request.adminUserId)
-          val pm = planCommander.addPaymentMethod(request.orgId, realToken, attribution)
-          planCommander.changeDefaultPaymentMethod(request.orgId, pm.id.get, attribution)
-          Ok
+        stripeClient.getPermanentToken(token, s"Card for Org ${request.orgId} added by user ${request.request.userId} with admin ${request.request.adminUserId}").flatMap { realToken =>
+          stripeClient.getLastFourDigitsOfCard(realToken).map { lastFour =>
```
camhashemi, line 60: @stkem seems like we can easily create a new method that returns both `(permanentToken, lastFour)` since they come from the same model. worth it?

stkem, line 60: Not sure it's worth it. This is the only place that is used (and it would still be two calls to stripe).

kifi-backend/modules/shoebox/app/com/keepit/payments/AccountEventRepo.scala
```diff
@@ -91,7 +91,7 @@ class AccountEventRepoImpl @Inject() (
   }
 
   def getByAccountAndState(accountId: Id[PaidAccount], state: State[AccountEvent])(implicit session: RSession): Seq[AccountEvent] = {
-    (for (row <- rows if row.accountId === accountId && row.state === state) yield row).list
+    (for (row <- rows if row.accountId === accountId && row.state === state) yield row).sortBy(row => row.eventTime desc).list
```
camhashemi, line 94: @stkem per the admin page's pr

kifi-backend/modules/shoebox/app/com/keepit/payments/PlanManagementCommander.scala
```diff
@@ -564,10 +565,20 @@ class PlanManagementCommanderImpl @Inject() (
...
     }
+    val extraInfo = event.action match {
+      case PaymentMethodAdded(_, lastFour) => Some(s"Card ending in $lastFour added")
+      case DefaultPaymentMethodChanged(_, _, newLastFour) => Some(s"Changed to card ending in $newLastFour")
+      case PlanBillingCharge() => {
+        val chargeIdNotFoundText = s"not found, please contact ${SystemEmailAddress.BILLING}"
+        Some(s"Invoice # ${event.chargeId.getOrElse(chargeIdNotFoundText)}")
+      }
+      case _ => None
+    }
```
camhashemi, line 574: @stkem

PR #17593, opened 2015-10-07
Moved Eliza's MessageThread into common

ryanpbrewster:
It's time to link keeps and message threads

@leogrim


ryanpbrewster:
@andrewconner @stkem

andrewconner:
Why is this needed? If just for `Id` references, Léo has a more elegant solution for graph (it deals with typed `Id`s of models it can't see)

stkem:
What Andrew said

ryanpbrewster:
Okay, I'll grab Leo and make him reveal his secrets.

PR #17594, opened 2015-10-07
Add upsell for integration. Fix up integration markup/layout.

cvializ:
@andrewconner @JRuff42


cvializ:
Accidentally added code that wasn't ready

PR #17603, opened 2015-10-07
Turning live mode off

andrewconner:
@cvializ @stkem Eishay was a bit quick on the uptake, turning live mode off so we can build the frontend without fear :)


andrewconner:
@eishay fyi

cvializ:
:clap:

stkem:
:+1:

PR #17617, opened 2015-10-08
remove nested db sessions from OrganizationChecker

camhashemi:
@ryanpbrewster @andrewconner turns out `unsafeModifyLibrary` also had sessions-in-sessions. it used a batch query at some point, which I had to remove. but like we said, better to be slow than wrong.


kifi-backend/modules/shoebox/app/com/keepit/commanders/LibraryCommander.scala
```diff
@@ -314,41 +315,30 @@ class LibraryCommanderImpl @Inject() (
...
-        val keepIds = keepRepo.getByLibraryIdAndExcludingVisibility(lib.id.get, Some(viz), 500).map(_.id.get).toSet ++ keepRepo.getByLibraryWithInconsistentOrgId(lib.id.get, lib.organizationId, Limit(500))
-        val keeps = keepRepo.getByIds(keepIds).values.toSeq
-        (keeps, lib, viz)
-      }
-      if (keeps.nonEmpty && curViz == changedVisibility) {
-        db.readWriteBatch(keeps, attempts = 5) { (s, k) =>
-          implicit val session: RWSession = s
-          keepCommander.syncWithLibrary(k, lib)
-        }
+    def updateKeepVisibility(changedVisibility: LibraryVisibility, iter: Int)(implicit session: RWSession): Future[Unit] = Future {
```
ryanpbrewster, line 329: This is going to be hard to avoid. It opens a `Future` to avoid really long waiting times, but Futures in sessions is rough.

PR #17622, opened 2015-10-08
add accept/decline buttons to invited orgs (home page)

camhashemi:
@JRuff42 @cvializ @andrewconner thanks for the help. happy to make changes per the team's response.

screen shot 2015-10-08 at 4 26 16 pm

### DIFF
--- a/kifi-backend/modules/shoebox/angular/src/orgProfile/orgProfileHeader.js
+++ b/kifi-backend/modules/shoebox/angular/src/orgProfile/orgProfileHeader.js
@@ -138,7 +138,6 @@ angular.module('kifi')
               .acceptOrgMemberInvite(scope.profile.id, $location.search().authToken)
               .then(function () {
                 scope.acknowledgedInvite = true;
-                orgProfileService.invalidateOrgProfileCache();
                 $state.reload();
               });
           }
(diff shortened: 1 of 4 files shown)


kifi-backend/modules/shoebox/angular/src/orgProfile/orgProfileHeader.js
```diff
@@ -138,7 +138,6 @@ angular.module('kifi')
               .acceptOrgMemberInvite(scope.profile.id, $location.search().authToken)
               .then(function () {
                 scope.acknowledgedInvite = true;
-                orgProfileService.invalidateOrgProfileCache();
                 $state.reload();
```
cvializ, line 141: Did you intentionally remove this? I think it is used to fetch the new team and secret libraries that the user is able to view when they accept the invitation. If it works without this, it's fine.

camhashemi, line 141: I removed it because `orgProfileService.acceptOrgMemberInvite` calls `invalidateOrgProfileCache`. No need to do it twice, right? I can add it back if that was inaccurate.

cvializ, line 141: oh good catch! yes, the service should be handling `invalidateOrgProfileCache` here

cvializ:
:+1: epic

PR #17624, opened 2015-10-09
Blocking wildcards

andrewconner:
@eishay @leogrim


andrewconner:
Merge if that logic looks good.

kifi-backend/modules/common/app/com/keepit/common/controller/BlockingFilter.scala
```diff
@@ -8,7 +8,7 @@ import scala.concurrent.Future
 object BlockingFilter extends Filter {
   def apply(nextFilter: (RequestHeader) => Future[Result])(requestHeader: RequestHeader): Future[Result] = {
     val ip = requestHeader.headers.get("X-Forwarded-For").getOrElse(requestHeader.remoteAddress)
-    if (blocked.contains(ip) && !requestHeader.path.contains("logout") && requestHeader.session.get(KifiSession.FORTYTWO_USER_ID).isDefined) {
+    if (blocked.exists(b => ip.indexOf(b) == 0) && !requestHeader.path.contains("logout") && requestHeader.session.get(KifiSession.FORTYTWO_USER_ID).isDefined) {
```
leogrim: `.startsWith`?

andrewconner: You're right.

PR #17625, opened 2015-10-09
Stripe integration

cvializ:
@andrewconner

### DIFF
--- a/kifi-backend/modules/shoebox/angular/bower.json
+++ b/kifi-backend/modules/shoebox/angular/bower.json
@@ -24,6 +24,7 @@
     "angular-moment": "0.8.0",
     "angular-resource": "~1.3.15",
     "angular-sanitize": "~1.3.15",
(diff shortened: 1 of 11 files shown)


kifi-backend/modules/shoebox/angular/bower.json
```diff
@@ -24,6 +24,7 @@
     "angular-moment": "0.8.0",
     "angular-resource": "~1.3.15",
     "angular-sanitize": "~1.3.15",
+    "angular-stripe-checkout": "git://github.com/cvializ/angular-stripe-checkout.git#eaa3071ae936594fcdbb61615421e6724bf3619b",
```
cvializ, line 27: I opened a pull request in the original author's repo, but I don't want to wait for them.

https://github.com/tobyn/angular-stripe-checkout/pull/9

Not sure what we ought to do for this. Move my fork into the kifi space?

andrewconner:
:money_with_wings:

kifi-backend/modules/shoebox/angular/bower.json
andrewconner, line 27: Feel free to do that if you want, or leave it where it is (hopefully he/she merges it)

